Issue #447, opened 2014-12-13
Users x.y and x.y2 are both shown as "X Y" in the UI


karyon:
Currently, when there are two users with the same first name and last name, they will have different usernames (usually by appending a number to the newer user), but they will be shown as "Firstname Lastname" in all the lists in the EvaP UI, making them indistinguishable. 

@janno42 and I think the best solution in terms of code and UI simplicity is to add this number also to the last name of the respective user. This needs to be done on production.

The actual todo for this issue is to add a warning in the importer when a new user with the same first and last name as an existing user is about to be added.

WGierke:
Ich würde mich mal ransetzen wollen :)

PR #473, opened 2015-01-15
solved #447 warning for importer

WGierke:
This pull request should solve #447: if someone wants to import a student whose first and last name already exist in the database then a warning is being displayed.

### DIFF
--- a/evap/staff/importers.py
+++ b/evap/staff/importers.py
@@ -172,7 +172,11 @@ class ExcelImporter(object):
 
             try:
                 user_same_name = UserProfile.objects.get(first_name=user_data.first_name, last_name=user_data.last_name)
-                self.warnings.append(u"Warning: The user {} {} already exists.".format(user_same_name.first_name, user_same_name.last_name))
+                if user_same_name.username != user_data.username:
+                    self.warnings.append(_(u"Warning: The existing user") + 
+                            u" {} ({} {} {}, {}) ".format(user_same_name.username, user_same_name.title or "", user_same_name.first_name, user_same_name.last_name, user_same_name.email) +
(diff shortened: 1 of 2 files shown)


evap/staff/importers.py
```diff
@@ -165,6 +165,13 @@ def check_user_data_sanity(self):
             except UserProfile.DoesNotExist:
                 # nothing to do here
                 pass
+                
+            try:
+                user_same_name = UserProfile.objects.get(first_name=user_data.first_name, last_name=user_data.last_name)
+                self.warnings.append(u"Warning: The user {} {} already exists.".format(user_same_name.first_name, user_same_name.last_name))
```
karyon: please wrap the string in _("string goes here"), this tells django to translate it.

also, @janno42 and i discussed this and we think we need the full data of the existing and new user.

something like
"Warning: The existing user {} ({} {} {}, {}) has the same first and last name as the following user from the enrollment data: {} ({} {} {}, {})". you could copypaste that structure from a few lines above. if you have an idea how to not duplicate the code, please do that as well ;)

evap/staff/importers.py
```diff
@@ -140,6 +140,10 @@ def check_user_data_correctness(self):
                 self.errors.append(_(u'Emailaddress {}: Username cannot be empty for non-external users.').format(user_data.email))
                 return # to avoid duplicate errors with validate
             try:
+                user = UserProfile.objects.get(email=user_data.email)
```
WGierke: Don't exactly know why I added this :D

karyon:
rebased and added the additional check i had in the pipeline. moved to #497

karyon:
i forgot the tahnk you, so: thank you! :)

PR #477, opened 2015-01-15
solve #47

lmm-git:
ref #47

Issue #301, opened 2014-05-30
use continuous colors in the export


karyon:
i had the idea to use continuous/stepless colors in the export instead of green/yellow/orange. this would allow to make more accurate estimates of a courses' evaluation just by looking at the colors.

additionally, the current colors make no distinction beween a 2.1 and a 3.6 which is a HUGE difference imo.

Paul-Geppert:
Please Assign Me, Thanks!

PR #484, opened 2015-01-15
Fix for #301: Continuous grade color support to excel export

Paul-Geppert:
references #301 
relies on the changes of #477


karyon:
rebased and manually merged, thanks!

PR #485, opened 2015-01-15
Fix #346

### LINKED ISSUE #346: Search field for contributor selection
The current contributor selection is a simple dropdown list with all users. Therefore you can only search for people when starting to type their first name.
This should be changed so that the dropdown is replaced by an edit field that has a dropdown option which can be manually triggered and that opens automatically when starting to type and which then only holds names that contain the typed substring (like the multiselect input alread does).

lawrencebenson:
changed contributor selection from static drop down select to input
based drop down

### DIFF
--- a/evap/contributor/templates/contributor_course_form.html
+++ b/evap/contributor/templates/contributor_course_form.html
@@ -34,6 +34,7 @@
                 </thead>
                 <tbody>
                     {% for form_element in formset %}
+
                         {% if form_element.non_field_errors %}
                             <tr><td colspan=100>{{ form_element.non_field_errors }}</td></tr>
                         {% endif %}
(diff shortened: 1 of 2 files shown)


evap/contributor/templates/contributor_course_form.html
```diff
@@ -34,6 +34,7 @@
                 </thead>
                 <tbody>
                     {% for form_element in formset %}
+
```
Bartzi: could you remove this line?

evap/contributor/templates/contributor_course_form.html
```diff
@@ -46,11 +46,17 @@
...
                             <td>
                                 {{form_element.contributor.errors}}
+                                
                                 {% if form_element.responsible.value == True %}
-                                    <select class="form-control" disabled="True"><option selected="selected">{{ responsible }} ({% trans "responsible" %})</option></select>
+                                    <input value='{{ responsible }} ({% trans "responsible" %})' disabled='True'/>
                                     {{ form_element.contributor.as_hidden }}
                                 {% else %}
-                                    {{ form_element.contributor }}
+                                    <input list="contributor_list" style="width:80%"/>
```
Bartzi: could you please add the css class form-control to each input?

Bartzi:
* could you have a look at the tests and why they fail?
* could you also add the functionality you implemented to the "course_edit" view in "/staff/views.py"?
you can find it here: 
https://github.com/fsr-itse/EvaP/blob/master/evap/staff/views.py#L307

karyon:
will be superseded by select2/chosen in #529.

Issue #249, opened 2014-03-03
redirect to correct tab when reviewing comments


karyon:
when reviewing comments on `/fsr/semester/x/course/y/review?tab=6`, clicking on "save" or "Save an next" when there are no more comments, leads me to `/fsr/semester/x?tab=1`.

I suspect this also happens when e.g. sending an email to participants or deleting a course, but I didn't test this.

PR #487, opened 2015-01-15
Hide comments marked for further review

### LINKED ISSUE #256: Do not show comments marked for further review again
Comments marked for further review should not be shown again in the current review session, but only after one clicked again on the review button on the semester page.

rs22:
Fixes #256 
IDs of Answers that are marked for further review are passed as a GET parameter to the next review page and are re-submitted in the HTML form on every review step.

Additionally the previously selected tab on the semester page is selected again when the review is done.


karyon:
there was this `offset` varaible in that view, and i think it was supposed to do that, but didn't work at all. sorry for not having said that earlier. it might be that your approach is the better one, i'll need to investigate a bit and then remove the offset stuff if need be.

regarding the selected tab, could you have a look at the last comment in #249 and say what you think about post-paramater vs cookie vs something else?

karyon:
okay, we'll take your approach. could you remove all that offset stuff and base_queryset and everything that is not needed anymore? also, please rename everything hidden* to something else, maybe skipped. hidden is used by the TextAnswer model (and the ReviewTextAnswerForm) to indicate that a comment should not shown to the public.

rs22:
Alright, should I also implement the persistent tabs (#249)?

I would also suggest a solution using cookies or localStorage, similar to this one:
http://stackoverflow.com/a/10524697 (I'd prefer localStorage, it's [broadly supported](http://caniuse.com/#feat=namevalue-storage))
From what I've seen, we don't need the selected tab information on the server side at all...

I think it should persist the selected tab per Semester, should the first tab always be selected when opening the page from the Navigation link?

karyon:
well, if you want to, sure!

i don't know much about cookies vs localStorage and have no opinion there, i guess it's your call.

no, we don't need that on the server side.

i don't think it's necessary to do this per semester, switching between semesters is usually not done when using EvaP. yes, i think the tab selection should be reset when clicking on a semester in the staff overview page or in the menu.

evap/staff/templates/staff_course_review.html
```diff
@@ -4,13 +4,15 @@
...
-    <form action="{% url "evap.staff.views.course_review" semester.id course.id offset %}" method="POST" class="form-horizontal multiselect-form">
+    <form action="{% url "evap.staff.views.course_review" semester.id course.id %}" method="POST" class="form-horizontal multiselect-form">
     <fieldset>
         <legend>{% trans "Answer Review" %}</legend>
         <p>{% blocktrans with TextAnswer.elements_per_page as elements_per_page and course.open_textanswer_set|length as open and course.textanswer_set|length as total %}Note: For better usability only {{ elements_per_page }} comments are shown at once. For this course {{ open }} out of {{ total }} comments have not been reviewed.{% endblocktrans %}</p>
 
         {% csrf_token %}
         {{ formset.management_form }}
+        <input type="hidden" name="skipped_answers" value="{{ skipped_answers }}" id="skipped_answers" />
+        <input type="hidden" name="tab" value="{{ tab }}" id="tab" />
```
karyon: there's a leftover here

evap/staff/templates/staff_semester_view.html
```diff
@@ -154,8 +154,26 @@
...
 
 {% block on_document_ready %}
-    $("a[data-toggle='tab']").on("shown.bs.tab", function(e) {
-        history.pushState(null, null, "?tab=" + $(this).data("tab"));
+    var defaultTab = "#tab1";
+
+    var storedHash = localStorage.getItem('evap.staff_semester_view.last_tab');
+    var hash = window.location.hash || storedHash || defaultTab;
+    history.replaceState(null, null, hash);
+    $('ul.nav a[href="' + hash + '"]').tab('show');
```
karyon: please add some comment on what these lines do

evap/staff/templates/staff_semester_view.html
```diff
@@ -154,8 +154,26 @@
...
+    });
+
+    // when navigating back
+    $(window).on("popstate", function() {
+        var hash = window.location.hash || defaultTab;
+        $('ul.nav a[href="' + hash + '"]').tab('show');
+    });
+
+    $('a[data-toggle="tab"]').on('shown.bs.tab', function (e) {
+        localStorage.setItem('evap.staff_semester_view.last_tab', this.hash);
```
karyon: is there a reason why this is not in the  click event handler?

karyon:
@janno42 noticed the skipped parameter gets written into the URL, but never read. if you remove that, then we're good to go i think.

evap/staff/views.py
```diff
@@ -356,51 +347,46 @@ def course_review(request, semester_id, course_id, offset=None):
...
-    if offset is None:
-        # get offset for current course
-        key_name = "course_%d_offset" % course.id
-        offset = cache.get(key_name) or 0
-
-        # store offset for next page view
-        cache.set(key_name, (offset + TextAnswer.elements_per_page) % base_queryset.count())
-    else:
-        offset = int(offset)
+    skipped_answers = request.POST.get("skipped_answers", "") if request.POST else request.GET.get("skipped_answers", "")
```
rs22: The skipped parameter is read from the URL here - is there a way to make this more explicit?
It is needed because we're doing a redirect to the same view when opening the next review-form

evap/staff/templates/staff_course_review.html
```diff
@@ -4,13 +4,14 @@
...
     {{ block.super }}
-    <form action="{% url "evap.staff.views.course_review" semester.id course.id offset %}" method="POST" class="form-horizontal multiselect-form">
+    <form action="{% url "evap.staff.views.course_review" semester.id course.id %}" method="POST" class="form-horizontal multiselect-form">
     <fieldset>
         <legend>{% trans "Answer Review" %}</legend>
         <p>{% blocktrans with TextAnswer.elements_per_page as elements_per_page and course.open_textanswer_set|length as open and course.textanswer_set|length as total %}Note: For better usability only {{ elements_per_page }} comments are shown at once. For this course {{ open }} out of {{ total }} comments have not been reviewed.{% endblocktrans %}</p>
 
         {% csrf_token %}
         {{ formset.management_form }}
+        <input type="hidden" name="skipped_answers" value="{{ skipped_answers }}" id="skipped_answers" />
```
janno42: i'm still a little bit confused. do you need this hidden field in the form and the url parameters?

karyon:
rebased an merged manually, thanks!

janno42:
Thanks!

PR #494, opened 2015-01-16
use chevrons on collapsibles

janno42:
This adds little chevrons on the collapsibles in the FAQ and the student voting page:
chevrons


Bartzi:
looks pretty good, but there seems to be a layout error on the student vote page:

image

after you fixed that error: :shipit: :+1:

Bartzi:
the identified problem seems to be caused by the way fonts are rendered in browsers
I've opened a new Issue for that: #495

PR #500, opened 2015-01-17
Support for Vagrant Hyper-V Provider

rs22:
On many Windows 8 Systems Hyper-V is built in - once enabled you cannot use VirtualBox on the same machine.

There's support for using Vagrant in combination with Hyper-V but unfortunately Vagrant's networking features are not supported. As a result some manual configuration is required to make the dev environment work with Hyper-V (see below).

Because the Puppetlabs-Box used with VirtualBox is not available for Hyper-V, I had to start off with a standard Ubuntu. I reviewed the [Puppetlabs Ubuntu configuration](https://github.com/puppetlabs/puppet-vagrant-boxes/tree/master/definitions/ubuntu-server-12042-x64-vbox4210), but ended up with just installing the latest puppet version as a ruby gem before the actual puppet provisioning happens (see hyperv.sh).

**How to use**
 1. Enable Hyper-V (note the warning on [this page](https://docs.vagrantup.com/v2/hyperv/index.html))
 2. Configure Networking

     You can either connect the VM to your external network directly by creating an *External Virtual Switch* in Hyper-V Manager->Virtual Switch Manager, or use "NAT-Mode" by creating an *Internal Virtual Switch* and setting up [Windows Internet Connection Sharing](http://windows.microsoft.com/en-us/windows/using-internet-connection-sharing#1TC=windows-7). I recommend the latter.
 3. `vagrant up --provider hyperv`

      Select the virtual switch created in step 2 and enter your Windows host credentials when prompted, in order to give the virtual machine access to the working directory via SMB.

 4.  Because port forwarding doesn't work with Hyper-V, you have to access the EvaP website using the VM's IP address. When using Django development server, make sure that it binds to all IP addresses (ie `python manage.py runserver 0.0.0.0:8000`)
 5.  That's it! :relieved:


rs22:
Please don't merge yet, I tried this on a different machine and I'm getting a TemplateEncodingError when migrating the database. Still trying to figure out why...

karyon:
sorry for not replying earlier, that was introduced by the release of django 1.8a1, which the pip 1.0 in the vagrant vm installs. try our latest master, it should work there.

rs22:
Thanks! It's working now...

karyon:
added documentation to installation.rst, rebased and merged manually. thanks!

PR #490, opened 2015-01-15
Fixes #172 

hrantzsch:
fix #172

PR #501, opened 2015-01-17
Sorting of questionnaires

### LINKED ISSUE #172: Easier sorting of questionnaires
Currently, questionnaires are sorted by an index, which is an attribute of each questionnaire. This attribute should not need to be edited directly, but through a drag'n'drop GUI.

Bartzi:
fixes #172 
and addresses all comments made in #490.

There is just one thing I did not do:
* I did not delete the hidden field rendering in ```faq_index.html``` and ```faq_section.html``` because they do not use the ```as_div``` function declared in the BootstrapMixin.

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -87,7 +87,7 @@ class Questionnaire(models.Model):
     teaser_en = models.TextField(verbose_name=_(u"teaser (english)"), blank=True, null=True)
     teaser = Translate
 
-    index = models.IntegerField(verbose_name=_(u"ordering index"))
+    index = models.IntegerField(verbose_name=_(u"ordering index"), default=0)
 
     is_for_contributors = models.BooleanField(verbose_name=_(u"is for contributors"), default=False)
     obsolete = models.BooleanField(verbose_name=_(u"obsolete"), default=False)
--- a/evap/staff/templates/staff_course_review.html
+++ b/evap/staff/templates/staff_course_review.html
@@ -27,9 +27,6 @@
                     </h3>
                 </div>
                 <div class="panel-body">
-                    {% for hidden in form.hidden_fields %}
-                        {{ hidden }}
-                    {% endfor %}
                     {% if form.non_field_errors %}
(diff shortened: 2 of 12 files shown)


evap/evaluation/models.py
```diff
@@ -10,6 +10,7 @@
 from django_fsm.db.fields import FSMField, transition
 import django.dispatch
 from django.contrib.auth.models import AbstractBaseUser, BaseUserManager, PermissionsMixin, Group
+from django import forms
```
karyon: please remove that again

evap/static/js/csrf_utils.js
```diff
@@ -0,0 +1,32 @@
...
+		for (var i = 0; i < cookies.length; i++) {
+			var cookie = jQuery.trim(cookies[i]);
+			// Does this cookie string begin with the name we want?
+			if (cookie.substring(0, name.length + 1) == (name + '=')) {
+				cookieValue = decodeURIComponent(cookie.substring(name.length + 1));
+				break;
+			}
+		}
+	}
+	return cookieValue;
```
karyon: you could just `return decodeURIFoo()` inside the loop.

also, is it really that complicated to retrieve some cookie value? :) (i honestly don't know)

evap/staff/templates/staff_course_review.html
```diff
@@ -27,9 +27,6 @@ <h3 class="panel-title">
                     </h3>
                 </div>
                 <div class="panel-body">
-                    {% for hidden in form.hidden_fields %}
-                        {{ hidden }}
-                    {% endfor %}
```
karyon: what about the other forms where we render hidden like this, should they be changed as well?

Bartzi: They should not be changed as they do not use the ```as_div``` method of the ```BootstrapMixin```

Bartzi:
okay I made changes that address all comments :tada:

karyon:
please rebase and then :ship:

Issue #371, opened 2014-08-05
Archiving functionality


karyon:
current state:
 * the number of participants and voters can be made independent of the userprofiles. this is done by setting participant_count and voter_count of the course. this is what the remove_participant_information.py script does, currently only for all courses in a specified semester. this process is called "archiving".
 * all the answers don't need to be dealt with as they are independent from the voter anyways.

remaining problems:
 1. we want users to be able to see all the courses they enrolled in. thus we can't just call the script for a semester once it's fully evaluated.
 2. some users will exist for two years (master), some for three (bachelor), some for five or six (bachelor and master, possibly internship), and some will exist even longer when working at the HPI. While we can probably delete all enrollments after five or six years, it would maybe be cool to keep all contributions as long as the user exists. Basically, we can't just delete users after a certain time, but rather have to delete them when they actually don't exist anymore.
 3. currently there is no way to make the results independent from contributor userprofiles. as we ideally keep the results for forever, this probably needs to be dealt with.

possible solutions for these problems:
  * [x] for 1.: we could easily fix that by just setting the counts in the script, but not deleting the participant and voters list.
  * [ ] for 2.: We find out per UserProfile, whether the Kerberos account still exists. We should ask the Admins whether that's possible. This would probably make EvaP even more HPI-specific. i don't have other ideas how to solve this 100% correctly. The Todo here is to ask the admins and to provide functionality in EvaP to batch delete users.
  * [x] For 3.: We can change the usernames for contributors instead of deleting them. @janno42 and I decided that it's not worth to do that in code. The importer already warns on naming collisions (same username, different email adress), and the importing person has then to deal with this on demand.

other todos:
  * [x] check whether when deleting users, other database objects with foreign keys on these users need to be manually updated or whether that happens automatically. CHECKED: object are deleted CASCADE.
  * [x] introduce Course.archive and Course.is_archived and use it in the script.
  * [x] in the code, make sure that users cannot be deleted when they are enrolled in a non-archived course so that results are not invalidated.
  * [x] for extra safety, we should add `on_delete=models.SET_NULL` as field constructor parameter to Course.last_modified_user. It shouldn't be possible to delete contributors but you never know.
  * [x] implement prerequisites to archiving a semester: all courses must be new or published.
  * [x] throw exceptions when something is archived twice.
  * [x] all views that do something in the semester should check whether the semester is already archived and yield some error then.
  * [x] add a "archive" button to the semester view.
[...]

PR #505, opened 2015-01-20
Archiving

karyon:
yes, archiving. #371 is not completely fixed, mainly the documentation in the wiki is still missing.

this might regress a bit on the user index page loading performance (i haven't measured it), but there is an open ticket for that, right?


janno42:
Please disable the on/off buttons for the reward points and re-enable the link to all courses (of course read-only).

janno42:
:shipit:

Issue #316, opened 2014-06-11
Editing questionnaires can destroy results


janno42:
When editing a questionnaire after it was used in an evaluation, the results can be destroyed because answers are not necessarily linked to the same questions anymore.

Changes in questionnaires should not influence running and past evaluations.

PR #507, opened 2015-01-22
Prevent editing questionnaires

karyon:
questionnaires that are in use may not be edited anymore. that's a very safe solution for #316, but not an elegant one. when closing #316, i'll create a new issue for the feature request of non-destructively editing questionnaires.

the second commit is a UI tweak.

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -101,8 +101,12 @@ class Questionnaire(models.Model):
         return self.name
 
     @property
-    def can_staff_delete(self):
+    def can_staff_edit(self):
         return not self.contributions.exists()
+    
+    @property
+    def can_staff_delete(self):
+        return self.can_staff_edit
 
 
 class Course(models.Model):
--- a/evap/staff/templates/staff_index.html
+++ b/evap/staff/templates/staff_index.html
@@ -23,11 +23,11 @@
         </div>
         <div class="col-md-3">
             <h3>{% trans "Questionnaires" %}</h3>
+            <ul><li><a href="{% url "evap.staff.views.questionnaire_index" %}">{% trans "All questionnaires" %}</a></li></ul>
             <h4>{% trans "Courses" %}</h4>
             <p>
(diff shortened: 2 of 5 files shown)


evap/staff/templates/staff_index.html
```diff
@@ -23,11 +23,11 @@
         </div>
         <div class="col-md-3">
             <h3>{% trans "Questionnaires" %}</h3>
+            <a href="{% url "evap.staff.views.questionnaire_index" %}">{% trans "All questionnaires" %}</a>
```
karyon: @janno42, if you could check this out and see whether you can pretty it up, that would be cool.

evap/staff/templates/staff_questionnaire_index_list.html
```diff
@@ -25,7 +25,7 @@ <h3 class="panel-title">{% trans headline %}</h3>
                             {% blocktrans count questionnaire.contributions.count as count %}used {{ count }} time {% plural %} used {{ count }} times {% endblocktrans %}
                         </td>
                         <td class="right">
-                            {% if not questionnaire.obsolete %}
```
karyon: so previously obsolete questionnaires could not be edited. now, as used questionnaires cannot be edited, i thought that wouldn't matter anymore. @janno42 please speak up if you think otherwise.

evap/evaluation/models.py
```diff
@@ -101,8 +101,12 @@ def __unicode__(self):
         return self.name
 
     @property
+    def can_staff_edit(self):
+        return self.contributions.exists()
```
janno42: not? ;)

karyon: oops

janno42:
there are also edit and delete buttons on the questionnaire preview page. could you please disable them as well?

janno42:
and could you please also add a check in the ```questionnaire_edit``` view?

janno42:
squash and go :+1:

PR #510, opened 2015-01-23
Importer fixes

### LINKED ISSUE #503: importer: message linebreaks missing and a possible crash
#497 introduced two problems: one new error message does not have the required line breaks, and it added a get() that may throw due to multiple objects returned.

karyon:
so i attempted to fix #503 and then did other things as well.


janno42:
:+1:

PR #511, opened 2015-01-23
Enable contributor page also for contributors who can't edit courses

### LINKED ISSUE #432: All contributors should get "My courses" page
Currently only editors can see the contributor overview page with all their courses. This should be enabled for all contributors.

On this page users should have different options per course based on their role in this course and the current state:
- Editors (and their delegates):
  - when the course is ```prepared```: they have an edit and a preview option
  - when the course is between ```lecturerApproved``` and ```reviewed``` : they have a view button linking to the read only edit page and a preview option
  - when the course is ```published```: they have a button linking to the course's results
- Contributors who are not editors:
  - when the course is between ```prepared``` and ```reviewed``` : they have a preview option
  - when the course is ```published```: they have a button linking to the course's results

janno42:
fix #432

### DIFF
--- a/evap/contributor/templates/contributor_index.html
+++ b/evap/contributor/templates/contributor_index.html
@@ -66,10 +66,12 @@
                                 {% if course.state == 'published' %}
                                     <a href="{% url "evap.results.views.course_detail" semester.id course.id %}" class="btn btn-sm btn-default">{% trans "Results" %}</a>
                                 {% else %}
-                                    {% if course.state == 'prepared' %}
-                                        <a href="{% url "evap.contributor.views.course_edit" course.id %}" class="btn btn-sm btn-primary">{% trans "Edit" %}</a>
-                                    {% else %}
-                                        <a href="{% url "evap.contributor.views.course_view" course.id %}" class="btn btn-sm btn-default">{% trans "View" %}</a>
+                                    {% if course|is_user_editor_or_delegate:user %}
+                                        {% if course.state == 'prepared' %}
+                                            <a href="{% url "evap.contributor.views.course_edit" course.id %}" class="btn btn-sm btn-primary">{% trans "Edit" %}</a>
+                                        {% else %}
+                                            <a href="{% url "evap.contributor.views.course_view" course.id %}" class="btn btn-sm btn-default">{% trans "View" %}</a>
+                                        {% endif %}
                                     {% endif %}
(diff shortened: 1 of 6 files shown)


evap/contributor/views.py
```diff
@@ -28,7 +28,7 @@ def index(request):
     semesters = Semester.objects.all()
     semester_list = [dict(semester_name=semester.name, id=semester.id, courses=[course for course in all_courses if course.semester_id == semester.id]) for semester in semesters]
 
-    template_data = dict(semester_list=semester_list, delegated_courses=delegated_courses)
+    template_data = dict(semester_list=semester_list, delegated_courses=delegated_courses, user=user)
```
karyon: could you check whether this is necessary? we should automatically have a user variable in templates, if not, we have request.user.

evap/contributor/templates/contributor_index.html
```diff
@@ -66,10 +66,12 @@ <h3 class="panel-title">{{ semester.semester_name }}</h3>
...
-                                    {% else %}
-                                        <a href="{% url "evap.contributor.views.course_view" course.id %}" class="btn btn-sm btn-default">{% trans "View" %}</a>
+                                    {% if course|is_user_editor_or_delegate:user %}
+                                        {% if course.state == 'prepared' %}
+                                            <a href="{% url "evap.contributor.views.course_edit" course.id %}" class="btn btn-sm btn-primary">{% trans "Edit" %}</a>
+                                        {% else %}
+                                            <a href="{% url "evap.contributor.views.course_view" course.id %}" class="btn btn-sm btn-default">{% trans "View" %}</a>
+                                        {% endif %}
                                     {% endif %}
                                     <a href="{% url "evap.contributor.views.course_preview" course.id %}" class="btn btn-sm btn-default">{% trans "Preview" %}</a>
```
karyon: the ticket says the preview should only be available in prepared, but i don't really care.

karyon:
:ship:

evap/contributor/templates/contributor_index.html
janno42: that was a mistake. the ticket also says "Contributors who are not editors: when the course is between prepared and reviewed : they have a preview option"

evap/contributor/views.py
janno42: removed.

PR #514, opened 2015-01-25
Fix testdata

karyon:
see commit messages for what has been done.


janno42:
i didn't test the ```dump_testdata``` and just assume it's working.
:+1:

PR #524, opened 2015-02-06
add legal notice

### LINKED ISSUE #523: Legal notice
We need a page for a legal notice (Impressum).

janno42:
fix #523.

adds the infrastructure for a legal notice page (impressum).

for using this page, a link name and the name of the file holding the actual content have to be set in the settings (```LEGAL_NOTICE_LINK``` and ```LEGAL_NOTICE_FILE```). only if the first value is set, a link to the page is shown in the footer on every page.
the extra file is used to enable updates without changing the content of the legal notice but also not including the data into an own model in the database.

this explanation and the following example for a (german) "impressum_text.html" will be added to the docs, when this pr is getting merged.

```
<p>
	Die Evaluierungsplattform wird betrieben von:<br />
	<br />
	<b>Maximilian Mustermann</b><br />
	Musterstraße 1<br />
	12345 Musterstadt<br />
	<br />
	E-Mail: <a href="mailto:maximilian@mustermann.de">maximilian@mustermann.de</a><br />
	Inhaltlich Verantwortlicher gemäß § 5 TMG: Maximilian Mustermann<br />
	<br/>
	<br/>
	Trotz größtmöglicher Sorgfalt kann nicht ausgeschlossen werden, dass einzelne Informationen auf dieser Seite veraltet oder nicht zutreffend sind. Es wird keine Haftung für Aktualität, Genauigkeit und Vollständigkeit der publizierten Informationen übernommen.<br />
	Für den Inhalt verlinkter externer Webseiten sind lediglich deren Betreiber verantwortlich.
</p>

```

### DIFF
--- a/evap/context_processors.py
+++ b/evap/context_processors.py
@@ -1,0 +1,10 @@
+from django.conf import settings
+
+def feedback_email(request):
+    return {'FEEDBACK_EMAIL': settings.FEEDBACK_EMAIL}
+
+def legal_notice_active(request):
+	return {'LEGAL_NOTICE_ACTIVE': settings.LEGAL_NOTICE_ACTIVE}
+
+def tracker_url(request):
+    return {'TRACKER_URL': settings.TRACKER_URL}
--- a/evap/evaluation/templates/legal_notice.html
+++ b/evap/evaluation/templates/legal_notice.html
@@ -1,0 +1,15 @@
+{% extends "base.html" %}
+
+{% load i18n %}
+
+{% block title %}{% trans "Legal Notice" %} - {{ block.super }}{% endblock %}
+
+{% block subtitle %}
+    {{ block.super }}
+    <li>{% trans "Legal Notice" %}</li>
+{% endblock %}
+
+{% block content %}
+    <h2>{% trans "Legal Notice" %}</h2>
+    {% if LEGAL_NOTICE_ACTIVE %}{% include "legal_notice_text.html" %}{% endif %}
+{% endblock %}
--- a/evap/evaluation/views.py
+++ b/evap/evaluation/views.py
@@ -2,6 +2,7 @@ from django.contrib import messages
 from django.contrib.auth import login as auth_login
 from django.shortcuts import redirect, render
 from django.utils.translation import ugettext as _
+from django.conf import settings
 
 from evap.evaluation.forms import NewKeyForm, LoginKeyForm, LoginUsernameForm
 from evap.evaluation.models import UserProfile, FaqSection, EmailTemplate
@@ -80,3 +81,6 @@ def index(request):
 
 def faq(request):
     return render(request, "faq.html", dict(sections=FaqSection.objects.all()))
+
+def legal_notice(request):
+    return render(request, "legal_notice.html", dict())
--- a/evap/settings.py
+++ b/evap/settings.py
@@ -97,6 +97,10 @@ REPLY_TO_EMAIL = DEFAULT_FROM_EMAIL
 if DEBUG:
     EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 
+# Config for legal notice
+# The HTML file which should be used must be located in evap\templates\legal_notice_text.html
+LEGAL_NOTICE_ACTIVE = False
+
 
 ### Application definition
 
@@ -137,6 +141,9 @@ TEMPLATE_CONTEXT_PROCESSORS = (
     "django.core.context_processors.static",
     "django.core.context_processors.request",
     "django.contrib.messages.context_processors.messages",
+    "evap.context_processors.feedback_email",
+    "evap.context_processors.legal_notice_active",
+    "evap.context_processors.tracker_url",
 )
 
 AUTHENTICATION_BACKENDS = (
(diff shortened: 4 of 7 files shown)


evap/context_processors.py
```diff
@@ -0,0 +1,10 @@
+from django.conf import settings
+
+def feedback_email(request):
+    return {'FEEDBACK_EMAIL': settings.FEEDBACK_EMAIL}
+
+def legal_notice(request):
+    return {'LEGAL_NOTICE_LINK': settings.LEGAL_NOTICE_LINK}
+
+def tracker_url(request):
+    return {'TRACKER_URL': settings.TRACKER_URL}
```
karyon: as we are using all these settings in only one view each, i would prefer to not add them to every context. passing them as template parameters is probably the easiest way. alternatively you could add a templatetags file "settings" and include it where needed.

evap/evaluation/views.py
```diff
@@ -80,3 +81,6 @@ def index(request):
 
 def faq(request):
     return render(request, "faq.html", dict(sections=FaqSection.objects.all()))
+
+def impressum(request):
+    return render(request, "impressum.html", dict(LEGAL_NOTICE_FILE=settings.LEGAL_NOTICE_FILE))
```
karyon: is there a reason why this is done as a template parameter and teh other settings are not?

evap/settings.py
```diff
@@ -97,6 +97,10 @@
 if DEBUG:
     EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 
+# Config for legal notice
+LEGAL_NOTICE_LINK = '' # e.g. "Impressum"
+LEGAL_NOTICE_FILE = '' # e.g. "impressum_text.html" which then has to be in evap\templates
```
karyon: why is the setting called "legal notice" while the template and view name is "impressum"?

evap/context_processors.py
janno42: they are used in the base file

evap/evaluation/views.py
janno42: because it's just needed in one view and the others in every view, because they are used in the base file.

evap/context_processors.py
karyon: alternatively you could add a templatetags file "settings" and include it where needed.

evap/settings.py
janno42: yes, quite inconsistent. there is no real "impressum" in english - shall we call it "legal notice" everywhere?

evap/context_processors.py
janno42: what's the benefit? i thought context processors are meant for this exact situation, when you need some variables in all contexts.

karyon: what i find weird is that we have this settings available in all templates. when doing template tags, we would only have them available in base.html and impressum.html. in my opinion that's better, but i'm not completely sure.

evap/settings.py
karyon: i guess so...

janno42: renamed

evap/context_processors.py
janno42: we'll keep it

evap/evaluation/templates/legal_notice.html
```diff
@@ -0,0 +1,15 @@
+{% extends "base.html" %}
+
+{% load i18n %}
+
+{% block title %}{% trans "Impressum" %} - {{ block.super }}{% endblock %}
```
karyon: trans "Legal Notice"

evap/evaluation/templates/legal_notice.html
```diff
@@ -0,0 +1,15 @@
...
+
+{% block subtitle %}
+    {{ block.super }}
+    <li><a href="{% url "evap.evaluation.views.legal_notice" %}">{% trans "Legal Notice" %}</a></li>
+{% endblock %}
+
+{% block content %}
+    <h2>Impressum</h2>
```
karyon: trans "Legal Notice"

karyon:
what do you think of hard-coding the path, adding a dummy template and replacing the existing settings by "ENABLE_LEGAL_NOTICE" and telling the users in the docs to use `git update-index --assume-unchanged legal_notice_something.html`? alternatively we could hard-code the path, not add the file and on the comment for ENABLE_LEGAL_NOTICE tell the user where to put that file.

janno42:
path hard-coded, file not added, comment in the settings

evap/evaluation/templates/legal_notice.html
```diff
@@ -0,0 +1,15 @@
+{% extends "base.html" %}
+
+{% load i18n %}
+
+{% block title %}{% trans "Legal Notice" %} - {{ block.super }}{% endblock %}
+
+{% block subtitle %}
+    {{ block.super }}
+    <li><a href="{% url "evap.evaluation.views.legal_notice" %}">{% trans "Legal Notice" %}</a></li>
```
karyon: is this the breadcrumb? if yes, please remove the link. i think there shouldn't be any other pages which have a link on themselves.

karyon:
after the last comment is fixed, :shipit:

evap/evaluation/templates/legal_notice.html
janno42: Stupid breadcrumbs ;)

PR #525, opened 2015-02-07
Remove unused imports, fix .landscape.yaml and minor code issues

cyroxx:
I checked Landscape.io and fixed a couple of issues mentioned there, especially the ones about unused imports.

### DIFF
--- a/evap/contributor/urls.py
+++ b/evap/contributor/urls.py
@@ -1,6 +1,6 @@
 from django.conf.urls import url
 
-from evap.contributor.views import *
+from evap.contributor.views import index, profile_edit, course_view, course_edit, course_preview
 
 urlpatterns = [
     url(r"^$", index),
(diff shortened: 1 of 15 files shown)


evap/contributor/urls.py
```diff
@@ -1,6 +1,6 @@
 from django.conf.urls import url
 
-from evap.contributor.views import *
+from evap.contributor.views import index, profile_edit, course_view, course_edit, course_preview
```
janno42: i would argue that this is a valid import of everything (because every view will get a url) and might even be better. but okay, let's do it for landscape ;)

cyroxx: Well, I restricted the import to the views only. Everything else would just unnecessarily clutter the module's namespace, as can be seen here ("Unused imports from wildcard import..."): https://landscape.io/github/fsr-itse/EvaP/370/modules/evap/contributor/urls.py#L3

PR #528, opened 2015-02-17
added missing migration

janno42:
auto-created migration that was still missing from latest commits


karyon:
could you make the commit message "added missing migration from <sha1>"? and, i don't know whether there is some convention at this, but could you give the migration a name, something like "fsm_upgrade"?

janno42:
okay

PR #529, opened 2015-02-17
Replace jquery.uix.multiselect with chosen

PFischbeck:
This replaces the old (ugly) multiselect with [chosen](https://github.com/harvesthq/chosen), a simple and clean jQuery plugin.

chosen


karyon:
@janno42 please talk to me before merging.

PFischbeck:
The plugin could replace all `select` elements if we want it that way... For example, it could add the search function to the contributors list on the course edit page.
chosen course

karyon:
that would be very nice. ref #346

evap/templates/base.html
```diff
@@ -193,7 +191,11 @@
 
             // run the jquery multiselect plugin
             $("select[multiple='multiple']").each(function(i,e) {
-                $(e).multiselect({ locale: "{{ LANGUAGE_CODE }}" });
+                $(e).chosen({
+                    search_contains: true,
+                    no_results_text: "{% trans "No results match" %}",
+                    placeholder_text_multiple: "{% trans "Select some options" %}"
```
janno42: Please use ```No results for``` and ```Please select...```

janno42:
Very nice. Can you please leave out the translations for now? Just remove the po file from your commit.

evap/templates/base.html
```diff
@@ -193,7 +191,12 @@
 
             // run the jquery multiselect plugin
```
karyon: please update this comment :)

karyon:
browsing a bit around chosen, i must say i actually like [select2](https://github.com/select2/select2) better. it seems more actively maintained are pull request are usually merged. regarding functionality we need they are equivalent i think.

i still need to check whether i like this widget for participant lists of courses. when there are 70 people selected this is probably much less clear than the old widget, i need to check that.

karyon:
i think this is not really cool for the participant lists, maybe we should keep the old multiselect (or some alternative) there. yes, it's ugly, but at least we have a proper list of participants with it. but i definitely want chosen/select2 for all the other cases we use multiselect right now. @janno42, what do you say?

janno42:
yes, it's not perfect for the participant list. but the multiselect definitely isn't as well. in the select version  i like that all participants are visible without scrolling - that would not be an option if all of them would be listed below each other, because that would result in a too long list.

i can't think of any other widget that would be better for the participant list right now, but i still prefer the proposed select over the old multiselect and would like to use it.

janno42:
replacing the other occurrences is definitely a very good idea.

karyon:
alright then. @PFischbeck, would you like to replace chosen by select2? otherwise i'd do it, i guess. Also, if you want, you could make the ontributor selection use this. otherwise, well, i'd do it :)

PFischbeck:
@karyon I'd prefer if you did it. I can help reviewing your PR :wink:
I suggest you also take a look at [select2-bootstrap-css](https://github.com/t0m/select2-bootstrap-css/tree/bootstrap3) for a bootstrap-styling of select2.

Issue #251, opened 2014-03-05
make reviewing-page more compact


karyon:
when reviewing comments, i need more time for scrolling than for actually reading comments.

the input field should be only as large as it needs to be. additionally, one could try to move the two checkboxes to the right of the text field, right now they need a lot of horizontal space.

additionally, 10 instead of 5 comments should be shown per page, as reload times are quite noticable.

PR #530, opened 2015-02-17
WIP: New comment review

### LINKED ISSUE #522: Show selected text answers only to evaluated person
When reviewing text answers there should be an option to make single text answers visible for only the evaluated person and not for the course's responsible person.
This is necessary to publish text answers which are positive for the reviewed person but might be understood negatively by the responsible person.

The solution should include a message to the evaluated persons next to these answers to let them know that only they can see them.

janno42:
Fix #522, #251 

Changes how the comments are reviewed. See commit messages for details.

Todo:
- [x] Enable editing comments
- [x] Update TextAnswer states in test data


evap/evaluation/migrations/0011_remove_textanswer_checked_and_hidden.py
```diff
@@ -0,0 +1,22 @@
...
+    operations = [
+    	migrations.RemoveField(
+            model_name='textanswer',
+            name='checked',
+        ),
+        migrations.RemoveField(
+            model_name='textanswer',
+            name='hidden',
+        ),
+    ]
```
karyon: the migrations could be squashed, what do you think?

evap/evaluation/models.py
```diff
@@ -186,12 +186,16 @@ def save(self, *args, **kw):
...
         return not self.open_textanswer_set.exists()
 
-    def is_fully_checked_except(self, ignored_answers):
-        """Shortcut for finding out if all text answers to this course have been checked except for specified answers"""
+    def is_not_fully_reviewed(self):
+        """Shortcut for finding out whether some text answers to this course have not yet been reviewed"""
+        return self.open_textanswer_set.exists()
+
+    def is_fully_reviewed_except(self, ignored_answers):
+        """Shortcut for finding out if all text answers to this course have been reviewed except for specified answers"""
```
karyon: Please replace all the "Shortcut for finding out whether..." by "checks whether" or "returns whether" or something like that.

evap/evaluation/models.py
```diff
@@ -218,7 +222,7 @@ def can_staff_delete(self):
 
     @property
     def can_staff_review(self):
-        return self.state in ['inEvaluation', 'evaluated'] and not self.is_fully_checked()
+        return self.state in ['inEvaluation', 'evaluated', 'reviewed'] and len(self.textanswer_set) > 0
```
karyon: please use .exists, otherwise the query will be fully evaluated

evap/evaluation/models.py
```diff
@@ -252,10 +256,14 @@ def evaluation_begin(self):
     def evaluation_end(self):
         pass
 
-    @transition(field=state, source='evaluated', target='reviewed', conditions=[is_fully_checked])
+    @transition(field=state, source='evaluated', target='reviewed', conditions=[is_fully_reviewed])
     def review_finished(self):
         pass
 
+    @transition(field=state, source='reviewed', target='evaluated', conditions=[is_not_fully_reviewed])
```
karyon: how can a course be in the state reviewed and at the same time not be fully reviewed?

evap/evaluation/models.py
```diff
@@ -362,12 +370,12 @@ def textanswer_set(self):
     @property
     def open_textanswer_set(self):
         """Pseudo relationship to all text answers for this course"""
-        return TextAnswer.objects.filter(contribution__in=self.contributions.all(), checked=False)
+        return TextAnswer.objects.filter(contribution__in=self.contributions.all(), state='')
```
karyon: i think it is very weird that the default state is the empty string. is there a good reason for that?

evap/evaluation/migrations/0011_remove_textanswer_checked_and_hidden.py
janno42: https://docs.djangoproject.com/en/1.7/topics/migrations/#data-migrations says
```
Migrations that alter data are usually called “data migrations”; they’re best written as separate migrations, sitting alongside your schema migrations.
```
So let's keep them and optimize them when we do the next general migration optimization.

karyon: k

evap/evaluation/models.py
```diff
@@ -512,19 +520,35 @@ class TextAnswer(Answer):
...
-    elements_per_page = 5
-
     reviewed_answer = models.TextField(verbose_name=_("reviewed answer"), blank=True, null=True)
     original_answer = models.TextField(verbose_name=_("original answer"), blank=True)
-
-    checked = models.BooleanField(verbose_name=_("answer checked"), default=False)
-    hidden = models.BooleanField(verbose_name=_("hide answer"), default=False)
+    
+    TEXT_ANSWER_STATES = (
+        ('N', _('not published')),
```
karyon: i think "not published" is a bit unortunate. i like "rejected" or "hidden" better, but they are not optimal as well imo.

evap/evaluation/tools.py
```diff
@@ -94,7 +95,11 @@ def mix(a, b, alpha):
     return alpha * a + (1 - alpha) * b
 
 
-def get_answers(course, contribution, question):
+def get_all_textanswers(course, contribution, question):
+    return get_answers(course, contribution, question, exclusion=[])
+
+
+def get_answers(course, contribution, question, exclusion=['', 'N']):
```
karyon: is there a difference between the textanswers of `get_all_textanswers` and answers from `get_answers`? if not, please remove/add the text to the other name.

also, i think it is not good to have a "get results" and a "get all results" method. it is totally not obvious that the "get results" excludes some of the results by default. i think it should be "get results" and "get filtered results".

karyon: also, i'd rename that exclusion parameter to exclude or exclude_states.

evap/evaluation/models.py
janno42: when a staff user opens the comments page of a fully reviewed course and then removes the publish decision for one or more comments. this is intentional.

evap/evaluation/tools.py
```diff
@@ -214,7 +218,7 @@ def calculate_results(course, staff_member=False):
                 if answers:
                     results.append(TextResult(
                         question=question,
-                        texts=[answer.answer for answer in answers]
+                        answers=[answer for answer in answers]
```
karyon: well, that doesn't make too much sense anymore.

evap/results/fixtures/minimal_test_data.json
```diff
@@ -0,0 +1,669 @@
...
+},
+{
+    "pk": 1,
+    "model": "rewards.semesteractivation",
+    "fields": {
+        "semester": 1,
+        "is_active": false
+    }
+}
+]
```
karyon: if this is copied from the other minimal_test_data.json, please remove this one and use the other one. if not, please remove all the data that is not used for the tests.

evap/results/tests.py
```diff
@@ -0,0 +1,79 @@
...
+        self.assertNotIn(".orig.", page)
+        self.assertIn(".changed.", page)
+        self.assertIn(".responsible_orig_published.", page)
+        self.assertNotIn(".responsible_orig_not_published.", page)
+        self.assertNotIn(".responsible_orig.", page)
+        self.assertIn(".responsible_changed.", page)
+        self.assertIn(".responsible_orig_private.", page)
+        self.assertNotIn(".responsible_orig_unchecked.", page)
+        self.assertIn(".contributor_orig_published.", page)
+        self.assertNotIn(".contributor_orig_private.", page) # private comment not visible
```
karyon: it took me a while to grok why which comment should be visible or not, but can't think of something better now, lets talk about that.

evap/evaluation/models.py
cyroxx: Did you maybe forget to use `is_not_fully_reviewed` here?

evap/results/views.py
```diff
@@ -102,14 +105,17 @@ def course_detail(request, semester_id, course_id):
             staff=request.user.is_staff)
     return render(request, "results_course_detail.html", template_data)
 
+def user_can_see_text_answer(user, course, text_answer):
+    if user == None:
+        return False
```
karyon: how can this happen?

evap/results/views.py
```diff
@@ -102,14 +105,17 @@ def course_detail(request, semester_id, course_id):
...
-
-    if section.contributor in user.represented_users.all():
+    contributor = text_answer.contribution.contributor
+    if contributor == user:
         return True
+    if text_answer.published:
+        if course.is_user_responsible_or_delegate(user):
+            return True
+        if contributor in user.represented_users.all():
+            return True
```
karyon: this and `text_answer.contribution.contributor` above smell like a whole bunch of database accesses. please check that you don't regress significantly here. maybe even add a test for that...? :)

evap/results/views.py
```diff
@@ -62,9 +62,12 @@ def course_detail(request, semester_id, course_id):
...
         for section in sections:
-            if not user_can_see_textresults(request.user, course, section):
-                for i, result in list(enumerate(section.results))[::-1]:
-                    if isinstance(result, TextResult):
+            for i, result in list(enumerate(section.results))[::-1]:
+                if isinstance(result, TextResult):
+                    for j, answer in list(enumerate(result.answers))[::-1]:
+                        if not user_can_see_text_answer(request.user, course, answer):
+                            del result.answers[j]
+                    if not result.answers:
```
karyon: please check whether you can do this easier, maybe in two passes and using the filter builtin or list comprehensions.

something like
```
for section in sections:
        result.answers = [answer for answer in result.answers if user_can_see_text_answer(blah)]
sections = (section for section in sections if section.answers)
```

karyon:
i didn't finish, will continue tomorrow. also, the tests are red.

evap/evaluation/models.py
cyroxx: I would prefer more descriptive (i.e. longer) names for the different states instead of `'N'`, `'Y'`, `'P'` and `''`. Especially because it is easy to mix up `'Y'` and `'P'`.

I don't know whether the the (translated) state names are appropriate, since I am not sure whether I understand how all the different states should work. Is this already documented somewhere? Especially, what is is the definition of `published privately` (note the oxymoron) and how exactly is it enforced to be private?

evap/evaluation/models.py
```diff
@@ -512,19 +520,35 @@ class TextAnswer(Answer):
...
+    )
+    state = models.CharField(max_length=1, choices=TEXT_ANSWER_STATES, verbose_name=_('state of answer'), default='')
 
     class Meta:
         verbose_name = _("text answer")
         verbose_name_plural = _("text answers")
 
     @property
+    def reviewed(self):
+        return self.state != ''
```
cyroxx: Instead of using the hard-coded strings everywhere for comparing or setting the state, please use constants. An example can be found [in the choices docs](https://docs.djangoproject.com/en/dev/ref/models/fields/#field-choices).

karyon: we should check if we should do this for the other tuples in that style as well.

evap/evaluation/models.py
janno42: no, this is intentional. it is now possible to use the comment review page even when the course was fully reviewed (but not yet published) and by removing some publishing decisions the course can even go one state back (becoming evaluated).

evap/evaluation/models.py
janno42: we could also use any other character like a ```-```, but i think an empty character defines very good that no decision on the publishing state has been made yet.

evap/evaluation/models.py
janno42: longer names are no problem, the complete name is never used so far. so also translation are irrelevant.
there is no documentation yet, but [the wiki](https://github.com/fsr-itse/EvaP/wiki/Publishing-Results) will hold this information.
#522 describes why the feature is needed.

evap/results/views.py
janno42: right, that shouldn't happen anymore

evap/evaluation/models.py
janno42: i will change it to constants. however, i will not change other occurrences (namely the courses' statuses) in this pr - this would need to go somewhere else.

evap/evaluation/models.py
karyon: still, please use exists.

evap/evaluation/models.py
karyon: suppose joe looks at this code and says "here a state is checked for empty string, i guess that should be some real state?" "no, emtpystring is also a state" "alright what does it mean?" "well, it means that the comment has not been reviewed yet"

if you change this to use constants, this is a bit moot, but i think an emptystring as a state is not acceptable.

evap/evaluation/models.py
janno42: yes of course

evap/evaluation/models.py
janno42: comments will be removed completely (it's really not hard to see what is done there) and is_fully_reviewed_except wasn't used anywhere so i'll remove the method.

evap/evaluation/models.py
janno42: i will use the constants ```published```, ```hidden```, ```private``` and ```not_reviewed```

evap/evaluation/tools.py
janno42: will use only one method and call the parameter ```exclude_text_answer_states```

evap/evaluation/tools.py
janno42: will replace it

evap/results/views.py
janno42: unfortunately named tuples are immutable and can't just be updated. i'll have a look into this later.

karyon: where is the named tuple? i don't see it. also, if i got this right i modify nothing where del wasn't used before.

karyon:
could you please check where the 2k removed lines in the test data come from?

evap/results/views.py
janno42: section and result are both named tuples. but as i said, i'll have a look into it.

janno42:
in the test data all hidden and checked fields have been replaced by the state in all answers. that's quite a lot. i just saw that i have to change the state's names there again...

evap/staff/templates/staff_course_comments.html
```diff
@@ -1,33 +1,49 @@
...
+            <div class="panel-heading">
+                <span class="panel-title">{% trans "Comments about the Contributors" %}</span>
+            </div>
+            <div class="panel-body">
+                {% regroup contributor_sections by questionnaire as sections_grouped %}
+                {% include "staff_course_comments_section.html" %}
+            </div>
+        </div>
+    {% endif %}
+{% endblock %}
```
karyon: since you are displaying all comments of a course on one page (i think...?), please pay special attention to loading times when a course has a few hundred comments.

evap/staff/views.py
```diff
@@ -479,17 +426,59 @@ def course_comments(request, semester_id, course_id):
...
+        else:
+            contributor_sections.append(CommentSection(questionnaire, contribution.contributor, contribution.responsible, comments))
 
-    template_data = dict(semester=semester, course=course, textanswers_by_question=textanswers_by_question)
+    template_data = dict(semester=semester, course=course, course_sections=course_sections, contributor_sections=contributor_sections)
     return render(request, "staff_course_comments.html", template_data)
 
 
 @staff_required
+def course_comments_update_publish(request):
```
karyon: why is this not in the same view, checking whether the request is POST or not, like all the other views do it?

evap/staff/views.py
```diff
@@ -479,17 +426,59 @@ def course_comments(request, semester_id, course_id):
...
 
-    textanswers = course.textanswer_set.filter(checked=True)
-
-    textanswers_by_question = []
-    for question_id in textanswers.values_list("question", flat=True).distinct():
-        textanswers_by_question.append((Question.objects.get(id=question_id), textanswers.filter(question=question_id)))
+    course_sections = []
+    contributor_sections = []
+    for questionnaire, contribution in questionnaires_and_contributions(course):
+        comments = []
```
karyon: maybe rename this to text_results?

evap/staff/views.py
```diff
@@ -479,17 +426,59 @@ def course_comments(request, semester_id, course_id):
...
+        for question in questionnaire.question_set.all():
+            if question.is_text_question:
+                answers = get_filtered_answers(course, contribution, question, exclude_text_answer_states=[])
+                if answers:
+                    comments.append(TextResult(
+                        question=question,
+                        answers=answers,
+                    ))
```
karyon: no newlines here please

evap/staff/views.py
```diff
@@ -479,17 +426,59 @@ def course_comments(request, semester_id, course_id):
...
+        if not comments:
+            continue
+        elif contribution.is_general:
```
karyon: make this an if

evap/staff/views.py
```diff
@@ -479,17 +426,59 @@ def course_comments(request, semester_id, course_id):
...
+    course = Course.objects.get(pk=course_id)
+    answer = TextAnswer.objects.get(pk=comment_id)
+
+    if publish_status == 'publish':
+        answer.publish()
+    elif publish_status == 'private':
+        answer.make_private()
+    elif publish_status == 'hide':
+        answer.hide()
+    elif publish_status == 'unreview':
```
karyon: unreview, hide and so on are not really statuses but rather actions, aren't they? or you make the strings "unreviewed", "hidden" etc.

evap/staff/views.py
janno42: because it's not a form that is being sent. it uses the same approach as the new questionnaire reordering and sends posts without reloading the page.

janno42:
Addressing your concerns about the number of database queries and loading times: I tried loading the pages with about 130 comments and didn't notice any delay in any view with any user. Looks quite fine.

evap/evaluation/models.py
```diff
@@ -512,19 +514,39 @@ class TextAnswer(Answer):
...
+    
+    HIDDEN = 'hidden'
+    PUBLISHED = 'published'
+    PRIVATE = 'private'
+    NOT_REVIEWED = 'not_reviewed'
+    TEXT_ANSWER_STATES = (
+        (HIDDEN, _('hidden')),
+        (PUBLISHED, _('published')),
+        (PRIVATE, _('private')),
+        (NOT_REVIEWED, _('not reviewed')),
```
karyon: the example in the django docs use short two-character strings for the constants, i would prefer that.

janno42: come on, just because you two wanted it i renamed them to this. i don't think two letters are better, so i will not change it again.

karyon: i actually meant changing `HIDDEN = 'hidden'` to `HIDDEN = 'HI'`, so just the constant contents, and `max_length=20` to `max_length=2`. there would be no other changes needed, i think.

evap/staff/templates/staff_course_comments_section.html
```diff
@@ -0,0 +1,48 @@
...
+                            </tr>
+                        </thead>
+                        <tbody>
+                            {% for answer in result.answers %}
+                                <tr>
+                                    <td>{{ answer.id }}</td>
+                                    <td>
+                                        {% if answer.reviewed_answer %}
+                                            <a class="comment-link" href="{% url "evap.staff.views.course_comment_edit" semester.id course.id answer.id %}">{{ answer.reviewed_answer }}</a><br />
+                                            <span class="comment-original">({{ answer.original_answer }})</span>
```
karyon: i would prefer an "edit"-button to the answer-as-link, but have no good reasons for that right now. what do you think?

evap/staff/templates/staff_course_comments.html
```diff
@@ -1,33 +1,54 @@
...
+        if($("#"+id+"-"+action).hasClass("active")) {
+            $("#"+id+"-buttons").children().removeClass("active");
+            action = "unreview";
+        }
+        else {
+            $("#"+id+"-buttons").children().removeClass("active");
+            $("#"+id+"-"+action).addClass("active");
+        }
+        var parameters = {"id": id, "action": action, "course_id": {{ course.id }}};
+        $.post('{% url "evap.staff.views.course_comments_update_publish" %}', parameters);
```
karyon: if the user loses internet connection or we introduce a bug that makes the post request fail on the server side, the user will not notice it and quite a bit of work might be lost. i think we should at least do an alert or even only do the switch on the button after we got a positive response from the server.

karyon:
i admit this comes a bit late, but with the previous concept the user had to do a click if he wanted to skip a comment or hide it. now, the user has to do a click for every single comment, which might be annoying. think of the amount of clicking you have to do when reviewing half a thousand comments, it's about half a thousand times. and you'll have to stop reading comments and look at the buttons to reposition the cursor for each single answer.

i'd actually like to require a test with a few hundred comments on real-life data and a reassurance that this concept works fine, or a better concept, but i see that this might be quite annoying for you. just please be ready to open a new ticket if you experience that this concept is not working out as good as you imagined.

keyboard controls might be a solution for this.


if someone goes through a courses comments and does not decide on a few of them, as we usually do, what happens with the rest? scrolling through the page and searching for the two button groups that have no selection yet in two hundred comments might be a bit of a workaround. i guess re-ordering the unreviewed answers to the top does not fit in the concept. maybe highlight them?

also, apropos re-ordering, when reviewing comments before the evaluation is finished, the comments that get added later will be scattered all over the page, right? this might be suboptimal as well. again, for this concept, highlighting or hiding/fading out reviewed answers or something like that may help here.

evap/evaluation/migrations/0010_fill_textanswer_state.py
```diff
@@ -0,0 +1,42 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('evaluation', '0009_add_textanswer_state'),
+    ]
+
+    operations = [
+    	migrations.RunPython(update_states, backward_update_states),
+    ]
```
karyon: there are mixed spaces and tabs in here, please use spaces

evap/staff/fixtures/minimal_test_data.json
```diff
@@ -297,192 +297,176 @@
...
   "pk": 2, 
   "model": "evaluation.textanswer", 
   "fields": {
-    "checked": true, 
     "reviewed_answer": "censored!", 
     "question": 3, 
     "original_answer": "asdf", 
     "contribution": 17, 
-    "hidden": false
+    "state": "Y"
```
karyon: these are not up to date, are they?

evap/results/tests.py
karyon: i think it would help if each comment had the format [course|contributor|responsible]_[orig|changed]_[unchecked|published|etc], right now a lot of it is implicit/missing. also, please remove the comments, or add more of them.

evap/staff/views.py
```diff
@@ -479,17 +426,72 @@ def course_comments(request, semester_id, course_id):
...
+    comment_id = request.POST["id"]
+    action = request.POST["action"]
+    course_id = request.POST["course_id"]
+
+    course = Course.objects.get(pk=course_id)
+    answer = TextAnswer.objects.get(pk=comment_id)
+
+    if action == 'publish':
+        answer.publish()
+    elif action == 'private':
```
karyon: could you call that `make_private` as well?

karyon:
also, when editing a comment and saving it, you get back to the top of the page. which is good because you see the message then, but which is also bad because you have to scroll down to where you were before. these all are nitpicks of course, but i think they might worsen the user experience especially when reviewing large courses.

evap/staff/templates/staff_course_comments.html
```diff
@@ -1,33 +1,54 @@
...
+            </div>
+        </div>
+    {% endif %}
+    {% if contributor_sections %}
+        <div class="panel panel-info">
+            <div class="panel-heading">
+                <span class="panel-title">{% trans "Comments about the Contributors" %}</span>
+            </div>
+            <div class="panel-body">
+                {% regroup contributor_sections by questionnaire as sections_grouped %}
```
karyon: why the regroup here and above?

evap/staff/views.py
```diff
@@ -479,17 +426,72 @@ def course_comments(request, semester_id, course_id):
...
+        answer.make_private()
+    elif action == 'hide':
+        answer.hide()
+    elif action == 'unreview':
+        answer.unreview()
```
karyon: could you make this method fail loudly in an else here?

karyon:
could you add a test that tests the POST view you added with the various actions? that should be fairly easy.

karyon:
in the comment edit view, could you pre-fill the lower text field with the original answer if there is no reviewed answer yet? i think the first thing every user is going to do is copypasting the original answer, and also this suggests that you should enter nothing new but rather edit the original.

karyon:
that's it for now. i hope i didn't go overboard with all the comments.

evap/staff/templates/staff_course_comments_section.html
janno42: tried it. didn't like it. i thought not adding new buttons would be good.

evap/staff/forms.py
```diff
@@ -391,3 +360,18 @@ class Meta:
...
+        super().__init__(*args, **kwargs)
+        self.fields['original_answer'].widget.attrs['readonly'] = "True"
+
+    class Meta:
+        model = TextAnswer
+        fields = ("original_answer", "reviewed_answer",)
+
+    def clean_reviewed_answer(self):
+        reviewed_answer = self.cleaned_data.get('reviewed_answer')
+        if reviewed_answer == self.instance.original_answer or reviewed_answer == '':
```
karyon: the original code used normalize_newlines around reviewed/original_answer. could you check whether that might be necessary again, maybe do a blame on why that was added?

evap/staff/urls.py
```diff
@@ -23,12 +23,14 @@
     url(r"^semester/(\d+)/course/create$", course_create),
     url(r"^semester/(\d+)/course/(\d+)/edit$", course_edit),
     url(r"^semester/(\d+)/course/(\d+)/delete$", course_delete),
-    url(r"^semester/(\d+)/course/(\d+)/review$", course_review),
-    url(r"^semester/(\d+)/course/(\d+)/review/(\d+)$", course_review),
     url(r"^semester/(\d+)/course/(\d+)/email$", course_email),
     url(r"^semester/(\d+)/course/(\d+)/unpublish$", course_unpublish),
     url(r"^semester/(\d+)/course/(\d+)/preview$", course_preview),
     url(r"^semester/(\d+)/course/(\d+)/comments$", course_comments),
+    url(r"^semester/(\d+)/course/(\d+)/comments/(\w+)$", course_comments),
```
karyon: can't you use these ?foo=bar parameters? this looks like a new page, but it's not.

karyon:
the success/failure highlights are ace.

when you had a look at the last two comments, :shrimpit

evap/staff/forms.py
janno42: according to https://code.djangoproject.com/ticket/19251 our friends the linebreaks can be different in some situations. to be safe i'll add the normalization again.

evap/staff/urls.py
janno42: changed

karyon:
:shipit:

PR #533, opened 2015-03-03
headline fixes

janno42:
use consistent headline formatting all over the place


karyon:
wow so much consistency

:ship:

PR #540, opened 2015-03-24
Tooltip refactoring

janno42:
- use data-toggle="tooltip" instead of twipsify
- fix tooltips for disabled elements
- fix tooltips in course form
- use linebreaks for resultbar tooltips

### DIFF
--- a/evap/evaluation/templates/evap_course_edit_js.html
+++ b/evap/evaluation/templates/evap_course_edit_js.html
@@ -27,13 +27,6 @@
             };
         }
 
-        function twipsify() {
-            $(".twipsify").tooltip({
-                html: true,
-                placement: "left"
-            });
-        }
-
         function getContributorCount() { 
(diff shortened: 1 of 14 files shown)


evap/evaluation/templates/evap_course_edit_js.html
```diff
@@ -67,8 +59,7 @@
             return id || (name && name != "---------");
         };
         row_added = function(row) {
-            twipsify();
-            assignClickHandlers()
+            assignClickHandlers();
```
karyon: how do new rows get their tooltip?

janno42: it works with the changes in the base file

karyon: please elaborate. i was talking about the rows that are added via javascript, because the code in the base file is run only once.

karyon:
:+1:

PR #541, opened 2015-03-24
Questionnaire fixes

janno42:
- change questionnaire sorting (obsolete questionnaires should not always be at the end)
- fix obsolete label
- fix when questionnaires can be edited


karyon:
:ship:

PR #547, opened 2015-03-28
Random improvements

janno42:
see comment messages for details


janno42:
it's on the top right now. before it was on the bottom right when the message had more than one line. following the bootstrap documentation the close button has to be added in front of all other elements in the message.

karyon:
:shipit:

PR #548, opened 2015-03-28
Course selection in semester view

### LINKED ISSUE #407: select courses directly inside the semester overview
currently, when approving/resetting/publishing courses, one visits the semester overview, goes to the respective state, clicks on a button which brings up a new page where all courses of that state are selectable, selects some courses, and clicks another button.

proposed design: on the semester overview, there is a checkbox to the left of each course. when one checks some of the boxes and clicks on the approve/publish button, one gets to a page that says "please confirm that you want to approve/publish the following courses:" followed by a list of selected courses.

this allows the user to stay on the semester overview while selecting the courses, which is cool because he has the evaluation period or participant count available there. and it removes that second page which feels out of place.

janno42:
fix #407 

This allows selecting courses directly in the semester view when performing state changes.

selecting courses:
selection

confirm operation:
confirmation


evap/staff/templates/staff_semester_view.html
```diff
@@ -91,35 +91,36 @@ <h3 class="panel-title">{% trans "Courses" %}</h3>
...
+                                <button name="operation" value="prepare" type="submit" class="btn btn-default" {{ disable_if_archived }}>{% trans "Enable courses for lecturer review" %}</button>
+                            {% endifequal %}
+                            {% ifequal state "prepared" %}
+                                <button name="operation" value="revertToNew" type="submit" class="btn btn-default" {{ disable_if_archived }}>{% trans "Revert courses to new" %}</button>
+                            {% endifequal %}
+                            {% if state == 'new' or state == 'prepared' or state == 'lecturerApproved' %}
+                                <button name="operation" value="approve" type="submit" class="btn btn-default" {{ disable_if_archived }}>{% trans "Approve courses" %}</button>
+                            {% endif %}
+                            {% ifequal state "lecturerApproved" %}
+                                <button name="operation" value="reenableLecturerReview" type="submit" class="btn btn-default" {{ disable_if_archived }}>{% trans "Re-enable courses for lecturer review" %}</button>
```
karyon: we should rename all this lecturer stuff to something else, shouldnt we?

janno42: what do you have in mind?

karyon: well, responsible or contributor...

janno42: responsible sounds good

evap/staff/tests.py
```diff
@@ -304,16 +304,11 @@ def test_all_urls(self):
...
-            ("test_staff_semester_x_publish", "/staff/semester/1/publish", "evap"),
             # staff semester course
             ("test_staff_semester_x_course_y_edit", "/staff/semester/1/course/5/edit", "evap"),
             ("test_staff_semester_x_course_y_email", "/staff/semester/1/course/1/email", "evap"),
             ("test_staff_semester_x_course_y_preview", "/staff/semester/1/course/1/preview", "evap"),
             ("test_staff_semester_x_course_y_comments", "/staff/semester/1/course/5/comments", "evap"),
-            ("test_staff_semester_x_course_y_unpublish", "/staff/semester/1/course/8/unpublish", "evap"),
             ("test_staff_semester_x_course_y_delete", "/staff/semester/1/course/1/delete", "evap"),
             # staff questionnaires
             ("test_staff_questionnaire", "/staff/questionnaire/", "evap"),
```
karyon: we should decide whether we want to have this kind of test and then either keep the url list complete or remove it altogether. i vote for the former. if you agree, iirc there was a PR of yours that also did not add a new url, please find that out and add it as well.

evap/staff/tests.py
```diff
@@ -956,12 +939,9 @@ def test_raise_403(self):
 
         self.get_assert_403("/staff/semester/4/import", "evap")
         self.get_assert_403("/staff/semester/4/assign", "evap")
-        self.get_assert_403("/staff/semester/4/approve", "evap")
-        self.get_assert_403("/staff/semester/4/contributorready", "evap")
         self.get_assert_403("/staff/semester/4/course/create", "evap")
         self.get_assert_403("/staff/semester/4/course/7/edit", "evap")
         self.get_assert_403("/staff/semester/4/course/7/delete", "evap")
-        self.get_assert_403("/staff/semester/4/course/7/unpublish", "evap")
```
karyon: same as above

evap/staff/views.py
```diff
@@ -91,6 +90,100 @@ def semester_view(request, semester_id):
...
+def semester_course_operation(request, semester_id):
+    semester = get_object_or_404(Semester, id=semester_id)
+    raise_permission_denied_if_archived(semester)
+
+    course_ids = request.GET.getlist('course')
+    operation = request.GET.get('operation')
+    if operation not in ['revertToNew', 'prepare', 'reenableLecturerReview', 'approve', 'publish', 'unpublish']:
+        messages.error(request, _("Unsupported operation: ") + str(operation))
+        return custom_redirect('evap.staff.views.semester_view', semester_id)
+    courses = [get_object_or_404(Course, id=course_id) for course_id in course_ids]
```
karyon: i think a simple get (and thus a crash if the course does not exist) is more appropriate.

evap/staff/views.py
```diff
@@ -91,6 +90,100 @@ def semester_view(request, semester_id):
...
+                "Successfully enabled %(courses)d courses for lecturer review.", len(courses)) % {'courses': len(courses)})
+            try:
+                EmailTemplate.get_review_template().send_to_users_in_courses(courses, ['editors'])
+            except Exception:
+                messages.error(request, _("An error occured when sending the notification emails to the lecturers."))
+
+        elif operation == 'approve':
+            approved = 0
+            for course in courses:
+                if course.has_enough_questionnaires:
```
karyon: it shouldnt be possible to select these in the first place.

janno42: we can't just deactivate the checkboxes because in some states approving is just one of the possible operations. should i already remove these courses from the list on the confirmation page and display the error message there?

evap/staff/views.py
```diff
@@ -91,6 +90,100 @@ def semester_view(request, semester_id):
...
+                "Successfully published %(courses)d courses.", len(courses)) % {'courses': len(courses)})
+            for user, user_courses in user_publish_notifications(courses).items():
+                try:
+                    EmailTemplate.get_publish_template().send_to_user(user, courses=list(user_courses))
+                except Exception:
+                    messages.error(request, _("An error occured when sending the notification email to %s.") % user.username)
+
+        elif operation == 'unpublish':
+            for course in courses:
+                course.revoke()
```
karyon: if there is no semantic difference between revoke and unpublish, please decide on one of them and use that everywhere.

evap/staff/views.py
```diff
@@ -91,6 +90,100 @@ def semester_view(request, semester_id):
...
+                course.save()
+            messages.success(request, __("Successfully unpublished %(courses)d course.",
+                "Successfully unpublished %(courses)d courses.", len(courses)) % {'courses': len(courses)})
```
karyon: i suggest you put all the different operations into methods and keep only the if-elif construct and the method calls here

evap/staff/views.py
karyon: this is not a condition of the transition but should be imo. please check whether that is possible and then add the condition.

karyon: i think removing the courses from the list is better, yes.

evap/staff/views.py
janno42: check

evap/staff/views.py
janno42: check, using unpublish

evap/staff/tests.py
janno42: check, list will be updated

evap/staff/tests.py
janno42: check

evap/staff/views.py
janno42: check

evap/staff/views.py
janno42: check

evap/staff/templates/staff_semester_view.html
janno42: we should do the renaming outside of this pr: #549

karyon:
:shipit:

PR #553, opened 2015-04-02
Result refactoring

karyon:
lots of refactoring. the only functional change is that empty sections and questions without answers are not removed anymore.

best reviewed per commit, i think.


janno42:
Please make sure that contributors can only see the voting results if they have been published and add "No answers" to those text questions, where no answers exist.
I think then you are ready to go.

janno42:
:+1:

PR #561, opened 2015-04-03
replace the old anonymizer with a new management command

### LINKED ISSUE #556: rewrite anonymizer in python
this should result in a lot less complexity in there. maybe make it a management command.

karyon:
fixes #556


janno42:
:+1:

PR #565, opened 2015-04-06
Better results caching

### LINKED ISSUE #475: calculate results upon publishing, cache indefinitely
currently, results are calculated when a user wants to view them, and they are cached for one day. that means a lot of people will need to wait ~10s for the results.

karyon:
see commit messages. fixes #475.


janno42:
:+1:

PR #566, opened 2015-04-06
Add colors to average grades on results page

### LINKED ISSUE #554: colorize average grades on course detail pages
the numbers in the lower right should be css'd like the ones in the upper left.

janno42:
fix #554

### DIFF
--- a/evap/evaluation/templates/result_bar.html
+++ b/evap/evaluation/templates/result_bar.html
@@ -1,0 +1,26 @@
+{% load morefilters %}
+{% load i18n %}
+
+{% if show_grades and result.count > 0 %}
+    {% spaceless %}
+    <span data-toggle="tooltip" data-placement="left"
+        title="{% for answer, rel in result.distribution.items %}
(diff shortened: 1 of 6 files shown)


janno42:
looks like this:
color_average

evap/evaluation/templates/result_bar.html
```diff
@@ -0,0 +1,26 @@
...
+{% if show_grades and result.count > 0 %}
+    {% spaceless %}
+    <span data-toggle="tooltip" data-placement="left"
+        title="{% for answer, rel in result.distribution.items %}
+            {% if result.question.is_likert_question %}{{ answer|likertname }}{% else %}{{ answer }}{% endif %}: {{ rel|floatformat }}%{% if not forloop.last %}</br>{% endif %}
+        {% endfor %}">
+        
+        <div class="grade-bg-result-bar text-center {% if questionnaire_warning or result.warning %}participants-warning{% endif %}" style="background-color: {{ result.average|gradecolor }}">
+            {{ result.average|floatformat:1 }}
+        </div>
```
karyon: why is the code for the average before the distribution code?

janno42: because the distribution bar should take up the remaining space. the layout doesn't work when the fixed-size element doesn't come first. it's the same for the participation bar.

karyon:
:ship:

PR #567, opened 2015-04-08
Django18 prep

karyon:
some cleanups and django 1.8 preparations to prevent them from bitrotting.

the actual update needs to wait as there is a bug (https://code.djangoproject.com/ticket/24513) that makes tests fail randomly, which is fixed in 1.8.1.

5985864 speeds up the tests, from 42s to 32s here.


janno42:
5a002640 is great!

karyon:
because evaluation didn't have a urls.py and i didn't bother to change that.

janno42:
:+1:

karyon:
added a urls.py for the evaluation app, :ship:?

janno42:
:shipit:

Issue #315, opened 2014-06-11
Introduce Course Type model to enable translations and easier changes


karyon:
course types are not translatable.

maybe this should be solved at the import as an additional column, analogue to the german and english course names. alternatively, one could provide a GUI for adding translations after the import. i vote for the former, it seems more consistent.

ref #250

karyon:
turns out there are a lot of different possible solutions. i may have overdone it a bit...

 * if one simply made Course.type a `Translate` field, that ... doesn't feel good
   * Easy to get unclean data.
   * Course type merging would operate on a bunch of strings. That might work ok

 * We also want to be able to easily merge course types.
   * This is used to simplify course types for publishing. When importing courses, tpyes may be very verbose (Lecture, Lecture with exercise, lecture with seminar etc), which helps when bulk assigning questionnaires, but for publishing, one may only want to display "lecture" in this case.

So, first idea is to introduce CourseType model, which contains one Translate field. One open question is whether both languages are mandatory or not.
 * If yes, the translation must be provided with the import.
   * The problem here is that one might have to specify translations for course types that are going to be merged away anyway.
   * On the other hand there are not going to be many course types overall, so the additional work is minimal and must be done only once as types are not going to be deleted.
   * The import contains two columns, one of which may be empty if the database already contains the course type.
 * If no, then the question is whether the german or english translation should be mandatory or we permit any of them to exist alone.
   * One is mandatory: The import contains one column for the course type, or two, the second one of which may be empty.
     * we will need a setting to specify which language is the mandatory one. but then we can change that at runtime, so this also boils down to the following
   * either may exist alone: The import should contain two columns, and in each row one of them may be empty.
     * this might lead to some courses having only german, and some only english names, so users of both languages might see missing translations.
   * Problem of both solutions
     * the staff views will not always have translated course types
     * we will have to display warning in the staff view "translation for course type missing" before moving courses to "prepared"
     * The course type merge UI will have to use the other language if the translation for the active one is missing. that might be a bit of code complexity

Another solution might be to let the course type have a verbose, imported name with unspecified language, and a translatable display name.
 * str() of this can use the short translated names if available or fallback to the import name (if the translations are optional)
 * Existing column in the import is the import name.
   * if the translations are mandatory, we add two columns for the display names, which may be empty if the DB contains this type already
   * if they are optional, we might add these columns as well and make them always optional.
     * in this case we need a warning "translations missing" again.
[...]

PR #568, opened 2015-04-17
rename course.kind and question.kind to type

karyon:
renaming, that's all this PR does. i started working on #315 and this is all i managed.


janno42:
rename all the kinds! :+1:

PR #569, opened 2015-04-18
Introduce Select2

### LINKED ISSUE #346: Search field for contributor selection

karyon:
this PR replaces the multiselect plugin with select2, and uses that to fix #346. thanks to @PFischbeck for starting #529.

notes:
 * there is one UI issue left, see screenshot. @janno42, could you have a look at this?
 * there is a small delay when the contributor list opens. here it is acceptable, but on production with more users and for people with slower machines it might be worse.
 * i removed .map and non-minified js files. while they are cool to have we never needed them and i'd rather solve this with a proper asset pipeline.
 * afaik we use jqueryui only for the datepicker now. i considered downloading a custom jqueryui with only the datepicker or replacing it with a bootstrap plugin, but thought it wasn't important enough.


janno42:
when adding a new contributor to the course the select element does nothing when clicking on it

karyon:
should be finished now. @janno42 i changed your commits so please review these as well.

12 hours ago select2 4.0 final came out, that's included as well.

janno42:
Travis failed...

janno42:
After the tests are fixed, you can :ship: it!

PR #570, opened 2015-04-19
Replace median by variance

### LINKED ISSUE #558: replace median by variance
in the export, result overview and course detail pages, we show a total median rating for the course. we realized that we don't know what this should actually mean, but that an average over all variances would be helpful. so these three places, the median should be replaced by the variance.

karyon:
fixes #558.
see screenshot. are the colors ok? or is the 2.2. to dark? a problem might be the grey background of every second line.


janno42:
the 2.2 is definitely too dark. the gray background of every second line is a problem.

karyon:
two commits added, please review.

PR #574, opened 2015-04-24
Fix headlines

janno42:
This fixes the headline issues on mobile devices and removes one unused template.


karyon:
the conversion is not consistent, large is sometimes h1 and h2, without class is h2 and h3, and small is h3 and h4. if that is intended, :ship:

janno42:
i decided for each headline individually.

PR #575, opened 2015-04-24
fix #550; add "select all" option in semester view

### LINKED ISSUE #550: Select all option in semester overview
The course lists in the semester view should get a select all option.

janno42:
adding a "select all" button


karyon:
:dowit

Issue #218, opened 2014-01-08
UI Design issues


karyon:
this issue is meant to be more of a meta-issue, feel free to add more issues in comments or in the list.
- when using the search function in datables, it makes the page (primarily the results page) move as the table may get smaller which can cause the page to scroll up. this can be confusing. see #369 for details.

janno42:
(see #491) The head of table shown on the questionnare overview page is not correctly aligned to the table body. see the attached image
![wrong_align](https://cloud.githubusercontent.com/assets/2399349/5766184/c3364804-9cff-11e4-84a1-976918948fab.png)

janno42:
The light gray rows on the voting page (and maybe other places) have a white rectangle when the line gets too high. it's slightly visible under the vote buttons in the image below:
![capture](https://cloud.githubusercontent.com/assets/1781719/5884524/1d32e9d8-a35f-11e4-89d9-6ff9c82cb88a.PNG)

janno42:
if warnings have more than one line of text the close button is displayed too far on the bottom. it should be on the height of the first line of text.
![error message](https://cloud.githubusercontent.com/assets/1781719/5884528/422bd0d8-a35f-11e4-85a8-822ede21d3cd.PNG)

janno42:
when displaying the platform in german the action buttons don't fit in one line and have no margin between the two rows:

![actions](https://cloud.githubusercontent.com/assets/1781719/6881140/f5f11c16-d552-11e4-8e6a-28b7417152d2.PNG)

janno42:
The questionnaire tooltips are too wide for smaller screens:
![untitled](https://cloud.githubusercontent.com/assets/1781719/7349266/1d799292-ecf8-11e4-8ad3-162d2301e469.png)

karyon:
![2015 05 07 19_17_08-000036](https://cloud.githubusercontent.com/assets/1891915/7521464/3bd88326-f4ee-11e4-94e4-e84f7f14ebf3.png)
see screenshot. this is in staff_questionnaire_form.html. @janno42, could you fix this?

janno42:
The result pages must show the semester name. Currently this name is only visible in the breadcrumbs, which are not displayed for students.

karyon:
lets discuss whether there will ever be a solution for this.

janno42:
[...]

PR #581, opened 2015-04-26
Move semester related buttons

janno42:
contributes to #218
semester related buttons were moved next to a new headline, so the remaining buttons will fit into one line:
capture


karyon:
cool :+1:

PR #582, opened 2015-05-07
Django 1.8

karyon:
1.8.1 was released with a fix that made our tests run again, so here is django 1.8.!

highlight are probably the faster tests, they are down to 22s (from 32s)


janno42:
go for it :+1:
(and thanks for the good work)

PR #584, opened 2015-05-14
Introduce SingleResults (Einzelergebnisse)

janno42:
this pull request makes the following changes:
- combining all LikertAnswers and GradeAnswers with the same contribution, question and answer into single LikertAnswerCounters and GradeAnswerCounters =)
- adding a form and views for creating and editing SingleResults, which are special courses without participants, a single event date and only one predefined question for the one responsible person. the form allows manually editing the answers for these courses.
- updating views and templates for SingleResults. for example the semester results page directly shows a result bar with the distribution of these courses.

and it does some general cleaning:
- removing unused statistics command
- removing complete message texts from email_templates

example from the semester results page:

capture


janno42:
I'm not quite sure about why Travis failed and would be happy if someone has an idea.

evap/staff/forms.py
```diff
@@ -95,6 +95,79 @@ def validate_unique(self):
             self._update_errors(e)
 
 
+class SingleResultForm(forms.ModelForm, BootstrapMixin):
+    last_modified_time_2 = forms.DateTimeField(label=_("Last modified"), required=False, localize=True)
+    last_modified_user_2 = forms.CharField(label=_("Last modified by"), required=False)
+    event_date = forms.DateField(label=_("Event date"), localize=True)
+    responsible = forms.ChoiceField(choices=[(u.username, u) for u in UserProfile.objects.order_by('username')], label=_("Responsible"))
```
karyon: this happens at module load time, that might be the problem of the travis failure.

evap/evaluation/models.py
```diff
@@ -522,26 +535,34 @@ class Meta:
...
+class LikertAnswerCounter(Answer):
     """A Likert-scale answer to a question with `1` being *strongly agree* and `5`
     being *strongly disagree*."""
 
     answer = models.IntegerField(verbose_name=_("answer"))
+    count = models.IntegerField(verbose_name=_("count"), default=0)
 
     class Meta:
         verbose_name = _("Likert answer")
         verbose_name_plural = _("Likert answers")
```
karyon: could you try ensuring uniqueness of an answer value for one question? like, there should be no two answercounters with the same answer for the same question. should be possible somehow with the "unique" attribute of the meta class.

evap/staff/forms.py
janno42: local tests and actually using the form work without problems. do you have any recommendation what could be changed?

karyon: the [ModelChoiceField](https://docs.djangoproject.com/en/1.8/ref/forms/fields/#fields-which-handle-relationships) looks more appropriate and probably solves this.

janno42:
The ModelChoiceField works perfectly, thanks! I will have a look at the other stuff later today.

karyon:
more general remarks:

have you considered creating all five counters for each question upfront and not on demand when saving a vote? i think there a few places where the code would be a bit more straightforward, and it would fit my mental model of such counters better.

i still think it's weird to separate grade results and likert results in the model, i think of them as different "skins" for the same data... what do you think?

karyon:
finished reviewing btw

janno42:
addressing your general remarks: isn't that a perfect task for people like @karyon? ;)

karyon:
the second one maybe, the first one i'd like to see handled in this PR if we decide that it's the better solution.

janno42:
i'm for the proposed second change.
i don't like creating counters for all possible answers. if we would do it, the only place where it would make sense is when the evaluation period for a course starts. but i don't see a real benefit from it.

janno42:
i refactored all mentioned things. please review.

karyon:
please check whether the overall grade calculation still makes sense when we have courses that have no likert questions at all.

janno42:
the grade calculation doesn't care about whether likert questions exist or not.

karyon:
please add one (or more) single result to test_data and minimal_test_data, and add urls that accessto the url test and add tests for the create and edit forms with valid and invalid data. should be easy to copypaste that from the other form tests.

karyon:
finished reviewing.

janno42:
updating test_data will not be done in this pr and is now issue #587

janno42:
another update :)

karyon:
:+1:

PR #592, opened 2015-05-31
Check whether email already exists

### LINKED ISSUE #590: Editing email of user to existing email causes Internal Server Error
When editing a user and setting the email address to one that is already used by another user in the system causes the following error:

```
Internal Server Error: /staff/user/create Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
psycopg2.IntegrityError: duplicate key value violates unique constraint "evaluation_userprofile_email_key"
DETAIL:  Key (email)=(a@b.com) already exists.
```

The form should check the address during cleaning and show an error in the UI.

janno42:
fix #590, validating unique before saving instance

### DIFF
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -367,19 +367,9 @@ class UserForm(forms.ModelForm, BootstrapMixin):
                                                                           required=False)
         self.fields['courses_participating_in'].help_text = ""
 
-    def clean_username(self):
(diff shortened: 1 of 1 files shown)


karyon:
nice! :shipit:

janno42:
replaced by #595

Issue #589, opened 2015-05-22
Exclude courses from reward points requirement


janno42:
Courses should have an option making them not being required for getting reward points for the respective semester.

PR #594, opened 2015-05-31
Add option to define courses as optional for rewards

janno42:
implement #589, enable defining courses as optional for rewards

### DIFF
--- a/evap/rewards/views.py
+++ b/evap/rewards/views.py
@@ -65,10 +65,13 @@ def semester_reward_points(request, semester_id):
 
     data = []
     for participant in participants:
-        number_of_courses = Course.objects.filter(semester=semester, participants=participant).count()
-        number_of_courses_voted_for = Course.objects.filter(semester=semester, voters=participant).count()
+        number_of_required_courses = Course.objects.filter(semester=semester, participants=participant, is_required_for_reward=True).count()
+        number_of_required_courses_voted_for = Course.objects.filter(semester=semester, voters=participant, is_required_for_reward=True).count()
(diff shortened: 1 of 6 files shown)


evap/rewards/views.py
```diff
@@ -65,10 +65,13 @@ def semester_reward_points(request, semester_id):
...
-        number_of_courses_voted_for = Course.objects.filter(semester=semester, voters=participant).count()
+        number_of_required_courses = Course.objects.filter(semester=semester, participants=participant, is_required_for_reward=True).count()
+        number_of_required_courses_voted_for = Course.objects.filter(semester=semester, voters=participant, is_required_for_reward=True).count()
+        number_of_optional_courses = Course.objects.filter(semester=semester, participants=participant, is_required_for_reward=False).count()
+        number_of_optional_courses_voted_for = Course.objects.filter(semester=semester, voters=participant, is_required_for_reward=False).count()
         earned_reward_points = RewardPointGranting.objects.filter(semester=semester, user_profile=participant).exists()
-        data.append((participant, number_of_courses_voted_for, number_of_courses, earned_reward_points))
+        data.append((participant, number_of_required_courses_voted_for, number_of_required_courses,
+            number_of_optional_courses_voted_for, number_of_optional_courses, earned_reward_points))
```
karyon: this smells like (#num participants in semester) * 5 queries, i'd expect this page to be quite slow on production. how frequently is it used? do you think the performance is acceptable?

janno42: even on the test system it load way faster than the user list. and the reward points view is not used very often, so...

karyon:
apart from the performance concern :+1:, so that's your decision.

PR #595, opened 2015-05-31
Model mommy and various

### LINKED ISSUE #590: Editing email of user to existing email causes Internal Server Error

karyon:
introduces model_mommy and replaces some fixtures with it, fixes #590, some minor cleanups/fixes.


janno42:
sorry for the migration conflict ;)

evap/staff/forms.py
```diff
@@ -368,32 +367,41 @@ def __init__(self, *args, **kwargs):
...
+        return self.cleaned_data['first_name'].strip()
+
+    def clean_last_name(self):
+        return self.cleaned_data['last_name'].strip()
 
-        super()._post_clean(*args, **kw)
+    def clean_title(self):
+        return self.cleaned_data['title'].strip()
+
+    def save(self):
```
janno42: why not ```save(self, *args, **kw)```?

karyon: because that would be the right thing to do!

karyon:
updated and two more commits added, please review.

evap/settings.py
```diff
@@ -252,6 +252,9 @@ def show_toolbar(request):
         'SHOW_TOOLBAR_CALLBACK': 'evap.settings.show_toolbar',
     }
 
+MOMMY_CUSTOM_FIELDS_GEN = {
```
janno42: maybe a comment here?

janno42:
looks good now! you could add the one comment in the settings and then... :shipit:

PR #597, opened 2015-06-03
Allow switching to public view in results

### LINKED ISSUE #588: Add "public view" on results pages for staff users
For staff users there should be an option on the results pages to switch between "public view" (without comments, like students would see the page) and "complete view" (the current view including all comments).

janno42:
fix #588

This allows changing to the public view on a results page if a user is staff, contributor or a delegate. In public view an information message is shown above the results to explain what happened.

### DIFF
--- a/evap/results/views.py
+++ b/evap/results/views.py
@@ -72,11 +72,14 @@ def course_detail(request, semester_id, course_id):
 
     sections = calculate_results(course)
 
+    public_view = request.GET.get('public_view', 'false') # default: show own view
+    public_view = {'true': True, 'false': False}.get(public_view.lower()) # convert parameter to boolean
+
     for section in sections:
         results = []
         for result in section.results:
             if isinstance(result, TextResult):
-                answers = [answer for answer in result.answers if user_can_see_text_answer(request.user, answer)]
+                answers = [answer for answer in result.answers if user_can_see_text_answer(request.user, answer, public_view)]
                 if answers:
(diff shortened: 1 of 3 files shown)


evap/templates/base.html
```diff
@@ -127,8 +127,10 @@
         {% block rawcontent %}
             <div class="row">
                 <div class="col-md-12">
-                    {% block content %}
-                    {% endblock %}
+                    <div>
```
janno42: this change moves the buttons next to the heading into a nicer position. they are now aligned with the text.

evap/results/views.py
```diff
@@ -72,11 +72,14 @@ def course_detail(request, semester_id, course_id):
...
 
+    public_view = request.GET.get('public_view', 'false') # default: show own view
+    public_view = {'true': True, 'false': False}.get(public_view.lower()) # convert parameter to boolean
+
     for section in sections:
         results = []
         for result in section.results:
             if isinstance(result, TextResult):
-                answers = [answer for answer in result.answers if user_can_see_text_answer(request.user, answer)]
+                answers = [answer for answer in result.answers if not public_view and user_can_see_text_answer(request.user, answer)]
```
karyon: how about passing a parameter `public` into `user_can_see_text_answer` and returning false from there? it would be cool to keep the decision whether a comment is displayed or not in one place. but then it needs to have a new name; i didn't find one but maybe you do.

janno42:
updated

karyon:
:+1:

PR #600, opened 2015-06-15
Fix has enough questionnaires

### LINKED ISSUE #585: Multiple empty general contributions
When saving a course form it makes sure that a general contribution exists. However, this can result in adding multiple empty contributions without questionnaires. When looking at a course in the Django Admin one can see these contributions.
This results in unwanted side effects (see https://github.com/janno42/EvaP/commit/8d5f9a8e3ac436757ca046881e7a74b55eafeb98#commitcomment-11211525).

These contributions should be removed or at least it has to be made sure that there will be at most one and that it doesn't interfere with other logic.

karyon:
kinda fixes #585. the real problem weren't multiple general contributions (there weren't any on evap-dev), but screwed logic in Course.has_enough_questionnaires.

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -196,7 +196,7 @@ class Course(models.Model, metaclass=LocalizeModelBase):
     course_evaluated = django.dispatch.Signal(providing_args=['request', 'semester'])
 
     class Meta:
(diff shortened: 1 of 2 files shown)


janno42:
:shipit:

Issue #587, opened 2015-05-17
Create new test_data


janno42:
The test_data used for development should be updated. In the following some missing items are listed, but this list might not be complete:
- [x] RewardPoints (Grantings, Redemptions and Events)
- [x] SingleResults
- [x] private comments
- [x] GradeDocuments
  - incl. grade publishers and maybe even some actual grade document files in e.g. the fixtures dir
- [x] updated EmailTemplates
- [x] add GradeQuestions and courses with GradeAnswerCounters

PR #603, opened 2015-06-21
Grade publishing

### LINKED ISSUE #352: Publish exam grades on EvaP
Publishing exam grades should be connected to EvaP. Currently the university management is uploading PDF files to a certain place - these documents should then be uploaded to EvaP instead.
These documents can be linked to courses and once they are uploaded and the evaluation of this course is complete, the evaluation results can be automatically published.
In addition the platform can send a message to all voters that the grades and evaluation results are now available on the platform. Every voter should be able to opt-out from this message in the respective questionnaire of the course.
Other participants won't get a notification - this will probably again result in more voters.

janno42:
fix #352

This adds functionality for uploading and downloading grade documents for courses, along with changes for the automated publishing of courses and updates to the publish notification emails.

Tests are still missing and will soon be implemented, but this pull request should already be complete regarding the functionality and allows to review the changes.

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -830,6 +830,10 @@ class UserProfile(AbstractBaseUser, PermissionsMixin):
             return True
         return is_external_email(self.email)
 
+    @property
+    def can_download_grades(self):
+        return not self.is_external
+
     @classmethod
     def email_needs_login_key(cls, email):
         # do the import here to prevent a circular import
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -477,14 +477,10 @@ class Course(models.Model, metaclass=LocalizeModelBase):
                     course.evaluation_end()
                     if course.is_fully_reviewed():
                         course.review_finished()
-                        if not course.is_graded:
+                        if not course.is_graded or course.final_grade_documents.exists():
                             course.publish()
                             evaluation_results_courses.append(course)
-                        elif course.final_grade_documents.exists():
-                            course.publish()
-                            grade_document_courses.append(course)
-                            evaluation_results_courses.append(course)
-                    elif course.final_grade_documents.exists():
+                    if course.final_grade_documents.exists():
                         grade_document_courses.append(course)
                     course.save()
             except Exception:
@@ -971,19 +967,19 @@ class EmailTemplate(models.Model):
         else:
             cc_addresses = []
 
-        grade_documents = len(grade_document_courses) > 0
-        evaluation_results = len(evaluation_results_courses) > 0
+        grade_documents_exist = len(grade_document_courses) > 0
+        evaluation_results_exist = len(evaluation_results_courses) > 0
 
         mail = EmailMessage(
             subject = self.render_string(self.subject, {
-                'grade_documents': grade_documents,
-                'evaluation_results': evaluation_results
+                'grade_documents_exist': grade_documents_exist,
+                'evaluation_results_exist': evaluation_results_exist
             }),
             body = self.render_string(self.body, {
                 'user': user,
                 'courses': courses,
-                'grade_documents': grade_documents,
-                'evaluation_results': evaluation_results,
+                'grade_documents_exist': grade_documents_exist,
+                'evaluation_results_exist': evaluation_results_exist,
                 'grade_document_courses': grade_document_courses,
                 'evaluation_results_courses': evaluation_results_courses
             }),
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -460,7 +460,7 @@ class Course(models.Model, metaclass=LocalizeModelBase):
 
     @classmethod
     def update_courses(cls):
-        from evap.evaluation.tools import send_publish_notifications, send_evaluation_started_notifications
+        from evap.evaluation.tools import send_publish_notifications
         today = datetime.date.today()
 
         courses_new_in_evaluation = []
@@ -483,7 +483,7 @@ class Course(models.Model, metaclass=LocalizeModelBase):
             except Exception:
                 logging.getLogger(__name__).exception(('An error occured when updating the state of course "{}".').format(course.name))
 
-        send_evaluation_started_notifications(courses_new_in_evaluation)
+        EmailTemplate.send_evaluation_started_notifications(courses_new_in_evaluation)
         send_publish_notifications(evaluation_results_courses=evaluation_results_courses)
 
 
@@ -899,25 +899,22 @@ class EmailTemplate(models.Model):
     subject = models.CharField(max_length=1024, verbose_name=_("Subject"), validators=[validate_template])
     body = models.TextField(verbose_name=_("Body"), validators=[validate_template])
 
-    @classmethod
-    def get_review_template(cls):
-        return cls.objects.get(name="Editor Review Notice")
+    EDITOR_REVIEW_NOTICE = "Editor Review Notice"
+    STUDENT_REMINDER = "Student Reminder"
+    PUBLISHING_NOTICE = "Publishing Notice"
+    LOGIN_KEY_CREATED = "Login Key Created"
+    EVALUATION_STARTED = "Evaluation Started"
 
-    @classmethod
-    def get_reminder_template(cls):
-        return cls.objects.get(name="Student Reminder")
 
-    @classmethod
-    def get_publish_template(cls):
-        return cls.objects.get(name="Publishing Notice")
 
-    @classmethod
-    def get_login_key_template(cls):
-        return cls.objects.get(name="Login Key Created")
+    EMAIL_RECIPIENTS = (
+        ('all_participants', _('all participants')),
+        ('due_participants', _('due participants')),
+        ('responsible', _('responsible person')),
+        ('editors', _('all editors')),
+        ('contributors', _('all contributors'))
+    )
 
-    @classmethod
-    def get_evaluation_started_template(cls):
-        return cls.objects.get(name="Evaluation Started")
 
     @classmethod
     def recipient_list_for_course(cls, course, recipient_groups):
@@ -938,11 +935,13 @@ class EmailTemplate(models.Model):
 
         return recipients
 
+
     @classmethod
-    def render_string(cls, text, dictionary):
+    def __render_string(cls, text, dictionary):
         return Template(text).render(Context(dictionary, autoescape=False))
 
-    def send_to_users_in_courses(self, courses, recipient_groups):
+    @classmethod
+    def send_to_users_in_courses(self, template, courses, recipient_groups):
         user_course_map = {}
         for course in courses:
             responsible = course.responsible_contributor
@@ -951,9 +950,13 @@ class EmailTemplate(models.Model):
                     user_course_map.setdefault(user, []).append(course)
 
         for user, courses in user_course_map.items():
-            self.send_to_user(user, courses)
+            subject_params = {}
+            body_params = {'user': user, 'courses': courses}
+            self.__send_to_user(user, template, subject_params, body_params, cc=True)
 
-    def send_to_user(self, user, courses=[], grade_document_courses=[], evaluation_results_courses=[], cc=True):
+
+    @classmethod
+    def __send_to_user(cls, user, template, subject_params, body_params, cc):
         if not user.email:
             logging.getLogger(__name__).warning("{} has no email address defined. Could not send email.".format(user.username))
             return
@@ -964,45 +967,66 @@ class EmailTemplate(models.Model):
         else:
             cc_addresses = []
 
-        grade_documents_exist = len(grade_document_courses) > 0
-        evaluation_results_exist = len(evaluation_results_courses) > 0
-
-        mail = EmailMessage(
-            subject = self.render_string(self.subject, {
-                'grade_documents_exist': grade_documents_exist,
-                'evaluation_results_exist': evaluation_results_exist
-            }),
-            body = self.render_string(self.body, {
-                'user': user,
-                'courses': courses,
-                'grade_documents_exist': grade_documents_exist,
-                'evaluation_results_exist': evaluation_results_exist,
-                'grade_document_courses': grade_document_courses,
-                'evaluation_results_courses': evaluation_results_courses
-            }),
-            to = [user.email],
-            cc = cc_addresses,
-            bcc = [a[1] for a in settings.MANAGERS],
-            headers = {'Reply-To': settings.REPLY_TO_EMAIL})
-        
-        try:
-            mail.send(False)
-        except Exception:
-            logging.getLogger(__name__).exception(('An error occured when sending email "{}" to {}.').format(subject, user.username))
-
-    @classmethod
-    def send_reminder_to_user(cls, user, due_in_number_of_days, due_courses):
-        if not user.email:
-            return
-
-        template = cls.get_reminder_template()
-        subject = template.render_string(template.subject, {'user': user, 'due_in_number_of_days': due_in_number_of_days})
-        body = template.render_string(template.body, {'user': user, 'due_in_number_of_days': due_in_number_of_days, 'due_courses': due_courses})
+        subject = cls.__render_string(template.subject, subject_params)
+        body = cls.__render_string(template.body, body_params)
 
         mail = EmailMessage(
             subject = subject,
             body = body,
             to = [user.email],
+            cc = cc_addresses,
             bcc = [a[1] for a in settings.MANAGERS],
             headers = {'Reply-To': settings.REPLY_TO_EMAIL})
-        mail.send(False)
+
+        try:
+            mail.send(False)
+        except Exception:
+            logging.getLogger(__name__).exception(('An error occured when sending email "{}" to {}.').format(subject, user.username))
+
+
+    @classmethod
+    def send_reminder_to_user(cls, user, due_in_number_of_days, due_courses):
+        template = cls.objects.get(name=cls.STUDENT_REMINDER)
+        subject_params = {'user': user, 'due_in_number_of_days': due_in_number_of_days}
+        body_params = {'user': user, 'due_in_number_of_days': due_in_number_of_days, 'due_courses': due_courses}
+
+        cls.__send_to_user(user, template, subject_params, body_params, cc=False)
+
+    @classmethod
+    def send_login_key_to_user(cls, user):
+        template = cls.objects.get(name=cls.LOGIN_KEY_CREATED)
+        subject_params = {}
+        body_params = {'user': user}
+
+        cls.__send_to_user(user, template, subject_params, body_params, cc=False)
+
+    @classmethod
+    def send_publish_notifications_to_user(cls, user, grade_document_courses=[], evaluation_results_courses=[]):
+        template = cls.objects.get(name=cls.PUBLISHING_NOTICE)
+
+        grade_documents_exist = len(grade_document_courses) > 0
+        evaluation_results_exist = len(evaluation_results_courses) > 0
+
+        subject_params = {
+                'grade_documents_exist': grade_documents_exist,
+                'evaluation_results_exist': evaluation_results_exist
+            }
+        body_params = {
+                'user': user,
+                'grade_documents_exist': grade_documents_exist,
+                'evaluation_results_exist': evaluation_results_exist,
+                'grade_document_courses': grade_document_courses,
+                'evaluation_results_courses': evaluation_results_courses
+            }
+
+        cls.__send_to_user(user, template, subject_params, body_params, cc=True)
+
+    @classmethod
+    def send_review_notifications(cls, courses):
+        template = cls.objects.get(name=cls.EDITOR_REVIEW_NOTICE)
+        cls.send_to_users_in_courses(template, courses, ['editors'])
+
+    @classmethod
+    def send_evaluation_started_notifications(cls, courses):
+        template = cls.objects.get(name=cls.EVALUATION_STARTED)
+        cls.send_to_users_in_courses(template, courses, ['all_participants'])
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -464,7 +464,6 @@ class Course(models.Model, metaclass=LocalizeModelBase):
         today = datetime.date.today()
 
         courses_new_in_evaluation = []
-        grade_document_courses = []
         evaluation_results_courses = []
 
         for course in cls.objects.all():
@@ -480,14 +479,12 @@ class Course(models.Model, metaclass=LocalizeModelBase):
                         if not course.is_graded or course.final_grade_documents.exists():
                             course.publish()
                             evaluation_results_courses.append(course)
-                    if course.final_grade_documents.exists():
-                        grade_document_courses.append(course)
                     course.save()
             except Exception:
                 logging.getLogger(__name__).exception(('An error occured when updating the state of course "{}".').format(course.name))
 
         send_evaluation_started_notifications(courses_new_in_evaluation)
-        send_publish_notifications(grade_document_courses=grade_document_courses, evaluation_results_courses=evaluation_results_courses)
+        send_publish_notifications(evaluation_results_courses=evaluation_results_courses)
 
 
 class Contribution(models.Model):
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -778,7 +778,7 @@ class UserProfile(AbstractBaseUser, PermissionsMixin):
     def is_active(self):
         return True
 
-    @cached_property
+    @property
     def is_staff(self):
         return self.groups.filter(name='Staff').exists()
 
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -454,7 +454,7 @@ class Course(models.Model, metaclass=LocalizeModelBase):
         return self.grade_documents.filter(type=GradeDocument.FINAL_GRADES)
 
     @property
-    def preliminary_grade_documents(self):
+    def midterm_grade_documents(self):
         from evap.grades.models import GradeDocument
         return self.grade_documents.exclude(type=GradeDocument.FINAL_GRADES)
     
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -957,7 +957,11 @@ class EmailTemplate(models.Model):
             cc = cc_addresses,
             bcc = [a[1] for a in settings.MANAGERS],
             headers = {'Reply-To': settings.REPLY_TO_EMAIL})
-        mail.send(False)
+        
+        try:
+            mail.send(False)
+        except Exception:
+            logging.getLogger(__name__).exception(('An error occured when sending email "{}" to {}.').format(subject, user.username))
 
     @classmethod
     def send_reminder_to_user(cls, user, due_in_number_of_days, due_courses):
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -457,7 +457,43 @@ class Course(models.Model, metaclass=LocalizeModelBase):
     def midterm_grade_documents(self):
         from evap.grades.models import GradeDocument
         return self.grade_documents.exclude(type=GradeDocument.FINAL_GRADES)
-    
+
+    @classmethod
+    def update_courses(cls):
+        from evap.evaluation.tools import send_publish_notifications
+        today = datetime.date.today()
+
+        courses_new_in_evaluation = []
+        grade_document_courses = []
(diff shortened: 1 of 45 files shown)


evap/evaluation/models.py
```diff
@@ -447,6 +448,17 @@ def is_archiveable(self):
...
 
+    @property
+    def final_grade_documents(self):
+        from evap.grades.models import GradeDocument
+        return self.grade_documents.filter(type=GradeDocument.FINAL_GRADES)
+
+    @property
+    def preliminary_grade_documents(self):
+        from evap.grades.models import GradeDocument
+        return self.grade_documents.exclude(type=GradeDocument.FINAL_GRADES)
```
karyon: `FINAL_GRADES`?

janno42: exclude. i did this in case there might be some more types in the future.

evap/grades/admin.py
```diff
@@ -0,0 +1,10 @@
+from django.contrib import admin
+from evap.grades.models import GradeDocument
+
+class GradeDocumentAdmin(admin.ModelAdmin):
+    model = GradeDocument
+    list_display = ('__str__', 'course', 'description', 'type')
```
karyon: please use `__unicode__`, that's the preferred method in python 3

evap/grades/forms.py
```diff
@@ -0,0 +1,45 @@
...
+        super(StrippedCharField, self).to_python(value)
+        if value is None:
+            return None
+        return value.strip()
+
+
+class GradeDocumentForm(forms.ModelForm, BootstrapMixin):
+    description = StrippedCharField(max_length=255)
+    last_modified_time_2 = forms.DateTimeField(label=_("Last modified"), required=False, localize=True)
+    last_modified_user_2 = forms.CharField(label=_("Last modified by"), required=False)
```
karyon: could you please figure out why this _2 stuff is necessary and if it's not, remove it here and wherever you copied it from?

evap/grades/templates/grades_delete.html
```diff
@@ -0,0 +1,13 @@
+{% extends "grades_course_base.html" %}
+
+{% load i18n %}
+
+{% block content %}
+    {{ block.super }}
+    <form method="POST" style="display: inline;">
+        {% csrf_token %}
+        <p>{% blocktrans with semester=semester.name course=course.name name=grade_document.description %}Do you really want to delete the grade document <em>{{ name }}</em> for the course {{ course }} ({{ semester }})?{% endblocktrans %}</p>
```
karyon: please add some line breaks

evap/grades/templates/grades_semester_base.html
```diff
@@ -0,0 +1,23 @@
...
+        {% endif %}
+    {% endif %}
+{% endblock %}
+
+{% block title %}
+    {% if semester %}
+        {{ semester.name }} -
+    {% else %}
+        {% trans "Semesters" %} -
+    {% endif %}
```
karyon: could you explain this part please?

evap/grades/templates/grades_semester_view.html
```diff
@@ -0,0 +1,85 @@
...
+                                        {{ degree }}{% if not forloop.last %}, {% endif %}
+                                    {% endfor %}
+                                </td>
+                                <td>{{ course.type }}</td>
+                                <td class="text-center">{% if has_final_grades %}<span class="glyphicon glyphicon-ok"></span>{% endif %}</td>
+                                <td>
+                                    <div class="btn-group">
+                                        <button type="button" class="btn btn-sm btn-default dropdown-toggle" data-toggle="dropdown" aria-expanded="false" {{ disable_if_archived }}>{% trans "Upload grades" %} <span class="caret"></span></button>
+                                        <ul class="dropdown-menu" role="menu">
+                                            <li><a href="{% url "grades:upload_grades" semester.id course.id %}">Preliminary grades</a></li>
```
karyon: are you sure preliminary is the right word? to me preliminary grades sounds like they are almost final but are checked again and might be changed.

evap/grades/views.py
```diff
@@ -0,0 +1,174 @@
...
+import os
+from sendfile import sendfile
+
+from evap.evaluation.auth import grade_publisher_required, grade_downloader_required
+from evap.evaluation.models import Semester, Contribution, Course
+from evap.grades.models import GradeDocument
+from evap.grades.forms import GradeDocumentForm
+from evap.evaluation.tools import send_publish_notifications
+
+def get_graded_courses_with_prefetched_data(semester):
```
karyon: i'd move this directly above the view that uses it

evap/grades/views.py
```diff
@@ -0,0 +1,174 @@
...
+    )
+    return render(request, "grades_semester_view.html", template_data)
+
+
+@grade_publisher_required
+def course_view(request, semester_id, course_id):
+    semester = get_object_or_404(Semester, id=semester_id)
+    course = get_object_or_404(Course, id=course_id)
+
+    grade_documents = GradeDocument.objects.filter(course=course)
```
karyon: you can use course.grade_documents here and put this directly into the template data

evap/grades/views.py
```diff
@@ -0,0 +1,174 @@
...
+        if instance:
+            created = False
+        form = GradeDocumentForm(request.POST, request.FILES, instance=instance)
+        if form.is_valid():
+            instance = form.save(commit=False)
+            instance.last_modified_user = request.user
+            instance.course = course
+            if final_grades:
+                instance.type = GradeDocument.FINAL_GRADES
+            instance.save()
```
karyon: why does this not happen in form.save()?

evap/grades/views.py
```diff
@@ -0,0 +1,174 @@
...
+        return redirect('grades:course_view', semester.id, course.id)
+    else:
+        template_data = dict(
+            semester=semester,
+            course=course,
+            form=form,
+            final_grades=final_grades,
+        )
+        return render(request, "grades_upload_form.html", template_data)
+
```
karyon: the logic of this view is quite convoluted. the helper method that returns three values (on two of which the code branches) contributes to that. i suggest you try inlining the helper which should eliminate half of the branches, and see whether some code should be extracted again afterwards. also, please see whether you can create the form in one line like the other views do.

evap/grades/views.py
```diff
@@ -0,0 +1,174 @@
...
+@grade_downloader_required
+def download_grades(request, grade_document_id):
+    if not request.method == "GET":
+        return HttpResponseBadRequest()
+
+    grade_document = get_object_or_404(GradeDocument, id=grade_document_id)
+
+    # final grades can only be downloaded when the evaluation is finished
+    if grade_document.type == GradeDocument.FINAL_GRADES and grade_document.course.state not in ['evaluated', 'reviewed', 'published']:
```
karyon: shouldn't courses that have final grades attached always be published? if yes, maybe do an assertion on this here. is there any code that explicitly enforces this?

evap/grades/views.py
```diff
@@ -0,0 +1,174 @@
...
+        return HttpResponseForbidden()
+
+    filename = os.path.join(settings.MEDIA_ROOT, grade_document.file.name)
```
karyon: please see whether grade_document.file.path works instead of the manual joining.

karyon:
please also have a look at the landscape issues

evap/grades/views.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+@grade_publisher_required
+def edit_grades(request, semester_id, course_id, grade_document_id):
+    semester = get_object_or_404(Semester, id=semester_id)
+    course = get_object_or_404(Course, id=course_id)
+    grade_document = get_object_or_404(GradeDocument, id=grade_document_id)
+    final_grades = grade_document.type == GradeDocument.FINAL_GRADES
+
+    success, created, form = helper_grade_upload(request, course, final_grades=final_grades, instance=grade_document)
```
karyon: i see now why you have the helper, but still i would try inlining it and then extracting again.

evap/grades/templates/grades_upload_form.html
```diff
@@ -0,0 +1,25 @@
...
+
+{% load i18n %}
+
+{% block content %}
+    {{ block.super }}
+    
+    {% if final_grades %}
+        <div class="bs-callout bs-callout-info">
+            <b>{% trans "Automated publishing" %}</b><br />
+            {% blocktrans %}After this file with total grades was uploaded, the course will be published automatically once the evaluation ends. If the evaluation for this course has already ended, it will be published immediately. All contributors and participants will receive an email informing them about the published evaluation results and the available grade file document.{% endblocktrans %}
```
karyon: is there a difference between total and final grades? any reason to not use the same word in code and UI?

evap/grades/views.py
```diff
@@ -0,0 +1,174 @@
...
+
+    if success:
+        messages.success(request, _("Successfully updated grades."))
+        return redirect('grades:course_view', semester.id, course.id)
+    else:
+        template_data = dict(
+            semester=semester,
+            course=course,
+            form=form,
+            final_grades=False,  # don't show publishing information
```
karyon: does this parameter do anything besides controlling whether that information should be displayed? if yes, please give it a different name, e.g. `show_automated_publishing_warning/info`

evap/settings.py
```diff
@@ -242,6 +277,14 @@
...
 MEDIA_URL = '/media/'
 
+# the Backend used for downloading attachments may be one of the following:
+# `sendfile.backends.development` - for use with django development server only. DO NOT USE IN PRODUCTION
+# `sendfile.backends.simple` - "simple" backend that uses Django file objects to attempt to stream files from disk (note middleware may cause files to be loaded fully into memory)
+# `sendfile.backends.xsendfile` - sets X-Sendfile header (as used by mod_xsendfile/apache and lighthttpd)
+# `sendfile.backends.mod_wsgi` - sets Location with 200 code to trigger internal redirect (daemon mode mod_wsgi only - see below)
+# `sendfile.backends.nginx` - sets X-Accel-Redirect header to trigger internal redirect to file
+# see https://github.com/johnsensible/django-sendfile for further information
+SENDFILE_BACKEND = 'sendfile.backends.simple'
```
karyon: maybe the last line from the comment lines is enough?

evap/staff/management/commands/run_tasks.py
```diff
@@ -26,12 +29,23 @@ def update_courses(self):
...
+                            grade_document_courses.append(course)
+                            evaluation_results_courses.append(course)
+                    elif course.final_grade_documents.exists():
+                        grade_document_courses.append(course)
                     course.save()
             except Exception:
                 pass
 
         if courses_new_in_evaluation:
             EmailTemplate.get_evaluation_started_template().send_to_users_in_courses(courses_new_in_evaluation, ['all_participants'])
```
karyon: please bring this in line with the style one line below (i.e. one method that gets a list of courses)

evap/staff/management/commands/run_tasks.py
```diff
@@ -26,12 +29,23 @@ def update_courses(self):
                     course.evaluation_end()
                     if course.is_fully_reviewed():
                         course.review_finished()
+                        if not course.is_graded:
+                            course.publish()
```
karyon: this is new, right? that courses without grades are published automatically

evap/staff/management/commands/run_tasks.py
```diff
@@ -26,12 +29,23 @@ def update_courses(self):
...
+                            evaluation_results_courses.append(course)
+                        elif course.final_grade_documents.exists():
+                            course.publish()
+                            grade_document_courses.append(course)
+                            evaluation_results_courses.append(course)
+                    elif course.final_grade_documents.exists():
+                        grade_document_courses.append(course)
```
karyon: i'm not quite sure whether it's cool to have this much code in a management command. i mean, it's an integral part of a course's life cycle that is in a completely different place. if you don't disagree, please move as much code as possible into the Course class.

evap/staff/tests.py
```diff
@@ -241,19 +241,19 @@ def test_remove_responsibility(self):
...
+#         """
+#             ensures that the number of queries in the user list is constant
+#             and not linear to the number of users
+#         """
+#         num_users = 50
+#         mommy.make(UserProfile, username="staff.user", groups=[Group.objects.get(name="Staff")])
+#         mommy.make(UserProfile, _quantity=num_users)
+
+#         with self.assertNumQueries(FuzzyInt(0, num_users-1)):
+#             self.app.get("/staff/user/", user="staff.user")
```
karyon: i just found out about https://docs.python.org/dev/library/unittest.html#skipping-tests-and-expected-failures, please use that.

evap/student/templates/student_index.html
```diff
@@ -81,18 +81,44 @@
...
                                     {% endif %}
                                 </td>
-                                <td>
-                                    {% if course.state == 'published' %}
-                                        {% if course.can_publish_grades %}
-                                            <a href="{% url "results:course_detail" semester.id course.id %}" class="btn btn-sm btn-default">{% trans "Results" %}</a>
-                                        {% else %}
-                                            <div data-toggle="tooltip" data-placement="left" class="disabled-tooltip" title="{% trans "We did not receive enough votes to publish the results of this course" %}"><a class="btn btn-sm btn-default" disabled="disabled">{% trans "Results" %}</a></div>
+                                <td class="text-right">
+                                    {% if course.state == 'evaluated' or course.state == 'reviewed' or course.state == 'published' %}
```
karyon: again, shouldn't courses with grades always be published?

evap/student/templates/student_index.html
```diff
@@ -81,18 +81,44 @@
...
+                                            {% else %}
+                                                <div data-toggle="tooltip" data-placement="left" class="disabled-tooltip" title="{% trans "We did not receive enough votes to publish the results of this course" %}"><a class="btn btn-sm btn-default" disabled="disabled">{% trans "Results" %}</a></div>
+                                            {% endif %}
+                                        {% endif %}
+                                    {% else %}
+                                        {% if course.preliminary_grade_documents.count > 0 and can_download_grades %}
+                                            <div class="btn-group">
+                                                <button type="button" class="btn btn-sm btn-default dropdown-toggle" data-toggle="dropdown" aria-expanded="false" {{ disable_if_archived }}>{% trans "Download grades" %} <span class="caret"></span></button>
+                                                <ul class="dropdown-menu" role="menu">
+                                                    {% for grade_document in course.preliminary_grade_documents %}
```
karyon: why not iterate over all grade documents here as well?

evap/student/templates/student_index.html
```diff
@@ -81,18 +81,44 @@
...
+                                                        <li><a href="{% url "grades:download_grades" grade_document.id %}">{{ grade_document.description }}</a></li>
+                                                    {% endfor %}
+                                                </ul>
+                                            </div>
```
karyon: these ~10 lines are exactly the same as the ones starting at line 86, except for preliminary vs final. please check the comment above and merge these parts if possible.

evap/student/templates/student_index.html
```diff
@@ -81,18 +81,44 @@
...
-                                        {% else %}
-                                            <a href="{% url "student:vote" course.id %}" class="btn btn-sm btn-primary">{% trans "Evaluate" %}</a>
+                                        {% if course.state == 'inEvaluation' %}
+                                            {% if course in voted_courses %}
+                                                <div data-toggle="tooltip" data-placement="left" class="disabled-tooltip" title="{% trans "You already evaluated this course" %}"><a class="btn btn-sm btn-default" disabled="disabled">{% trans "Evaluate" %}</a></div>
+                                            {% else %}
+                                                <a href="{% url "student:vote" course.id %}" class="btn btn-sm btn-primary">{% trans "Evaluate" %}</a>
+                                            {% endif %}
                                         {% endif %}
```
karyon: here the logic for the different buttons is quite mixed together although they could be separated.

how about 
```
{% if course.state == 'inEvaluation' %}
    vote button stuff

{% if course.state == 'evaluated' or course.state == 'reviewed' or course.state == 'published' %}
    some grade button stuff
{% else %}
    some other grade button stuff

{% if course.state == 'published' %}
    results button stuff
```

evap/grades/templates/grades_base.html
```diff
@@ -0,0 +1,15 @@
+{% extends "base.html" %}
+
+{% load i18n %}
+{% load static %}
+
+{% block title %}{% trans "Grade publishing" %} - {{ block.super }}{% endblock %}
+
+{% block subtitle %}
+    {{ block.super }}
+    {% if disable_breadcrumb_staff %}
```
karyon: that doesn't help here does it?

evap/grades/views.py
```diff
@@ -0,0 +1,174 @@
...
+    course = get_object_or_404(Course, id=course_id)
+
+    grade_documents = GradeDocument.objects.filter(course=course)
+
+    template_data = dict(
+        semester=semester,
+        course=course,
+        grade_documents=grade_documents,
+        disable_if_archived="disabled=disabled" if semester.is_archived else "",
+        disable_breadcrumb_course=True,
```
karyon: why disable the breadcrumb here?

evap/evaluation/models.py
```diff
@@ -916,9 +937,22 @@ def send_to_user(self, user, courses=None, cc=True):
...
+                'grade_documents': grade_documents,
+                'evaluation_results': evaluation_results
+            }),
+            body = self.render_string(self.body, {
+                'user': user,
+                'courses': courses,
+                'grade_documents': grade_documents,
+                'evaluation_results': evaluation_results,
+                'grade_document_courses': grade_document_courses,
+                'evaluation_results_courses': evaluation_results_courses
```
karyon: maybe we should provide example email templates in the initial data. what do you think? right now it is completely unclear what they should look like

evap/grades/templates/grades_semester_base.html
janno42: the title is used in the title bar of the bowser. here we add the semester name (or the string `Semesters` if we don't have a semester at hand, e.g. when we are in a semester overview).

evap/grades/views.py
janno42: the instance type must be set in advance, otherwise the form wouldn't be valid at all. the user can't select the type by hand so it is set in this method.

evap/grades/views.py
janno42: no, the grades will not be published as long as the course is in evaluation.

evap/staff/management/commands/run_tasks.py
janno42: yes

evap/staff/management/commands/run_tasks.py
janno42: would you prefer an `update` method within the course that does stuff?

evap/student/templates/student_index.html
janno42: see above

evap/evaluation/tools.py
```diff
@@ -271,30 +276,52 @@ def is_external_email(email):
...
+
+    for user, course_lists in publish_notifications.items():
+        try:
+            EmailTemplate.get_publish_template().send_to_user(
+                user, 
+                grade_document_courses=list(course_lists.grade_document_courses),
+                evaluation_results_courses=list(course_lists.evaluation_results_courses)
+            )
+        except Exception:
+            logging.getLogger(__name__).exception(("An error occured when sending the notification email to {}.").format(user.username))
```
karyon: i suggest putting this around the actual send_mail, which is the place where we kinda "expect" exceptions.

evap/student/templates/student_index.html
janno42: because the course is not yet evaluated

evap/student/templates/student_index.html
janno42: see above

evap/grades/templates/grades_base.html
janno42: nope :)

evap/staff/management/commands/run_tasks.py
karyon: as discussed, please make evap send mails to the admins if something goes wrong here.

evap/grades/views.py
janno42: because we are viewing the course itself. following e.g. the course edit pages for staff users we don't put a link for the course in this situation. we need the course only be linked when viewing an upload form for the course.

evap/evaluation/models.py
janno42: -> #587

evap/evaluation/tools.py
karyon: and then maybe merge this with user_publish_notifications again.

evap/grades/views.py
karyon: erm, here you're setting the type AFTER is_valid is called...

janno42: oh, right. because there is a default value. well i guess then it can be moved.

evap/staff/management/commands/run_tasks.py
karyon: i initially thought this should go into the transition itself, but django-fsm looks like you can't do more transitions inside a transition.

...maybe the post_transition signal could be used? could you check how that would look like?

karyon: as discussed, please look how moving all this into a class method of Course.

evap/grades/forms.py
janno42: https://docs.djangoproject.com/en/dev/ref/models/fields/#datefield says that `As currently implemented, setting auto_now to True will cause the field to have editable=False`. and the description for `editable` then explains `If False, the field will not be displayed in the admin or any other ModelForm`.
that means we cannot show the `last_modified_time` field in the form, thus we have to create a new one. also, i would like to keep the current approach for `last_modified_user`, because otherwise we would have a disabled select field there which doesn't make sense.

karyon: ok. a comment in one place and references to that comment in the other places would be cool.

evap/grades/views.py
janno42: i don't really know what you want here. we should probably talk about that in person.

janno42:
i addressed most of the comments, but some things are still missing. first of all refactoring the course updating in run_tasks.

evap/evaluation/models.py
```diff
@@ -447,6 +448,51 @@ def is_archiveable(self):
...
+                        course.review_finished()
+                        if not course.is_graded:
+                            course.publish()
+                            evaluation_results_courses.append(course)
+                        elif course.final_grade_documents.exists():
+                            course.publish()
+                            grade_document_courses.append(course)
+                            evaluation_results_courses.append(course)
+                    elif course.final_grade_documents.exists():
+                        grade_document_courses.append(course)
```
karyon:
```
if course.is_fully_reviewed():
    course.review_finished()
    if not course.is_graded or course.final_grade_documents.exists():
        course.publish()
        evaluation_results_courses.append(course)
if course.final_grade_documents.exists():
    grade_document_courses.append(course)
```
could you check whether that's equivalent and then use that?

evap/evaluation/models.py
```diff
@@ -916,14 +971,31 @@ def send_to_user(self, user, courses=None, cc=True):
...
             to = [user.email],
             cc = cc_addresses,
             bcc = [a[1] for a in settings.MANAGERS],
             headers = {'Reply-To': settings.REPLY_TO_EMAIL})
-        mail.send(False)
+        
+        try:
+            mail.send(False)
+        except Exception:
+            logging.getLogger(__name__).exception(('An error occured when sending email "{}" to {}.').format(subject, user.username))
```
karyon: what about the body in case the admins want to re-send this?

evap/evaluation/models.py
```diff
@@ -916,14 +971,31 @@ def send_to_user(self, user, courses=None, cc=True):
         else:
             cc_addresses = []
 
+        grade_documents = len(grade_document_courses) > 0
+        evaluation_results = len(evaluation_results_courses) > 0
```
karyon: these names sound like lists but they aren't, please think of something else. maybe prepend with "has_"

evap/grades/templates/grades_semester_menu.html
```diff
@@ -0,0 +1,5 @@
+{% load i18n %}
+
+{% for semester in semesters %}
+    <li><a style="padding-left: 30px;" href="{% url "grades:semester_view" semester.id %}">{{ semester.name }}</a></li>
```
karyon: just a reminder that you wanted such style stuff to be somewhere else... :)

evap/grades/templates/grades_semester_view.html
```diff
@@ -0,0 +1,85 @@
...
+                                </td>
+                                <td>{{ course.type }}</td>
+                                <td class="text-center">{% if has_final_grades %}<span class="glyphicon glyphicon-ok"></span>{% endif %}</td>
+                                <td>
+                                    <div class="btn-group">
+                                        <button type="button" class="btn btn-sm btn-default dropdown-toggle" data-toggle="dropdown" aria-expanded="false" {{ disable_if_archived }}>{% trans "Upload grades" %} <span class="caret"></span></button>
+                                        <ul class="dropdown-menu" role="menu">
+                                            <li><a href="{% url "grades:upload_grades" semester.id course.id %}">Midterm grades</a></li>
+                                            <li><a href="{% url "grades:upload_grades" semester.id course.id %}?final=true">Final grades</a></li>
+                                        </ul>
```
karyon: i don't know how that button works, but shouldn't the links be disabled as well?

evap/grades/tests.py
```diff
@@ -0,0 +1,140 @@
...
+            for grade_document in course.grade_documents.all():
+                grade_document.file.delete()
+
+    def get_assert_403(self, url, user):
+        try:
+            self.app.get(url, user=user, status=403)
+        except AppError as e:
+            self.fail('url "{}" failed with user "{}"'.format(url, user))
+
+    def helper_upload_grades(self, course, final_grades=False):
```
karyon: please remove the default parameter

evap/grades/views.py
karyon: despite that, why would we hold back publishing the grades when evaluation is still running? surely that might influence the evaluation but whether grades are published or not shouldn't be our decision.

evap/grades/forms.py
```diff
@@ -0,0 +1,53 @@
...
+        self.fields['last_modified_time_2'].initial = self.instance.last_modified_time
+        self.fields['last_modified_time_2'].widget.attrs['readonly'] = "True"
+        if self.instance.last_modified_user:
+            self.fields['last_modified_user_2'].initial = self.instance.last_modified_user.full_name
+        self.fields['last_modified_user_2'].widget.attrs['readonly'] = "True"
+
+        self.instance.course = course
+        if final_grades:
+            self.instance.type = GradeDocument.FINAL_GRADES
+        self.instance.last_modified_user = user
```
karyon: how does this work when using this form for creating new documents? i thought `instance` is none then.

karyon:
done reviewing.

evap/grades/forms.py
janno42: the instance given to the form during initialization is none. but once you have the form its data is stored in the instance. so setting a value for the instance allows saving it afterwards.

evap/evaluation/models.py
janno42: the body might be quite long and there might be quite a lot of failing messages if something goes wrong. should we put that much text into the log file?

evap/grades/templates/grades_semester_view.html
janno42: the button is a dropdown menu. if it's disabled it doesn't do anything when clicking on it, so the contained links won't be displayed.

evap/grades/views.py
janno42: yeah... right. all grade documents will now be always published immediately.

evap/evaluation/models.py
```diff
@@ -447,6 +448,47 @@ def is_archiveable(self):
...
+                    courses_new_in_evaluation.append(course)
+                elif course.state == "inEvaluation" and course.vote_end_date < today:
+                    course.evaluation_end()
+                    if course.is_fully_reviewed():
+                        course.review_finished()
+                        if not course.is_graded or course.final_grade_documents.exists():
+                            course.publish()
+                            evaluation_results_courses.append(course)
+                    if course.final_grade_documents.exists():
+                        grade_document_courses.append(course)
```
karyon: this might be wrong now. if it has grades, an email for that should've been sent already.

karyon:
:+1:

janno42:
:+1:

PR #610, opened 2015-07-16
Grant reward points only if required courses exist

janno42:
fix 607; add additional check before granting reward points


karyon:
:+1

PR #493, opened 2015-01-16
[WIP] Some cleanup of the html code

lmm-git:
Some corrections to produce valid HTML5 code and some CSS corrections (#489).

Tasks to do:
- [ ] Check student form for valid functionality due to some field-id changes - for this I need the fix of the JavaScript functions on this page; is there already an issue open?
- [x] the function ```{{ form.as_div }}``` in evap/contributor/templates/contributor_course_form.html produces invalid ids (same at some other forms like /staff/semester/XX/edit...)
- [x] the function ```{{ form.as_div }}``` in evap/contributor/templates/contributor_course_form.html produces invalid table-form mixes
- [x] Check all student html files (or frontend functions)
- [x] Check all contributor html files (or frontend functions)
- [x] Check all evaluation html files (or frontend functions)
- [x] Check all results html files (or frontend functions)
- [ ] Check all rewards html files (or frontend functions)
- [ ] Check html files located in templates

PR #611, opened 2015-07-18
Html cleanup

karyon:
rebased and squashed version of #493. the last commit is new.

i left out three parts of the original PR: https://github.com/lmm-git/EvaP/commit/0e19dd80d83f17a4cfd1448a5b0d2d984b983055 because it didn't seem to be worth it, https://github.com/lmm-git/EvaP/commit/3b2444a985e0b510a158b1ac2d1c01cd8c9a94c9 because it introduced an !important, and the changes in this file here https://github.com/lmm-git/EvaP/commit/cfa3efd927aa2aa142667cb36d0091b0ded66037#diff-22f5bba5e5ef3223a601be53cdb7d28fL189 because i had no idea what they were doing.


janno42:
looks good then :+1:

PR #613, opened 2015-07-19
Grades for staff

janno42:
this allows staff users to access the list of grade documents per course (linked from the semester view pages)

### DIFF
--- a/evap/evaluation/auth.py
+++ b/evap/evaluation/auth.py
@@ -109,6 +109,15 @@ def grade_publisher_required(view_func):
     return user_passes_test(check_user)(view_func)
 
 
+def grade_publisher_or_staff_required(view_func):
+    """
+    Decorator for views that checks that the user is logged in and a grade publisher or a staff member
+    """
+    def check_user(user):
+        return user.is_grade_publisher or user.is_staff
+    return user_passes_test(check_user)(view_func)
+
+
 def grade_downloader_required(view_func):
     """
     Decorator for views that checks that the user is logged in and can download grades
--- a/evap/grades/templates/grades_base.html
+++ b/evap/grades/templates/grades_base.html
@@ -7,5 +7,9 @@
 
 {% block subtitle %}
     {{ block.super }}
-    <li><a href="{% url "grades:index" %}">{% trans "Grade publishing" %}</a></li>
+    {% if is_grade_publisher %}
+        <li><a href="{% url "grades:index" %}">{% trans "Grade publishing" %}</a></li>
+    {% else %}
+        <li>{% trans "Grade publishing" %}</li>
+    {% endif %}
 {% endblock %}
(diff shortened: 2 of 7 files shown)


evap/evaluation/auth.py
```diff
@@ -109,6 +109,15 @@ def check_user(user):
     return user_passes_test(check_user)(view_func)
 
 
+def grade_publisher_or_staff_required(view_func):
+    """
+    Decorator for views that checks that the user is logged in and a grade publisher
```
karyon: comment is wrong

evap/grades/templates/grades_base.html
```diff
@@ -7,5 +7,9 @@
 
 {% block subtitle %}
     {{ block.super }}
-    <li><a href="{% url "grades:index" %}">{% trans "Grade publishing" %}</a></li>
+    {% if is_grade_publisher %}
+    	<li><a href="{% url "grades:index" %}">{% trans "Grade publishing" %}</a></li>
+    {% else %}
+		<li>{% trans "Grade publishing" %}</li>
```
karyon: tab spotted here. also, why shouldn't staff see the index page? seems harmless to allow that...

evap/grades/templates/grades_course_base.html
```diff
@@ -1,11 +1,12 @@
 {% extends "grades_semester_base.html" %}
 
 {% load i18n %}
+{% load morefilters %}
 
 {% block subtitle %}
     {{ block.super }}
     {% if course %}
-    	{% if disable_breadcrumb_course %}
+    	{% if disable_breadcrumb_course or is_grade_publisher|is_false %}
```
karyon: i noticed that disable_breadcrumb_course is set in exactly one place, and there it is always set to true. please check that and check whether it's necessary to do this check here.

evap/grades/templates/grades_base.html
janno42: we could also allow staff users to see the grades index and semester_view pages. they don't really need them and if they can access them they should also get a menu entry for that. that would mean even more menu items...
what do you think?

evap/grades/templates/grades_course_base.html
karyon: oh no. is that again the "if it's not there we'll just assume it's false"-story?

janno42: no that is checking whether it's false. if it's not existing this will not return true:
```
@register.filter
def is_false(arg):
    return arg is False
```

janno42: oh, but if `disable_breadcrumb_course` doesn't exist, its false. yes.

evap/grades/templates/grades_course_base.html
```diff
@@ -1,6 +1,7 @@
 {% extends "grades_semester_base.html" %}
 
 {% load i18n %}
+{% load morefilters %}
```
karyon: this can be removed now

karyon:
uh, what should i be reviewing again?

janno42:
i don't know either. just tell me everything is okay ;)

karyon:
yeah i think everything is okay.

PR #617, opened 2015-07-23
Remove standard deviation for single results and improve result bar

janno42:
This removes the standard deviation for single results from the results page and includes the number of voters into the result bar (also used for each rating question on the course result detail pages).

The new result bar:
result bar


janno42:
updated

karyon:
:light_rail:

PR #618, opened 2015-07-23
Allow selection question type in questionnaire form

### LINKED ISSUE #604: Fix question type selection in questionnaire creation
When creating a new questionnaire and adding additional questions the select box for the question type does not work.

janno42:
This fixes #604.


karyon:
:facepunch:

PR #621, opened 2015-08-05
Update puppet

karyon:
so today i was doing a fresh vagrant up and it didn't work. it looks like they decided to include puppet 4.0 in a new box version (1.0.2), which has breaking changes. here are changes that make our vagrant setup work again. additionally i added the createcachetable command to init.pp which i forgot back then.


janno42:
:+1:

janno42:
:shipit:

Issue #512, opened 2015-01-24
Menu bar hides content


janno42:
The navigation on top of the page hides parts of the content if its size increases. This happens in smaller browser windows, where the navigation bar has more than one line but is not yet compressed into a single icon.
The content should always be visible.

![capture](https://cloud.githubusercontent.com/assets/1781719/5887235/763386a4-a3c4-11e4-8b93-c2999d8ca74a.PNG)

karyon:
reopening. we opted to fix #641 instead of this one here. right now i don't know of a good solution that fixes both issues, apart from making the navbar static instead of fixed to top, which we didn't want.

maybe bootstrap 4 fixes it :)

PR #635, opened 2015-09-17
Add django-compressor, compile bootstrap from source

karyon:
in this PR i add django-compressor to the mix to concatenate all javascript and css files for performance, as well as to compile bootstrap from source. the latter helped in fixing #512.

in debug mode, nothing should change except for bootstrap being recompiled whenever css/bootstrap.less is modified. with `DEBUG=False`, css and js files get concatenated. There is also a COMPRESS_OFFLINE setting that, when enabled, further speeds up page load times but requires you to run `manage.py compress` after `collectstatic`.

i recommend you just do vagrant destroy && vagrant up, as there is a new submodule, some packages to install and a new dependency.


evap/templates/base.html
```diff
@@ -5,20 +5,23 @@
...
-    <link rel="stylesheet" href="{% get_static_prefix %}css/select2.min.css" />
-    <link rel="stylesheet" href="{% get_static_prefix %}css/jquery.dataTables.min.css" />
-    <link rel="stylesheet" href="{% get_static_prefix %}css/dataTables.bootstrap.css" />
-    <link rel="stylesheet" href="{% get_static_prefix %}css/jquery.reject.css" />
-    <link rel="shortcut icon" href="{% get_static_prefix %}images/favicon.ico" />
+        <title>{% block title %}EvaP{% endblock %}</title>
+        {% compress css %}
+            <link rel="stylesheet" href="{% get_static_prefix %}css/jquery.ui.css">
+            <link type="text/less" rel="stylesheet" href="{% get_static_prefix %}css/bootstrap.less">
+            <link rel="stylesheet" href="{% get_static_prefix %}css/evap.css">
```
janno42: any reason why evap.css should not be at the end? it might prevent future hours of looking for the source of styling problems.

janno42:
:+1:

evap/templates/base.html
karyon: i can move it to the bottom, if you test it and tell me everything still looks ok :)

janno42: looks good. and while you're at it: could you please change the line endings to consistent use of ` />`?

karyon: done, could you review again?

janno42:
perfect! :rabbit:

PR #636, opened 2015-09-20
fix apache locale, fixes #626 (at least in vagrant)

### LINKED ISSUE #626: Special characters in grade file names
When a grade file name contains special characters like umlauts, an error occurs during the upload and the uploader gets a message that the admins have been informed. EvaP should either be able to handle these file names (preferred solution) or at least show a message that explains that the file name must not contain special characters.

karyon:
fixes #626.

the problem is actually in some apache config, this PR fixes it in the vagrant setup. see diff for details. 

to fix this on production, run this:
`sudo sed -i s,#.\ /etc/default/locale,.\ /etc/default/locale,g /etc/apache2/envvars` and then restart apache.

### DIFF
--- a/deployment/testing_environment/manifests/site.pp
+++ b/deployment/testing_environment/manifests/site.pp
@@ -58,6 +58,13 @@ node default {
     # apache environment
     class { 'apache':
         default_vhost => false
+    } ->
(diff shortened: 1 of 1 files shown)


janno42:
:+1:

Issue #392, opened 2014-09-10
PDF export for course results


janno42:
We need an option to export the results of a course as a PDF document.
This should be available for staff users and responsible contributors.

karyon:
server-side pdf rendering would probably either use headless chrome or phantomjs, which is webkit-based. the resulting pdf will probably not differ from chrome's save-to-pdf functionality we use today.

in firefox, the output isn't as pretty yet (see #1013). additionally, firefox doesn't support pdf rendering, one would need to use a pdf printer, resulting in larger filesizes.

so, advantages of server-side pdf rendering would be:
* users save some clicks
* users don't need to use chrome (partly addressed in #1013)

disadvantages:
* implementation effort
* additional infrastructure complexity that adds maintenance burden
* less flexibility (we loose the options in the printing dialog)

we decided it's not worth it.

Issue #495, opened 2015-01-16
Make usage of Font consistent across platforms


Bartzi:
Different operating systems use different fonts and therefore certain layout errors as seen below may occur. The task is to investigate whether this is actually the case.
If this is the problem we need to make sure that every client uses the same font!

![image](https://cloud.githubusercontent.com/assets/2399349/5784538/27fe4c1e-9dcd-11e4-945b-412e87e46e5f.png)

karyon:
using the same font might be sufficient, but in theory this could still occur since this is also dependent on font rendering. we could also hack this by adding a few px to the page/button area width or making the font or margins inside buttons a bit smaller.

karyon:
@janno42 says it should be possible to first render the buttons and then let the question column take up the remaining space.

janno42:
issue shown above is fixed. we decided against sending a font. other occurrences will be solved case by case. a general solution will be flexboxes in Bootstrap 4.

PR #638, opened 2015-09-21
Print optimization

janno42:
This allows printing results pages with colored result bars when activating "Print background graphics" in the browser.

It's a step to #392.

### DIFF
--- a/evap/results/templates/results_course_detail.html
+++ b/evap/results/templates/results_course_detail.html
@@ -22,10 +22,10 @@
 
     {% if staff or contributor %}
         {% if public_view %}
-            <div class="alert alert-info">{% trans "This is the public view. Users who are neither contributors of this course nor their delegates can only see the results below." %}</div>
-            <a href="{% url "results:course_detail" course.semester.id course.id %}?public_view=false" class="pull-right btn btn-sm btn-primary">{% trans "Show my view" %}</a>
+            <div class="alert alert-info hidden-print">{% trans "This is the public view. Users who are neither contributors of this course nor their delegates can only see the results below." %}</div>
+            <a href="{% url "results:course_detail" course.semester.id course.id %}?public_view=false" class="pull-right btn btn-sm btn-primary hidden-print">{% trans "Show my view" %}</a>
         {% else %}
-            <a href="{% url "results:course_detail" course.semester.id course.id %}?public_view=true" class="pull-right btn btn-sm btn-default">{% trans "Show public view" %}</a>
+            <a href="{% url "results:course_detail" course.semester.id course.id %}?public_view=true" class="pull-right btn btn-sm btn-default hidden-print">{% trans "Show public view" %}</a>
         {% endif %}
     {% endif %}
     <h2>{{ course.name }} ({{ course.semester.name }})</h2>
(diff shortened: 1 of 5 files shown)


evap/results/templates/results_course_detail_questionnaires.html
```diff
@@ -16,31 +16,33 @@
     <table class="table table-striped">
         <tbody>
             {% for result in results %}
-                <tr>
                 {% if result.question.is_rating_question %}
-                    <td colspan="2">{{ result.question.text }}</td>
-                    <td>
-                        {% include_result_bar result show_grades questionnaire_warning %}
-                    </td>
+                    <tr class="no-break">
```
karyon: this no-break doesn't help with #495 by any chance?

evap/results/templates/results_course_detail.html
```diff
@@ -48,12 +48,12 @@
...
-                        <th class="col-sm-3">{% trans "Responsible" %}</th>
-                        <th class="col-sm-2">{% trans "Average Grade" %}</th>
-                        <th class="col-sm-2">{% trans "Standard Deviation" %}</th>
-                        <th class="col-sm-2">{% if course.is_single_result|is_false %}{% trans "Voters" %}{% endif %}</th>
+                        <th class="col-xs-1">{% trans "Degree" %}</th>
+                        <th class="col-xs-2">{% trans "Type" %}</th>
+                        <th class="col-xs-3">{% trans "Responsible" %}</th>
+                        <th class="col-xs-2">{% trans "Average Grade" %}</th>
+                        <th class="col-xs-2">{% trans "Standard Deviation" %}</th>
+                        <th class="col-xs-2">{% if course.is_single_result|is_false %}{% trans "Voters" %}{% endif %}</th>
```
karyon: why?

evap/static/css/bootstrap.less
```diff
@@ -10,7 +10,7 @@
 
 // Reset and dependencies
 @import "../bootstrap/less/normalize.less";
-@import "../bootstrap/less/print.less";
+//@import "../bootstrap/less/print.less"; // don't include bootstrap print css to enable printing colored results pages
```
karyon: could you change this to `don't include this to enable ...`, i had difficulties parsing it :)

evap/static/css/evap.css
```diff
@@ -146,15 +150,56 @@ td.vote-question.text-field {
...
+    }
+    .grade-bg {
+        width: 30px;
+    }
+    .deviation-bg {
+        width: 30px;
+    }
+    .alert {
+        padding: 5px;
+    }
```
karyon: that looks like everything from print.less except the background stuff. wouldn't it be cooler to just include what helps us?

evap/results/templates/results_course_detail_questionnaires.html
janno42: it might help - but it's a different issue ;)

evap/static/css/evap.css
janno42: i wrote that all by myself. we need it just like that.

evap/results/templates/results_course_detail.html
janno42: using `xs` this also works for printing. it seems like printed pages are dealt with as xs-devices (see bootstrap docu for details, it doesn't work without in our case).
we could think about refactoring these on all pages.

karyon:
:+1:

PR #639, opened 2015-09-21
Fix ordering after error

### LINKED ISSUE #456: Errors during saving formset resets ordering
When saving a formset results in an error the ordering of e.g. the contributors is reset. The form should keep the ordering that was set by the user.

karyon:
fix #456. also fixes a massive test slowdown introduced in #635.


janno42:
this is SO GREAT :trophy:

Issue #627, opened 2015-09-16
Confirm finished grading process without grade upload


janno42:
Grade uploaders need an option to confirm that the grading process for a course is finished, without actually uploading a grade document. This is necessary for e.g. Bachelor projects, where the participants get individual information about their grades, so there exists no overview document for the whole project.

Courses for which this confirmation was given should be handled exactly like courses where final grades have been uploaded when it comes to publishing the evaluation results.

PR #640, opened 2015-09-21
Grade publishing

janno42:
This allows graders
- to see also ungraded courses (there are courses which are ungraded but where documents will be uploaded, stating whether participants passed or not)
- to confirm that the final grades for a course have been submitted but will not be uploaded to the evaluation platform (#627).

Confirming that no grade documents will be uploaded leads to the following:
- If the course is already reviewed: it will be published.
- If the evaluation for this course ends and it is already reviewed: it will be published.

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -177,6 +177,9 @@ class Course(models.Model, metaclass=LocalizeModelBase):
     # default is True as that's the more restrictive option
     is_graded = models.BooleanField(verbose_name=_("is graded"), default=True)
 
+    # graders can set this to True, then the course will be handled as if final grades have already been uploaded
+    gets_no_grade_documents = models.BooleanField(verbose_name=_("gets no grade documents"), default=False)
+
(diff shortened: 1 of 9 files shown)


janno42:
please review again

evap/evaluation/models.py
```diff
@@ -177,6 +177,9 @@ class Course(models.Model, metaclass=LocalizeModelBase):
     # default is True as that's the more restrictive option
     is_graded = models.BooleanField(verbose_name=_("is graded"), default=True)
 
+    # graders can set this to True, then the course will be handled as if final grades have already been uploaded
+    gets_no_grades = models.BooleanField(verbose_name=_("gets no grades"), default=False)
```
karyon: as discussed, please rename to gets_no_grade_documents

janno42: done

karyon:
:package:

karyon:
that is, apart from that one comment i did 10 minutes ago, which i can't find right now...

PR #642, opened 2015-09-26
Use sortable.js for sorting

karyon:
because it feels so much better.


janno42:
please also apply for
- contributor_course_form
- staff_questionnaire_index_list

karyon:
fixed.

janno42:
:+1:

PR #643, opened 2015-09-27
Text results for contributors

### LINKED ISSUE #551: Add option to publish Text Results for additional contributors
The responsible person should be able to allow single contributors of a course to view all Text Results about the course or all Text Results about the course and all contributors once they are published.
This could be done with two extra fields in the form, next to "can edit".

janno42:
fix #551 

This allows staff users and editors to define on the course edit page, which contributors can see which comments after the course was published.

### DIFF
--- a/evap/contributor/templates/contributor_course_form.html
+++ b/evap/contributor/templates/contributor_course_form.html
@@ -3,6 +3,7 @@
 {% load i18n %}
 {% load static %}
 {% load morefilters %}
+{% load evaluation_templatetags %}
 
 {% block subtitle %}
     {{ block.super }}
@@ -54,32 +55,11 @@
                             </td>
                             <td>{{ form.questionnaires.errors }} {{ form.questionnaires }}</td>
                             <td>                        
-                                {% trans "Responsibility" %}:
-                                <div class="btn-group" data-toggle="buttons">
-                                    {% for choice in form.responsibility %}
-                                        <label class="btn btn-sm btn-option{% if form.responsibility.value == "RESPONSIBLE" %} disabled{% endif %}{% if form.responsibility.value == choice.choice_value %} active{% endif %}{% if form.responsibility.errors %} choice-error{% endif %}" name="{{ choice.name }}" data-toggle="tooltip" data-placement="top" title=
-                                        {% if choice.choice_value == "CONTRIBUTOR" %}"{% trans "Default value. No special rights." %}"
-                                        {% elif choice.choice_value == "EDITOR" %}"{% trans "Able to edit the course to prepare the evaluation." %}"
-                                        {% elif choice.choice_value == "RESPONSIBLE" %}"{% trans "Name will be shown next to course. Able to see all comments about all contributors of this course. Every course has exactly one responsible person." %}"
-                                        {% endif %}>
-                                            <input id="{{ choice.id_for_label }}" name="{{ choice.name }}" type="radio" value="{{ choice.choice_value }}" autocomplete="off" {% if form.responsibility.value == choice.choice_value %}checked{% endif %} />
-                                            {{ choice.choice_label }}
-                                        </label>
-                                    {% endfor %}
-                                </div><br /><br />
-                                {% trans "Comment availability" %}:
-                                <div class="btn-group" data-toggle="buttons">
-                                    {% for choice in form.comment_visibility %}
-                                        <label class="btn btn-sm btn-option{% if form.responsibility.value == "RESPONSIBLE" %} disabled{% endif %}{% if form.comment_visibility.value == choice.choice_value %} active{% endif %}{% if form.comment_visibility.errors %} choice-error{% endif %}" name="{{ choice.name }}" data-toggle="tooltip" data-placement="top" title=
-                                        {% if choice.choice_value == "OWN" %}"{% trans "Default value. Will only see own comments." %}"
-                                        {% elif choice.choice_value == "COURSE" %}"{% trans "Will see own comments and comments about the course." %}"
-                                        {% elif choice.choice_value == "ALL" %}"{% trans "Will see all comments about the course and about all contributors." %}"
-                                        {% endif %}>
-                                            <input id="{{ choice.id_for_label }}" name="{{ choice.name }}" type="radio" value="{{ choice.choice_value }}" autocomplete="off" {% if form.comment_visibility.value == choice.choice_value %}checked{% endif %} />
-                                            {{ choice.choice_label }}
-                                        </label>
-                                    {% endfor %}
-                                </div>
+                                {% trans "Responsibility" %}:<br />
+                                {% include_responsibility_buttons form %}
+                                <br /><br />
+                                {% trans "Comment availability" %}:<br />
+                                {% include_comment_visibility_buttons form %}
                             </td>
                             <td>{% if not form.responsibility.value == "RESPONSIBLE" %}{{ form.DELETE }}{% endif %}</td>
                         </tr>
--- a/evap/contributor/templates/contributor_course_form.html
+++ b/evap/contributor/templates/contributor_course_form.html
@@ -58,7 +58,7 @@
                                 {% trans "Responsibility" %}:<br />
                                 {% include_responsibility_buttons form %}
                                 <br /><br />
-                                {% trans "Comment availability" %}:<br />
+                                {% trans "Comment visibility" %}:<br />
                                 {% include_comment_visibility_buttons form %}
                             </td>
                             <td>{% if not form.responsibility.value == "RESPONSIBLE" %}{{ form.DELETE }}{% endif %}</td>
--- a/evap/contributor/templates/contributor_course_form.html
+++ b/evap/contributor/templates/contributor_course_form.html
@@ -28,43 +28,60 @@
                     <tr>
                         <th></th>
                         <th class="col-sm-4">{% trans "Contributor" %}</th>
-                        <th class="col-sm-5">{% trans "Questionnaires" %}</th>
-                        <th class="col-sm-2">{% trans "Can edit course" %}</th>
+                        <th class="col-sm-4">{% trans "Questionnaires" %}</th>
+                        <th class="col-sm-3">{% trans "Options" %}</th>
                         <th class="col-sm-1"></th>
                     </tr>
                 </thead>
                 <tbody>
-                    {% for form_element in formset %}
-                        {% if form_element.non_field_errors %}
-                            <tr><td colspan="100">{{ form_element.non_field_errors }}</td></tr>
-                        {% endif %}
-
-                        <tr class="contribution select2tr sortable">
-                            <td style="width: 10px;"><span style="font-size: 16px; top: 8px; cursor: move;" class="glyphicon glyphicon-move"></span></td>
+                    {% for form in formset %}
+                    {% if form.non_field_errors %}
+                        <tr><td colspan=100>{{ form.non_field_errors }}</td></tr>
+                    {% endif %}
+                    <tr class="contribution select2tr sortable">
+                        {% for field in form.hidden_fields %}
+                            {{ field }}
+                        {% endfor %}
+                        <td style="width: 10px;"><span style="font-size: 16px; top: 8px; cursor:move;" class="glyphicon glyphicon-move"></span></td>
                             <td>
-                                {% for field in form_element.hidden_fields %}
-                                    {{ field }}
-                                {% endfor %}
-                                {{ form_element.responsible.as_hidden }}
-                                {{ form_element.contributor.errors }}
-                                {% if form_element.responsible.value == True %}
+                                {{ form.contributor.errors }}
+                                {% if form.responsibility.value == "RESPONSIBLE" %}
                                     <select class="form-control" disabled><option selected>{{ responsible }} ({% trans "responsible" %})</option></select>
-                                    {{ form_element.contributor.as_hidden }}
+                                    {{ form.contributor.as_hidden }}
                                 {% else %}
-                                    {{ form_element.contributor }}
+                                    {{ form.contributor }}
                                 {% endif %}
                             </td>
-                            <td>{{ form_element.questionnaires.errors }} {{ form_element.questionnaires }}</td>
-                            <td>
-                                {{ form_element.can_edit.errors }}
-                                {% if form_element.responsible.value == True %}
-                                    <input type="checkbox" checked disabled />
-                                    {{ form_element.can_edit.as_hidden }}
-                                {% else %}
-                                    {{ form_element.can_edit }}
-                                {% endif %}
+                            <td>{{ form.questionnaires.errors }} {{ form.questionnaires }}</td>
+                            <td>                        
+                                {% trans "Responsibility" %}:
+                                <div class="btn-group" data-toggle="buttons">
+                                    {% for choice in form.responsibility %}
+                                        <label class="btn btn-sm btn-option{% if form.responsibility.value == "RESPONSIBLE" %} disabled{% endif %}{% if form.responsibility.value == choice.choice_value %} active{% endif %}{% if form.responsibility.errors %} choice-error{% endif %}" name="{{ choice.name }}" data-toggle="tooltip" data-placement="top" title=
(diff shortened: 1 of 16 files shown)


evap/staff/forms.py
```diff
@@ -204,6 +213,14 @@ def validate_unique(self):
         except forms.ValidationError as e:
             self._update_errors(e)
 
+    def save(self, *args, **kwargs):
+        super().save(*args, **kwargs)
+        responsibility = self.cleaned_data['responsibility']
+        self.instance.responsible = responsibility == Contribution.IS_RESPONSIBLE
+        self.instance.can_edit = (responsibility == Contribution.IS_RESPONSIBLE) or (responsibility == Contribution.IS_EDITOR)
+        self.instance.save()
+        return self.instance
```
karyon: please do the stuff first and then do `return super().save(*args, **kwargs)` without `self.instance.save()`, i have the feeling that right now you would change values in the db even if the formset generates some errors.

karyon:
* why is it no longer possible to make a contributor the responsible by clicking on their "responsible"-button? right now i have to de-select the current responsible and then change the new one.
* when loading the page, the responsible has "own" comments selected. making him editor and then responsible again fixes that. i think responsibility should always imply "all" availability, even in the database. the form's save() method should enforce that, and the migration would need to handle that as well.
* when clicking those buttons, the tooltip stays there. that was ok for the questionnaires, but it looks really bad for the buttons imo, especially because a tooltip on the lower buttons overlaps the upper buttons
* just an idea, i don't know whether it makes sense: maybe editor should imply "course" availability?

janno42:
1. will change
2. i forgot that in the migration, yes. i will also add it in save.
3. no idea what to do
4. nah, i don't think that should have anything to do with that.

janno42:
addressed 2 and 3. 1 and 4 are okay.
there are still webtest form parsing errors left.

evap/contributor/templates/contributor_course_form.html
```diff
@@ -28,43 +28,60 @@
...
+                                        </label>
+                                    {% endfor %}
+                                </div><br /><br />
+                                {% trans "Comment availability" %}:
+                                <div class="btn-group" data-toggle="buttons">
+                                    {% for choice in form.comment_visibility %}
+                                        <label class="btn btn-sm btn-option{% if form.responsibility.value == "RESPONSIBLE" %} disabled{% endif %}{% if form.comment_visibility.value == choice.choice_value %} active{% endif %}{% if form.comment_visibility.errors %} choice-error{% endif %}" name="{{ choice.name }}" data-toggle="tooltip" data-placement="top" title=
+                                        {% if choice.choice_value == "OWN" %}"{% trans "Default value. Will only see own comments." %}"
+                                        {% elif choice.choice_value == "COURSE" %}"{% trans "Will see own comments and comments about the course." %}"
+                                        {% elif choice.choice_value == "ALL" %}"{% trans "Will see all comments about the course and about all contributors." %}"
```
karyon: please put these if elifs into a tag, or put the whole widget into a separate file. i'd prefer the latter to duplicate less code with staff_course_form.html

karyon:
i have a feeling i commented this already, but couldn't find it. do you think it makes sense to show the always disabled responsible-button on the contributor pages? it think you should at least add some explanation why that can't be changed, if not design the ui controls differently for contributors.

evap/contributor/templates/contributor_course_form.html
karyon: also, please use more line breaks in the label tag

evap/results/views.py
```diff
@@ -134,6 +134,11 @@ def user_can_see_text_answer(user, text_answer, public_view=False):
         if contributor == user or contributor in user.represented_users.all():
             return True
         if text_answer.contribution.course.is_user_responsible_or_delegate(user):
+            return True        
+        if text_answer.contribution.course.contributions.filter(contributor=user, comment_visibility=Contribution.ALL_COMMENTS).exists():
+            return True
+        if text_answer.contribution.is_general and \
+            text_answer.contribution.course.contributions.filter(contributor=user, comment_visibility=Contribution.COURSE_COMMENTS).exists():
```
karyon: what about delegates?

good luck :)

evap/staff/forms.py
```diff
@@ -183,16 +183,25 @@ def validate_unique(self):
...
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields['contributor'].widget.attrs['class'] = 'form-control'
 
+        if self.instance.responsible:
+            self.initial['responsibility'] = Contribution.IS_RESPONSIBLE
+        elif self.instance.can_edit:
+            self.initial['responsibility'] = Contribution.IS_EDITOR
+        else:
+            self.initial['responsibility'] = Contribution.IS_CONTRIBUTOR
```
karyon: if this works, could you please check the difference between self.initial[fieldname] and self.fields[fieldname].initial?

evap/evaluation/templates/evap_course_edit_js.html
```diff
@@ -2,48 +2,90 @@
...
-                    $("#id_contributions-" + i + "-can_edit").prop("checked", true);
-                    $("#id_contributions-" + i + "-can_edit").prop("disabled", true);
+                // check "all comments" and disable comment buttons
+                checkButtons("input[name=contributions-" + i + "-comment_visibility][value=ALL]");
+                disableButtons("input[name=contributions-" + i + "-comment_visibility]");
+
+                // disable all other responsible buttons
+                for(var j = 0; j < contributorCount; j++){
+                    if (j != i)
+                        disableButtons("input[name=contributions-" + j + "-responsibility][value=RESPONSIBLE]");
```
karyon: this shouldn't be necessary right?

evap/evaluation/templates/evap_course_edit_js.html
```diff
@@ -2,48 +2,90 @@
...
                 }
-                // if the responsible checkbox is deselected
-                else {
-                    // enable can-edit
-                    $("#id_contributions-" + i + "-can_edit").prop("disabled", false);
+            }
+            // unchecked responsible right now
```
karyon: could you clarify that comment? this is an if-elseif, a "right now" in between doesn't make sense to me

evap/evaluation/templates/evap_course_edit_js.html
```diff
@@ -2,48 +2,90 @@
...
+    }
 
-        function getContributorCount() {
-            return $("#contribution_table select").length;
+    function makeDisabledClickHandler() {
+        return function () {
+            // don't do anything on disabled buttons
+            if ($(this).hasClass("disabled"))
+                return false;
         }
```
karyon: what is this good for?

evap/evaluation/templates/evap_course_edit_js.html
```diff
@@ -2,48 +2,90 @@
...
+    function enableButtons(selector){
+        $(selector).parent().removeClass("disabled");
+    }
+
+    function getContributorCount() {
+        return $("#contribution_table select").length;
+    }
+
+    function assignClickHandlers() {
+        for(var i = 0; i < getContributorCount(); i++)
```
karyon: please put the result of getContributorCount() into a variable

evap/evaluation/templates/evap_course_edit_js.html
```diff
@@ -2,48 +2,90 @@
...
     $(document).ready(function() {
-        assignClickHandlers()
+        assignClickHandlers();
 
         // apply initial settings
         for(var i = 0; i < getContributorCount(); i++){
-            if($("#id_contributions-" + i + "-responsible").is(":checked")) {
-                $("#id_contributions-" + i + "-can_edit").prop("checked", true);
-                $("#id_contributions-" + i + "-can_edit").prop("disabled", true);
+            if($("input[name=contributions-" + i + "-responsibility][value=RESPONSIBLE]").is(":checked")) {
```
karyon: you use `.is(":checked")` here but `.prop("checked")` above, please make it consistent

evap/evaluation/templates/evap_course_edit_js.html
```diff
@@ -2,48 +2,90 @@
...
+                // disable comment buttons for responsible contributor
+                disableButtons("input[name=contributions-" + i + "-comment_visibility]");
+                // disable all other responsible buttons
+                for(var j = 0; j < getContributorCount(); j++){
+                    if (j != i)
+                        disableButtons("input[name=contributions-" + j + "-responsibility][value=RESPONSIBLE]");
+                }
```
karyon: where do you initially check "all" in the comment widget for responsibles? why doesn't that happen here?

evap/evaluation/templates/evap_course_edit_js.html
```diff
@@ -2,48 +2,90 @@
...
+        assignClickHandlers();
 
         // apply initial settings
         for(var i = 0; i < getContributorCount(); i++){
-            if($("#id_contributions-" + i + "-responsible").is(":checked")) {
-                $("#id_contributions-" + i + "-can_edit").prop("checked", true);
-                $("#id_contributions-" + i + "-can_edit").prop("disabled", true);
+            if($("input[name=contributions-" + i + "-responsibility][value=RESPONSIBLE]").is(":checked")) {
+                // disable comment buttons for responsible contributor
+                disableButtons("input[name=contributions-" + i + "-comment_visibility]");
```
karyon: in the UI this is called "comment availability", here its visibility, please make it consistent

evap/staff/forms.py
janno42: From the documentation: `Note that if a Field defines initial and you include initial when instantiating the Form, then the latter initial will have precedence.`
Otherwise they do the same thing.

evap/evaluation/templates/evap_course_edit_js.html
janno42: when selecting a responsible button, all other responsible buttons must be disabled, because it should only be possible to select one contributor as responsible.

evap/staff/forms.py
karyon: alright, if you're not using those in semantically different ways, please be consistent and use only one way.

evap/evaluation/templates/evap_course_edit_js.html
janno42: the part below is done after a responsible button has been unchecked. what should i write instead?

evap/evaluation/templates/evap_course_edit_js.html
karyon: yeah i know... for some reason i thought that shouldn't be necessary here, but i don't remember why...

evap/evaluation/templates/evap_course_edit_js.html
janno42: when not using this, buttons with the class "disabled" can still be toggled.

evap/evaluation/templates/evap_course_edit_js.html
janno42: responsible contributors already have "all" saved in the db. should i make a second override here?

evap/evaluation/templates/evap_course_edit_js.html
karyon: "user unchecked responsible button" or "if the responsible button is deselected" or "user changed selection from responsible to something else". at first i misunderstood this as a description of the state, which was confusing in between the if-elseif, but it should be more a description of the action that was performed.

evap/staff/forms.py
janno42: check

evap/evaluation/templates/evap_course_edit_js.html
karyon: could you state that problem somehow in the comment?

evap/evaluation/templates/evap_course_edit_js.html
janno42: check

evap/evaluation/templates/evap_course_edit_js.html
janno42: check

evap/evaluation/templates/evap_course_edit_js.html
janno42: check

evap/evaluation/templates/evap_course_edit_js.html
karyon: no, probably not.

evap/evaluation/templates/evap_course_edit_js.html
janno42: check

evap/evaluation/templates/evap_course_edit_js.html
janno42: check

janno42:
addressed comments. now also delegates of qualified users can see comments and the responsible button is not shown for contributors (except when active).

evap/evaluation/templates/comment_visibility_buttons.html
```diff
@@ -0,0 +1,19 @@
...
+            {% trans "Will see all comments about the course and about all contributors." as tooltip %}
+        {% elif choice.choice_value == "COURSE" %}
+            {% trans "Will see own comments and comments about the course." as tooltip %}
+        {% elif choice.choice_value == "OWN" %}
+            {% trans "Default value. Will only see own comments." as tooltip %}
+        {% endif %}
+        {% if form.responsibility.value == "RESPONSIBLE" %}
+            {% include_choice_button form.comment_visibility choice False tooltip %}
+        {% else %}
+            {% include_choice_button form.comment_visibility choice True tooltip %}
```
karyon: can you do something like `form.responsibility.value == "RESPONSIBLE" as disabled` and then use `disabled` as parameter? (i mixed up True and False but you get the idea)

evap/evaluation/templates/responsibility_buttons.html
```diff
@@ -0,0 +1,27 @@
...
+        {% else %}
+            {% if choice.choice_value == "EDITOR" %}
+                {% trans "Able to edit the course to prepare the evaluation." as tooltip %}
+            {% elif choice.choice_value == "CONTRIBUTOR" %}
+                {% trans "Default value. No special rights." as tooltip %}
+            {% endif %}
+            {% if form.responsibility.value == "RESPONSIBLE" and can_change_responsible|is_false %}
+                {% include_choice_button form.responsibility choice False tooltip %}
+            {% else %}
+                {% include_choice_button form.responsibility choice True tooltip %}
```
karyon: same here

evap/results/views.py
```diff
@@ -135,5 +135,12 @@ def user_can_see_text_answer(user, text_answer, public_view=False):
             return True
         if text_answer.contribution.course.is_user_responsible_or_delegate(user):
             return True
```
karyon: these two lines should be superfluous now if i got that correctly. also, please put `list(user.represented_users.all()) into a variable to avoid hitting the database twice in the common case (which is (i think) that a comment is not visible)

evap/staff/forms.py
```diff
@@ -204,6 +213,14 @@ def validate_unique(self):
         except forms.ValidationError as e:
             self._update_errors(e)
 
+    def save(self, *args, **kwargs):
+        responsibility = self.cleaned_data['responsibility']
+        self.instance.responsible = responsibility == Contribution.IS_RESPONSIBLE
+        self.instance.can_edit = (responsibility == Contribution.IS_RESPONSIBLE) or (responsibility == Contribution.IS_EDITOR)
+        if responsibility == Contribution.IS_RESPONSIBLE:
+            self.instance.comment_visibility = Contribution.ALL_COMMENTS
```
karyon: please put `responsibility == Contribution.IS_RESPONSIBLE` and `responsibility == Contribution.IS_EDITOR` into a variable

karyon:
done reviewing.

evap/evaluation/templates/comment_visibility_buttons.html
janno42: i think variable assignment is only possible with `with`and something like `{% with disabled = form.responsibility.value == "RESPONSIBLE" %}` does not work. if you have any other idea, please let me know. and keep in mind that we would actually need `enabled`, and as you know django template doesn't like negations.

karyon:
:+1:

PR #648, opened 2015-10-11
Only show graded courses to graders

### LINKED ISSUE #645: Allow to exclude courses from grader view
It should be possible to exclude courses from the semester overview on the grading page.
Single results should never be listed on the page and there should be a new attribute for each course defining whether the course is shown for graders or not.
This must also include a symbol for staff users, showing them that this course can't get grade documents.

janno42:
fix #645
the existing attribute `is_graded` is used to define whether courses are visible for graders.


karyon:
how is the "gradedness" of a course determined until now? if that happens in the importer, we should maybe remove it there?

janno42:
Or we could keep it. Just a few days ago I used the importer to create an ungraded course.

karyon:
k

:+1:

PR #652, opened 2015-10-30
Add contribution labels

### LINKED ISSUE #651: Contribution label
There should be an option to add a label to contributors which is then shown in the questionnaire.

A use case might be an event where each contributor should be evaluated for one specific talk. The talk's name can be added as additional information along the contributor's name so that participants can remember who the person was.

janno42:
Fix #651
This allows adding labels to contributions. The label is then shown in the questionnaire like shown below (here: "EvaP Talk"):
label


karyon:
:baby_chick:

PR #656, opened 2015-11-06
Fix non-sortable questionnaires

### LINKED ISSUE #653: Contributor questionnaires not sortable
On the questionnaire page the contributor questionnaires are not sortable.

karyon:
Fixes #653

### DIFF
--- a/evap/static/js/sortable_query_binding.js
+++ b/evap/static/js/sortable_query_binding.js
@@ -1,0 +1,61 @@
+/**
+ * jQuery plugin for Sortable
+ * @author	RubaXa   <trash@rubaxa.org>
+ * @license MIT
+ */
+(function (factory) {
+	"use strict";
+
+	if (typeof define === "function" && define.amd) {
+		define(["jquery"], factory);
+	}
+	else {
+		/* jshint sub:true */
+		factory(jQuery);
+	}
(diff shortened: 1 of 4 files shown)


evap/static/js/sortable_query_binding.js
```diff
@@ -0,0 +1,61 @@
+/**
```
janno42: why do we need that?

karyon: to make those other changes possible. before i was getting raw DOM elements with get(), and now i can directly call sortable() on a jquery object. with this i also don't need to iterate over the two tables in the questionnaire index manually.

janno42: so much code. i especially like the `/* CODE */`. that's helpful. but you're the clean-up guy. so if you're okay with it, it's fine.

janno42:
:+1:

evap/static/js/sortable_query_binding.js
karyon: i'm not sure what you mean, this is a lib just like the core sortable.js or bootstrap...?

karyon: yeah i use only a few lines of this plugin but that's how using libs works.

PR #659, opened 2015-11-08
Replace profile with settings page

### LINKED ISSUE #262: remove lecturer profile, introduce settings
the lecturer profile is largely useless right now. lecturers are not supposed to change their name or email address. the picture is useless anyway

The only useful part is the delegate multiselect box and the list of "People for which you are a delegate". those could be moved to a dedicated "Delegate" page or merged into the lecturer overview page.

janno42:
fix #262 

settings


evap/contributor/forms.py
```diff
@@ -87,10 +87,13 @@ def clean(self):
...
 
-class UserForm(forms.ModelForm, BootstrapMixin):
+class DelegatesForm(forms.ModelForm, BootstrapMixin):
+    delegate_of = forms.IntegerField()
+    cc_user_of = forms.IntegerField()
+
     class Meta:
         model = UserProfile
-        fields = ('title', 'first_name', 'last_name', 'email', 'delegates')
+        fields = ('delegates', 'cc_users',)
```
karyon: why are delegate_of and cc_user_of up there and not here?

evap/contributor/forms.py
```diff
@@ -98,25 +101,22 @@ def __init__(self, *args, **kwargs):
...
-        return self.cleaned_data['first_name'].strip()
+        represented_users = self.instance.represented_users.all()
+        self.fields['delegate_of'] = forms.ModelMultipleChoiceField(represented_users, initial=represented_users, help_text="", required=False)
+        self.fields['delegate_of'].widget.attrs['disabled'] = True
 
-    def clean_last_name(self):
-        return self.cleaned_data['last_name'].strip()
+        ccing_users = self.instance.ccing_users.all()
+        self.fields['cc_user_of'] = forms.ModelMultipleChoiceField(ccing_users, initial=ccing_users, help_text="", required=False)
+        self.fields['cc_user_of'].widget.attrs['disabled'] = True
```
karyon: why do the first two have their help text set but these here not?

evap/contributor/forms.py
```diff
@@ -98,25 +101,22 @@ def __init__(self, *args, **kwargs):
...
 
-    def clean_title(self):
-        return self.cleaned_data['title'].strip()
+    def clean_cc_users(self):
+        return self.instance.cc_users.all()
```
karyon: why is this field cleaned/reset and the other two disabled ones not?

evap/contributor/forms.py
janno42: they are not part of the model, so they must be included differently.

evap/contributor/forms.py
janno42: the others are not part of the model and are filled manually. they are not saved on saving the form.

evap/contributor/forms.py
```diff
@@ -98,25 +101,22 @@ def __init__(self, *args, **kwargs):
...
 
+    def save(self, *args, **kw):
+        user = kw.pop("user")
+        super().save(*args, **kw)
+        self.instance.save()
```
karyon: what does this do? i'm pretty sure super().save also saves the instance. also, when is the user parameter different from self.instance?

evap/contributor/templates/contributor_settings.html
```diff
@@ -0,0 +1,111 @@
...
+    <li>{% trans "Settings" %}</li>
+{% endblock %}
+
+{% block content %}
+    {{ block.super }}
+    <h2>{% trans "Settings" %}</h2>
+
+    <div class="panel panel-default">
+        <div class="panel-heading">
+            <span class="panel-title">{% trans "Personal information" %} <span data-toggle="tooltip" data-placement="left" class="glyphicon glyphicon-info-sign pull-right" title="{% trans "To change any of these settings, please contact the student representatives." %}"></span></span>
```
karyon: line breaks please

evap/contributor/templates/contributor_settings.html
```diff
@@ -0,0 +1,111 @@
...
+            </table>
+        </div>
+    </div>
+
+    <form method="POST" class="form-horizontal multiselect-form" enctype="multipart/form-data">
+        {% csrf_token %}
+        {{ form.non_field_errors }}
+        <div class="panel panel-default">
+            <div class="panel-heading">
+                <span class="panel-title">{% trans "Delegates and CC users" %} <span data-toggle="tooltip" data-placement="left" class="glyphicon glyphicon-info-sign pull-right" title="{% trans "You can only change your own delegates. To become a delegate of another person this person has to add you. Please contact the student representatives to change the CC settings." %}"></span></span>
```
karyon: line breaks

evap/contributor/templates/contributor_settings.html
```diff
@@ -0,0 +1,111 @@
...
+                        </div>
+                    </div>
+                    <div class="col-sm-6">
+                        <div class="panel panel-default">
+                            <div class="panel-heading">
+                                <span class="panel-title">{% trans "Emails sent to these users will be CC'ed to you" %}</span>
+                            </div>
+                            <div class="panel-body">
+                                {{ form.cc_user_of.errors }}
+                                {{ form.cc_user_of }}
```
karyon: the last three can't have any errors, can they?

evap/staff/tests.py
```diff
@@ -1229,20 +1228,14 @@ def test_user_with_same_email(self):
...
         form = UserForm(data=data)
         self.assertFalse(form.is_valid())
-        form = ContributorUserForm(data=data)
-        self.assertFalse(form.is_valid())
 
         data = {"username": "uiae", "email": user.email.upper()}
         form = UserForm(instance=user, data=data)
         self.assertTrue(form.is_valid())
-        form = ContributorUserForm(instance=user, data=data)
-        self.assertTrue(form.is_valid())
```
karyon: please replace these tests with new ones instead of just removing them.

evap/static/css/evap.css
```diff
@@ -474,3 +474,7 @@ li.menu-indent {
 .table-striped tbody tr.sortable-ghost {
     background: #E5E5E5;
 }
+
+select {
+    width: 100%;
+}
```
karyon: doesn't this affect other pages? how does it look there?

evap/templates/base.html
```diff
@@ -239,7 +239,12 @@
                     placeholder: "{% trans "Please select..." %}",
                     minimumResultsForSearch: 15,});
             };
-            apply_select2($('select').not('.form-template select'));
+            apply_select2($('select').not('.form-template select').not('[disabled]'));
+
+            apply_select2_disabled = function(elem) {
+                elem.select2();
+            };
+            apply_select2_disabled($('select[disabled]').not('.form-template select'));
```
karyon: why is this necessary? why can't you use `apply_select2` to disabled elements?

evap/contributor/forms.py
janno42: help text for all fields removed

evap/contributor/forms.py
janno42: thanks. changed.

evap/contributor/forms.py
karyon: also, wouldn't it be cooler to declare `delegate_of` and `cc_user_of` as `ModelMultipleChoiceField` right away (probably with `queryset=None`) and then set `queryset` and `initial` (and `required` if you actually need to) here?

karyon: (if you think that would be cooler, please also change this is staff.UserForm)

(even if you don't think it's cooler, have a look at the user form and see whether there is similar code that diverged from this one now)

evap/contributor/templates/contributor_settings.html
janno42: done

evap/contributor/templates/contributor_settings.html
janno42: done

evap/contributor/templates/contributor_settings.html
janno42: they shouldn't. do you want me to remove those parts?

evap/staff/tests.py
janno42: the part tested here doesn't exist anymore.

evap/static/css/evap.css
janno42: it's fine. other form-elements are styled explicitly and i didn't find any other occurrences which would now change their layout.

evap/templates/base.html
janno42: i could, but i don't want to have the "Please select..." text when there is nothing the user can select ;)

evap/staff/tests.py
karyon: i know, but still, you removed some code and replaced it with new code, and removed the tests without replacing them with new ones. what i want to say is that a test testing the form would be cool.

evap/templates/base.html
karyon: as this wasn't really obvious, please add a comment.

evap/static/css/evap.css
karyon: i don't know much about how to do css, but shouldn't this rule rather apply to only the select2 fields on that specific page? i mean if we generally do not want it to be 100%...

evap/contributor/forms.py
janno42: it's cooler. changed.

evap/static/css/evap.css
janno42: we also do on other pages, but there they are already sized correctly. i can think about something else if need be.

janno42:
test added

karyon:
if you really think the css is done the right way, squash and then :shipit: :

PR #661, opened 2015-11-09
Added error message when 0 points redeemed. Fixes  #632

### LINKED ISSUE #632: Reward point redemption error message
When users try to redeem 0 reward points they get the error message "You don't have enough reward points". The error message should be something like "You must select the number of reward points you want to redeem" instead.

klaraseitz:


evap/rewards/tools.py
```diff
@@ -15,8 +15,11 @@ def save_redemptions(request, redemptions):
     total_points_available = reward_points_of_user(request.user)
     total_points_redeemed = sum(redemptions.values())
 
-    if total_points_redeemed == 0 or total_points_redeemed > total_points_available:
-        return False
+    if total_points_redeemed <= 0:
+        return False, "zero_points"
+
+    if total_points_redeemed > total_points_available:
+        return False, "not_enough_points"
```
karyon: such a "stringly typed" is quite error-prone, for example it's easy to make a typo when checking for those error strings, which would be hard to notice. could you try replacing those string error codes with exceptions? so, instead of returning a bool and string either return nothing (on success) or raise an exception on failure. you can use two different exceptions (which you would need to add) for the two errors, and then catch these two here to print out the different error messages.

PR #664, opened 2015-11-09
Edit public view to not show results for not published courses. Fixes #625

### LINKED ISSUE #625: public view shows results in courses even if they were not published
see screenshot. it even says the results were not published, but i can see them although i'm in the public view.

when implementing this, please see whether it is possible to actually use the same code for the "public view" of contributors and the view in the results app.

simonkrogmann:


karyon:
ref #625

we discussed this and, as this public view does not exist at all when the course was not published, we think it's better to just remove the "public view" button when the course was not published.

so, with your check whether a user can see a result, could you pass a new parameter to the template and not render that button depending on that?

...although i don't know whether the copying is sane, i'll need to check that

janno42:
there already is a `sufficient_votes_warning` which triggers the warning message when a course's results have not been published. that could be used to hide the toggle button for the public view.

simonkrogmann:
Now the public view button is only shown for courses that are publically viewable.

janno42:
please only check against `elif not sufficient_votes_warning` :)

simonkrogmann:
Done

janno42:
and now one final request :)
could you please squash all the commits together into one?

simonkrogmann:
Also done

janno42:
you have the `evaluation_warning` check back in there...

simonkrogmann:
Oops... I must have made a mistake while squashing. Now it should be correct.

janno42:
thanks!

Issue #623, opened 2015-09-07
Translation of Grade Documents


Nef10:

janno42:
Could you please be so kind and give us a short problem description? :)

Nef10:
missing translation 
<img width="616" alt="screenshot 2015-09-07 12 57 39" src="https://cloud.githubusercontent.com/assets/4551135/9715417/266eace0-5560-11e5-8628-e3c8f406807a.png">

janno42:
In the dropdown menu where students can download their grades the name of the document is not translated, beacuse it's a name that the Studienreferat enters. The respective field is getting a default entry "Endnoten" in case they don't put a different name there. So there is no translation yet.

Paul-Geppert:
Please assign me to this ticket. My suggestion is the same as janno42 has ("We could allow entering two names (German and English) to have a translation.").
I'm also assigned to the issue #628, which needs the solution of this issue.

janno42:
fixed in #666

Issue #628, opened 2015-09-16
Direct download


janno42:
For courses where only one grade document exists, this document should be downloaded directly when clicking on the download button in the student overview or on the course's results page.
Currently a dropdown menu is used even when only one document exists, this should only be the case if there are multiple documents available.

Paul-Geppert:
Please assign me, I would like to do this ticket.

PR #666, opened 2015-11-09
Download grades directly Fix for #628 ---> Translations?

Paul-Geppert:
Fix for #628 
On students site, I changed the student_index.html, that only one button for downloading the grades is shown, if only one document with grade exists.
It would be great, for having translations for the description of the grade-documents (like Midterm Notes --> Zwischennote), but I didn't find a translation script for it. The version for downloading more than one grade file didn't use a translation for the description too, only for "Download Grades".


janno42:
for translation see #623

evap/student/templates/student_index.html
```diff
@@ -117,4 +121,4 @@
         </div>
         {% endif %}
     {% endfor %}
-{% endblock %}
+{% endblock %}
```
Steditor: Please add a newline at the end of the file.

janno42:
Looks good! Can you please do the same on the results page (`results/templates/results_course_detail.html`)?

And please squash all you commits into a single commit when you are done.

Paul-Geppert:
I finished it. It's possible to download the grades directly from student and course side now.
I will squash my commits, thanks for this. This commit should be the very last for this issue, if you don't count #623 .

evap/results/templates/results_course_detail.html
```diff
@@ -33,7 +33,11 @@
     <div class="panel panel-info">
         <div class="panel-heading">
             <span class="panel-title">{% trans "Overview" %}</span>
-            {% if course.grade_documents.count > 0 and can_download_grades %}
+            {% if course.grade_documents.count = 1 and can_download_grades %}
```
janno42: ==

evap/student/templates/student_index.html
```diff
@@ -87,7 +87,9 @@
                                             <a href="{% url "student:vote" course.id %}" class="btn btn-sm btn-primary">{% trans "Evaluate" %}</a>
                                         {% endif %}
                                     {% endif %}
-                                    {% if course.grade_documents.count > 0 and can_download_grades %}
+                                    {% if course.grade_documents.count = 1 and can_download_grades %}
```
janno42: `==`

Paul-Geppert:
This commit should resolve #623 and #628. I removed the drop-down menu for downloading grades, if only one grade document is available.
To have a translation for grade documents, I added a description-field for german and english and removed the old field for description.
To reach that, we added migration files, which add the new attributes to the database, save the old data, set the unique-dependencies (AlterUniqueTogether) and remove the attribute 'description'.
Beware: It is not revertable! That comes from the fact, that the AlterUniqueTogether-operation needs existing column and data to work. When reverting, one has to fill the 'description'-column with data first, so the position of the AlterUniqueTogether-operation is wrong for reverting.
Would that be changed, the migration has it now is, wouldn't work.

evap/grades/forms.py
```diff
@@ -26,11 +25,17 @@ def __init__(self, *args, **kwargs):
...
+    def clean_description_de(self):
+        description_de = self.cleaned_data.get('description_de')
+        if GradeDocument.objects.filter(course=self.instance.course, description_de=description_de).exclude(id=self.instance.id).exists():
             raise ValidationError(_("This description for a grade document was already used for this course."))
-        return description
+        return description_de
+
+    def clean_description_en(self):
+        description_en = self.cleaned_data.get('description_en')
+        if GradeDocument.objects.filter(course=self.instance.course, description_en=description_en).exclude(id=self.instance.id).exists():
```
karyon: while you're at it, could you make this filter ignore case so we can catch more user errors here? ctrl-f for `iexact` in the django queryset docs.

evap/student/templates/student_index.html
karyon: did you talk to @janno42 about this and the other similar comment or address this somewhere else?

evap/grades/migrations/0009_translate_description.py
```diff
@@ -0,0 +1,23 @@
...
+    operations = [
+        migrations.AlterUniqueTogether(
+            name='gradedocument',
+            unique_together=set([('course', 'description_de'), ('course', 'description_en')]),
+        ),
+        migrations.RemoveField(
+            model_name='gradedocument',
+            name='description',
+        ),
+    ]
```
karyon: i'll think about the migration tomorrow. just a quick thought, @janno42 is it possible to just rename description to description_de, create a new description_en and add a unique together on that one? (maybe make description a CharField beforehand)

evap/student/templates/student_index.html
Paul-Geppert: No, we didn't talk about it. I tested both versions, and both work. If == instead of = is better, I could fix it.

evap/grades/forms.py
Paul-Geppert: Okay, I will think about more details for the filter.

evap/student/templates/student_index.html
janno42: yes, please use `==`

evap/grades/migrations/0009_translate_description.py
janno42: yes, i think that should work. before changing the unique_together we still need the data migration that puts stuff into description_en, but then it should be possible.

karyon: @Paul-Geppert do you want to try this? what i'd do is start over with the base commit, rename description_de, make it a CharField, create CharField description_en, create migration. then add the data migration, and then add the second unique_together and create another migration. do a commit, and then somehow replace the migrations in your second commit with the new ones.

if you want to do something else, let me know, then i'd do this. or we look at it together during the next hacking hours.

Paul-Geppert: Yes, I will try it and show you the result next meeting.

karyon: nice. let me know if you encounter any problems.

Paul-Geppert:
@karyon please take over the migrations. I did little bugfixes discussed last meeting like == instead of =

karyon:
added new migrations and merged it manually, thanks!

PR #668, opened 2015-11-11
Set comment_visibility for responsible person on import

### LINKED ISSUE #654: Comment visibility selection after import
After importing a course, the responsible person does not have `All` selected for the comment visibilty in the UI.
The importer should set `comment_visibility` on creation.

janno42:
fix #654: set comment_visibility on import and update test_data

### DIFF
--- a/evap/evaluation/fixtures/test_data.json
+++ b/evap/evaluation/fixtures/test_data.json
@@ -113990,7 +113990,7 @@
         2
       ],
       "is_superuser": true,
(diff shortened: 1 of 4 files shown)


karyon:
as discussed, please override save() on the contribution anddo something like`if responsible: assert can_edit and comment_visibilty == "ALL" and

janno42:
please check again :)

karyon:
:package:

PR #670, opened 2015-11-19
Switch back to original puppetlabs-concat and puppetlabs-postgresql as karyon's changes are merged.

Steditor:
Original cause for using the fork: https://github.com/puppetlabs/puppetlabs-concat/pull/373

This also has the benefit that https://github.com/puppetlabs/puppetlabs-concat/pull/374 will be merged into the original project soon (hopefully) and resolve the issue of wrong line endings of quoted_paths_spec.rb (which occurs on windows as well and leads to the annoying message, that the submodule had been changed and is on a detached head).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "puppetlabs-concat"]
 	path = deployment/testing_environment/modules/concat
-    url = https://github.com/karyon/puppetlabs-concat.git
+    url = https://github.com/puppetlabs/puppetlabs-concat.git
 [submodule "puppetlabs-stdlib"]
 	path = deployment/testing_environment/modules/stdlib
     url = https://github.com/puppetlabs/puppetlabs-stdlib.git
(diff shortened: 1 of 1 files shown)


karyon:
i think a commit hash of the submodule is also part of our repo and pins the commit that is used. this would need to be updated as well, but we should wait until https://github.com/puppetlabs/puppetlabs-concat/pull/374 is merged. also, i'd do the same thing for the postgres submodule so we don't use my forks anymore.

Steditor:
Somehow I can't pin down the file in which the commit hash is stored. I can see it in your original pull request #655, but not in the repository itself :confused:

karyon:
i think you need to go into the submodule with your bash/sourcetree/whatever, checkout a new commit, and then you should be able to commit the changed pin in the parent project. but as i said i'd wait until they have merged the PR.

Steditor:
I see. The problem why git did not update the hash reference on my machine was also due to the error https://github.com/puppetlabs/puppetlabs-concat/pull/374 will solve.
I'll wait for them to merge and will then update the PR.

Steditor:
The CRLF-fixes were merged into puppeltlabs-concat in https://github.com/puppetlabs/puppetlabs-concat/pull/375
I have updated this PR and now also included the commit hash - should be ready for merge now.

karyon:
could you do the same with the postgresql submodule?

PR #672, opened 2015-11-23
fix #660

### LINKED ISSUE #660: Remove single results from course list in user form
On the user editing page, single results should not be listed in the course list (Courses participating in (active semester)).

drs-m:
exclude all courses for wich is_single_result() returns true

fixing #660


karyon:
could you describe in the commit message what has been done? something like "made that button blue, fixes #9001"

drs-m:
done

PR #684, opened 2015-12-08
Exclude single results from export

### LINKED ISSUE #683: Remove single results from export
When exporting a semester with single results a `ZeroDevisionError` occurs, because there are no participants.
Single results should not be part of the semester export.

janno42:
fix #683
exclude single results from export and avoid division by zero if a course has no participants

### DIFF
--- a/evap/results/exporters.py
+++ b/evap/results/exporters.py
@@ -72,6 +72,8 @@ class ExcelExporter(object):
     def export(self, response, ignore_not_enough_answers=False):
         courses_with_results = list()
         for course in self.semester.course_set.filter(state="published").all():
+            if course.is_single_result():
+                continue
             results = OrderedDict()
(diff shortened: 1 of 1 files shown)


evap/results/exporters.py
```diff
@@ -72,10 +72,11 @@ def deviation_to_style(self, deviation):
     def export(self, response, ignore_not_enough_answers=False):
         courses_with_results = list()
         for course in self.semester.course_set.filter(state="published").all():
-            results = OrderedDict()
-            for questionnaire, contributor, label, data, section_warning in calculate_results(course):
-                results.setdefault(questionnaire.id, []).extend(data)
-            courses_with_results.append((course, results))
+            if not course.is_single_result():
```
karyon: if is_single_result: continue

karyon:
:+1:

PR #688, opened 2015-12-16
Compute average for all contributors of each type on export

### LINKED ISSUE #687: Exporter only uses first contributor per type
The exporter only uses the results of the first contributor per type.
It should instead compute the average of all contributors per type.

janno42:
fix #687


karyon:
:mega:

PR #689, opened 2015-12-16
Weigh grades for exported averages based on how many votes went into it

janno42:
Right now, the average grade for each type of contributor in the export is computed as the average of all averages. That gives unwanted results if e.g. one of the averages is based on 10 votes and the other one only on 1.

This weighs the persons' averages based on how many votes they got to compute an overall average for a question.

### DIFF
--- a/evap/evaluation/tools.py
+++ b/evap/evaluation/tools.py
@@ -8,6 +8,7 @@ from collections import OrderedDict, defaultdict
 from collections import namedtuple
 from functools import partial
 from math import ceil, sqrt
+from statistics import pstdev, median
 
 GRADE_COLORS = {
     1: (136, 191, 74),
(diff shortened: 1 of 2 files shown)


karyon:
okay, but i don't like that you're expanding the counts like that. multiplying the average with the count and keeping track of how many answers were in there would be more appropriate, i think.

we have exactly the same problem in calculate_average_grades_and_deviation in evaluation/tools.py. please have a look at that and see whether you can build an averaging function that doesn't expand the counts. while at it, you could also look at _calculate_results_impl which does expand the counts as well :)

thinking about it, the averaging of deviations is bollocks. suppose we have a contributor getting a 1 and one getting a 5, both with no deviation, the export would show a 3 with no deviation. that might not be a problem for the averaging of deviations for the overall grade, because there we want to have "how much did the participants disagree with each other" instead of "how much did the contributors differ from each other".

janno42:
we need the individual answers in `_calculate_results_impl` to calculate the standard deviation. the values there can't be aggregated in advance.

the deviation is supposed to show the same thing everywhere. i think we should keep the current implementation. averaging the values for multiple contributors is always tricky but for the standard deviation we should have the same meaning everywhere: how much did the individual answers differ from each other. to my mind averaging the values over all contributors is the only (but admittedly just slighty) useful option. it makes no sense to put the answers from different contributors together and compute an overall deviation. so i vote for keeping the average just like now.

janno42:
updated

janno42:
go for it :diamond_shape_with_a_dot_inside:

PR #690, opened 2015-12-22
Update to Django 1.9

karyon:
this updates EvaP to django 1.9.


karyon:
three more commits added, one of which addresses the problem you found.

janno42:
:gift:

PR #693, opened 2015-12-29
Form cleanup

karyon:
See commit messages.

I'm not quite sure about the last commit, as i didn't find out the reason why it was written in the first place... i manually tested it and copying seems to work, so....


janno42:
looks good. even the last one ;)
:+1:

PR #694, opened 2015-12-29
Keep obsolete questionnaires

### LINKED ISSUE #593: Add "staff_visible" attribute to questionnaires
When editing a course that uses obsolete questionnaires these are not shown in the form and not being saved on submitting the form. Basically this removes obsolete questionnaires when editing a course.
What should happen is that obsolete questionnaires that are used by a course should still be included in the form, only the other obsolete questionnaires should be removed from the list.

karyon:
Fixes #593 

what do you think about the third commit? i thought it was missing but maybe it's intentional...


evap/staff/views.py
```diff
@@ -644,21 +644,46 @@ def questionnaire_create(request):
...
 @staff_required
 def questionnaire_edit(request, questionnaire_id):
     questionnaire = get_object_or_404(Questionnaire, id=questionnaire_id)
-    InlineQuestionFormset = inlineformset_factory(Questionnaire, Question, formset=AtLeastOneFormSet, form=QuestionForm, extra=1, exclude=('questionnaire',))
+    editable = questionnaire.can_staff_edit
 
-    form = QuestionnaireForm(request.POST or None, instance=questionnaire)
-    formset = InlineQuestionFormset(request.POST or None, instance=questionnaire)
+    if not editable:
+        messages.info(request, _("Some fields are disabled as this questionnaire is already in use."))
```
janno42: calling this here shows the message also when the form was valid and the user was redirected to the index page after saving changes.

evap/contributor/forms.py
```diff
@@ -13,7 +14,7 @@
 
 
 class CourseForm(forms.ModelForm, BootstrapMixin):
```
janno42: please do the equivalent changes also in the staff edit form. obsolete questions must be shown in the list if they are in use-

karyon:
updated.

janno42:
looks good :hamster:

PR #695, opened 2015-12-29
Export per type

### LINKED ISSUE #692: Export per type
The exporter should allow exporting only a subset of course types. The resulting Excel file should hold a configurable number of sheets each containing a configurable subset of course types. E.g. one holds all lectures and seminars, another one all projects.

janno42:
fix #692; enable exporting courses per type and move export view into staff app

### DIFF
--- a/evap/evaluation/tools.py
+++ b/evap/evaluation/tools.py
@@ -2,7 +2,7 @@ from django.conf import settings
 from django.core.cache import cache
 from django.utils.translation import ugettext_lazy as _
 from django.db.models import Sum
-from evap.evaluation.models import TextAnswer, EmailTemplate
+from evap.evaluation.models import TextAnswer, EmailTemplate, Course
 
 from collections import OrderedDict, defaultdict
 from collections import namedtuple
@@ -330,3 +330,6 @@ def sort_formset(request, formset):
     if request.POST: # if not, there will be no cleaned_data and the models should already be sorted anyways
         formset.is_valid() # make sure all forms have cleaned_data
         formset.forms.sort(key=lambda f: f.cleaned_data.get("order", 9001) )
+
+def course_types_in_semester(semester):
+    return Course.objects.filter(semester=semester).values_list('type', flat=True).order_by().distinct()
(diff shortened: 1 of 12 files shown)


evap/evaluation/tools.py
```diff
@@ -330,3 +330,10 @@ def sort_formset(request, formset):
     if request.POST: # if not, there will be no cleaned_data and the models should already be sorted anyways
         formset.is_valid() # make sure all forms have cleaned_data
         formset.forms.sort(key=lambda f: f.cleaned_data.get("order", 9001) )
+
+def course_types_in_semester(semester):
+    courses = Course.objects.filter(semester=semester)
+    types = set()
+    for course in courses:
+        types.add(course.type)
+    return types
```
karyon: `return Course.objects.filter(semester=semester).values_list('type', flat=True).distinct()`
or something like that... and then you could inline it. add an `order_by` if necessary.

evap/staff/views.py
```diff
@@ -285,6 +289,28 @@ def semester_import(request, semester_id):
...
 
 @staff_required
+def semester_export(request, semester_id):
+    semester = get_object_or_404(Semester, id=semester_id)
+
+    ExportSheetFormset = formset_factory(form=ExportSheetForm, can_delete=True, extra=1)
+    formset = ExportSheetFormset(request.POST or None, form_kwargs={'semester': semester})
+
+    if formset.is_valid():
+        ignore_not_enough_answers = request.POST.get('ignore_not_enough_answers', False)
```
karyon: you sure you don't need to convert to a boolean here?

janno42: if it's checked it might be something like `on` or `selected`. but all non-empty strings are true.

evap/staff/views.py
```diff
@@ -285,6 +289,28 @@ def semester_import(request, semester_id):
...
+        course_types = []
+        for form in formset:
+            course_types.append(form.cleaned_data.get('selected_course_types'))
```
karyon: course_types = [form.cleaned_data['selected_course_types'] for form in formset]

evap/staff/views.py
```diff
@@ -285,6 +289,28 @@ def semester_import(request, semester_id):
...
+
+        filename = "Evaluation-%s-%s.xls" % (semester.name, get_language())
```
karyon: are you aware that the semester name is localized as well? so it would be "sommersemester 13-de" and "summer term 13-en". also, the spaces look a bit weird to me, or rather, that there are none around the dashes.

karyon:
:facepunch:

PR #696, opened 2015-12-29
Revert "Prevent the navbar from overflowing. fixes #512"

### LINKED ISSUE #512: Menu bar hides content

### LINKED ISSUE #641: hamburger menu shows up unnecessarily
with fixing https://github.com/fsr-itse/EvaP/issues/512 the hamburger menu now shows up when the page width is below something like 1360px. this happens even when the user does not have all the menu items, and it's confusing for the user as there is no apparent reason for this.

karyon:
Reverts 531adb4cc5a2b31af26bb814c715a2a344e4338e. The navbar will now overflow, but just for the few users who have too many menu items. For most users, this will be an improvement. Fixes #641.

### DIFF
--- a/evap/static/css/bootstrap.less
+++ b/evap/static/css/bootstrap.less
@@ -57,7 +57,3 @@
 
 
 // Custom variables
-
-// prevent the navbar from overflowing
-@grid-float-breakpoint:         1360px;
-@nav-link-padding:              10px 13px;


janno42:
:dart:

PR #698, opened 2016-01-04
Add option to export email addresses in send mail form. Closes #619

### LINKED ISSUE #619: Option for staff to get list of all participants' email addresses

simonkrogmann:
There is no an export button in the send email form, which displays the email addresses of the currently selected groups to a field below.

### DIFF
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -202,8 +202,11 @@ class CourseEmailForm(forms.Form, BootstrapMixin):
 
     def __init__(self, *args, **kwargs):
         self.instance = kwargs.pop('instance')
+        self.export = kwargs.pop('export', False)
         self.template = EmailTemplate()
         super().__init__(*args, **kwargs)
+        self.fields['subject'].required = not self.export
+        self.fields['body'].required = not self.export
 
     def clean(self):
         self.recipient_groups = self.cleaned_data.get('recipients')
@@ -213,10 +216,16 @@ class CourseEmailForm(forms.Form, BootstrapMixin):
 
         return self.cleaned_data
 
-    # returns the number of recepients without an email address
+    # returns the number of recipients without an email address
     def missing_email_addresses(self):
         recipients = self.template.recipient_list_for_course(self.instance, self.recipient_groups)
         return len([user for user in recipients if not user.email])
+
+    def email_addresses(self):
+        if self.recipient_groups is None:
+            return []
+        recipients = self.template.recipient_list_for_course(self.instance, self.recipient_groups)
+        return set(user.email for user in recipients if user.email)
 
     def send(self):
         self.template.subject = self.cleaned_data.get('subject')
--- a/evap/staff/templates/staff_course_email.html
+++ b/evap/staff/templates/staff_course_email.html
@@ -17,14 +17,14 @@
                     <div class="form-row">
                         {% if field.name == "recipients" %}
                             <div class="form-label vertically-aligned {% if field.errors %}error{% endif %}" name="{{ field.name }}">{{ field.label }}</div>
-                            <div class="form-field btn-group" data-toggle="buttons">
+                            <div class="btn-group" data-toggle="buttons">
                                 {% for choice in field %}
                                     <label class="btn btn-sm btn-default {% if choice.choice_value in field.value %}active{% endif %} {% if field.errors %}error{% endif %}" name="{{ field.name }}" onclick="changedFieldValue('{{ field.name }}')">
                                         <input id="{{ choice.id_for_label }}" name="{{ choice.name }}" type="checkbox" value="{{ choice.choice_value }}" autocomplete="off" {% if choice.choice_value in field.value %}checked{% endif %} />
                                         {{ choice.choice_label }} {{ choice.id }}
                                     </label>
                                 {% endfor %}
-                            </div>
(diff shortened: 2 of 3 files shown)


evap/staff/views.py
```diff
@@ -524,6 +524,12 @@ def course_email(request, semester_id, course_id):
     course = get_object_or_404(Course, id=course_id)
     form = CourseEmailForm(request.POST or None, instance=course)
 
+    if 'export' in request.POST and 'recipients' in request.POST:
+        form.recipient_groups = request.POST['recipients']
+        email_addresses = ", ".join(form.email_addresses())
+        form2 = CourseEmailForm(None, instance=course)
+        return render(request, "staff_course_email.html", dict(semester=semester, course=course, form=form2, addresses=email_addresses))
```
karyon: did you try using the original `form` here, did that work? because as i understand it, if the user would use the export as "show me the emailadresses before i hit send", now his selection of recipients would be reset?

evap/staff/templates/staff_course_email.html
```diff
@@ -31,13 +33,16 @@
                         {% elif field.name == "body" %}
                             <div class="form-label {% if field.errors %}error{% endif %}" name="{{ field.name }}">{{ field.label }}</div>
                             <div class="form-field"><textarea class="form-control {% if field.errors %}error{% endif %}" cols="40" rows="10" id="{{ field.id_for_label }}" name="{{ field.name }}" oninput="changedFieldValue('{{ field.name }}');"></textarea></div>
+                        {% elif field.name == "addresses" and addresses %}
+                            <div class="form-label " name="{{ field.name }}">{{ field.label }}</div>
+                            <div class="form-field"><input class="form-control {% if field.errors %}error{% endif %}" id="{{ field.id_for_label }}" name="{{ field.name }}" type="text" value="{{ addresses }}"></div>
```
karyon: if i understand it correctly, you're not really using that `addresses` form field for user input, and even the data you render here comes as a additional template parameter not coupled to the form field, right? if yes, could you remove that field and render the data outside the form? @janno42 suggested to just put it into a message (search for `messages.info` in staff/views.py), one email address per line.

evap/staff/forms.py
```diff
@@ -217,6 +218,12 @@ def missing_email_addresses(self):
         recipients = self.template.recipient_list_for_course(self.instance, self.recipient_groups)
         return len([user for user in recipients if not user.email])
 
+    def email_addresses(self):
+        if self.recipient_groups == None:
```
karyon: `is None`

simonkrogmann:
Updated

evap/staff/forms.py
```diff
@@ -197,27 +197,41 @@ def save(self, *args, **kwargs):
...
-            raise forms.ValidationError(_("No recipient selected. Choose at least one group of recipients."))
+        if self.export:
+            if not self.recipient_groups:
+                raise forms.ValidationError(_("No recipient selected. Choose at least one group of recipients."))
+            return self.cleaned_data
+
+        for field_name in ['recipients', 'subject', 'body']:
+            data = self.cleaned_data.get(field_name)
+            if not data:
+                self.add_error(field_name, _("Please fill out all required fields and select at least one recipient group."))
```
karyon: i'm sorry, another suggestion. could you try handing over the export parameter into the constructor? and then pop it just like for the instance parameter. and then you could set the required parameter on the fields in the constructor depending on the export variable, and then remove those four lines here (and consequently show the `No recipient selected` always and not just `if export`.

evap/staff/templates/staff_course_email.html
```diff
@@ -17,13 +17,15 @@
...
-                            <div class="form-field btn-group" data-toggle="buttons">
+                            <div class="btn-group form-field">
+                            <div class="btn-group" data-toggle="buttons">
                                 {% for choice in field %}
                                     <label class="btn btn-sm btn-default {% if choice.choice_value in field.value %}active{% endif %} {% if field.errors %}error{% endif %}" name="{{ field.name }}" onclick="changedFieldValue('{{ field.name }}')">
                                         <input id="{{ choice.id_for_label }}" name="{{ choice.name }}" type="checkbox" value="{{ choice.choice_value }}" autocomplete="off" {% if choice.choice_value in field.value %}checked{% endif %} />
                                         {{ choice.choice_label }} {{ choice.id }}
                                     </label>
                                 {% endfor %}
+                            </div><input name="export" type="submit" value="{% trans "Export" %}" class="btn btn-sm btn-primary"/>
```
janno42: can you please move the button behind the `btn-group`? and please call it `Show recipients`

evap/staff/views.py
```diff
@@ -524,7 +524,13 @@ def course_email(request, semester_id, course_id):
     course = get_object_or_404(Course, id=course_id)
     form = CourseEmailForm(request.POST or None, instance=course)
 
+    form.export = 'export' in request.POST
+
     if form.is_valid():
+        if form.export:
+            email_addresses = ', '.join(form.email_addresses())
```
janno42: please use a semicolon (`;`)

evap/staff/views.py
```diff
@@ -524,7 +524,13 @@ def course_email(request, semester_id, course_id):
...
+            messages.info(request, _('Recipients: ') + email_addresses)
```
janno42: please use `messages.info(request, _('Recipients: ') + '\n' + email_addresses)`

Issue #489, opened 2015-01-15
Clean up inline css


janno42:
Style sheets currently defined within the templates should be moved into ```evap.css``` wherever possible.

lmm-git:
fixed by #735 (hopefully)

PR #699, opened 2016-01-04
fix table column width for grades semester view

### LINKED ISSUE #691: Buttons rendered wrong in firefox in grades semester view
should be side by side instead of one above the other.

schumannd:
fix #691 

I this solution counterproductive to #489 ?


karyon:
crazey man

@janno42 says in this case it's okay, since he wouldn't put css stuff for single elements into css files.

PR #701, opened 2016-01-04
add staff and gradeUser group checkboxes to user edit form. fixes #576

### LINKED ISSUE #576: Staff field on user page is missing
The user edit page does not have a staff field anymore.
Superusers should be able to add or remove the group there.

drs-m:


evap/staff/forms.py
```diff
@@ -333,6 +336,10 @@ def __init__(self, *args, **kwargs):
         self.fields['courses_participating_in'].initial = courses_of_current_semester.filter(participants=self.instance) if self.instance.pk else ()
         self.fields['courses_participating_in'].label = _("Courses participating in (active semester)")
         self.fields['courses_participating_in'].required = False
+        self.fields['is_staff'].initial = Group.objects.filter(user=self.instance, name="Staff").exists()
+        self.fields['is_staff'].label = _("Staff user")
+        self.fields['is_grade_user'].initial = Group.objects.filter(user=self.instance, name="Grade publisher").exists()
+        self.fields['is_grade_user'].label = _("Grade user")
```
karyon: there are `self.instance.is_staff` and `self.instance.is_grade_publisher`. also, could you specify the label as parameter to `BooleanField`?

evap/staff/forms.py
```diff
@@ -361,6 +368,18 @@ def clean_email(self):
...
+        grade_user_group = Group.objects.get(name="Grade publisher")
+        if (self.cleaned_data.get('is_staff')):
+            staff_group.user_set.add(self.instance)
+        else:
+            staff_group.user_set.remove(self.instance)
+
+        if (self.cleaned_data.get('is_grade_user')):
+            grade_user_group.user_set.add(self.instance)
+        else:
+            grade_user_group.user_set.remove(self.instance)
```
karyon: ifs without the brackets please :) and i'd prefer `self.instance.groups.add(group)` over `group.user_set...` it makes it clearer that we're modifying a user and not a group.

karyon:
if you address the two comments i made we can merge this. please amend to your commit and rebase on current master when you're done. thanks for the PR!

drs-m:
done

evap/staff/forms.py
```diff
@@ -333,6 +336,8 @@ def __init__(self, *args, **kwargs):
         self.fields['courses_participating_in'].initial = courses_of_current_semester.filter(participants=self.instance) if self.instance.pk else ()
         self.fields['courses_participating_in'].label = _("Courses participating in (active semester)")
         self.fields['courses_participating_in'].required = False
+        self.fields['is_staff'].initial = self.instance.is_staff
+        self.fields['is_grade_user'].initial = self.instance.is_grade_publisher
```
karyon: this broke the tests, it's probably enough to put these two lines into a `if self.instance.pk`. the problem is that you can't access m2m relationships (like groups on a user) before it has been saved to the database once because you need the user's id for that. this happens e.g. when using this form for creating a user.

i saw some other stuff i'd like to change in this method so i'll fix the tests while i'm at it. thanks again for the PR!

PR #703, opened 2016-01-08
Management commands

### LINKED ISSUE #700: Database reset command
manage.py should allow executing a command `reset_database` which
- resets the database completely,
- creates the cachetable,
- loads the test data file

### LINKED ISSUE #702: merge_users fails because of deprecated import
The `merge_users` command fails, because `get_models` can't be imported anymore.

karyon:
See commit messages. fixes #702, fixes #700. Enables `./manage.py run`.

Apache now runs always and is accessible on port 8001 on the host.

I called the new management command reload_testdata, i thought that was cooler than reset_testdata, but i wouldn't mind changing that.

The massive test_data change is due to a different ordering of django's default dumpdata. Should happen only this one time. dump_testdata now has a progress bar as a bonus.

I'm pretty tired so i have missed stuff that might break due to port changes, but i couldn't think of anything.


karyon:
@Bartzi the pycharm docs said that the debugger doesn't need any additional port. could you confirm? port 8000 (iirc it was you who forwarded that to 8080) was just the default one for the django development server, right? is there any reason to let it run on a different port when using pycharm that when using vagrant directly?

evap/evaluation/merge.py
```diff
@@ -45,7 +46,9 @@ def merge_model_objects(primary_object, alias_objects, keep_old=False):
     # Loop through all alias objects and migrate their data to the primary object.
     for alias_object in alias_objects:
         # Migrate all foreign key references from alias object to primary object.
-        for related_object in alias_object._meta.get_all_related_objects():
+        for related_object in alias_object._meta.get_fields():
```
janno42: could you name this `related_field`?

evap/evaluation/merge.py
```diff
@@ -59,11 +62,13 @@ def merge_model_objects(primary_object, alias_objects, keep_old=False):
                     obj.save()
 
         # Migrate all many to many references from alias object to primary object.
-        for related_many_object in alias_object._meta.get_all_related_many_to_many_objects():
+        for related_many_object in alias_object._meta.get_fields(include_hidden=True):
```
janno42: same as above

janno42:
looks good to me :rabbit:

Bartzi:
@karyon the pycharm debug helper do not need any ports as they use ssh tunneling (afaik). If someone wants to use the pycharm debugger with django he only needs access to the port opened by the django development server. If you open port 8000 everything should be fine.

karyon:
@Paul-Geppert fyi you'll likely get merge conflicts because i updated test_data. you don't need to handle that if you don't want to, i think it's time to get your PR merged and not give you more paperwork to do :)

PR #708, opened 2016-01-12
Import users into existing course (Fixes #248)

### LINKED ISSUE #248: Bulk import participants to a course
There should be an option for student representatives to add a list of external students to a course without having to first create each user individually.

This will be needed for D-School courses where the student representatives will get a list of participants (names and email addresses) of external students who are not part of the initial enrolment data.

The bulk import should create the given users (if they don't exist yet) and add them to a specified course.

schumannd:
Fixes #248


evap/staff/views.py
```diff
@@ -350,7 +357,7 @@ def semester_lottery(request, semester_id):
             if not courses.exclude(voters=user).exists():
                 eligible.append(user)
 
-        winners = random.sample(eligible, min([form.cleaned_data['number_of_winners'], len(eligible)]))
+        winners = sample(eligible, min([form.cleaned_data['number_of_winners'], len(eligible)]))
```
karyon: why?

karyon:
there's a file permission change on manage.py, why?

schumannd:
That is a mistake, I will fix it

evap/staff/views.py
```diff
@@ -538,6 +545,37 @@ def course_email(request, semester_id, course_id):
...
+        excel_file = form.cleaned_data['excel_file']
+
+        test_run = operation == 'test'
+
+        # Parse table.
+        imported_users = UserImporter.process(request, excel_file, test_run)
+
+        # Test run, or an error occurred while parsing -> stay and display error.
+        if test_run or not imported_users:
+            return render(request, "staff_import_participants.html", dict(course=course, form=form))
```
karyon: where does the "display error" part happen?

evap/staff/views.py
schumannd: It was suggested by PyCharm. Also makes the line shorter. random.sample might be easier to understand though, I see your point.

evap/staff/views.py
```diff
@@ -538,6 +545,37 @@ def course_email(request, semester_id, course_id):
...
+        else:
+            # Add users to course participants. * converts list into parameters.
+            course.participants.add(*imported_users)
```
karyon: could you check whether += works?

evap/staff/views.py
schumannd: unfortunately this is very implicit. UserImporter.process() sets message.errors which will be displayed on the same website in the case that imported_users == None instead of the semester_view page.

evap/staff/tests.py
```diff
@@ -85,6 +85,18 @@ def test_sample_user_xls(self):
...
+        page = self.app.get("/staff/semester/1/course/1/importparticipants", user='user')
+
+        original_participant_count = course.participants.count()
+
+        form = lastform(page)
+        form["excel_file"] = (os.path.join(settings.BASE_DIR, "static", "sample_user.xls"),)
+        form.submit(name="operation", value="import")
+
+        self.assertEqual(course.participants.count(), original_participant_count + 2)
+
```
karyon: that would be your opportunity to add an example how to properly test a view ;) you could add a new class somewhere in here and put all the tests in there.

evap/staff/tests.py
```diff
@@ -77,7 +77,7 @@ def test_sample_xls(self):
     def test_sample_user_xls(self):
         page = self.app.get("/staff/user/import", user='user')
 
-        original_user_count = UserProfile.objects.all().count()
+        original_user_count = UserProfile.objects.count()
```
karyon: if it's not too much work, please separate the unrelated changes into a new commit.

evap/staff/views.py
schumannd: It does not worl. += seems to not be implemented for the django RelatedManager

evap/staff/tests.py
schumannd: done

karyon: well, i meant the imports and whitespace changes as well.

karyon:
@janno42 what about users that already exist? with this implemention, the importer will update the user with new data from the input excel sheet and just continue. do we want that?

janno42:
no, this importer must also allow to check changes in user data beforehand. an import test like the one of the enrollment importer is suggested.

karyon:
you're right, the warnings for that are already in place and should just work.

schumannd:
all comments done

evap/staff/templates/staff_import_participants.html
```diff
@@ -0,0 +1,21 @@
+{% extends "staff_semester_base.html" %}
+
+{% load i18n %}
+{% load static %}
+
+{% block content %}
+    {{ block.super }}
+    <p>
+        {% trans "Upload course participants data from excel file (" %}<a href="{% get_static_prefix %}sample_user.xls">{% trans "Sample File"%}</a>{% trans "). This will create all containing students and enroll them into the course: " %} {{ course }}
```
janno42: Please use `This will create user accounts for all contained students and adds them as participants for the course`

evap/staff/templates/staff_semester_view_course.html
```diff
@@ -90,11 +90,13 @@
             <span class="glyphicon glyphicon-user"></span>
             <span>{{ course.num_participants }}</span>
         </p>
+        <a href="{% url "staff:course_import_participants" semester.id course.id %}" class="btn btn-sm btn-default" data-toggle="tooltip" data-placement="top" title="{% trans "Import Students" %}"><span class="glyphicon glyphicon-import" aria-hidden="true"></span></a>
```
janno42: please use `glyphicon-add`

evap/staff/tests.py
```diff
@@ -1240,3 +1231,61 @@ def test_user_with_same_username(self):
...
+    def test_suspicious_operation(self):
+        page = self.app.get("/staff/semester/1/course/1/importparticipants", user='user')
+
+        form = lastform(page)
+        form["excel_file"] = (os.path.join(settings.BASE_DIR, "staff/fixtures/valid_user_import.xls"),)
+
+        # Should throw SuspiciousOperation Exception.
+        reply = form.submit(name="operation", value="hackit", expect_errors=True)
+
+        self.assertEqual(reply.status_code, 400)
```
karyon: do you think it would make sense to do a plain post to the page instead of visiting the page and submitting the form?

evap/staff/views.py
janno42: please add a message to tell the user how many participants have been added to the course.

PR #709, opened 2016-01-17
Grade archiving

janno42:
This allows archiving grade documents by selecting whether or not grade documents of courses in a semester can be downloaded.

The PR also exchanges the course button texts with icons in the staff's semester view and moves the three options for adding courses into a new place. Courses now all have the same height in the table whether they show progress bars or not. Course names can get longer without creating line breaks.

*before:*
before

*now:*
after


karyon:
regarding the UI:

i'm ok with the x as delete button, but if we use it there we should use it everywhere where we have delete buttons, shouldn't we? it might clash with the other buttons not being icons though, but still...

the right part looks really dense now, could you add some horizontal space?

(Nothing to do with this PR, but what about vertically centering the buttons and the evaluation period? And what about spelling out M and F, there is enough space for that. Not sure how we would do that in german though. OR we put the M and F besides the progress bars, not below, would also help with the right part being quite packed)

What about calling the Add button something like "Create/Import"? might not look so cool but would much better describe what's in there.

karyon:
i'm wondering about `grades_active`, why not `grades_downloadable` besides it being longer? the model name gets a bit awkward, the best i could think of was `SemesterGradeDownloadActivation`. the benefit would be having no discrepancy between code and UI.

janno42:
`grades_downloadable` it is.

evap/grades/tests.py
```diff
@@ -181,3 +185,22 @@ def test_toggle_no_grades(self):
         self.get_submit_assert_302(toggle_url, "grade_publisher")
         course = Course.objects.get(id=course.id)
         self.assertFalse(course.gets_no_grade_documents)
+
+    def helper_grade_activation(self, semester, active):
+        activation, created = SemesterGradeDownloadActivation.objects.update_or_create(
+        semester=semester,
+        defaults={'is_active': active})
```
karyon: indent those please

evap/grades/urls.py
```diff
@@ -15,4 +15,6 @@
     url(r"^semester/(\d+)/course/(\d+)/edit/(\d+)$", edit_grades, name="edit_grades"),
     url(r"^semester/(\d+)/course/(\d+)/delete/(\d+)$", delete_grades, name="delete_grades"),
     url(r"^semester/(\d+)/course/(\d+)/togglenogrades$", toggle_no_grades, name="toggle_no_grades"),
+
+    url(r"^semester_grade_activation/(\d+)/(\w+)$", semester_grade_activation, name="semester_grade_activation"),
```
karyon: why not `semester/(\d)/grade_activation/(\w)`?

evap/grades/views.py
```diff
@@ -186,3 +191,14 @@ def delete_grades(request, semester_id, course_id, grade_document_id):
...
         )
         return render(request, "grades_delete.html", template_data)
+
+
+@staff_required
+def semester_grade_activation(request, semester_id, active):
+    active = active == 'on'
+
+    activation, created = SemesterGradeDownloadActivation.objects.update_or_create(
+        semester=Semester.objects.get(id=semester_id),
```
karyon: please use `semester = get_object_or_404(Semester, id=semester_id)` at the beginning of the method and then use `semester` here

janno42:
updated

PR #711, opened 2016-01-18
Force line breaks in list elements

### LINKED ISSUE #650: Force line breaks in comments
When posting a link into a comment, it might get too long for the page and breaks the layout.
linebreak

Line breaks should be forced.

maxifischer:
Fixes #650

### DIFF
--- a/evap/static/css/evap.css
+++ b/evap/static/css/evap.css
@@ -482,3 +482,6 @@ li.menu-indent {
 .settings-form select {
     width: 100%;
 }
+li {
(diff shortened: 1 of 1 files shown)


janno42:
this unfortunately always breaks hard at the end of the line (see below, e.g. "magn-a"). it should only break if necessary, when a word is too long to fit on a line.
do you want to have another look into the css?

capture

janno42:
fixed in e17fa26fdf2abdbbd9c349988f949a7b995bf8a1

PR #712, opened 2016-01-18
Read-only course details

### LINKED ISSUE #133: Course details should be accessible read-only for student representatives in all stages
At the moment, for courses which are in one of the states "evaluated", "reviewed" or "published", student representatives cannot see the details (like lecturer, participants, and so on) of this course.

It is true that the details of a course MUST NOT be edited in those states. However, it is desirable to have read-only access to this information e.g. to check the list of lecturers and to identify potential permission issues arising from it ("X is proxy of Y, but why can't X access Y's course results? Is Y really the main lecturer [='Dozent' flag]?)

kateyy:
Published and archived courses can now be opened in a read-only form. For reviewed courses, a warning message will be displayed on the edit form. (Closes #133)

### DIFF
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -17,6 +17,11 @@ import logging
 logger = logging.getLogger(__name__)
 
 
+def disable_all_fields(form):
+    for field in form.fields.values():
+        field.disabled = True
+
+
 class ImportForm(forms.Form, BootstrapMixin):
     vote_start_date = forms.DateField(label=_("First day of evaluation"), localize=True)
     vote_end_date = forms.DateField(label=_("Last day of evaluation"), localize=True)
@@ -107,6 +112,10 @@ class CourseForm(forms.ModelForm, BootstrapMixin):
         if self.instance.state in ['inEvaluation', 'evaluated', 'reviewed']:
             self.fields['vote_start_date'].disabled = True
 
+        if not self.instance.can_staff_edit:
+            # form is used as read-only course view
+            disable_all_fields(self)
+
     def clean(self):
         super().clean()
         vote_start_date = self.cleaned_data.get('vote_start_date')
@@ -150,6 +159,9 @@ class SingleResultForm(forms.ModelForm, BootstrapMixin):
 
         if self.instance.vote_start_date:
             self.fields['event_date'].initial = self.instance.vote_start_date
+
+        if not self.instance.can_staff_edit:
+            disable_all_fields(self)
 
         if self.instance.pk:
             self.fields['responsible'].initial = self.instance.responsible_contributor
@@ -222,6 +234,10 @@ class ContributionForm(forms.ModelForm, BootstrapMixin):
 
         self.fields['questionnaires'].queryset = Questionnaire.objects.filter(is_for_contributors=True).filter(
             Q(obsolete=False) | Q(contributions__course=self.course)).distinct()
+        
+        if not self.course.can_staff_edit:
+            # form is used as read-only course view
+            disable_all_fields(self)
 
     def save(self, *args, **kwargs):
         responsibility = self.cleaned_data['responsibility']
--- a/evap/staff/templates/staff_course_form.html
+++ b/evap/staff/templates/staff_course_form.html
@@ -35,11 +35,13 @@
                     {% if form.non_field_errors %}
                         <tr><td colspan=100>{{ form.non_field_errors }}</td></tr>
                     {% endif %}
-                    <tr class="contribution select2tr sortable">
+                    <tr class="contribution select2tr {% if course.can_staff_edit %} sortable {% endif %}">
                         {% for field in form.hidden_fields %}
                             {{ field }}
                         {% endfor %}
-                        <td class="movable"><span class="movable-icon glyphicon glyphicon-move"></span></td>
+                        <td class="movable">
+                            {% if course.can_staff_edit %} <span class="movable-icon glyphicon glyphicon-move" /> {% endif %}
+                        </td>
 
                         <td>{{ form.contributor.errors }} {{ form.contributor }}</td>
                         <td>{{ form.questionnaires.errors }} {{ form.questionnaires }}</td>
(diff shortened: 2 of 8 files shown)


evap/staff/templates/staff_course_form.html
```diff
@@ -64,10 +68,18 @@
                 {% if form.instance.state == "inEvaluation" %}
                     <div class="alert alert-warning">{% trans "You are editing a course, which is already in evaluation. Please note that only the students who did not evaluate yet will see your changes." %}</div>
                 {% endif %}
+                {% if form.instance.state == "evaluated" %}
+                    <div class="alert alert-warning">{% trans "You are editing a course, for which the evaluation already finished. Please be careful to not destroy any results." %}</div>
+                {% endif %}
+                {% if form.instance.state == "reviewed" %}
+                    <div class="alert alert-warning">{% trans "You are editing a course, which is already review. Please note that only the students who did not evaluate yet will see your changes." %}</div>
```
karyon: could you remove this and just do `if form.instance.state == "evaluated" or form.instance.state == "reviewed"` above?

evap/staff/templates/staff_course_form.html
```diff
@@ -78,5 +90,7 @@
 
 {% block additional_javascript %}
     {% include "evap_course_edit_js.html" %}
-    <script type="text/javascript"> preventAccidentalClosing = true; </script>
+    <script type="text/javascript">
+        preventAccidentalClosing =  {{ form.instance.can_staff_edit|yesno:"true,false" }};
+    </script>
```
karyon: could you do 
```
{% if form.instance.can_staff_edit %}
    <script type="text/javascript"> preventAccidentalClosing = true; </script>
{% endif %}
```

evap/staff/templates/staff_single_result_form.html
```diff
@@ -11,12 +11,16 @@
...
             </div>
         </div>
     </form>
 {% endblock %}
 
 {% block additional_javascript %}
-    <script type="text/javascript"> preventAccidentalClosing = true; </script>
+    <script type="text/javascript"> 
+        preventAccidentalClosing = {{ form.instance.can_staff_edit|yesno:"true,false" }};
+    </script>
```
karyon: same here

evap/staff/templates/staff_single_result_form.html
```diff
@@ -11,12 +11,16 @@
 
         <div class="form-group well">
             <div class="col-sm-offset-2 col-sm-6">
-                <input type="submit" value="{% trans "Save course" %}" class="btn btn-primary" onclick="preventAccidentalClosing = false;"/>
+                {% if form.instance.can_staff_edit %}
+                    <input type="submit" value="{% trans "Save course" %}" class="btn btn-primary" onclick="preventAccidentalClosing = false;"/>
+                {% endif %}
```
karyon: maybe this if should be around the whole `<div class="form-group well">`?

evap/staff/views.py
```diff
@@ -462,6 +456,10 @@ def helper_course_edit(request, semester, course):
     if form.is_valid() and formset.is_valid():
         if operation not in ('save', 'approve'):
             raise SuspiciousOperation("Invalid POST operation")
+
+        if (not course.can_staff_edit or course.is_archived) and operation == 'save':
+            raise SuspiciousOperation("Modifying this course is not allowed.")
```
karyon: i think this check should also apply for the `approve` operation, right? we don't want people to approve courses that they cannot edit.

evap/staff/views.py
```diff
@@ -494,6 +492,9 @@ def helper_single_result_edit(request, semester, course):
     form = SingleResultForm(request.POST or None, instance=course, initial=initial)
 
     if form.is_valid():
+        if (not course.can_staff_edit or course.is_archived):
```
karyon: remove parentheses please

karyon:
@kateyy do you know why i didn't review this two weeks ago? were there any major problems we talked about?

evap/staff/forms.py
```diff
@@ -183,6 +192,11 @@ def __init__(self, *args, **kwargs):
 
         self.fields['questionnaires'].queryset = Questionnaire.objects.filter(is_for_contributors=True).filter(
             Q(obsolete=False) | Q(contributions__course=self.course)).distinct()
+        
+        if not self.course.can_staff_edit:
+            # form is used as read-only course view
+            for field in self.fields.values():
+                field.disabled = True
```
karyon: before i forget it again, this should go into a method `def disable_all_fields(form):`

karyon:
@kateyy any chance you'll complete this e.g. tomorrow or say, in the next two weeks? otherwise i'd take this over.

evap/staff/templates/staff_course_form.html
kateyy: The warning messages are (intentionally) different, so I wouldn't change this.

janno42: `reviewed` comes after `evaluated`, the course is put into that state once all comments have been reviewed. the states are otherwise the same, so we should actually show the same message for both.

evap/staff/templates/staff_course_form.html
kateyy: done

evap/staff/templates/staff_single_result_form.html
kateyy: Yep. The same would be nice for staff_course_form.html, but I didn't find a solution for not checking all these ifs again around the div.

janno42: what do you mean by "all these ifs"?
what about this?:
```
{% if form.instance.can_staff_edit %}
    <div class="form-group well">
        <div class="col-sm-offset-2 col-sm-6">
            <input type="submit" value="{% trans "Save course" %}" class="btn btn-primary" onclick="preventAccidentalClosing = false;"/>
        </div>
    </div>
{% endif %}
```

janno42: oh, i see. nevermind, for staff_course_form we don't need it because the save button is always there.

evap/staff/views.py
kateyy: Done

evap/staff/tests.py
```diff
@@ -1089,7 +1088,6 @@ def test_raise_403(self):
         self.get_assert_403(semester_url + "import", "evap")
         self.get_assert_403(semester_url + "assign", "evap")
         self.get_assert_403(semester_url + "course/create", "evap")
-        self.get_assert_403(semester_url + "course/7/edit", "evap")
```
janno42: why did you remove this?

evap/staff/templates/staff_course_form.html
```diff
@@ -64,10 +68,18 @@
                 {% if form.instance.state == "inEvaluation" %}
```
janno42: please add an if around the well div. it probably will be quite long, but an empty div makes no sense.

evap/staff/views.py
```diff
@@ -462,6 +456,10 @@ def helper_course_edit(request, semester, course):
     if form.is_valid() and formset.is_valid():
         if operation not in ('save', 'approve'):
             raise SuspiciousOperation("Invalid POST operation")
+
+        if (not course.can_staff_edit or course.is_archived) and operation in ('save', 'approve'):
```
janno42: the operation must be either `save` or `approve`at this point, so `if not course.can_staff_edit or course.is_archived:` would be enough.

evap/staff/templates/staff_course_form.html
```diff
@@ -59,24 +61,33 @@
             </tbody>
         </table>
 
+        {% if state == "inEvaluation" or state == "evaluated" or state == "reviewed" or state == "new" or state == "prepared" or state == "editorApproved" or course.can_staff_edit %}
```
janno42: i'm sorry that i didn't completely think this through before. the states listed here are all states but `published` (`approved` is missing, but should actually be listed as well). so in all states in which the course can be edited we need this div. that means we can simply put a `{% if course.can_staff_edit %}` here.

The `Save` button can always be rendered (because we already checked for the `can_staff_edit`), the `Save and approve` button keeps the if with its three states.

kateyy:
saving courses in evaluated or reviewed now triggers TransitionNotAllowed at reopen_evaluation...

janno42:
`is_in_evaluation_period` is not a property and must be called `is_in_evaluation_period()`. this is currently wrong in two places in the code: `staff/views.py` line 491, and `evaluation/models.py` line 271.
when you fix that it should work.

kateyy:
okay fixed.

janno42:
when editing a published course, the disabled contribution formset does not render correctly (the selection fields with the contributors' names are too short). also, it is currently possible to add and delete forms in this formset.
please let me know if you don't want to look into that because it's too much trouble.

kateyy:
``class="movable"`` only sets ``width: 10px``, so it's fixed now, but it's maybe not the best style to use this class for non-movable elements?

janno42:
i'm okay with that class there. it's the element for moving stuff around even if it's not enabled ;)
do you want to look into the add and remove formset stuff?

kateyy:
yep

kateyy:
Is there a way to remove the "add another" without modifying the jquery.formset.js plugin?

janno42:
you could simply not call the `make_form_sortable` in `evap_course_edit_js.html`, then you get rid of all the formset stuff (you might not even need the movable td then anymore).
the only thing left then is the delete checkbox which you need to hide.

janno42:
it looks good now! thanks for fixing all the small thing i had to criticize.
could you now please squash at least some of the commits together (if you want they can all go into one)? then we're ready to merge.

kateyy:
Done =)

PR #718, opened 2016-01-31
Staff UI updates

janno42:
- show in how many days the evaluation for courses in state `approved` begins, if in 7 days or less
label

- show a message if in a state exist no courses.
nocourses

- show the number of participants also for single results
sr

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -263,6 +263,18 @@ class Course(models.Model, metaclass=LocalizeModelBase):
 
         return self.contributions.get(responsible=True).questionnaires.filter(name_en=Questionnaire.SINGLE_RESULT_QUESTIONNAIRE_NAME).exists()
 
+    def num_single_result_participants(self):
+        if not self.is_single_result:
+            return 0
+
+        num_participants = 0
+        ratinganswercounters = RatingAnswerCounter.objects.filter(
+            question=Question.objects.get(questionnaire=Questionnaire.get_single_result_questionnaire()),
+            contribution=self.contributions.get(contributor=self.responsible_contributor))
+        for counter in ratinganswercounters:
+            num_participants += counter.count
+        return num_participants
+
     @property
     def can_staff_edit(self):
         return not self.is_archived and self.state in ['new', 'prepared', 'editorApproved', 'approved', 'inEvaluation', 'evaluated', 'reviewed']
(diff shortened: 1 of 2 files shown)


evap/staff/templates/staff_semester_view.html
```diff
@@ -115,11 +115,15 @@ <h3 class="panel-title">{% trans "Actions" %}</h3>
...
-                                {% endfor %}
+                                {% if courses %}
+                                    {% for course in courses %}
+                                        <tr>
+                                            {% include_staff_semester_view_course semester state course disable_if_archived %}
+                                        </tr>
+                                    {% endfor %}
+                                {% else %}
+                                    <tr><td><i>{% trans "There are no courses in this state." %}</i><td></tr>
+                                {% endif %}
```
karyon: you can use https://docs.djangoproject.com/es/1.9/ref/templates/builtins/#for-empty here

evap/evaluation/models.py
```diff
@@ -263,6 +263,18 @@ def is_single_result(self):
...
+        if not self.is_single_result:
+            return 0
+
+        num_participants = 0
+        ratinganswercounters = RatingAnswerCounter.objects.filter(
+            question=Question.objects.get(questionnaire=Questionnaire.get_single_result_questionnaire()),
+            contribution=self.contributions.get(contributor=self.responsible_contributor))
+        for counter in ratinganswercounters:
+            num_participants += counter.count
+        return num_participants
```
karyon: this should be a private method and be called by num_voters and num_participants if self.is_single_result.

actually, i think we should put this directly into the _participant_count and _voter_count variable. unfortunately, those variables is currently used for determining whether a course is archived, which is used to determine whether a semester is archived. we could allow semesters with archived and non-archived courses at the same time, but a semester with only one course which is a single result would still be archived then.

i think this might require an is_archived flag on Semester and/or Course to to correctly. lets discuss that tomorrow.

janno42:
moved the first two changes into #719

karyon:
to not forget it again: semester gets an `archived` flag. semester.archive just sets that flag after checking the prerequisites. a course is archived if and only if its semester is archived. single results have their number of participants saved in `_num_participants` (and possibly `_num_voters`)

karyon:
superseded by #738

PR #719, opened 2016-02-01
Staff UI updates (Part 2)

janno42:


evap/staff/templates/staff_semester_view_course.html
```diff
@@ -36,6 +36,14 @@
 <td style="width: 10%">
     {{ course.vote_start_date|date:"SHORT_DATE_FORMAT" }}{% if course.is_single_result|is_false %} - {{ course.vote_end_date|date:"SHORT_DATE_FORMAT" }}{% endif %}
 
+    {% if course.state == 'approved' %}
+        <br>
+        {% if course.days_until_evaluation <= 1 %}
+            <span class="label label-danger">{% trans "starts tomorrow" %}</span>
```
karyon: the cronjob does not have to run on 00:00. so, after midnight, evap will tell you "starts tomorrow" even though that's not true.

karyon:
with the comment addressed +1.

evap/staff/templates/staff_semester_view_course.html
janno42: same goes for "ends today" and i don't really care :)

PR #721, opened 2016-02-01
Prevent invalid redirects after logging in. Fix #716

### LINKED ISSUE #716: Only internal redirects
The platform should only redirect to internal pages after logging in.

(handled in `evaluation/views.py index`)

simonkrogmann:
e. g. previously `http://localhost:8000/?next=/http://google.com/` would redirect to google after logging in.

### DIFF
--- a/evap/evaluation/views.py
+++ b/evap/evaluation/views.py
@@ -2,13 +2,14 @@ from django.contrib import messages
 from django.contrib.auth import login as auth_login
 from django.shortcuts import redirect, render
 from django.utils.translation import ugettext as _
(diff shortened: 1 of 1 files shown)


janno42:
unfortunately this doesn't help

simonkrogmann:
Now it should also handle `http://localhost:8000/?next=http://google.com/` correctly.

janno42:
looks good, thanks :)
if you rebase on the current master the tests shouldn't fail anymore.

karyon:
thanks!

PR #728, opened 2016-02-07
Participants warning

### LINKED ISSUE #726: Warning in courses with small number of participants
In courses with 5 or less participants a warning should be shown above the course's questionnaire:

_This course has only a small number of participants. Please remember that your comments will be visible for the responsible person and the contributors you're evaluating. If two or more people evaluate the course, the results of all voting questions will also be published._

janno42:
fix #726; show warning to students if course has small number of participants


karyon:
:koala:

PR #730, opened 2016-02-07
Use modals instead of confirmation pages

### LINKED ISSUE #723: Replace confirmation pages with modals
All confirmation pages (like e.g. `staff/semester/00/course/000/delete`) should be replaced with Bootstrap modals (http://getbootstrap.com/javascript/#modals).

- [x] user deletion
- [x] confirmation that final grades have been submitted
- [x] revoke confirmation that final grades have been submitted
- [x] grade document deletion
- [x] semester deletion
- [x] semester archiving
- [x] course deletion
- [x] questionnaire deletion
- [x] reward point redemption event deletion
- [x] lecturer confirmation (see #722)

no confirmation modals will be added for:
- course operations
- degree deletion
- FAQ section deletion
- FAQ question deletion

janno42:
fix #723
replaces confirmation pages with modals


evap/contributor/templates/contributor_course_form.html
```diff
@@ -92,6 +94,20 @@
     </form>
 {% endblock %}
 
+{% block modals %}
+    {{ block.super }}
+    {% trans "Approve course" as title %}
+    {% trans "If you approve this course you won't be able to make any additional changes afterwards, unless the student representatives reopen the course for changes.<br /><strong>Please approve the course when you're done editing it.</strong>" as question %}
```
karyon: To me this sounds like "please go back and finish editing"

janno42: Hm. It's meant as "please do not leave the course unapproved". Any suggestions?

evap/evaluation/templatetags/morefilters.py
```diff
@@ -88,3 +88,9 @@ def is_choice_field(field):
 @register.filter
 def is_user_editor_or_delegate(course, user):
     return  course.is_user_editor_or_delegate(user)
+
+
+# escaping ' so that names with ' can still be used as javascript arguments
+@register.filter
+def js_name(name):
+    return name.replace("'", "\\'")
```
karyon: could https://docs.djangoproject.com/es/1.9/ref/templates/builtins/#std:templatefilter-escape work here?

janno42: No, but `escapejs` might. Will try.

evap/grades/views.py
```diff
@@ -117,29 +117,19 @@ def upload_grades(request, semester_id, course_id):
 
 
 @grade_publisher_required
-def toggle_no_grades(request, semester_id, course_id):
-    semester = get_object_or_404(Semester, id=semester_id)
+def toggle_no_grades(request):
+    course_id = request.POST.get("course_id")
     course = get_object_or_404(Course, id=course_id)
```
karyon: why not keep the previous url scheme and drop the respective POST parameters?

evap/grades/views.py
```diff
@@ -117,29 +117,19 @@ def upload_grades(request, semester_id, course_id):
 
 
 @grade_publisher_required
-def toggle_no_grades(request, semester_id, course_id):
-    semester = get_object_or_404(Semester, id=semester_id)
+def toggle_no_grades(request):
```
karyon: i just found https://docs.djangoproject.com/en/1.9/topics/http/decorators/#django.views.decorators.http.require_POST, could you add that to all our views that are only for POST requests?

janno42: Okay

evap/grades/views.py
janno42: URLs can't be resolved in JavaScript. But only on client side I know which parameter to use. Or do you mean something else?

evap/grades/templates/grades_semester_view.html
```diff
@@ -97,6 +99,40 @@ <h3 class="panel-title">{% trans "Courses" %}</h3>
...
+        function confirm_laterupload_modal_action(data_id) {
+            $.ajax({
+                type: "POST",
+                url: "{% url "grades:toggle_no_grades" %}",
+                data: {"course_id": data_id},
+                success: function(){ location.reload(); },
+                error: function(){}
+            });
+        };
+    </script>
```
karyon: these two js functions appear identical to me.

evap/grades/templates/grades_semester_view.html
```diff
@@ -97,6 +99,40 @@ <h3 class="panel-title">{% trans "Courses" %}</h3>
...
+{% endblock %}
```
karyon: if i understood this correctly, most views fade out the item that is deleted, which might be enough visual feedback. here, the page is reloaded. is this intended? if yes, could you add a message in the view?

janno42: This was intended, because one ui element is removed and another one added. I will see to animate that as well.
We didn't have a message before and the changed entry should be enough confirmation for the user. But as said I will try to animate the change, making the reload unnecessary.

evap/rewards/views.py
```diff
@@ -114,18 +114,15 @@ def reward_point_redemption_event_edit(request, event_id):
...
-            event.delete()
-            return redirect('rewards:reward_point_redemption_events')
-        else:
-            return render(request, "rewards_reward_point_redemption_event_delete.html", dict(event=event))
+        event.delete()
+        return HttpResponse() # 200 OK
     else:
-        messages.warning(request, _("This event cannot be deleted because some users already redeemed points for it."))
-        return redirect('rewards:reward_point_redemption_events')
+        return HttpResponse(status=400) # 400 Bad Request
```
karyon: suggestion:
```
if not event.can_delete:
    raise SuspiciousOperation("Deleting redemption event not allowed")

event.delete()
return HttpResponse() # 200 OK
```
two things here, 1. i like the if-structure more this way and 2. raising a SuspiciousOperation documents that this could only happen if the user is forging requests, and it gets logged. somewhere.

janno42: Isn't this the same for all the other deletions as well? The buttons are always disabled in the UI when you're not allowed to use them so we could throw the SuspiciousOperation everywhere?

evap/staff/views.py
```diff
@@ -896,19 +876,15 @@ def user_edit(request, user_id):
...
-        if request.method == 'POST':
-            user.delete()
-            messages.success(request, _("Successfully deleted user."))
-            return redirect('staff:user_index')
-        else:
-            return render(request, "staff_user_delete.html", dict(user_to_delete=user))
+        user.delete()
+        return HttpResponse() # 200 OK
     else:
-        messages.warning(request, _("The user '%s' cannot be deleted, because he lectures courses.") % user.full_name)
```
karyon: there really was no good reason we had these error messages or was there? users could only see this by manually typing URLs right?

evap/staff/views.py
```diff
@@ -896,19 +876,15 @@ def user_edit(request, user_id):
...
-        return redirect('staff:user_index')
+        return HttpResponse(status=400) # 400 Bad Request
```
karyon: same comment about if-structure and suspiciousOperation applies everywhere here

evap/staff/views.py
janno42: Right

evap/rewards/views.py
karyon: well so there should be no user-facing change here as far as i know, as SuspiciousOperation also returns a 400. but basically, yes, i think.

evap/grades/views.py
karyon: what do you mean by resolving URLs in javascript?

what i mean is basically leave the urls.py and the URL creation in the templates unchanged, and in the view, leaving the view parameters as they are. so, continue doing

in the template:
```
{% url "grades:toggle_no_grades" semester.id course.id %}
```
and pass this whole url into the javascript function. the url tag can also do `as variable_name` if that helps.

urls.py
```
url(r"^semester/(\d+)/course/(\d+)/delete/(\d+)$", delete_grades, name="delete_grades"),
```

in the view
```
def toggle_no_grades(request, semester_id, course_id):
    course = get_object_or_404(Course, id=course_id)
    blahbla
```
just like it was before.

janno42: you mean instead of just passing the `data_id`? what's the advantage? i never really liked the current structure where you have to hand over unnecessary variables (like the semester_id, when you're actually working on a course which is already unique by its own id).

evap/grades/templates/grades_semester_view.html
janno42: right. will change.

evap/grades/views.py
karyon: well whether the semester should be part of it is another discussion.

what i meant is continue using the URL to contain the variable(s), just like we do for GET requests. in other words, why change this, what's the advantage?

janno42: hm. it's just passing the real data (one integer) around. not doing string joining and un-joining. it's shorter and to me it seems more logical to just do it that way. but if you insist...

karyon: if you show me the user list view loads slowly because of creating all those urls, that would be a reason to change this that i'd consider :)

evap/contributor/templates/contributor_course_form.html
karyon: why not `If you approve blah blah. Are you sure you're done editing it.?` to make clear what this modal want from you or `Are you sure you want to approve it?` to be in line with the approval button's text?

Or something like `Are you sure you want to approve this course? This will allow the student representatives to proceed with its evaluation, but you won't be able to make any further changes.`

about the reopening part, advertising it basically tells them "nevermind if you make mistakes, we can reopen them anytime", not sure if that is what you want.

janno42: reopening: yes, that's kind of what i want. otherwise some editors will fear to make irreversible changes when approving an therefore hesitate to do that. but maybe we can just ignore that.

evap/grades/templates/grades_semester_view.html
janno42: although possible i won't implement it. the reload will stay, otherwise every change in the page layout would need changes in the javascript as well and there are too many elements involved which have to be changed so there is a high chance of creating inconsistencies.

karyon: ok. what about the message? do you think the reload with some changed elements is enough feedback?

janno42: yes

janno42:
everything but the url resolving has been addressed

karyon:
ok, looks good. as i don't see a good reason to change the url scheme, please revert the respective changes. if there are performance issues, the code in the template gets ugly or there are other drawbacks, let's talk about it again.

janno42:
let's talk tomorrow.

karyon:
k sure

karyon:
i reverted the URL stuff for the grades app and came to the conclusion that it makes things slightly worse and the little bit more consistency is not worth it. i rebased the branch and added a small cleanup, feel free to merge.

Issue #704, opened 2016-01-10
Test management commands


karyon:
Because in three years, run_tasks will silently fail on the production system and nobody will notice.

* [x] **run_tasks** - shouldn't be too hard and is rather important
* [x] **anonymize** - might be a bit of work to cover it properly, but should be straightforward.
* [x] **refresh_results_cache** - should be easy
* [x] **dump_testdata** - don't know how not to overwrite the file during testing, but should be possible

the other commands are already tested or rather unsuitable for testing
* [x] **merge_users** - already has a test (#703) and is shown to be pretty broken.
* [x] **run** - don't know how to test this and there isn't really anything that could break. still, somehow running it to check that it doesn't crash right away on e.g. imports would be cool
* [x] **reload_testdata** - don't know whether it's possible at all to test that, i mean it drops the whole database...
* [ ] **import_ad** - we never used it and i don't know whether it's feasible to mock ldap

use `self.stdout.write` instead of `print` and `call_command("command_name", stdout=StringIO())` to avoid console output during tests. don't know what to do about calls to `input`.

schumannd:
calls to input can be mocked as usual:
```
from mock import patch

with patch('builtins.input') as mock_input:
    # do stuff
```

PR #731, opened 2016-02-08
started testing management commands (#704)

schumannd:
Cron job will have to be changed on production.

#704 

edit by @karyon: let's not forget to update the corresponding wiki pages with the new cronjobs

### DIFF
--- a/evap/evaluation/management/commands/refresh_results_cache.py
+++ b/evap/evaluation/management/commands/refresh_results_cache.py
@@ -5,17 +5,16 @@ from evap.evaluation.models import Course
 from evap.evaluation.tools import calculate_results
 
 
-
 class Command(BaseCommand):
     args = ''
     help = 'Clears the cache and pre-warms it with the results of all courses'
 
     def handle(self, *args, **options):
-        print("Clearing cache...")
+        self.stdout.write("Clearing cache...")
         cache.clear()
 
-        print("Calculating results for all courses...")
+        self.stdout.write("Calculating results for all courses...")
         for course in Course.objects.all():
             calculate_results(course)
 
-        print("Done.")
+        self.stdout.write("Done with updating cache.")
--- a/evap/evaluation/management/commands/send_reminders.py
+++ b/evap/evaluation/management/commands/send_reminders.py
@@ -11,16 +11,13 @@ logger = logging.getLogger(__name__)
 
 
 class Command(BaseCommand):
-    args = '<kind of jobs>'
-    help = 'Runs updates/tasks based on time events'
+    help = 'Sends email reminders X days before course evaluation ends.'
 
-    def update_courses(self):
-        """ Updates courses state, when evaluation time begins/ends."""
-        Course.update_courses()
-
-    def check_reminders(self):
-        logger.info("check_reminders called.")
+    def handle(self, *args, **options):
+        logger.info("send_reminders called.")
         check_dates = []
+
+        # Collect end-dates of courses whose participants need to be reminded today.
         for number_of_days in settings.REMIND_X_DAYS_AHEAD_OF_END_DATE:
             check_dates.append(datetime.date.today() + datetime.timedelta(days=number_of_days))
 
@@ -33,14 +30,11 @@ class Command(BaseCommand):
             for course in Course.objects.filter(participants=recipient, state='inEvaluation').exclude(voters=recipient):
                 due_courses[course] = (course.vote_end_date - datetime.date.today()).days
             first_due_in_days = min(due_courses.values())
-            # sort courses by number of days left for evaluation and bring them to following format: [(course, due_in_days), ...]
+
+            # Sort courses by number of days left for evaluation and bring them to following format:
+            # [(course, due_in_days), ...]
             due_courses = sorted(due_courses.items(), key=operator.itemgetter(1))
 
             EmailTemplate.send_reminder_to_user(recipient, first_due_in_days=first_due_in_days, due_courses=due_courses)
-        logger.info("check_reminders finished.")
-
-    def handle(self, *args, **options):
-        if len(args) > 0 and args[0] == 'daily':
-            self.check_reminders()
-        else:
-            self.update_courses()
+        logger.info("send_reminders finished.")
+        logger.info("sent reminders to %s people." % len(recipients))
--- a/evap/evaluation/management/commands/send_reminders.py
+++ b/evap/evaluation/management/commands/send_reminders.py
@@ -37,4 +37,4 @@ class Command(BaseCommand):
 
             EmailTemplate.send_reminder_to_user(recipient, first_due_in_days=first_due_in_days, due_courses=due_courses)
         logger.info("send_reminders finished.")
-        logger.info("sent reminders to %s people." % len(recipients))
+        logger.info("sent reminders to %s people.", len(recipients))
--- a/evap/evaluation/tests/test_commands.py
+++ b/evap/evaluation/tests/test_commands.py
@@ -1,4 +1,6 @@
 import datetime
+from django.utils.six import StringIO
+
 from django.core import management, mail
 from unittest.mock import patch
 from django.test import TestCase
@@ -6,6 +8,15 @@ from django.test.utils import override_settings
 from model_mommy import mommy
 
 from evap.evaluation.models import UserProfile, Course
+
+
+class TestRefreshResultsCacheCommand(TestCase):
+    def test_calls_calculate_results(self):
+        mommy.make(Course)
+        with patch('evap.evaluation.tools.calculate_results') as mock:
+            management.call_command('refresh_results_cache', stdout=StringIO())
+
+        self.assertEqual(mock.call_count, Course.objects.count())
 
 
 class TestUpdateCourseStatesCommand(TestCase):
--- a/evap/evaluation/tests/test_commands.py
+++ b/evap/evaluation/tests/test_commands.py
@@ -1,6 +1,11 @@
-from django.core import management
+import datetime
+from django.core import management, mail
 from unittest.mock import patch
-from django.test.testcases import TestCase
+from django.test import TestCase
+from django.test.utils import override_settings
+from model_mommy import mommy
+
+from evap.evaluation.models import UserProfile, Course
 
 
 class TestUpdateCourseStatesCommand(TestCase):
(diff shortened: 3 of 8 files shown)


evap/evaluation/management/commands/send_reminders.py
```diff
@@ -11,16 +11,17 @@
 
 
 class Command(BaseCommand):
-    args = '<kind of jobs>'
-    help = 'Runs updates/tasks based on time events'
+    help = 'Runs updates based on time events'
+    args = 'daily'
+
+    def handle(self, *args, **options):
+        assert (len(args) > 0 and args[0] == 'daily')
```
karyon: daily can be removed now, right? it was used to differentiate between the two tasks which isn't necessary now.

evap/evaluation/management/commands/send_reminders.py
```diff
@@ -11,16 +11,17 @@
 
 
 class Command(BaseCommand):
-    args = '<kind of jobs>'
-    help = 'Runs updates/tasks based on time events'
+    help = 'Runs updates based on time events'
```
karyon: this would need an update

evap/evaluation/management/commands/update_course_state.py
```diff
@@ -0,0 +1,14 @@
...
+
+from django.core.management.base import BaseCommand
+
+from evap.evaluation.models import Course
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Updates courses state, when evaluation time begins/ends.'
```
karyon: "Updates the state of all courses whose evaluation period starts or ends today."

evap/evaluation/management/commands/update_course_state.py
```diff
@@ -0,0 +1,14 @@
+import logging
+
+from django.core.management.base import BaseCommand
+
+from evap.evaluation.models import Course
+
+logger = logging.getLogger(__name__)
```
karyon: whats up with this unused stuff here. i thought you'd use pycharm, didn't you.

;)

evap/evaluation/tests/test_commands.py
```diff
@@ -0,0 +1,67 @@
...
+                vote_end_date=cls.today + datetime.timedelta(days=1),
+                participants=[cls.user_not_to_remind])
+
+    def test_handle(self):
+        with patch('evap.evaluation.models.EmailTemplate.send_reminder_to_user') as mock:
+            management.call_command('send_reminders', 'daily')
+
+        self.assertEquals(mock.call_count, 2)
+        mock.assert_any_call(self.user_to_remind, first_due_in_days=2, due_courses=[(self.course1, 2)])
+        mock.assert_any_call(self.user_to_remind_twice, first_due_in_days=0, due_courses=[(self.course2, 0), (self.course1, 2)])
```
karyon: is this test finished? i imagine there might be a lot more one could assert.

karyon: ...are there reasons to not split this into individual test methods? for each "type" of course one, and maybe one to ensure only one mail is sent in case multiple courses are affected or something like tha.

evap/evaluation/tests/test_models.py
```diff
@@ -0,0 +1,29 @@
...
+
+        with patch('evap.evaluation.models.EmailTemplate.send_evaluation_started_notifications') as mock:
+            Course.update_courses()
+
+        mock.assert_called_once_with([course])
+
+        course = Course.objects.get(pk=course.pk)
+        self.assertEqual(course.state, 'inEvaluation')
+
+    def test_fully_reviewed(self):
```
karyon: please make all the test names descriptive enough to not need a description, or add a comment.

karyon: also, for this test, please add a comment explaining why this course is fully reviewed, that's not immediately obvious.

evap/evaluation/management/commands/update_course_state.py
schumannd: PyCharm is not what it used to be...

schumannd:
I incorporated your comments. They are all (more or less ;) ) valid.
There is this really ugly send_publish_notifications() method that needs to be tested. but it's not center part of these commands.

evap/evaluation/tests/test_commands.py
```diff
@@ -0,0 +1,71 @@
...
+                state='inEvaluation',
+                vote_start_date=self.today - datetime.timedelta(days=1),
+                vote_end_date=self.today + datetime.timedelta(days=2),
+                participants=[user_no_remind],
+                voters=[user_no_remind])
+
+        with patch('evap.evaluation.models.EmailTemplate.send_reminder_to_user') as mock:
+            management.call_command('send_reminders', 'daily')
+
+        self.assertEqual(mock.call_count, 0)
```
karyon: please have a look at https://docs.djangoproject.com/en/1.9/topics/testing/tools/#email-services and especially the `len(mail.outbox)`part, i think asserting that no email have been sent at all makes more sense than just checking this one method. for the other tests, consider adding a check that no other mails have been sent.

evap/evaluation/management/commands/send_reminders.py
```diff
@@ -33,14 +30,10 @@ def check_reminders(self):
...
                 due_courses[course] = (course.vote_end_date - datetime.date.today()).days
             first_due_in_days = min(due_courses.values())
-            # sort courses by number of days left for evaluation and bring them to following format: [(course, due_in_days), ...]
+
+            # Sort courses by number of days left for evaluation and bring them to following format:
+            # [(course, due_in_days), ...]
             due_courses = sorted(due_courses.items(), key=operator.itemgetter(1))
 
             EmailTemplate.send_reminder_to_user(recipient, first_due_in_days=first_due_in_days, due_courses=due_courses)
         logger.info("check_reminders finished.")
```
karyon: send_reminders

evap/evaluation/tests/test_commands.py
```diff
@@ -0,0 +1,71 @@
...
+        mommy.make(
+                Course,
+                state='inEvaluation',
+                vote_start_date=self.today - datetime.timedelta(days=1),
+                vote_end_date=self.today + datetime.timedelta(days=2),
+                participants=[user_no_remind],
+                voters=[user_no_remind])
+
+        with patch('evap.evaluation.models.EmailTemplate.send_reminder_to_user') as mock:
+            management.call_command('send_reminders', 'daily')
```
karyon: remove `daily` here and above

evap/evaluation/tests/test_commands.py
schumannd: I disagree. I just want to test that send_reminder_to_user is not called. Whether this function actually sends an email should be tested as a model_test imo.

karyon: maybe i was unclear, i don't want to test whether  ̀send_reminder_to_user` sends emails or not, i want to ensure that this management command does not send any emails at all by `self.assertEqual(len(mail.outbox), 0)`

schumannd:
run_tasks (now split into send_reminders and update_course_state) and refresh_results_cache are now tested. The underlying "tools.py" should be tested as well. But I am not touching that monster now (wth is _calculate_results_impl? looks like a very hacky serializer)

evap/evaluation/management/commands/refresh_results_cache.py
```diff
@@ -11,11 +13,11 @@ class Command(BaseCommand):
...
+        logger.info("Clearing cache...")
         cache.clear()
 
-        print("Calculating results for all courses...")
+        logger.info("Calculating results for all courses...")
         for course in Course.objects.all():
             calculate_results(course)
 
-        print("Done.")
+        logger.info("Done with updating cache.")
```
karyon: please make sure this gets printed on the console as well when calling this command

evap/evaluation/tests/test_commands.py
```diff
@@ -0,0 +1,83 @@
...
+from django.test.utils import override_settings
+from model_mommy import mommy
+
+from evap.evaluation.models import UserProfile, Course
+from evap.evaluation.tests.test_utils import SilenceStd
+
+
+class TestRefreshResultsCacheCommand(TestCase):
+    def test_calls_calculate_results(self):
+        with SilenceStd():
```
karyon: please have a look at https://github.com/fsr-itse/EvaP/blob/master/evap/staff/tests.py#L1096 regarding silencing management commands

evap/evaluation/tests/test_commands.py
```diff
@@ -0,0 +1,83 @@
...
+
+from evap.evaluation.models import UserProfile, Course
+
+
+class TestRefreshResultsCacheCommand(TestCase):
+    def test_calls_calculate_results(self):
+        with patch('evap.evaluation.tools.calculate_results') as mock:
+            management.call_command('refresh_results_cache', stdout=StringIO())
+
+        self.assertEqual(mock.call_count, Course.objects.count())
```
karyon: ...could it be that `Course.objects.count())` is exactly zero here?

schumannd: Probably. I am not so sure about this test anyway. The Command basically consists of one for loop. Should we test this at all? I just added it for coverage. in case python 4 comes along :)

karyon: for that exact reason, or django API changes, i'd really like to have that test.

evap/evaluation/tests/test_tools.py
```diff
@@ -0,0 +1,18 @@
+from asyncio.test_utils import TestCase
```
karyon: man never thought i'd need to review imports as well...

evap/evaluation/tools.py
```diff
@@ -141,7 +144,7 @@ def calculate_results(course):
     if course.state != "published":
         return _calculate_results_impl(course)
 
-    cache_key = str.format('evap.staff.results.tools.calculate_results-{:d}', course.id)
+    cache_key = 'evap.staff.results.tools.calculate_results-%d' % course.id
```
karyon: please use .format, see the note here: https://docs.python.org/3.5/library/stdtypes.html#printf-style-string-formatting

karyon: i mean either leave it like it was or do `'foo'.format(...)` if that helps, but don't use the percent operator.

evap/evaluation/tools.py
```diff
@@ -74,6 +74,9 @@
 RatingResult = namedtuple('RatingResult', ('question', 'total_count', 'average', 'deviation', 'counts', 'warning'))
 TextResult = namedtuple('TextResult', ('question', 'answers'))
 
+CourseLists = namedtuple('CourseLists', ('grade_document_courses', 'evaluation_results_courses'))
+
+
```
karyon: was this intended to stay?

evap/evaluation/management/commands/send_reminders.py
```diff
@@ -33,14 +30,11 @@ def check_reminders(self):
...
             EmailTemplate.send_reminder_to_user(recipient, first_due_in_days=first_due_in_days, due_courses=due_courses)
-        logger.info("check_reminders finished.")
-
-    def handle(self, *args, **options):
-        if len(args) > 0 and args[0] == 'daily':
-            self.check_reminders()
-        else:
-            self.update_courses()
+        logger.info("send_reminders finished.")
+        logger.info("sent reminders to %s people." % len(recipients))
```
karyon: use `logger.info("sent reminders to %s people.", len(recipients))`

...i wonder why landscape hasn't complained about that...

evap/evaluation/tests/test_tools.py
```diff
@@ -0,0 +1,17 @@
...
+
+class TestCalculateResults(TestCase):
+    def test_caches_published_course(self):
+        course = mommy.make(Course, state='published')
+
+        self.assertIsNone(cache.get('evap.staff.results.tools.calculate_results-%d' % course.id))
+
+        calculate_results(course)
+
+        self.assertIsNotNone(cache.get('evap.staff.results.tools.calculate_results-%d' % course.id))
```
karyon: .format here as well

evap/evaluation/management/commands/send_reminders.py
karyon: ...or use .format here as well.

schumannd:
all done

evap/evaluation/tests/test_tools.py
```diff
@@ -0,0 +1,17 @@
...
+
+class TestCalculateResults(TestCase):
+    def test_caches_published_course(self):
+        course = mommy.make(Course, state='published')
+
+        self.assertIsNone(cache.get('evap.staff.results.tools.calculate_results-%d'.format(course.id)))
+
+        calculate_results(course)
+
+        self.assertIsNotNone(cache.get('evap.staff.results.tools.calculate_results-%d'.format(course.id)))
```
karyon: this does not work, you need to do {} inside the string.

karyon:
@janno42 could you review this as well?

janno42:
:+1: nice job

schumannd:
all done

karyon:
thanks! i updated https://github.com/fsr-itse/EvaP/wiki/Installation. @janno42 don't forget to update production.

PR #733, opened 2016-02-13
Delete navbar cache

### LINKED ISSUE #732: Update navbar cache
When adding or deleting semesters the cache must be refreshed so that the navbar shows the existing semesters.

Currently changes are not immediately visible in the navbar because cached entries are used for rendering.

janno42:
fix #732
delete navbar cache when creating or deleting semester


karyon:
:v:

Issue #715, opened 2016-01-26
Backup student form data in local storage


janno42:
Storing form data from partly filled questionnaires during the evaluation in the browser's cache can help to prevent data loss on connection errors.
In rare cases the connection to the server fails when submitting a questionnaire and all entered data is lost. The data should be stored locally so that it can be restored in such a case.

Possible tool: http://sisyphus-js.herokuapp.com/
See also #355

janno42:
replaced by #760

PR #734, opened 2016-02-15
[WIP] Save votes in local storage

TimNN:
Part of a potential solution to #715.

This uses the proposed sisyphus library.

While user information is currently preserved across reloads and navigation actions, user data is automatically cleared when the form is submitted.

Although this can be trivially disabled, this would lead to user data never being deleted from local storage (which is very undesirable I think).

Potential solutions:
* clear local storage on confirmation page load **Problem:** the user is redirect to index, making it non-trivial to pass the required data to the confirmation page
* submit the form using ajax and only clear the form upon success

I believe that the second solution would be preferable.

Currently only tested on Chrome on OS X.


janno42:
yes, please go for the ajax post request and clean on success.

evap/student/templates/student_vote.html
```diff
@@ -101,11 +104,29 @@
...
+            sisyphus.saveAllData()
         }
         function selectedAnswer(fieldName) {
             // remove error highlighting when an answer was selected
             $(".choice-error[name=" + fieldName + "]").removeClass("choice-error");
         }
         preventAccidentalClosing = true;
+
+        $(function() {
+            sisyphus = $("#student-vote-form").sisyphus({
```
janno42: can you replace the variable definition in line 93 above with `$("#student-vote-form").sisyphus({...});` and remove the rest?

evap/student/templates/student_vote.html
```diff
@@ -101,11 +104,29 @@
...
+                locationBased: true,
+                excludeFields: $("input[type=hidden]"), // exclude the csrf token
+                onRestore: function () {
+                    // If the form contains errors, is submitted, then changes are made and the page is reloaded,
+                    // some labels will have been set active by the server but the underlying checkbox is no longer
+                    // checked because it has been updated by sisyphus
```
janno42: these comments were not perfectly understandable to me at first. can you please just write a more general `update vote button states` before the two lines of code?

TimNN:
I currently don't have time to continue working on this, especially since using ajax for submission will require a non-trivial amount of work as all the error display would have to be handled in JS as well.

janno42:
@TimNN allright, then i'll take over. thanks so far :)

janno42:
replaced by #760

PR #736, opened 2016-02-20
User merge

### LINKED ISSUE #135: User merge: Allow user merging via the web interface
At the moment, duplicate user accounts can only be merged by using the `merge_user` management command on the command line. This is rather inconvenient and error-prone (e.g. you have to specify the correct user ids and the merge doesn't ask for confirmation, which may lead to merging the wrong accounts).

The merge feature should be available in the user management part of the web UI.

### LINKED ISSUE #705: merge_users discards votes and delegates
will probably be fixed by #135 either way but anyways.

we have users a and b. b voted for a course and has delegate.

if b gets merged into a, a does not appear in the voters list of the course and still has no delegate.

janno42:
fix #135, fix #705
adds views for merging users and merges all attributes

*Example:*
mergeusers


evap/staff/tools.py
```diff
@@ -20,3 +24,85 @@ def delete_navbar_cache():
...
+        course.save()
+
+    # update last_modified_user for grade documents
+    for grade_document in GradeDocument.objects.filter(last_modified_user=other_user):
+        grade_document.last_modified_user = main_user
+        grade_document.save()
+
+    # add all groups of other user
+    for group in merged_user['groups']:
+        main_user.groups.add(group)
```
karyon: for all these attributes, can't you just do `main_user.groups = merged_user['groups']`

...to be a bit more radical, would something like
```
for key, value in merged_user.items():
    setattr(main_user, key, value)
```
work? that would be pretty cool as its a lot less code and a lot less room for errors.

evap/staff/tools.py
```diff
@@ -20,3 +24,85 @@ def delete_navbar_cache():
...
+
+    # update last_modified_user for course
+    for course in Course.objects.filter(last_modified_user=other_user):
+        course.last_modified_user = main_user
+        course.save()
+
+    # update last_modified_user for grade documents
+    for grade_document in GradeDocument.objects.filter(last_modified_user=other_user):
+        grade_document.last_modified_user = main_user
+        grade_document.save()
```
karyon: please try something like 
`GradeDocument.objects.filter(last_modified_user=other_user).update(last_modified_user=main_user)` here and above

evap/staff/tests.py
```diff
@@ -1081,28 +1085,96 @@ def load_test_data(self):
...
+            first_name="Main",
+            last_name="",
+            email="main@test.com",
+            groups=[cls.group1],
+            delegates=[cls.user1, cls.user2],
+            represented_users=[cls.user3],
+            cc_users=[cls.user1],
+            ccing_users=[]
+        )
+        cls.mergeuser = mommy.make(UserProfile,
```
karyon: could you call these `main_user` and `other_user` to be in line with the implementation?

evap/evaluation/models.py
```diff
@@ -918,6 +918,16 @@ def generate_login_key(self):
...
         self.login_key_valid_until = datetime.date.today() + datetime.timedelta(settings.LOGIN_KEY_VALIDITY)
 
+    def get_sorted_contributions(self):
+        return Contribution.objects.filter(contributor=self).order_by('course__semester__created_at', 'course__name_de')
+
+    def get_sorted_courses_participating_in(self):
+        return Course.objects.filter(participants=self).order_by('semester__created_at', 'name_de')
+
+    def get_sorted_courses_voted_for(self):
+        return Course.objects.filter(voters=self).order_by('semester__created_at', 'name_de')
```
karyon: with the changes proposed in another comment:
`self.contributions, self.courses_participating_in, self.courses_voted_for` instead of `Course.objects.filter(...)`

evap/evaluation/management/commands/merge_users.py
```diff
@@ -1,20 +0,0 @@
...
-    args = '<primary user ID> <duplicate user ID>'
-    help = 'Merges duplicate user into primary user'
-
-    def handle(self, *args, **options):
-        try:
-            user1 = UserProfile.objects.get(pk=int(args[0]))
-            user2 = UserProfile.objects.get(pk=int(args[1]))
-
-            self.stdout.write("Merging user '{1}' into user '{0}'".format(user1, user2))
-            merge_model_objects(user1, user2)
```
karyon: any reason not to remove the actual merge script itself? i think it shouldn't be used anywhere else, or are you using it manually?

evap/staff/templates/staff_user_merge.html
```diff
@@ -0,0 +1,267 @@
...
+                        {{ user.full_name }}{% if forloop.last|is_false %},{% endif %}
+                    {% endfor %}
+                </td>
+                <td{% if merged_user.ccing_users %} class="success"{% endif %}>
+                    {% for user in merged_user.ccing_users %}
+                        {{ user }}{% if forloop.last|is_false %},{% endif %}
+                    {% endfor %}
+                </td>
+            </tr>
+            <tr{% if "contributions" in errors %} class="danger"{% endif %}>
```
karyon: these errors are explained nowhere in the UI. is that intended?

karyon: ...nvm :)

evap/staff/templates/staff_user_merge.html
```diff
@@ -0,0 +1,267 @@
...
+                        <ul>
+                            {% for course in courses.list %}
+                                <li>{{ course.name }}</li>
+                            {% endfor %}
+                        </ul>
+                    {% endfor %}
+                </td>
+                {% regroup merged_user.voted_for by semester as voting_list %}
+                <td{% if voting_list|length > 0 %} class="success"{% endif %}>
+                    {% for courses in voting_list %}
```
karyon: `courses` is misleading imo, since it's actually a group. to be in line with the example from the django docs this would be `semester`, or simply call it `group`, or something else.

evap/staff/templates/staff_user_merge.html
```diff
@@ -0,0 +1,267 @@
...
+                        <ul>
+                            {% for course in courses.list %}
+                                <li>{{ course.name }}</li>
+                            {% endfor %}
+                        </ul>
+                    {% endfor %}
+                </td>
+            </tr>
+        </tbody>
+    </table>
```
karyon: man this really is a huge amount of copy-pasta. but i couldn't see a good way to reduce it, let's talk about this tomorrow some more

karyon:
and let's not forget, we need a test that helps us not forgetting to add new attributes to the merge view.

evap/evaluation/management/commands/merge_users.py
janno42: i don't know what you mean. i removed the script here?

evap/staff/tools.py
janno42: the second suggestion probably wouldn't work for the contributions

janno42: works like a charm.

evap/staff/tools.py
janno42: check

evap/evaluation/management/commands/merge_users.py
karyon: i mean [this](https://github.com/fsr-itse/EvaP/blob/master/evap/evaluation/merge.py)

evap/staff/tools.py
karyon: nice!

evap/staff/tests.py
janno42: check

evap/evaluation/models.py

evap/evaluation/management/commands/merge_users.py
janno42: check

evap/staff/templates/staff_user_merge.html
janno42: check

janno42:
very refactorings, such nicer. wow.

evap/staff/tools.py
```diff
@@ -20,3 +24,55 @@ def delete_navbar_cache():
...
+        return (merged_user, errors)
+
+
+    # update last_modified_user for courses and grade documents
+    Course.objects.filter(last_modified_user=other_user).update(last_modified_user=main_user)
+    GradeDocument.objects.filter(last_modified_user=other_user).update(last_modified_user=main_user)
+
+    # email must not exist twice
+    other_user.email = ""
+    other_user.save()
```
karyon: you probably can just delete `other_user` here. if saving `main_user` fails, we still got the transaction.

janno42: no, not before the contributions have been changed.

karyon: right. could you put that into a comment?

janno42: k

evap/staff/tests.py
```diff
@@ -1081,28 +1085,96 @@ def load_test_data(self):
...
 
-        # see https://github.com/fsr-itse/EvaP/issues/705
-        #self.assertEqual(set(user1.delegates.all()), set([delegate1, delegate2]))
-        #self.assertEqual(set(course2.voters.all()), set([user1]))
+    def test_merge_users(self):
+        merged_user, errors = merge_users(self.main_user, self.other_user)  # merge should fail
+        self.assertSequenceEqual(errors, ['contributions', 'courses_participating_in'])
+
+        # assert that nothing has changed
+        self.assertEqual(self.main_user.username, "main_user")
```
karyon: you will need to do a `self.main_user.refresh_from_db()`, otherwise the attributes are guaranteed to not change :)

evap/staff/tests.py
```diff
@@ -1081,28 +1085,96 @@ def load_test_data(self):
...
+        self.assertSequenceEqual(self.course3.participants.all(), [self.other_user])
+        self.assertSequenceEqual(self.course3.voters.all(), [self.other_user])
+
+        # fix data
+        self.course1.participants = [self.main_user]
+        self.contribution2.delete()
+
+        merged_user, errors = merge_users(self.main_user, self.other_user)  # merge should succeed
+
+        self.assertEqual(self.main_user.username, "main_user")
```
karyon: same here, do a `refresh_from_db`

evap/staff/tests.py
```diff
@@ -1081,28 +1085,96 @@ def load_test_data(self):
...
+        self.assertSequenceEqual(self.course2.participants.all(), [self.main_user])
+        self.assertSequenceEqual(self.course2.voters.all(), [self.main_user])
+        self.assertSequenceEqual(self.course3.participants.all(), [self.other_user])
+        self.assertSequenceEqual(self.course3.voters.all(), [self.other_user])
+
+        # fix data
+        self.course1.participants = [self.main_user]
+        self.contribution2.delete()
+
+        merged_user, errors = merge_users(self.main_user, self.other_user)  # merge should succeed
```
karyon: assert errors is empty?

evap/staff/tests.py
janno42: why did the email change without reloading from db?

evap/staff/tests.py
janno42: yes

evap/staff/tests.py
janno42: hmkay

evap/staff/tools.py
```diff
@@ -20,3 +24,55 @@ def delete_navbar_cache():
...
+
+    # update values for main user
+    for key, value in merged_user.items():
+        setattr(main_user, key, value)
+    main_user.save()
+
+    # delete other user
+    other_user.delete()
+
+    return (merged_user, errors)
```
karyon: ...i'm not quite sure, shouldn't this return `main_user`? the return value when not using the preview is not used as far as i can see, but still...

also, please remove the parentheses here and above in the return statements

janno42: i don't think that returning a UserProfile object here and a dictionary above is a good idea.

evap/staff/views.py
```diff
@@ -914,6 +914,35 @@ def user_delete(request, user_id):
...
 @staff_required
+def user_merge_selection(request):
+    form = UserMergeSelectionForm(request.POST or None, request.FILES or None)
+
+    if form.is_valid():
+        main_user = form.cleaned_data['main_user']
+        other_user = form.cleaned_data['other_user']
+        return redirect('staff:user_merge', main_user.id, other_user.id)
+    else:
+        return render(request, "staff_user_merge_selection.html", dict(form=form))
```
karyon: wouldn't it be helpful to have an additional button in the user list or in a userprofile's edit page "merge..." or "merge into..."? you have these pages open anyways when searching for the duplicates. but you are the user, so it's your call i guess dr

janno42: yes, i thought about that. it saves entering one username. but it would preferably have both options you mentioned and i don't know whether it's worth it. i will find out when actually using it on production and add it later if necessary :)

evap/staff/tests.py
karyon: ...that's probably because you pass `main_user` into `merge_users`, and in there `main_user` is modified. so it might actually not be necessary... i'd add it anyways to make clear what should be tested and to be independent of `merge_users`'s implementation.

evap/staff/tools.py
karyon: okay, i thought of this method to generally return a UserProfile, but in the case of previewing, return a kind of duck-typed mock that can be used as if it was a UserProfile. i mean the point of the `merged_user` mock is that the caller doesn't know it actually didn't get a real UserProfile...

evap/staff/views.py
karyon: k

evap/staff/tools.py
janno42: ah. there is no use case so far and the object is not needed, so i would keep it that way for now.

janno42:
updated

karyon:
:last_quarter_moon_with_face:

PR #738, opened 2016-02-21
Archiving changes and participant count

karyon:
this does "show number of participants for single results", the remaining part of #718. To enable doing that in a nice way, it also does some changes to the archiving concept:

* semester now has an `is_archived` flag.
* `semester.archive` sets that flag after checking the prerequisites (and still writes to `_participant_count` and `_voter_count` for each course).
* a course is archived if and only if its semester is archived.
* single results have their number of participants saved in `_participant_count` and `_voter_count`.


karyon:
just ignore the first five commits, they're from that other pull request.

janno42:
looks good then

karyon:
Three commits added. the last one fixes another oversight, i forgot to apply the last migration to the test_data.json. Can you confirm there are only two single results in the test data?

janno42:
yes, one reviewed, one published.

PR #740, opened 2016-02-22
instant evaluation start check

janno42:
make sure that instant evaluation start is only possible if vote_end_date lies in the future


karyon:
:hamster:

PR #741, opened 2016-02-22
Introduce CourseType model

### LINKED ISSUE #315: Introduce Course Type model to enable translations and easier changes

janno42:
fix #315


evap/staff/importers.py
```diff
@@ -247,6 +248,14 @@ def check_course_data_correctness(self, semester):
             if not Degree.objects.filter(name_de=degree_name).exists():
                 self.errors.append(_("Error: The degree \"{}\" does not exist yet. Please manually create it first.").format(degree_name))
 
+        course_type_names = set()
+        for course_data in self.courses.values():
+            course_type_names.add(course_data.type_name)
```
karyon: you could do `course_type_names = set(course_data.type_name for course_data in self.courses.values())`

evap/staff/tests.py
```diff
@@ -842,13 +846,14 @@ class SingleResultFormTests(TestCase):
...
+        course_type = mommy.make(CourseType)
         form_data = {
             "name_de": "qwertz",
             "name_en": "qwertz",
-            "type": "a type",
+            "type": course_type.id,
             "degrees": ["1"],
             "event_date": "02/1/2014",
-            "responsible": responsible.pk,
+            "responsible": responsible.id,
```
karyon: i think `pk` is actually recommended, since it points to whatever field is used as primary key, which isn't necessarily called `id`

evap/evaluation/models.py
```diff
@@ -159,6 +159,26 @@ class Meta:
     def __str__(self):
         return self.name
 
+    def can_delete(self):
+        return not self.courses.all().exists()
```
karyon: on all other models this is called `can_staff_delete`, could you use that here as well?

evap/evaluation/models.py
```diff
@@ -159,6 +159,26 @@ class Meta:
...
+    name = Translate
+
+    class Meta:
+        ordering = ['name_de', ]
+
+    def __str__(self):
+        return self.name
+
+    def can_delete(self):
+        return not self.courses.all().exists()
```
karyon: same here

evap/staff/forms.py
```diff
@@ -48,6 +47,18 @@ class Meta:
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.fields["name_de"].widget = forms.TextInput(attrs={'class': 'form-control'})
+        self.fields["name_en"].widget = forms.TextInput(attrs={'class': 'form-control'})
+
+    class Meta:
+        model = CourseType
+        fields = "__all__"
+
```
karyon: i must say i'd feel more comfortable if we checked that for the instance of deleted forms `can_staff_delete` is actually True. in theory, if we added another criterion to `can_staff_delete` that wouldn't be catched by database constraints, we'd likely forget to add that check, to the form deletion.

if i remember correctly, you were able to detect deletion inside `clean()` of this class. what if you raised a SuspiciousOperation or PermissionDenied there instead of ValidationError?

...same would apply for Degree then.

janno42:
addressed all your comments

karyon:
looks fine apart from having zero test coverage. may i ask you to add tests? one that creates a new course type via the UI would probably be enough.

a similar test would be cool for the degrees but not required in this PR if you really don't want to.

janno42:
I really don't want to write a test for the course types either. Once I have one, the other one is almost the same but as I really don't want to it may take some more days.

janno42:
updated

karyon:
:mailbox:

Issue #170, opened 2013-12-03
Provide a simple feedback form


karyon:
some button at the side of the page which expands as you click on it and shows exactly one input field and a submit button. this could be sent as mail to the admins.

karyon:
implemented in #776

Issue #212, opened 2014-01-02
Evaluation feedback question in questionnaires


janno42:
The questionnaire of each course should provide a feedback question to write down remarks on the evaluation itself.
The information is meant to help the student representatives to update questionnaires and the evaluation process. The comments should only be visible to the student representatives.

schumannd:
If there is a feedback form accessible from anywhere as implemented in #753, would an extra section in the questionnaire still be necessary? Or would a simple sentence like "If you have any issues filling out the questionnaire, don't hesitate to ask for help via the feedback-button" or something similar suffice?

janno42:
see https://github.com/janno42/EvaP/commit/323f776c010b64a78487c639dba9593892fa4d87 for proposed changes to layout and functionality (save current user instead of email field and allow only logged in users to use the feedback form).

we do need some additional features as well:
- we need to save a time stamp for each entry
- staff users must be able to:
  - delete feedback entries
  - mark feedback entries as done
  - sort the table by timestamp, user, marked or not (we can use a datatable like for the results)

karyon:
i think it would still be good to have since the entry barrier is lower. if there's an input field readily available people might just enter something there, whereas the feedback button is more like when people have specific issues.

this feature requires an option per course to disable that question for certain courses. this option should not be visible for editors, just staff members.

karyon:
again to make this easier, just as in #170 we could simply send an email to the admins/staff, this way we don't need to store anything.

janno42:
the new feature in #1507 is planned to be set up for certain keywords to tell students that they should use the feedback form for general evap feedback instead of the evaluation questionnaires.

PR #753, opened 2016-03-10
Add Form to send Feedback/Questions from anywhere on the site

schumannd:
Implements #170 

A lot still open to debate, such as:
- General Design of: Button, Form, view_feedback view
- Feedback for successful message. (currently: changes feedback button text...)
- Location of Link to View Feedback Page
- Auto send mail function


schumannd:
Currently blocked by issue #754

janno42:
see https://github.com/janno42/EvaP/commit/323f776c010b64a78487c639dba9593892fa4d87 for proposed changes to layout and functionality (save current user instead of email field and allow only logged in users to use the feedback form).

we do need some additional features as well:
- we need to save a time stamp for each entry
- staff users must be able to:
  - delete feedback entries
  - mark feedback entries as done
  - sort the table by timestamp, user, marked or not (we can use a datatable like for the results)

janno42:
alternatively we could just send an email to the admins when the feedback form is submitted, which includes the username and the message. that would be way easier and we don't need to store anything in the database.

schumannd:
The points you listed should be easy to implement. So it is simply a matter of preference.

schumannd:
Implemented all features except for ordering of the table. Is this essential?

janno42:
yes, but it should be very easy, you just need to apply a datatable to the table element.

by the way: we need to send an email to the feedback address (not the admins) anyway, so that they know there is a new entry. i still didn't make up my mind whether we should only have the email - @karyon would most certainly tell us that this feature here is not necessary.

schumannd:
I developed this to support future implementation of #212. Is that issue not wanted then anymore? Should #212 also be implemented with an email only, not an extra model?

karyon:
indeed i'd just send an email with the feedback to some address and be done with it. having feedback organization functions inside evap sounds like overkill and feature creep to me. emails have a time stamp, can be deleted, marked as read, and sorted. the only difference is that they won't survive if the staff people change, but if that's so important, maybe you can get an email account from the admins or something.

PR #758, opened 2016-03-21
Login page redesign

janno42:
This changes the login page layout. External users often seem to have problems understanding how to login to the platform. This new layout will hopefully help them to find their way. The success message also explains, that the key was sent via email and the email address is cleared from the form to prevent multiple requests.

**Current layout:**
old
<hr>

**Proposed layout:**
new

### DIFF
--- a/evap/evaluation/templates/index.html
+++ b/evap/evaluation/templates/index.html
@@ -9,14 +9,15 @@
     {{ block.super }}
     <h1>{% trans "Welcome to the evaluation platform!" %}</h1>
 
-    <p>{% trans "You are currently not logged in." %}</p>
+    <i>{% trans "You are currently not logged in." %}</i>
+    <br /><br />
 
     <div class="row">
-        <div class="col-md-6">
+        <div class="col-md-5">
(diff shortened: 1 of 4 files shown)


karyon:
let's discuss this login key thing once more. this PR looks good but also makes the login key look even more absurd :)

an alternative: drop the login key. replace "request key" by "request login link" on the front page. the login link is a link that works only once and creates a session. to log in after the session has expired or on another machine/browser, the user needs to get a new link. all emails to external users contain such a link. only the newest link is valid for each user. if a user clicks on a link that has expired by timeout or by using it, the page says "sorry this link has expired, but we sent you a new one, please use that one."

so basically, it's like the login key, but it only works once, is used only via links and automatically sent in all the emails so people generally don't need to use the form on the front page.

i think the common case for external users is: they get an email that they need to review a course, evaluate a course or that results are available, and then want to access this stuff once in reaction to that email. having a link in those emails that just works would be super cool for that reason.

janno42:
replaced by #767

PR #760, opened 2016-03-29
Save votes in local storage

### LINKED ISSUE #715: Backup student form data in local storage

janno42:
fix #715. this supersedes #734.


evap/student/templates/student_vote.html
```diff
@@ -29,8 +29,9 @@
         </div>
     </small>
     {% endif %}
+    <div id="restore-info" class="bs-callout bs-callout-info" style="display: none">{% trans "Your answers have been restored for you. You can continue answering questions and then submit the questionnaire." %}</div>
```
karyon: does this help? i'd say this PR is a bugfix from a UX point of view and this says "oh yeah that bug really doesn't happen anymore"

evap/rewards/tests.py
```diff
@@ -132,5 +132,5 @@ def test_grant_reward_points(self):
         course.voters = []
         course.save()
         response = form.submit()
-        self.assertRedirects(response, reverse('student:index'))
+        self.assertEqual(response.status_code, 200)
```
karyon: why does this work? the view still redirects doesn't it?

evap/student/templates/student_index.html
```diff
@@ -120,3 +120,26 @@
...
+            for (var i = 0; i < localStorage.length; i++){
+                if (localStorage.key(i).substring(0,prefix.length) == prefix) {
+                    entries.push(localStorage.key(i));
+                }
+            }
+            // remove the items by key
+            for (var i = 0; i < entries.length; i++) {
+                localStorage.removeItem(entries[i]);
+            }
+        </script>
```
karyon: sysiphus has manuallyReleaseData which should do all this, only that it is probably not possible to directly use this as the form doesn't exist anymore. could you try sending the form via ajax and deleting everything on success? this [http://stackoverflow.com/questions/11534690/how-to-do-a-jquery-callback-after-form-submit](SO question) (the second answer looks better imo) suggests that it should be fairly simple to do.

janno42:
rework. now posting the form via ajax and only displaying a message if the server can't be reached.

evap/student/templates/student_vote.html
```diff
@@ -81,14 +81,56 @@
...
+                // update vote button states
+                $("div.vote-inputs input[type=radio]:not(:checked)").parent().removeClass("active");
+                $("div.vote-inputs input[type=radio]:checked").parent().addClass("active");
+            }
+        });
+        // save all data after loading the page
+        // (the data gets deleted every time the form is submitted, i.e. also when the form had errors and is displayed again)
+        sisyphus.saveAllData();
+
+        var frm = $('#student-vote-form');
```
karyon: come on give it an o please :)

evap/student/templates/student_vote.html
```diff
@@ -81,14 +81,56 @@
...
+            $.ajax({
+                type: frm.attr('method'),
+                url: frm.attr('action'),
+                data: frm.serialize(),
+                success: function (data) {
+                    sisyphus.manuallyReleaseData(); // delete the data from local storage
+                    // replace the html with the response data
+                    document.open();
+                    document.write(data);
+                    document.close();
```
karyon: ...and this does work? i.e. redirects to index if the vote succeeded?

janno42: just try it ;)

karyon:
like :+1: 

just add the o please

janno42:
done

PR #761, opened 2016-03-30
Move tests around

### LINKED ISSUE #707: Restructuring the tests to a more logical and newcomer-friendly structure
Having 90% of the tests reside in staff/tests.py does not seem very practical to me.

I would suggest a structure as outlined [here](https://realpython.com/blog/python/testing-in-django-part-1-best-practices-and-examples/)

Tests testing the Model "Course" would be placed in the following structure:

```
└── evaluation
    └── tests
        ├── __init__.py
        ├── test_utils.py
        ├── test_forms.py
        ├── test_models.py    <- they go here, e.g. class TestCourse(TestCase):
        └── test_views.py
```
Tests for views can now be easily located or placed at the appropriate place.

e.g. tests for the view under /staff/semester/create will be found in /staff/tests/test_views.py in the class TestSemesterCreateView

Utils, edge cases and other tests which are hard to classify could go into test_utils.py or a similar file. This should be the exception.

What do you think of this suggestion?

karyon:
Mostly fixes #707, i'd make a new issue for removing the coverage tests.

@schumannd please review this as well.


evap/staff/tests/test_views.py
```diff
@@ -60,16 +81,17 @@ def test_deletes_users(self):
 
 class TestSemesterExportView(ViewTest):
     url = '/staff/semester/1/export'
-    users = ['staff']
+    test_users = ['staff']
```
schumannd: how did this ever work? Are there other tests that have this bug?

karyon: it didn't. since "test_users" was not overridden and thus empty, the test simply did nothing. there are no other tests with this bug.

janno42:
:+1:

PR #762, opened 2016-03-30
Remove import_ad.py, we never used it

### LINKED ISSUE #704: Test management commands

karyon:
Remove import_ad.py since we never used it.

Fixes #704.


janno42:
:+1: get rid of it!

PR #763, opened 2016-03-31
Sort single results correctly

janno42:
- sort single results by total grade
- fix course type dropdown values in single results form


karyon:
:+1:

PR #765, opened 2016-04-01
Speedup user index

### LINKED ISSUE #755: Pagination for User view
The view takes several seconds to load. We might need a pagination for the future feedback view as well.

karyon:
kinda fixes #755 

based on #761, the last four commits are new.


janno42:
:+1:

PR #767, opened 2016-04-01
Update login procedure for external users

janno42:
This PR will change the way external users can login on the platform.
The proposed changes are open for discussion. What is missing, what should be changed?
- Login keys are only sent as login URLs for one-time use.
- When getting an editor review notification, an evaluation start notification, an evaluation reminder or a publishing notice, external users get a login URL included into the message.
- If the message is sent to other users in CC, the login URL is not included, but sent in a separate email right afterwards.
- The login page only has an email address field for external users where they can request a new login URL.


evap/evaluation/auth.py
```diff
@@ -41,16 +41,22 @@ def process_request(self, request):
...
+            user.login_key_valid_until = date.today() - timedelta(1)
+            user.save()
+        elif user:
+            # A user exists, but the login key is not valid anymore. Send the user a new one.
+            user.generate_login_key()
+            EmailTemplate.send_login_key_to_user(user)
+            messages.warning(request, _("The login URL was already used. We sent you a new one to your email address."))
         else:
-            messages.warning(request, _("Invalid login key"))
+            messages.warning(request, _("Invalid login URL. Please request a new one to your email address."))
```
karyon: request something to someone, is that correct? sounds weird to me.

evap/evaluation/models.py
```diff
@@ -1028,19 +1028,23 @@ def __render_string(cls, text, dictionary):
...
             responsible = course.responsible_contributor
-            for user in self.recipient_list_for_course(course, recipient_groups):
-                if user not in responsible.cc_users.all() and user not in responsible.delegates.all():
-                    user_course_map.setdefault(user, []).append(course)
+            all_recipients = self.recipient_list_for_course(course, recipient_groups)
+            cc_recipients = []
+            if responsible in all_recipients and cc == True:
+                cc_recipients.extend(responsible.cc_users.all())
+                cc_recipients.extend(responsible.delegates.all())
+            recipients = [recipient for recipient in all_recipients if recipient not in cc_recipients]
```
karyon: this code removes users from the recipient list if they'll get the email anyways because they are cc'ed in the email to a course's responsible person so they don't get the email twice, right? and this will break if they are in the recipient list for another course as well, right? and we don't care because that won't happen often enough in production, right?

if yes, could you add some comment along those lines to the code? i needed quite a while right now to figure that out.

evap/evaluation/models.py
```diff
@@ -1028,19 +1028,23 @@ def __render_string(cls, text, dictionary):
         return Template(text).render(Context(dictionary, autoescape=False))
 
     @classmethod
-    def send_to_users_in_courses(self, template, courses, recipient_groups):
+    def send_to_users_in_courses(self, template, courses, recipient_groups, cc):
```
karyon: can we find something better that `cc` as variable name? best i could think of is `use_cc` but that's not ideal either.

evap/evaluation/models.py
```diff
@@ -1083,9 +1098,10 @@ def send_reminder_to_user(cls, user, first_due_in_days, due_courses):
     def send_login_key_to_user(cls, user):
```
karyon: could you grep for `key` and replace it with url where appropriate?

evap/evaluation/models.py
```diff
@@ -1083,9 +1098,10 @@ def send_reminder_to_user(cls, user, first_due_in_days, due_courses):
...
         template = cls.objects.get(name=cls.LOGIN_KEY_CREATED)
         subject_params = {}
-        body_params = {'user': user}
+        body_params = {'user': user, 'login_url': settings.PAGE_URL + "?userkey=" + str(user.login_key)}
```
karyon: could you put that into a method?

evap/evaluation/auth.py
janno42: k

evap/evaluation/models.py
janno42: why does it break? if a user is a delegate of the responsible person of a course, this course will not be added to the list for the delegate, that's right. but other courses where this person is a recipient will be added to this users course list as usual - or am i wrong here?

evap/evaluation/models.py
janno42: for me `cc` is okay ;)

evap/evaluation/models.py

evap/evaluation/models.py
karyon: what i meant is that the person will get the email twice in that case, but that might actually be correct, thinking about it. still, please add some comment.

evap/evaluation/models.py
karyon: could you briefly describe what the `cc` variable does?

evap/evaluation/models.py
```diff
@@ -1028,19 +1028,23 @@ def __render_string(cls, text, dictionary):
...
+    def send_to_users_in_courses(self, template, courses, recipient_groups, cc):
         user_course_map = {}
         for course in courses:
             responsible = course.responsible_contributor
-            for user in self.recipient_list_for_course(course, recipient_groups):
-                if user not in responsible.cc_users.all() and user not in responsible.delegates.all():
-                    user_course_map.setdefault(user, []).append(course)
+            all_recipients = self.recipient_list_for_course(course, recipient_groups)
+            cc_recipients = []
+            if responsible in all_recipients and cc == True:
```
karyon: just `and cc:`

evap/evaluation/models.py

evap/evaluation/models.py

evap/evaluation/models.py
janno42: defining whether an email may be sent to other users in cc. if it's False, nobody will receive the email in cc. if it's True, delegates and cc users will receive a copy of the message if such users exist.

karyon: still in favor of `use_cc`.

evap/evaluation/models.py
```diff
@@ -1054,6 +1064,15 @@ def __send_to_user(cls, user, template, subject_params, body_params, cc):
         else:
             cc_addresses = []
 
+        send_separate_login_link = False
+        body_params['login_url'] = ""
+        if user.is_external:
```
karyon: maybe `user.needs_login_key` is more appropriate?

actually, same goes for all the email templates in the test data...

evap/evaluation/models.py

evap/evaluation/models.py

evap/evaluation/models.py
```diff
@@ -1028,32 +1034,45 @@ def __render_string(cls, text, dictionary):
...
             responsible = course.responsible_contributor
-            for user in self.recipient_list_for_course(course, recipient_groups):
-                if user not in responsible.cc_users.all() and user not in responsible.delegates.all():
-                    user_course_map.setdefault(user, []).append(course)
+            all_recipients = self.recipient_list_for_course(course, recipient_groups)
+            cc_recipients = []
+            if responsible in all_recipients and use_cc:
+                cc_recipients.extend(responsible.cc_users.all())
+                cc_recipients.extend(responsible.delegates.all())
+            recipients = [recipient for recipient in all_recipients if recipient not in cc_recipients]
```
karyon: explaining comment for all this please

evap/evaluation/tests/test_auth.py
```diff
@@ -0,0 +1,24 @@
...
+        user = mommy.make(UserProfile, email="extern@extern.com")
+        user.generate_login_key()
+        page = self.app.get(reverse("results:index") + "?userkey=%s" % user.login_key)
+        self.assertContains(page, 'Logged in as ' + user.full_name)
+        page = self.app.get(reverse("django-auth-logout")).follow()
+        self.assertContains(page, 'Not logged in')
+        page = self.app.get(reverse("results:index") + "?userkey=%s" % user.login_key).follow()
+        self.assertContains(page, 'The login URL was already used')
+        self.assertEqual(len(mail.outbox), 1)  # a new login key was sent
+        user = UserProfile.objects.get(pk=user.pk)
```
karyon: `user.refresh_from_db()`

evap/evaluation/models.py
janno42: oh... check.

evap/evaluation/tests/test_auth.py
janno42: check

evap/evaluation/tests/test_models.py
```diff
@@ -222,3 +223,49 @@ def test_archiving_doesnt_change_single_results_participant_count(self):
...
+    def test_login_url_when_nobody_in_cc(self):
+        # message is not sent to others in cc
+        EmailTemplate.send_to_users_in_courses(self.template, [self.course], "contributors", use_cc=True)
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertTrue("userkey" in mail.outbox[0].body)  # message does contain the login url
+
+    def test_login_url_when_use_cc_is_false(self):
+        # message is not sent to others in cc
+        self.user.delegates.add(self.other_user)
+        self.user.save()
```
karyon: the saves shouldn't be necessary, you're editing a m2m

janno42: check

evap/evaluation/tests/test_models.py
```diff
@@ -222,3 +223,49 @@ def test_archiving_doesnt_change_single_results_participant_count(self):
...
+        self.assertFalse("userkey" in mail.outbox[0].body)  # message does not contain the login url
+        self.assertTrue("userkey" in mail.outbox[1].body)  # separate email with login url was sent
+
+    def test_no_login_url_when_cc_users_in_cc(self):
+        self.user.cc_users.add(self.other_user)
+        self.user.save()
+        EmailTemplate.send_to_users_in_courses(self.template, [self.course], "contributors", use_cc=True)
+        self.assertEqual(len(mail.outbox), 2)
+        self.assertFalse("userkey" in mail.outbox[0].body)  # message does not contain the login url
+        self.assertTrue("userkey" in mail.outbox[1].body)  # separate email with login url was sent
```
karyon: maybe assert that the second email is sent to the right person and no-one is in cc?

evap/evaluation/tests/test_models.py
```diff
@@ -222,3 +223,49 @@ def test_archiving_doesnt_change_single_results_participant_count(self):
...
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertTrue("userkey" in mail.outbox[0].body)  # message does contain the login url
+
+    def test_login_url_when_use_cc_is_false(self):
+        # message is not sent to others in cc
+        self.user.delegates.add(self.other_user)
+        self.user.save()
+        EmailTemplate.send_to_users_in_courses(self.template, [self.course], "contributors", use_cc=False)
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertTrue("userkey" in mail.outbox[0].body)  # message does contain the login url
```
karyon: actually, is there any reason why this is called userkey and not loginkey?

evap/evaluation/tests/test_models.py
janno42: check

evap/evaluation/tests/test_models.py
janno42: not really...

karyon: ok pls rename it :)

janno42: check

evap/evaluation/models.py
```diff
@@ -1028,32 +1034,48 @@ def __render_string(cls, text, dictionary):
...
             return
 
-        if cc:
+        if use_cc:
             cc_users = set(user.delegates.all() | user.cc_users.all())
             cc_addresses = [p.email for p in cc_users if p.email]
         else:
             cc_addresses = []
 
+        send_separate_login_link = False
```
karyon: should be url perhaps

evap/settings.py
```diff
@@ -77,6 +77,9 @@
 # localhost is listed for easier development. Remove it in production environments.
 ALLOWED_HOSTS = ["localhost"]
 
+# The page URL that is used for links in email templates.
```
karyon: here as well

karyon:
:+1:

evap/evaluation/models.py

evap/settings.py
janno42: check

PR #770, opened 2016-04-02
Allow editing unsaved courses

### LINKED ISSUE #766: Fix create course form
When creating a new course the form is not rendered correctly because `course.can_staff_edit` is not defined in this situation.

janno42:
fix #766

### DIFF
--- a/evap/staff/templates/staff_course_form.html
+++ b/evap/staff/templates/staff_course_form.html
@@ -35,12 +35,12 @@
                     {% if form.non_field_errors %}
                         <tr><td colspan=100>{{ form.non_field_errors }}</td></tr>
                     {% endif %}
-                    <tr class="contribution select2tr {% if course.can_staff_edit %} sortable {% endif %}">
+                    <tr class="contribution select2tr {% if editable %} sortable {% endif %}">
                         {% for field in form.hidden_fields %}
                             {{ field }}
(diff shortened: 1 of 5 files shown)


evap/staff/templates/staff_course_form.html
```diff
@@ -85,7 +85,7 @@
```
karyon: what about the "You are editing a course which has not been approved by the editor yet." warning? i'd expect that to show up on the course create page.

karyon: oh and the "save and approve" button as well...

janno42: why would you want to have the message? that doesn't make sense to me.
i would never use the save and approve button in this situation because i would always want to check the preview first, but if you insist i can think about adding it.

karyon: nice misunderstanding :)

from looking at the code, i'd expect those to show up: `{% if state == "new" ...` but they wouldn't make any sense, right. the question is whether you made sure they don't show up.

janno42:
i didn't explicitly made sure that they don't show up, but they don't because `state` is not set

karyon:
actually state is set, but course is not set in the template...

karyon:
:bus:

PR #774, opened 2016-04-04
Remove lastform

### LINKED ISSUE #754: Find alternative to lastform() method used in many tests.
While trying to implement a feedback form in the base.html file I broke about 1/3 of all tests:

Some 35 test rely on using lastform() to identify the last form in the page. This breaks when any forms are added at the bottom of the page.
If I add a form to the top, about 15 other tests break. They rely on the form to be tested to be at a certain index.

A fix would be to name all forms and call them by their identifiers instead of relying on their location.

karyon:
fixes #754


janno42:
:mortar_board:

PR #776, opened 2016-04-05
Feedback form

schumannd:
Feature #170


evap/evaluation/tests/test_views.py
```diff
@@ -48,3 +52,12 @@ class TestLegalNoticeView(ViewTest):
...
     test_users = ['']
+
+
+class TestSendFeedbackView(WebTest):
+    csrf_checks = False
+
+    def test_sends_mail(self):
+        with patch('django.core.mail.EmailMessage.send') as mock:
+            self.app.post('/feedback/send', {'message': 'feedback message', 'sender_email': 'unique@mail.de'}, user='')
+        mock.assert_called_once_with(False)
```
karyon: put your hammer out of your hand for a sec and use [`mail.outbox`](https://docs.djangoproject.com/ja/1.9/topics/testing/tools/#email-services) :)

evap/static/js/feedback_form.js
```diff
@@ -0,0 +1,19 @@
...
+        url : "/feedback/send",
+        type : "POST",
+        data : { sender_email : email, message: message },
+
+        success : function(json) {
+            $('#feedback_spinner').hide();
+            $('#feedback_modal').modal('hide');
+            $('#feedback_button').html('Message Sent!');
+        }});
+});
```
karyon: newline here

evap/evaluation/views.py
```diff
@@ -94,3 +101,26 @@ def faq(request):
 
 def legal_notice(request):
     return render(request, "legal_notice.html", dict())
+
+
+@require_POST
+def feedback_send(request):
+    sender_email = request.POST.get("sender_email")
```
karyon: this is optional, right?

if the user doesn't provide one, django uses `DEFAULT_FROM_EMAIL`. maybe we should add a setting like `DEFAULT_FEEDBACK_EMAIL = 'feedback@localhost'` and use that if the user doesn't enter an address?

evap/templates/base.html
```diff
@@ -160,6 +160,36 @@
...
+                        <label for="sender-email" class="control-label">Email:</label>
+                        <input type="text" class="form-control" id="sender_email">
+                    </div>
+                    <div class="form-group">
+                        <label for="message-text" class="control-label">Question/Feedback:</label>
+                        <textarea class="form-control" id="message_text" style="resize:vertical;"></textarea>
+                    </div>
+                    <button type="button" class="btn btn-default" data-dismiss="modal">Cancel</button>
+                    <input type="submit" class="btn btn-primary">
+                    <img id="feedback_spinner" src="{% get_static_prefix %}images/spinner.gif" width="30px" height="30px" style="display: none;">
```
karyon: @janno42 should we use dashes consistently?

evap/templates/base.html
```diff
@@ -160,6 +160,36 @@
...
+                </form>
+            </div>
+        </div>
+    </div>
+</div>
```
karyon: the button probably shouldn't be shown if the user is not authenticated, otherwise someone will somewhen write a script and spam the crap out of us.

also, we could hide it for staff users, they won't use it.

karyon: the view would need a @login_required then.

evap/templates/base.html
```diff
@@ -160,6 +160,36 @@
...
+                    <div class="form-group">
+                        <label for="sender-email" class="control-label">Email:</label>
+                        <input type="text" class="form-control" id="sender_email">
+                    </div>
+                    <div class="form-group">
+                        <label for="message-text" class="control-label">Question/Feedback:</label>
+                        <textarea class="form-control" id="message_text" style="resize:vertical;"></textarea>
+                    </div>
+                    <button type="button" class="btn btn-default" data-dismiss="modal">Cancel</button>
+                    <input type="submit" class="btn btn-primary">
```
karyon: the button in the modal currently says "submit query" how about just a "submit"?

evap/templates/base.html
```diff
@@ -160,6 +160,36 @@
         {% endblock %}
     </div>
 </div>
+
+<button id="feedback_button" type="button" class="btn btn-warning" data-toggle="modal" data-target="#feedback_modal">?</button>
```
karyon: i'm not sure i like the `?`, @janno42 what do you think?

evap/static/css/evap.css
```diff
@@ -518,3 +518,12 @@ li.menu-indent {
     font-weight: bold;
     background-color: #F0F0F0;
 }
+
+/*feedback form + button*/
```
karyon: spaces after `/*` and before `*/` please

evap/evaluation/urls.py
```diff
@@ -9,4 +9,5 @@
     url(r"^$", index, name="index"),
     url(r"^faq$", faq, name="faq"),
     url(r"^legal_notice$", legal_notice, name="legal_notice"),
+    url(r"feedback/send$", feedback_send, name="send_feedback"),
```
karyon: `name="feedback_send"`

evap/templates/base.html
schumannd: I don't see submit query anywhere. What exactly do you mean?

evap/templates/base.html
janno42: I redesigned all the UI stuff here https://github.com/janno42/EvaP/commit/323f776c010b64a78487c639dba9593892fa4d87, can you please copy it from there?

evap/evaluation/views.py
janno42: i addressed this in https://github.com/janno42/EvaP/commit/323f776c010b64a78487c639dba9593892fa4d87. please use the email address of the currently logged in user. unauthenticated users don't need to see the feedback button.

evap/templates/base.html
janno42: it's not really important, but yes. let's start spreading some consistency :)

evap/templates/base.html
janno42: again, this was part of the changes in https://github.com/janno42/EvaP/commit/323f776c010b64a78487c639dba9593892fa4d87. only logged in users should get access to the feedback form. i don't mind having the button also for staff users.

schumannd:
Fixed all issues except for the "Submit Query" which I could not reproduce. Also the design decision to use "?" as a button was not yet addressed.

karyon:
there is the "submit query".

the button text's language seems to correspond to the browser language, not the language set in EvaP. maybe you'll need to set the text manually in the template?

evap/evaluation/tests/test_utils.py
```diff
@@ -1,5 +1,6 @@
 from django_webtest import WebTest as DjangoWebTest
 from model_mommy import mommy
+from webtest import AppError
```
karyon: ?

evap/evaluation/tests/test_views.py
```diff
@@ -48,3 +47,15 @@ class TestLegalNoticeView(ViewTest):
...
     url = '/faq'
     test_users = ['']
+
+
+class TestSendFeedbackView(ViewTest):
+    csrf_checks = False
+
+    @classmethod
+    def setUpTestData(cls):
+        mommy.make(UserProfile, username='evap', password=make_password('evap'))
```
karyon: hwy do you need to set the password?

evap/evaluation/tests/test_utils.py
schumannd: It was not imported, but used in the file. I don't know why it hasn't crashed before

evap/evaluation/views.py
```diff
@@ -94,3 +102,27 @@ def faq(request):
 
 def legal_notice(request):
     return render(request, "legal_notice.html", dict())
+
+
+@require_POST
+@login_required
+def feedback_send(request):
+    sender_email = request.POST.get("sender_email")
```
karyon: please use the email of the logged in user instead.

@janno42 actually, do we want this? it wouldn't be anonymous then anymore... maybe we should at least put a disclaimer on top. and from a technical POV, does owa (still) allow sending from arbitrary email addresses?

janno42: you could add a disabled field showing the email address. then everybody knows, that the adress will be included.

OWA does not allow that, the emails have to be sent from our own account (evaluierung). the email adress should be included in the message.

evap/evaluation/views.py
```diff
@@ -94,3 +102,27 @@ def faq(request):
...
+    if message:
+        mail = EmailMessage(
+            from_email=sender_email,
+            subject=subject,
+            body=message,
+            to=[settings.FEEDBACK_EMAIL])
+
+        try:
+            mail.send(False)
+            logger.info(('Sent email "{}"').format(mail.message()))
```
karyon: False is the default, you can remove it.

also, how does `str(mail.message())` look like? does it simply print the message body?

also, please make the text `Sent feedback email: "{}". and maybe with a newline?

evap/evaluation/views.py
```diff
@@ -94,3 +102,27 @@ def faq(request):
...
+
+        except Exception:
+            logger.exception('An exception occurred when sending the following feedback email:"{}":\n{}\n'.format(mail.message()))
```
karyon: two times `{}` in here, intended?

evap/static/css/evap.css
```diff
@@ -511,10 +511,19 @@ li.menu-indent {
 }
 
 .text-answer li,a {
-    word-break: break-word;
+    word-wrap: break-word;
```
karyon: this seems like a merge fail

evap/evaluation/views.py
schumannd: It looks something like this

Content-Type: text/b; charset="us-ascii"
MIME-Version: 1.0
Content-Transfer-Encoding: 7bit
Subject: subj
From: asdf@asdf.de

body stuff here

schumannd:
Done

evap/static/css/evap.css
karyon: this doesn't seem to be resolved

evap/templates/base.html
```diff
@@ -160,6 +160,37 @@
         {% endblock %}
     </div>
 </div>
+
+{% if user.is_authenticated and not user.is_staff %}
```
janno42: please show for staff as well. sorry that i didn't specify before. we don't have any features that are disabled for staff yet and we shouldn't start now. besides, staff users are also students and not all of them actively organize the evaluations, so they should also be able to use the feedback form.

janno42: the modal is currently always visible on the bottom of the page. please hide it until someone clicks the button ;)
also, please show the user's email address in the modal so that the user knows, that it will be connected to the message (use a disabled form field).

evap/static/js/feedback_form.js
```diff
@@ -0,0 +1,22 @@
...
+        data : { message: message },
+
+        success : function(json) {
+            $('#feedback-spinner').hide();
+            $('#feedback-modal').modal('hide');
+            var text = $('#feedback_button button').html();
+            $('#feedback-button button').html('Message sent.');
+            setTimeout(function(){
+                $('#feedback-button button').html(text);
+            }, 3000);
```
janno42: this does also not work, please check all the javascript for the modal and form again.

evap/evaluation/views.py
```diff
@@ -94,3 +102,27 @@ def faq(request):
...
+            subject=subject,
+            body=message,
+            to=[settings.FEEDBACK_EMAIL])
+
+        try:
+            mail.send()
+            logger.info(('Sent feedback email: \n{}\n').format(mail.message()))
+
+        except Exception:
+            logger.exception('An exception occurred when sending the following feedback email:\n{}\n'.format(mail.message()))
```
karyon: please have a look at the landscape issues, they're right i think

evap/evaluation/views.py
```diff
@@ -94,3 +102,27 @@ def faq(request):
...
+@require_POST
+@login_required
+def feedback_send(request):
+    sender_email = request.user.email
+    message = request.POST.get("message")
+    subject = "Feedback from {}".format(sender_email)
+
+    if message:
+        mail = EmailMessage(
+            from_email=sender_email,
```
janno42: please don't set the `from` - it probably won't work on the mailserver.

evap/static/css/evap.css
schumannd: What do you mean resolved? break-word is a property of word-wrap and not word-break. The way it is right now it shouldn't work:

http://www.w3schools.com/cssref/css3_pr_word-wrap.asp

http://www.w3schools.com/cssref/css3_pr_word-break.asp

evap/evaluation/views.py
schumannd: They suggest  %  string formatting? But I don't understand why.

schumannd:
done

evap/static/css/evap.css
karyon: well, i commented "this seems like a merge fail". you didn't respond and when you said "done" i assumed you missed my comment. if it's a intended fix it's fine of course

evap/evaluation/views.py
karyon: lmgtfy: https://www.simonmweber.com/2014/11/24/python-logging-traps.html

evap/evaluation/views.py
```diff
@@ -94,3 +102,26 @@ def faq(request):
...
+
+    if message:
+        mail = EmailMessage(
+            subject=subject,
+            body=message,
+            to=[settings.FEEDBACK_EMAIL])
+
+        try:
+            mail.send()
+            logger.info('Sent feedback email: \n%s\n' % mail.message())
```
karyon: the lint suggested to pass the formatting parameters to logger.info, not interpolate the string via %.

karyon:
alright after the string formatting is changed i say +1, @janno42 please +1 the UI.

evap/static/css/evap.css
schumannd: Ok, i thought that comment was referring to the possible origin of the original error.

schumannd:
changed the string formatting

karyon:
* i think the input field is pretty small initially
* the button looks cool but i suspect that no-one will notice it. maybe explicitly write "feedback" on it?

let's wait for @janno42's comments

janno42:
almost done ;)
please add a label "Message" above the message field and as @karyon said, please increase its size (maybe 5 lines).

PR #782, opened 2016-04-25
Fix single result ordering

### LINKED ISSUE #743: single results not sortable by result
On the results site of a semester are the single results not sortable by the result. When clicking on the results table header the results get sorted by something, but not the result.

karyon:
fixes #743


evap/static/dataTables/de.json
```diff
@@ -5,6 +5,7 @@
 	"sInfoFiltered": 	"(gefiltert von _MAX_ Einträgen)",
 	"sInfoPostFix":  	"",
 	"sInfoThousands":  	".",
+	"sDecimal":         ",",
```
janno42: indentation

janno42:
:hamster:

PR #785, opened 2016-05-01
Raw data export

### LINKED ISSUE #681: Export raw data for staff users
Each semester should be exportable as raw data for staff users. One line per course, each line should hold the following:

- course name
- course degree
- course type
- number of voters
- number of participants
- number of comments
- average grade

janno42:
Fix #681 

This allows exporting raw course data per semester as well as a participation overview in CSV format. Both are helpful for creating detailed statistics about a semester.


evap/staff/views.py
```diff
@@ -357,6 +361,60 @@ def semester_export(request, semester_id):
 
 
 @staff_required
+def semester_raw_export(request, semester_id):
+    semester = get_object_or_404(Semester, id=semester_id)
+
+    filename = "Evaluation-%s-%s_raw.csv" % (semester.name, get_language())
```
karyon: could you use the `{}` formatting style?

evap/staff/views.py
```diff
@@ -357,6 +361,60 @@ def semester_export(request, semester_id):
...
+    response = HttpResponse(content_type="text/csv")
+    response["Content-Disposition"] = "attachment; filename=\"%s\"" % filename
+
+    writer = csv.writer(response, delimiter=";")
+    writer.writerow([_('Name'), _('Degrees'), _('Type'), _('Single result'), _('State'), _('#Voters'),
+        _('#Participants'), _('#Comments'), _('Average degree')])
+    for course in semester.course_set.all():
+        degrees = ",".join([degree.name for degree in course.degrees.all()])
```
karyon: maybe `", "`(with a space?)

evap/staff/views.py
```diff
@@ -357,6 +361,60 @@ def semester_export(request, semester_id):
...
+        course.avg_grade, course.avg_deviation = calculate_average_grades_and_deviation(course)
+        if course.state in ['evaluated', 'reviewed', 'published'] and course.avg_grade is not None:
+            course.avg_grade = ('%.1f' % float(course.avg_grade)).replace('.',',')
```
karyon: you want to localize with this, try using the floatformat template tag
```
from django.template.defaultfilters import floatformat
floatformat(1.22, 1)
```

evap/staff/views.py
```diff
@@ -357,6 +361,60 @@ def semester_export(request, semester_id):
...
+        else:
+            course.avg_grade = ""
+        writer.writerow([course.name, degrees, course.type.name, course.is_single_result(), course.state,
+            course.num_voters, course.num_participants, len(course.textanswer_set.all()), course.avg_grade])
+
+    return response
```
karyon: how about putting all this in a template? it's probably overkill, i just want to make sure you have thought about this

evap/staff/views.py
```diff
@@ -357,6 +361,60 @@ def semester_export(request, semester_id):
...
+    writer = csv.writer(response, delimiter=";")
+    writer.writerow([_('Name'), _('Degrees'), _('Type'), _('Single result'), _('State'), _('#Voters'),
+        _('#Participants'), _('#Comments'), _('Average degree')])
+    for course in semester.course_set.all():
+        degrees = ",".join([degree.name for degree in course.degrees.all()])
+        course.avg_grade, course.avg_deviation = calculate_average_grades_and_deviation(course)
+        if course.state in ['evaluated', 'reviewed', 'published'] and course.avg_grade is not None:
+            course.avg_grade = ('%.1f' % float(course.avg_grade)).replace('.',',')
+        else:
+            course.avg_grade = ""
```
karyon: please put the formatted avg_grade into a variable and don't write into the object

evap/staff/views.py
```diff
@@ -357,6 +361,60 @@ def semester_export(request, semester_id):
...
+        writer.writerow([course.name, degrees, course.type.name, course.is_single_result(), course.state,
+            course.num_voters, course.num_participants, len(course.textanswer_set.all()), course.avg_grade])
```
karyon: .count instead of len

evap/staff/views.py
```diff
@@ -357,6 +361,60 @@ def semester_export(request, semester_id):
...
+
+    return response
+
+
+@staff_required
+def semester_participation_export(request, semester_id):
+    semester = get_object_or_404(Semester, id=semester_id)
+    courses = Course.objects.filter(semester=semester)
```
karyon: semester.courses.all()

evap/staff/views.py
```diff
@@ -357,6 +361,60 @@ def semester_export(request, semester_id):
...
+    participants = set()
+    for course in courses:
+        for participant in course.participants.all():
+            if can_user_use_reward_points(participant):
+                participants.add(participant)
+    participants = sorted(participants, key=attrgetter('username'))
```
karyon: could you try replacing all this with a db query? something like 
```
semester.courses.participants.order_by("username")  # .distinct() if necessary
participants = [p for p in participants if can_user_use_reward_points(p)]
```

evap/staff/views.py
```diff
@@ -357,6 +361,60 @@ def semester_export(request, semester_id):
...
+    semester = get_object_or_404(Semester, id=semester_id)
+
+    filename = "Evaluation-%s-%s_raw.csv" % (semester.name, get_language())
+    response = HttpResponse(content_type="text/csv")
+    response["Content-Disposition"] = "attachment; filename=\"%s\"" % filename
+
+    writer = csv.writer(response, delimiter=";")
+    writer.writerow([_('Name'), _('Degrees'), _('Type'), _('Single result'), _('State'), _('#Voters'),
+        _('#Participants'), _('#Comments'), _('Average degree')])
+    for course in semester.course_set.all():
```
karyon: courses are by default sorted by name_de, is that okay here?

speaking of which, that default actually isn't helpful...

evap/staff/views.py
```diff
@@ -357,6 +361,60 @@ def semester_export(request, semester_id):
...
+    response["Content-Disposition"] = "attachment; filename=\"%s\"" % filename
+
+    writer = csv.writer(response, delimiter=";")
+    writer.writerow([_('Username'), _('#Required courses voted for'), _('#Required courses'),
+        _('#Optional courses voted for'), _('#Optional courses'), _('Earned reward points')])
+    for participant in participants:
+        number_of_required_courses = Course.objects.filter(semester=semester, participants=participant, is_required_for_reward=True).count()
+        number_of_required_courses_voted_for = Course.objects.filter(semester=semester, voters=participant, is_required_for_reward=True).count()
+        number_of_optional_courses = Course.objects.filter(semester=semester, participants=participant, is_required_for_reward=False).count()
+        number_of_optional_courses_voted_for = Course.objects.filter(semester=semester, voters=participant, is_required_for_reward=False).count()
```
karyon: `Semester.courses.filter` instead of `Course.objects.filter(semester=semester` would shorten those lines a bit

evap/staff/views.py
```diff
@@ -357,6 +361,60 @@ def semester_export(request, semester_id):
...
+            if can_user_use_reward_points(participant):
+                participants.add(participant)
+    participants = sorted(participants, key=attrgetter('username'))
+
+    filename = "Evaluation-%s-%s_participation.csv" % (semester.name, get_language())
+    response = HttpResponse(content_type="text/csv")
+    response["Content-Disposition"] = "attachment; filename=\"%s\"" % filename
+
+    writer = csv.writer(response, delimiter=";")
+    writer.writerow([_('Username'), _('#Required courses voted for'), _('#Required courses'),
```
karyon: i must say i don't like that we have the username flying around everywhere in the rewards part where it's not actually necessary. i propose replacing that with the user ID where possible. that effectively hides the username and prevents ctrl-f'ing it but still makes it possible to debug stuff.

karyon:
oh and please add tests :)

evap/staff/views.py
```diff
@@ -347,16 +351,64 @@ def semester_export(request, semester_id):
...
+        writer.writerow([course.name, degrees, course.type.name, course.is_single_result(), course.state,
+            course.num_voters, course.num_participants, course.textanswer_set.count(), avg_grade])
+
+    return response
+
+
+@staff_required
+def semester_participation_export(request, semester_id):
+    semester = get_object_or_404(Semester, id=semester_id)
+    participants = [participant for participant in UserProfile.objects.filter(courses_participating_in__semester=semester).order_by("username")]
```
karyon: you skipped the `can_user_use_reward_points`, intentionally? if you add it, please put the list comprehension on a second line

karyon:
could you add an explanation why that page got removed to the commit message? also, check out the landscape issues.

evap/staff/views.py
janno42: oh, no. but now i want to keep it that way. added a column stating whether the value is true or false.

karyon: ok then you can remove the `[participant for participant in`, it does nothing.

janno42:
added tests. i don't like them. let the hate begin.

evap/staff/tests/test_views.py
```diff
@@ -247,3 +247,57 @@ def test_no_second_update(self):
...
+        cls.course2 = mommy.make(Course, type=cls.course_type, semester=cls.semester, participants=[cls.student_user],
+            name_de="Veranstaltung 2")
+        mommy.make(Contribution, course=cls.course1, responsible=True, can_edit=True, comment_visibility=Contribution.ALL_COMMENTS)
+        mommy.make(Contribution, course=cls.course2, responsible=True, can_edit=True, comment_visibility=Contribution.ALL_COMMENTS)
+
+    def test_view_downloads_csv_file(self):
+        response = self.app.get(self.url, user='staff')
+        degrees1 = ", ".join([degree.name for degree in self.course1.degrees.all()])
+        degrees2 = ", ".join([degree.name for degree in self.course2.degrees.all()])
+        expected_content = "Name;Degrees;Type;Single result;State;#Voters;#Participants;#Comments;Average degree\r\n" + \
```
karyon: does the exporter really put `\r`s in there?

evap/staff/tests/test_views.py
```diff
@@ -247,3 +247,57 @@ def test_no_second_update(self):
...
+            self.course1.name_en + ";" + degrees1 + ";" + self.course_type.name_en + ";False;" + self.course1.state + ";" + \
+            str(self.course1.num_voters) + ";" + str(self.course1.num_participants) + ";" + str(self.course1.textanswer_set.count()) + ";\r\n" + \
+            self.course2.name_en + ";" + degrees2 + ";" + self.course_type.name_en + ";False;" + self.course2.state + ";" + \
+            str(self.course2.num_voters) + ";" + str(self.course2.num_participants) + ";" + str(self.course2.textanswer_set.count()) + ";\r\n"
+        self.assertEqual(response.content, expected_content.encode("utf-8"))
```
karyon: why is the `encode` necessary?

evap/staff/tests/test_views.py
```diff
@@ -247,3 +247,57 @@ def test_no_second_update(self):
...
+
+    def test_view_downloads_csv_file(self):
+        response = self.app.get(self.url, user='staff')
+        degrees1 = ", ".join([degree.name for degree in self.course1.degrees.all()])
+        degrees2 = ", ".join([degree.name for degree in self.course2.degrees.all()])
+        expected_content = "Name;Degrees;Type;Single result;State;#Voters;#Participants;#Comments;Average degree\r\n" + \
+            self.course1.name_en + ";" + degrees1 + ";" + self.course_type.name_en + ";False;" + self.course1.state + ";" + \
+            str(self.course1.num_voters) + ";" + str(self.course1.num_participants) + ";" + str(self.course1.textanswer_set.count()) + ";\r\n" + \
+            self.course2.name_en + ";" + degrees2 + ";" + self.course_type.name_en + ";False;" + self.course2.state + ";" + \
+            str(self.course2.num_voters) + ";" + str(self.course2.num_participants) + ";" + str(self.course2.textanswer_set.count()) + ";\r\n"
```
karyon: how about just putting the expected string here, so without accessing all the attributes, just the result? makes it less ugly and more strict

karyon: also you can put the string in parentheses instead of using the backslash, that looks better imo.

evap/staff/tests/test_views.py
janno42: in the file it's a byte string. after encoding this also is a byte string.

evap/staff/tests/test_views.py
janno42: well it works, doesn't it? yes, the file had `\r`s in my tests.

evap/staff/tests/test_views.py
janno42: check

karyon:
:fuelpump:

PR #789, opened 2016-05-07
Landscape fixes

### LINKED ISSUE #780: Landscape sweep and style consistency
* fix landscape issues as far as possible
* prevent more "some lints" commits by @schumannd, possibly by tuning landscape to also complain about those minor issues, or provide config files for pylint/pep8/whatnot

karyon:
Lots of style and smell fixes, some minor functional changes and refactorings. Adds a pylintrc file, tweaks landscape.

two things to consider:
* i decided against `logger.info("foo %s", bar)` in favor of `logger.info("foo {}".format(bar))` so we can use the newer `{}` formatting everywhere, although the former has some (rather theoretical) advantages (see [here](https://www.simonmweber.com/2014/11/24/python-logging-traps.html). on the other hand it's kinda pointless because we use the `%s` formatting still in about half of all places.
* i disabled bad-continuation because we consistently do it differently that pep8.

The goal is not to come down to zero issues, that would require quite a few comment in the code to disable lints that are not fixable in those particular cases. I disabled some lints that have a particularly high number of false-positives. For all the remaining issues i couldn't find an appropriate fix and I didn't want to disable the respective lint completely.

Sorry it got so large. I tried to keep the commits well-separated and reviewable individually.

fixes #780.


evap/evaluation/models.py
```diff
@@ -518,14 +523,14 @@ def update_courses(cls):
...
-                    if course.is_fully_reviewed():
+                    if course.is_fully_reviewed:
                         course.review_finished()
                         if not course.is_graded or course.final_grade_documents.exists() or course.gets_no_grade_documents:
                             course.publish()
                             evaluation_results_courses.append(course)
                     course.save()
             except Exception:
-                logger.exception(('An error occured when updating the state of course "{}" (id {}).').format(course, course.id))
+                logger.exception('An error occured when updating the state of course "%s" (id %s).', course, course.id)
```
janno42: didn't you want to use `{}`?

.landscape.yaml
```diff
@@ -7,6 +7,7 @@ ignore-paths:
 test-warnings: true
 python-targets:
     - 3
+max-line-length: 200
```
janno42: is this now official? in the past i tried to break after 120, but i'd love to be allowed to use 200 :)

karyon: i'd say the goal is still to keep lines as short as possible but as long as necessary... the 200 is rather arbitrary to keep the noise low. i also thought about disabling it completely, would that be better to not make it look like we agreed on some number?

janno42: yes, either define something that is a strict rule for our project or define nothing at all and comment on pull requests if lines get too long.

karyon: alright then i'll disable the line length lints.

janno42:
whenever you're happy: :shipit:

PR #793, opened 2016-05-09
Iss777

tomkellog:
changed title tags in student result view and staff manage course view to show current course name
#firstpullrequestever


karyon:
i may close your PR but your commits live on in #829 :)

PR #794, opened 2016-05-09
Update update.sh, fixes #787

### LINKED ISSUE #787: Deal with update.sh
update.sh is out of date. We can either
* remove it
* update it
* replace it with something else

having a script would be pretty cool to document what needs to be done when updating a production server.

maybe this can go into a management command.

an idea for additional automation would be a git post-commit-hook that checks out the release branch and updates everything when someone pushes to the production server via ssh. logs of the update could be sent via email to the admins.

karyon:
Fixes #787.


janno42:
:palm_tree:

PR #795, opened 2016-05-09
Remove passwords from crash reports, fixes #792

### LINKED ISSUE #792: Remove passwords from error messages
EvaP sends error messages to the admins when a server error occurs. Post data sent to the page where the error occurred will be included in the message, but passwords must not be included in these messages.
This currently happens e.g. when a user is logging in who has two accounts like described in #791.

karyon:
Fixes #792


janno42:
:dart:

PR #804, opened 2016-05-12
fix question formset

janno42:
This fixes the question formset when editing questionnaires (only one add another link, add another only shown when editable and correct select2 application to additional forms).


karyon:
please put the actual fixes into the commit message, on a second line if it gets too long.

:+1:

PR #805, opened 2016-05-14
Fix adding multiple questions at once

### LINKED ISSUE #803: Adding multiple questions at once does not work
When trying to add multiple questions to a questionnaire by adding forms to the formset and saving it only the last entry is saved in the database.

janno42:
fix #803; use correct form prefix and make questionnaire form editable when copying or creating new version


karyon:
:video_game:

PR #807, opened 2016-05-22
Private courses

### LINKED ISSUE #784: Internal courses
Allow courses to be flagged as internal. Results should be available to contributors and participants (and staff) as usual, but all other users of the platform should not be able to see the course or to view its results.

janno42:
fix #784
allows to create private courses. results of these courses are only visible for staff users, contributors and participants of the course.


evap/results/templates/results_course_detail.html
```diff
@@ -21,9 +21,19 @@
...
 
     {% if staff or contributor %}
-        {% if public_view %}
+        {% if course.is_private %}
+            <div class="alert alert-info hidden-print">
+                {% trans "This course is private. Only contributors and participants can see the results." %}
+            </div>
+        {% endif %}
+        {% if public_view and course.is_private %}
+            <div class="alert alert-info hidden-print">{% trans "This is the participant's view. Participants who are not contributors of this course nor their delegates can only see the results below." %}</div>
```
karyon: maybe drop the delegate part? it makes the sentence quite hard to parse.

evap/results/templates/results_course_detail.html
```diff
@@ -21,9 +21,19 @@
...
+            <a href="{% url "results:course_detail" course.semester.id course.id %}?public_view=false" class="pull-right btn btn-sm btn-primary hidden-print">{% trans "Show my view" %}</a>
+        {% elif public_view %}
             <div class="alert alert-info hidden-print">{% trans "This is the public view. Users who are neither contributors of this course nor their delegates can only see the results below." %}</div>
             <a href="{% url "results:course_detail" course.semester.id course.id %}?public_view=false" class="pull-right btn btn-sm btn-primary hidden-print">{% trans "Show my view" %}</a>
```
karyon: this line is exactly the same line as 3 lines above

evap/results/templates/results_semester_detail.html
```diff
@@ -47,8 +51,16 @@
...
                                 <td class="text-center"><div class="deviation-bg" style="background-color: {{ course.avg_deviation|deviationcolor }};">{{ course.avg_deviation|floatformat:1 }}</div></td>
                             {% else %}
-                                <td class="text-center"><div class="grade-bg grade-bg-disabled" data-toggle="tooltip" data-placement="left" title="{% trans "Not enough answers were given." %}">&mdash;</div></td>
-                                <td class="text-center"><div class="deviation-bg deviation-bg-disabled" data-toggle="tooltip" data-placement="left" title="{% trans "Not enough answers were given." %}">&mdash;</div></td>
+                                <td class="text-center"><div class="grade-bg grade-bg-disabled" data-toggle="tooltip" data-placement="left" title="
+                                    {% if course.is_private %}{% trans "The results of this course are private." %}
+                                    {% else %}{% trans "Not enough answers were given to publish the results." %}{% endif %}">
+                                        &mdash;
```
karyon: line breaks for the if-else please
also, the mdash is indented wrongly

evap/results/tests/test_views.py
```diff
@@ -2,7 +2,7 @@
 from model_mommy import mommy
 
 from evap.evaluation.models import Semester, UserProfile, Course, Contribution, Questionnaire
-from evap.evaluation.tests.test_utils import ViewTest
+from evap.evaluation.tests.test_utils import ViewTest, WebTest
```
karyon: unused import

evap/results/tests/test_views.py
```diff
@@ -52,12 +52,29 @@ def setUpTestData(cls):
         mommy.make(Contribution, course=cls.course, contributor=responsible, can_edit=True, responsible=True, comment_visibility=Contribution.ALL_COMMENTS)
         mommy.make(Contribution, course=cls.course, contributor=contributor, can_edit=True)
 
+        # Private course
+        student = UserProfile.objects.get(username="student")
+        test1 = mommy.make(UserProfile, username="test1")
+        test2 = mommy.make(UserProfile, username="test2")
+        cls.private_course = mommy.make(Course, state='published', is_private=True, semester=cls.semester, participants=[student, test1, test2], voters=[test1, test2])
+        mommy.make(Contribution, course=cls.private_course, contributor=responsible, can_edit=True, responsible=True, comment_visibility=Contribution.ALL_COMMENTS)
+        mommy.make(Contribution, course=cls.private_course, contributor=contributor, can_edit=True)
```
karyon: could you move the generation of data for specific tests to those specific tests?

evap/results/templates/results_course_detail.html
janno42: k

evap/results/templates/results_course_detail.html
janno42: yes, but i don't see a version that doesn't make everything even worse right now. do you have any suggestion?

evap/results/templates/results_semester_detail.html
janno42: k

evap/results/tests/test_views.py
janno42: k

evap/results/tests/test_views.py
janno42: k

evap/results/templates/results_course_detail.html
karyon:
```
{% if public_view  %}
    {% if private %}
        This is the participants view
    {% else %}
        This is the public view
    {% endif %}
    show my view button
{% elif foo  %}
...
```
?

evap/evaluation/models.py
```diff
@@ -284,7 +287,11 @@ def can_user_see_results(self, user):
         if user.is_staff:
             return True
         if self.state == 'published':
-            return self.can_publish_grades or self.is_user_contributor_or_delegate(user)
+            if self.is_user_contributor_or_delegate(user):
+                return True
+            if not self.can_publish_grades:
+                return False
+            return not self.is_private or user in self.participants.all()
```
karyon: maybe
```
if self.is_private and user not in self.participants.all():
    return False
return True
```
That's clearer in my opinion.

janno42: really? that would be
```
if user.is_staff:
    return True
if self.state == 'published':
    if self.is_user_contributor_or_delegate(user):
        return True
    if not self.can_publish_grades:
        return False
    if self.is_private and user not in self.participants.all():
        return False
    return True
return False
```
that's also not very easy to parse

evap/results/templates/results_course_detail.html
janno42: k

evap/evaluation/models.py
karyon: i thought, that is_private variable is supposed to restrict access if a condition is not met, and basically thats in the code
```
if private and not condition:
    restrict access
otherwise not
```
which i think is more straightforward than the current `give access exactly if not private and condition`

besides, i like the structure of handling three special cases, and then returning the default rather than handling two special cases and then returning the result of a third special case which has the default hidden in it. (only talking about the contents of the `if published` branch)

janno42: i will change it.
i see what you mean and it's okay for me and better than before but still not optimal. shouldn't restricting be actually the default case, so that we only explicitly allow some cases and restrict everything else? now there's a `return True` that might do unexpected stuff if we should ever have unhandled cases in there.
but i don't mind having at that way because i don't have a better solution in mind right now.

karyon: i thought about `return False` as default but didnt find something nice. i thought True as default would be ok since we're in state published, so everything is published by default except for special cases.

False as default might look like this:
```
if self.state == 'published':
    if self.is_user_contributor_or_delegate(user):
        return True
    if self.can_publish_grades:
        if not self.is_private or user in self.participants.all():
            return True
    return False
return False
```
i think that might be better. your call :)

janno42: meh. `if if` is also not really nice. let's keep the current one.

evap/results/templates/results_course_detail.html
```diff
@@ -21,9 +21,24 @@
...
         {% if public_view %}
-            <div class="alert alert-info hidden-print">{% trans "This is the public view. Users who are neither contributors of this course nor their delegates can only see the results below." %}</div>
+            {% if course.is_private %}
+                <div class="alert alert-info hidden-print">
+                    {% trans "This is the participant's view. Participants who are not contributors of this course can only see the results below." %}
+                </div>
+            {% else %}
+                <div class="alert alert-info hidden-print">
+                    {% trans "This is the public view. Users who are not contributors of this course can only see the results below." %}
+                </div>
```
karyon: pull the div out of the if else

evap/results/templates/results_semester_detail.html
```diff
@@ -47,8 +51,22 @@
...
+                                        {% trans "Not enough answers were given to publish the results." %}
+                                    {% endif %}">
+                                    &mdash;
+                                </div></td>
+                                <td class="text-center"><div class="deviation-bg deviation-bg-disabled" data-toggle="tooltip" data-placement="left" title="
+                                    {% if course.is_private %}
+                                        {% trans "The results of this course are private." %}
+                                    {% else %}
+                                        {% trans "Not enough answers were given to publish the results." %}
+                                    {% endif %}">
```
karyon: any way to deduplicate this if else?

besides, why is the course shown at all here if it is private?

evap/results/tests/test_views.py
```diff
@@ -26,7 +26,7 @@ def setUpTestData(cls):
 
 
 class TestResultsSemesterCourseDetailView(ViewTest):
-    url = '/results/semester/2/course/21'
+    url = '/results/semester/1/course/1'
```
janno42: @karyon can you please take over the tests? i don't even know what this part should be doing and why it worked before with the other course.

karyon: there is one automatic test that asserts that the view doesn't crash. if uses the data created in setUpTestData. in there, a course with id 21 is created, which is why this worked before.

please don't use the fixtures, i planned to remove them. what was the problem with creating the data via model mommy?

janno42: the course still is created - why doesn't it work now?
i would be happy creating the data with mommy, but i couldn't get the test running. please feel free to change everything.

karyon: because the course is created with the id 21, and the url accesses the course with the id 1, which may not exist at all.

janno42: i changed it to 1 because that works. the old one didn't.

karyon: um actually, the tests are green....?

evap/evaluation/models.py
```diff
@@ -283,17 +283,22 @@ def can_user_vote(self, user):
...
             return True
         if self.state == 'published':
-            if self.is_user_contributor_or_delegate(user):
-                return True
             if not self.can_publish_grades:
                 return False
-            if self.is_private and user not in self.participants.all():
-                return False
-            return True
+            return self.can_user_see_course(user)
```
karyon: this changes the methods behaviour. previously, contributors or delegates could always access the results, now only if course.can_publish_grades. i guess this is also the reason for the test failure.

evap/results/fixtures/minimal_test_data_results.json
```diff
@@ -428,7 +473,29 @@
...
+    "last_modified_time": "2014-09-16T23:01:13.059",
+    "_voter_count": 17,
+    "state": "published",
+    "last_modified_user": 2,
+    "participants": [5],
+    "vote_end_date": "2014-09-30",
+    "name_en": "Private course",
+    "type": 1,
+    "vote_start_date": "2014-09-10",
+    "is_private": true
```
karyon: please use model mommy again

evap/results/tests/test_views.py
```diff
@@ -58,6 +58,27 @@ def test_single_result_course(self):
         response = self.app.get(url, user=user)
         self.assertEqual(response.status_code, 200, 'url "{}" failed with user "{}"'.format(self.url, user))
 
+    def test_private_course(self):
+        private_course = Course.objects.get(name_en='Private course')
+
+        page = self.app.get("/results/semester/%s" % (private_course.semester.id), user='random')
```
karyon: please put the url into a variable

evap/results/tests/test_views.py
```diff
@@ -58,6 +58,36 @@ def test_single_result_course(self):
...
+        mommy.make(UserProfile, username="random")
+        private_course = mommy.make(Course, state='published', is_private=True, semester=self.semester, participants=[student, test1, test2], voters=[test1, test2])
+        mommy.make(Contribution, course=private_course, contributor=responsible, can_edit=True, responsible=True, comment_visibility=Contribution.ALL_COMMENTS)
+        mommy.make(Contribution, course=private_course, contributor=contributor, can_edit=True)
+
+        url = '/results/semester/%s' % (self.semester.id)
+        page = self.app.get(url, user='random')
+        self.assertNotIn(private_course.name, page)
+        page = self.app.get(url, user='student')
+        self.assertIn(private_course.name, page)
```
janno42: the test fails here, because no courses are listed on the results page. but i don't know why.

karyon:
:ambulance:

Issue #769, opened 2016-04-02
Remove test_coverage and test_usecases


karyon:
test_coverage contains some "real" tests which should be moved to appropriate places, and some tests that don't really test anything which should be removed.

test_usecases has tests that basically start on the main page, navigate to some view and then test that view. we should probably remove the navigating and make ordinary view tests out of them.

successor to #707

PR #811, opened 2016-05-30
Better testing, refactorings

karyon:
Some test refactoring towards #769. Not finished yet but since i got multiple conflicts already and it got quite large, i thought i keep it from rotting.

Some unrelated refactorings are in between.


evap/rewards/tools.py
```diff
@@ -68,10 +67,11 @@ def grant_reward_points(sender, **kwargs):
     if required_courses.exclude(voters=request.user).exists():
         return
     # did the user not already get reward points for this semester?
-    if not RewardPointGranting.objects.filter(user_profile=request.user, semester=semester):
-        granting = RewardPointGranting(user_profile=request.user, semester=semester, value=settings.REWARD_POINTS_PER_SEMESTER)
-        granting.save()
-        messages.success(request, _("You just have earned reward points for this semester because you evaluated all your courses. Thank you very much!"))
+    if RewardPointGranting.objects.filter(user_profile=request.user, semester=semester).exists():
```
janno42: what about `get_or_create` and just show the message if it was created?

evap/evaluation/models.py
```diff
@@ -1062,6 +1062,7 @@ def send_to_users_in_courses(cls, template, courses, recipient_groups, use_cc):
             # Collect recipients for all courses.
             # Don't include delegates and CC users of the responsible person of a course to the recipient list of this
             # course, because they will get the notification in CC anyway.
+            # TODO why only for the responsible, not for all contributors
```
janno42: and you will address this in the near future? i would prefer having a ticket over TODOs in the code ;)

evap/evaluation/models.py
```diff
@@ -1087,7 +1087,9 @@ def send_to_users_in_courses(cls, template, courses, recipient_groups, use_cc):
     @classmethod
     def __send_to_user(cls, user, template, subject_params, body_params, use_cc):
         if not user.email:
-            logger.warning("{} has no email address defined. Could not send email.".format(user.username))
+            warning_message = "{} has no email address defined. Could not send email.".format(user.username)
```
janno42: translate

evap/evaluation/models.py
janno42: never mind

evap/rewards/tools.py
karyon: k

evap/evaluation/models.py
karyon: well, i can apply the translation only in messages.warning.

janno42: that's probably the best option.

evap/rewards/tools.py
karyon: just checked, the method has 5 returns if a bunch of conditions is not met. i think it's nice to stay in that schema. also, the intent is expressed explicitly with this: if points were granted, do nothing, otherwise, grant points.

janno42:
:key:

PR #817, opened 2016-06-05
Safer update_production.sh

### LINKED ISSUE #799: +x on update.sh, earlier apache restart
update_production.sh is missing the x bit, also because of the cache clearing the apache is restarted 2min after the code has changed.

karyon:
Fixes #799.

The basic problem is to update production in a safe, simple way with no downtimes. specifically, the question is how to apply the migrations. updating static files might have similar problems (but cannot produce server crashes, just UI weirdnesses i think)

* Option 1: migrate, then reload. simple and no downtime, but there might be time windows where things crash.
* Option 2: apache stop, migrate, apache start. safe and simple, but with downtimes.
* Option 3: do lots of engineering work explained in [this blog post](http://pankrat.github.io/2015/django-migrations-without-downtimes/). safe and no downtimes, but not simple.

Option 3 is not an option, so we can either be safe and accept downtimes or unsafe and accept potential crashes. We did the latter before and this PR does not change that.

I thought about whether to apply the migrations before or after reloading apache. consider the following:

time in which code version and database migration level do not match (because of which things could break):
* first migrate, then reload: first migration commit until reload. close to zero in case of a single migration.
* first reload, then migrate: reload until last migration commit. at least the duration of one migration

The first option seemed like the better one.

Keep in mind that I didn't really know what i was doing, all this is unverified and might be plain wrong. but I did test this on the staging server :)


deployment/update_production.sh
```diff
@@ -6,13 +6,16 @@ set -x # print executed commands
...
-sudo -u evap ./manage.py migrate
-sudo -u evap ./manage.py collectstatic --noinput
 sudo -u evap ./manage.py compilemessages
+sudo -u evap ./manage.py collectstatic --noinput
 sudo -u evap ./manage.py compress --verbosity=0
+sudo -u evap ./manage.py migrate
+# reload only after static files are updated, so the new code finds all the files it expects.
+# also, reload after migrations happened. see https://github.com/fsr-itse/EvaP/pull/817 for a discussion.
+sudo service apache2 reload
+# update chaches. this can take minutes but doesn't need a reload.
```
janno42: chaches ;)

janno42:
:e-mail:

PR #822, opened 2016-06-06
Set question order

### LINKED ISSUE #798: Add Question order to test data
After introducing an order field to Questions, the test data should get meaningful values instead of the current -1 for all Questions.

karyon:
fixes #798


janno42:
:baby_chick:

Issue #788, opened 2016-05-05
Deal with BootstrapMixin and QuestionnaireSelectMultiple


karyon:
evaluation/forms.py contains `QuestionnaireSelectMultiple`, which i think adds html titles to the individual choices, and `BootstrapMixin`, of which i don't exactly know what it does. Since they are pretty hard to understand and undocumented, we should either

* figure out what they do and properly document them or
* replace them.

the source of BootstrapMixin says that one should use https://github.com/maraujop/django-crispy-forms instead. i don't know whether that makes sense for us, but we should investigate.

PR #823, opened 2016-06-06
Refactor QuestionnaireMultipleChoiceField handling of additional description text

Steditor:
Contributes to #788

### DIFF
--- a/evap/evaluation/forms.py
+++ b/evap/evaluation/forms.py
@@ -3,7 +3,6 @@ from itertools import chain
 from django import forms
 from django.contrib.auth import authenticate
 from django.forms import widgets
-from django.forms.models import ModelChoiceIterator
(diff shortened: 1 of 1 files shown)


karyon:
hey, will you be there on monday? if not, could you rebase this?

Steditor:
I'm not yet sure, whether I'll be there - but I've just rebased anyway.

karyon:
do you remember why we can't render this by ourselves in the template?

karyon:
i have something in the pipeline which kinda makes this obsolete, i now render the widget manually in the templates. but, as overwriting some widget renderers by now actually seems like the technically cleaner solution to me, i'll merge this first, it might be a starting point for more investigation... somewhen... in a few years...

karyon:
thanks!

PR #827, opened 2016-06-13
Remove courses with not enough answers from default export

### LINKED ISSUE #814: Export contains unpublished results even if option is not checked
The excel export of results contains columns for unpublished courses and shows a total grade even if the export option "Include courses where not enough answers where given in the export" is not selected.

In this situation, the course should not be contained in the export.

simonkrogmann:
Fix #814


karyon:
:+1: @janno42, i'll leave merging to you

PR #828, opened 2016-06-13
More javascript compression

### LINKED ISSUE #800: Some javascript files are not compressed
Some javascript files, e.g. datatables, are not run through django-compressor. Thus the cache busting does not work, and browsers might use old versions after a production update.

karyon:
fixes #800


janno42:
:lollipop:

PR #833, opened 2016-06-28
use django-bootstrap3

### LINKED ISSUE #788: Deal with BootstrapMixin and QuestionnaireSelectMultiple

karyon:
use django-bootstrap3.

one functional change is that, when a form error occurs, all correct form fields are now green. the other is that in the staff course form, the headline is now the course form, not "General Options".

there's still something missing, i'd like to look for places where we can use bootstrap3's error rendering. search term would be `errors }}` and there are files such as staff_course_email.html which could be radically easier e.g. using the required attribute

fixes #788


evap/staff/forms.py
```diff
@@ -212,10 +216,10 @@ def save(self, *args, **kw):
         self.instance.save()
 
 
-class ContributionForm(forms.ModelForm, BootstrapMixin):
+class ContributionForm(forms.ModelForm):
     responsibility = forms.ChoiceField(widget=forms.RadioSelect(), choices=Contribution.RESPONSIBILITY_CHOICES)
     course = forms.ModelChoiceField(Course.objects.all(), disabled=True, required=False, widget=forms.HiddenInput())
-    questionnaires = QuestionnaireMultipleChoiceField(Questionnaire.objects.filter(is_for_contributors=True, obsolete=False), label=_("Questionnaires"))
+    questionnaires = forms.ModelMultipleChoiceField(Questionnaire.objects.filter(is_for_contributors=True, obsolete=False), widget=CheckboxSelectMultiple, label=_("Questionnaires"))
```
janno42: why didn't you put some linebreaks here? :)

janno42:
:crescent_moon:

PR #838, opened 2016-06-30
Fix datepicker

### LINKED ISSUE #836: Fix datepicker
With #833, the datepicker in form fields does not work anymore.

karyon:
Fixes #836.

The fix is taken from 1327. it's not pretty but quite effective :) it requires any date fields' name to end with (or contain?) `date`. I think the alternative would be to extend the renderers of bootstrap3.


janno42:
:tangerine:

PR #840, opened 2016-07-11
Show questionnaire preview to editors in all states before published

janno42:


karyon:
well done, janno, well done.

PR #842, opened 2016-07-11
Backup in update_production.sh

### LINKED ISSUE #837: Create backup in update_production
It would be nice if the `update_production` script would optionally create a backup database dump to a configurable path.

karyon:
Fixes #837

giving the parameter a name would've become ugly in bash. the alternative would've been to make this a python script, which i didn't feel like doing now.


janno42:
:cactus:

PR #844, opened 2016-07-11
Fix questionnaire formset in questionnaire create form, fixes #835

### LINKED ISSUE #835: Fix formset for creating questionnaire
When creating a new questionnaire, the formset does not have an `add another` link and the remove checkbox is not replaced by a `remove` link.

karyon:
Fixes #835


janno42:
:large_orange_diamond:

PR #845, opened 2016-07-11
Make update_courses set last_modified_user to the new cronjob user

### LINKED ISSUE #819: Update "Last modified by" in cronjob
When the cronjob changes a course's state, the field `Last modified` is updated with the respective time, but the field `Last modified by` is not changed. This field should then be empty.

karyon:
Fixes #819


karyon:
fixed the test failure.

janno42:
maybe you could remove the cronjob user from all user lists? especially in the contributor selection form that would be nice. but it could also be removed in the staff user list.

karyon:
added the custom manager.
```
>>> UserProfile.objects.filter(username="cronjob")
[]
>>> UserProfile.objects.with_cronjob_user().filter(username="cronjob")
[<UserProfile: cronjob>]
```

janno42:
:v:

PR #848, opened 2016-07-16
Ensure django-auth-kerberos creates lowercase usernames

### LINKED ISSUE #791: Username case sensitivity
Usernames are case sensitive. The importer makes all usernames lowercase, but automatically created accounts when logging in with Kerberos authentification can have uppercase letters.
This can lead to two users having the same username and then the system crashed on login.

Automatically created accounts should also get lowercase usernames, even if the user enters the name differently.

karyon:
i couldn't test this properly, but in authenticate(), django-auth-kerberos creates the userprofile, so this seemed to be the appropriate place to lowercase the username.

fixes #791


janno42:
:panda_face:

PR #849, opened 2016-07-16
Rewards expired events 846

### LINKED ISSUE #846: Allow redeeming points only for open events
Currently there is no check that the event for which points should be redeemed is still available for redemptions. If the `redeem_end_date` is reached, the event can't be selected on the rewards index page anymore, but a correct post request can still redeem points of a user for this event.
This should not be possible.

### LINKED ISSUE #847: Count number of people for redemption events instead of number of redemptions
Currently the number of redemptions is counted per event. Instead, the number of users who redeemed points should be shown on the staff page.

karyon:
i'm open for suggestions for the ui strings.
fixes #847, fixes #846

### DIFF
--- a/evap/rewards/models.py
+++ b/evap/rewards/models.py
@@ -14,6 +14,11 @@ class NotEnoughPoints(Exception):
     pass
 
 
+class RedemptionEventExpired(Exception):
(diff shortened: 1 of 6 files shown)


janno42:
:fast_forward:

PR #852, opened 2016-07-23
No double votes

### LINKED ISSUE #850: Votes might get counted twice
`student vote` checks whether a user can vote and raises a `PermissionDenied` if not.
but if the post request with the votes for a course is sent twice immediately after each other this check can in both cases return `True` if the vote was not saved before the second request is handled. then, the votes would be saved twice.

karyon:
fixes #850 

first and third commits are alternatives. i like the first one better. the second one is more explicit, but both require the explaining comment, and with that, the first one is enough imo. what do you think?

### DIFF
--- a/evap/student/templates/student_vote.html
+++ b/evap/student/templates/student_vote.html
@@ -108,7 +108,12 @@
         var form = $('#student-vote-form');
         form.submit(function (event) {
             event.preventDefault(); // don't use the default submission
-            $('#vote-submit-btn').prop('disabled', true)
+            var submitButton = $('#vote-submit-btn');
+            var originalText = submitButton.val();
+
+            submitButton.val("Submitting...");
+            submitButton.prop('disabled', true);
+
             $.ajax({
                 type: form.attr('method'),
                 url: form.attr('action'),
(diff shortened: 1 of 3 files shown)


janno42:
as voting is the most important feature of the whole system, i would prefer the second version and would add an explicit entry to the log file explaining from which user and for which course the double vote was recorded so that we could further investigate if any problems with the voting procedure should occur in the future.

karyon:
well, if we just log it, nobody is going to notice, just like in the last few years.

janno42:
then make it an error and send an email to the admins :)

karyon:
the button disable thing is ok? it might be slightly confusing, because when clicking the button, it gets disabled and the mouse pointer turns into that red forbidden-sign before the new page is loaded...

janno42:
maybe you could also change the button text to something like "Submitting questionnaire..."? then the user gets instant feedback,

karyon:
As discussed, the SuspiciousOperation gets logged and send to the admins by default, so there's no need for any additional logging.

evap/student/templates/student_vote.html
```diff
@@ -108,6 +108,12 @@
         var form = $('#student-vote-form');
         form.submit(function (event) {
             event.preventDefault(); // don't use the default submission
+            var submitButton = $('#vote-submit-btn');
+            var originalText = submitButton.val();
+
+            submitButton.val("Submitting...");
```
janno42: `Submitting ...`

janno42:
:rabbit2:

PR #866, opened 2016-10-17
Fix #865 adds username to user selection list on user merging

### LINKED ISSUE #865: User merge: user selection shows only first name and last name
When selecting users for merging, their first and last name is shown. this breaks if both have the same first and last name, in which case the staff user cannot know which one should be the main or other user. a possible solution would be to additionally show the username.

leonardMerzljak:
fixes #865

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -900,6 +900,13 @@ class UserProfile(AbstractBaseUser, PermissionsMixin):
         else:
             return self.username
 
+    @property
+    def full_name_with_username(self):
+        name = self.full_name
+        if self.username not in name:
+            name += " (" + self.username + ")"
+        return name
+
     def __str__(self):
         return self.full_name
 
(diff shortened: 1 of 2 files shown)


evap/evaluation/models.py
```diff
@@ -900,6 +900,18 @@ def full_name(self):
...
+    def full_name_with_username(self):
+        name = self.username
+        if self.last_name:
+            name = self.last_name
+            if self.first_name:
+                name = self.first_name + " " + name
+            if self.title:
+                name = self.title + " " + name
+            name += " (" + self.username + ")"
+        return name
```
karyon:
```
name = self.full_name
if self.username not in name:
    name += " (" + self.username + ")"
return name
```
should do the same i think?

karyon:
oh, and landscape complains about missing blank lines, adding these would be nice as well :)

karyon:
sorry for the delay. @leonardMerzljak the PR looks good, but we've merged other stuff in the meantime, could you give it a rebase? should be something like `git fetch upstream` and `git rebase upstream/master` (and then a force push,  `git push --force-with-lease`)

karyon:
Testing out that "squash and merge" feature of github. @leonardMerzljak no rebase needed anymore :)

PR #870, opened 2016-10-17
Prevent staff users from removing responsible contributors

### LINKED ISSUE #781: Responsible contributor must not be removed by staff
Currently staff users can remove the responsible contributor from a course in the UI. Saving fails, but there is no easy way to assign another contributor as responsible.
The remove link should not be enabled for the contributor set as responsible.

frederike-ramin:
Fixes #781

### DIFF
--- a/evap/evaluation/templates/contribution_formset.html
+++ b/evap/evaluation/templates/contribution_formset.html
@@ -52,9 +52,7 @@
                 </td>
                 {% if editable %}
                     <td>
-                        {% if staff or responsible %}
-                            {{ form.DELETE }}
-                        {% endif %}
+                        {{ form.DELETE }}
                     </td>
                 {% endif %}
             </tr>
--- a/evap/evaluation/templates/contribution_formset.html
+++ b/evap/evaluation/templates/contribution_formset.html
@@ -52,7 +52,7 @@
                 </td>
                 {% if editable %}
                     <td>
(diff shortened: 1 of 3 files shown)


evap/evaluation/templates/contribution_formset.html
```diff
@@ -52,7 +52,7 @@
                 </td>
                 {% if editable %}
                     <td>
-                        {% if staff or not form.responsibility.value == "RESPONSIBLE" %}
+                        {% if staff or responsible %}
```
karyon: what was the intention behind this? because right now it checks whether the `responsible` variable contains something, which it always does.

evap/evaluation/templates/evap_course_edit_js.html
```diff
@@ -20,6 +20,8 @@
                     if (j != i)
                         disableButtons("input[name=contributions-" + j + "-responsibility][value=RESPONSIBLE]");
                 }
+                //disable remove link
```
karyon: space after // please

evap/evaluation/templates/contribution_formset.html
karyon: @janno42 that if can simply be removed right? we always render the link, and the JS code removes it for the responsible as far as i understand

janno42: yes, please remove the `if` completely

karyon:
thanks!

PR #874, opened 2016-11-27
Safer post requests

### LINKED ISSUE #853: Verify all POSTS can be sent twice safely
since we had #850 and #846, we should check all the POSTS that we have whether they can be safely twice, delayed, with malformed data etc.

karyon:
fixes #853.

the problem of clicking multiple times on a button and sending the post requests multiple times can only happen if we overwrote the button's action with an ajax request. otherwise, the browser will prevent sending multiple requests.

so, to prevent nervous users from sending multiple requests, we need to check only the requests that are sent with ajax. the vote button (#852) was such a case, and in this PR i've modified the feedback form to prevent multiple submits.

what's left is malicious users who manually send POST requests. i've hardened the reward point redemption view, but apart from that didn't see anything where users can do any damage. however, i did not check the staff and grades apps, that seemed too much work and not worth it, since i'd say if staff users or grade uploaders are malicious, we have a totally different kind of problem.


karyon:
@janno42 could you confirm that when deleting GradeDocuments, the actual file is not deleted? i think we should do something about that, shouldn't we?

Issue #786, opened 2016-05-01
Sorting by name


janno42:
Currently objects (e.g. courses) are sorted by their name in multiple use cases. On the results pages courses are sorted in the datatable on client side. This is done after translating the name into the current language.

But e.g. in exports they are always sorted by their German name. This should also adapt to the current language if possible.

PR #876, opened 2016-11-29
#786: Exports are now sorted correctly. Added tests for the sorting.

He3lixxx:


evap/results/exporters.py
```diff
@@ -107,6 +107,7 @@ def export(self, response, course_types_list, include_not_enough_answers=False,
                     used_questionnaires.add(questionnaire)
                 courses_with_results.append((course, results))
 
+            courses_with_results.sort(key=lambda cr: cr[0].name)
             courses_with_results.sort(key=lambda cr: cr[0].type)
```
karyon: found this: http://stackoverflow.com/questions/4233476/sort-a-list-by-multiple-attributes
```
courses_with_results.sort(key=lambda cr: (cr[0].type, cr[0].name))
```

evap/staff/tests/test_views.py
```diff
@@ -84,9 +84,40 @@ def test_deletes_users(self):
...
+        page = self.app.get(self.url, user='staff').body.decode("utf-8")
+        position_course1 = page.find("Course 1")
+        position_course2 = page.find("Course 2")
+        self.assertLess(position_course1, position_course2)
+
+
 class TestSemesterExportView(ViewTest):
     url = '/staff/semester/1/export'
     test_users = ['staff']
+    extra_environ = {'HTTP_ACCEPT_LANGUAGE': 'en'}
```
karyon: is this still necessary here (other test as well)? if not, i'd prefer to just set this inside the test

He3lixxx: I moved it to the app.get() and form.submit() calls.

karyon: that's even better.

PR #880, opened 2016-12-05
Django 1.10 upgrade

### LINKED ISSUE #861: Django 1.10 upgrade
I have a branch open where all tests are passing and all deprecations are fixed, but the UI is not tested yet and there are some cool new features we might make use of.

Release notes of interest from https://docs.djangoproject.com/en/1.10/releases/1.10/:
- Required form fields now have the required HTML attribute. Set the new Form.use_required_attribute attribute to False to disable it. The required attribute isn’t included on forms of formsets because the browser validation may not be correct when adding and deleting formsets.
- Required form fields now have the required HTML attribute. Set the Form.use_required_attribute attribute to False to disable it. You could also add the novalidate attribute to `<form>` if you don’t want browser validation.
- [since the html required attribute does not work well with the fixed bootstrap navbar, we might need to disable it for every single field which is dumb. the global setting does not work, see https://github.com/dyve/django-bootstrap3/issues/325, so we might need to wait for https://github.com/dyve/django-bootstrap3/issues/337]
- Added the is and is not comparison operators to the if tag.
- Request and response handling of the runserver command is sent to the django.server logger instead of to sys.stderr. If you disable Django’s logging configuration or override it with your own, you’ll need to add the appropriate logging configuration if you want to see that output:

ref https://github.com/django-webtest/django-webtest/pull/56

karyon:
fixes #861

### DIFF
--- a/evap/evaluation/auth.py
+++ b/evap/evaluation/auth.py
@@ -46,7 +46,7 @@ class RequestAuthMiddleware(object):
 
         # If we already have an authenticated user don't try to login a new user. Show an error message if another user
         # tries to login with a URL in this situation.
-        if request.user.is_authenticated():
+        if request.user.is_authenticated:
             if user != request.user:
                 messages.error(request, _("Another user is currently logged in. Please logout first and then use the login URL again."))
             return
@@ -96,7 +96,7 @@ def user_passes_test(test_func):
     def decorator(view_func):
         @wraps(view_func, assigned=available_attrs(view_func))
         def _wrapped_view(request, *args, **kwargs):
-            if not request.user.is_authenticated():
+            if not request.user.is_authenticated:
                 return redirect_to_login(request.get_full_path())
             if not test_func(request.user):
                 raise PermissionDenied()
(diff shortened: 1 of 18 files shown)


requirements.txt
```diff
@@ -1,8 +1,8 @@
-django >= 1.9, < 1.10
+django >= 1.10, < 1.11
 xlrd == 1.0.0
 xlwt == 1.1.2
 django-fsm == 2.4.0
-django-webtest == 1.8.0
+django-webtest >= 1.8.0
```
janno42: why do you allow all later versions (including major and minor version changes) here but not anywhere else?

requirements.txt
```diff
@@ -1,8 +1,8 @@
...
 django-widget-tweaks == 1.4.1
 WebTest==2.0.23
```
janno42: what about some spaces here? ;)

evap/evaluation/auth.py
```diff
@@ -23,6 +23,16 @@ class RequestAuthMiddleware(object):
 
     field_name = "loginkey"
 
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        self.process_request(request)
+
```
janno42: why are there so many empty lines?

PR #887, opened 2017-01-02
fix #869; rename user fsr to staff

### LINKED ISSUE #869: staff test user is called "fsr"
should be called "staff"

leonardMerzljak:


karyon:
thanks!

PR #892, opened 2017-01-02
fix #875

### LINKED ISSUE #875: When deleting GradeDocuments, the file is not deleted
GradeDocuments have an actual file stored somewhere, but when deleting the GradeDocument, the file is not deleted. It should be deleted.

leonardMerzljak:


karyon:
can you fix the landscape warnings? you can't do anything about the unused arguments, but at least the blank lines thing.

also, could you give the commit a more verbose commit message? like "Implement foo/fix crash in bar/make button green, fixes #875", because otherwise you have no idea what the commit does without looking into the issue.

karyon:
Finished in #913.

PR #893, opened 2017-01-02
Modifiable Email Template on course operations. Fixes #350

### LINKED ISSUE #350: Customized email when sending course back to responsible person
The platform needs a feature that allows to return a course to a lecturer for approval with a customized message instead of the standard "New courses ready for approval" email.

schumannd:
Fixes #350

### DIFF
--- a/evap/evaluation/tools.py
+++ b/evap/evaluation/tools.py
@@ -256,8 +256,11 @@ def is_external_email(email):
     return not any([email.endswith("@" + domain) for domain in settings.INSTITUTION_EMAIL_DOMAINS])
 
 
-def send_publish_notifications(courses, template):
+def send_publish_notifications(courses, template=None):
     publish_notifications = defaultdict(set)
+    
+    if not template:
+        template = EmailTemplate.objects.get(name=EmailTemplate.PUBLISHING_NOTICE)
 
     for course in courses:
         # for published courses all contributors and participants get a notification
--- a/evap/evaluation/tools.py
+++ b/evap/evaluation/tools.py
@@ -256,7 +256,7 @@ def is_external_email(email):
     return not any([email.endswith("@" + domain) for domain in settings.INSTITUTION_EMAIL_DOMAINS])
 
 
-def send_publish_notifications(courses):
+def send_publish_notifications(courses, template=None):
     publish_notifications = defaultdict(set)
 
     for course in courses:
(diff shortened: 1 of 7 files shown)


evap/staff/templates/staff_course_operation.html
```diff
@@ -20,11 +20,17 @@
...
         </table>
         {% if show_email_checkbox %}
+        <div class="form-group">
             <div class="checkbox">
                 <label>
                     <input type="checkbox" name="send_email" checked> {% trans "Send email notifications" %}
                 </label>
             </div>
+            <label>Email subject:</label> <br>
+            <input type="text" name="email_subject" value="{{ email_template.subject}}" style="width: 100%;"> <br>
```
janno42: the class `form-control` is missing

evap/staff/templates/staff_course_operation.html
```diff
@@ -20,11 +20,17 @@
...
+            <label>Email body:</label>
+            <textarea class="form-control" name="email_body" style="resize:vertical;min-height:300px;"> {{ email_template.body }} </textarea>
+        </div>
         {% endif %}
         <button type="submit" name="operation" value="{{ operation }}" class="btn btn-primary">{% trans "Confirm" %}</button>
```
janno42: - please put the buttons into a `<div class="well submit-area text-center">` (see email pages of courses)
- and please use the form design we have on other pages (e.g. the email page of courses), where form labels are besides the input fields and have no colon.

evap/staff/tests/test_views.py
```diff
@@ -35,6 +35,23 @@ def test_num_queries_is_constant(self):
...
+class TestSemesterCourseOperationView(ViewTest):
+    @classmethod
+    def setUpTestData(cls):
+        mommy.make(UserProfile, username='staff', groups=[Group.objects.get(name='Staff')])
+
+    def test_operation_start_evaluation(self):
+        sem = mommy.make(Semester,pk=1)
+        mommy.make(Course, pk=1, state='approved', semester=sem)
+
+        url = '/staff/semester/1/courseoperation?course=1&operation=startEvaluation'
```
schumannd: this view does indeed need some setup for the simple 200_OK check. So I decided not to set test_users

evap/evaluation/tools.py
```diff
@@ -275,7 +275,10 @@ def send_publish_notifications(courses):
             publish_notifications[course.responsible_contributor].add(course)
 
     for user, course_set in publish_notifications.items():
-        EmailTemplate.send_publish_notifications_to_user(user, list(course_set))
+        if not template:
+            template = EmailTemplate.objects.get(name=EmailTemplate.PUBLISHING_NOTICE)
```
schumannd: It would be 3 places where this line would be put into the method call. I would prefer to keep it that way as executing this method with a specific template != PUBLISHING_NOTICE seems to be the special case.

karyon: you're right, i missed the other two places.

please put this line outside the for loop, preferably at the beginning of the function.

karyon:
thanks :)

PR #894, opened 2017-01-02
fix #890

### LINKED ISSUE #890: Make number of enrollments before warning is shown a setting
the importers show a warning `Warning: User {} has {} enrollments, which is a lot.` if a user has more than a certain number of enrollments in the import. this number should be configurable in the settings file.

leonardMerzljak:
fixes #890


evap/settings.py
```diff
@@ -60,6 +60,7 @@
 # the importer accepts only these two strings in the 'graded' column
 IMPORTER_GRADED_YES = "yes"
 IMPORTER_GRADED_NO = "no"
+IMPORTER_MAX_ENROLLMENTS = 7
```
karyon: please add a comment above which explains what the setting does.

karyon:
i added the missing comment. would be nice to see you again!

PR #895, opened 2017-01-03
Remove on_document_ready leftovers

karyon:
note the removal in rewards_index.html. to me it looked like it did nothing.

### DIFF
--- a/evap/evaluation/templates/index.html
+++ b/evap/evaluation/templates/index.html
@@ -66,6 +66,8 @@
     </div>
 {% endblock %}
 
-{% block on_document_ready %}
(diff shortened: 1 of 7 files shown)


evap/evaluation/templates/index.html
```diff
@@ -65,7 +65,3 @@
         </div>
     </div>
 {% endblock %}
-
-{% block on_document_ready %}
-    $("#id_username").focus();
```
janno42: does that remove the initial focus from the username field?

karyon: yeah of course it does. i'll remove that diff.

PR #899, opened 2017-01-19
Misc

karyon:
* Move coverage tests to better places
* Move and rename some templates and filters
* update dependencies
* refine update_production.sh. Now the pip dependencies are installed only for the evap user. wsgi should now run in daemon mode.


karyon:
reminder to self: mail exceptions in management commands to admins

karyon:
Ready for review. I need to verify the logged exception actually gets sent as mail to the admins, but that's just a logging config change at most.

Also, when calling a management command with that decorator from the command line, the exception will be printed twice. i could swallow the exception in the decorator but thought that wouldn't be cool, since that decorator should log exceptions and not swallow them.

PR #906, opened 2017-01-31
Fixes #900, adds (usernames) to ModelChoiceFields (Simple and Multiple)

### LINKED ISSUE #900: Show username in all user selection fields
#866 implemented that the username is shown along the first and last name in the user merge selection fields. This should also be done for all other user selection fields, e.g. on the lecturer's course edit page.

MerlindlH:
This commit adds corresponding Usernames in brackets to persons in selectionfields for one or multiple persons.

Fixes #900

### DIFF
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -9,7 +9,7 @@ from django.db.models import Q
 
 from evap.evaluation.models import Course, UserProfile, Questionnaire, Semester
 from evap.staff.forms import ContributionForm
-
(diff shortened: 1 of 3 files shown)


karyon:
there are test failures :) i've read a bit in the docs and here's why: https://docs.djangoproject.com/en/1.10/topics/forms/modelforms/#overriding-the-default-fields (in the green box if you scroll down a bit). also, reading this, i noticed maybe it's cooler to use `Meta.field_classes` instead of explicitly defining that field to avoid that problem.

Issue #824, opened 2016-06-07
Copy participants from other course


janno42:
It should be possible to not only import participants to a course from an excel sheet, but also to copy the participants from another course.

karyon:
what's the use case for this?

janno42:
e.g. courses for participants in bachelor projects. they participate in multiple preparation seminars that are not part of the enrollment data. after manually adding them all to one course it would be nice to just be able to copy them instead of adding them one by one again.

janno42:
Actually, copying a whole course would probably be even better.

He3lixxx:
I think this commit (#881) introduces another problem since the UserImportForm is used in two places:
- user import (/staff/user/import)
- course participant import (e.g. staff/semester/18/course/1603/participant_import).

With the current changes, the course selection is also shown when you want to import users which I don't understand.

Also, the idea of having the import-from-another-course functionality in the same form as the import-from-excel-file functionality makes the interface more confusing when implementing the 2-stage file import mechanism as discussed on Issue #678. I'd suggest leaving it on the same page but somehow making a distinction by putting them in two seperate areas that are completely independent on eachother.

karyon:
> With the current changes, the course selection is also shown when you want to import users which I don't understand.
That's a bit weird indeed...

If you separate these parts, it might make sense to put the import-from-course into a separate form, which is displayed on the one page, but not on the other.

Issue #724, opened 2016-02-02
More verbose user imports


janno42:
Re #708 
When importing participants into a course the importer should show if participants already exist in the course which are also included in the import file. It should list the respective names in a warning.
Also, the message informing about the number of added participants should not be the length of the import list but the actual number of added participants (so without the ones already existing).

Also, when importing users or enrollment data, the importer should show which users have been newly created, not only the count.

He3lixxx:
Also, on the user, semester and course participant import test runs, there should be an indication on what effects the import would have (i.e. how many users would be added). See the comment I added in #910 on eval/staff/views.py, course_participant_import, line 675.

PR #910, opened 2017-02-01
Seperated importer warnings, forced test run for importers

### LINKED ISSUE #678: Separate importer warnings
The importer warning when user data changes should be separated into different groups of warnings. One for each of the following:
- username changed
- name changed
- email changed

Each of these warning groups should be collapsible.

He3lixxx:
also, Copy Participants UI introduced in #824 is fixed.

fixes #678.


evap/staff/views.py
```diff
@@ -311,27 +313,51 @@ def semester_import(request, semester_id):
...
+                delete_import_file(str(request.user.id))
+                return render(request, "staff_semester_import.html", dict(semester=semester,
+                    excel_form=excel_form, warnings=warnings, errors=errors, test_passed=False))
+
+        elif import_run:
+            import_filename = get_import_file_name_if_exists(str(request.user.id))
+            if import_filename:
+                vote_start_date = excel_form.cleaned_data['vote_start_date']
+                vote_end_date = excel_form.cleaned_data['vote_end_date']
+                if vote_start_date is None or vote_end_date is None:
```
karyon: please check if you can make these field required based on the action and do a form.is_valid instead of this here

evap/staff/views.py
```diff
@@ -626,40 +652,65 @@ def course_participant_import(request, semester_id, course_id):
...
+                return render(request, "staff_course_participant_import.html", dict(
+                    course=course, excel_form=excel_form, copy_form=copy_form, errors=errors, semester=semester, participant_test_passed=False))
 
-        # Import user from either excel file or other course
-        imported_users = []
-        if excel_file:
-            imported_users = UserImporter.process(request, excel_file, test_run)
         else:
-            imported_users = import_course.participants.all()
+            return render(request, "staff_course_participand_import.html", dict(course=course, excel_form=excel_form,
```
karyon: participand

evap/staff/views.py
```diff
@@ -626,40 +652,65 @@ def course_participant_import(request, semester_id, course_id):
...
+
+        elif copy_run and copy_form.is_valid():
+            import_course = copy_form.cleaned_data['course']
+            if import_course is not None:
+                imported_users = import_course.participants.all()
+                # fallthrough to end
+            else:
+                errors = ["You have to select one course from the drop-down."]
+                return render(request, "staff_course_participant_import.html", dict(
+                    course=course, excel_form=excel_form, copy_form=copy_form, errors=errors, semester=semester, participant_test_passed=False))
```
karyon: please check if you can use more fallthroughs and have `render(request, "staff_course_participant_import.html", ...)` in fewer, ideally only one place.

evap/staff/views.py
```diff
@@ -1033,21 +1084,39 @@ def user_create(request):
...
-        excel_file = form.cleaned_data['excel_file']
-        UserImporter.process(request, excel_file, test_run)
-        if test_run:
-            return render(request, "staff_user_import.html", dict(form=form))
-        return redirect('staff:user_index')
+        import_run = operation == 'import'
+
+        if test_run and excel_form.is_valid():
+            excel_file = excel_form.cleaned_data['excel_file']
+            _, warnings, errors = UserImporter.process(request, excel_file, test_run)
```
karyon: `__` please, `_` is an import, i don't want to know in how many ways this can break things :)

evap/staff/tools.py
```diff
@@ -7,11 +8,40 @@
 from django.core.cache import cache
 from django.core.cache.utils import make_template_fragment_key
 from django.db import transaction
+from django.conf import settings
 
 from evap.evaluation.models import UserProfile, Course, Contribution
 from evap.grades.models import GradeDocument
 
 
+def generate_import_file_name(user_id):
```
karyon: please use filename without underscore everywhere

evap/staff/tools.py
```diff
@@ -7,11 +8,40 @@
...
+
+
+def delete_import_file(user_id):
+    filename = generate_import_file_name(user_id)
+    if not os.path.isfile(filename):
+        return False
+    os.remove(filename)
+
+
+def get_import_file_name_if_exists(user_id):
```
karyon: please check whether returning the actual file (not name) or None simplifies things

evap/staff/importers.py
```diff
@@ -86,18 +87,34 @@ def store_in_database(self, vote_start_date, vote_end_date, semester):
...
     def read_book(self, excel_file):
-        self.book = xlrd.open_workbook(file_contents=excel_file.read())
+        if excel_file is None:
+            self.errors.append(_("Please select an Excel file."))
+            return
+        try:
+            if isinstance(excel_file, UploadedFile):
+                self.book = xlrd.open_workbook(file_contents=excel_file.read())
+            else:
+                self.book = xlrd.open_workbook(excel_file)
```
karyon: i think i get it now. can you check whether this can be made unnecessary when applying my comment on `get_import_file_name_if_exists`?

evap/staff/tools.py
```diff
@@ -7,11 +8,40 @@
...
+    with open(filename, "wb") as file:
+        for chunk in excel_file.chunks():
+            file.write(chunk)
+    excel_file.seek(0)
+
+
+def delete_import_file(user_id):
+    filename = generate_import_file_name(user_id)
+    if not os.path.isfile(filename):
+        return False
```
karyon: return type is never used. if you want to return nothing, simply use `return`

evap/staff/views.py
```diff
@@ -311,27 +313,51 @@ def semester_import(request, semester_id):
...
+
+        if test_run and excel_form_valid:
+            excel_file = excel_form.cleaned_data['excel_file']
+            warnings, errors = EnrollmentImporter.process(request, excel_file, semester, None, None, test_run)
+            if not errors:
+                save_import_file(excel_file, str(request.user.id))
+                return render(request, "staff_semester_import.html", dict(semester=semester,
+                    excel_form=excel_form, warnings=warnings, errors=errors, test_passed=True))
+            else:
+                delete_import_file(str(request.user.id))
```
karyon: why delete it, i think it hasn't been saved at all?

evap/staff/tools.py
```diff
@@ -7,11 +8,40 @@
...
+    os.remove(filename)
```
karyon: check out http://stackoverflow.com/a/10840586/4312944

evap/evaluation/tests/test_misc.py
```diff
@@ -29,9 +29,13 @@ def test_sample_xls(self):
         original_user_count = UserProfile.objects.count()
 
         form = page.forms["semester-import-form"]
+        form["excel_file"] = (os.path.join(settings.BASE_DIR, "static", "sample.xls"),)
+        form.submit(name="operation", value="test")
+
+        page = self.app.get("/staff/semester/1/import", user='user')
```
He3lixxx: Changed

evap/staff/importers.py
```diff
@@ -86,18 +87,34 @@ def store_in_database(self, vote_start_date, vote_end_date, semester):
 
 
 class ExcelImporter(object):
+    W_NAME = 'name'
+    W_EMAIL = 'email'
+    W_DUPL = 'duplicate'
+    W_GENERAL = 'general'
+
     def __init__(self, request):
```
He3lixxx: Yes, I removed django messages as much as possible, they are not only used after redirecting the user when an import was successful.

evap/staff/importers.py
He3lixxx: Changed it so the importer is given the file's content and doesn't handle the possble difference of whether it's given an UploadedFile or a file path.

evap/staff/importers.py
```diff
@@ -197,23 +219,22 @@ def check_user_data_sanity(self):
...
                 warningstring += "<br> - {} ({} {} {}, {})".format(user_data.username, user_data.title or "", user_data.first_name, user_data.last_name, user_data.email)
                 warningstring += _(" (new)")
-                self.warnings.append(mark_safe(warningstring))
-
-    def show_errors_and_warnings(self):
-        for error in self.errors:
-            messages.error(self.request, error)
-        for warning in self.warnings:
-            messages.warning(self.request, warning)
+                self.warnings[self.W_DUPL].append(mark_safe(warningstring))
```
He3lixxx: Done

evap/staff/importers.py
```diff
@@ -197,23 +219,22 @@ def check_user_data_sanity(self):
...
 
 
 class EnrollmentImporter(ExcelImporter):
+    # extension of ExcelImporter.warnings + dictionary to translate to UI strings
```
He3lixxx: Changed

evap/staff/templates/staff_user_import.html
```diff
@@ -2,20 +2,58 @@
...
+                {% else %}
+                <button name="operation" value="test" type="submit" class="btn btn-default form-submit-btn">{% trans "Upload and Test" %}</button>
+                <div class="form-submit-btn-divider"></div>
+                <button name="operation" value="import" type="submit" class="btn btn-primary form-submit-btn">{% trans "Import previously uploaded file" %}</button>
+                {% endif %}
             </div>
         </div>
     </form>
-{% endblock %}
+    {% endblock %}
```
He3lixxx: Fixed

evap/staff/tools.py
He3lixxx: Fixed

evap/staff/tools.py
He3lixxx: Changed

evap/staff/tools.py
He3lixxx: Thanks

evap/staff/tools.py
He3lixxx: Changed

evap/staff/tools.py
```diff
@@ -7,11 +8,59 @@
...
+    except OSError:
+        pass
+
+
+def import_file_exists(user_id):
+    filename = generate_import_filename(user_id)
+    if os.path.isfile(filename):
+        return True
+    else:
+        return False
```
karyon: `return os.path.isfile(filename)` :)

karyon:
i should add that most of the comments in any of the views apply to all three views since they are so similar.

janno42:
please take care when merging with the changes made in #915

evap/staff/views.py
```diff
@@ -311,27 +313,51 @@ def semester_import(request, semester_id):
...
         test_run = operation == 'test'
-
-        # parse table
-        EnrollmentImporter.process(request, excel_file, semester, vote_start_date, vote_end_date, test_run)
-        if test_run:
-            return render(request, "staff_semester_import.html", dict(semester=semester, form=form))
-        return redirect('staff:semester_view', semester_id)
+        import_run = operation == 'import'
+
+        excel_form_valid = excel_form.is_valid()
```
He3lixxx: Fixed

evap/staff/views.py
```diff
@@ -311,27 +313,51 @@ def semester_import(request, semester_id):
...
+
+        if test_run and excel_form_valid:
```
He3lixxx: Removed

evap/staff/views.py
He3lixxx: Added a comment, this was to remove an old file that may still be on the hard drive so after the page reloads the import button will not be displayed anymore

evap/staff/views.py
```diff
@@ -311,27 +313,51 @@ def semester_import(request, semester_id):
...
+            excel_file = excel_form.cleaned_data['excel_file']
+            warnings, errors = EnrollmentImporter.process(request, excel_file, semester, None, None, test_run)
+            if not errors:
+                save_import_file(excel_file, str(request.user.id))
+                return render(request, "staff_semester_import.html", dict(semester=semester,
+                    excel_form=excel_form, warnings=warnings, errors=errors, test_passed=True))
+            else:
+                delete_import_file(str(request.user.id))
+                return render(request, "staff_semester_import.html", dict(semester=semester,
+                    excel_form=excel_form, warnings=warnings, errors=errors, test_passed=False))
```
He3lixxx: Changed

evap/staff/views.py
```diff
@@ -311,27 +313,51 @@ def semester_import(request, semester_id):
...
+
+        elif import_run:
+            import_filename = get_import_file_name_if_exists(str(request.user.id))
+            if import_filename:
```
He3lixxx: Changed

evap/staff/views.py
He3lixxx: Changed the code to this behaviour.

evap/staff/views.py
```diff
@@ -311,27 +313,51 @@ def semester_import(request, semester_id):
...
+                    delete_import_file(str(request.user.id))
+                    return redirect('staff:semester_view', semester_id)
+            else:
+                raise SuspiciousOperation("No test run performed previously.")
+
+    if get_import_file_name_if_exists(str(request.user.id)):
+        return render(request, "staff_semester_import.html", dict(errors=errors, warnings=warnings, semester=semester, excel_form=excel_form, test_passed=True))
     else:
-        return render(request, "staff_semester_import.html", dict(semester=semester, form=form))
+        return render(request, "staff_semester_import.html", dict(errors=errors, warnings=warnings, semester=semester, excel_form=excel_form, test_passed=False))
```
He3lixxx: Changed

evap/staff/views.py
He3lixxx: Fixed

evap/staff/tools.py
He3lixxx: Changed

evap/staff/views.py
He3lixxx: fixed

evap/staff/views.py
```diff
@@ -626,40 +652,65 @@ def course_participant_import(request, semester_id, course_id):
...
         test_run = operation == 'test'
+        import_run = operation == 'import'
+        copy_run = operation == 'copy'
+
+        if test_run and excel_form.is_valid():
+            excel_file = excel_form.cleaned_data['excel_file']
+            imported_users, warnings, errors = UserImporter.process(request, excel_file, test_run)
+            if not errors:
+                save_import_file(excel_file, str(request.user.id))
+                # messages.success(request, "%d Participants would be added to course %s" % (len(imported_users), course.name))
```
He3lixxx: The commented code was a result of the UserImporter.process method currently not returning a correct list if you're doing a test run, thus always resulting in the message "0 Users would be imported." I changed the comment from the code to an explanation and I will add a reference on Issue #724

evap/staff/views.py
He3lixxx: Changed the functions to use as much fallthroughs as possible.

evap/staff/views.py
```diff
@@ -626,40 +652,65 @@ def course_participant_import(request, semester_id, course_id):
...
+        if operation not in ('test', 'import', 'copy'):
+            raise SuspiciousOperation("Invalid POST operation")
 
         test_run = operation == 'test'
+        import_run = operation == 'import'
+        copy_run = operation == 'copy'
+
+        if test_run and excel_form.is_valid():
+            excel_file = excel_form.cleaned_data['excel_file']
+            imported_users, warnings, errors = UserImporter.process(request, excel_file, test_run)
```
He3lixxx: Changing the API would result in the need of always creating a WhateverImporter object so in the end, when rendering, you can use WhateverImporter.warnings etc., even when there was no test or import request. I don't know whether this would be beneficial. Fixed the underscore issue for now.

evap/staff/views.py
```diff
@@ -311,27 +313,49 @@ def semester_import(request, semester_id):
...
+            excel_form.excel_file_required = True
+            if excel_form.is_valid():
+                excel_file = excel_form.cleaned_data['excel_file']
+                file_content = excel_file.read()
+                success_messages, warnings, errors = EnrollmentImporter.process(file_content, semester, None, None, test_run)
+                test_passed = len(errors) == 0
+                if test_passed:
+                    save_import_file(excel_file, str(request.user.id))
+                else:
+                    delete_import_file(str(request.user.id))  # delete file that may exist from old upload
```
karyon: please do the string conversion inside those methods

evap/staff/views.py
```diff
@@ -311,27 +313,49 @@ def semester_import(request, semester_id):
...
+            excel_form.vote_dates_required = True
+            if excel_form.is_valid():
+                vote_start_date = excel_form.cleaned_data['vote_start_date']
+                vote_end_date = excel_form.cleaned_data['vote_end_date']
+                success_messages, warnings, __ = EnrollmentImporter.process(file_content, semester, vote_start_date, vote_end_date, test_run)
+                forward_messages(request, success_messages, warnings)
+                delete_import_file(str(request.user.id))
+                return redirect('staff:semester_view', semester_id)
+
+    test_passed = import_file_exists(str(request.user.id))
```
karyon: you set a test_passed further above, is this here necessary?

evap/staff/views.py
```diff
@@ -1037,21 +1080,44 @@ def user_create(request):
...
+
+        elif import_run:
+            file_content = get_import_file_content_if_exists(str(request.user.id))
+            if file_content:
+                __, success_messages, warnings, __ = UserImporter.process(file_content, test_run)
+                forward_messages(request, success_messages, warnings)
+                delete_import_file(str(request.user.id))
+                return redirect('staff:user_index')
+            else:
+                raise SuspiciousOperation("No test run performed previously.")
```
karyon: guard clause

evap/staff/views.py
```diff
@@ -1037,21 +1080,44 @@ def user_create(request):
...
-        return render(request, "staff_user_import.html", dict(form=form))
+            excel_form.excel_file_required = True
+            if excel_form.is_valid():
+                excel_file = excel_form.cleaned_data['excel_file']
+                file_content = excel_file.read()
+                __, success_messages, warnings, errors = UserImporter.process(file_content, test_run)
+                if not errors:
+                    save_import_file(excel_file, str(request.user.id))
+                else:
+                    delete_import_file(str(request.user.id))  # remove old file if still exists
```
karyon: to me, it would make more sense to do the delete directly below the `if test_run:` line, because that's the place where we know we don't need the old file. what do you think?

evap/staff/views.py
```diff
@@ -311,27 +313,49 @@ def semester_import(request, semester_id):
...
-        return redirect('staff:semester_view', semester_id)
-    else:
-        return render(request, "staff_semester_import.html", dict(semester=semester, form=form))
+            excel_form.excel_file_required = True
+            if excel_form.is_valid():
+                excel_file = excel_form.cleaned_data['excel_file']
+                file_content = excel_file.read()
+                success_messages, warnings, errors = EnrollmentImporter.process(file_content, semester, None, None, test_run)
+                test_passed = len(errors) == 0
+                if test_passed:
```
karyon: `if not errors` like you do it in the other methods?

evap/staff/templates/staff_message_rendering_template.html
```diff
@@ -0,0 +1,37 @@
...
+            </span>
+        </div>
+        <div id="{{ key }}" class="panel-collapse collapse in">
+            <div class="panel-body">
+                {% for warning in warninglist %}
+                <div class="alert alert-warning alert-dismissable">
+                    <button type="button" class="close" data-dismiss="alert" aria-hidden="true">&#215;</button>
+                    {{ warning }}
+                </div>
+                {% endfor %}
```
karyon: indentation

evap/staff/templates/staff_message_rendering_template.html
```diff
@@ -0,0 +1,37 @@
...
+            </div>
+        </div>
+    </div>
+    {% endif %}
```
karyon: indentation

evap/staff/templates/staff_semester_import.html
```diff
@@ -5,16 +5,27 @@
...
+                {% bootstrap_form excel_form layout='horizontal' %}
+            </div>
+            <div class="panel-footer">
+                {% if not test_passed %}
+                <button name="operation" value="test" type="submit" class="btn btn-primary form-submit-btn">{% trans "Upload and Test" %}</button>
+                {% else %}
+                <button name="operation" value="test" type="submit" class="btn btn-default form-submit-btn">{% trans "Upload and Test" %}</button>
+                <div class="form-submit-btn-divider"></div>
+                <button name="operation" value="import" type="submit" class="btn btn-primary form-submit-btn">{% trans "Import previously uploaded file" %}</button>
+                {% endif %}
```
karyon: indentation

evap/staff/templates/staff_user_import.html
```diff
@@ -5,16 +5,27 @@
...
+                {% bootstrap_form excel_form layout='horizontal' %}
+            </div>
+            <div class="panel-footer">
+                {% if not test_passed %}
+                <button name="operation" value="test" type="submit" class="btn btn-primary form-submit-btn">{% trans "Upload and Test" %}</button>
+                {% else %}
+                <button name="operation" value="test" type="submit" class="btn btn-default form-submit-btn">{% trans "Upload and Test" %}</button>
+                <div class="form-submit-btn-divider"></div>
+                <button name="operation" value="import" type="submit" class="btn btn-primary form-submit-btn">{% trans "Import previously uploaded file" %}</button>
+                {% endif %}
```
karyon: indentation

evap/staff/tests/test_views.py
```diff
@@ -108,6 +108,79 @@ def test_view_list_sorting(self):
...
+    url = "/staff/semester/1/import"
+    test_users = ["staff"]
+    filename_valid = os.path.join(settings.BASE_DIR, "staff/fixtures/test_enrollment_data.xls")
+    filename_invalid = os.path.join(settings.BASE_DIR, "staff/fixtures/invalid_enrollment_data.xls")
+
+    @classmethod
+    def setUpTestData(cls):
+        semester = mommy.make(Semester, pk=1)
+        mommy.make(UserProfile, username="staff", groups=[Group.objects.get(name="Staff")])
+        cls.course = mommy.make(Course, pk=1, semester=semester)
```
karyon: if you don't depend on the primary keys being a specific value, don't set them here.

i also saw that in @bjrne's PR. is that copied from somewhere? if yes, could you fix it there as well?

evap/staff/templates/staff_user_import.html
```diff
@@ -5,16 +5,27 @@
...
+                {% trans "Upload user data from excel file (" %}<a href="{% get_static_prefix %}sample_user.xls">{% trans "Sample File"%}</a>{% trans "). This will create all containing users." %}
+                </p>
+                {% csrf_token %}
+                {% bootstrap_form excel_form layout='horizontal' %}
+            </div>
+            <div class="panel-footer">
+                {% if not test_passed %}
+                <button name="operation" value="test" type="submit" class="btn btn-primary form-submit-btn">{% trans "Upload and Test" %}</button>
+                {% else %}
+                <button name="operation" value="test" type="submit" class="btn btn-default form-submit-btn">{% trans "Upload and Test" %}</button>
```
karyon: as far as i can see, the only difference is the button class. in that case, i'd restrict the if/else to just contain that class. just put all of that in one line (maybe the button text on a new line if it gets too ugly)

evap/staff/tools.py
```diff
@@ -7,12 +8,57 @@
...
+    return os.path.isfile(filename)
+
+
+def get_import_file_content_if_exists(user_id):
+    filename = generate_import_filename(user_id)
+    if os.path.isfile(filename):
+        with open(filename, "rb") as file:
+            return file.read()
+    else:
+        return None
```
karyon: guard clause

evap/staff/views.py
```diff
@@ -311,27 +313,47 @@ def semester_import(request, semester_id):
...
+                excel_file = excel_form.cleaned_data['excel_file']
+                file_content = excel_file.read()
+                success_messages, warnings, errors = EnrollmentImporter.process(file_content, semester, None, None, test_run)
+                if not errors:
+                    save_import_file(excel_file, request.user.id)
+
+        elif import_run:
+            file_content = get_import_file_content_if_exists(request.user.id)
+            if not file_content:
+                raise SuspiciousOperation("No test run performed previously.")
```
karyon: how about rausing the `SuspiciosOperation` inside `get_import_file_content_if_exists`?

evap/staff/views.py
```diff
@@ -628,40 +650,58 @@ def course_participant_import(request, semester_id, course_id):
     course = get_object_or_404(Course, id=course_id, semester=semester)
     raise_permission_denied_if_archived(course)
 
-    form = UserImportForm(request.POST or None, request.FILES or None)
+    excel_form = UserImportForm(request.POST or None, request.FILES or None)
+    copy_form = CourseParticipantCopyForm(request.POST or None, request.FILES or None)
```
karyon: the second parameter isn't necessary or is it?

evap/staff/views.py
```diff
@@ -628,40 +650,58 @@ def course_participant_import(request, semester_id, course_id):
...
             course.participants.add(*imported_users)
             messages.success(request, "%d Participants added to course %s" % (len(imported_users), course.name))
             return redirect('staff:semester_view', semester_id)
-    else:
-        return render(request, "staff_course_participant_import.html", dict(course=course, form=form, semester=semester))
+
+    test_passed = import_file_exists(request.user.id)
+    return render(request, "staff_course_participant_import.html", dict(
+        course=course, excel_form=excel_form, copy_form=copy_form, semester=semester, success_messages=success_messages,
+        warnings=warnings, errors=errors, participant_test_passed=test_passed,))
```
karyon: could you give the parameters here and in the other methods a consistent ordering? something like semester, course, forms, messages, test_passed

and there's a trailing comma here :)

He3lixxx: The trailing comma was intended (and missing on the other dicts, I admit :) ). I heard @janno42 support this style because the line won't be changed if you append another item in a new line (see http://stackoverflow.com/questions/11597901/why-does-python-allow-a-trailing-comma-in-list#11597911). Since the lines are already pretty full, the next item would probably be appended on the next line, so this would apply. What do you think?

evap/staff/tests/test_views.py
He3lixxx: I assumed the "1" in `url = '/staff/semester/1/import'` was this key. The `pk=1` part is all over the file and also not used / needed in some other places except for the link.
Reading through more of the code I juss realized that in the TestSemesterExportView class, there is no pk set for the semester and it's still accessed with .../semester/1/...
I'm new to this stuff, so is my assumption right or not? And if it's right, why is the TestSemesterExportView class working correctly?

evap/staff/templates/staff_user_import.html
He3lixxx: There's not just the changed button class, there's also two more lines in the else clause (one for the seperator and one for the import button), or am I misunderstanding you?

karyon: yeah, totally missed that.

evap/staff/tests/test_views.py
karyon: you assumption is right, I missed that the 1 is part of the url. TestSemesterExportView likely works by chance, as the first object of a model that will be created will usually (though is not guaranteed to) get the pk 1.

evap/staff/forms.py
```diff
@@ -49,10 +72,8 @@ def __init__(self, *args, **kwargs):
         self.fields['course'].choices = choices
 
     def clean(self):
-        if self.cleaned_data['course'] and self.cleaned_data['excel_file']:
-            raise ValidationError('Please select only one of course or Excel file.')
-        if not self.cleaned_data['course'] and not self.cleaned_data['excel_file']:
-            raise ValidationError('Please select either course or Excel File.')
+        if self.cleaned_data['course'] is None:
+            raise ValidationError("Please select a Course from the drop-down.")
```
He3lixxx: I'd agree but if I change it to use `required=True` it will give an error when using the excel import form even though is_valid is never called on the participant copy form. So going to the course participant import view, selecting an excel file and clicking on "Upload" will cause the upload and test to pass, but after the page reloads it will say "This field is required" on the course selection, which is distracting. Any ideas to circumvent that?

evap/staff/views.py
karyon: i'm in favor of trailing commas in lists where each element is on a new line. in the case at hand i don't really care, for consistency with the rest of the code i would tend to omit it.

evap/staff/tests/test_views.py
He3lixxx: Alright, I'll look over the file and try to reduce it if possible and add it where necessary.

evap/staff/views.py
He3lixxx: Okay :)

evap/staff/forms.py
karyon: you would probably need to not pass the POST data into the form's constructor in case the other form got submitted. if it's too much of a hassle, leave it as it is.

He3lixxx: That would make the view's code larger again, so I think keeping it this way is better. I just found out that the unnecessary error message also exists in this commit, but i've fixed it now.

evap/staff/forms.py
```diff
@@ -26,19 +26,44 @@ def disable_all_fields(form):
...
+    excel_file = forms.FileField(label=_("Excel file"), required=False)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.excel_file_required = False
+        self.vote_dates_required = False
+
+    def clean(self):
+        if self.excel_file_required and self.cleaned_data['excel_file'] is None:
+            raise ValidationError("Please select an Excel file.")
```
janno42: translation

evap/staff/forms.py
```diff
@@ -26,19 +26,44 @@ def disable_all_fields(form):
...
+        if self.vote_dates_required:
+            if self.cleaned_data['vote_start_date'] is None or self.cleaned_data['vote_end_date'] is None:
+                raise ValidationError("Please enter an evaluation period.")
```
janno42: translation

evap/staff/forms.py
```diff
@@ -26,19 +26,44 @@ def disable_all_fields(form):
...
 class UserImportForm(forms.Form):
     excel_file = forms.FileField(label=_("Import from Excel file"), required=False)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.excel_file_required = False
+
+    def clean(self):
+        if self.excel_file_required and self.cleaned_data['excel_file'] is None:
+            raise ValidationError("Please select an Excel file.")
```
janno42: translation

evap/staff/forms.py
```diff
@@ -49,10 +74,8 @@ def __init__(self, *args, **kwargs):
         self.fields['course'].choices = choices
 
     def clean(self):
-        if self.cleaned_data['course'] and self.cleaned_data['excel_file']:
-            raise ValidationError('Please select only one of course or Excel file.')
-        if not self.cleaned_data['course'] and not self.cleaned_data['excel_file']:
-            raise ValidationError('Please select either course or Excel File.')
+        if self.course_selection_required and self.cleaned_data['course'] is None:
+            raise ValidationError("Please select a Course from the drop-down.")
```
janno42: translation and please use `Please select a course from the dropdown menu.`

karyon:
also, if you could squash the last five commits into one, that would be nice.

evap/staff/importers.py
```diff
@@ -318,13 +315,13 @@ def process(cls, excel_file, semester, vote_start_date, vote_end_date, test_run)
             importer = cls()
             importer.read_book(excel_file)
             if importer.errors:
-                return importer.success_messages, importer.warnings, importer.errors
+                return importer.success_messages, dict(importer.warnings), importer.errors
```
He3lixxx: Iterating defaultdicts in a django template does not work (see https://code.djangoproject.com/ticket/16335). You have to convert them to normal dicts before. I could move the conversion to where the dict is passed to the template though. I thought it might be simpler and cleaner the way I did it since this approach creates more encapsulation (The "default"-part of the dict is not needed anywhere outside the importer).

karyon: okay, in that case i'd do the conversion where this thing is passed to the template.

karyon: nvm.

karyon:
thanks!

PR #912, opened 2017-02-06
Remove contents from copied forms, fixes #644

### LINKED ISSUE #644: "add another" copies last form's content
to reproduce: 
1. go and edit some course, choose a contributor name for the last (empty) contributor form, but no questionnaire.
2. save the form.
3. it reloads as choosing no questionnaire is an error.
4. click "add another". the newly added form is a copy of the last form, and as the last form now wasn't empty, the new form copied it's contents.

the problem is that the `extra` parameter in the inlineformsetfactory doesn't seem to add more extra forms on the reload.

jakobmache:


janno42:
in a button-group please also select the first option when resetting

evap/staff/templates/staff_course_type_index.html
```diff
@@ -50,6 +50,7 @@
                     </tr>
                 {% endfor %}
             </tbody>
+
```
janno42: why did you add this? ;)

PR #913, opened 2017-02-06
Fix for #875

scento:
This pull request should supersede the pull request #892. Deleting a GradeDocument now also deletes the corresponding file on disk. In addition to that an old document gets deleted when it is replaced with another one.

### DIFF
--- a/evap/grades/models.py
+++ b/evap/grades/models.py
@@ -48,13 +48,15 @@ class GradeDocument(models.Model, metaclass=LocalizeModelBase):
     def filename(self):
         return os.path.basename(self.file.name)
 
-@receiver(models.signals.post_delete, sender=GradeDocument)
-def deleteFileOnDelete(sender, instance, **kwargs):
+@receiver(models.signals.pre_delete, sender=GradeDocument)
+
+def delete_file_pre_delete(sender, instance, **kwargs):
     if instance.file:
         instance.file.delete(False)
 
 @receiver(models.signals.pre_save, sender=GradeDocument)
-def deleteFileOnChange(sender, instance, **kwargs):
+
+def delete_file_pre_save(sender, instance, **kwargs):
     if not instance.pk:
         return False
     try:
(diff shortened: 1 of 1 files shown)


evap/grades/models.py
```diff
@@ -47,6 +49,25 @@ def filename(self):
...
+
+
+@receiver(pre_save, sender=GradeDocument)
+def delete_file_pre_save(sender, instance, **kwargs):
+    if not instance.pk:
+        return False
+    try:
+        oldFile = GradeDocument.objects.get(pk=instance.pk).file
+    except GradeDocument.DoesNotExist:
+        return False
```
karyon: could you check whether this except is necessary?

evap/grades/models.py
```diff
@@ -47,6 +49,25 @@ def filename(self):
...
+@receiver(pre_delete, sender=GradeDocument)
+def delete_file_pre_delete(sender, instance, **kwargs):
+    if instance.file:
+        instance.file.delete(False)
+
+
+@receiver(pre_save, sender=GradeDocument)
+def delete_file_pre_save(sender, instance, **kwargs):
+    if not instance.pk:
+        return False
```
karyon: comment plz when this is the case

evap/grades/models.py
```diff
@@ -47,6 +49,25 @@ def filename(self):
...
+    try:
+        oldFile = GradeDocument.objects.get(pk=instance.pk).file
+    except GradeDocument.DoesNotExist:
+        return False
+    newFile = instance.file
+    if not oldFile == newFile:
+        oldFile.delete(False)
```
karyon: also a comment please that explains why this receiver exists.

evap/grades/models.py
scento: The try-except is not necessary.

scento: It is necessary for fixtures.

evap/grades/models.py
scento: Done.

karyon:
thanks!

PR #920, opened 2017-02-07
Reviewer user group

janno42:
Adds a new user group for reviewers, gives them access to results and reviewing pages, adds a menu entry and semester overviews for reviewing.

Reviewer semester view:
review


evap/results/views.py
```diff
@@ -41,7 +41,7 @@ def semester_detail(request, semester_id):
             for degree in course.degrees.all():
```
karyon: does can_user_see_course need adapting? it is used in this method here

evap/templates/base.html
```diff
@@ -92,6 +92,15 @@
                             </li>
                         {% endif %}
 
+                        {% if user.is_reviewer %}
```
karyon: staff users see this as well, right? isn't that a bit annoying?

janno42: i thought it would be nice to have that for staff users as well if they just want to review sth.

evap/results/views.py
janno42: hm, yes, that probably needs adapting.

janno42: fixed

janno42:
- reviewers now can only see published courses:
  - that was a side effect of separating the results from the reviewing process. i know what you mean though - do you have a good idea about how to show the unpublished courses? i don't like adding the to the results page.
- the redirection stuff in evaluation/views.py probably needs adapting
  - the main view for reviewers will be their student view, that's intended.

evap/templates/base.html
karyon: k

karyon:
regarding redirection: at least the redirection-after-login needs adapting, since reviewers would not get redirected to staff/semester/review.

also, i wanted to have the decision that the main view is the student view documented, but couldn't think of anything that looks good.


regarding published courses: what about enabling the staff semester view for them, and in there remove/disable everything that reviewers are not allowed to do? then they can still see the current status of the evaluation which is nice, and they can see all courses regardless of their state. actually, that removes the need for the additional views/menu items....

janno42:
- redirect
  - okay, now i know what you mean
- main view
  - well, it's not always the student view. if a reviewer is also a contributor this will be the main view. so the role as a reviewer should just not change anything about that.
- course results of unpublished courses:
  - i really like the clean and focused reviewer view. using the large staff view is not really something i would like to have. the easiest solution would probably be to show unpublished results on the results page for reviewers (and staff).

karyon:
here i hacked it:


it's not super-nice but quite okay i think. i pushed a commit into my fork. i like this solution because it matches my mental model of reviewers being restricted staff users better, and it's very little code and needs less maintenance. what do you think?

the alternative would likely be to show all courses in the results page for staff and reviewers as you said. maybe we should do that regardless of whether we use the reviewer or semester view? so staff users have a place where they see all the grades whether they are published or not. on the other hand, it might be of value to have that view match the view the students have... i' not sure.

karyon:
also i noticed the user form needs a new checkbox to make users reviewers :)

janno42:
please review again

PR #921, opened 2017-02-09
Fix results cache after user merge

### LINKED ISSUE #907: Refresh cache after merging users
When merging users and thereby affecting contributions, the cached results are not correct anymore. This can result in errors, e.g., when exporting results.

karyon:
Fixes #907.

I also moved all the results calculation to the results app, i thought that made sense. @janno42, what do you think should i also move staff.views.semester_export, semester_raw_export and semester_participation_export?


janno42:
keep the exports in the staff app :)

PR #915, opened 2017-02-06
Contributor import

bjrne:
You can now add contributors in the same way as participants, either from an existing course or from an excel file.

Fixes #898

PR #931, opened 2017-02-28
Contributor import

He3lixxx:
Applied changes from @bjrne 's pr (#915) to new importing system.
Tests for copying participants and contributors still missing (I have a list of 15 tests that should be implemented on the importers, I'd do that in another issue).
Fun fact: the webtest parser fails when there is a comment in the page. All forms after the comment will not be in the page.forms dict. Just cost me about an hour to figure out why the form was not there...


evap/staff/views.py
```diff
@@ -677,48 +678,91 @@ def course_participant_import(request, semester_id, course_id):
...
-            if excel_form.is_valid():
-                excel_file = excel_form.cleaned_data['excel_file']
-                file_content = excel_file.read()
-                # if on a test run, the process method will not return a list of users that would be
-                # imported so there is currently no way to show how many users would be imported.
-                __, success_messages, warnings, errors = UserImporter.process(file_content, test_run)
-                if not errors:
-                    save_import_file(excel_file, request.user.id)
+        result, success_messages, warnings, errors = helper_course_person_import(
+                request, excel_form, copy_form, course, semester_id)
```
karyon: i think i would inline that method again. it would be a different thing if we had two of these methods, but right now it's just one. of course the view gets quite massive, but now it makes things even worse by e.g. making that redirect weirdness necessary.

evap/staff/views.py
```diff
@@ -677,48 +678,91 @@ def course_participant_import(request, semester_id, course_id):
...
 
-        test_run = operation == 'test'
-        import_run = operation == 'import'
-        copy_run = operation == 'copy'
+        if 'participants' in operation:
+            excel_form = participant_excel_form
+            copy_form = participant_copy_form
+        else:                                                 # 'contributors' in operation
+            excel_form = contributor_excel_form
+            copy_form = contributor_copy_form
```
karyon: if you're creating the magic string with the one-line-if-else, it might make sense to use that here too.

evap/staff/views.py
```diff
@@ -1105,27 +1150,24 @@ def user_import(request):
...
                 if not errors:
-                    save_import_file(excel_file, request.user.id)
+                    save_import_file(excel_file, request.user.id, magic)
 
-        elif import_run:
-            file_content = get_import_file_content_or_raise(request.user.id)
-            __, success_messages, warnings, __ = UserImporter.process(file_content, test_run)
+        elif operation == 'import':
+            file_content = get_import_file_content_or_raise(request.user.id, magic)
+            __, success_messages, warnings, __ = UserImporter.process(file_content, False)
```
karyon: i think i would actually name that parameter, `test_run=False`, here and in the other views. the views are complex enough without looking up parameter names.

evap/staff/views.py
```diff
@@ -677,48 +678,91 @@ def course_participant_import(request, semester_id, course_id):
...
-            forward_messages(request, success_messages, warnings)
-            return redirect('staff:semester_view', semester_id)
-
-        elif copy_run:
-            copy_form.course_selection_required = True
-            if copy_form.is_valid():
-                import_course = copy_form.cleaned_data['course']
+        else:
+            helper_add_contributors(imported_users, course)
+            success_messages.append(("{} Contributors added to course {}").format(len(imported_users), course.name))
```
karyon: could you check how that would look like if you moved these 6 lines into a helper method, re-used that method in the `copy` branch of this view (you'd still need to determine `imported_users` outside the helper)?

evap/staff/views.py
```diff
@@ -677,48 +678,91 @@ def course_participant_import(request, semester_id, course_id):
...
+            if magic == 'participant':
                 imported_users = import_course.participants.all()
                 course.participants.add(*imported_users)
-                messages.success(request, "%d Participants added to course %s" % (len(imported_users), course.name))
-                return redirect('staff:semester_view', semester_id)
+                success_messages.append(("{} Participants added to course {}").format(len(imported_users), course.name))
+            else:
+                imported_users = UserProfile.objects.filter(contributions__course=import_course)
+                helper_add_contributors(imported_users, course)
+                success_messages.append(("{} Contributors added to course {}").format(len(imported_users), course.name))
```
karyon: all these messages are missing i18n

evap/staff/views.py
He3lixxx: I don't quite understand what you mean here. There are two assignments in each block here, collapsing this to one line would be very unreadable, but perhaps I'm just misunderstanding you.

evap/staff/urls.py
```diff
@@ -24,7 +24,7 @@
     url(r"^semester/(\d+)/course/(\d+)/edit$", views.course_edit, name="course_edit"),
     url(r"^semester/(\d+)/course/(\d+)/email$", views.course_email, name="course_email"),
     url(r"^semester/(\d+)/course/(\d+)/preview$", views.course_preview, name="course_preview"),
-    url(r"^semester/(\d+)/course/(\d+)/participant_import", views.course_participant_import, name="course_participant_import"),
+    url(r"^semester/(\d+)/course/(\d+)/person_import", views.course_person_import, name="course_person_import"),
```
janno42: person_import was my suggestion, so i'd keep it.

evap/staff/views.py
karyon:
```
magic = 'participant' if 'participants' in operation else 'contributor'
excel_form = participant_excel_form if 'participants' in operation else contributor_excel_form
copy_form = participant_copy_form if 'participants' in operation else contributor_copy_form
```

karyon:
I put in bjarne's commit so his contribution is properly credited. richard, if you pull this branch again, you'll likely get some weird messages because i forcepushed into it.

karyon:
@janno42 leaving it open because i didn't know whether you wanted to review. if you want to merge, please do not squashmerge this one.

PR #933, opened 2017-03-06
Iss812

### LINKED ISSUE #812: Add times to evaluation period
The evaluation period could be defined by days and time instead of just days like currently implemented. That would allow to configure a precisely timed start/end for the evaluation period.

For this, the cron job checking for the start of an evaluation has to run every minute and the documentation must be updated accordingly.

tomkellog:
fixed #812 (hopefully)


karyon:
Superseded by #944

PR #942, opened 2017-03-20
Multiple responsible contributors

### LINKED ISSUE #896: Multiple responsible lecturers
In some cases multiple lecturers are equally responsible for a course. What would we need to reflect that on the platform?

jakobmache:
Fix #896

### DIFF
--- a/evap/grades/templates/grades_semester_view.html
+++ b/evap/grades/templates/grades_semester_view.html
@@ -21,7 +21,7 @@
                     <thead>
                         <tr>
                             <th class="col-sm-5">{% trans "Name" %}</th>
-                            <th class="col-sm-3">{% trans "Responsible" %}</th>
+                            <th class="col-sm-3">{% trans "Responsibles" %}</th>
                             <th class="col-sm-1">{% trans "Evaluation completed" %}</th>
                             <th class="col-sm-1">{% trans "Midterm grade documents" %}</th>
                             <th class="col-sm-1">{% trans "Final grade documents" %}</th>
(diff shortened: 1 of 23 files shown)


karyon:
I had a look into the ordering thing, in the end the solution was trivial but i don't fully understand why it works but we have a test for that so i guess it's fine.

i searched for remaining places where `responsible_contributor` is accessed (using regex `responsible_contributor[^s]` and found two more, could you search and fix those as well? then this should be good to go.

karyon:
oh and there's one remaining landscape issue. maybe just rename the index parameter to i.

evap/staff/forms.py
```diff
@@ -471,8 +467,6 @@ def clean(self):
 
         if count_responsible < 1:
             raise forms.ValidationError(_('No responsible contributor found. Each course must have exactly one responsible contributor.'))
```
janno42: This message must be updated

evap/grades/templates/grades_semester_view.html
```diff
@@ -21,7 +21,7 @@ <h3 class="panel-title">{% trans "Courses" %}</h3>
                     <thead>
                         <tr>
                             <th class="col-sm-5">{% trans "Name" %}</th>
-                            <th class="col-sm-3">{% trans "Responsible" %}</th>
+                            <th class="col-sm-3">{% trans "Responsibles" %}</th>
```
janno42: please keep `Responsible`. i will update the translation to `Verantwortlich`.

evap/results/templates/results_course_detail.html
```diff
@@ -64,7 +64,7 @@
                     <tr>
                         <th class="col-xs-1">{% trans "Degree" %}</th>
                         <th class="col-xs-2">{% trans "Type" %}</th>
-                        <th class="col-xs-3">{% trans "Responsible" %}</th>
+                        <th class="col-xs-3">{% trans "Responsibles" %}</th>
```
janno42: same here

evap/results/templates/results_semester_detail.html
```diff
@@ -22,7 +22,7 @@
                         <tr>
                             <th class="col-sm-4">{% trans "Course" %}</th>
                             <th class="col-sm-1">{% trans "Type" %}</th>
-                            <th class="col-sm-3">{% trans "Responsible" %}</th>
+                            <th class="col-sm-3">{% trans "Responsibles" %}</th>
```
janno42: same here

evap/results/templates/results_semester_detail.html
```diff
@@ -77,7 +81,7 @@
...
```
janno42: same here

PR #944, opened 2017-03-21
Add times to evaluation period 

### LINKED ISSUE #812: Add times to evaluation period

tomkellog:
changed the date format for evaluation periods to include times and added corresponding datetimepicker with locale support

fixes #812


karyon:
could you rebase your branch on the current master? (or merge the master, at your option) also, the js libraries are missing.

karyon:
hey @tomkellog, i noticed you based your work not on the current fsr-itse master, but on your own master, which unfortunately is almost a year old... so this really does need a rebase :)

a minor UI thing i noticed is that when editing a course, the vote start and end date fields are blank for me. when clicking into them, they suddenly are filled correctly.

also, there are a bunch of test failures, which should give a good guidance on which other places need adapting.

will you join us on monday? if not, could you give us a rough ETA? also if you don't want to continue on this let us know.

karyon:
contained in #952

Issue #936, opened 2017-03-13
Too many enrollments warning for external user does not contain username


He3lixxx:
The "User has too many enrollments" warning should contain the username, which it currently doesn't if the user is external and the username is generated during the import.

The test for the warning should also be modified to test this behaviour then.

karyon:
fixed by #984

PR #948, opened 2017-03-27
Verbose user imports

### LINKED ISSUE #724: More verbose user imports

He3lixxx:
- user import: correct number and ID of imported users in test and import run
- participant and contributor import: warning if imported users are already participating / contributing to the course
- participant and contributor import: correct number and ID of added participants / contributors
- enrollment import: list of newly created users

fixes #724

### DIFF
--- a/evap/staff/importers.py
+++ b/evap/staff/importers.py
@@ -287,17 +287,17 @@ class EnrollmentImporter(ExcelImporter):
                 self.warnings[self.W_MANY].append(_("Warning: User {} has {} enrollments, which is a lot.").format(username, len(enrollments)))
 
     def write_enrollments_to_db(self, semester, vote_start_date, vote_end_date):
-        students_created = 0
-        responsibles_created = 0
+        students_created = []
+        responsibles_created = []
 
         with transaction.atomic():
             for user_data in self.users.values():
-                created = user_data.store_in_database()
+                __, created = user_data.store_in_database()
                 if created:
                     if user_data.is_responsible:
-                        responsibles_created += 1
+                        responsibles_created.append(user_data)
                     else:
-                        students_created += 1
+                        students_created.append(user_data)
             for course_data in self.courses.values():
                 course_data.store_in_database(vote_start_date, vote_end_date, semester)
 
@@ -306,8 +306,26 @@ class EnrollmentImporter(ExcelImporter):
                 student = UserProfile.objects.get(email=student_data.email)
                 course.participants.add(student)
 
-        self.success_messages.append(_("Successfully created {} course(s), {} student(s) and {} contributor(s).").format(
-            len(self.courses), students_created, responsibles_created))
+        msg = _("Successfully created {} course(s), {} student(s) and {} contributor(s):").format(
+            len(self.courses), len(students_created), len(responsibles_created))
+        for user in students_created + responsibles_created:
+            msg += "<br>"
+            msg += "{} {} ({})".format(user.first_name, user.last_name, user.username)
+        self.success_messages.append(mark_safe(msg))
+
+    def create_test_success_messages(self):
+        filtered_users = []
+
+        for user_data in self.users.values():
+            if not user_data.user_already_exists():
+                filtered_users.append(user_data)
+
+        self.success_messages.append(_("The test run showed no errors. No data was imported yet."))
+        msg = _("The import run will create {} courses and {} users:").format(len(self.courses), len(filtered_users))
+        for user in filtered_users:
+            msg += "<br>"
+            msg += "{} {} ({})".format(user.first_name, user.last_name, user.username)
+        self.success_messages.append(mark_safe(msg))
 
     @classmethod
     def process(cls, excel_content, semester, vote_start_date, vote_end_date, test_run):
@@ -338,9 +356,10 @@ class EnrollmentImporter(ExcelImporter):
             if importer.errors:
                 importer.errors.append(_("Errors occurred while parsing the input data. No data was imported."))
             elif test_run:
-                importer.success_messages.append(_("The test run showed no errors. No data was imported yet."))
+                importer.create_test_success_messages()
             else:
                 importer.write_enrollments_to_db(semester, vote_start_date, vote_end_date)
+
             return importer.success_messages, importer.warnings, importer.errors
         except Exception as e:
             importer.errors.append(_("Import finally aborted after exception: '%s'" % e))
--- a/evap/staff/importers.py
+++ b/evap/staff/importers.py
@@ -42,6 +42,9 @@ class UserData(CommonEqualityMixin):
             user.refresh_login_key()
         return user, created
 
+    def user_already_exists(self):
+        return UserProfile.objects.filter(username=self.username).exists()
+
     def validate(self):
         user = UserProfile()
         user.username = self.username
@@ -361,6 +364,7 @@ class UserImporter(ExcelImporter):
             occur because of the data already in the database.
         """
         new_participants = []
+        created_users = []
         with transaction.atomic():
             users_count = 0
             for (sheet, row), (user_data) in self.associations.items():
@@ -369,14 +373,42 @@ class UserImporter(ExcelImporter):
                     new_participants.append(user)
                     if created:
                         users_count += 1
+                        created_users.append(user)
 
                 except Exception as e:
                     self.errors.append(_("A problem occured while writing the entries to the database."
                                          " The original data location was row %(row)d of sheet '%(sheet)s'."
                                          " The error message has been: '%(error)s'") % dict(row=row+1, sheet=sheet, error=e))
                     raise
-        self.success_messages.append(_("Successfully created %(users)d user(s).") % dict(users=users_count))
+
+        msg = _("Successfully created {} user(s):").format(users_count)
+        for user in created_users:
+            msg += "<br>"
+            msg += ("{} {} ({})").format(user.first_name, user.last_name, user.username)
+        self.success_messages.append(mark_safe(msg))
         return new_participants
+
+    def get_user_profile_list(self):
+        new_participants = []
+        for user_data in self.users.values():
+            if UserProfile.objects.filter(username=user_data.username).exists():
+                new_participants.append(UserProfile.objects.get(username=user_data.username))
+
+        return new_participants
+
+    def create_test_success_messages(self):
+        self.success_messages.append(_("The test run showed no errors. No data was imported yet."))
+
+        filtered_users = []
+        for (sheet, row), (user_data) in self.associations.items():
+            if not user_data.user_already_exists():
+                filtered_users.append(user_data)
+
+        msg = _("{} user(s) would be created:").format(len(filtered_users))
+        for user in filtered_users:
+            msg += "<br>"
+            msg += ("{} {} ({})").format(user.first_name, user.last_name, user.username)
+        self.success_messages.append(mark_safe(msg))
 
     @classmethod
     def process(cls, excel_content, test_run):
@@ -405,8 +437,8 @@ class UserImporter(ExcelImporter):
                 importer.errors.append(_("Errors occurred while parsing the input data. No data was imported."))
                 return [], importer.success_messages, importer.warnings, importer.errors
             if test_run:
-                importer.success_messages.append(_("The test run showed no errors. No data was imported yet."))
-                return [], importer.success_messages, importer.warnings, importer.errors
+                importer.create_test_success_messages()
+                return importer.get_user_profile_list(), importer.success_messages, importer.warnings, importer.errors
             else:
                 return importer.save_users_to_db(), importer.success_messages, importer.warnings, importer.errors
 
--- a/evap/staff/importers.py
+++ b/evap/staff/importers.py
@@ -11,6 +11,14 @@ from evap.evaluation.models import Course, UserProfile, Degree, Contribution, Co
 from evap.evaluation.tools import is_external_email
 
 
+def create_user_list_string_for_message(users):
+    msg = ""
+    for user in users:
+        msg += "<br>"
+        msg += "{} {} ({})".format(user.first_name, user.last_name, user.username)
+    return msg
+
+
 # taken from https://stackoverflow.com/questions/390250/elegant-ways-to-support-equivalence-equality-in-python-classes
 class CommonEqualityMixin(object):
 
@@ -45,13 +53,17 @@ class UserData(CommonEqualityMixin):
     def user_already_exists(self):
         return UserProfile.objects.filter(username=self.username).exists()
 
-    def validate(self):
+    def create_user_profile(self):
         user = UserProfile()
         user.username = self.username
         user.first_name = self.first_name
         user.last_name = self.last_name
         user.email = self.email
         user.password = "asdf"  # clean_fields needs that...
+        return user
+
+    def validate(self):
+        user = self.create_user_profile()
         user.clean_fields()
 
 
@@ -308,23 +320,15 @@ class EnrollmentImporter(ExcelImporter):
 
         msg = _("Successfully created {} course(s), {} student(s) and {} contributor(s):").format(
             len(self.courses), len(students_created), len(responsibles_created))
-        for user in students_created + responsibles_created:
-            msg += "<br>"
-            msg += "{} {} ({})".format(user.first_name, user.last_name, user.username)
+        msg += create_user_list_string_for_message(students_created + responsibles_created)
         self.success_messages.append(mark_safe(msg))
 
     def create_test_success_messages(self):
-        filtered_users = []
-
-        for user_data in self.users.values():
-            if not user_data.user_already_exists():
-                filtered_users.append(user_data)
+        filtered_users = [user_data for user_data in self.users.values() if not user_data.user_already_exists()]
 
         self.success_messages.append(_("The test run showed no errors. No data was imported yet."))
         msg = _("The import run will create {} courses and {} users:").format(len(self.courses), len(filtered_users))
-        for user in filtered_users:
-            msg += "<br>"
-            msg += "{} {} ({})".format(user.first_name, user.last_name, user.username)
+        msg += create_user_list_string_for_message(filtered_users)
         self.success_messages.append(mark_safe(msg))
 
     @classmethod
@@ -385,13 +389,11 @@ class UserImporter(ExcelImporter):
         new_participants = []
         created_users = []
         with transaction.atomic():
-            users_count = 0
             for (sheet, row), (user_data) in self.associations.items():
                 try:
                     user, created = user_data.store_in_database()
                     new_participants.append(user)
                     if created:
-                        users_count += 1
                         created_users.append(user)
 
                 except Exception as e:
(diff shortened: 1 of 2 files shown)


evap/staff/importers.py
```diff
@@ -303,8 +306,26 @@ def write_enrollments_to_db(self, semester, vote_start_date, vote_end_date):
...
+        for user_data in self.users.values():
+            if not user_data.user_already_exists():
+                filtered_users.append(user_data)
+
+        self.success_messages.append(_("The test run showed no errors. No data was imported yet."))
+        msg = _("The import run will create {} courses and {} users:").format(len(self.courses), len(filtered_users))
+        for user in filtered_users:
+            msg += "<br>"
+            msg += "{} {} ({})".format(user.first_name, user.last_name, user.username)
+        self.success_messages.append(mark_safe(msg))
```
karyon: all these loops that assemble the user strings, could you put that into a method?

evap/staff/importers.py
```diff
@@ -303,8 +306,26 @@ def write_enrollments_to_db(self, semester, vote_start_date, vote_end_date):
...
                 course.participants.add(student)
 
-        self.success_messages.append(_("Successfully created {} course(s), {} student(s) and {} contributor(s).").format(
-            len(self.courses), students_created, responsibles_created))
+        msg = _("Successfully created {} course(s), {} student(s) and {} contributor(s):").format(
+            len(self.courses), len(students_created), len(responsibles_created))
+        for user in students_created + responsibles_created:
+            msg += "<br>"
+            msg += "{} {} ({})".format(user.first_name, user.last_name, user.username)
+        self.success_messages.append(mark_safe(msg))
```
karyon: any specific reason why this is not in an extra method? i think it's a bit confusing that the test run has a separate method, but this here does not

evap/staff/importers.py
```diff
@@ -303,8 +306,26 @@ def write_enrollments_to_db(self, semester, vote_start_date, vote_end_date):
...
+
+    def create_test_success_messages(self):
+        filtered_users = []
+
+        for user_data in self.users.values():
+            if not user_data.user_already_exists():
+                filtered_users.append(user_data)
```
karyon: filtered_users = [user_data for user_data in self.users.values() if not user_data.user_already_exists()]

evap/staff/importers.py
```diff
@@ -369,15 +392,43 @@ def save_users_to_db(self):
...
+                        created_users.append(user)
 
                 except Exception as e:
                     self.errors.append(_("A problem occured while writing the entries to the database."
                                          " The original data location was row %(row)d of sheet '%(sheet)s'."
                                          " The error message has been: '%(error)s'") % dict(row=row+1, sheet=sheet, error=e))
                     raise
-        self.success_messages.append(_("Successfully created %(users)d user(s).") % dict(users=users_count))
+
+        msg = _("Successfully created {} user(s):").format(users_count)
```
karyon: remove users_count and use len(created_users)?

evap/staff/views.py
```diff
@@ -732,14 +735,63 @@ def course_person_import(request, semester_id, course_id):
...
+
+    if already_related:
+        if import_type == "participant":
+            msg = _("The following {} user(s) are already course participants in course {}:").format(len(already_related), course.name)
+        else:  # import_type == "contributor"
+            msg = _("The following {} user(s) are already contributing to course {}:").format(len(already_related), course.name)
+
+        for participant in already_related:
+            msg += "<br>"
+            msg += ("{} {} ({})").format(participant.first_name, participant.last_name, participant.username)
```
karyon: dude how often did you copypaste that?! :)

karyon: i'm a bit confused why these messages are created here and not in the importer, any specific reasons for that?

evap/staff/views.py
```diff
@@ -691,16 +692,17 @@ def course_person_import(request, semester_id, course_id):
             if excel_form.is_valid():
                 excel_file = excel_form.cleaned_data['excel_file']
                 file_content = excel_file.read()
-                # if on a test run, the process method will not return a list of users that would be
-                # imported so there is currently no way to show how many users would be imported.
-                __, success_messages, warnings, errors = UserImporter.process(file_content, test_run=True)
+                imported_users, success_messages, warnings, errors = UserImporter.process(file_content, test_run=True)
```
karyon: variable name

evap/staff/views.py
```diff
@@ -691,16 +692,17 @@ def course_person_import(request, semester_id, course_id):
...
+                imported_users = helper_handle_already_related(import_type, course, imported_users, warnings[UserImporter.W_GENERAL])
```
karyon: different variable name

evap/staff/importers.py
```diff
@@ -369,15 +392,43 @@ def save_users_to_db(self):
...
+        self.success_messages.append(mark_safe(msg))
         return new_participants
 
+    def get_user_profile_list(self):
+        new_participants = []
+        for user_data in self.users.values():
+            if UserProfile.objects.filter(username=user_data.username).exists():
+                new_participants.append(UserProfile.objects.get(username=user_data.username))
+
+        return new_participants
```
karyon: check again, you probably need to create UserProfiles for users that are not in the database yet

evap/staff/views.py
```diff
@@ -691,16 +692,17 @@ def course_person_import(request, semester_id, course_id):
...
                 if not errors:
+                    success_messages.append(helper_create_test_success_message(import_type, course, imported_users))
                     save_import_file(excel_file, request.user.id, import_type)
 
         elif 'import' in operation:
             file_content = get_import_file_content_or_raise(request.user.id, import_type)
             imported_users, success_messages, warnings, __ = UserImporter.process(file_content, test_run=False)
             delete_import_file(request.user.id, import_type)
+            imported_users = helper_handle_already_related(import_type, course, imported_users, warnings[UserImporter.W_GENERAL])
```
karyon: maybe call that added_users? (here and above)

evap/staff/importers.py
```diff
@@ -369,15 +392,43 @@ def save_users_to_db(self):
...
+            msg += "<br>"
+            msg += ("{} {} ({})").format(user.first_name, user.last_name, user.username)
+        self.success_messages.append(mark_safe(msg))
         return new_participants
 
+    def get_user_profile_list(self):
+        new_participants = []
+        for user_data in self.users.values():
+            if UserProfile.objects.filter(username=user_data.username).exists():
+                new_participants.append(UserProfile.objects.get(username=user_data.username))
```
karyon: to avoid the two queries:
```
try:
    new_participant = UserProfile.objects.get(username=user_data.username)
except: 
    new_participant = UserProfile(...)
new_participants.append(new_participant)
```

evap/staff/views.py
```diff
@@ -732,14 +735,63 @@ def course_person_import(request, semester_id, course_id):
...
+    return mark_safe(msg)
+
+
+def helper_handle_already_related(import_type, course, user_list, warnings):
+    already_related = []
+
+    if import_type == "participant":
+        for new_participant in user_list:
+            if new_participant in course.participants.all():
+                already_related.append(new_participant)
```
karyon: would something like this work `already related = course.participants.all() & user_list` (& is intersection)?

evap/staff/views.py
```diff
@@ -732,14 +735,63 @@ def course_person_import(request, semester_id, course_id):
...
+    else:  # import_type == "contributor"
+        for new_contributor in user_list:
+            if Contribution.objects.filter(course=course, contributor=new_contributor).exists():
+                already_related.append(new_contributor)
```
karyon: same here

evap/staff/views.py
```diff
@@ -732,14 +735,63 @@ def course_person_import(request, semester_id, course_id):
...
+            msg = _("The following {} user(s) are already course participants in course {}:").format(len(already_related), course.name)
+        else:  # import_type == "contributor"
+            msg = _("The following {} user(s) are already contributing to course {}:").format(len(already_related), course.name)
+
+        for participant in already_related:
+            msg += "<br>"
+            msg += ("{} {} ({})").format(participant.first_name, participant.last_name, participant.username)
+        warnings.append(mark_safe(msg))
+
+    filtered_list = list(filter(lambda x: x not in already_related, user_list))
```
karyon: please test if `user_list - already_related` works

evap/staff/views.py
He3lixxx: Should work in the first case but since this is not an intersection between two lists I don't see how it could work here.

evap/staff/views.py
He3lixxx: Typed it every time, always thinking that maybe at some point, we could want different formattings for different warnings ;P

He3lixxx: The importer does not know about any of the participant or contributor stuff. It just handles the user import. All the course related importing is happening in the view then, so this is the innermost place where the warnings could be generated.

He3lixxx: Changed it so all the import stuff happens in importers.py in a new importer (PersonImporter)

evap/staff/views.py
He3lixxx: does not work

evap/staff/importers.py
```diff
@@ -42,13 +50,20 @@ def store_in_database(self):
             user.refresh_login_key()
         return user, created
 
-    def validate(self):
+    def user_already_exists(self):
+        return UserProfile.objects.filter(username=self.username).exists()
+
+    def create_user_profile(self):
```
karyon: `UserProfile.objects.create(...)` is the method to create objects in the database, so using `create` here is problematic i think. maybe construct, init or instantiate

evap/staff/importers.py
```diff
@@ -417,6 +470,93 @@ def process(cls, excel_content, test_run):
...
+            raise ValueError("Expected one of the two positional arguments: file_content or source_course")
+
+        importer.handle_already_related()
+
+        if test_run:
+            importer.create_test_success_messages()
+        else:
+            importer.create_relations()
+
+        return importer.success_messages, importer.warnings, importer.errors
```
karyon: every method in this class except this one branches on `if self.import_type == "participant":`, and the individual branches don't contain that much code. suggestion: do that branch exactly once and call one of two methods, e.g. `process_contributors` and `process_participants`, which have the code from the other methods in this class inlined, and using local variables instead of object attributes.

karyon: with the intention to reduce branching, method calls and state, all of which are sources of complexity and which (i think) are unnecessary here.

evap/staff/importers.py
```diff
@@ -417,6 +470,93 @@ def process(cls, excel_content, test_run):
...
+        if already_related:
+            if self.import_type == "participant":
+                msg = _("The following {} user(s) are already course participants in course {}:").format(len(already_related), self.course.name)
+            else:  # import_type == "contributor"
+                msg = _("The following {} user(s) are already contributing to course {}:").format(len(already_related), self.course.name)
+            msg += create_user_list_string_for_message(already_related)
+
+        # since the user profiles are not necessarily saved to the database, they are not guaranteed to have a pk yet which
+        # makes anything relying on hashes unusable here (for a faster list difference)
+        self.users_to_add = list(filter(lambda x: x not in already_related, self.user_list))
```
karyon: in other places we use list comprehensions (`[user for user in...`) for exactly the same purpose, so i'd suggest using that here as well.

evap/staff/importers.py
```diff
@@ -361,23 +387,50 @@ def save_users_to_db(self):
...
+        new_participants = []
+        for user_data in self.users.values():
+            try:
+                new_participant = UserProfile.objects.get(username=user_data.username)
+            except UserProfile.DoesNotExist:
+                new_participant = user_data.create_user_profile()
+
+            new_participants.append(new_participant)
+
+        return new_participants
```
karyon: i would remove the two blank lines, but you don't have to :)

evap/staff/importers.py
```diff
@@ -361,23 +387,50 @@ def save_users_to_db(self):
...
+
+    def create_test_success_messages(self):
+        self.success_messages.append(_("The test run showed no errors. No data was imported yet."))
+
+        filtered_users = []
+        for (sheet, row), (user_data) in self.associations.items():
```
karyon: isn't there a list of user_data lying around that you can use instead of `associations`?

karyon: in fact, it looks like `associations` contains one entry for every line in the import, which would lead to duplicated messages.

karyon: can you match the structure of the other `create_test_success_messages` in line 326?

evap/staff/views.py
```diff
@@ -691,31 +691,22 @@ def course_person_import(request, semester_id, course_id):
...
             copy_form.course_selection_required = True
             if copy_form.is_valid():
                 import_course = copy_form.cleaned_data['course']
-                if import_type == 'participant':
-                    imported_users = import_course.participants.all()
-                else:
-                    imported_users = UserProfile.objects.filter(contributions__course=import_course)
-                # Import happens in this call:
-                success_messages.append(helper_person_import(imported_users, course, import_type))
+                success_messages, warnings, errors = PersonImporter.process(import_type, course, test_run=False, source_course=import_course)
```
karyon: i like this very much. the view definitely had too much going on so moving these out of it is a good thing in my opinion.

evap/staff/importers.py
He3lixxx: Changed it, but keep in mind that for user_data, no external user names are generated, so this extends Issue #936 then. I tried to circumvent that by using the associations.

evap/staff/importers.py
```diff
@@ -417,6 +470,93 @@ def process(cls, excel_content, test_run):
...
+        self.errors = []
+
+    def handle_already_related(self):
+        already_related = []
+
+        if self.import_type == "participant":
+            usernames = [user.username for user in self.user_list]
+            already_related = self.course.participants.filter(username__in=usernames).all()
+        else:  # import_type == "contributor"
+            already_related = Contribution.objects.filter(course=self.course, contributor__in=self.user_list).all()
```
He3lixxx: If I understand you correctly, then this code should fail (since a newly created UserProfile doesn't have an id, right?):
```
u = [UserProfile()]
already_related = Contribution.objects.filter(course=self.course, contributor__in=(self.user_list + u)).all()
```
It works fine although I can't find any guarantees on that in the documentation. Just appending an integer works too. A string fails, the error message being that it can't be casted to int. What are your thoughts? Is relying on "it's working right now" okay or do we need guarantees in the documentation?

He3lixxx:
I did all the requested changes, but take a look at the comments I answered, there are probably still changes needed.

evap/staff/views.py
He3lixxx: :)

evap/staff/importers.py
```diff
@@ -417,6 +464,83 @@ def process(cls, excel_content, test_run):
...
+        # makes anything relying on hashes unusable here (for a faster list difference)
+        users_to_add = [user for user in user_list if user not in already_related]
+
+        if test_run:
+            msg = _("{} participants would be added to the course {}:").format(len(users_to_add), course.name)
+            msg += create_user_list_string_for_message(users_to_add)
+        else:
+            course.participants.add(*users_to_add)
+            msg = _("{} Participants added to course {}").format(len(users_to_add), course.name)
+            msg += create_user_list_string_for_message(users_to_add)
```
karyon: could you check how it looks like if you pull the creation of msg out of the if/else and put the `would be` into another format parameter? (with `.format(..., "would be " if test_run else "", ...)` ). side benefit: the actual meat of this part (`course.participants.add(*users_to_add)`) stands out more clearly

karyon: same applies to `process_contributors`. if you don't do this, lowercase `p` in `Participants`, same for contributions below.

evap/staff/importers.py
```diff
@@ -417,6 +464,83 @@ def process(cls, excel_content, test_run):
...
+    @classmethod
+    def process(cls, import_type, course, test_run, file_content=None, source_course=None):
+        importer = cls()
+
+        if file_content is None and source_course is None:
+            raise ValueError("Expected one of the two positional arguments: file_content or source_course")
+
+        if import_type == 'participant':
+            importer.process_participants(course, test_run, file_content=file_content, source_course=source_course)
+        else:  # import_type == 'contributors'
```
karyon: `contributor`

evap/staff/importers.py
```diff
@@ -417,6 +464,83 @@ def process(cls, excel_content, test_run):
...
+            msg += create_user_list_string_for_message(users_to_add)
+
+        self.success_messages.append(mark_safe(msg))
+
+    @classmethod
+    def process(cls, import_type, course, test_run, file_content=None, source_course=None):
+        importer = cls()
+
+        if file_content is None and source_course is None:
+            raise ValueError("Expected one of the two positional arguments: file_content or source_course")
```
karyon: maybe make this a `==` instead of `and` put `exactly one` into the error message?

evap/staff/importers.py
```diff
@@ -417,6 +464,83 @@ def process(cls, excel_content, test_run):
...
+
+        if test_run:
+            msg = _("{} contributors would be added to the course {}:").format(len(users_to_add), course.name)
+            msg += create_user_list_string_for_message(users_to_add)
+        else:
+            for user in users_to_add:
+                order = Contribution.objects.filter(course=course).count()
+                contribution = Contribution(course=course, contributor=user, order=order)
+                contribution.save()
+                course.contributions.add(contribution)
```
karyon: check out https://docs.djangoproject.com/en/1.11/ref/models/querysets/#create. also, the last line shouldn't be necessary, because you already told the contribution it belongs to that course.

evap/staff/importers.py
```diff
@@ -417,6 +464,83 @@ def process(cls, excel_content, test_run):
...
+
+        self.success_messages.append(mark_safe(msg))
+
+    def process_contributors(self, course, test_run, file_content=None, source_course=None):
+        if file_content:
+            user_list, self.success_messages, self.warnings, self.errors = UserImporter.process(file_content, test_run)
+        elif source_course:
+            user_list = list(UserProfile.objects.filter(contributions__course=source_course))
+
+        already_related = Contribution.objects.filter(course=course, contributor__in=user_list).all()
```
karyon: already_related contains Contribution objects, that's not right, is it?

evap/staff/importers.py
```diff
@@ -417,6 +464,83 @@ def process(cls, excel_content, test_run):
...
+        self.errors = []
+
+    def process_participants(self, course, test_run, file_content=None, source_course=None):
+        if file_content:
+            user_list, self.success_messages, self.warnings, self.errors = UserImporter.process(file_content, test_run)
+        elif source_course:
+            user_list = list(source_course.participants.all())
+
+        usernames = [user.username for user in user_list]
+        already_related = course.participants.filter(username__in=usernames).all()
```
karyon: i noticed that each element of user_list is either in already_related or users_to_add, right? i thought it might be nice to make that explicit. [relevant SO post](http://stackoverflow.com/questions/949098/python-split-a-list-based-on-a-condition). i thought about this :
```
already_related = []
users_to_add = []
participants = course.participants
already_related.append(user) for user in user_list if user in participants else users_to_add.append(user)
```
but it's a bit weird.

alternatively:
```
already_related = [user for user in user_list if user in course_participants]
users_to_add = [user for user in user_list if user not in course_participants]
```
takes double the time but i don't mind actually. what do you think?

evap/staff/importers.py
karyon: i think this is fine then.

evap/staff/importers.py
karyon: actually.... it bothers me a bit that the caller knows which of the parameters is used, but we discard that knowledge and check for it again. how about two methods, `process_file_content` and `process_source_course`? each one would take only one of those two arguments. they would have some duplication, but this check here becomes unnecessary and the if/elif at the beginning of the other two process methods can be removed, they would then only get a user_list parameter. (and btw this would also fix the landscape issue ;) )

you can try that out if you want, but i'm not sure myself whether it's an improvement.

PR #954, opened 2017-05-16
Upgrade to Django 1.11

karyon:
Simplest way to update is probably `vagrant provision`.

A more or less functional change is the EmailField, all code should now put None instead of an empty string in there, previously that was converted by a custom model field. Also, i edited the initial migration, but i hope that breaks nothing because the change has nothing to do with the actual database scheme.


evap/urls.py
```diff
@@ -16,7 +16,7 @@
     url(r"^rewards/", include('evap.rewards.urls')),
     url(r"^grades/", include('evap.grades.urls')),
 
-    url(r"^logout$", django.contrib.auth.views.logout, {'next_page': "/"}, name="django-auth-logout"),
+    url(r"^logout$", django.contrib.auth.views.LogoutView.as_view(next_page="/"), {'next_page': "/"}, name="django-auth-logout"),
```
janno42: why do we need the `next_page` thing twice?

karyon: whoops, that's a leftover.

PR #962, opened 2017-05-26
Single results

janno42:
This fixes several bugs with Single Results.

### DIFF
--- a/evap/evaluation/tests/test_models.py
+++ b/evap/evaluation/tests/test_models.py
@@ -7,8 +7,8 @@ from django.core import mail
 
 from model_mommy import mommy
 
-from evap.evaluation.models import Course, UserProfile, Contribution, Semester, \
-                                   Questionnaire, CourseType, NotArchiveable, EmailTemplate
+from evap.evaluation.models import (Contribution, Course, CourseType, EmailTemplate, NotArchiveable, Questionnaire,
+                                    RatingAnswerCounter, Semester, UserProfile)
 from evap.results.tools import calculate_average_grades_and_deviation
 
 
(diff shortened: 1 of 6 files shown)


evap/staff/views.py
```diff
@@ -634,6 +634,9 @@ def course_delete(request):
 
     if not course.can_staff_delete:
         raise SuspiciousOperation("Deleting course not allowed")
+    if course.is_single_result:
+        for counter in RatingAnswerCounter.objects.filter(contribution__course=course):
+            counter.delete()
```
karyon: `RatingAnswerCounter.objects.filter(contribution__course=course).delete()`?

evap/results/templates/results_semester_detail.html
```diff
@@ -87,8 +87,13 @@
                     </thead>
                     <tbody>
                     {% for single_result, result in single_results %}
-                        <tr>
-                            <td>{{ single_result.name }}</td>
+                        <tr{% if single_result.state == 'reviewed' %} class="preview"{% endif %}>
```
karyon: space before the `{%` and not inside it please

karyon: also, i would prefer `state != 'published'`

evap/results/templates/results_semester_detail.html
```diff
@@ -87,8 +87,13 @@
...
+                            <td>
+                                {% if single_result.state == 'reviewed' %}
```
karyon: same here

janno42: check

evap/results/templates/results_semester_detail.html
janno42: check

evap/staff/views.py
janno42: check

janno42:
i would suggest some tests, too. would you like to write them? ;)

karyon:
no.

janno42:
tests. i don't like them but they test stuff. creating single results is ugly.

evap/evaluation/tests/test_models.py
```diff
@@ -116,6 +117,39 @@ def test_responsible_contributors_ordering(self):
...
+            "answer_3": 2,
+            "answer_4": 0,
+            "answer_5": 2,
+            "semester": course.semester.pk
+        }
+        form = SingleResultForm(form_data, instance=course)
+        self.assertTrue(form.is_valid())
+        form.save(user=mommy.make(UserProfile))
+
+        self.assertTrue(Course.objects.filter(pk=course.pk).exists())
```
karyon: can you replace all that by directly creating a singleResult with model mommy? this test doesn't need to use a form at all as far as i can see.

janno42: k

PR #965, opened 2017-05-26
Cronjob email error

### LINKED ISSUE #596: Warning if email address is missing when sending emails for course
When sending emails for a course (e.g. when publishing it), a warning must be shown if the email address is missing for a user to whom an email should be sent.
This can be a Django warning in the UI when sending the emails was triggered by a staff user, but must be handled differently when the situation occurrs when a cronjob triggers it. Maybe a message should be send to the admins.

janno42:
fix #596 
send email to admins when cronjob tries to send email to user without email address

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -1111,9 +1111,14 @@ class EmailTemplate(models.Model):
     def send_to_user(cls, user, template, subject_params, body_params, use_cc, request=None):
         if not user.email:
             warning_message = "{} has no email address defined. Could not send email.".format(user.username)
-            logger.warning(warning_message)
+            # If this method is triggered by a cronjob changing course states, the request is None.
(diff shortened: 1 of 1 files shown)


evap/evaluation/models.py
```diff
@@ -1111,9 +1111,11 @@ def send_to_users_in_courses(cls, template, courses, recipient_groups, use_cc, r
     def send_to_user(cls, user, template, subject_params, body_params, use_cc, request=None):
         if not user.email:
             warning_message = "{} has no email address defined. Could not send email.".format(user.username)
-            logger.warning(warning_message)
             if request is not None:
+                logger.warning(warning_message)
                 messages.warning(request, _(warning_message))
+            else:
+                logger.error(warning_message)
```
karyon: please add one or two comments explaining both what `if request is not None` means and why this is a warning or an error depending on that.

PR #967, opened 2017-05-29
Update datepicker localization files

### LINKED ISSUE #924: Wrong date format is used when selecting from datepicker
When loading the page, the correct date format for German is shown. But when you change it with the datepicker, it now uses the wrong date format.

PFischbeck:
This fixes #924. Because the localization files were outdated, the German date format was never applied.


karyon:
thanks!

Issue #926, opened 2017-02-19
Grade download button


janno42:
The grade download button on the students' index page should be removed for courses where the results have already been published without uploaded grade documents.

PR #968, opened 2017-05-29
Remove grade download button on the students' index page for courses …

JenniferStamm:
…where the results have already been published without uploaded grade documents.

This solves #926 .


karyon:
thanks!

PR #969, opened 2017-05-29
Remove comment IDs from UI

### LINKED ISSUE #925: Remove IDs from comment reviewing
The comment IDs are not helpful, and they even allow guessing which comments were made by the same participant, so let's remove them.

JenniferStamm:
Fixes #925 

This does not attribute the fact that comment IDs are still used in the url. So while there is no obvious display of the ids, the information is still retrievable.


karyon:
thanks!

PR #971, opened 2017-05-29
Importer bugfixes and tests

He3lixxx:
Bugfixes for small bugs (double spaces, messages not being returned)
Tests for warnings and more advances success messages that were added in #948

### DIFF
--- a/evap/staff/importers.py
+++ b/evap/staff/importers.py
@@ -478,12 +478,11 @@ class PersonImporter:
         if already_related:
             msg = _("The following {} user(s) are already course participants in course {}:").format(len(already_related), course.name)
             msg += create_user_list_string_for_message(already_related)
+            self.warnings[ExcelImporter.W_GENERAL].append(mark_safe(msg))
 
         if not test_run:
             course.participants.add(*users_to_add)
-            msg = _("{} participants added to the course {}:").format(len(users_to_add), course.name)
-        else:
-            msg = _("{} participants would be added to the course {}:").format(len(users_to_add), course.name)
+        msg = _("{} participants {}added to the course {}:").format(len(users_to_add), "would be " if test_run else "", course.name)
         msg += create_user_list_string_for_message(users_to_add)
 
         self.success_messages.append(mark_safe(msg))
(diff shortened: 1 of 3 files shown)


He3lixxx:
The order of the users in the success message given by the UserImporter is randomly shuffeled at the moment, so one test might fail randomly. I'll try to figure out why the order is not deterministic and change it so it is sorted alphabetically. Until then, don't pull please.

evap/staff/tests/test_importers.py
```diff
@@ -13,14 +13,58 @@ class TestUserImporter(TestCase):
...
+        self.assertEqual(original_users, list(UserProfile.objects.all()))
+
+    def test_test_and_notest_equality(self):
+        # success messages are supposed to be different
+        list_test, __, warnings_test, errors_test = UserImporter.process(self.valid_excel_content, test_run=True)
+        list_notest, __, warnings_notest, errors_notest = UserImporter.process(self.valid_excel_content, test_run=False)
+
+        notest_string_list = ["{} {}".format(user.full_name, user.email) for user in list_notest]
+        for user in list_test:
+            self.assertIn("{} {}".format(user.full_name, user.email), notest_string_list)
```
karyon: scratch that, what about making such a string_list for both `list_test`and `list_notest` and comparing these lists for equality?

evap/staff/tests/test_importers.py
```diff
@@ -13,14 +13,58 @@ class TestUserImporter(TestCase):
...
+        user_list, v1, v2, v3 = UserImporter.process(self.valid_excel_content, test_run=False)
+
+        self.assertIn("Successfully read sheet 'Users'.", v1)
+        self.assertIn('Successfully created 2 user(s):<br>Lucilia Manilium (lucilia.manilium)<br>Bastius Quid (bastius.quid.ext)', v1)
+        self.assertIn('Successfully read Excel file.', v1)
+        self.assertEqual(v2, {})
+        self.assertEqual(v3, [])
+
+        self.assertEqual(len(user_list), 2)
+        self.assertEqual(2 + original_user_count, UserProfile.objects.count())
```
karyon: original + 2?

karyon: also, swap the two arguments to `assertEqual`. otherwise it gets close to yoda expressions, `2 == some_variable`

evap/staff/importers.py
```diff
@@ -478,10 +478,11 @@ def process_participants(self, course, test_run, user_list):
         if already_related:
             msg = _("The following {} user(s) are already course participants in course {}:").format(len(already_related), course.name)
             msg += create_user_list_string_for_message(already_related)
+            self.warnings[ExcelImporter.W_GENERAL].append(mark_safe(msg))
```
He3lixxx: Nope, it didn't

PR #978, opened 2017-06-19
Dynamically changed sample files with correct institution domain (Issue #929)

### LINKED ISSUE #929: Update example files
The example files must be updated so that they can be imported without errors.

He3lixxx:
- View that delivers a changed sample file with the currently used institution domain
- Test whether the sample file is correct

fixes #929

### DIFF
--- a/evap/staff/tests/test_views.py
+++ b/evap/staff/tests/test_views.py
@@ -20,6 +20,31 @@ def helper_delete_all_import_files(user_id):
     for filename in glob.glob(file_filter):
         os.remove(filename)
 
+# Staff - Sample Files View
+class TestDownloadSampleXlsView(ViewTest):
+    test_users = ['staff']
+    url = '/staff/download_sample_xls/sample.xls'
+    email_placeholder = "institution.com"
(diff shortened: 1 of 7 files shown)


evap/staff/tests/test_views.py
```diff
@@ -20,6 +20,32 @@ def helper_delete_all_import_files(user_id):
...
+        found_institution_domain = False
+        book = xlrd.open_workbook(file_contents=page.body)
+        for sheet_index in range(book.nsheets):
+            sheet = book.sheet_by_index(sheet_index)
+            for row in range(sheet.nrows):
+                for col in range(sheet.ncols):
+                    value = sheet.cell(row, col).value
+                    self.assertNotIn(self.email_placeholder, value)
+                    if settings.INSTITUTION_EMAIL_DOMAINS[0] in value:
+                        found_institution_domain = True
```
karyon: can you use `for sheet in book.sheets()` and for `row in sheet.get_rows` and `for cell in row` (not sure about that one)

PR #982, opened 2017-07-03
Simplify staff_course_email form generation

### LINKED ISSUE #834: Error highlighting does not work in staff_course_email.html
in this commit https://github.com/fsr-itse/EvaP/commit/968d85c62a1de0ebd42d9ba7b04abbc03d20f9d4, the `form-group` class was removed from a div in `staff_course_email.html`. this prevents the new button to go on the next line, but broke the error highlighting when not selecting any recipient list.

but actually, i think all the error handling in there can be removed. it only adds a bit eye candy, and that's a lot of template code for a bit of eye candy. and it's not even a good solution, the proper one solution would be to not allow the user to submit invalid values at all through the html5 `required` class and checking with javascript whether any recipient group has been selected.

PFischbeck:
This fixes #834.
Should there still be additional client-side validation that a recipient group was selected?


karyon:
hey, do you plan to complete this?

PFischbeck:
Yeah, I am planning to work on it during today's hacking hours 👍

Issue #950, opened 2017-04-04
Inactive users


janno42:
Users should get an additional attribute, stating whether they are active or not (default: active).
The bulk deletion should get an additional option, that all users who would be deleted but can't (because they are participating in or contributing to a non-archived course or delegates or cc users of such users) will be marked inactive instead.

Inactive users are then not to be shown in participant or contributor selection fields, the importer should show a warning if an inactive user will be added to a course and an option should be added to the staff user list, whether inactive users should be shown (if so they will be marked with a respective label).

karyon:
django's AbstractBaseUser, from which our UserProfile inherits, already has an is_active field specifically for this purpose: https://docs.djangoproject.com/en/1.11/ref/contrib/auth/#django.contrib.auth.models.User.is_active. we should use that.

we could filter out inactive users by default similar to how we filter out the cronjob user by default: https://github.com/fsr-itse/EvaP/blob/master/evap/evaluation/models.py#L799. The UserProfileManager could get a `include_inactive_users` (possibly taking a boolean parameter if that simplifies things) method, analogue to `with_cronjob_user`.

He3lixxx:
assign please

PR #988, opened 2017-07-24
Inactive users (#950)

He3lixxx:


evap/staff/forms.py
```diff
@@ -566,6 +568,13 @@ def save(self, *args, **kw):
         else:
             self.instance.groups.remove(reviewer_group)
 
+        if self.cleaned_data.get('is_inactive'):
+            self.instance.is_active = False
+        else:
+            self.instance.is_active = True
```
karyon: look again

evap/staff/importers.py
```diff
@@ -221,17 +228,22 @@ def _create_user_name_collision_warning(self, user_data, users_with_same_names):
...
             try:
-                user = UserProfile.objects.get(username=user_data.username)
+                user = UserProfile.objects.with_inactive_users().get(username=user_data.username)
                 if user.email != user_data.email:
                     self.warnings[self.W_EMAIL].append(self._create_user_data_mismatch_warning(user, user_data))
                 if ((user.title is not None and user.title != user_data.title)
                         or user.first_name != user_data.first_name
                         or user.last_name != user_data.last_name):
                     self.warnings[self.W_NAME].append(self._create_user_data_mismatch_warning(user, user_data))
+                if user.is_active != True:
```
karyon: if not user.is_active

evap/staff/importers.py
```diff
@@ -538,17 +553,30 @@ def process_source_course(cls, import_type, course, test_run, source_course):
...
         if import_type == 'participant':
             user_list = list(source_course.participants.all())
+            if not test_run:
+                cls.make_users_active(user_list)
             importer.process_participants(course, test_run, user_list)
         else:  # import_type == 'contributor'
-            user_list = list(UserProfile.objects.filter(contributions__course=source_course))
+            user_list = list(UserProfile.objects.with_inactive_users().filter(contributions__course=source_course))
+            if not test_run:
+                cls.make_users_active(user_list)
```
karyon: why isn't this inside process_users?

evap/evaluation/models.py
```diff
@@ -810,10 +810,16 @@ class Meta:
...
+            .exclude(username=UserProfile.CRONJOB_USER_USERNAME)
+            .exclude(is_active=False)
+        )
 
     def with_cronjob_user(self):
-        return super().get_queryset()
+        return super().get_queryset().exclude(is_active=False)
+
+    def with_inactive_users(self):
+        return super().get_queryset().exclude(username=UserProfile.CRONJOB_USER_USERNAME)
```
karyon: as discussed, please do not filter by default.

karyon: following proposal:

this manager gets two methods, `cronjob_user`, returning the one user (remove the method in UserProfile then), and `exclude_inactive_users`, doing the filtering.

karyon:
also, check out the landscape issues :)

evap/staff/importers.py
```diff
@@ -324,6 +336,9 @@ def write_enrollments_to_db(self, semester, vote_start_datetime, vote_end_date):
                 course = Course.objects.get(semester=semester, name_de=course_data.name_de)
                 # This is safe because the user's email address is checked before in the importer (see #953)
                 student = UserProfile.objects.get(email=student_data.email)
+                if not student.is_active:
+                    student.is_active = True
+                    student.save()
```
karyon: isn't this sort of contained in one of the `self.make_users_active(user_list)` below?

evap/staff/importers.py
```diff
@@ -510,6 +526,8 @@ def process_contributors(self, course, test_run, user_list):
         users_to_add = [user for user in user_list if user not in already_related]
 
         if not test_run:
+            self.make_users_active(user_list)
```
karyon: do you need to do this at all? it says this here above
```
for user_data in self.users.values():
__, created = user_data.store_in_database()
```
and as far as i understand, this should set every user's `is_active` to True, or might any user from the excel file be missing?

evap/staff/views.py
```diff
@@ -1141,7 +1153,11 @@ def user_import(request):
 
 @staff_required
 def user_edit(request, user_id):
-    user = get_object_or_404(UserProfile, id=user_id)
+    try:
+        user = UserProfile.objects.get(id=user_id)
+    except UserProfile.DoesNotExist:
+        raise Http404('No UserProfile matches the given query.')
```
karyon: have you checked whether get_object_or_404 can be used again since we're overriding the default manager?

karyon:
@janno42 now would be the time to review this if you want :)

janno42:
looks good to me

Issue #963, opened 2017-05-26
Error highlighting in course forms is broken


janno42:
When editing a course and changing e.g. the name to an already existing one, the form returns with an error but only displays a generic "was not saved, please resolve errors below", but the actual name field is rendered in green and does not show an error.
This must be fixed for all forms.

He3lixxx:
Django's Form validation fails because the unique_together constraint of the Courses "name_de" or "name_en" fields together with the "semester" field is violated in that case. Since two fields are involved, Django does not assign this error to one of the fields. We don't render the non_field_errors, the error is not shown anywhere on the form. It would be nice to have the error assigned to the name_xy field involved since the semester field is not visible on the Form.

PR #993, opened 2017-08-07
Fixed non-field-bound errors not being shown

He3lixxx:
Fixed Issue #963 just as described there. I don't know if there are any more elegant ways to achieve this though. Also, I don't know if there is a guarantee on the order of the field names (perhaps it depends on the order in the ModelForm's `fields` member?), so maybe could just always use e.params["unique_check"][1].

### DIFF
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -180,13 +180,13 @@ class CourseForm(forms.ModelForm):
         super().validate_unique()
         # name_xy and semester are unique together. This will be treated as a non-field-error since two
         # fields are involved. Since we only show the name_xy field to the user, assign that error to this
-        # field.
+        # field. This hack is not documented, so it might be broken when you are reading this.
         for e in self.non_field_errors().as_data():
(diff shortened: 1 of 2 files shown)


evap/staff/forms.py
```diff
@@ -176,6 +176,18 @@ def __init__(self, *args, **kwargs):
...
+    def validate_unique(self):
+        super().validate_unique()
+        # name_xy and semester are unique together. This will be treated as a non-field-error since two
+        # fields are involved. Since we only show the name_xy field to the user, assign that error to this
+        # field.
+        for e in self.non_field_errors().as_data():
+            if e.code == "unique_together" and "unique_check" in e.params:
+                if "semester" in e.params["unique_check"]:
+                    field = e.params["unique_check"][0]
+                    field = e.params["unique_check"][1] if field == "semester" else field
```
karyon: these two lines are rather unclear. i would suggest to rename `field` to `name_field` and actually i would suggest to simply access the second field and be done with it.

PR #997, opened 2017-08-16
Add positive and negative yes-no questions

### LINKED ISSUE #345: New question type: Yes/No
The platform could use a new question type that allows to answer with only `yes` or `no`. This can be used for questions like "Would you recommend this lecture to future students?".

The answers to this type of questions will not influence the grade of the course and they should be displayed with a green/red bar in the design of the Likert question's results.

PFischbeck:
This fixes #345.

How it will look like in questionnaire creation:
questionnaire
How it will look like when voting:
student-form
How it will look like in the results (note the approval percentage on the right):
results
How it will look like in the export:
export


evap/results/templates/result_bar.html
```diff
@@ -6,15 +6,19 @@
     <div data-toggle="tooltip" data-placement="left"
         title="{% if questionnaire_warning or result.warning %}{% trans "Only a few participants answered this question." %}</br></br>{% endif %}
         {% for answer, count in result.counts.items %}
-            {% if result.question.is_likert_question %}{{ answer|likertname }}{% else %}{{ answer }}{% endif %}: {{ count }}/{{ result.total_count }} ({{ count|percentage_one_decimal:result.total_count }}){% if not forloop.last %}</br>{% endif %}
+            {% if result.question.is_likert_question %}{{ answer|likertname }}{% elif result.question.is_positive_yes_no_question %}{{ answer|positiveyesnoname }}{% elif result.question.is_negative_yes_no_question %}{{ answer|negativeyesnoname }}{% else %}{{ answer }}{% endif %}: {{ count }}/{{ result.total_count }} ({{ count|percentage_one_decimal:result.total_count }}){% if not forloop.last %}</br>{% endif %}
```
karyon: can you replace that if/elif with a new templatetag like this `{{ result.question|get_answer_name:answer }}` that handles all four answer types?

Issue #981, opened 2017-07-03
Only staff should be able to define responsibles


janno42:
Revert parts of #942 so that only staff users can define responsible contributors for a course. Editors should not be able to remove responsible contributors, to change the respective users (only their questionnaires and label) or to define new responsible contributors.

An information tooltip should be added to explain that responsible contributors can only be changed or added by the student representatives.

PR #1001, opened 2017-08-22
Editors can no longer change responsibles - Issue #981 - WIP

He3lixxx:
This only contains GUI changes and the old server side validation at the moment, I will add additional server side validation soon.
@janno42: Please confirm that the GUI handles the permissions correctly with these changes.
Code review is welcome as always.

I suspect that the validation was not done correctly in the past (see changes made in #942). The check was first implemented in 007203608438881ee2b01613612f0e30e6944a9d and then refactored in  98015c199afb9725ff8a4fede6bbeb9f6b38324e. We disabled the form's critical fields if the person is responsible which makes an editor unable to degrade a responsible person, but we didn't disable them if the form's instance is not responsible, so in theory, a privilege escalation from being editor to being responsible was possible here. The test introduced in 007203608438881ee2b01613612f0e30e6944a9d didn't test that. I guess we want this case covered too, so I'll need to do additional server side validation.
Additionally, I don't know whether a disabled field causes the form to be undeletable, so it may be necessary to also check for POST hacking when contributions are deleted.

edit: references #981

### DIFF
--- a/evap/contributor/tests/test_forms.py
+++ b/evap/contributor/tests/test_forms.py
@@ -66,7 +66,6 @@ class ContributionFormsetTests(TestCase):
     def test_editors_cannot_degrade_responsibles(self):
         course = mommy.make(Course)
         user = mommy.make(UserProfile)
-        questionnaire = mommy.make(Questionnaire, is_for_contributors=True)
         contribution = mommy.make(Contribution, course=course, contributor=user, responsible=True,
                                   can_edit=True, comment_visibility=Contribution.ALL_COMMENTS)
         InlineContributionFormset = inlineformset_factory(Course, Contribution, formset=ContributionFormSet, form=EditorContributionForm, extra=1)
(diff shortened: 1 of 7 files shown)


janno42:
ui seems to work as far as i tested it

evap/contributor/tests/test_forms.py
```diff
@@ -63,6 +63,115 @@ def test_staff_only(self):
...
+            "contributions-0-questionnaires": "{}".format(questionnaire.pk),
+            "contributions-0-responsibility": "RESPONSBILE",
+            "contributions-0-comment_visibility": "OWN",
+            "contributions-0-label": "",
+            "contributions-0-DELETE": "1",
+        }
+
+        formset = InlineContributionFormset(data, instance=course, can_edit_responsibles=False, form_kwargs={'course': course})
+
+        self.assertFalse(formset.is_valid())
```
karyon: i would prefer to first have some valid data, asserting that the form is valid, and then changing the necessary data to make the data invalid, and then asserting that the form is invalid.

there might be all kinds of other reasons for the data to be invalid, and right now it's not guaranteed that the test actually tests what you want it to test.

evap/staff/forms.py
```diff
@@ -391,10 +391,11 @@ def clean(self):
 
 
 class ContributionFormSet(AtLeastOneFormSet):
-    def __init__(self, data=None, *args, **kwargs):
+    def __init__(self, data=None, can_edit_responsibles=True, *args, **kwargs):
```
karyon: in the template this is called `can_change_responsible`, please make that consistent

evap/staff/forms.py
```diff
@@ -474,11 +475,21 @@ def clean(self):
...
 
+        if not self.can_edit_responsibles:
+            old_responsibles = list(self.instance.responsible_contributors)
+
+            if len(old_responsibles) != len(responsible_users):
+                raise ValidationError(_("You are not allowed to change responsible contributors"))
+
+            for old_responsible in old_responsibles:
+                if old_responsible not in responsible_users:
+                    raise ValidationError(_("You are not allowed to change responsible contributors"))
```
karyon: cast those to sets and compare those for equality

evap/evaluation/templates/responsibility_buttons.html
```diff
@@ -3,12 +3,16 @@
...
             {% trans "Able to edit the course to prepare the evaluation." as tooltip %}
         {% elif choice.data.value == "CONTRIBUTOR" %}
             {% trans "Default value. No special rights." as tooltip %}
         {% endif %}
-        {% include "choice_button.html" with formelement=form.responsibility choice=choice tooltip=tooltip %}
+        {% if choice.data.value != "RESPONSIBLE" or can_change_responsible %}
+            {% include "choice_button.html" with formelement=form.responsibility choice=choice tooltip=tooltip disabled=False %}
+        {% else %}
+            {% include "choice_button.html" with formelement=form.responsibility choice=choice tooltip=tooltip disabled=True %}
+        {% endif %}
```
karyon: can you invert that if? i find "disable this if it's responsible and the user cannot change responsibles" easier to parse than "enable this if it is not responsible and the user can change it"

evap/contributor/tests/test_forms.py
```diff
@@ -63,6 +63,145 @@ def test_staff_only(self):
...
+            "contributions-0-label": "",
+            "contributions-0-DELETE": "",
+        }
+        formset = InlineContributionFormset(data, instance=course, can_change_responsible=False, form_kwargs={'course': course})
+        self.assertTrue(formset.is_valid())
+
+        data["contributions-0-responsibility"] = "EDITOR"
+        formset = InlineContributionFormset(data, instance=course, can_change_responsible=False, form_kwargs={'course': course})
+        self.assertTrue(formset.is_valid())
+        self.assertTrue(formset.forms[0].fields["responsibility"].disabled)
```
karyon: please assert that it's not disabled in the first run, and maybe drop a comment that for some reason, django does stuff differently in this case and you check for the disabledness therefore.

PR #1002, opened 2017-08-28
Use UUIDs for random comment ids

### LINKED ISSUE #970: Random comment IDs
While looking into #925, we noticed it would actually be pretty easy to assign random IDs to comments, mostly eliminating the chance to find multiple comments created by the same user.

It might be as easy as specifying a primary key as shown here: https://docs.djangoproject.com/en/1.11/ref/models/fields/#uuidfield and creating a corresponding migration.

PFischbeck:
This fixes #970.


karyon:
@PFischbeck, i opened https://code.djangoproject.com/ticket/28541#ticket and https://code.djangoproject.com/ticket/28542#ticket.

karyon:
so we either use postgres for testing, both local and on travis, or wait on a fix from django (nobody knows if/when that will happen), or provide that fix ourselves (which might be complicated)

@PFischbeck do you want to either switch the travis setup to postgres or dig into django to fix this? otherwise i would take over (and probably just do the switch to postgres).

PFischbeck:
@karyon I would prefer the test setup switch for now. You can take over, otherwise I will do it on Monday.

.travis.yml
```diff
@@ -1,8 +1,12 @@
 language: python
 python:
   - "3.4"
+services:
+- postgresql
+before_script:
+  - psql -c 'create database evap;' -U postgres
 install:
-  - npm install -g less
+  - pip install psycopg2
```
karyon: lets just put that into requirements.txt and remove it from here and the vagrant setup.

karyon:
also, if you could rebase the merge commits out of the branch, that would be cool. might be a bit hard now that you have two of those with commits in between...

karyon:
master got a new migration, sorry :( on a positive note, minimal_test_data is now gone.

karyon:
we just realized that besides switching over to UUIDs, we need to ensure that the ordering of comments is guaranteed to not be the order of submission. so either django/postgres have some guarantee in that direction or we specify the default ordering by UUID.

@PFischbeck please let us know as soon as you don't want to continue working on this so we can prevent unnecessary bitrot. i can also take care of the merge commits or the migration thing if you want.

PFischbeck:
@karyon I gave this another try. The commits are restructured, and the text anwers are ordered by UUID.
Unfortunately, some test cases are still failing, and I can't see why. Some of them should have failed before (the ones with "302 != 200"), and for some I don't have any explanation.
You can take over here, I'm out of ideas...

PR #1006, opened 2017-09-04
Student voting test

### LINKED ISSUE #856: Test student voting
From a quick search, the voting process has exactly one test, which is not much for the primary feature of the platform.

jakobmache:
Fixes #856


evap/student/tests/test_views.py
```diff
@@ -18,63 +19,126 @@ def setUp(self):
...
+
+        self.assertEqual(RatingAnswerCounter.objects.filter(question=self.general_grade_question).count(), 1)
+        self.assertEqual(RatingAnswerCounter.objects.get(question=self.general_grade_question).answer, 3)
+
+        self.assertEqual(RatingAnswerCounter.objects.filter(question=self.contributor_likert_question).count(), 2)
+        self.assertEqual(RatingAnswerCounter.objects.filter(question=self.contributor_likert_question)[0].answer, 4)
+        self.assertEqual(RatingAnswerCounter.objects.filter(question=self.contributor_likert_question)[1].answer, 2)
+
+        self.assertEqual(TextAnswer.objects.filter(question=self.general_text_question).count(), 2)
+        self.assertEqual(TextAnswer.objects.filter(question=self.contributor_text_question).count(), 4)
```
karyon: can you check that the answers belong to the right contribution? also, can you check the contents of some (not necessarily all text answers?

karyon:
oh i almost forgot, this was the last test using minimal_test_data.json. so, you have the honor to remove that file ;)

evap/student/tests/test_views.py
```diff
@@ -1,9 +1,13 @@
...
 from django.test.utils import override_settings
 from django.urls import reverse
 from model_mommy import mommy
 
-from evap.evaluation.models import UserProfile, Course, Questionnaire, Question, Contribution
+from evap.evaluation.models import UserProfile, Course, Questionnaire, Question, Contribution, TextAnswer, RatingAnswerCounter
 from evap.evaluation.tests.tools import WebTest, ViewTest
+from evap.student.tools import question_id
+
+import pdb
```
karyon: some leftover here :)

evap/student/tests/test_views.py
```diff
@@ -17,64 +21,127 @@ def setUp(self):
...
+        cls.general_grade_question = mommy.make(Question, questionnaire=cls.general_questionnaire, type="G")
+
+        cls.contribution1 = mommy.make(Contribution, contributor=cls.contributor1, questionnaires=[cls.contributor_questionnaire],
+                                       course=cls.course)
+        cls.contribution2 = mommy.make(Contribution, contributor=cls.contributor2, questionnaires=[cls.contributor_questionnaire],
+                                       course=cls.course)
+
+        cls.course.general_contribution.questionnaires.set([cls.general_questionnaire])
+
+        cls.url = cls.url + str(cls.course.pk)
```
karyon: just specify the course pk in the mommy call and write that into the url directly. not very pretty, but all the other tests do it this way too.

evap/student/tests/test_views.py
```diff
@@ -17,64 +21,127 @@ def setUp(self):
...
         form = page.forms["student-vote-form"]
-        form["question_17_2_3"] = "some text"
-        form["question_17_2_4"] = 1
-        form["question_17_2_5"] = 6
-        form["question_18_1_1"] = "some other text"
-        form["question_18_1_2"] = 1
-        form["question_19_1_1"] = "some more text"
-        form["question_19_1_2"] = 1
-        form["question_20_1_1"] = "and the last text"
+        self.fill_form(form, False)
```
karyon: here and in the other calls to fill_form i would specify the parameter name `fill_complete=False`, since the code here is then easier to understand without looking at the fill_form method.

evap/student/forms.py
```diff
@@ -47,9 +46,14 @@ def __init__(self, *args, contribution, questionnaire, **kwargs):
...
-            identifier = make_form_identifier(contribution,
-                                              questionnaire,
-                                              question)
+            identifier = question_id(contribution,
+                                     questionnaire,
+                                     question)
+
+            identifier = question_id(self.contribution,
+                                     self.questionnaire,
+                                     question)
```
karyon: you're assigning identifier two times?

evap/student/tests/test_views.py
```diff
@@ -72,7 +68,7 @@ def test_incomplete_form(self):
         """
         page = self.get_assert_200(self.url, user=self.voting_user1.username)
         form = page.forms["student-vote-form"]
-        self.fill_form(form, False)
+        self.fill_form(form)
```
karyon: sorry for being unclear, i meant specifying the parameter when calling the method (`self.fill_form(form, fill_complete=False)`), not in the method definition.

PR #1012, opened 2017-09-12
Remove tooltip transparency

MrSerth:
Removes the tooltip transparency, because some tooltips are more difficult to read when they appear over the navbar.

opacity 0.9:
opacity 0 9

opacity 1.0:
opacity 1 0

### DIFF
--- a/evap/static/css/evap.scss
+++ b/evap/static/css/evap.scss
@@ -171,6 +171,10 @@ a.btn:not([href]):not(.disabled).btn-dark {
     line-height: 1;
 }
 
+.tooltip.show{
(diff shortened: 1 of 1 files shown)


evap/static/css/evap.scss
```diff
@@ -171,6 +171,10 @@ a.btn:not([href]):not(.disabled).btn-dark {
     line-height: 1;
 }
 
+.tooltip.show{
```
janno42: please add missing space before `{`

MrSerth:
Combined with #1011.

PR #1020, opened 2017-09-25
Reused login keys if they are still valid

### LINKED ISSUE #1018: Resend old login key if unused
When a user requests a new login key or receives one because of a course transaction, a new key should only be generated, if the previous one was not already used. An email should still be sent, but include the same key as before.

There should also be a test testing that behavior.

He3lixxx:
fixes #1018

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -1005,6 +1005,9 @@ class UserProfile(AbstractBaseUser, PermissionsMixin):
         return UserProfile.email_needs_login_key(self.email)
 
     def generate_login_key(self):
+        if self.login_key and self.login_key_valid_until > date.today():
+            return
+
         while True:
             key = random.randrange(0, UserProfile.MAX_LOGIN_KEY)
             if not UserProfile.objects.filter(login_key=key).exists():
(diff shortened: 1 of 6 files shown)


evap/evaluation/models.py
```diff
@@ -1005,6 +1005,9 @@ def needs_login_key(self):
         return UserProfile.email_needs_login_key(self.email)
 
     def generate_login_key(self):
```
karyon: maybe rename this to `ensure_login_key`? because now this does not always generate a new key.

evap/evaluation/tests/test_auth.py
```diff
@@ -38,3 +39,12 @@ def test_inactive_external_users_can_not_login(self):
         page = self.app.get(reverse("results:index") + "?loginkey=%s" % self.inactive_external_user.login_key).follow()
         self.assertContains(page, "Inactive users are not allowed to login")
         self.assertNotContains(page, "Logged in")
+
+    def test_login_key_resend_if_still_valid(self):
+        page = self.app.post("/", params={"submit_type": "new_key", "email": self.external_user.email}).follow()
+
+        new_key = UserProfile.objects.get(id=self.external_user.id).login_key
+        self.assertEqual(self.external_user.login_key, new_key)
```
karyon: could you make it more explicit you're comparing the original key to the new one? like `old_key = ...; page = ...; new_key = ...; self.assertEqual(...)`

PR #1024, opened 2017-10-01
Staff notification modals

### LINKED ISSUE #985: Easy staff notification for editors
Editors should be provided with an easy way to contact staff members in all places with texts like "in this case please contact the student representatives".
This should be implemented by placing a button next to these places which opens a modal similar to the general feedback modal, from which an email can be sent to the staff members. The message text could be initialized to already describe the general problem so that the user only has to specify what exactly needs to be changed.

This is necessary at the following places:

- [x] participant list on editor course edit pages
- [x] user settings
- [x] contributor list on editor course edit page (if a user does not already exist in the system) *(new)*

janno42:
fix #985
adds modals for contributors to easily notify staff about things that need to be changed

### DIFF
--- a/evap/contributor/templates/contributor_course_form.html
+++ b/evap/contributor/templates/contributor_course_form.html
@@ -74,7 +74,7 @@
 {% block modals %}
     {{ block.super }}
     {% trans "Approve course" as title %}
-    {% trans "Do you want to approve this course? This will allow the student representatives to proceed with the preparation, but you won't be able to make any further changes." as question %}
+    {% trans "Do you want to approve this course? This will allow the evaluation team to proceed with the preparation, but you won't be able to make any further changes." as question %}
     {% trans "Approve course" as action_text %}
     {% include "confirmation_modal.html" with modal_id="approveCourseModal" title=title question=question action_text=action_text btn_type="primary" %}
     <script type="text/javascript">
--- a/evap/contributor/templates/contributor_course_form.html
+++ b/evap/contributor/templates/contributor_course_form.html
@@ -22,8 +22,15 @@
         {% include 'contribution_formset.html' with formset=formset staff=False editable=editable%}
 
         <fieldset>
-            <legend>{% trans "Students" %}</legend>
-            <p>{% trans "These students should be participating in your course. If this data is not correct, e.g. when students are missing from the list, please contact the student representatives." %}</p>
+            <div class="d-flex">
+                <legend class="mr-auto">{% trans "Students" %}</legend>
+                <div>
+                    <a role="button" class="btn btn-sm btn-light" onclick="changeParticipantRequestModalShow();" id="changeParticipantRequestModalButton">
+                        {% trans "Request changes" %}
+                    </a>
+                </div>
+            </div>
+            <p>{% trans "These students should be participating in your course. Please let us know if this data is not correct." %}</p>
             <ul>
                 {% for p in course.participants.all|dictsort:"last_name" %}
                     <li>{{ p.full_name }}</li>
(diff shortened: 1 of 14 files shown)


janno42:
i have no idea what led to the test failure

evap/evaluation/templates/base.html
```diff
@@ -27,6 +27,23 @@
...
+            <div class="modal-content">
+                <div class="modal-header">
+                    <h5 class="modal-title" id="successMessageModalLabel">{% trans "Message sent" %}</h5>
+                </div>
+                <div class="modal-body">
+                    {% trans "Your message was successfully sent." %}
+                </div>
+            </div>
+        </div>
+    </div>
```
karyon: into `contact_modal.html`?

evap/evaluation/templates/base.html
```diff
@@ -27,6 +27,23 @@
...
+    {% if user.is_authenticated %}
+        {% trans "Feedback" as title %}
```
karyon: Contact?

evap/evaluation/templates/base.html
```diff
@@ -27,6 +27,23 @@
...
+        {% trans "You are welcome to submit feedback regarding the evaluation platform or the evaluation of specific courses. Please let us know how we can improve your experience on EvaP." as teaser %}
+        {% include "contact_modal.html" with modal_id="feedbackModal" user=request.user title=title teaser=teaser %}
```
karyon: contactModal?

evap/contributor/templates/contributor_settings.html
```diff
@@ -9,105 +9,97 @@
...
-        <div class="card card-submit-area text-center mb-3">
-            <div class="card-body">
-                <button type="submit" class="btn btn-primary">{% trans "Save changes" %}</button>
-            </div>
-        </div>
-    </form>
+{% block modals %}
+    {{ block.super }}
+    {% trans "Request user setting changes" as title %}
+    {% trans "Please tell us, which of your settings we should change." as teaser %}
```
karyon: no comma i think

evap/contributor/templates/contributor_course_form.html
```diff
@@ -76,6 +83,12 @@ <h5 class="modal-title" id="previewModalLabel">{% trans 'Preview' %}</h5>
             $('#course-form').submit();
         };
     </script>
+    {% blocktrans asvar title %}Request participant changes for {{ course }}{% endblocktrans %}
+    {% trans "Please tell us, what changes to the participants list we should make." as teaser %}
+    {% include "contact_modal.html" with modal_id="changeParticipantRequestModal" user=request.user title=title teaser=teaser %}
+    {% trans "Request user profile creation" as title %}
+    {% trans "Please tell us, which new user profiles we should create. We need the name and email for all new user profiles." as teaser %}
```
karyon: no commas again

evap/evaluation/templates/contribution_formset.html
```diff
@@ -1,7 +1,16 @@
 {% load evaluation_filters %}
 
 <fieldset>
-    <legend>{% trans "Contributors" %}</legend>
+    <div class="d-flex">
+        <legend class="mr-auto">{% trans "Contributors" %}</legend>
+        {% if not staff %}
+            <div>
+                <a role="button" class="btn btn-sm btn-light" onclick="createUserRequestModalShow();" id="createUserRequestModalButton">
+                    {% trans "Request creation of new user profile" %}
```
karyon: just `user`? if at all it would be an `account` i think

karyon:
regarding the test failure, check whether there are some primary keys in that test that might be different with postgres. why this should show up only with this PR i don't know...

evap/evaluation/templates/base.html
janno42: i had it there first but on some pages there are multiple contact modals and that would duplicate the `successMessageModal` on the page. probably not what we want.

evap/evaluation/templates/base.html
janno42: it's the feedback button. should we change that?

janno42:
the test probably fails because of the `form[2]` thing in `get_submit_assert_200`. will check tomorrow.

about the initialized message text: the modal title is part of the email that is sent, so the general topic is known to staff. the teaser text tells the user what to request, but yeah, they are free to formulate their specific text. i would assume that most users will send brief requests already, so i didn't want to include a predefined message. if it shows to be necessary we can add it later on.

evap/evaluation/templates/base.html
karyon: as discussed, let's do it anyways.

PR #1026, opened 2017-10-09
add warning if evaluation period is in last hours

### LINKED ISSUE #990: End of evaluation warning
When opening an evaluation questionnaire after the last day of evaluation (in the X hour period in which the evaluation is still possible), an information about the exact end of the evaluation should be shown on top of the page (or maybe in a modal).

maikz:
fixes #990 
+ add warning if evaluation period is in last hours
+ fix assertion and related tests

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -425,7 +425,7 @@ class Course(models.Model, metaclass=LocalizeModelBase):
     def time_left_for_evaluation(self):
         return self.vote_end_datetime - datetime.now()
 
-    def ends_soon(self):
+    def evaluation_ends_soon(self):
         return self.time_left_for_evaluation.total_seconds() < EVALUATION_END_WARNING_PERIOD * 3600
 
     @property
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -20,7 +20,7 @@ from django_fsm.signals import post_transition
 # see evaluation.meta for the use of Translate in this file
 from evap.evaluation.meta import LocalizeModelBase, Translate
 from evap.evaluation.tools import date_to_datetime
-from evap.settings import EVALUATION_END_OFFSET_HOURS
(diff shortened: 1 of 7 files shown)


evap/student/views.py
```diff
@@ -80,7 +81,11 @@ def vote(request, course_id):
             contributor_form_groups=contributor_form_groups,
             course=course,
             participants_warning=course.num_participants <= 5,
-            preview=False)
+            preview=False,
+            vote_end_datetime=course.vote_end_datetime,
```
karyon: is that variable used in the template?

evap/evaluation/models.py
```diff
@@ -262,7 +262,7 @@ def save(self, *args, **kw):
             self.contributions.create(contributor=None)
             del self.general_contribution  # invalidate cached property
 
-        assert self.vote_end_date >= self.vote_end_date
+        assert self.vote_end_date >= self.vote_start_datetime.date()
```
karyon: i would prefer such a change to be in a separate commit since it has little to do with the feature you implemented, but you don't need to change it for this PR.

evap/grades/tests.py
```diff
@@ -24,8 +24,8 @@ def setUpTestData(cls):
         cls.course = mommy.make(
             Course,
             name_en="Test",
-            vote_start_datetime=datetime.now() - timedelta(10),
-            vote_end_date=datetime.now() + timedelta(10),
+            vote_start_datetime=datetime.now() - timedelta(days=10),
+            vote_end_date=date.today() + timedelta(10),
```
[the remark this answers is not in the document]
maikz: days is the first argument. I think like this it's more specific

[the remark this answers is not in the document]
karyon: yeah, i was suggesting adding it here, since in the line above it is specified.

evap/student/views.py
karyon: yes it is.

karyon:
forgot to say thanks, so in case you're not there tomorrow: thanks! :)

PR #1033, opened 2017-11-20
Improve scrolling behavior on FAQ page (fix #539)

### LINKED ISSUE #539: Clean up jumping to content
The FAQ and the comments page make use of jumping to specific elements by using fragment identifiers in the URL.
Because the navbar hides some content in the topmost 50 px, this currently involves JavaScript which scrolls the page after some timeout. The timeout is needed, because Chrome doesn't seem to scroll otherwise.
The code used for this workaround should be cleaned up.

felixrindt:
Since fragment identifier navigation is only used on the FAQ page, the `scrollBy` mechanics were removed from `base.html`. 

In the FAQ template, a `scrollIntoView` call is used to bring an uncollapsed card into the middle of the screen. The `block: center` used is available in the current Chrome browser and in Firefox 58 (planned release in January 2018). 

To further simplify, it is now mandatory to use the `-q` suffix for question fragment identifiers.

### DIFF
--- a/evap/evaluation/templates/faq.html
+++ b/evap/evaluation/templates/faq.html
@@ -14,13 +14,13 @@
         {% for section in sections %}
             <h4 id="{{ section.id }}-s">{{ section.title }}</h4>
             {% for question in section.questions.all %}
-                <div id="card-{{ question.id }}" class="card{% if not forloop.last %} mb-1{% else %} mb-4{% endif %}">
+                <div id="{{ question.id }}-card" class="card{% if not forloop.last %} mb-1{% else %} mb-4{% endif %}">
                     <div class="card-header d-flex" role="tab" id="{{ question.id }}-q">
                         <a class="collapse-toggle collapsed" data-toggle="collapse" href="#collapse-{{ question.id }}" aria-expanded="false" aria-controls="collapse-{{ question.id }}">
                             {{ question.question }}
                         </a>
                     </div>
-                    <div id="collapse-{{ question.id }}" class="collapse" role="tabpanel" aria-labelledby="{{ question.id }}-q" data-parent="#accordion">
+                    <div id="{{ question.id }}-a" class="collapse" role="tabpanel" aria-labelledby="{{ question.id }}-q" data-parent="#accordion">
                         <div class="card-body">
                             {{ question.answer|safe }}
                         </div>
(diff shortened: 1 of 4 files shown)


evap/evaluation/templates/faq.html
```diff
@@ -14,7 +14,7 @@ <h3 class="mb-4">{% trans "These are the Frequently Asked Questions (FAQs)" %}</
         {% for section in sections %}
             <h4 id="{{ section.id }}-s">{{ section.title }}</h4>
             {% for question in section.questions.all %}
-                <div class="card{% if not forloop.last %} mb-1{% else %} mb-4{% endif %}">
+                <div id="card-{{ question.id }}" class="card{% if not forloop.last %} mb-1{% else %} mb-4{% endif %}">
```
karyon: how about we call this `{{ question.id }}-card` or something like that and rename the div with the answer `{{ question.id }}-a` to be more consistent??

evap/evaluation/templates/faq.html
```diff
@@ -33,19 +33,20 @@ <h4 id="{{ section.id }}-s">{{ section.title }}</h4>
...
-        type = 'q';
-        if(anchor.length > 1)
-            type = anchor[1];
+        const anchor = window.location.hash.replace("#", "").split('-');
+        const id = anchor[0];
+        const type = anchor[1];
 
         if(type == "q" && id != ""){
-            $("#collapse-" + id).collapse('show');
+            const collapse = $("#collapse-" + id);
```
karyon: with the div renamed as suggested above, maybe call this `answer_div` or similar?

evap/evaluation/templates/faq.html
```diff
@@ -33,19 +33,20 @@ <h4 id="{{ section.id }}-s">{{ section.title }}</h4>
...
             window.history.pushState('', id, '/faq#' + id + '-q');
+            collapse.collapse('show');
+            collapse.on('shown.bs.collapse', function () {
+                $("#card-" + id)[0].scrollIntoView({
+                    behavior: "smooth",
+                    block: "center"
+                });
+            });
```
karyon: how does this behave when clicking on a link with `-s`? the heading might be hidden again by the menubar, could you check?

felixrindt: It does indeed hide the heading behind the menu bar. Not removing the code from `base.html` would fix this.

karyon: yes, please try that out. although i'm wondering whether it is obvious enough which section the link pointed to, since there is no visual hint like there is when selecting a specific question...

karyon:
thanks!

evap/evaluation/templates/faq.html
```diff
@@ -47,6 +48,11 @@ <h4 id="{{ section.id }}-s">{{ section.title }}</h4>
                     block: "center"
                 });
             });
-        }
+		} else if(type == "s") {
+			// scroll section heading below navbar
+			setTimeout(function(){
```
felixrindt: I just checked. The timeout is still necessary for my Chrome installation.

PR #1034, opened 2017-11-28
Questionnaire sections

### LINKED ISSUE #975: Provide a grouping feature for questionnaires
In many events the evaluation consists of multiple sections (e.g. General Questions, Food, Contents etc.) leading to many questionnaires cluttering up the staff area. One way to better organize those questionnaires could be a grouping feature, which allows one to hide and seek many questionnaires at the same time.

Currently, we just put all questions in one (long) questionnaire - which makes organizing the different sections hard and also looks unpleasing to the users completing the questionnaire.

He3lixxx:
fixes #975

Work in progress, I still need to do tests, but I'd like @karyon to take a look at the code because I'm not sure whether this is the way we want to do it.

Especially for  results/views.py:81:
- Should we refactor the filter loops to methods?
- I made it two loops because imo it becomes unreadable when you put both filters in one loop. What do you think?
- What about the code duplication for the one-time-execution after the loop? Do you know any cool way to prevent that? We could use a dictionary, but I don't think that this would improve the code.
- Is there any possible use case where we would want two headings without any other questions between them? This is not possible with the current approach. (@janno42)

The same questions pretty much apply for results/exports.py:80


evap/results/exporters.py
```diff
@@ -77,6 +77,34 @@ def deviation_to_style(self, deviation, total=False):
             style_name += "_total"
         return style_name
 
+    def filter_text_and_heading_questions(self, questions):
+        filtered_questions = []
+        current_heading_question = None
+        questions_in_current_heading = []
+
+        for question in questions:
```
janno42: i would suggest adding a comment here that briefly explains what the following lines do

evap/results/templates/results_course_detail_questionnaires.html
```diff
@@ -12,7 +12,14 @@ <h4>{{ section.questionnaire.public_name }}</h4>
 <table class="table table-striped{% if not last %} mb-3{% endif %}">
     <tbody>
         {% for result in section.results %}
-            {% if result.question.is_rating_question %}
+            {% if result.question.is_heading_question %}
+    </tbody>
+</table>
+<h5{% if not forloop.first %} class="mt-3"{% endif %}> {{ result.question.text }} </h5>
```
janno42: please remove the spaces around `{{ result.question.text }}`

evap/results/templates/results_course_detail_questionnaires.html
```diff
@@ -12,7 +12,14 @@ <h4>{{ section.questionnaire.public_name }}</h4>
...
+<table class="table table-striped {% if not last %} mb-3{% endif %}">
```
janno42: please remove the space before `{% if not last %}`

evap/results/templates/results_course_detail_questionnaires.html
```diff
@@ -12,7 +12,14 @@ <h4>{{ section.questionnaire.public_name }}</h4>
...
+    <tbody>
+
```
janno42: empty line?

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -9,34 +9,42 @@
...
-                    {% else %}
-                        <td class="vote-text-answer">
-                            <div class="vote-inputs {% if preview %}preview{% endif %}">
-                                {{ field.errors }}
-                                <textarea id="{{ field.id_for_label }}" name="{{ field.name }}">{{ field.value|default_if_none:"" }}</textarea>
-                            </div>
+                {% if field|is_heading_field %}
+        </tbody>
+    </table>
+    <h5{% if not forloop.first %} class="mt-3"{% endif %}> {{ field.label }} </h5>
```
janno42: spaces

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -9,34 +9,42 @@
...
+                                    <label class="btn btn-sm btn-light vote-btn vote-btn-{{ choice.data.value }} d-flex{% if field.value == choice.data.value %} active{% endif %}{% if field.errors %} choice-error{% endif %}" name="{{ choice.data.name }}"{% if not preview %} onclick="selectedAnswer('{{ choice.data.name }}');"{% endif %}>
+                                        <input id="{{ choice.id_for_label }}" name="{{ choice.data.name }}" type="radio" value="{{ choice.data.value }}" autocomplete="off" {% if field.value == choice.data.value %}checked{% endif %} />
+                                        <span class="m-auto vote-btn-text">{{ choice.choice_label }} {{ choice.id }}</span>
+                                    </label>
+                                {% endfor %}
+                                </div>
+                            </td>
+                        {% else %}
+                            <td class="vote-text-answer">
+                                <div class="vote-inputs {% if preview %}preview{% endif %}">
```
janno42: please change to `vote-inputs{% if preview %} preview{% endif %}`

evap/results/exporters.py
```diff
@@ -77,6 +77,34 @@ def deviation_to_style(self, deviation, total=False):
...
+                if questions_in_current_heading:
+                    if current_heading_question:
+                        filtered_questions.append(current_heading_question)
+                    filtered_questions += questions_in_current_heading
+
+                current_heading_question = question
+                questions_in_current_heading = []
+
+            elif question.is_text_question:
+                continue
```
karyon: i would put that into a separate list comprehension

evap/results/exporters.py
```diff
@@ -77,6 +77,34 @@ def deviation_to_style(self, deviation, total=False):
...
+
+        for question in questions:
+            if question.is_heading_question:
+                if questions_in_current_heading:
+                    if current_heading_question:
+                        filtered_questions.append(current_heading_question)
+                    filtered_questions += questions_in_current_heading
+
+                current_heading_question = question
+                questions_in_current_heading = []
```
karyon: this can possibly be simplified by using a simple loop with a counter and checking the current element and the next element, if they are both heading questions, then do not use the first question.

evap/results/templates/results_course_detail_questionnaires.html
```diff
@@ -12,7 +12,14 @@ <h4>{{ section.questionnaire.public_name }}</h4>
 <table class="table table-striped{% if not last %} mb-3{% endif %}">
     <tbody>
         {% for result in section.results %}
-            {% if result.question.is_rating_question %}
+            {% if result.question.is_heading_question %}
+    </tbody>
+</table>
```
karyon: can you try out not reducing the indentation? or put some comments around this part so one can visually better understand what's going on?

evap/results/views.py
```diff
@@ -78,6 +78,31 @@ def course_detail(request, semester_id, course_id):
...
+
+                current_heading = result
+                results_in_current_heading = []
+            else:
+                results_in_current_heading.append(result)
+
+        if results_in_current_heading:
+            if current_heading:
+                results_to_store.append(current_heading)
+            results_to_store += results_in_current_heading
```
karyon: same as above

evap/results/templates/results_course_detail_questionnaires.html
```diff
@@ -12,7 +12,14 @@ <h4>{{ section.questionnaire.public_name }}</h4>
...
         {% for result in section.results %}
-            {% if result.question.is_rating_question %}
+            {% if result.question.is_heading_question %}
+                {# We want to start a new section here, so we close the table, render the heading and reopen the table #}
+                </tbody>
+                </table>
+                <h5{% if not forloop.first %} class="mt-3"{% endif %}>{{ result.question.text }}</h5>
+                <table class="table table-striped{% if not last %} mb-3{% endif %}">
+                <tbody>
+            {% elif result.question.is_rating_question %}
```
He3lixxx: @karyon: Do you like the indentation this way? I feel like there is no way to not make it look ugly or misleading...

He3lixxx: Besides that, all requested changes were done, so if you approve, we can merge this.

evap/results/exporters.py
```diff
@@ -77,6 +77,22 @@ def deviation_to_style(self, deviation, total=False):
...
 
+    def filter_text_and_heading_questions(self, questions):
+        # remove text questions:
+        questions = [question for question in questions if not question.is_text_question]
+
+        # remove heading questions if they have no "content" below them
+        filtered_questions = []
+        for index in range(len(questions)):
+            question = questions[index]
+            # filter out if there are no more questions or the next question is also a heading question
```
karyon: i'd move the comment one line below

evap/results/templates/results_course_detail_questionnaires.html
karyon: yeah that's probably as good as it will get.

janno42:
On the comment review page we could easily have multiple empty headings when some parts of the questionnaire don't contain text questions. So after thinking about it again I think we can just leave that page as it currently is, without the headings.

janno42:
much headings. very export. wow.

evap/staff/tests/test_views.py
```diff
@@ -665,6 +665,31 @@ def test_view_downloads_excel_file(self):
...
+        contribution = mommy.make(Contribution, course=course, questionnaires=[questionnaire], contributor=contributor)
+        mommy.make(RatingAnswerCounter, question=likert_question, contribution=contribution, answer=3, count=100)
+
+        page = self.app.get(self.url, user='staff')
+        form = page.forms["semester-export-form"]
+        form.set('form-0-selected_course_types', 'id_form-0-selected_course_types_0')
+        form.set('include_not_enough_answers', 'on')
+        response = form.submit()
+
+        workbook = xlrd.open_workbook(file_contents=response.content)
```
karyon: can you check whether it's reasonably easy to test the exporter directly without using the view?

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -9,34 +9,42 @@
...
-                            <div class="vote-inputs {% if preview %}preview{% endif %}">
-                                {{ field.errors }}
-                                <textarea id="{{ field.id_for_label }}" name="{{ field.name }}">{{ field.value|default_if_none:"" }}</textarea>
-                            </div>
+                {% if field|is_heading_field %}
+        </tbody>
+    </table>
+    <h5{% if not forloop.first %} class="mt-3"{% endif %}>{{ field.label }}</h5>
+    <table class="table table-striped">
+        <tbody>
```
karyon: can you unify the indentation style with the other template?

evap/staff/tests/test_views.py
He3lixxx: Yes it's possible using this approach:
```
        binary_content = BytesIO()
        ExcelExporter(course.semester).export(binary_content, [[course.type.id]], True, True)
        binary_content.seek(0)
        workbook = xlrd.open_workbook(file_contents=binary_content.read())
```

I'd move the new test function to the export tests. Should I also move `test_view_excel_file_sorted` there? It also just tests whether the exporter correclty uses `course.name` to sort the courses. The name is localized by django. I don't quite know how to elegantly tell it which language to use if I don't go through the view.

PR #1043, opened 2017-12-11
Dont use subjunctive after importing users #1038

### LINKED ISSUE #1038: Importer warnings wording
After importing courses for a semester, the warnings state that, e.g., an existing user *would* be overwritten with other data, although at this moment the data already *was* overwritten.

frederike-ramin:
closes #1038

### DIFF
--- a/evap/staff/importers.py
+++ b/evap/staff/importers.py
@@ -209,20 +209,21 @@ class ExcelImporter(object):
 
     @staticmethod
     def _create_user_data_mismatch_warning(user, user_data, test_run):
-        msg = "The existing user would be overwritten with the following data:" \
-            if test_run \
-            else "The existing user was overwritten with the following data:"
+        if test_run:
+            msg = "The existing user would be overwritten with the following data:"
+        else:
+            msg = "The existing user was overwritten with the following data:"
         return (mark_safe(_(msg) +
             "<br> - " + ExcelImporter._create_user_string(user) + _(" (existing)") +
             "<br> - " + ExcelImporter._create_user_string(user_data) + _(" (new)")))
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


evap/staff/importers.py
```diff
@@ -208,14 +208,20 @@ def _create_user_string(user):
...
             "<br> - " + ExcelImporter._create_user_string(user_data) + _(" (new)")))
 
     @staticmethod
-    def _create_user_inactive_warning(user):
-        return mark_safe(_("The following user is currently marked inactive and will be marked active upon importing:") + " "
+    def _create_user_inactive_warning(user, test_run):
+        msg = "The following user is currently marked inactive and will be marked active upon importing:" \
+            if test_run \
+            else "The following user was previously marked inactive and is now marked active upon importing:"
+        return mark_safe(_(msg) + " "
```
karyon: return statement on a single line

evap/staff/importers.py
```diff
@@ -208,14 +208,20 @@ def _create_user_string(user):
         return "{} ({} {} {}, {})".format(user.username, user.title or "", user.first_name, user.last_name, user.email or "")
 
     @staticmethod
-    def _create_user_data_mismatch_warning(user, user_data):
-        return (mark_safe(_("The existing user would be overwritten with the following data:") +
+    def _create_user_data_mismatch_warning(user, user_data, test_run):
+        msg = "The existing user would be overwritten with the following data:" \
+            if test_run \
+            else "The existing user was overwritten with the following data:"
```
karyon: i would prefer this as a regular if/else, since your syntax is intended for single line statements and you put that on multiple lines ;)

frederike-ramin: :'(

Issue #415, opened 2014-10-19
when removing a contributor and adding him again, an error occurs


karyon:
when removing a contributor on the course page, and then adding the same contributor again, an error is shown: "Duplicate contributor found. Each contributor should only be used once."

clicking on the "remove" link next to a contributor sets django's `delete` field and uses javascript to hide the table row. the actual deleting takes part during saving the form. thus, when adding the contributor again, a duplicate is created which is not visible to the user and which results in the error message described above.

PR #1058, opened 2017-12-13
Fix crash in contribution form validation, fixes #1057

### LINKED ISSUE #1057: Removing empty contributor field throws KeyError
When removing the empty contributor field at the end of a course form (as an editor or staff member) and saving the course, a `KeyError` is thrown.

karyon:
fixes #1057

next time this form crashes we rewrite the whole thing.

### DIFF
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -422,8 +422,10 @@ class ContributionFormSet(AtLeastOneFormSet):
         for form_with_errors in self.forms:
             if not form_with_errors.errors:
                 continue
+            if 'contributor' not in form_with_errors.cleaned_data:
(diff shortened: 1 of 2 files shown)


karyon:
as far as i can see, this must've been broken since adding the workaround for #415 in December 2014, although this seems hard to imagine...

besides the one described in #1057, there was another way to trigger the crash: delete any contributor and also change some data (e.g. select a questionnaire, NOT select a contributor) in the extra formset, and then click save.

janno42:
i'm pretty sure it wasn't broken since 2014 because we didn't have issues with that before and a lot of them today ;)
very strange.

PR #1063, opened 2017-12-23
Django20

### LINKED ISSUE #1035: Update to Django 2.0
Django 2.0 was released: https://docs.djangoproject.com/en/dev/releases/2.0/

* [x] run the tests and see what happens
* [x] run the tests with warnings enabled (-Wall?) and check for deprecations
* [x] (maybe re-export test_data so later diffs are cleaner)
* [x] https://docs.djangoproject.com/en/dev/releases/2.0/#simplified-url-routing-syntax, we could update to that
* [x] https://docs.djangoproject.com/en/dev/releases/2.0/#abstractuser-last-name-max-length-increased-to-150, we need to create that migration
* [x] Subclasses of AbstractBaseUser are no longer required to implement get_short_name() and get_full_name(), we can remove those

And: "A model instance’s primary key now appears in the default Model.\_\_str__() method, e.g. Question object (1)" :)

karyon:
fixes #1035 

some tests are disabled because of https://github.com/vandersonmota/model_mommy/issues/344. i suggest we wait a while and if that takes too long, merge anyway. or maybe workaround it.

### DIFF
--- a/evap/contributor/templates/contributor_course_form.html
+++ b/evap/contributor/templates/contributor_course_form.html
@@ -71,7 +71,7 @@
                         <a role="button" class="close" data-dismiss="modal" aria-label="Close"><span class="fa fa-times"></span></a>
                     </div>
                     <div class="modal-body">
-                        {{ preview_html|safe|escape }}
+                        {{ preview_html|safe }}
                     </div>
                     <div class="modal-footer">
                         <a role="button" class="btn btn-light" data-dismiss="modal">{% trans "Close" %}</a>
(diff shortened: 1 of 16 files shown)


janno42:
I just found that the course preview is broken after applying the changes.

preview

karyon:
removed the files. unfortunately i cannot reproduce the broken course preview, both on windows and the vagrant vm...

karyon:
fixed the broken preview as well. now we just wait for model_mommy to be fixed, don't know whether i'll do that myself anytime soon...

karyon:
i'm gonna merge this now. i [have fixed model_mommy](https://github.com/vandersonmota/model_mommy/pull/355) but received no reply for two weeks, our requirements now specify my fork until that is merged. a `pip3 install --user --upgrade -r requirements.txt` will be required, note the `--upgrade`.

Issue #883, opened 2016-12-12
Be more mobile  friendly


karyon:
First, all pages should be checked on some mobile device, there are likely some broken ones.

Then, we could re-design some parts. e.g. voting: questions could be displayed one after another, displaying the next one when a answer was selected

LinqLover:
Maybe notebooks should be displayed on top of the page on mobile:

![image](https://github.com/e-valuation/EvaP/assets/38782922/f6a8a50f-e062-4586-83a2-bbab0aff47af)

richardebeling:
@LinqLover what resolution / screen simulation setting was this screenshot taken at? There is move-to-top logic for the notebook with mobile users, we may want to make the trigger more aggressive though.

LinqLover:
@richardebeling Most mobile phones in landscape orientation, e.g. iPhone SE or Pixel 5:

![image](https://github.com/e-valuation/EvaP/assets/38782922/af7f9d20-c724-41fe-b325-e1fe102af494)

PR #1073, opened 2018-01-02
Background images

janno42:
How about adding some background images and getting rid of the plain white pages?

login
results


karyon:
i like the grey background. the login page is too colorful in my opinion. for instance, the text becomes a bit harder to read. i don't know yet how i feel about adding the pattern to EvaP's "branding" because it does not appear anywhere else apart from the webpage background (we had it on some poster but that does not count).

janno42:
maybe this less colorful version?

login

karyon:
i liked the lower-contrast background in the first version better. also in the first version i didn't even notice that the two boxes in the middle had a structured background as well, and in the second version i find it distracting, i would suggest using the first version for the boxes or remove the background for them altogether.

janno42:
login

He3lixxx:
I'm fine with both the colorful and the less colorful version of the background image. Also, I don't really care about the contrast on the background image.

I agree on removing the background from the boxes. Guides the focus more towards them.

So, in conclusion, I'm fine with version 3 but I wouldnt mind changes towards less color or less contrast.

janno42:
remember to add cards behind course edit form and maybe other places

janno42:
cards have now been added on many pages. ready for review.

evap/staff/templates/staff_course_operation.html
```diff
@@ -2,43 +2,51 @@
...
-            <fieldset id="email_form">
-                <div class="email-form">
-                    <div class="form-group d-flex">
-                        <div class="form-label">Email subject</div>
-                        <div class="form-field"><input class="form-control" name="email_subject" type="text" value="{{ email_template.subject}}" /></div>
-                    </div>
-                    <div class="form-group d-flex">
-                        <div class="form-label">Email body</div>
-                        <div class="form-field"><textarea class="form-control" cols="40" rows="10" name="email_body">{{ email_template.body }}</textarea></div>
+        <div class="card mb-3">
```
karyon: one more indentation level, but actually i don't really care, it's way too many levels anyways :)

janno42:
page content got a little narrower in all places where card have been added, but not much. in some places vertical distances are now a little bit more, but also not much. in general everything stays the same.

karyon:
2018-01-10 18_26_55-presenting succesfully - how to make an even better impression - evaluation - ev

the breadcrumb is about the average color of the background, maybe make it brighter?

also, some headings are on the structured background. i think that looks a bit weird. making the breadcrumb brighter might make this worse by enclosing the background-less heading with elements that have background. but giving it it's own background like the breadcrumb might look even worse, i'm not sure what to do here, maybe make the heading larger? or maybe just leave it as it is...

janno42:
the breadcrumbs background is now white. i actually like the heading being directly on the background. so if you don't mind too much i'd keep it that way.

capture

janno42:
(and by "some headings" i hope you mean all of them? i thought they're all the same.)

evap/staff/templates/staff_user_merge.html
```diff
@@ -11,308 +11,311 @@
 {% block content %}
     {{ block.super }}
     <h3>{% trans 'Merge users' %}</h3>
```
He3lixxx: This looks buggy on firefox 57:
grafik

evap/contributor/templates/contributor_course_form.html
```diff
@@ -17,26 +17,37 @@
     <form method="POST" class="form-horizontal multiselect-form select2form" id="course-form">
         {% csrf_token %}
 
-        {% include 'course_form_general_questions.html' with course=course course_form=form %}
-
-        {% include 'contribution_formset.html' with formset=formset staff=False editable=editable%}
+        <div class="card mb-3">
+            <div class="card-body">
+                {% include 'course_form_general_questions.html' with course=course course_form=form %}
```
He3lixxx: Are there any rules by which we decide whether a title for a card becomes part of it body or its heading? I find it a little inconsistent. Especially on the contributor and staff course edit page I feel like having some sort of line or other visual guidance that distincts the title from the content could make finding it easier. Compare the two attached screenshots.
grafik

grafik

evap/staff/templates/staff_user_merge.html
janno42: resolutions below 1200px are currently not fully supported. thoughts went only into contributor and student pages so far. would be part of #883.

He3lixxx: Hmm, strange, I'm pretty sure I had my browser close to 1920px wide when I took the screenshot - the screenshot is ~1400px wide, which is exactly how wide the merge cardbox is on my screen when I have the browser open fullscreen. I also counter checked against the current master, the page looked fine with that on the same browser window.

But: I just tried to replicate that and couldn't, so I guess, I messed up somewhere. #833 it is, then.

evap/contributor/templates/contributor_course_form.html
janno42: okay, card titles are also used in other places, where i introduced cards following the current bootstrap 4 default (e.g. user form, settings page, person import, staff index). would you prefer all of them to be unified?

He3lixxx: I guess unified looks more... unified? But I don't know yet which of the looks I'd prefer.
Since this was not unified before anyways and I only noticed on for the heading "Course Information" on the course edit page, I'll approve this and think about it some more. If I come up with a good idea, I'll create an issue just for that.

janno42: perfect, thanks. it won't be the last ui update ;)

PR #1079, opened 2018-01-10
Replace puppet with shell script

karyon:
pros:
- There is much less stuff in the repo
- look at the script, it's beautifully explicit
- The box is only 300MB instead of 600MB to download
- The finished VM is only 2GB instead of 3GB
- as a bonus, i added `git submodule update --init` to the provisioning script, so the remaining submodule (bootstrap) won't cause any issues. it seems a bit dirty doing that from within the VM, but i don't care.

cons:
- the secret key in the localsettings.py is hardcoded
- with older virtualbox versions, `vagrant up` aborts because the machine enters the state `unknown`. with a newer virtualbox, this still happened to me once at random out of maybe 10 tests. if that happens, one can simply run `vagrant provision`.
- requires new vagrant and virtualbox versions, the latter is not automatically checked

it would be cool if you both @janno42 and @He3lixxx could test this, ideally with a fresh checkout without `--recurse-submodules` to test that part as well. after the initial vagrant up, evap should be available on localhost:8001, and with a `./manage.py run` on 8000 as well, just as before.


karyon:
oh, and the progress bar of the last step got printed onto a few dozen lines like this
```
[.                                                                          ]
[..                                                                         ]
[...                                                                        ]
[....                                                                       ]
[.....                                                                      ]
[......                                                                     ]
[.......                                                                    ]
[........                                                                   ]
[.........                                                                  ]
[..........                                                                 ]
[...........                                                                ]
[............                                                               ]
[.............                                                              ]
[..............                                                             ]
[...............                                                            ]
[................                                                           ]
[.................                                                          ]
[..................                                                         ]
[...................                                                        ]
[....................                                                       ]
```
strangely, only in the last few attempts of mine, and i don't see how i could have caused that. did you see that as well?

janno42:
Yes, I had this as well. But I think it's not really a problem.

deployment/localsettings.py
```diff
@@ -3,11 +3,11 @@
...
-        'PASSWORD': 'evap',             # Not used with sqlite3.
+        'PASSWORD': 'evap',                     # Not used with sqlite3.
         'HOST': '127.0.0.1',                    # Set to empty string for localhost. Not used with sqlite3.
         'PORT': '',                             # Set to empty string for default. Not used with sqlite3.
         'CONN_MAX_AGE': 600,
     }
 }
 
-SECRET_KEY = "<%= @secret_key -%>"
+SECRET_KEY = "xvSFTFgwc8v5oYdB2F6oRVCShbZNAC"
```
He3lixxx: What do you think about leaving some kind of placeholder here and replacing it from the provision_vagrant_vm.sh script? We are already using sed in that file. If we leave this line like `SECRET_KEY = "${SECRET_KEY}"`, a call to sed like `sed -n -e "s/\${SECRET_KEY}/$(sudo head /dev/urandom | tr -dc A-Za-z0-9 | head -c 32)" localsettings.py` could do the job.

Vagrantfile
```diff
@@ -1,36 +1,22 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
-# needed for the hyper-v stuff
-Vagrant.require_version ">= 1.7.4"
+Vagrant.require_version ">= 2.0.1"
```
He3lixxx: Ubuntu has 1.8.1 in the official package repositories for the 16.04 LTS version and 1.9.1 for the most recent version. This would require linux users to install third party packages. Personally, I don't mind that much, but other people might...

deployment/localsettings.py
karyon: yeah, i'll use that, thanks. admittedly i was simply too lazy to create that oneliner and i didn't care enough.

Vagrantfile
karyon: oh wow, 1.9.1 is over a year old, and that's in 17.10...

anyways, i had 1.9.4 installed and that had a bug (https://github.com/hashicorp/vagrant/issues/8520) which destroyed the box directly after creation. it does not occur in 1.9.3 and 1.9.5. i simply didn't care and required the newest version.

i'm a bit unwilling to re-test this with older vagrant versions. how hard is it to install a debian package on ubuntu? (or to install the newest vagrant version by any other means)

janno42: on both ubuntu and debian `apt-get install vagrant` should do the trick.

He3lixxx: Are you aware of any new feature that is not in 1.8.1 that you require? If not, I'd test whether it works with 1.8.1.

He3lixxx: It works fine with vagrant 1.8.1 on my machine. I also get the progress bar printed on multiple lines, but I don't care about that. Could you change the Vagrantfile and set the required vagrant version to 1.8.1?

PR #1084, opened 2018-01-15
Remove some hardcoded pks in tests

karyon:
In rewards test_tools.py, this fixed a test failure that depended on the order of the tests.


evap/rewards/tests/test_views.py
```diff
@@ -121,7 +121,7 @@ class TestEventEditView(ViewTest):
     @classmethod
     def setUpTestData(cls):
         mommy.make(UserProfile, username='staff', groups=[Group.objects.get(name='Staff')])
-        mommy.make(RewardPointRedemptionEvent, pk=1, name='old name')
+        cls.event = mommy.make(RewardPointRedemptionEvent, pk=1, name='old name')
```
He3lixxx: The pk=1 is required here?

karyon: yes, in the url above.

just to be clear, i didn't touch explicit pks whenever they are used in urls, which is in a lot of places, because that would require changing the structure of the tests (url is defined at class level, the pk is available in setUpTestData)

...thinking about it, one could probably replace the pks in the urls with `{}` and then apply a `.format` in setUpTestData. on the other hand, maybe it's just not worth it.

He3lixxx: Ah alright I didn't see that. Never mind then. I agreethat generating the url in two places is not worth it.

PR #1086, opened 2018-01-15
Move datetimepicker and moment.js into their own html file

### LINKED ISSUE #1083: Move datetimepicker and moment.js into their own html file
Since it's a lot of js code used in only very few places, it would be nice to have the includes to moment.js, moment_de.js, bootstrap-datetimepicker.js as well as the js code from base.html (search for datetimepicker) into its own html file, which is included if needed.

We do the same thing for datatables, look at datatables.html.

It's needed on the course edit pages, the course import page, and reward point redemption event pages.

phoeinx:
fixes #1083


evap/evaluation/templates/bootstrap_datetimepicker.html
```diff
@@ -0,0 +1,40 @@
+{% load compress %}
+{% load static %}
+{% get_current_language as LANGUAGE_CODE %}
+
+{% compress js %}
+
+    <script type="text/javascript"  src="{% get_static_prefix %}js/moment.min.js" ></script>
+    <script type="text/javascript"  src="{% get_static_prefix %}js/locale/moment_de.js" ></script>
+    <script type="text/javascript"  src="{% get_static_prefix %}js/bootstrap-datetimepicker.min.js" ></script>
+
```
karyon: remove the two empty lines here and the double spaces in the script tags

evap/evaluation/templates/bootstrap_datetimepicker.html
```diff
@@ -0,0 +1,40 @@
...
+        format: "YYYY-MM-DD HH:mm",
+        sideBySide: true,
+        icons: {
+            up: "fa fa-chevron-up",
+            down: "fa fa-chevron-down",
+            next: "fa fa-chevron-right",
+            previous: "fa fa-chevron-left"
+        }
+    });
+
```
karyon: remove the empty line here and at the beginning of the script tag

PR #1096, opened 2018-02-05
Fix table row anchors

### LINKED ISSUE #1095: No way to open results in a new tab
From the results semester view, one cannot open the results for the individual course with a middle click or right click -> open in new tab.

@janno42 can you fix this please?

janno42:
fix #1095; improve table row anchors and button hover effects

### DIFF
--- a/evap/static/css/evap.scss
+++ b/evap/static/css/evap.scss
@@ -736,22 +736,33 @@ div.disabled-tooltip {
     padding-bottom: 0.4rem;
 }
 
-.table-hover-evap tbody tr:hover:not(.nohover) {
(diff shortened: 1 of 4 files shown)


evap/static/css/evap.scss
```diff
@@ -736,22 +736,32 @@ div.disabled-tooltip {
...
 }
-tr.hover-results:hover .btn-results {
+.table-hover-evap tbody tr td {
+    overflow: hidden;
+}
+.table-hover-evap tbody tr td a:not(.btn) {
+    color: $black;
+    display: block;
+    margin: -10em;
+    padding: 10em;
```
karyon: some explanation somewhere please why this helps

Issue #1078, opened 2018-01-08
Scrolling with a dragged item does not work


He3lixxx:
For example on:
- staff questionnaires edit page
- contributor edit formset (course edit page)
- staff questionnaire edit page
- faq sections
- faq questions
- degrees

If you drag an item, you can not scroll while you have it dragged. Tested with firefox 57.0.3 and Chrome 63.0.3239.132

janno42:
closed by https://github.com/fsr-itse/EvaP/pull/1098

PR #1098, opened 2018-02-05
Fix scrolling with a dragged item (#1078)

felixrindt:


karyon:
in this demo here: http://jsbin.com/tutuzeh/edit?html,js,output it works as expected, and it should be the same scenario as we have. do you want to investigate? please say if you don't, we might just merge anyways or i'll look into it.

janno42:
@karyon on the page you mentioned the same problem exists: if you add more content on top of the table and then drag an item up so far that the table is not visible anymore you also can't stop scrolling.

karyon:
yeah, weird. i actually added a second list and experimented a bit and couldn' reproduce it, but now i can. i'll leave the decision on what to do to you.

felixrindt:
I was also able to reproduce it pretty easily. I'd guess that's it's a bug within sortable.js then.

Anyway, as the corresponding issue is considered FSR major, this would at least make scrolling possible again.

felixrindt:
:tada:

Issue #1042, opened 2017-12-11
Display random backronyms / slogans for EvaP


He3lixxx:
It would be cool to have random slogans like on https://git-scm.com/

The slogans should not occupy any more space or require moving the content of the page down. Also, we do not want to put the text above the logo.

We would need to have slogans in German and English and select one based on the user's language.

Propositions for the slogans in German:
- Evaluierungen verlässlich ausführen und präsentieren
- Exzellenzinitiative zur Verbesserung außergewöhnlicher Programme
- Ein voll atemberaubendes Projekt
- Evaluierungs-Vereinfachung aus Potsdam
- Elegante Verwaltung automatisierter Performancemessungen
- Effektive Vermeidung von anstrengendem Papierkram

English:
- extremely valuable automated processing
- exploring various answers professionally
- enables values and perfection
- enhances verbally adressed petitions
- evading very annoying paperwork
- Engineered voluntarily and passionately
- elegantly and voluntarily administrated performance measurements
- efficient voting and processing
- everyone values awesome products

PR #1099, opened 2018-02-08
Menu redesign and slogans

janno42:
I had some ideas for redesigning the navbar and footer. This is in preparation for #1042 but also to give them a sleeker design in general. The navbar is now smaller and the footer is not fixed anymore (the navbar remains fixed on top).
Note that the results link will also become a single link without dropdown in the future. Non-staff users will then just have the user dropdown for the appropriate choices of settings/rewards/logout in the upper right corner.
Many individual things differ from the current implementation and I'm looking forward to a discussion and even more ideas.

Here are some screenshots:
login
student
responsible
res_top
res_bottom


He3lixxx:
I really like it 👍 

One thing we may want to think about: The menu items in the header are grey on dark-grey. The backronym is white on dark-grey so it's a bit more eye-catching. This doesn't really decrease usability, but we may want to think about whether we want more focus on the slogan than on the menu. Maybe we want to increase the contrast on the menu items a bit and then make the menu items and the slogan the same color?

janno42:
I added slogans and refactored the navbar and footer.
Now ready for review.

janno42:
One test fails because it asserts a certain body of a page, but the slogan is randomly selected on each page load so the message might differ. Any suggestions for a good test solution?

karyon:
you could try `random.seed(42)` at the beginning of the test.

janno42:
that's nice, thanks.

karyon:
on a second look, that test seems to be entirely broken. it visits `/results/semester/<id>`, which does not have the public view parameter, `/results/semester/<id>/course/<id>` does. could you fix that please?

janno42:
check

karyon:
* the text in the navbar is not vertically centered for me, there is more padding to the top (screenshot below)
* i like the language buttons. they seem to be correctly centered, however, the text inside them is not again (in this zoom level by one pixel only though, in other zoom levels by two)
* the hamburger menu would benefit from some vertical padding i think.
* the problems/feedback uses a smaller font, which looks a bit weird (i realize the button would be quite large otherwise, but still)
* actually, by now i think the navbar is too small. i'm used to rather large navbars from most sites, also it looks a bit old-fashioned. maybe there's a middle ground?
* i would prefer to have the rewards in the navbar and not a dropdown to make it more visible.


image

janno42:
note to self: tooltips are broken

janno42:
- navbar height will be set to 42px
- center content will stay centered between the right and left content so that we don't lose too much space (sorry @He3lixxx)
- vertical centering stuff will hopefully be done
- text size in feedback button will be increased
- language change buttons will move into user dropdown
- rewards will move to main navbar
- currently either tooltips are positioned correctly *or* the footer stays at the bottom for pages with content smaller than one screen. this will be fixed but is one of the most annoying things possible so i just wrote done what needs to be done and will wait until i actually have enough motivation to do that.

janno42:
note to self: think about an integration of the breadcrumbs into the navbar.

karyon:
dump from the chat:
* removing the breadcrumb altogether is still weird. e.g. grade publishers have no link to the course when uploading a document for the course, except for the back button.
* we'll need to reevaluate caching of the navbar if there's changing stuff in it
* staff users could have an additional "current semester" in the navbar. (there's more room for that with #1109!)
* or a click on "manage" in the navbar leads to the current semester, while the other stuff is accessible through a smaller dropdown-symbol beneath that
* a second navbar below the first one for staff users was proposed

janno42:
- done
  - height was increased
  - feedback button is larger
  - rewards are in navbar
  - tooltips are correctly positioned
- won't fix
  - i didn't move the language buttons into the user dropdown - i don't like hiding them there because users might want to change the language per course during evaluation
  - vertical alignment: this seems to be a general issue of text positioning in firefox. it usually displays vertically centered text lower than every other browser (see images below). if it's really important we can add some browser-specific settings but i would prefer not having to.
    - chrome: chrome
    - firefox: firefox
- todos:
  - correct collapsing
  - breadcrumbs?

janno42:
i now added a suggestion for breadcrumbs. the breadcrumbs bar is integrated into the navbar and fixed to the top.

on staff and grades views breadcrumbs work as before (only without the leading "EvaP", because that linked to the same page as the large logo right above the breadcrumbs does).
the course edit page for contributors, the course vote page for students and the course result pages for all users now also show the course name (and semester) on the breadcrumbs bar.
breadcrumbs have been removed from all other pages because they were unnecessary.

breadcrumbs
breadcrumbs2

janno42:
collapsing was also changed. ready for review.

karyon:
very nice!

image

on smaller screens, there is a bit too little spacing between the breadcrumb and the headline. if there is any reasonably easy way around that, please do that, otherwise i think we can live with that.

evap/evaluation/templates/base.html
```diff
@@ -37,20 +37,23 @@
...
 {% get_current_language as LANGUAGE_CODE %}
-{% cache 3600 navbar request.session.session_key LANGUAGE_CODE %}
-    {% include_navbar user LANGUAGE_CODE %}
-{% endcache %}
-
+<div class="fixed-top d-print-none">
+    {% cache 3600 navbar request.session.session_key LANGUAGE_CODE %}
+        {% include_navbar user LANGUAGE_CODE %}
+    {% endcache %}
+    {% if user.is_staff or user.is_grade_publisher %}
```
karyon: reviewers don't get the breadcrumb, is that still correct?

evap/grades/templates/grades_base.html
```diff
@@ -4,9 +4,9 @@
 
 {% block breadcrumb %}
     {{ block.super }}
-    {% if is_grade_publisher %}
-        <li class="breadcrumb-item"><a href="{% url 'grades:index' %}">{% trans 'Grade publishing' %}</a></li>
-    {% else %}
+    {% if disable_breadcrumb_grades or not user.is_grade_publisher %}
         <li class="breadcrumb-item">{% trans 'Grade publishing' %}</li>
```
karyon: why is this displayed even if `not user.is_grade_publisher`?

evap/staff/templates/staff_base.html
```diff
@@ -2,11 +2,17 @@
...
-        <li class="breadcrumb-item">{% trans 'Manage' %}</li>
-    {% else %}
-        <li class="breadcrumb-item"><a href="{% url 'staff:index' %}">{% trans 'Manage' %}</a></li>
-    {% endif %}
+{% block breadcrumb_bar %}
+    <div>
+        <ul class="breadcrumb">
+            {% block breadcrumb %}
+                {% if disable_breadcrumb_staff or not user.is_staff %}
+                    <li class="breadcrumb-item">{% trans 'Manage' %}</li>
```
karyon: same here, when is this shown to non-staff users?

evap/results/templates/results_course_detail.html
```diff
@@ -1,14 +1,8 @@
...
 {% load results_templatetags %}
 
-{% block title %}{{ course.name }} - {{ course.semester.name }} - {{ block.super }}{% endblock %}
-
-{% block breadcrumb %}
-    {{ block.super }}
-    <li class="breadcrumb-item"><a href="{% url 'results:semester_detail' course.semester.id %}">{{ course.semester.name }}</a></li>
-    <li class="breadcrumb-item">{{ course.name }}</li>
-{% endblock %}
+{% block title %}{% trans 'Results' %} - {{ course.name }} - {{ course.semester.name }} - {{ block.super }}{% endblock %}
```
karyon: order of this stuff as discussed

evap/evaluation/templates/base.html
janno42: this is outdated :)

evap/grades/templates/grades_base.html
janno42: for staff users who can access the page from the staff semester view

evap/staff/templates/staff_base.html
janno42: for reviewers :)

evap/grades/templates/grades_base.html
```diff
@@ -2,11 +2,17 @@
...
-    {% endif %}
+{% block breadcrumb_bar %}
+    <div>
+        <ul class="breadcrumb">
+            {% block breadcrumb %}
+                {% if disable_breadcrumb_grades or not user.is_grade_publisher %}
+                    <li class="breadcrumb-item">{% trans 'Grade publishing' %}</li>
+                {% else %}
+                    <li class="breadcrumb-item"><a href="{% url 'grades:index' %}">{% trans 'Grade publishing' %}</a></li>
+                {% endif %}
```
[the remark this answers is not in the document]
janno42: we decided to keep it

PR #1105, opened 2018-02-12
check vote success before releasing data (fix #1100)

### LINKED ISSUE #1100: Release Sisyphus data only after successful post
When a user enters answers on the student vote page and then logs out in another window before submitting the form, Sisyphus releases the form data on the form submit, because a 302 redirect to the login page is not an error case.
The data should be kept in browser storage until the vote was successfully counted.

felixrindt:
- [x]  Add serverside test for the 3 cases
- [x] Improve redirect comments in js

fixes #1100


evap/student/templates/student_vote.html
```diff
@@ -141,15 +141,19 @@ <h3 class="mb-3">{{ course.name }} ({{ course.semester.name }})</h3>
...
-                        // Manually perform the redirect: first replace content
-                        document.open();
-                        document.write(data);
-                        document.close();
-                        // then set the URL
-                        window.history.pushState('', '', xhr.responseURL);
-                        // scroll to the top of the page
-                        $('html,body').scrollTop(0);
+                        if(data === "{{ success_magic_string }}") {
+                            sisyphus.manuallyReleaseData(); // delete the data from local storage
```
karyon: can you remove the comment? i think it adds very little value to the code.

evap/student/templates/student_vote.html
```diff
@@ -141,15 +141,19 @@ <h3 class="mb-3">{{ course.name }} ({{ course.semester.name }})</h3>
...
+                            window.location.replace("{{ success_redirect_url }}");
+                        } else {
+                            // Manually perform the redirect: first replace content
```
karyon: maybe "manually perform the page load: first replace content", that would be less wrong

karyon:
Some explanation:

there are several things the server can return:
1. The user opens the page or submits the form with form errors
    - 200 OK with the page is returned
2. The user submits the form successfully
    - 200 OK with the magic success string is returned
3. The user has logged out
    - 302 redirect to login is returned
4. A server error occurred
    - 500 internal server error is returned

Before this PR, the second case would also return a 302. Our JS code would treat 1, 2 and 3 as success (while only 2 would've been correct) and discard the saved form data. Case 3 was reported in #1100. In case 1 (not reported) the form data is not saved anywhere after a form error until the user enters some data and causes sisyphus to save the data again.

With this PR, only case 2 causes the form data to be deleted from local storage.

janno42:
I didn't review yet but would vote for adding tests now. Better right than fast. The most important evaluation phase on production already ended so there's no pressure.

felixrindt:
Ok. I will add tests as soon as I find time for this :)

PR #1107, opened 2018-02-17
Ui updates

### LINKED ISSUE #1070: UI updates
- [ ] Once Bootstrap's stable v4.0.0 is available we could update from the current beta.
- [ ] Also, Font Awesome 5 is now available.

janno42:
- update bootstrap and font awesome (fix #1070)
- fix missing confirmation modal after message was sent
- add missing hover effect on option buttons


evap/static/css/evap.scss
```diff
@@ -426,6 +430,15 @@ td.vote-question.choice-error {
...
     width: 90px;
+    border-color: $medium-gray;
+    background-color: $lighter-gray;
+}
+.vote-btn.active {
+    color: $black;
+    box-shadow: inset 0 0 5px $dark-gray;
+}
+.vote-btn:hover {
+    background-color: lighten($light-gray, 5%);
```
karyon: what do these changes do?

janno42:
note to self: font is broken where set with `content: "\f078";`

evap/evaluation/templates/base.html
```diff
@@ -14,7 +14,9 @@
         <title>{% block title %}EvaP{% endblock %}</title>
 
         {% compress css %}
-            <link rel="stylesheet" href="{% get_static_prefix %}font-awesome/css/font-awesome.min.css" />
+            <link rel="stylesheet" href="{% get_static_prefix %}font-awesome/css/fontawesome.min.css" />
+            <link rel="stylesheet" href="{% get_static_prefix %}font-awesome/css/fa-solid.min.css" />
+            <link rel="stylesheet" href="{% get_static_prefix %}font-awesome/css/fa-regular.min.css" />
```
karyon: if you could give a brief explanation or point to an explanation why these additional files are necessary and why each of them has 5 different versions, that would be nice.

evap/static/css/evap.scss
janno42: active `btn-light` buttons looks differently after the update, this recreates the previous design for the  vote buttons.

janno42:
chevrons in `::before` elements are now fixed.

note to reviewers: font awesome 5 recommends "SVG with JS" which creates `svg` elements for the icons via javascript. i don't see any real advantages, but
- it adds additional java script
- it breaks tooltips on icons
- and there was a third point i can't remember right now
so we keep the "traditional" webfont approach.

evap/evaluation/templates/base.html
janno42: there is now only one version left (i wasn't quite done). after consulting https://www.w3schools.com/css/css3_fonts.asp, we're using WOFF.

what kind of explanation should i add? this is plain font awesome 5, it consists of several fonts internally - these are the ones we need for the icons we're using. no magic involved.

janno42:
@karyon you're welcome to browse through some pages to check whether you spot anything that's broken. i think i opened every page once, but with the ui you can never be sure.

evap/student/templates/student_index.html
```diff
@@ -65,10 +65,10 @@
...
-                                                <span class="badge badge-success" data-toggle="tooltip" data-placement="top" title="{% trans 'You already evaluated this course.' %}"><span class="fa fa-smile-o" aria-hidden="true"></span></span>
+                                                <span class="text text-success" data-toggle="tooltip" data-placement="top" title="{% trans 'You already evaluated this course.' %}"><span class="far fa-smile" aria-hidden="true"></span></span>
                                             {% endif %}
                                         {% elif course.state == 'evaluated' or course.state == 'reviewed' or course.state == 'published' %}
-                                                <span class="badge badge-primary" data-toggle="tooltip" data-placement="top" title="{% trans 'You did not evaluate this course.' %}"><span class="fa fa-frown-o" aria-hidden="true"></span></span>
+                                                <span class="text text-primary" data-toggle="tooltip" data-placement="top" title="{% trans 'You did not evaluate this course.' %}"><span class="far fa-frown" aria-hidden="true"></span></span>
```
karyon: if that's not font-awesome-related, can you move that to the other commits?

evap/evaluation/templates/base.html
karyon: firefox and chrome both will first request the woff2 font, and upon getting a 404, fall back to the woff font. i would suggest re-adding the woff2 versions - and maybe even removing the other fonts from the small css files for correctness? btw https://caniuse.com/#search=woff tells a better story for woff2 than the w3schools page.

karyon:
i found it striking that icons on the questionnaires page look much more heavy now. i think it's less pretty now, but i don't really mind.

image
image

evap/student/templates/student_index.html
janno42: it's related because the icons don't look good otherwise. i can still move it somewhere else if you tell me where exactly you would prefer that. i don't know where you want to have something if you say "to other commits" and i don't want to move it twice.

janno42:
about the heavy thing: yes, the default icons in font awesome 5 are solid (i.e. bold). there is a regular version, but only a few of them are available in the free version. we could get a pro license if that's something we want to have.
but we could also just discard the changes and keep the icons we had before.

evap/evaluation/templates/base.html
janno42: done

evap/student/templates/student_index.html
karyon: no then it's fine. looked like a bootstrap thing to me.

karyon:
i don't mind the heavier icons and the PR can be merged as it is.

PR #1110, opened 2018-02-22
Fix collapsing of contributors, add disabled button for public view

### LINKED ISSUE #1071: Fix results view for courses where results can't be published
The results page for a course where the results can't be published because it didn't get enough votes is not displayed correctly:
- Contributors can't see any non-text answers (that's correct) - but not all contributor cards are collapsed accordingly. If at least one vote exists, the respective contributor card is open, the warning text is not shown and no useful information is shown inside the card (see screenshot 1). The card should be closed in this case.
- Staff users initially see the "public" view and can change the view to `Myself`. On results pages of courses that didn't get enough votes, the `Public` button is missing (that was by design, because a public view does not exist for these courses) - now it's confusing because the initial view is something that's not accessible via the buttons.

Screenshot 1 (Contributor):
cardopen

Screenshot 2 (Staff):
viewbtn

PFischbeck:
This fixes #1071.


evap/results/templates/results_course_detail.html
```diff
@@ -35,16 +35,28 @@ <h3 class="mr-auto">{{ course.name }} ({{ course.semester.name }})</h3>
...
+                            </a>
+                        </span>
+                    {% else %}
+                        {% if course.is_private %}
+                            <a href="{% url 'results:course_detail' course.semester.id course.id %}?public_view=true" role="button" class="btn btn-sm btn-light{% if public_view %} active{% endif %}" data-toggle="tooltip" data-placement="top"
+                                title="{% trans 'Shows results available for the participants.' %}">
+                                {% trans 'Participant' %}
+                            </a>
+                        {% else %}
+                            <a href="{% url 'results:course_detail' course.semester.id course.id %}?public_view=true" role="button" class="btn btn-sm btn-light{% if public_view %} active{% endif %}" data-toggle="tooltip" data-placement="top"
```
He3lixxx: Isn't this exactly the same line as in line 50? Could you move it out of the if block? Same for the `</a>`

evap/results/views.py
```diff
@@ -107,10 +114,15 @@ def course_detail(request, semester_id, course_id):
                                             {'total_votes': 0, 'sections': []})['sections'].append(section)
 
             # Sum up all Sections for this contributor.
-            # If section is not a RatingResult:
-            # Add 1 as we assume it is a TextResult or something similar that should be displayed.
-            contributor_sections[section.contributor]['total_votes'] +=\
-                sum([s.total_count if isinstance(s, RatingResult) else 1 for s in section.results])
+            for s in section.results:
+                if isinstance(s, TextResult):
+                    # If section is a TextResult:
```
He3lixxx: `s` is not a section here. Maybe remove the comment completely? I don't feel like these comments add any valuable information. I also feel like we should replace the variable `s` with something more descriptive like `result`

evap/results/templates/results_course_detail.html
```diff
@@ -35,16 +35,28 @@ <h3 class="mr-auto">{{ course.name }} ({{ course.semester.name }})</h3>
...
-                            title="{% trans 'Shows results available for the participants.' %}">
-                            {% trans 'Participant' %}
-                        </a>
-                    {% elif not sufficient_votes_warning %}
-                        <a href="{% url 'results:course_detail' course.semester.id course.id %}?public_view=true" role="button" class="btn btn-sm btn-light{% if public_view %} active{% endif %}" data-toggle="tooltip" data-placement="top"
-                            title="{% trans 'Shows results available for everyone logged in.' %}">
-                            {% trans 'Public' %}
-                        </a>
+                    {% if sufficient_votes_warning %}
+                        <span class="d-inline-block" tabindex="0" data-toggle="tooltip" data-placement="top" title="{% blocktrans %}The results of this course have not been published because it didn't get enough votes.{% endblocktrans %}">
```
janno42: please use the following instead, this is how we add disabled buttons at other places:
```
<div class="disabled-tooltip" data-toggle="tooltip" data-placement="top" title="{% blocktrans %}The results of this course have not been published because it didn't get enough votes.{% endblocktrans %}">
    <a href="#" role="button" class="btn btn-sm btn-light disabled">
        {% if course.is_private %}
            {% trans 'Participant' %}
        {% else %}
            {% trans 'Public' %}
        {% endif %}
    </a>
</div>
```

PR #1111, opened 2018-02-22
Fix tests for multiple responsibles

### LINKED ISSUE #1059: Review ContributionFormsetTests:test_contribution_form_set
We introduced multiple responsibles in pr #942, which closed issue #896.
Still, [evap/staff/tests/test_forms.py:188](https://github.com/fsr-itse/EvaP/blob/master/evap/staff/tests/test_forms.py#L188) contains the following lines:

```
        # two responsibles
        data['contributions-1-contributor'] = user2.pk
        data['contributions-1-responsibility'] = "RESPONSIBLE"
        self.assertFalse(ContributionFormset(instance=course, form_kwargs={'course': course}, data=data).is_valid())
```

This is semantically incorrect since the form should be valid with two responsibles.

PFischbeck:
This fixes #1059.
The `test_contribution_form_set` test validation returned false because of a missing comment visibility.
I also adjusted `test_dont_validate_deleted_contributions` so that the test would fail if we validated deleted contributions.

### DIFF
--- a/evap/staff/tests/test_forms.py
+++ b/evap/staff/tests/test_forms.py
@@ -181,11 +181,12 @@ class ContributionFormsetTests(TestCase):
         data['contributions-1-course'] = course.pk
         data['contributions-1-questionnaires'] = questionnaire.pk
         data['contributions-1-order'] = 1
+        data['contributions-1-comment_visibility'] = "ALL"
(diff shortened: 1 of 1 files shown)


karyon:
could you modify the second test so it first assertFalse on invalid input data set, and then set the deleted to `on` and assertTrue so this cannot happen again?

He3lixxx:
I agree with @karyon, otherwise I'm fine with the code.

PR #1114, opened 2018-02-26
Course counts

janno42:
the new staff semester page does not have counts for the individual filter options like the course states had before. this pr adds these numbers.

### DIFF
--- a/evap/staff/templates/staff_semester_view.html
+++ b/evap/staff/templates/staff_semester_view.html
@@ -110,6 +110,73 @@
             {% endif %}
         </div>
         <div class="card-body">
+            <div class="btn-switch mb-2" id="course-filter-buttons">
+                <div class="btn-switch-label my-auto">{% trans 'Filter' %}</div>
+                <div class="btn-group icon-buttons">
+                    <a role="button" class="btn btn-sm btn-light" data-filter="fa-circle icon-yellow"
+                        data-toggle="tooltip" data-placement="top" title="{% trans 'In preparation' %}">
+                        <span class="fas fa-circle icon-yellow"></span>
+                    </a>
+                    <a role="button" class="btn btn-sm btn-light" data-filter="fa-square icon-gray"
+                        data-toggle="tooltip" data-placement="top" title="{% trans 'Awaiting editor review' %}">
+                        <span class="far fa-square icon-gray"></span>
+                    </a>
+                    <a role="button" class="btn btn-sm btn-light" data-filter="fa-check-square icon-yellow"
+                        data-toggle="tooltip" data-placement="top" title="{% trans 'Approved by editor, awaiting staff review' %}">
+                        <span class="far fa-check-square icon-yellow"></span>
+                    </a>
+                    <a role="button" class="btn btn-sm btn-light" data-filter="fa-check-square icon-green"
+                        data-toggle="tooltip" data-placement="top" title="{% trans 'Approved by staff' %}">
+                        <span class="far fa-check-square icon-green"></span>
+                    </a>
+                </div>
+                <div class="btn-group icon-buttons ml-2">
+                    <a role="button" class="btn btn-sm btn-light" data-filter="fa-pause icon-gray"
+                        data-toggle="tooltip" data-placement="top" title="{% trans 'Evaluation did not start yet' %}">
+                        <span class="fas fa-pause icon-gray"></span>
+                    </a>
+                    <a role="button" class="btn btn-sm btn-light" data-filter="fa-play icon-gray"
+                        data-toggle="tooltip" data-placement="top" title="{% trans 'In evaluation' %}">
+                        <span class="fas fa-play icon-gray"></span>
+                    </a>
+                    <a role="button" class="btn btn-sm btn-light" data-filter="fa-stop icon-green"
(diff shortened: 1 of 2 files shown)


evap/staff/views.py
```diff
@@ -87,9 +87,61 @@ def __init__(self):
             self.first_start = datetime(9999, 1, 1)
             self.last_end = date(2000, 1, 1)
 
+    # course counts
```
He3lixxx: remove please :D

Edit: I like the rest, so I'll approve when this is changed :)

evap/staff/views.py
```diff
@@ -87,9 +87,61 @@ def __init__(self):
...
+        if course.state in ['new', 'prepared', 'editor_approved', 'approved', 'in_evaluation', 'evaluated']:
+            course_counts.num_evaluation_not_finished += 1
+        elif course.state == 'reviewed':
+            if course.is_single_result or course.final_grade_documents or course.gets_no_grade_documents:
+                course_counts.num_not_published_warning += 1
+            else:
+                course_counts.num_not_published += 1
+        if course.state == 'published':
+            course_counts.num_published += 1
+
```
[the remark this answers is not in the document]
karyon: maybe you can simply iterate over all the buttons and add that span for each after applying datatables.

janno42:
now the counts are added via javascript. also the selected state is restored after page reload.

evap/staff/templates/staff_semester_view.html
```diff
@@ -292,50 +359,49 @@ <h3>
...
+                    };
+                    // add counts to each filter button
+                    $("#course-filter-buttons [data-filter]").each(function(i, obj) {
+                        var count = countIcons(obj);
+                        if (count > 0) {
+                            $('<span class="badge badge-pill badge-btn">' + count + '</span>').appendTo(obj);
+                        }
+                        else {
+                            $('<span class="badge badge-pill badge-btn-zero">' + count + '</span>').appendTo(obj);
+                        }
```
karyon: so i just learned about [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals), how about this:

``$(`<span class="badge badge-pill ${count == 0 ? 'badge-btn-zero' : 'badge-btn'}> ${count} </span>`).appendTo(obj);``

evap/staff/templates/staff_semester_view.html
```diff
@@ -292,50 +359,49 @@ <h3>
...
+                        localStorage.removeItem('course-filter_{{ semester.id }}');
+                        if($("[data-filter].active").length)
+                        {
+                            localStorage['course-filter_{{ semester.id }}'] = $("[data-filter].active")[0].dataset.filter;
+                        }
+                    });
+                    // helper function to count number of lines with certain icon
+                    countIcons = function(obj) {
+                        var count = 0;
+                        $.each([1, 2, 3, 4], function(index, value) {
```
karyon: you could use forEach and the arrow notation for something like this:
```
[1,2,3,4].forEach( i => { 
    count += table.column(i).data().filter( (value, index) => value.includes(obj.dataset.filter)).count();
}
```

okay fancy one-liner incoming
`return [1,2,3,4].reduce( (sum, i) => sum + table.column(i).data().filter( (value, index) => value.includes(obj.dataset.filter)).count(), 0);`

:) but yeah, that probably doesn't help anymore...

in any case, i would inline the function (and maybe call the variable iconCount).

evap/staff/templates/staff_semester_view.html
```diff
@@ -292,50 +359,49 @@ <h3>
...
+                        $("[data-filter].active").not(button).removeClass("active");
                         table.draw();
-                    } );
-                    activateTooltips('.filter-toolbar');
-                },
+
+                        // store which button was selected
+                        localStorage.removeItem('course-filter_{{ semester.id }}');
+                        if($("[data-filter].active").length)
+                        {
```
karyon: brace on the same line please

evap/staff/templates/staff_semester_view.html
```diff
@@ -292,50 +359,49 @@ <h3>
...
+                    // add counts to each filter button
+                    $("#course-filter-buttons [data-filter]").each(function(i, obj) {
+                        var count = countIcons(obj);
+                        if (count > 0) {
+                            $('<span class="badge badge-pill badge-btn">' + count + '</span>').appendTo(obj);
+                        }
+                        else {
+                            $('<span class="badge badge-pill badge-btn-zero">' + count + '</span>').appendTo(obj);
+                        }
+                    });
```
karyon: maybe swap this with the `.click` part above so the localstorage stuff is located together?

evap/staff/templates/staff_semester_view.html
janno42: `${count}` works fine. the inline if does work on it's own but when adding it like in your snippet, nothing happens at all.

karyon:
Random guess, die you use backticks as required for template literals?

janno42:
yes, i am now using template literals, just the `if ? :` part does not work

karyon:
``$(`<span class="badge badge-pill ${iconCount == 0 ? 'badge-btn-zero' : 'badge-btn'}">${iconCount}</span>`).appendTo(obj);`` works.

PR #1115, opened 2018-02-26
Stuff

karyon:
See commit messages.

The staff semester view was sped up from 1.3s to 0.7s in my tests.

My reasoning for removing all those checks on the redirect variable was: In my understanding, they were redirecting a user to their start page if they followed a link to a page that they weren't allowed to see. However, if someone follows a link, logs in, and then sees their default start page, that looks more like a bug. If a 403 is shown, it's obvious what happened.

### DIFF
--- a/evap/evaluation/views.py
+++ b/evap/evaluation/views.py
@@ -64,25 +64,7 @@ def index(request):
         # check for redirect variable
         redirect_to = request.GET.get("next", None)
         if redirect_to is not None:
-            if redirect_to.startswith("/staff/"):
-                if request.user.is_reviewer:
-                    return redirect(redirect_to)
-            elif redirect_to.startswith("/grades/"):
-                if request.user.is_grade_publisher:
-                    return redirect(redirect_to)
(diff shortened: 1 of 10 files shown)


evap/evaluation/views.py
```diff
@@ -85,16 +85,14 @@ def index(request):
                     return redirect(redirect_to)
 
         # redirect user to appropriate start page
-        if request.user.is_staff:
+        if request.user.is_reviewer:
```
janno42: reviewers can't access staff index page

karyon:
fixed tests. fixed your comment in the second commit. added two new commits.

requirements.txt
```diff
@@ -6,7 +6,7 @@ psycopg2 == 2.7.3.2
 django-fsm == 2.6.0
 django-webtest == 1.9.2
 WebTest == 2.0.29
-git+git://github.com/karyon/model_mommy.git@development#egg=model-mommy
+model_mommy = 1.5.1
```
janno42: you have a typo here ;)

PR #1119, opened 2018-03-02
clear cache for course results when the course is (re-) published

### LINKED ISSUE #1032: Missing cache refresh on course state change
You can unpublish a course that is in the state "published". Then, you can change the courses questionnaires. If the course gets published after changing the questionnaires, the old cache will not be invalidated. If someone opens the courses result page, the old questionnaires will still be shown.
The cache entry should be deleted when a staff user unpublishes it.

pixunil:
Fixes #1032

Please note that I am unsure whether the cache belongs to the model of a MVC, but it fits quite well as transition side effect imo.

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -6,6 +6,7 @@ import uuid
 from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth.models import AbstractBaseUser, BaseUserManager, Group, PermissionsMixin
+from django.core.cache import cache
 from django.core.exceptions import ValidationError
 from django.core.mail import EmailMessage
 from django.db import models, transaction
@@ -388,7 +389,8 @@ class Course(models.Model, metaclass=LocalizeModelBase):
 
     @transition(field=state, source='published', target='reviewed')
     def unpublish(self):
-        pass
+        from evap.results.tools import get_results_cache_key
+        cache.delete(get_results_cache_key(self))
 
     @property
     def student_state(self):
--- a/evap/results/tests/test_tools.py
+++ b/evap/results/tests/test_tools.py
@@ -7,7 +7,7 @@ from django.test import override_settings
 from model_mommy import mommy
 
 from evap.evaluation.models import Contribution, RatingAnswerCounter, Questionnaire, Question, Course, UserProfile
-from evap.results.tools import get_answers, get_answers_from_answer_counters, calculate_average_grades_and_deviation, calculate_results
+from evap.results.tools import get_answers, get_answers_from_answer_counters, get_results_cache_key, calculate_average_grades_and_deviation, calculate_results
 from evap.staff.tools import merge_users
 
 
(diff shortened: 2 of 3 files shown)


pixunil:
Should I also add a test case that the cache should be invalidated when a course gets unpublished?
Are you fine with rebasing and squashing or do you prefer to keep already pushed commits?

janno42:
One can almost never have too many tests, so go ahead.
And yes, please rebase and squash.

karyon:
(for larger pull requests, i personally prefer to keep the commits of already reviewed stuff until merging to make the reviewing easier)

evap/evaluation/models.py
```diff
@@ -388,7 +389,10 @@ def publish(self):
 
     @transition(field=state, source='published', target='reviewed')
     def unpublish(self):
-        pass
+        from evap.results.tools import get_results_cache_key
```
He3lixxx: PEP8 says you should put the import at the top of the file. Is a specific reason to do the import here?

@karyon what do you think?

evap/results/tests/test_tools.py
```diff
@@ -7,19 +7,26 @@
...
 
+        self.assertIsNotNone(cache.get(get_results_cache_key(course)))
+
+    def test_cache_unpublished_course(self):
+        course = mommy.make(Course, state='published')
         calculate_results(course)
+        course.unpublish()
 
-        self.assertIsNotNone(cache.get('evap.staff.results.tools.calculate_results-{:d}'.format(course.id)))
+        self.assertIsNone(cache.get('evap.staff.results.tools.calculate_results-{:d}'.format(course.id)))
```
He3lixxx: Could you use the `get_results_cache_key` method here, too?

evap/evaluation/models.py
karyon: probably to avoid a cyclic import, we have that in other locations in that file as well.

evap/evaluation/models.py
```diff
@@ -388,7 +389,10 @@ def publish(self):
...
+        # invalidate cache
+        cache_key = get_results_cache_key(self)
+        cache.delete(cache_key)
```
karyon: i would prefer the variable inlined, since the method says that it's about to return a cache key. also, the comment adds little value now.

karyon:
looks good! one more nit, and what @He3lixxx said, then we're good to go. sorry for the late reply, i didn't get a notification for your newly pushed commits.

evap/evaluation/models.py
He3lixxx: Ah okay, its fine then :)

He3lixxx:
I agree to karyon, thanks! :)
Also, sorry for not reviewing earlier.

pixunil:
Okay, now should be everything be set. I might be mistaken, but I don't think that GitHub notifies you for single commits…

PR #1120, opened 2018-03-02
grant reward points when the removal of a participation fulfils the conditions

### LINKED ISSUE #543: Reward point check on participation change
When the participations of users in a course are removed this might qualify them to receive reward points in case all their other participations in the active semester have already been evaluated.
This could happen by removing a participant in a course form or removing a course in the user's profile.

The platform should check for these cases and grant reward points accordingly.

pixunil:
fixes #543

### DIFF
--- a/evap/rewards/tests/test_tools.py
+++ b/evap/rewards/tests/test_tools.py
@@ -1,4 +1,5 @@
 from django.conf import settings
+from django.test import TestCase
 from django.urls import reverse
 
 from model_mommy import mommy
@@ -37,7 +38,7 @@ class TestGrantRewardPoints(WebTest):
     def test_everything_works(self):
         SemesterActivation.objects.create(semester=self.course.semester, is_active=True)
         self.form.submit()
-        self.assertEqual(settings.REWARD_POINTS_PER_SEMESTER, reward_points_of_user(self.student))
+        self.assertEqual(reward_points_of_user(self.student), settings.REWARD_POINTS_PER_SEMESTER)
 
     def test_semester_activated_not_all_courses(self):
         SemesterActivation.objects.create(semester=self.course.semester, is_active=True)
@@ -50,3 +51,22 @@ class TestGrantRewardPoints(WebTest):
         mommy.make(RewardPointGranting, user_profile=self.student, value=0, semester=self.course.semester)
         self.form.submit()
         self.assertEqual(0, reward_points_of_user(self.student))
+
+class TestGrantRewardPointsParticipationChange(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.course = mommy.make(Course)
+        already_evaluated = mommy.make(Course, semester=cls.course.semester)
+        SemesterActivation.objects.create(semester=cls.course.semester, is_active=True)
+        cls.student = mommy.make(UserProfile, username="student", email="foo@institution.example.com",
+            courses_participating_in=[cls.course, already_evaluated], courses_voted_for=[already_evaluated])
+
+    def test_participant_removed_from_course(self):
+        self.course.participants.remove(self.student)
+
+        self.assertEqual(reward_points_of_user(self.student), settings.REWARD_POINTS_PER_SEMESTER)
+
+    def test_course_removed_from_participant(self):
+        self.student.courses_participating_in.remove(self.course)
+
+        self.assertEqual(reward_points_of_user(self.student), settings.REWARD_POINTS_PER_SEMESTER)
--- a/evap/rewards/tools.py
+++ b/evap/rewards/tools.py
@@ -2,12 +2,12 @@ from datetime import date
 
 from django.conf import settings
 from django.contrib import messages
-from django.db import transaction
+from django.db import models, transaction
 from django.utils.translation import ugettext as _
 from django.dispatch import receiver
 from django.contrib.auth.decorators import login_required
 
-from evap.evaluation.models import Course
+from evap.evaluation.models import Semester, Course, UserProfile
 
 from evap.rewards.models import RewardPointGranting, RewardPointRedemption, RewardPointRedemptionEvent, \
                                 SemesterActivation, NoPointsSelected, NotEnoughPoints, RedemptionEventExpired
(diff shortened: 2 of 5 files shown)


evap/rewards/tools.py
```diff
@@ -59,30 +59,45 @@ def reward_points_of_user(user):
...
+    request = kwargs['request']
+    semester = kwargs['semester']
+
+    if grant_reward_points(request.user, semester):
+        messages.success(request, _("You just have earned reward points for this semester because you evaluated all your courses. Thank you very much!"))
+
+@receiver(models.signals.m2m_changed, sender=Course.participants.through)
+def grant_reward_points_after_delete(instance, action, reverse, pk_set, **kwargs):
+    # if users do not need to evaluate a course anymore, they may have earned reward points
+    if action == "post_remove":
```
karyon: we have a loose rule which we didn't tell you yet that in python code user-facing strings use double quotes, and other strings single quotes. this would be one of those other strings :)

evap/rewards/tools.py
```diff
@@ -59,30 +59,45 @@ def reward_points_of_user(user):
...
+        if reverse:
+            for semester in Semester.objects.filter(course__pk__in=pk_set):
+                grant_reward_points(instance, semester)
+        else:
+            for user in UserProfile.objects.filter(pk__in=pk_set):
+                grant_reward_points(user, instance.semester)
```
karyon: i needed a while to understand this and would suggest the following to speed that up for others:

- to add comments behind the `if` and `else` that say `course got removed` and `participant got removed` respectively
- to add variables `user = instance` and `course = instance` in the two branches

i admit that after checking the documentation of `m2m_changed`, it's pretty clear what happens, but still, since we're not using this mechanic anywhere else, maybe additional help for the reader is still warranted. @He3lixxx opinions?

karyon: @janno42 i was wondering whether some message to the staff user that reward point changes happened would be desirable?

evap/rewards/tests/test_tools.py
```diff
@@ -45,6 +45,20 @@ def test_semester_activated_not_all_courses(self):
         self.form.submit()
         self.assertEqual(0, reward_points_of_user(self.student))
 
+    def test_removes_from_course(self):
+        SemesterActivation.objects.create(semester=self.course.semester, is_active=True)
+        mommy.make(Course, semester=self.course.semester, participants=[self.student])
+        self.form.submit()
+        self.student.courses_participating_in.set([self.course])
```
karyon: is something like `self.student.courses_participating_in.remove(self.course)` possible here as well? i think that would make clearer what's actually happening

evap/rewards/tests/test_tools.py
```diff
@@ -45,6 +45,20 @@ def test_semester_activated_not_all_courses(self):
...
+        self.assertEqual(settings.REWARD_POINTS_PER_SEMESTER, reward_points_of_user(self.student))
+
+    def test_removed_from_course(self):
```
karyon: maybe call the tests `course_removed_from_participant` and `participant_removed_from_course`?

karyon: nice tests. however, this test class currently requires you to use that form, but that's not really the part that's tested.

so, to make clearer what's actually tested, i would suggest to make a new `class TestGrantRewardPointsParticipationChange(TestCase):` (or similar). You probably won't need a WebTest since you don't need to use an actual view for this, this also makes the tests much faster. You probably also don't need most of the stuff in the two setup-methods.

evap/rewards/tools.py
janno42: yes, please.

evap/rewards/tests/test_tools.py
janno42: it would be nice to test that the staff user gets an information message though.

evap/rewards/tools.py
janno42: to me it was clear what happens but with the suggested changes it should be obvious even without having the detailed model structure in mind, so that would be nice.

pixunil: Agreed, I found it hard to express myself with the different sides altered. Same for the test names.

He3lixxx: For the sake of making this complete, altough this answer is way too late: I agree with karyon.

pixunil:
### Some explanation
I kept the `m2mchanged` receiver, as it grants the integrity of the rewards model. As an addition, it now sends a signal out when a participation change includes reward points granting. Views, which are known for altering the participation can listen to this signal to provide the message (through closures, they can use the `request` object).

#### Advantages

- Test of participation change → reward points can be done on model layer
- View tests only need to test the view (separation)
- Tool does not access messages

#### Disadvantages

- Repetition in model and view tests, as they all follow the same scheme (one course which participation gets deleted, one which is already evaluated and one or more students)
- Some inconsistency with the normal cause of reward points, which receives a custom event from the views instead of sending it to the views

### Preview

user-edit
course-edit

I chose `info`, as it is a neutral side effect and I haven't lost any thoughts about the order or a better text; suggestions are welcome.

evap/rewards/tests/test_tools.py
```diff
@@ -50,3 +51,22 @@ def test_already_got_points(self):
...
         self.form.submit()
         self.assertEqual(0, reward_points_of_user(self.student))
+
+class TestGrantRewardPointsParticipationChange(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.course = mommy.make(Course)
+        already_evaluated = mommy.make(Course, semester=cls.course.semester)
+        SemesterActivation.objects.create(semester=cls.course.semester, is_active=True)
+        cls.student = mommy.make(UserProfile, username="student", email="foo@institution.example.com",
```
pixunil: Here, I wasn't sure whether the attributes `student` and `email` are considered user-facing or not. In the test clearly not, but normally, they will be displayed on the page.

evap/staff/views.py
```diff
@@ -569,6 +570,12 @@ def course_edit(request, semester_id, course_id):
 
 @staff_required
 def helper_course_edit(request, semester, course):
+    @receiver(RewardPointGranting.granted_by_removal)
+    def notify_reward_points(users, **kwargs):
+        messages.info(request, ungettext("The removal of participants has granted %(affected)d user reward points.",
+            "The removal of participants has granted %(affected)d users reward points.",
```
janno42: can you maybe add a comma separated list of the affected user's usernames at the end?

evap/staff/views.py
```diff
@@ -1140,6 +1147,10 @@ def user_import(request):
 
 @staff_required
 def user_edit(request, user_id):
+    @receiver(RewardPointGranting.granted_by_removal)
+    def notify_reward_points(users, **kwargs):
+        messages.info(request, _("The removal of courses has granted {} reward points for the active semester.".format(users[0])))
```
janno42: can you please change this to `'The removal of courses has granted the user "{}" reward points for the active semester.'`?

evap/staff/views.py
pixunil: Do you want to cap it at a limit? Like `The removal of participants has granted x users reward points: "alfa", "beta", "charlie", "delta", "echo" and 22 additional users` or display all? Someone has explained the worst-case scenario that nearly all participants are affected (like a late-semester event which has been cancelled after all other lectures are open for evaluation and all students were very busy).

evap/staff/views.py
pixunil: Sorry for nitpicking, but wouldn't it be `"The removal of courses has granted the user \"{}\" reward points for the active semester."`? Also note that the message makes an assumption, that the form can only edit the active semester and therefore the message's content reflect this.

evap/staff/views.py
janno42: that's fine, just show them all. it's not a problem if in very special cases the lists can get very long - usually it will only be a short list.

evap/staff/views.py
janno42: i saw the active semester, that's fine. in some cases we're escaping the double quotation marks, in others we're using single quotation marks for the string when having doubles in there - i don't really care.

evap/rewards/tests/test_tools.py
karyon: the maxime is usually to not specify values that you're not testing against, so i would remove them here.

janno42: But it needs to be an internal user, so the email is required

evap/rewards/tests/test_tools.py
```diff
@@ -50,3 +51,22 @@ def test_already_got_points(self):
...
+
+    def test_participant_removed_from_course(self):
+        self.course.participants.remove(self.student)
+
+        self.assertEqual(settings.REWARD_POINTS_PER_SEMESTER, reward_points_of_user(self.student))
+
+    def test_course_removed_from_participant(self):
+        self.student.courses_participating_in.remove(self.course)
+
+        self.assertEqual(settings.REWARD_POINTS_PER_SEMESTER, reward_points_of_user(self.student))
```
karyon: i would say the variable is equal to the constant, and not the constant is equal to the variable, if you know what i mean.

karyon:
i like this approach very much. however, i have one concern: what happens if the view is run once, and connects a receiver, finishes running, and then somewhen later the signal is sent? is the receiver cleaned up somehow? or is it even going to register a new receiver every time the view runs? you could verify this by putting a print in the receiver and then grant reward points by removing an object e.g. through the shell.

karyon:
thanks, sounds good. congratulations on disregarding my suggestions and finding a better solution. it's a bit too complex for my taste, but it does the right thing, which is itself worth a lot.

pixunil:
So, I think the only waterproof solution would be to `disconnect` the receiver afterwards. Otherwise, as reference-counted gc is nearly always present and the receiver won't leak to other parts outside the view it should be fine.

karyon:
i think it's fine for our purposes to rely on the receiver being destroyed when the view finishes.

evap/staff/tests/test_views.py
```diff
@@ -139,6 +140,21 @@ def test_questionnaire_edit(self):
         form.submit()
         self.assertTrue(UserProfile.objects.filter(username='lfo9e7bmxp1xi').exists())
 
+    def test_remove_course(self):
```
He3lixxx: Maybe rename the test to "test_reward_points_granting_message"?
Maybe let mommy set the username and compare to the (random) name in the assertion?

evap/staff/tests/test_views.py
```diff
@@ -886,6 +902,40 @@ def test_remove_responsibility(self):
 
         self.assertIn("No responsible contributors found", page)
 
+    def test_remove_participant(self):
```
He3lixxx: Maybe rename to "test_participant_removal_reward_point_granting_message"?
Maybe let mommy set the username and compare to the (random) name in the assertion?

Issue #1040, opened 2017-12-11
Evaluate multiple contributors together


He3lixxx:
Users should be able to answer a questionnaire for multiple contributors. Each of the contributors should then be able to see the results of the questionnaire.

Use case: n students give a talk, the talk is evaluated and every student should be able to see the results.

janno42:
see discussion in #1122 for additional details and requirements

PR #1122, opened 2018-03-12
[WIP] Change contributions to have multiple contributors per contribution

PFischbeck:
This is a WIP for #1040.
There is still some stuff to do:
- [ ] Fix all tests
- [ ] Think about and implement uniqueness of contributions (should each user only occur at most once per course, so he cannot be in multiple contributions per course?)
- [ ] Write tests (and maybe fixtures) that actually use this new feature
- [ ] Evaluate whether this feature is worth the many changes


karyon:
another concern was the complexity of the course edit form which will increase, and we will need to find a way so users will absolutely not misuse this feature, otherwise it will cause more work for staff users correcting the errors of editors.

does stuff like a responsible contribution with multiple responsibles make sense at all?

we also need clarification what the actual need behind this feature request was. should multiple contributors be able to see text answers for a contribution? should participants see these names when voting? something else?

janno42:
- editors should not be able to add multiple contributors for one contribution, so the form for them can stay as it is. only staff members should be able to add multiple people to a contribution. we could use a multiple select input field for them and once multiple people have been added to a contribution, we could remove the input field on the editor page and just show the values as plain text there.
- otherwise, contributions could work the same way as they currently do:
  - that means a responsible contribution can have multiple contributors (and probably we can also have multiple contributions marked as responsible, each having multiple contributors). there is no specific use case for that but it probably makes the implementation easier when we're not adding additional restrictions here.
  - all contributors of a contribution are shown in the questionnaire (comma separated list)
  - all contributors of a contribution can see the respective results

the use case for this feature is having a large course where some parts have been done together by two or more contributors who can't (or shouldn't) be evaluated individually. there can be multiple of these "contributor groups" in a course, not all might apply to every participant. so adding to the course questions is no option.

there is an additional requirement that comes up: it must be possible to have the same contributor in multiple contributions in one course. this happens regularly when two or more people together give a talk (or workshop or whatever), are evaluated together, and some of these people also give other talks together with different (or the same) people and we need another questionnaire for that.

karyon:
what if two contributors give two talks together?

janno42:
That's a special case of the one described above and should also be possible.

karyon:
..that might have a few more implications since there is no uniqueness constraint left anymore. e.g. `calculate_results` saves result per contributor-questionnaire pair and disregards contributions completely. this would need to be changed to save contribution-questionnaire pairs instead.

PFischbeck:
Since I haven't been able to continue working on this for some time now, feel free to re-assign issue #1040 and use this WIP PR as base.

janno42:
so you won't join us in the near future to continue your work on it? :(

karyon:
@PFischbeck works on other stuff now and this PR is hopelessly out of date, closing.

PR #1125, opened 2018-03-12
Add general questions below contributor questions

### LINKED ISSUE #605: Move grade question below contributor questions
Currently the grade question is part of the general questions and thus appears above the contributor questions. IMO it should be the very last question to be answered, so evap should make it possible to arrange questions accordingly.

maybe the "Miscellaneous" question should also appear at the end.

He3lixxx:
Work in progress. Fixes #605 
The student vote page should work as well as all vote preview pages
The staff questionnaire index page should work. I haven't thought about allowing drag and drop between the lists yet.
The staff course edit page should work.
The result's course detail page should work.

Currently not working:
Staff questionnaire create (test fails, haven't manually tested).
Staff questionnaire edit - when you change the questionnaire's type, the order attribute will not be updated as intended in the QuestionnaireForm's save method. Looks like the `formset.is_valid()` call changes the questionnaire instance in the questionnaire_edit view, but I haven't taken a closer look yet.


evap/evaluation/models.py
```diff
@@ -111,25 +129,38 @@ class Questionnaire(models.Model, metaclass=LocalizeModelBase):
...
+        ordering = ('type', 'order', 'name_de')
         verbose_name = _("questionnaire")
         verbose_name_plural = _("questionnaires")
 
     def __str__(self):
         return self.name
 
     def __lt__(self, other):
-        return (self.is_for_contributors, self.index) < (other.is_for_contributors, other.index)
+        return (self.type, self.order) < (other.type, other.index)
```
He3lixxx: other.order

evap/evaluation/models.py
```diff
@@ -111,25 +129,38 @@ class Questionnaire(models.Model, metaclass=LocalizeModelBase):
...
 
     def __gt__(self, other):
-        return (self.is_for_contributors, self.index) > (other.is_for_contributors, other.index)
+        return (self.type, self.order) > (other.type, other.index)
```
He3lixxx: other.order

evap/staff/views.py
```diff
@@ -1002,7 +1003,7 @@ def questionnaire_update_indices(request):
     updated_indices = request.POST
     for questionnaire_id, new_index in updated_indices.items():
         questionnaire = Questionnaire.objects.get(pk=questionnaire_id)
-        questionnaire.index = new_index
+        questionnaire.order = new_index
```
He3lixxx: new_order

evap/contributor/forms.py
```diff
@@ -25,7 +25,7 @@ class Meta:
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        self.fields['general_questions'].queryset = Questionnaire.objects.filter(is_for_contributors=False).filter(
+        self.fields['general_questions'].queryset = Questionnaire.objects.general_questionnaires().filter(
```
karyon: we're already calling those "course questionnaires" somewhere in the code.

...actually, maybe the general contribution should be called "course contribution"? but oh well...

evap/evaluation/migrations/0064_questionnaire_position.py
```diff
@@ -0,0 +1,36 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('evaluation', '0063_add_heading_question_type'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='questionnaire',
+            options={'ordering': ('order', 'name_de'), 'verbose_name': 'questionnaire', 'verbose_name_plural': 'questionnaires'},
```
karyon: the ordering doesn't seem to reflect what's actually in the model?

evap/evaluation/migrations/0064_questionnaire_position.py
```diff
@@ -0,0 +1,36 @@
...
+        ),
+        migrations.AddField(
+            model_name='questionnaire',
+            name='type',
+            field=models.IntegerField(choices=[(0, 'Top questionnaire'), (1, 'Contributor questionnaire'), (2, 'Bottom questionnaire')], default=0, verbose_name='type'),
+            preserve_default=False,
+        ),
+        migrations.RunSQL(
+            "UPDATE evaluation_questionnaire SET type = 1 WHERE is_for_contributors = TRUE;",
+            reverse_sql="UPDATE evaluation_questionnaire SET is_for_contributors = TRUE WHERE type = 1"
```
karyon: semicolon vs no semicolon

karyon:
[optional] for reviewing (and actually for the migration) it would be cool to have the index renaming in a separate commit, and the introduction of the custom manager as well.

karyon:
...there already exists a test that uses the term "general_questionnaire". can you change that? maybe in a separate commit? :) maybe together with removing the unused imports? :D

evap/results/templates/results_course_detail.html
```diff
@@ -176,4 +176,18 @@ <h3>{{ course.name }} ({{ course.semester.name }})</h3>
             </div>
         </div>
     {% endif %}
+
+    {% if course_sections_bottom %}
+        <div class="card card-outline-primary">
```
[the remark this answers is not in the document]
He3lixxx: I thought the last card usually has no mb-3, see old line 136. @janno42?

He3lixxx:
voilà

evap/results/templates/results_course_detail.html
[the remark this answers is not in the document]
janno42: the last card should have no bottom margin, that's correct.

He3lixxx:
Rebased, all commits now pass all tests in my vm. Travis seems to have random test failures, I don't know why.

Still need to add tests.

evap/evaluation/migrations/0065_questionnaire_type.py
```diff
@@ -0,0 +1,31 @@
...
+        ),
+        migrations.AddField(
+            model_name='questionnaire',
+            name='type',
+            field=models.IntegerField(choices=[(0, 'Top questionnaire'), (1, 'Contributor questionnaire'), (2, 'Bottom questionnaire')], default=0, verbose_name='type'),
+            preserve_default=False,
+        ),
+        migrations.RunSQL(
+            "UPDATE evaluation_questionnaire SET type = 1 WHERE is_for_contributors = TRUE;",
+            reverse_sql="UPDATE evaluation_questionnaire SET is_for_contributors = TRUE WHERE type = 1;"
```
karyon: i know we talked about this before, but upon thinking about this again, there is no good reason to use plain sql here i think. apart from necessitating two methods, the ORM version should be similarly succinct.

evap/staff/forms.py
```diff
@@ -387,6 +387,11 @@ def send(self, request, courses):
 
 
 class QuestionnaireForm(forms.ModelForm):
+    def save(self, *args, **kwargs):
+        if self.cleaned_data['type'] != self.instance.type:
+            self.cleaned_data['order'] = Questionnaire.objects.filter(type=self.cleaned_data['type']).aggregate(Max('order'))['order__max'] + 1
```
karyon: does the condition work correctly when creating a new questionnaire?

evap/student/templates/student_vote.html
```diff
@@ -80,6 +82,16 @@ <h3 class="mb-3">{{ course.name }} ({{ course.semester.name }})</h3>
...
+                <div class="card-header">
+                    {% trans "Questions about the course" %}
+                </div>
+                <div class="card-body">
+                    {% include "student_vote_questionnaire_group.html" with questionnaire_group=course_form_group_bottom preview=preview %}
+                </div>
+            </div>
+        {% endif %}
         {% if not for_rendering_in_modal %}
         <div class="card card-submit-area{% if not preview %} card-submit-area-2{% endif %} text-center mb-3">
```
karyon: last card has bottom margin (i don't know what i'm talking about though)

evap/student/views.py
```diff
@@ -75,11 +81,15 @@ def vote(request, course_id):
 
         course_form_group = form_groups.pop(course.general_contribution)
 
+        course_form_group_top = [questions_form for questions_form in course_form_group if questions_form.questionnaire.is_above_contributors]
+        course_form_group_bottom = [questions_form for questions_form in course_form_group if questions_form.questionnaire.is_below_contributors]
+
         contributor_form_groups = list((contribution.contributor, contribution.label, form_group, helper_has_errors(form_group)) for contribution, form_group in form_groups.items())
```
karyon: what if we let `helper_create_voting_form_groups` return `(course_form_group_top, contributor_form_groups, course_form_group_bottom)`? the transformation of `contributor_form_groups` to that other format with the label etc. would stay here i think. for the `if not all...` form validation we would need a list along the lines of `course_form_group_top + contributor_form_groups.values() + course_form_group_bottom`.

or rather, maybe lets get your PR through and you (or I) do it afterwards.

karyon:
umm did we talk about the exporter yet?

:(

He3lixxx:
I will do the requested changes.

I think I asked @janno42 about the exporter and he answered that the distinction is not necessary there, but I'm not sure about that. Janno, is this correct?

The test failures had the following reason: Before, the is_for_contributors field had a default value (false). Thus, every call like mommy.make(Questionnaire) created a course questionnaire. I removed the default since the old default value was split into two options and I don't know if top-questionnaire is an appropriate default. Therefore, mommy always took a random choice, which led to test failures. Two options:
- Make either top or bottom the default (I guess it'd be top then)
- Change every test that createst a questionnaire to specify a type.
What do you think @karyon @janno42?

evap/student/views.py
He3lixxx: See #1144, will do afterwards.

evap/evaluation/migrations/0065_questionnaire_type.py
He3lixxx: done

evap/student/templates/student_vote.html
He3lixxx: :D done

evap/staff/forms.py
He3lixxx: When you're creating a new questionnaire, this doesn't make a difference since the order will be overriden in the questionnaire_create staff view.

This does not work when changing an existing questionnaire - it was still on my todo list. I'm not sure about whether I want to do what this code implies though. When you edit a questionnaire, you can edit both, the type and the order value. Now, if you have a questionnaire that's on the top of the list and you go to the edit page, you will have the values type top and order 0 in the input fields. If you just change the type and click submit, wouldn't it be strange if the order changed from 0 to to highest value?
Or even more strange, if it was the last and you wanted it to be the first, so you changed the type from top to bottom and the order from whatever to 0. You submit the form and the questionnaire is back on the bottom with a high order value.

From this perspective, I would not touch the order. It's not a big problem if we have duplicate order values in one type and reordering them with the drag-and-drop interface will fix up every messy value.

Also strange: When a questionnaire is in use, you can not edit it's order value on the edit page but you can change it with the drag-and-drop interface.

@janno42 what's your opinion?

karyon: you could check whether the instance has a pk and if not, assume this is a newly created questionnaire, and then do the number assignment here instead of in the view. i think that would be nicer.

for existing questionnaires, having the order in the form is weird i think because the user has little information what this means. even if they know how the sorting works, the only options that make any sense are zero (meaning "either above or below the previous top questionnaire" or "some large number" meaning at the bottom.

the simplest solution would probably be to remove the order field from the form, always put it to the bottom, and maybe output a message to the user after saving that they need to check the ordering.

karyon:
3. i think top questionnaire would be an appropriate default.

4. i'm fine with removing is_for_contributors.

5. yeah, good idea. the grade questionnaire or "other comments" or something like that maybe, if the test data have that and it's selected for most courses.

janno42:
2: I don't remember discussing that. I would argue that the order in the export should be the same as on the voting pages.

3: What karyon said. Having a default makes sense because the respective checkbox in the form must have a default anyways.

evap/staff/forms.py
janno42: What's the order field you're talking about? Did you add that to the form? Nobody will ever use such a field in the form. Ordering only happens via staff and drop on the overview page. Users will expect that new questionnaires will be added on the bottom and when changing the "type" moving it to the bottom (preferably with the message) is also okay.

He3lixxx: Ahh now that makes sense. Yes, I added it - by accident, when I renamed the "index" field to "order". The fields that are excluded in the form are only referenced by name as a string, so nothing broke and I didn't notice. Alright, so it'll just be excluded. Never mind then.

He3lixxx:
I did everything we mentioned here.

The exporter will be sorted properly because it relies on the Questionnaire's comparison functions __gt__ and __lt__ for sorting, I've changed these.

Todo:
- Change questionnaire form so it sets the order correctly. Remove corresponding code from staff/views.py:questionnaire_create
- Add tests

He3lixxx:
I squashed the commits together.

evap/staff/templates/staff_questionnaire_index.html
```diff
@@ -17,11 +17,13 @@
         </div>
     </div>
 
-    {% if course_questionnaires or contributor_questionnaires %}
-        {% trans 'Course questionnaires' as headline %}
-        {% include 'staff_questionnaire_index_list.html' with questionnaires=course_questionnaires headline=headline extra_classes='mb-3' %}
+    {% if course_questionnaires_top or course_questionnaires_bottom or contributor_questionnaires %}
+        {% trans 'Top course questionnaires' as headline %}
+        {% include 'staff_questionnaire_index_list.html' with questionnaires=course_questionnaires_top headline=headline extra_classes='mb-3' %}
```
janno42: it would be nice if you could make the three cards collapsible

evap/staff/tests/test_forms.py
```diff
@@ -3,13 +3,60 @@
...
+        data = {
+            'description_de': 'English description',
+            'description_en': 'German description',
+            'name_de': 'A name',
+            'name_en': 'A german name',
+            'public_name_en': 'A display name',
+            'public_name_de': 'A german display name',
+            'question_set-0-id': question.id,
+            'order': 20,
+            'type': Questionnaire.TOP,
```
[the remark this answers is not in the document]
He3lixxx: Creating a new questionnaire was not my main intention on this test.
`questionnaire = form.save(force_highest_order=True)` is what this is about. This also happens when a new form is created, yes. You rather want `test_manual_order_correction_on_questionnaire_creation` here? I don't feel like that's right.

The intention with setting the order was that the view actually supplies the default values for hidden fields, so on the view, we usually have order=0 supplied to the form. I just wanted to test that it ignores the default values. Should I still remove it or just set it to 0?

He3lixxx:
Everything resolved except the ongoing discussion about `test_manual_order_correction` in the staff forms test.

karyon:
> also, it would help a lot if the type could be changed from top course to bottom course in the questionnaire form even if the questionnaire is already in use
@janno42 why?

janno42:
because i don't want to make this change in the database if we decide to change the order. within each "section" we can change it by drag and drop but not from top to bottom.

evap/staff/forms.py
```diff
@@ -387,6 +387,20 @@ def send(self, request, courses):
...
+
+    def clean(self, *args, **kwargs):
+        cleaned_data = super().clean(*args, **kwargs)
+
+        if self.disallow_type_change_to_contributor and 'type' in self.changed_data:
+            if self.instance.type != Questionnaire.CONTRIBUTOR and self.cleaned_data['type'] == Questionnaire.CONTRIBUTOR:
+                self.add_error('type', _("This questionnaire is in use, so its type can only be changed between top and bottom."))
+
+        return cleaned_data
+
```
karyon: please move the Metaclass to the beginning of this class.

evap/staff/views.py
```diff
@@ -894,6 +894,10 @@ def make_questionnaire_edit_forms(request, questionnaire, editable):
 
     if not editable:
         editable_fields = ['staff_only', 'obsolete', 'name_de', 'name_en', 'description_de', 'description_en']
+        if questionnaire.type != Questionnaire.CONTRIBUTOR:
+            editable_fields.append('type')
+            form.disallow_type_change_to_contributor = True
+
```
karyon: can you check whether we can restrict the choices of the type field? that would have the advantage that the user sees only the valid choices, and maybe the validation is also done for us.

evap/staff/tests/test_forms.py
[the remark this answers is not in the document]
karyon: if you're testing the `force_highest_order` parameter, you could call the test `test_force_highest_order`. regarding the order parameter, if the form will usually be used with a zero in there, I think it's reasonable to put a zero in there.

evap/evaluation/migrations/0065_questionnaire_type.py
```diff
@@ -0,0 +1,41 @@
...
+    operations = [
+        migrations.AlterModelOptions(
+            name='questionnaire',
+            options={'ordering': ('type', 'order', 'name_de'), 'verbose_name': 'questionnaire', 'verbose_name_plural': 'questionnaires'},
+        ),
+        migrations.AddField(
+            model_name='questionnaire',
+            name='type',
+            field=models.IntegerField(choices=[(0, 'Top questionnaire'), (1, 'Contributor questionnaire'), (2, 'Bottom questionnaire')], default=0, verbose_name='type'),
+            preserve_default=True,
```
karyon: we haven't discussed this yet, but I think that is the dropout questionnaire by @jakobmache and the single result questionnaire are probably going to have their own type. the evaluation feedback (#212) might also get a new questionnaire type.

Now the question is, if we add those types, do we simply append those to the list of questionnaire types, since the ordering for them is irrelevant since they are special-cased anyways? Or do we want them in a certain position in the list (e.g. dropout questionnaire at the very top since it will be shown at the very top)? In that case I would leave generous gaps between the numbers to not require migrations for rewriting all the types on each change. But I think I'm fine with appending. What do you think?

evap/evaluation/templates/questionnaires_widget.html
```diff
@@ -2,6 +2,11 @@
 
 <ul class="inputs-list">
     {% for choice, questionnaire in field|zip:field.field.queryset %}
+        {% ifchanged questionnaire.type %}
```
janno42: didn't know this feature existed in django template. nice!

evap/evaluation/migrations/0065_questionnaire_type.py
He3lixxx: I guess @janno42 will have to decide on the ordering that will be required. Having the dropout questionnaire at the top seems logical to me. Which values would you set right now if you had to leave gaps? 1000, 2000 and 3000?

karyon: well, the ordering in the UI is technically irrelevant, since those special questionnaires are most likely special cased and explicitly inserted at some specific place. it's more a semantic thing. if we want the type to match the ordering even for those types where the type is not actually used for ordering.

initially i thought of 10, 20 and 30. but in the end it does not matter at all, you choose :)

He3lixxx: 10, 20, 30 it is :)

janno42:
I'm totally fine with that. Additional explanations besides the general info box on the top aren't necessary.

karyon:
yeah, das more like it.

He3lixxx:
Nice :smiley:

PR #1126, opened 2018-03-16
allow transition from 'editor_approved' to 'new'

janno42:
the transition from course state `editor_approved` to `new` was not allowed although the ui shows this option and the `semester_course_operation` can also handle this change.


karyon:
and in the UI this would just crash?

janno42:
yes ;)

PR #1138, opened 2018-03-25
Redis & session timeout

### LINKED ISSUE #1101: Consider longer session timeout
Right now, sessions expire after [the default duration of two weeks](https://docs.djangoproject.com/en/dev/ref/settings/#std:setting-SESSION_COOKIE_AGE). We should consider setting that to several months so users need to login less often.

### LINKED ISSUE #1102: Investigate a proper caching backend like redis
Right now, we're using the [database backend](https://github.com/fsr-itse/EvaP/blob/master/evap/settings.py#L101) for the cache. We should test using [django-redis](https://github.com/niwinz/django-redis), how much page loading time that saves (if at all), and balance that against the setup and maintenance cost.

What would move to redis is the cache itself as well as the [session engine](https://docs.djangoproject.com/en/dev/ref/settings/#session-engine)

Compared to redis, the alternative [memcached](https://docs.djangoproject.com/en/2.0/topics/cache/#memcached) is supported natively by django, but offers no persistence, i.e. in case of a server restart the cached data including sessions would be lost. That's not that bad and we should try that out if redis is e.g. too much of a hassle to setup.

karyon:
see commit messages.

as discussed, having redis installed is now mandatory. run `apt-get install redis-server` to install it.

fixes #1102, fixes #1101

### DIFF
--- a/evap/settings.py
+++ b/evap/settings.py
@@ -106,9 +106,17 @@ DATABASES = {
 CACHES = {
     'default': {
         'BACKEND': 'django.core.cache.backends.db.DatabaseCache',
-        'LOCATION': 'evap_db_cache',
+        'LOCATION': 'evap_db_default_cache',
         'OPTIONS': {
-            'MAX_ENTRIES': 1000  # note that the results alone need one entry per course
+            'MAX_ENTRIES': 1000
+        }
+    },
+    'results': {
+        'BACKEND': 'django.core.cache.backends.db.DatabaseCache',
+        'LOCATION': 'evap_db_results_cache',
+        'TIMEOUT': None,  # is always invalidated manually
+        'OPTIONS': {
+            'MAX_ENTRIES': 100000
         }
     }
 }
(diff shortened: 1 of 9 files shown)


evap/settings.py
```diff
@@ -105,17 +105,19 @@
...
             'MAX_ENTRIES': 1000
         }
     },
     'results': {
-        'BACKEND': 'django.core.cache.backends.db.DatabaseCache',
-        'LOCATION': 'evap_db_results_cache',
-        'TIMEOUT': None,  # is always invalidated manually
+        'BACKEND': 'django_redis.cache.RedisCache',
+        'LOCATION': 'redis://127.0.0.1:6379/1',
+        'TIMEOUT': None,  # is always invalidated manually
```
janno42: there is a trailing space here! omg!

evap/settings.py
```diff
@@ -98,10 +105,20 @@
 
 CACHES = {
     'default': {
-        'BACKEND': 'django.core.cache.backends.db.DatabaseCache',
-        'LOCATION': 'evap_db_cache',
+        'BACKEND': 'django_redis.cache.RedisCache',
+        'LOCATION': 'redis://127.0.0.1:6379/1',
```
He3lixxx: Change this to id 0.

PR #1139, opened 2018-03-26
Random UI improvements

janno42:
course results page:
- `grade-bg-result-bar` width was increased to prevent line breaks for yes-no-questions with 100% in some browsers

staff semester page:
- ungraded reviewed courses now also get a red icon and the `Results not yet published although they probably could be` tooltip

semester results page:
- courses where results could not be published because of too few answers are now marked with an `eye-slash` icon instead of a `lock` icon
- private courses are now marked with a `lock` icon
- the `pause` icon for courses which have not been published have been replaced with the respective yellow and red `chart-bar` icons

### DIFF
--- a/evap/results/templates/results_semester_detail.html
+++ b/evap/results/templates/results_semester_detail.html
@@ -38,13 +38,23 @@
                                 {% if course|can_user_see_results:request.user %}data-url="{% url 'results:course_detail' semester.id course.id %}"{% endif %}>
                                     <td data-order="{{ course.name }}">
                                         {% if course.state == 'in_evaluation' %}
-                                            <span data-toggle="tooltip" data-placement="top" class="fas fa-play" title="{% trans 'This course is still in evaluation.' %}"></span>
+                                            <span data-toggle="tooltip" data-placement="top" class="fas fa-play" title="{% trans 'In evaluation' %}"></span>
                                         {% endif %}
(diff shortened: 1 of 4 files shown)


evap/results/templates/results_semester_detail.html
```diff
@@ -38,13 +38,23 @@ <h3>{{ semester.name }}</h3>
...
+                                        {% if course.state == 'evaluated' %}
+                                            <span data-toggle="tooltip" data-placement="top" class="fas fa-chart-bar icon-yellow" title="{% trans 'Results not yet published' %}"></span>
+                                        {% endif %}
+                                        {% if course.state == 'reviewed' %}
+                                            {% if course.is_single_result or course.final_grade_documents or course.gets_no_grade_documents or not course.is_graded %}
+                                                <span data-toggle="tooltip" data-placement="top" class="fas fa-chart-bar icon-red" title="{% trans 'Results not yet published although they probably could be' %}"></span>
+                                            {% else %}
+                                                <span data-toggle="tooltip" data-placement="top" class="fas fa-chart-bar icon-yellow" title="{% trans 'Results not yet published' %}"></span>
+                                            {% endif %}
```
karyon: isn't there a lot more of the icons from the staff semester view that could be put in here? either put all of that in here or the bare minimum, i would say.

Issue #1124, opened 2018-03-12
Misguiding messages on contributor's course edit page


He3lixxx:
When a course is already approved, the contributor course edit page shows a message saying you can't edit this course anymore (introduced in #1076).
Also, a message instructing the user to edit the data is shown. This is misleading. The second message shouldn't be there.

![grafik](https://user-images.githubusercontent.com/13838962/37304966-601f88f0-2633-11e8-8ca8-e49b3c07dc29.png)

@janno42 please make sure I'm not missing out anything here.

janno42:
Correct, this message should be removed for courses already approved by an editor.

PR #1140, opened 2018-03-26
fixed Misguiding messages on contributor's course edit page

phSch08:
#1124

### DIFF
--- a/evap/contributor/views.py
+++ b/evap/contributor/views.py
@@ -66,14 +66,10 @@ def settings_edit(request):
 def course_view(request, course_id):
     user = request.user
     course = get_object_or_404(Course, id=course_id)
-    show_edit_message = True
 
     # check rights
     if not (course.is_user_editor_or_delegate(user) and course.state in ['prepared', 'editor_approved', 'approved', 'in_evaluation', 'evaluated', 'reviewed']):
         raise PermissionDenied
-
-    if course.is_user_editor_or_delegate(user):
-        show_edit_message = False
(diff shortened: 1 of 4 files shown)


evap/contributor/views.py
```diff
@@ -84,7 +81,7 @@ def course_view(request, course_id):
         for field in cform.fields.values():
             field.disabled = True
 
-    template_data = dict(form=form, formset=formset, course=course, editable=False,
+    template_data = dict(form=form, formset=formset, course=course, editable=False, show_edit_message=False,
```
karyon: ...now after the simplification it seems that show_edit_message is now equivalent to `editable`, so as far as i can see you could just remove `show_edit_message` and use `editable`. could you verify that and implement it?

evap/contributor/views.py
```diff
@@ -71,9 +71,6 @@ def course_view(request, course_id):
     if not (course.is_user_editor_or_delegate(user) and course.state in ['prepared', 'editor_approved', 'approved', 'in_evaluation', 'evaluated', 'reviewed']):
         raise PermissionDenied
```
karyon: as both @He3lixxx and I needed longer than necessary to parse this condition, and the erroneous condition you removed probably came from that, could you change this condition to `if not ... or course.state not in ...`?

karyon:
i'd be glad if you joined us again on monday! if you can't make it in the next few weeks, let us know, then we would finish and merge the PR.

PR #1141, opened 2018-03-26
Ui fixes

pixunil:
grafik

### DIFF
--- a/evap/evaluation/templates/contact_modal.html
+++ b/evap/evaluation/templates/contact_modal.html
@@ -48,6 +48,8 @@
     function {{ modal_id }}Show() {
         $('#{{ modal_id }}ActionButton').unbind().click(function(){ {{ modal_id }}Action(); });
         $('#{{ modal_id }}').modal();
+        // remove wrongly added padding by Bootstrap (https://github.com/twbs/bootstrap/blob/v4-dev/js/src/modal.js#L429)
+        $('body, .fixed-top, .fixed-bottom, .is-fixed, .sticky-top').css('padding-right', '');
     }
     function {{ modal_id }}Action() {
         var actionButton = $('#{{ modal_id }}ActionButton');
--- a/evap/static/css/evap.scss
+++ b/evap/static/css/evap.scss
@@ -147,6 +147,16 @@ a.btn:not([href]):not(.disabled).btn-dark {
     display: inline-flex;
     background-color: lighten($light-gray, 2%);
     border-radius: 0.2rem;
+    .btn-toolbar {
+        background: $white;
+    }
+    .btn-group {
+        background: lighten($light-gray, 2%);
+        &::before {
+            width: 0.5rem;
+            content: "";
+        }
+    }
 }
 .btn-switch-label {
     display: inline-block;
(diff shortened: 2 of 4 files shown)


evap/evaluation/templates/contact_modal.html
```diff
@@ -48,6 +48,8 @@ <h5 class="modal-title" id="{{ modal_id }}Label">{{ title }}</h5>
     function {{ modal_id }}Show() {
         $('#{{ modal_id }}ActionButton').unbind().click(function(){ {{ modal_id }}Action(); });
         $('#{{ modal_id }}').modal();
+        // remove wrongly added padding by Bootstrap
```
janno42: can you please add a comment where exactly (which file, which line) bootstraps adds the padding?

evap/static/css/evap.scss
```diff
@@ -141,6 +141,19 @@ a.btn:not([href]):not(.disabled).btn-dark {
     display: inline-flex;
     background-color: lighten($light-gray, 2%);
     border-radius: 0.2rem;
+
```
janno42: please remove empty lines here and below

evap/static/css/evap.scss
```diff
@@ -141,6 +141,19 @@ a.btn:not([href]):not(.disabled).btn-dark {
...
+    .btn-toolbar {
+        background: $white;
+    }
+
+    .btn-group {
+        background: lighten($light-gray, 2%);
+
+        &::before {
+            width: $spacer * 0.5;
```
janno42: please use `width: 0.5rem;`

evap/evaluation/templates/contact_modal.html
```diff
@@ -48,6 +48,8 @@ <h5 class="modal-title" id="{{ modal_id }}Label">{{ title }}</h5>
     function {{ modal_id }}Show() {
         $('#{{ modal_id }}ActionButton').unbind().click(function(){ {{ modal_id }}Action(); });
         $('#{{ modal_id }}').modal();
+        // remove wrongly added padding by Bootstrap (https://github.com/twbs/bootstrap/blob/v4-dev/js/src/modal.js#L429)
+        $('body, .navbar, .footer').css('padding-right', '');
```
karyon: if you think that is a bug, please consider reporting it. there have been previous (closed) reports of this (e.g. [here](https://github.com/twbs/bootstrap/issues/24934)), but still, please consider it.

janno42: you can remove the `.footer` now, it's not fixed anymore.

janno42:
unfortunately you're breaking the language button now. can you please make sure that the background color won't be changed there?

evap/evaluation/templates/contact_modal.html
pixunil: I think it is no bug on their side, the `padding` just does not play well with a `overflow: scroll !important` on the body, so I think it is our fault here

pixunil: Changed to a more redundant, but future-proof solution  (same selectors like bootstrap)

Issue #571, opened 2015-04-20
Results UI rework


karyon:
this is a placeholder and reminder for the results UI rework.

(there is an etherpad for that)

janno42:
The idea is to have one page containing all the information (all courses, all degrees, all semesters) in one large table. This table can then be filtered by selecting one or multiple filter checkboxes (e.g. only courses from the last 4 semesters in Bachelor degree) and by the already existing text box (e.g. only courses by a certain responsible).

Because this list will be quite long, only the last N semesters could be initially selected. The results list needs to be cached to be loaded fastly.

Filter settings should be kept when opening the page again.

janno42:
implemented in #1149
🎉 🎆

PR #1149, opened 2018-04-03
Results page rework

janno42:
this is part of #571.
it's still work in progress and things left to do will be collected here.

- [x] save filter settings in browser
- [x] option to clear all filters
~~- [ ] sorting without datatables~~ won't do that in this pr
- [x] provide degrees and course types directly from the view instead of collecting them via javascript
- [x] show tooltips for single result distribution
- [x] shadow box effect of checkboxes in filter lists is cut on the left side once vertical scrollbar is added
- [x] change what is shown when not enough answers were given
- [x] exporter layout is likely broken after removing the deviations
~~- [ ] semester short names, degree names and course type names must not include `"` characters~~
- [x] remove archived semesters from semester list on results page
- [x] fix single results on results page
- [x] results rows don't link to detail pages and hover effect on distribution bar is missing

@karyon's stuff
- [ ] test performance for all user groups
- [ ] test performance on crappy laptop
- [ ] any correctness tests?
- [ ] write performance tests


evap/results/templates/results_course_detail.html
```diff
@@ -111,7 +111,9 @@ <h3>{{ course.name }} ({{ course.semester.name }})</h3>
                         </td>
                         {% if show_grades %}
                             <td class="text-center"><div class="grade-bg" style="background-color: {{ course.avg_grade|gradecolor }};">{{ course.avg_grade|floatformat:1 }}</div></td>
-                            <td class="text-center"><div class="deviation-bg" style="background-color: {{ course.avg_deviation|deviationcolor }};">{{ course.avg_deviation|floatformat:1 }}</div></td>
+                            <td class="text-center">
+                                {% include 'distribution_bar.html' with answer_counts=course.distribution total_count=course.distribution.0 warning=False %}
+                            </td>
                         {% else %}
                             <td class="text-center"><div class="grade-bg grade-bg-disabled" data-toggle="tooltip" data-placement="left" title="{% trans 'Not enough answers were given.' %}">&mdash;</div></td>
                             <td class="text-center"><div class="deviation-bg deviation-bg-disabled" data-toggle="tooltip" data-placement="left" title="{% trans 'Not enough answers were given.' %}">&mdash;</div></td>
```
karyon: this needs to be changed as well

evap/staff/views.py
```diff
@@ -410,7 +410,7 @@ def semester_raw_export(request, semester_id):
         _('#Participants'), _('#Comments'), _('Average grade')])
     for course in semester.course_set.all():
         degrees = ", ".join([degree.name for degree in course.degrees.all()])
-        course.avg_grade, course.avg_deviation = calculate_average_grades_and_deviation(course)
+        course.avg_grade, course.avg_deviation, __ = calculate_average_grades_and_deviation_and_distribution(course)
```
karyon: the distribution should not go into the raw export?

janno42: i didn't think about that yet. what would you say?

karyon: i don't know! i'm not the user :) i actually have no idea right now how the raw export is used at all.

evap/results/templates/results_course_detail.html
janno42: right. i added it to the list.

evap/staff/views.py
janno42: well... the user would be me i guess. the raw data export is used to create additional statistics - what we want to find changes every time, that's why the "raw" data is a good way to just get some numbers to fiddle around with.
but currently i can't think of a use case where we would need the distribution in there.

evap/static/css/evap.scss
```diff
@@ -1033,6 +1033,10 @@ a.no-underline:hover {
     text-decoration: none;
 }
 
+.semester-short-name {
+    word-spacing: 30000px;
+}
```
karyon: what does this do?

evap/results/templates/results_index.html
```diff
@@ -1,22 +1,84 @@
...
+                            <th style="width: 10%">{% trans 'Semester' %}</th>
+                            <th style="width: 22%">{% trans 'Responsible' %}</th>
+                            <th style="width: 13%">{% trans 'Voters' %}</th>
+                            <th style="width: 20%">{% trans 'Distribution and average grade' %}</th>
+                        </tr>
+                    </thead>
+                    <tbody>
+                    {% for course in courses %}
+                        <tr class="results-hover {% if course.is_single_result or not course|can_user_see_results_page:request.user %}nohover{% endif %}{% if course.state == 'in_evaluation' or course.state == 'evaluated' or course.state == 'reviewed' %} preview{% endif %}"
+                        {% if not course.is_single_result and course|can_user_see_results_page:request.user %}data-url="{% url 'results:course_detail' course.semester.id course.id %}"{% endif %}>
```
karyon: fix indentation please, also the condition can be put into a variable

evap/results/templates/results_index.html
```diff
@@ -1,22 +1,84 @@
...
+                                {{ course.semester.short_name }}
+                            </td>
+                            <td data-order="{{ course.responsible_contributors.0.last_name }}">
+                                {% for contributor in course.responsible_contributors %}
+                                    {{ contributor.full_name }}{% if not forloop.last %}, {% endif %}
+                                {% endfor %}
+                            </td>
+                            {% with num_participants=course.num_participants num_voters=course.num_voters %}
+                                {% if course.is_single_result %}
+                                    <td class="text-center" data-order="{{ num_voters|percentage_value:num_participants|default:'0' }}"><span class="fas fa-user"></span>&nbsp;{{ num_participants }}</td>
```
karyon: this percentage thing doesn't make sense for single results

evap/results/templates/results_index.html
```diff
@@ -44,21 +98,23 @@ <h3>{% trans 'Evaluation results' %}</h3>
                                 <span class="course-name">{{ course.name }}</span>
                                 <div>
                                     {% for degree in course.degrees.all %}
-                                        <span class="badge badge-primary">{{ degree }}</span>
+                                        <span class="badge badge-primary badge-degree">{{ degree }}</span>
```
karyon: some of these classes are used for styling, and another, similarly-looking class is used for identification, is that the way this is usually done?

evap/results/templates/results_index.html
```diff
@@ -84,5 +140,114 @@ <h3>{% trans 'Evaluation results' %}</h3>
...
+        $(document).ready(function() {
+            // load previously selected filters
+            $('input#search').val(localStorage['results_enteredSearchValue']);
+            if(localStorage.getItem('results_selectedSemesters') != null){
+                var selectedSemesters = JSON.parse(localStorage['results_selectedSemesters']);
+                selectedSemesters.forEach(function(value){
+                    $(`input[name=semester][value="${value}"]`).prop('checked', true);
+                });
+            } else {
+                localStorage['results_selectedSemesters'] = '[]';
```
karyon: is this necessary?

evap/results/views.py
```diff
@@ -21,16 +21,28 @@ def index(request):
...
         course.distribution = calculate_average_distribution(course)
         course.avg_grade = distribution_to_grade(course.distribution)
 
-    template_data = dict(courses=courses)
+
+    template_data = dict(
+        courses=courses,
+        degrees=sorted(degrees, key=lambda degree: degree.order),
+        course_types=sorted(course_types, key=lambda course_type: course_type.name),
+        semesters=Semester.objects.all(),
```
karyon: replacing this by `semesters=semesters` here should fix what we discussed

evap/static/css/evap.scss
```diff
@@ -1037,6 +1037,28 @@ a.no-underline:hover {
...
+.clear-filter-button:hover {
+    color: $dark-gray;
+}
+
+.result-filter-list {
+    max-height: 145px;
+    overflow-y: auto;
+    padding-left: 4px;
+}
```
karyon: into its own file? at the very least i would add a comment that marks this a s the "results index" section

evap/results/views.py
```diff
@@ -25,6 +25,8 @@ def index(request):
     courses = [course for course in courses if course.can_user_see_course(request.user)]
 
     for course in courses:
+        if course.is_single_result:
+            course.single_result_rating_result = get_single_result_rating_result(course)
         course.distribution = calculate_average_distribution(course)
         course.avg_grade = distribution_to_grade(course.distribution)
```
karyon: for single results, i think it should be `course.avg_grade should = course.single_result_rating_result.average` and `course.distribution` shouldn't be written at all.

either we decide to get the results of single results from `calculate_average_distribution` via some hard-coded indexing into the datastructure, or we leave it like you proposed, then i would assert in `calculate_average_distribution` and `collect_results` that they are not used for single results.

evap/static/css/evap.scss
karyon: ...it makes the lines *always* break there, right?

in any case, please add a comment along those lines.

evap/results/templates/results_index.html
janno42: Usually one would like to use an ID for identification but those should be unique and it doesn't really fit here. A class is a good solution to my mind because it's actually quite similar to what we do for styling. There we define that all objects of a certain class look like this and that and here we're using the class selector to get all objects with this class to collect their data. It might be better to only use the class `badge-degree` and let the css map its looks to what is defined in `badge-primary`, but I'd say that's not necessary. So - I would continue using it in this way.

evap/static/css/evap.scss
janno42: Yes, for consistency with different semester name lengths between ST and WT. Will do.

evap/results/templates/results_index.html
janno42: everything inside the for loop was moved to a different file, but i can still fix it in this commit.

do you mean putting `course.is_single_result or not course|can_user_see_results_page:request.user` into a variable? `or` is unfortunately not allowed in `with`, so it would have to be two variables. or did you mean something else?

evap/results/templates/results_index.html
janno42: seems like it can be removed.

evap/results/views.py
```diff
@@ -2,44 +2,74 @@
...
     TextAnswer, TextResult, RatingResult, HeadingResult, get_single_result_rating_result
 
 
-@internal_required
-def index(request):
-    semesters = Semester.get_all_with_published_unarchived_results()
+def delete_template_cache(course):
+    assert course.state != 'published'
+    make_template_fragment_key('course_result', [course.id, True])
+    make_template_fragment_key('course_result', [course.id, False])
```
janno42: i don't really understand what this is doing. on delete you create the keys?

evap/results/views.py
```diff
@@ -2,44 +2,74 @@
...
+        additional_courses = get_courses_with_prefetched_data(additional_courses)
+        courses += additional_courses
+
+    course_pks = [course.pk for course in courses]
+    degrees = Degree.objects.filter(courses__pk__in=course_pks).distinct()
+    course_types = CourseType.objects.filter(courses__pk__in=course_pks).distinct()
     template_data = dict(
         courses=courses,
-        degrees=sorted(degrees, key=lambda degree: degree.order),
+        degrees=degrees,
```
janno42: the order is not guaranteed now anymore, right? it would be nice to make sure they are in the correct order when being added to the filter list.

janno42: nevermind, we have the ordering defines in the model.

evap/evaluation/migrations/0076_add_semester_short_names.py
```diff
@@ -0,0 +1,53 @@
+# Generated by Django 2.0.2 on 2018-02-27 21:19
+
+from django.db import migrations, models
+
+
+def populate_semester_short_names(apps, schema_editor):
```
janno42: do we want to manually remove the `schema_editor` here?

karyon: i've done it in this PR but can't see us enforcing that... i don't know. we'll see.

evap/evaluation/tests/test_models.py
```diff
@@ -167,10 +167,11 @@ def test_responsible_contributors_ordering(self):
     def test_single_result_can_be_deleted_only_in_reviewed(self):
         responsible = mommy.make(UserProfile)
         course = mommy.make(Course, semester=mommy.make(Semester), is_single_result=True)
-        mommy.make(Contribution,
+        contribution = mommy.make(Contribution,
             course=course, contributor=responsible, responsible=True, can_edit=True, comment_visibility=Contribution.ALL_COMMENTS,
             questionnaires=[Questionnaire.single_result_questionnaire()]
         )
+        mommy.make(RatingAnswerCounter, answer=1, count=1, question=Questionnaire.single_result_questionnaire().question_set.all()[0], contribution=contribution)
```
janno42: `question=Questionnaire.single_result_questionnaire().question_set.first()`?

evap/results/templates/results_index.html
```diff
@@ -1,22 +1,195 @@
...
+                $(this).toggleClass("d-none", !showRow);
+            });
+        };
+
+        function uniqueValues(array){
+            var unique = array.filter(function(item, i, array) {
+                return i == array.indexOf(item);
+            });
+            return unique;
+        };
```
karyon: this seems to be unused

janno42:
this is now actually the last part to fix #571

evap/evaluation/models.py
```diff
@@ -348,36 +361,22 @@ def can_user_vote(self, user):
...
-    @property
-    def is_single_result(self):
-        # early return to save some queries
-        if self.vote_start_datetime.date() != self.vote_end_date:
-            return False
-
-        return self.contributions.filter(responsible=True, questionnaires__name_en=Questionnaire.SINGLE_RESULT_QUESTIONNAIRE_NAME).exists()
+        if not self.can_publish_rating_results or self.semester.results_are_archived or not self.can_user_see_course(user):
+            if not self.is_user_contributor_or_delegate(user):
+                return False
```
He3lixxx: What do you think about `return self.is_user_contributor_or_delegate(user)`
Using the fallthrough for the other case seems more like a "we do that because it's easier to code" and not a "we do that because it's semantically more true.". So, it would be more precise (imo) and shorter

evap/evaluation/models.py
```diff
@@ -609,6 +611,24 @@ def update_courses(cls):
...
+        warm_up_template_cache([instance])
+
+
+@receiver(post_transition, sender=Course)
+def delete_cache_on_unpublish(instance, source, **_kwargs):
+    if source == 'published':
+        from evap.results.tools import get_collect_results_cache_key
+        from evap.results.views import delete_template_cache
+        caches['results'].delete(get_collect_results_cache_key(instance))
+        delete_template_cache(instance)
```
He3lixxx: We've had bugs due to invalid cache clearing. Do we want to add a test for this?

evap/results/views.py
```diff
@@ -1,59 +1,83 @@
...
-    for course in courses:
-        if course.is_single_result:
-            for degree in course.degrees.all():
-                question_result = collect_results(course).questionnaire_results[0].question_results[0]
-                courses_by_degree[degree].single_results.append((course, question_result))
-        else:
-            for degree in course.degrees.all():
-                courses_by_degree[degree].courses.append(course)
+    if request.user.is_reviewer:
+        additional_courses = Course.objects.filter(semester__in=semesters, state__in=['in_evaluation', 'evaluated', 'reviewed'])
```
He3lixxx: Huh? Doesn't course.can_user_see_course return true in these cases if the user is a reviewer?

evap/results/views.py
```diff
@@ -1,59 +1,83 @@
...
 
-    template_data = dict(semester=semester, courses_by_degree=courses_by_degree)
-    return render(request, "results_semester_detail.html", template_data)
+    course_pks = [course.pk for course in courses]
+    degrees = Degree.objects.filter(courses__pk__in=course_pks).distinct()
+    course_types = CourseType.objects.filter(courses__pk__in=course_pks).distinct()
+    template_data = dict(
+        courses=courses,
+        degrees=degrees,
+        course_types=sorted(course_types, key=lambda course_type: course_type.name),
```
He3lixxx: Maybe do the ordering in the database? (-> order_by on the QuerySet)

evap/evaluation/models.py
janno42: there are asserts for caches being cleared after unpublishing. if you want to add some tests, i won't try to stop you :)

evap/results/views.py
janno42: yes, but above we're only looking for published courses

evap/results/views.py
janno42: we would need to check the language first and then decide to order on which column. here we're using the translated name already. while possible i think this doesn't really improve the current situation. what do you think?

evap/evaluation/models.py
janno42: well, actually there aren't asserts after deletion. so tests would be even better.

He3lixxx:
I wrote some tests for the caching and had these results:
- All our methods work as expected (well, that was not unexpected^^)
- The cache in evap/results/templates/results_index-course.html currently is transparent. If you open the results page and there is a cache entry missing for a course, it will simply calculate that cache value at that time, all following requests for that course will be cached.

We do not intend to have this template not cached for any course, that's why we create that cache on publishing and delete it on unpublishing. But, we use redis for caches, so there might be cases where a cached entry gets lost. Currently, this wouldn't be much of a problem, the system would fix itself as soon as necessary.
If however we end up with a wrong cache precalculation, this might mean that we create this cache for all courses when the server gets the first request for the results page, which could then take a while. It could also mean that we end up with all entries cached twice, with two keys.

What do we want to do here? I think leaving it as it is and just adding tests that make sure that our keys for warming up the cache and using it in the template match is a good approach.

Another thing we still have to do: Currently, our cache key does not contain the language code, which breaks localization.

janno42:
Yes, I would say adding these tests is sufficient. I'm on the translation caching.

He3lixxx:
I added tests, can be found at https://github.com/He3lixxx/EvaP/commit/2bb41a696cfac9c05836877f3650bce68b2d10a9. I can push it into this branch too, I think, but since the tests will have to be adapted when we change the cache key to also include the language, I guess it's better to wait for that.

He3lixxx:
@karyon if you have a clue why Codacy complains here, this would be a real help. We are pretty sure by manual and automatic testing that these imports are ok.

Issue #991, opened 2017-07-31
Dropped courses


janno42:
Participants should be able to state that they dropped a course instead of starting its evaluation. In this case, a form should be shown, asking about the reasons for dropping the course. Submitting the form will do the following:

- The user will be removed from the list of participants.
- Reward points will be granted if the user evaluated all other required courses in this semester (the email to the admins will include a notification if points were granted by this action).
- The form data will be sent to the admins.

janno42:
Forget all of the above.
A course's voting page will have a button "I dropped this course". This will add another section at the top with questions about why the participant dropped the course (Likert about most frequent reasons and free text). "No answer" will be selected for all other questions of the course's questionnaire.
On the student's semester page next to the "evaluate" button of each course a button "I dropped this course" will be added - this will have the same behavior as opening the questionnaire normally and then clicking the "dropped course" button.

The results of the special questionnaire will be shown only to staff, reviewers and responsible contributors. The participants will see an explanatory text on top of the questionnaire during the vote stating this information as well as an invitation that all other questions below can still be answered.

He3lixxx:
For anyone tackling this: It may be helpful to browse through the old attempt to implement this in #1153. Some parts of this implementation are probably still usable. Make sure to have the design considerations from https://github.com/fsr-de/EvaP/pull/1153#pullrequestreview-112547210 in mind.

fekoch:
I would like to try tackling this

fekoch:
@janno42 
> Forget all of the above. A course's voting page will have a button "I dropped this course". 

Should this button be shown for all courses? Or only for specific course types? Or should this be a selectable option at the time of the course creation?

fekoch:
(@janno42) As discussed, we want:

- [ ] A check-box "Allow drop-out?" when creating new evaluations
- [ ] A setting "DROPOUT_QUESTIONNAIRE_ID" to set which questionnaire should be shown when using the drop-out function.
- [ ] The "I dropped this course" button should be shown only on the student index and only for evaluations that "allow drop-out"

PR #1153, opened 2018-04-09
Added dropped courses for #991

jakobmache:
Add dropped courses for issue #991


evap/student/templates/student_vote.html
```diff
@@ -30,16 +32,39 @@
...
+                <b>{% trans 'Questionnaire Preview' %}</b><br />
+                {% blocktrans %}This is a preview of the questionnaire for the course. Students will see the questions below when participating in the evaluation.{% endblocktrans %}
+            </div>
+        </small>
     {% endif %}
     <h3 class="mb-3">{{ course.name }} ({{ course.semester.name }})</h3>
     <form id="student-vote-form" method="POST" class="form-horizontal">
         {% csrf_token %}
 
+        {% if not preview %}
```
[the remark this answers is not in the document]
janno42: showing a disabled button in the card on the preview page should be easy. or would you like it to be functional as in opening the card and showing the dropout questionnaire?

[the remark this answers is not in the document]
karyon: i think the minimum is that the page *looks* exactly like it looks to students, that's kind of the point of a preview.

whether or not the page should *work* exactly like it does to students, i'm not sure... in this case it might be nice to actually show the dropout questionnaire. if you think of it as not really doing anything but rather showing another part of the page, it's a part of the preview. however, i don't have strong feeling there and if it's too much work, we can postpone that or leave it out.

evap/student/templates/student_vote.html
```diff
@@ -168,6 +188,21 @@ <h3 class="mb-3">{{ course.name }} ({{ course.semester.name }})</h3>
...
+            function markUnansweredAsNoAnswerFor(contributorId) {
+                rows = $("#vote-area-" + contributorId).find(".vote-row");
+                for (i = 0; i < rows.length; i++) {
+                    row = $(rows[i]);
+                    vote_inputs = row.find(".vote-inputs [type=radio]");
+
+                    if (!vote_inputs.parent().hasClass("active")) {
+                        row.find("[value=6]").prop('checked', true);
+                        row.find("[value=6]").parent().addClass("active");
+                        sisyphus.saveAllData();
```
He3lixxx: In case this method will be kept (it's called in 250 as far as i can see): Is this call really necessary once per row or can we just do it once after the loop completed?

evap/student/templates/student_vote.html
```diff
@@ -202,6 +242,30 @@ <h3 class="mb-3">{{ course.name }} ({{ course.semester.name }})</h3>
...
+
+                var contributorIds = [
+                {% for contributor, __ , __ , __ in contributor_form_groups %}
+                    {{ contributor.id }},
+                {% endfor %}
+                ];
+
+                for(var i = 0; i <  contributorIds.length; i++) {
+                    markNoAnswerFor(contributorIds[i])
+                }
```
He3lixxx: I understand that you're using building on top of an older function here, but passing in the IDs in the template generation to call the function for all ids feels wrong for me. What do you think about giving all contributor-cards that currently have a "#vote-area-X"-id a common class and writing a javascript method that iterates over all these divs by class? Imo that would be cleaner. I don't have much JS experience though. @karyon what do you think?

evap/student/views.py
```diff
@@ -89,7 +94,9 @@ def vote(request, course_id):
             minutes_left_for_evaluation=(course.time_left_for_evaluation.seconds//60)%60,
             success_magic_string=SUCCESS_MAGIC_STRING,
             success_redirect_url=reverse('student:index'),
-            evaluation_ends_soon=course.evaluation_ends_soon())
+            evaluation_ends_soon=course.evaluation_ends_soon(),
+            dropped=False)
```
He3lixxx: Is this unconditionally always false?

He3lixxx:
@jakobmache is it okay for you if we close this PR and add a comment in the issue for anyone attempting that there was already this pr?

He3lixxx:
I left a note about this PR in the issue. Closing as this will probably not be finished.

Issue #1094, opened 2018-02-02
Continuous text answer review


janno42:
When reviewing text answers, it should be possible to faster approve individual answers. Currently two ideas exist for the reviewing UI:
- Either approved answers are removed so that the next answer and its yes and no buttons move up to the exact same position where the old answer was. This allows to approve answers without moving the mouse every time.
- Another possibility (which would be even better) is to only show a single text answer at a time and to approve/hide it by pressing a key on the keyboard. After each choice the next answer is shown.

Additionally, after all text answers for a course have been reviewed, it should be possible to easily switch to the next course (e.g. navigation button to "next course" or another hotkey).

karyon:
@janno42, @pixunil you guys keep in mind that #1130 needs to be integrated somehow?

PR #1154, opened 2018-04-13
Add a quick interface to review text answers continously

pixunil:
screenshot

Addresses #1094 

### Changelog since last review

- [x] Refactor JS away from pseudoclasses
- [x] Tooltip for private button
- [x] More robust solution for overflow detection (scrollable by arrow keys)
- [x] More robust solution for handling mouse and keyboard events
- [x] Fixed breadcrumbs
- [x] Nicer greeting on courses with no text answers
- [x] Better magic for next course
- [x] Fixed the bug no one cared about (Untoggle a review button with mouse will keep the focus on this button)
- [x] Edit links
- [x] Refactor of `update` functions
- [x] Fix layers sliding in wrong direction when keeping keys pressed (cause known)
- [x] Clear discussion about sql
- [x] Make hotkeys and click events tied together (Still needs discussion)

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -820,6 +820,10 @@ class TextAnswer(Answer):
     def is_published(self):
         return self.state == self.PUBLISHED
 
+    @property
+    def is_reviewed(self):
+        return self.state != self.NOT_REVIEWED
+
     def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
         assert self.answer != self.original_answer
--- a/evap/staff/templates/quick-review.js
+++ b/evap/staff/templates/quick-review.js
@@ -1,0 +1,184 @@
+$(document).ready(() => {
+    var slider = $(".slider");
+    var items = [];
+    var index = 0;
+
+    slider.on("transitionend", ".slider-item", event => {
+        updateButtons();
+        $(event.target).removeClass("to-left to-right")
+            .css({top: "", height: ""});
+    });
+
+    slider.on("click", "[data-action]", event => {
+        reviewAction($(event.target).data("action"));
+    });
+    slider.on("click", "[data-slide]", event => {
+        let offset = $(event.target).data("slide") === "left" ? -1 : 1;
+        slideTo(index + offset);
+        updateButtons();
+    });
+    slider.on("click", "[data-startover]", event => {
+        startOver($(event.target).data("startover"));
+    });
+
+    $(document).keydown(event => {
+        let actions = {
+            37: "[data-slide=left]",          // arrow left
+            39: "[data-slide=right]",         // arrow right
+            74: "[data-action=publish]",      // j
+            75: "[data-action=make_private]", // k
+            76: "[data-action=hide]",         // l
+             8: "[data-action=unreview]",     // backspace
+            13: "[data-url=next-course]",     // enter
+            77: "[data-startover=unreviewed]",// m
+            78: "[data-startover=all]"        // n
+        };
+
+        if(actions[event.which] && !event.originalEvent.repeat) {
+            let element = slider.find(actions[event.which]);
+            if(element.length) {
+                element[0].click();
+            }
+            event.preventDefault();
+        }
+    });
+
+    startOver("unreviewed");
+
+    function slideTo(requestedIndex) {
+        requestedIndex = Math.min(Math.max(requestedIndex, 0), items.length);
+        let direction = requestedIndex >= index ? "right" : "left";
+        index = requestedIndex;
+
+        if(index === items.length) {
+            let reviewed = items.filter(":not([data-review])").length === 0;
+            let alert = slider.find(".alert");
+            alert.toggleClass("alert-warning", !reviewed);
+            alert.toggleClass("alert-success", reviewed && items.length !== 0);
+            alert.toggleClass("alert-secondary", reviewed && items.length === 0);
+            alert.find("[data-content=unreviewed]").toggleClass("d-none", reviewed);
+            alert.find("[data-content=reviewed]").toggleClass("d-none", !reviewed);
+            slider.find("[data-action=startover-unreviewed]").toggleClass("d-none", reviewed);
+            slider.find("[data-action=startover-all]").toggleClass("d-none", items.length === 0);
+            slideLayer(2, direction, alert);
+        } else {
+            slideLayer(2, direction, items.eq(index));
+        }
+
+        slider.find(".slider-side-left").toggleClass("shown", index > 0);
+        let reviewed = items.slice(0, index).filter("[data-review]").length;
+        slider.find("[data-counter=reviewed-left]").text(reviewed);
+        slider.find("[data-counter=unreviewed-left]").text(index - reviewed);
+
+        slider.find(".slider-side-right").toggleClass("shown", index < items.length);
+        reviewed = items.slice(index + 1).filter("[data-review]").length;
+        slider.find("[data-counter=reviewed-right]").text(reviewed);
+        let unreviewed = items.length - index - reviewed - 1;
+        slider.find("[data-counter=unreviewed-right]").text(Math.max(unreviewed, 0));
+    }
+
+    function slideLayer(layer, direction, element) {
+        // to preserve the vertical positions and heights during transition,
+        // the elements will be deactivated from bottom to top and activated from top to bottom
+
+        if(element && element.is(".active")) {
+            return;
+        }
+
+        let last = slider.find("[data-layer=" + layer + "].active, .alert.active");
+        if(last.length !== 0) {
+            let reversed = direction === "left" ? "right" : "left";
+            last.css({top: last.position().top, height: last.outerHeight()});
+            last.removeClass("active to-left to-right");
+            last.addClass("to-" + reversed);
+        }
+
+        if(layer > 0) {
+            if(index < items.length) {
+                let reference = element.prevAll("[data-layer=" + (layer - 1) + "]").first();
+                slideLayer(layer - 1, direction, reference);
+            } else {
+                slideLayer(layer - 1, direction, null);
+            }
+        }
+
+        if(element !== null) {
+            element.removeClass("to-left to-right");
+            element.position();
+            element.addClass("to-" + direction);
+            element.position();
+            element.addClass("active");
+        }
+    }
+
+    function startOver(action) {
+        let reviewed = slider.find("[data-layer=2][data-review]");
+        let unreviewed = slider.find("[data-layer=2]:not([data-review])");
+        let index = action === "unreviewed" && unreviewed.length ? reviewed.length : 0;
+        items = $.merge(reviewed, unreviewed);
+        slideTo(index);
+        updateButtons();
+    }
+
+    function reviewAction(action) {
+        if(index === items.length || action === "make_private" && !items.eq(index).data("contribution")) {
+            return;
+        }
+
+        let active = items.eq(index);
+        let parameters = {"id": active.data("id"), "action": action, "course_id": {{ course.id }}};
+        $.ajax({
+            type: "POST",
+            url: "{% url 'staff:course_comments_update_publish' %}",
+            data: parameters,
+            error: function(){ window.alert("{% trans 'The server is not responding.' %}"); }
+        });
+
+        if(action === "unreview") {
+            active.removeAttr("data-review");
+        } else {
+            active.attr("data-review", action);
+        }
+
+        updateButtonActive();
+
+        if(action !== "unreview") {
+            slideTo(index + 1);
+            slider.find("[data-action=" + action + "]").focus();
+        }
+    }
+
+    function updateButtons() {
+        slider.find("[data-action-set=reviewing]").toggleClass("d-none", index === items.length);
+        slider.find("[data-action-set=summary]").toggleClass("d-none", index < items.length);
+
+        if(index < items.length) {
+            updateButtonActive();
+            let notContributor = !items.eq(index).data("contribution");
+            let privateBtn = slider.find("[data-action=make_private]");
+            privateBtn.attr("disabled", notContributor);
+            privateBtn.tooltip(notContributor ? 'enable' : 'disable');
+        }
+    }
+
+    function updateButtonActive() {
+        let active = items.eq(index);
+        // focus the answer to enable scrolling with arrow keys
+        active.focus();
+        slider.find(".btn:focus").blur();
+
+        let actions = {
+            publish: "btn-success",
+            make_private: "btn-dark",
+            hide: "btn-danger"
+        };
+
+        let review = active.attr("data-review");
+        for(let action in actions) {
+            let btn = slider.find("[data-action=" + action + "]");
+            btn.toggleClass(actions[action], review === action);
+            btn.toggleClass("btn-outline-secondary", review !== action);
+        }
+        slider.find("[data-action=unreview]").attr("disabled", !review);
+    }
+});
--- a/evap/staff/templates/staff_course_comments_quick.html
+++ b/evap/staff/templates/staff_course_comments_quick.html
@@ -1,0 +1,138 @@
+{% extends 'staff_course_comments.html' %}
+
+{% block content %}
+    {{ block.super }}
+
+    <div class="modal fade" id="hotkeys-modal" tabindex="-1" role="dialog" aria-labelledby="hotkeys-modal-title" aria-hidden="true">
+        <div class="modal-dialog" role="document">
+            <div class="modal-content">
+                <div class="modal-header">
+                    <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
+                    <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                        <span aria-hidden="true">&times;</span>
+                    </button>
+                </div>
+                <table class="modal-body table table-striped vertically-aligned">
+                    <tr><td><span class="key">←</span> <span class="key">→</span></td><td>{% trans 'Navigate through answers' %}</td></tr>
+                    <tr><td><span class="key">J</span></td><td>{% trans 'Publish answer' %}</td></tr>
+                    <tr><td><span class="key">K</span></td><td>{% trans 'Make answer private to the contributor' %}</td></tr>
+                    <tr><td><span class="key">L</span></td><td>{% trans 'Hide answer' %}</td></tr>
+                    <tr><td><span class="key">&#x232b;</span></td><td>{% trans 'Unreview answer' %}</td></tr>
+                    <tr><td><span class="key">&#x21b2;</span></td><td>{% trans 'Review next course' %}</td></tr>
+                    <tr><td><span class="key">N</span></td><td>{% trans 'Show all again' %}</td></tr>
+                    <tr><td><span class="key">M</span></td><td>{% trans 'Show unreviewed' %}</td></tr>
+                </table>
+                <div class="modal-footer">
+                    <button type="button" class="btn btn-light" data-dismiss="modal">Close</button>
+                </div>
+            </div>
+        </div>
+    </div>
+
+    <div class="card card-outline-primary slider">
+        <div class="card-header">
+            {% trans 'Comments' %}
+            <span class="hotkey-btn" aria-hidden="true" data-toggle="tooltip" data-placement="right" title="{% trans 'Hotkeys' %}">
+                <span class="far fa-keyboard" data-toggle="modal" data-target="#hotkeys-modal"></span>
+            </span>
+        </div>
+        <div class="card-body slider-inner px-0">
+            <div class="slider-side slider-side-left">
+                <span class="badge badge-pill badge-primary" data-counter="reviewed-left" data-toggle="tooltip" data-placement="left" title="{% trans 'Reviewed comments' %}"></span>
+                <span class="fas fa-angle-left" data-slide="left"></span>
+                <span class="badge badge-pill badge-secondary" data-counter="unreviewed-left" data-toggle="tooltip" data-placement="left" title="{% trans 'Unreviewed comments' %}"></span>
+            </div>
+            <div class="slider-items">
+                {% for questionnaire, contributor, label, is_responsible, results in sections %}
+                    <div class="slider-item" data-layer="0">
+                        {{ questionnaire.public_name }}{% if contributor %}: {{ contributor.full_name }}{% if is_responsible %} ({% trans 'responsible' %}){% endif %}{% if label %} &ndash; <i>{{ label }}</i>{% endif %}{% endif %}
+                    </div>
+                    {% for result in results %}
+                        <div class="slider-item" data-layer="1">
+                            {{ result.question.text}}
+                        </div>
+                        {% for answer in result.answers %}
+                            <div class="slider-item" data-layer="2" data-id="{{ answer.id }}"{% if contributor %} data-contribution="yes"{% endif %}
+                                {% if answer.is_reviewed %}data-review="{% if answer.is_published %}publish{% elif answer.is_private %}make_private{% else %}hide{% endif %}"{% endif %}
+                            >
+                            <a class="comment-link" href="{% url 'staff:course_comment_edit' semester.id course.id answer.id %}">{{ answer.answer|linebreaksbr }}</a>
+                            {% if answer.original_answer %}
+                                <span class="comment-original">({{ answer.original_answer|linebreaksbr }})</span>
+                            {% endif %}
+                            </div>
+                        {% endfor %}
+                    {% endfor %}
+                {% endfor %}
+                <div class="slider-item alert text-center">
+                    {% if sections %}
+                        <span data-content="unreviewed">
+                            {% trans 'Some comments for this course are still unreviewed.' %}
+                        </span>
+                        <span data-content="reviewed">
+                            {% trans 'You have reviewed all comments for this course.' %}
+                        </span>
+                    {% else %}
+                        <span>
+                            {% trans 'There are no comments for this course.' %}
+                        </span>
+                    {% endif %}
+                    {% if next_course %}
+                        <span>
+                            {% blocktrans with name=next_course.name count answers=next_course.num_unreviewed_textanswers %}
+                                The next course "{{ name }}" has got {{ answers }} unreviewed text answer.
+                            {% plural %}
+                                The next course "{{ name }}" has got {{ answers }} unreviewed text answers.
+                            {% endblocktrans %}
+                        </span>
+                    {% endif %}
+                </div>
+            </div>
+            <div class="slider-side slider-side-right">
+                <span class="badge badge-pill badge-primary" data-counter="reviewed-right" data-toggle="tooltip" data-placement="right" title="{% trans 'Reviewed comments' %}"></span>
+                <span class="fas fa-angle-right" data-slide="right" title=""></span>
+                <span class="badge badge-pill badge-secondary" data-counter="unreviewed-right" data-toggle="tooltip" data-placement="right" title="{% trans 'Unreviewed comments' %}"></span>
+            </div>
+        </div>
+        <div class="card-footer d-flex">
+            <div class="lcr-left"></div>
+            <div class="lcr-center" data-action-set="reviewing">
+                <button role="button" data-action="publish" class="btn btn-sm btn-outline-secondary">
+                    {% trans 'yes' %}
+                </button>
+                <button role="button" data-action="make_private" class="btn btn-sm btn-outline-secondary" disabled=""
+                    title="{% blocktrans %}This answer is for a course question and can't be made private.{% endblocktrans %}" data-toggle="tooltip">
+                    {% trans 'private' %}
+                </button>
+                <button role="button" data-action="hide" class="btn btn-sm btn-outline-secondary">
+                    {% trans 'no' %}
+                </button>
+            </div>
+            <div class="lcr-center d-none" data-action-set="summary">
+                {% if next_course %}
+                    <a href="{% url 'staff:course_comments' next_course.semester.id next_course.id %}" data-url="next-course" class="btn btn-sm btn-primary">
+                        {% trans 'Review next course' %}
+                    </a>
+                {% endif %}
+                <button role="button" data-startover="all" class="btn btn-sm btn-outline-primary">
+                    {% trans 'Show all again' %}
+                </button>
+                <button role="button" data-startover="unreviewed" class="btn btn-sm btn-outline-primary">
+                    {% trans 'Show unreviewed' %}
+                </button>
+            </div>
+            <div class="lcr-right">
+                <div data-action-set="reviewing">
+                    <button role="button" data-action="unreview" class="btn btn-sm btn-outline-secondary">
+                        {% trans 'unreview' %}
+                    </button>
+                </div>
+            </div>
+        </div>
+    </div>
+{% endblock %}
+
+{% block additional_javascript %}
+    <script type="text/javascript">
+        {% include "quick-review.js" %}
+    </script>
+{% endblock %}
--- a/evap/staff/tools.py
+++ b/evap/staff/tools.py
@@ -9,11 +9,12 @@ from django.core.cache import cache
 from django.core.cache.utils import make_template_fragment_key
 from django.core.exceptions import SuspiciousOperation
 from django.db import transaction
+from django.db.models import Count
 from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
 from django.utils.safestring import mark_safe
 
-from evap.evaluation.models import UserProfile, Course, Contribution
+from evap.evaluation.models import UserProfile, Course, Contribution, TextAnswer
 from evap.grades.models import GradeDocument
 from evap.results.tools import collect_results
 
@@ -183,3 +184,12 @@ def merge_users(main_user, other_user, preview=False):
     other_user.delete()
 
     return merged_user, errors, warnings
+
+
+def find_next_unreviewed_course(semester, excluded):
+    return semester.course_set.exclude(pk__in=excluded) \
+        .exclude(state='published') \
+        .exclude(can_publish_text_results=False) \
+        .filter(contributions__textanswer_set__state=TextAnswer.NOT_REVIEWED) \
+        .annotate(num_unreviewed_textanswers=Count("contributions__textanswer_set")) \
+        .order_by('vote_end_date', '-num_unreviewed_textanswers').first()
--- a/evap/staff/views.py
+++ b/evap/staff/views.py
@@ -34,7 +34,7 @@ from evap.staff.forms import (AtLeastOneFormSet, ContributionForm, ContributionF
 from evap.staff.importers import EnrollmentImporter, UserImporter, PersonImporter
 from evap.staff.tools import (bulk_delete_users, custom_redirect, delete_import_file, delete_navbar_cache_for_users,
                               forward_messages, get_import_file_content_or_raise, import_file_exists, merge_users,
-                              save_import_file, get_parameter_from_url_or_session)
+                              save_import_file, get_parameter_from_url_or_session, find_next_unreviewed_course)
 from evap.student.forms import QuestionnaireVotingForm
 from evap.student.views import get_valid_form_groups_or_render_vote_page
 
@@ -796,7 +796,8 @@ def course_comments(request, semester_id, course_id):
     if not course.can_publish_text_results:
         raise PermissionDenied
 
-    filter_comments = get_parameter_from_url_or_session(request, "filter_comments")
+    view = request.GET.get('view', 'quick')
+    filter_comments = view == "unreviewed"
 
     CommentSection = namedtuple('CommentSection', ('questionnaire', 'contributor', 'label', 'is_responsible', 'results'))
     course_sections = []
@@ -815,9 +816,23 @@ def course_comments(request, semester_id, course_id):
             section_list = course_sections if contribution.is_general else contributor_sections
             section_list.append(CommentSection(questionnaire, contribution.contributor, contribution.label, contribution.responsible, text_results))
 
-    template_data = dict(semester=semester, course=course, course_sections=course_sections,
-            contributor_sections=contributor_sections, filter_comments=filter_comments)
-    return render(request, "staff_course_comments.html", template_data)
+    template_data = dict(semester=semester, course=course, view=view)
+
+    if view == 'quick':
+        visited = request.session.get('review-visited', set())
+        visited.add(course.pk)
+        next_course = find_next_unreviewed_course(semester, visited)
+        if not next_course and len(visited) > 1:
+            visited = {course.pk}
+            next_course = find_next_unreviewed_course(semester, visited)
+        request.session['review-visited'] = visited
+
+        sections = course_sections + contributor_sections
+        template_data.update(dict(sections=sections, next_course=next_course))
+        return render(request, "staff_course_comments_quick.html", template_data)
+    else:
+        template_data.update(dict(course_sections=course_sections, contributor_sections=contributor_sections))
+        return render(request, "staff_course_comments_full.html", template_data)
 
 
 @require_POST
--- a/evap/static/css/_quick-review.scss
+++ b/evap/static/css/_quick-review.scss
@@ -1,0 +1,94 @@
+.slider-inner {
+    display: flex;
+    overflow: hidden;
+}
+.slider-side {
+    width: 3rem;
+    flex: none;
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
+    // slider controls should not be overlapped by sliding slides
+    z-index: 5;
+    opacity: 0;
+    transition: 0.2s opacity;
+    &.shown {
+        opacity: 1;
+        .fas {
+            cursor: pointer;
+        }
+    }
+    .fas {
+        margin: 0.2rem 0;
+        font-size: 2rem;
+        color: grey;
+        transition: 0.2s opacity;
+    }
+}
+.slider-items {
+    height: 24rem;
+    flex: auto;
+    position: relative;
+    display: flex;
+    flex-direction: column;
+}
+.slider-item {
+    width: 100%;
+    flex: none;
+    position: absolute;
+    display: none;
+    overflow-y: auto;
+    opacity: 0;
+    transition: transform, opacity;
+    transition-duration: 0.2s;
+    &.alert, &[data-layer="2"] {
+        flex: auto;
+    }
+    &.alert {
+        margin: 0;
+        justify-content: center;
+    }
+    &.to-left, &.to-right, &.active {
+        display: flex;
+        flex-direction: column;
+    }
+    &.to-left {
+        transform: translateX(-10%);
+    }
+    &.to-right {
+        transform: translateX(10%);
+    }
+    &.active {
+        position: static;
+        transform: translateX(0);
+        opacity: 1;
+    }
+    &:focus {
+        outline: none;
+    }
+
+}
+[data-layer="0"] {
+    @extend .card-header, .border, .rounded-top;
+}
+[data-layer="1"] {
+    @extend .card-body, .bg-light, .border-left, .border-right;
+}
+[data-layer="2"] {
+    @extend .card-body, .border, .rounded-bottom;
+}
+.hotkey-btn {
+    float: right;
+    cursor: pointer;
+}
+.key {
+    background: $darker-gray;
+    color: $white;
+    border-radius: 0.2rem;
+    display: inline-block;
+    width: 2rem;
+    height: 2rem;
+    text-align: center;
+    line-height: 2rem;
+}
--- a/evap/static/css/evap.scss
+++ b/evap/static/css/evap.scss
@@ -45,6 +45,12 @@ div.container {
     flex: 1 0 auto;
 }
 
+// remove wrongly added padding by Bootstrap, caused by the always visible scrollbar
+// (https://github.com/twbs/bootstrap/blob/v4-dev/js/src/modal.js#L429)
+body, .fixed-top, .fixed-bottom, .is-fixed, .sticky-top {
+    padding-right: 0 !important;
+}
+
 .btn.fas {
     font-weight: 900;
 }
--- a/evap/static/css/evap.scss
+++ b/evap/static/css/evap.scss
@@ -1,5 +1,6 @@
 @import "evap-variables";
 @import "../bootstrap/scss/bootstrap";
+@import "quick-review";
 
 
 /* override bootstrap values */
@@ -43,6 +44,18 @@ h1, h2, h3, h4, h5, h6 {
 div.container {
     padding-top: 80px; // for navbar (and breadcrumbs)
     flex: 1 0 auto;
+}
+
+.lcr-left {
+    flex: 1;
+}
+.lcr-center {
+    flex: 5;
+    text-align: center;
+}
+.lcr-right {
+    flex: 1;
+    text-align: right;
 }
 
 // remove wrongly added padding by Bootstrap, caused by the always visible scrollbar
@@ -310,7 +323,7 @@ a.btn:not([href]):not(.disabled).btn-dark {
     margin-bottom: 0;
 }
 
-button {
+button:not(:disabled) {
     cursor: pointer;
 }
 
--- a/evap/evaluation/templates/contact_modal.html
+++ b/evap/evaluation/templates/contact_modal.html
@@ -48,8 +48,6 @@
     function {{ modal_id }}Show() {
         $('#{{ modal_id }}ActionButton').unbind().click(function(event){ {{ modal_id }}Action(event); });
         $('#{{ modal_id }}').modal();
-        // remove wrongly added padding by Bootstrap (https://github.com/twbs/bootstrap/blob/v4-dev/js/src/modal.js#L429)
-        $('body, .fixed-top, .fixed-bottom, .is-fixed, .sticky-top').css('padding-right', '');
     }
     function {{ modal_id }}Action(event) {
         var actionButton = $('#{{ modal_id }}ActionButton');
--- a/evap/staff/templates/staff_course_comments.html
+++ b/evap/staff/templates/staff_course_comments.html
@@ -13,13 +13,16 @@
     <div class="d-flex">
         <h3 class="mr-auto">{{ course.name }} ({{ course.semester.name }})</h3>
         <div class="btn-switch my-auto d-print-none">
-            <div class="btn-switch-label">{% trans 'Reviewed comments' %}</div>
+            <div class="btn-switch-label">{% trans 'View' %}</div>
             <div class="btn-group">
-                <a href="{% url 'staff:course_comments' semester.id course.id %}?filter_comments=false" role="button" class="btn btn-sm btn-light{% if not filter_comments %} active{% endif %}">
-                    {% trans 'Show' %}
+                <a href="{% url 'staff:course_comments' semester.id course.id %}" role="button" class="btn btn-sm btn-light{% if view == 'quick' %} active{% endif %}">
+                    {% trans 'Quick' %}
                 </a>
-                <a href="{% url 'staff:course_comments' semester.id course.id %}?filter_comments=true" role="button" class="btn btn-sm btn-light{% if filter_comments %} active{% endif %}">
-                    {% trans 'Hide' %}
+                <a href="{% url 'staff:course_comments' semester.id course.id %}?view=full" role="button" class="btn btn-sm btn-light{% if view == 'full' %} active{% endif %}">
+                    {% trans 'Full' %}
+                </a>
+                <a href="{% url 'staff:course_comments' semester.id course.id %}?view=unreviewed" role="button" class="btn btn-sm btn-light{% if view == 'unreviewed' %} active{% endif %}">
+                    {% trans 'Unreviewed' %}
                 </a>
             </div>
         </div>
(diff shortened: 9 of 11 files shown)


evap/static/css/evap.scss
```diff
@@ -1259,3 +1259,65 @@ table.dataTable thead>tr>th.sorting_asc, table.dataTable thead>tr>th.sorting_des
         flex: 0 0 40%;
     }
 }
+
+
+/* interactive slider */
```
janno42: `interactive text answer review slider` should make it easier to imagine the respective element

evap/static/css/evap.scss
```diff
@@ -1259,3 +1259,65 @@ table.dataTable thead>tr>th.sorting_asc, table.dataTable thead>tr>th.sorting_des
...
+    margin: 1.25rem;
+    top: 0;
+    left: 100%;
+    transition: top 0.5s, left 0.5s, border 0.5s;
+    position: absolute;
+
+    .card-body {
+        overflow-y: auto;
+    }
+
```
janno42: please remove the empty lines before the `&` entries

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,285 @@
...
+{% block breadcrumb %}
+    {{ block.super }}
+    <li class="breadcrumb-item">{% trans 'Comments' %}</li>
+{% endblock %}
+
+{% block content %}
+    {{ block.super }}
+
+    <div class="d-flex">
+        <h3 class="mr-auto">{{ course.name }} ({{ course.semester.name }})</h3>
```
janno42: can you please move this part and the one from `staff_course_comments.html` to a common base template?

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,285 @@
...
+                    {% trans 'Some comments are still unreviewed.' %}
+                </p>
+            </div>
+            <div class="alert alert-success slider-item text-center">
+                <p>
+                    {% trans 'You have reviewed all comments on this course.' %}
+                </p>
+            {% if next_course %}
+                <p>
+                    The course {{ next_course.name }} has got {{ next_course.num_unreviewed_textanswers }} unreviewed text answers.
```
janno42: translation missing

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,285 @@
...
+                </p>
+            {% endif %}
+            </div>
+        </div>
+        <div class="card-footer d-flex justify-content-between">
```
janno42: all following strings are not marked to be translated

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,285 @@
...
+
+            <div>
+                <div data-action-set="reviewing">
+                    <button role="button" data-action="publish" class="btn btn-outline-secondary">yes</button>
+                    <button role="button" data-action="make_private" class="btn btn-outline-secondary" disabled="">private</button>
+                    <button role="button" data-action="hide" class="btn btn-outline-secondary">no</button>
+                </div>
+                <div data-action-set="incomplete" class="d-none">
+                    <button role="button" data-action="repeat-unreviewed" class="btn btn-primary">Review unreviewed</button>
+                    <button role="button" data-action="repeat-start" class="btn btn-outline-primary">Review seen</button>
```
janno42: i'm not sure what this button is supposed to do. a button labeled `Show all again` and doing that would be fitting i guess.

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,285 @@
...
+            </div>
+
+            <div>
+                <div data-action-set="reviewing">
+                    <button role="button" data-action="publish" class="btn btn-outline-secondary">yes</button>
+                    <button role="button" data-action="make_private" class="btn btn-outline-secondary" disabled="">private</button>
+                    <button role="button" data-action="hide" class="btn btn-outline-secondary">no</button>
+                </div>
+                <div data-action-set="incomplete" class="d-none">
+                    <button role="button" data-action="repeat-unreviewed" class="btn btn-primary">Review unreviewed</button>
```
janno42: i would prefer `Show unreviewed again`

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,285 @@
...
+                    <button role="button" data-action="repeat-start" class="btn btn-outline-primary">Review seen</button>
+                </div>
+                <div data-action-set="done" class="d-none">
+                {% if next_course %}
+                    <a href="{% url 'staff:course_comments' next_course.semester.id next_course.id %}" class="btn btn-primary">Next review</a>
+                {% endif %}
+                    <button role="button" data-action="repeat-all" class="btn btn-outline-primary">Review all</button>
```
janno42: `Show all again`

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,285 @@
...
+                    <button role="button" data-action="make_private" class="btn btn-outline-secondary" disabled="">private</button>
+                    <button role="button" data-action="hide" class="btn btn-outline-secondary">no</button>
+                </div>
+                <div data-action-set="incomplete" class="d-none">
+                    <button role="button" data-action="repeat-unreviewed" class="btn btn-primary">Review unreviewed</button>
+                    <button role="button" data-action="repeat-start" class="btn btn-outline-primary">Review seen</button>
+                </div>
+                <div data-action-set="done" class="d-none">
+                {% if next_course %}
+                    <a href="{% url 'staff:course_comments' next_course.semester.id next_course.id %}" class="btn btn-primary">Next review</a>
```
janno42: `Review next course`

janno42:
this was just a first glance at the pr and i didn't fully review the code yet.

karyon:
oh did i already tell you that? yeah maybe i already told you that...

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,446 @@
...
+                {% endfor %}
+            {% endfor %}
+            </div>
+            <div data-layer="answer">
+            {% for questionnaire, contributor, label, is_responsible, results in sections %}
+                {% for result in results %}
+                    {% for answer in result.answers %}
+                        <div class="card-body slider-item border rounded-bottom" data-id="{{ answer.id }}"{% if contributor %} data-contribution="yes"{% endif %}{% if answer.is_reviewed %} data-review="{% if answer.is_published %}publish{% elif answer.is_private %}make_private{% else %}hide{% endif %}"{% endif %} data-reference="{% if contributor %}{{ contributor.id }}-{% endif %}{{ result.question.id }}">
+                        {% if answer.reviewed_answer %}
+                            <p>{{ answer.reviewed_answer|linebreaksbr }}</p>
```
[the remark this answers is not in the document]
pixunil: Changed by using `flex-direction: column`

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,446 @@
...
+        <div class="modal-dialog" role="document">
+            <div class="modal-content">
+                <div class="modal-header">
+                    <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
+                    <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                        <span aria-hidden="true">&times;</span>
+                    </button>
+                </div>
+                <table class="modal-body table table-striped">
+                    <tr><td><span class="key">←</span> <span class="key">→</span></td><td class="align-middle">{% trans 'Navigate through answers' %}</td></tr>
```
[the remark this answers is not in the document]
pixunil: Heaven :heart_eyes_cat:

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,446 @@
...
+                    {% endfor %}
+                {% endfor %}
+            {% endfor %}
+                <div class="alert alert-warning slider-item text-center">
+                    <span>
+                        {% trans 'Some comments are still unreviewed.' %}
+                    </span>
+                {% if next_course %}
+                    <span>
+                        The course {{ next_course.name }} has got {{ next_course.num_unreviewed_textanswers }} unreviewed text answers.
```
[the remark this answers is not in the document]
pixunil: grafik
Doing something wrong?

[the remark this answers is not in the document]
janno42: if using variables you have to use blocktrans (https://docs.djangoproject.com/en/2.0/topics/i18n/translation/#std:templatetag-blocktrans)

karyon:
hey, if that much code is necessary for a good UX, then i can live with that, but please try to keep it simple. we don't have infrastructure to test all of this and somebody needs to maintain it in five years.

@pixunil, @janno42 i can look into this whenever you tell me to, but please don't wait too long. if i request large-scale changes when you invested so much time already is kind of bad for everyone.

janno42:
i think we're feature complete and can now work on smaller details and the code. so @karyon go ahead with the review. i didn't look into the js code so far and up to now everything was feature and ux-related - i don't expect large scale change requests about that from your side.

now to the details i noticed:
the left and right border seem to randomly change sizes:



in general they should be a bit larger and please also increase the distance between pills and arrows. also, could you make the upper pill dark-blue? we're using that color for progress representation and it also can't be misinterpreted as "number of answers marked to be published" then.

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,436 @@
...
+                        this.active.addClass("active");
+                    }
+                }
+            };
+
+            function QuickSlider() {
+                this.setup.apply(this, arguments);
+            }
+
+            QuickSlider.prototype = {
```
karyon: why do we need this class? i know we need the contents, but why are they wrapped in a class?

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,436 @@
...
+                    this.active = element;
+
+                    if(this.active.length) {
+                        this.active.removeClass("to-left to-right");
+                        this.active.addClass("to-" + direction);
+                        this.active.position();
+                        this.active.addClass("active");
+                    }
+                }
+            };
```
karyon: is it possible to remove this whole class and put code into the other classes `slide` method that makes the three elements slide if necessary? would that be more straightforward?

evap/staff/views.py
```diff
@@ -764,9 +764,21 @@ def course_comments(request, semester_id, course_id):
...
-    template_data = dict(semester=semester, course=course, course_sections=course_sections,
-            contributor_sections=contributor_sections, filter_comments=filter_comments)
-    return render(request, "staff_course_comments.html", template_data)
+    template_data = dict(semester=semester, course=course, view=view)
+
+    if view == 'quick':
+        next_course = semester.course_set.annotate(num_unreviewed_textanswers=Count("contributions__textanswer_set", filter=Q(contributions__textanswer_set__state=TextAnswer.NOT_REVIEWED), distinct=True)) \
+            .exclude(pk=course.pk) \
+            .exclude(state='published') \
+            .filter(num_unreviewed_textanswers__gt=0) \
```
karyon: i would put the annotate behind the excludes, because semantically you don't want to annotate the whole `course_set`.

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,436 @@
...
+                    this.slider.find(".slider-inner > :last-child").toggleClass("disabled", !this.isReviewing);
+
+                    this.slider.find("[data-counter=reviewed-left]").text(this.numReviewed);
+                    this.slider.find("[data-counter=unreviewed-left]").text(this.index - this.numReviewed);
+                    this.slider.find("[data-counter=reviewed-right]").text(this.numAlreadyReviewed);
+                    let unreviewed = this.items.length - this.index - this.numAlreadyReviewed - 1;
+                    this.slider.find("[data-counter=unreviewed-right]").text(Math.max(unreviewed, 0));
+
+                    if(!this.isReviewing) {
+                        let reviewed = this.numReviewed === this.index;
```
karyon: i think it's a bit weird to use the index here, is there some kind of numTotal somewhere that you can use?

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,436 @@
...
+                    }
+
+                    this.items = reviewed.concat(unreviewed);
+
+                    let direction = action === "start"? "right" : "left";
+                    this.answers.slide(this.active, direction);
+                    this.updateState();
+                },
+
+                updateState: function(noFooter) {
```
karyon: updateState seems to be a catch-all "update all the things", aren't there separate parts (e.g. the arrows and the buttons) that don't always need to be updated at the same time?

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,436 @@
...
+                    }
+
+                    return this.items[this.index];
+                },
+
+                get numReviewed() {
+                    return $.grep(this.items.slice(0, this.index), item => item.is("[data-review]")).length;
+                },
+
+                get numAlreadyReviewed() {
```
karyon: i was completely confused by the names of these two functions (there are some variables somewhere else as well). i would suggest `numReviewedBefore` and `numReviewedAfter`, but I'm not really happy with that myself. maybe there's something better...

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,436 @@
...
+                        this.alert.toggleClass("alert-warning", !reviewed);
+                        this.alert.toggleClass("alert-success", reviewed);
+                        this.alert.find("[data-content=unreviewed]").toggleClass("d-none", reviewed);
+                        this.alert.find("[data-content=reviewed]").toggleClass("d-none", !reviewed);
+                        this.slider.find("[data-action=repeat-unreviewed]").toggleClass("d-none", reviewed);
+                    }
+
+                    if(noFooter) {
+                        return;
+                    }
```
karyon: why is this early return (and the noFooter parameter) necessary?

karyon:
this is definitely the sickest UI something that evap does not have yet.

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,436 @@
...
+                    {% trans 'Show unreviewed' %}
+                </button>
+            </div>
+        </div>
+    </div>
+{% endblock %}
+
+{% block additional_javascript %}
+    <script type="text/javascript">
+        (function($) {
```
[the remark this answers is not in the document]
pixunil: It is just a pattern I was seeing all along, it does two things:

1. Not polluting the `window` scope by using the scope of an anonymous function
2. Better portability if `$ != jQuery` _not our case_ (but a side effect with no harm)

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,436 @@
...
+                </div>
+                {% for result in results %}
+                    <div data-layer="question" data-id="{% if contributor %}{{ contributor.id }}-{% endif %}{{ result.question.id }}" data-reference="{% if contributor %}{{ contributor.id }}-{% endif %}{{ questionnaire.id }}">
+                        {{ result.question.text}}
+                    </div>
+                    {% for answer in result.answers %}
+                        <div data-layer="answer" data-id="{{ answer.id }}"{% if contributor %} data-contribution="yes"{% endif %}{% if answer.is_reviewed %} data-review="{% if answer.is_published %}publish{% elif answer.is_private %}make_private{% else %}hide{% endif %}"{% endif %} data-reference="{% if contributor %}{{ contributor.id }}-{% endif %}{{ result.question.id }}">
+                        {% if answer.reviewed_answer %}
+                            <p>{{ answer.reviewed_answer|linebreaksbr }}</p>
+                            <p class="comment-original">({{ answer.original_answer|linebreaksbr }})</p>
```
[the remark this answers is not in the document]
janno42: there is no way editing the answer, but if it was edited in another view it should be displayed here.

evap/staff/templates/staff_course_comments_quick.html
pixunil: Agreed, I'm still collecting ideas to refactor this code and this is definitely on my list

evap/staff/templates/staff_course_comments_quick.html
pixunil: It _could_ be a plain object, but it felt strange for me. Yeah, JavaScript is strange.

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
karyon: I would remove it. @janno42?

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,436 @@
...
+                <div data-layer="questionnaire" data-id="{% if contributor %}{{ contributor.id }}-{% endif %}{{ questionnaire.id }}">
+                    {{ questionnaire.public_name }}{% if contributor %}: {{ contributor.full_name }}{% if is_responsible %} ({% trans 'responsible' %}){% endif %}{% if label %} &ndash; <i>{{ label }}</i>{% endif %}{% endif %}
+                </div>
+                {% for result in results %}
+                    <div data-layer="question" data-id="{% if contributor %}{{ contributor.id }}-{% endif %}{{ result.question.id }}" data-reference="{% if contributor %}{{ contributor.id }}-{% endif %}{{ questionnaire.id }}">
+                        {{ result.question.text}}
+                    </div>
+                    {% for answer in result.answers %}
+                        <div data-layer="answer" data-id="{{ answer.id }}"{% if contributor %} data-contribution="yes"{% endif %}{% if answer.is_reviewed %} data-review="{% if answer.is_published %}publish{% elif answer.is_private %}make_private{% else %}hide{% endif %}"{% endif %} data-reference="{% if contributor %}{{ contributor.id }}-{% endif %}{{ result.question.id }}">
```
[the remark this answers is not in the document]
pixunil: Are they even allowed there? If yes, then I would have done definetely

evap/staff/templates/staff_course_comments_quick.html
karyon: no, not even a plain object, why not have just a bunch of functions? they all don't seem to have any state anyways.

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
karyon: i don't know to be honest, but please try :)

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
janno42: i would also remove it

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,436 @@
...
+                        // l
+                        this.press("hide");
+                        event.preventDefault();
+                    } else if(event.which === 8) {
+                        // backspace
+                        this.press("unreview");
+                        event.preventDefault();
+                    } else if(event.which === 13) {
+                        // enter
+                        this.slider.find("[data-action=next-course]")[0].click();
```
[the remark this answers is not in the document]
pixunil: Good point, but there is one exception to it: a click on the button should toggle it, but the hotkey should always do this action (there is backspace to remove the selection). But yeah, it's probably more future-proof in general

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
pixunil: Gosh, forget the links to do that :facepalm:

evap/staff/templates/staff_course_comments_quick.html
karyon: okay, that was not obvious to me. then why is `updateState` called without `noFooter=True` in `press`? 

also, i assume this is the purpose of the button timeout. would it be more appropriate to start that timeout inside updateState if noFooter was set?

karyon: ...i just noticed that `press` calls `updateState` and then `slide`, which also calls `updateState`.

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
karyon: yeah, true. i'm positive you'll find a way ;)

evap/staff/templates/staff_course_comments_quick.html
pixunil: Yes, that's correct, as the button needs to be coloured before sliding to the next answer. I'll just collect all this to find a semantically better version

pixunil:
So, I striped the shared state into four variables (where `active` and `index` are functional dependent, but I guess it is for the sake of readability) and bundled the functions a bit more, so they do more in less code. I added two comments which could help understanding why am I doing that, as I make use of an invariant in the markup that the referenced layer atop a lower layer is always before it. For CSS, I used nearly everywhere flexbox, but now it looks like a huge spaghessics due to flexboxes stacked in three layers. Could be that I shrunk the whole slider, if it is too small, just post a specific `rem` value :smirk_cat:

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,403 @@
...
+                active.removeAttr("data-review");
+                updateButtons();
+                return;
+            }
+
+            active.attr("data-review", action);
+            updateButtons();
+
+            slide("right");
+            slider.find("[data-action=" + action + "]").focus();
```
karyon: i would prefer something like
```
if(action === "unreview") {
    active.removeAttr("data-review");
} else {
    active.attr("data-review", action);
}
updateButtons();

if(action !== "unreview") {
      slide("right");
      slider.find("[data-action=" + action + "]").focus();
}
```
your call.

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,403 @@
...
+                let alert = slider.find(".alert");
+                alert.toggleClass("alert-warning", !reviewed);
+                alert.toggleClass("alert-success", reviewed && items.length !== 0);
+                alert.toggleClass("alert-secondary", reviewed && items.length === 0);
+                alert.find("[data-content=unreviewed]").toggleClass("d-none", reviewed);
+                alert.find("[data-content=reviewed]").toggleClass("d-none", !reviewed);
+                slider.find("[data-action=repeat-unreviewed]").toggleClass("d-none", reviewed);
+                slider.find("[data-action=repeat-all]").toggleClass("d-none", items.length === 0);
+            }
+        }
```
karyon: do you actually *want* to update all of this on every call to `handle`? please check whether there's a reasonable way to split this up and call the parts only when needed.

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,403 @@
...
+                    let btn = slider.find("[data-action=" + action + "]");
+                    btn.toggleClass(actions[action], review === action);
+                    btn.toggleClass("btn-outline-secondary", review !== action);
+                }
+
+                let notContributor = !active.data("contribution");
+                let privateBtn = slider.find("[data-action=make_private]");
+                privateBtn.attr("disabled", notContributor);
+                privateBtn.tooltip(notContributor? 'enable' : 'disable');
+            }
```
karyon: same for `updateButtons`, please check whether there's a reasonable way to split this up.

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,403 @@
...
+                    37: "slide-left",        // arrow left
+                    39: "slide-right",       // arrow right
+                    74: "publish",           // j
+                    75: "make_private",      // k
+                    76: "hide",              // l
+                     8: "unreview",          // backspace
+                    13: "next-course",       // enter
+                    77: "repeat-unreviewed", // m
+                    78: "repeat-all"         // n
+                };
```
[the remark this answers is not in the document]
pixunil: The problem with delegating hotkeys to their buttons can be lots of irritation of `:focus` and other stuff, at least it was like that when I begun with web dev. I know that a stringly typed interface is not great, but I can't think how I can avoid them on the buttons (currently `data-action` is an interface with strings, so it could be replaced with `data-action` for review, `data-slide` and `data-repeat` but it would cause more binding code imo. At least there has to be some unique identifier in the template)

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,403 @@
...
+                element.addClass("active");
+            }
+        }
+
+        function findActive(layer) {
+            if(layer === 2) {
+                return slider.find("[data-layer=2].active, .alert.active");
+            } else {
+                return slider.find("[data-layer=" + layer + "].active");
+            }
```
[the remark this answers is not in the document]
pixunil: Oh my… You found my greatest struggle point. Semantically, an `.slider-item[data-layer]` is something with real content, so `0` is questionnaires, `1` is questions and `2` is answers. So what are these alerts? They are like a substitute to all three layers. So they need to be slid away before a layer 2 is slid in (otherwise they will be pushed down), but they are no `[data-layer=2]` (read: no answer). `findActive()` is some poor naming but I'm still waiting for a nice idea

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,403 @@
...
+            let newIndex;
+
+            if(action !== "all" && unreviewed.length) {
+                newIndex = reviewed.length;
+            } else {
+                newIndex = 0;
+            }
+
+            items = reviewed.concat(unreviewed);
+            let direction = newIndex > index? "right" : "left";
```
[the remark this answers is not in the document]
pixunil: Official code style or begin of ideology wars? :smiling_imp:

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,403 @@
...
+            index = newIndex;
+            active = items[index];
+            slideLayer(2, direction, active);
```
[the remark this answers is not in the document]
pixunil: Isn't this even harder to read? When I apply `s/active/items[index]/` then this line would grow into
```
slideLayer(2, newIndex > index ? "right" : "left", items[index]);
```
I'm fine with it, but I thought not all readers will read it with ease

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,403 @@
...
+                }
+            });
+
+            let newIndex;
+
+            if(action !== "all" && unreviewed.length) {
+                newIndex = reviewed.length;
+            } else {
+                newIndex = 0;
+            }
```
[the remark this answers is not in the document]
pixunil: You should be greeted with the reviewed answers in their natural order and not with the green alert. If you followed a link to the comments, you did it on purpose (like, filtering in the semester view after reviewed courses) and want to see them. Other opinions?

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,373 @@
...
+            if(index === items.length) {
+                let reviewed = items.every(item => item.is("[data-review]"));
+                let alert = slider.find(".alert");
+                alert.toggleClass("alert-warning", !reviewed);
+                alert.toggleClass("alert-success", reviewed && items.length !== 0);
+                alert.toggleClass("alert-secondary", reviewed && items.length === 0);
+                alert.find("[data-content=unreviewed]").toggleClass("d-none", reviewed);
+                alert.find("[data-content=reviewed]").toggleClass("d-none", !reviewed);
+                slider.find("[data-action=repeat-unreviewed]").toggleClass("d-none", reviewed);
+                slider.find("[data-action=repeat-all]").toggleClass("d-none", items.length === 0);
```
pixunil: I came up with the idea to have delegated classes `.reviewed` and  `.unreviewed` which would hide or show the content in css.

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,403 @@
...
+            } else if(parts[0] === "repeat") {
+                repeat(parts[1]);
+                updateButtons();
+            } else if(action === "next-course" && !click) {
+                slider.find("[data-action=next-course]")[0].click();
+            } else if(["publish", "make_private", "hide", "unreview"].indexOf(action) !== -1) {
+                press(action, click);
+            }
+
+            updateState();
```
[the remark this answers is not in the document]
pixunil: `updateState` may be gone by now, but `updateButton` will not be always called after `slide`, for the review buttons giving their feedback. In the last option, you saw the `noFooter` parameter for it, so this is the replacement for it

evap/staff/tools.py
```diff
@@ -188,3 +189,10 @@ def merge_users(main_user, other_user, preview=False):
     other_user.delete()
 
     return merged_user, errors, warnings
+
+def find_next_unreviewed_course(semester, excluded):
+    return semester.course_set.exclude(pk__in=excluded) \
+        .exclude(state='published') \
+        .annotate(num_unreviewed_textanswers=Count("contributions__textanswer_set", filter=Q(contributions__textanswer_set__state=TextAnswer.NOT_REVIEWED), distinct=True)) \
+        .filter(num_unreviewed_textanswers__gt=0) \
```
karyon: did you try doing `.filter(contributions__textanswer_set__state=TextAnswer.NOT_REVIEWED)`? (not sure whether that works)

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
karyon: we don't have an official code style (yet), but if we had one it would surprise me if leaving out the space before the `?` were part of it. in fact, it would surprise me if that were part of any style guide anywhere. i have never seen that before, have you?

(my opinion: it's the operator with the lowest precedence right after assignment operators and the comma operator. we use spaces around all the other operators (except for the unary ones like ++ or !, which have very high precedence), and i don't see why of all operators the low-precedence ternary operator should not get spaces around it. in addition, in the case of `newIndex > index? foo : bar` leaving out the space is actually misleading, because it makes it look like `newIndex > (index? foo : bar)`. the only way i would find this remotely acceptable is by removing all the other spaces as well, or adding parenthesis, both of which seem unnecessary compared to adding the space.)

i would like to take the opportunity and thank my parents for raising me into such a privileged position that instead of worrying about housing and food, i have time to discuss this kind of stuff. thanks, mom and dad!

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -224,14 +224,29 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
...
+        function slide(to) {
+            let direction = to;
+            if(to === "left" && index > 0) {
                 index--;
-            } else if(direction === "right" && index < items.length) {
+            } else if(to === "right" && index < items.length) {
                 index++;
+            } else if(Number.isInteger(to)) {
+                direction = to >= index? "right" : "left";
+                index = to;
```
karyon: cool, you found a more general case of the slide method, which is to give it an arbitrary index instead of "left" or "right". this makes the "left" and "right" parameters look a bit like unnecessary convenience, how about we use the general case everywhere? so, instead of calling `slide("right")`, one would call `slide(index+1)`?

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
karyon: of course the buttons would need an ID which is a string, yes, but there is no problem with that or is there?

i think this solution could be much nicer if it works. can you try it please?

evap/evaluation/models.py
```diff
@@ -775,6 +775,10 @@ def is_private(self):
     def is_published(self):
         return self.state == self.PUBLISHED
 
+    @property
+    def is_reviewed(self):
+        return self.state != self.NOT_REVIEWED
```
karyon: test

evap/staff/tools.py
karyon: test

karyon:
also check out the outdated comments on that last review.

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -291,25 +291,9 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
...
-                if(item.is("[data-review]")) {
-                    reviewed.push(item);
-                } else {
-                    unreviewed.push(item);
-                }
-            });
-
+            let answers = slider.find("[data-layer=2]");
+            let reviewed = answers.filter("[data-review]").toArray().map($);
+            let unreviewed = answers.filter(":not([data-review])").toArray().map($);
```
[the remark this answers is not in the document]
pixunil: `toArray()` will put a jQuery collection (an array-like object) into a plain array, but as DOM objects. A map on `$` will convert them back into jQuery objects. Nothing else has been done in the loop, where the first line was `let item = $(this)`

evap/staff/templates/staff_course_comments_quick.html
pixunil: Okay, if they will underflow `index < 0` or overflow `index > items.length` then I can just reset them to bounds. Nice thing which I didn't had planned, but if you prefer an index driven interface rather than a string interface, it's fine on my side

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
karyon: but why do you need it at all? to me your answer reads as "...converts an array(-like object) with jquery objects to an array with jquery objects".

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
pixunil: Well, while I'm thinking about it... I think it is my personal style developed when I didn't understand the operator yet, so I tried to interpret the question mark as the end of a question and the colon as else. Like `(Est-ce que) [question]? [option] : [option]`, so I guess that your argumentation makes more sense.

Also thanks a lot for your stamina, while I'm still not convinced of all your suggestions I try them and look what will happen and I can really see the improvements.

evap/staff/tools.py
pixunil: Yes, still on my try list. I thought that the feature of not cycling through two courses was more important than this experiment. At least, if I will say no, I want to give you a good explanation.

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
pixunil: Well, did my research and hacking, and I have no nice solution yet:

- Based on the lookup table for key codes to action, I can use an identifier for the button. And then `.click()` it. Problem: jQuery will fire an own jQuery event which has two sides: I can detect in the handler that the action wasn't triggered by the mouse, but from code, but the link to next course will be ignored (hence I used `.find("[data-action=next-course]")[0]` to cast it in a DOM object). It would need an extra case for the link
- I can use a single bind handler for all `[data-action]` objects which will look a lot like `handle`(kind `[data-action=#{action}]`) now or split it in three handlers for review ,  or a tuple for type and subaction (kind`[data-#{type}=#{action}]`)
- The `unreview` action has no matching button, it is the toggle state of the review buttons

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
pixunil: Aha, you think that his is so easy? The array-like object can be indexed with `.get(i)` not `[i]`, which is not a great harm, but the problem lies in the combination of `unreviewed` and `reviewed` sets. Yes, sets is a good keyword: the ordering will reset to markup ordering. I could move the elements in the DOM, but this would result in a more complex markup (every item must know the name of their parent and the parent must know their name and the django template loop must be run three times otherwise the ordering of the layers would be wrong and then then we need CSS flex `order` to bring back peace into the anarchy), as the parent (or reference) is quite simply found in the markup by going backwards. Yes, this is one piece which I glued together from jQuery docs, an outdated (?) answer from SO and modern JS and I don't think it is ideal, but it guarantees an array with the ordering we want which makes less headache elsewhere.

evap/staff/tools.py
pixunil: With my small knowledge about Django and my failed attempt to implement this, I would say that it is not possible.

For reference: In the [docs](https://docs.djangoproject.com/en/2.0/topics/db/aggregation/#filtering-on-annotations) they show how to count all books rated higher than seven. So somehow it has to be done that the annotation filters out everything. I tried as alternative something like `Queryset("contributions__textanswer_set").filter(state=TextAnswer.NOT_REVIEWED).count()` but it does not connect correctly to the outer query.

karyon: i tried this here and it seemed to just work:
```
# current solution
>>> Semester.objects.first().course_set\
    .annotate(num_unreviewed_textanswers=Count("contributions__textanswer_set", filter=Q(contributions__textanswer_set__state=TextAnswer.NOT_REVIEWED), distinct=True))\
    .filter(num_unreviewed_textanswers__gt=0).count()
43

# proposed solution
>>> Semester.objects.first().course_set\
    .filter(contributions__textanswer_set__state=TextAnswer.NOT_REVIEWED).distinct().count()
43

```

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
karyon: would https://api.jquery.com/jQuery.merge/ be of any help?

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
karyon: why do you need to detect whether the event was fired by the mouse or not? an extra case for the link alone is not a show stopper i'd say.

the unreview hotkey would use a selector that does not select a specific id but rather the currently selected button. or does that not work?

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
karyon: thanks for developing such a large feature with us and for persevering in the face of seemingly endless reviews ;)

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,373 @@
...
+                return slider.find("[data-layer=2].active, .alert.active");
+            } else {
+                return slider.find("[data-layer=" + layer + "].active");
+            }
+        }
+
+        function repeat(action) {
+            let reviewed = slider.find("[data-layer=2][data-review]");
+            let unreviewed = slider.find("[data-layer=2]:not([data-review])");
+            let index = action === "unreviewed" && unreviewed.length? reviewed.length : 0;
```
pixunil: Yes, the `index` must be assigned before `merge()`, as it will modify the first argument and return the modified array. I played around with other options but this looked like the cleanest to me.

Just noted after push that `index` can be renamed to `to` in this scope.

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
pixunil: I need to difference between mouse and keyboard as the review buttons have toggle behaviour with mouse but the hotkeys have do-only-one-thing behaviour. (Sorry if I'm repeating myself there) With the same reason, I can't delegate the `unreview` action to an active button as a) the code has no chance to decide whether the key was activated by keyboard with a review or a unreview or b) we had a second extra case for the unreview.

Speaking of mental models, you want to have the hotkeys imitate the mouse. This works for sliding, repeating and forwarding, but with reviewing we have a model with slightly different behaviour. I think this is the cause why I have problems translating this into code. I can hack together that follows this model, but I have problems with reading the flow.

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -140,24 +140,38 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
...
 
                 if(actions[event.which]) {
-                    handle(actions[event.which]);
+                    let element = slider.find(actions[event.which]);
+                    // review hotkeys have no toggle behaviour
+                    if(actions[event.which].startsWith("[data-action")) {
+                        element.click();
+                    } else if(element.length) {
+                        element[0].click();
+                    }
```
pixunil: Please ignore how the condition is expressed, I will find something better than a bracket which will remain open to you the rest of the day. In fact, `publish`, `make_private` and `hide` all take the upper branch and `unreview` with all other actions take the lower.

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -203,25 +214,10 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
                 findActive(2).css({transform: "", opacity: ""});
             }
 
-            handle("repeat-unreviewed");
+            repeat("unreviewed");
+            updateButtons();
```
pixunil: To `repeat`, the call to `updateButtons` can be integrated, whereas in `slide` not

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -140,24 +140,38 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
             });
 
             slider.on("click", "[data-action]", event => {
-                handle($(event.target).data("action"), true);
+                press($(event.target).data("action"), !!event.originalEvent);
```
pixunil: Fancy way for a boolean cast. Urge to remove it...

evap/staff/tools.py
pixunil: Still, I am not sure where you want to apply this snippet. Your code will count the number of courses in a semester which have unreviewed text answers (That's for the filter toolbar in the semester overview).
The next course is the course with unreviewed text answers which has been ended most in the past. Do you want to propose a change how the next course is displayed or is this a replacement for the innermost query?

---------------------

code in git gives:
```sql
SELECT ..., COUNT(DISTINCT "evaluation_textanswer"."id")
FILTER (WHERE "evaluation_textanswer"."state" = NR) AS "num_unreviewed_textanswers"
FROM "evaluation_course"
LEFT OUTER JOIN "evaluation_contribution" ON ("evaluation_course"."id" = "evaluation_contribution"."course_id")
LEFT OUTER JOIN "evaluation_textanswer" ON ("evaluation_contribution"."id" = "evaluation_textanswer"."contribution_id")
WHERE ("evaluation_course"."semester_id" = 21 AND NOT ("evaluation_course"."state" = published))
GROUP BY "evaluation_course"."id"
ORDER BY "evaluation_course"."vote_end_date" ASC, "num_unreviewed_textanswers" DESC
```
your proposal gives:
```sql
SELECT ..., COUNT(DISTINCT "evaluation_textanswer"."id")
FILTER (WHERE "evaluation_textanswer"."state" = NR) AS "num_unreviewed_textanswers"
FROM "evaluation_course"
LEFT OUTER JOIN "evaluation_contribution" ON ("evaluation_course"."id" = "evaluation_contribution"."course_id")
LEFT OUTER JOIN "evaluation_textanswer" ON ("evaluation_contribution"."id" = "evaluation_textanswer"."contribution_id") WHERE "evaluation_course"."semester_id" = 21
GROUP BY "evaluation_course"."id"
HAVING COUNT(DISTINCT "evaluation_textanswer"."id")
FILTER (WHERE ("evaluation_textanswer"."state" = NR)) > 0
ORDER BY "evaluation_course"."name_de" ASC
```

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
pixunil: > Even infinities tend to age.

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
karyon: maybe you can do a similar thing as for unreview for the three buttons? such that their selector selects the corresponding id *and* on the activated attribute if there is any such thing. this way, the selector would find nothing if the button is pressed already.

evap/staff/tools.py
karyon: i proposed to replace the long annotate plus filter by a simple filter, sorry if that wasn't clear. the exclude and order need to stay of course.

evap/staff/tools.py
```diff
@@ -194,6 +194,6 @@ def merge_users(main_user, other_user, preview=False):
 def find_next_unreviewed_course(semester, excluded):
     return semester.course_set.exclude(pk__in=excluded) \
         .exclude(state='published') \
-        .annotate(num_unreviewed_textanswers=Count("contributions__textanswer_set", filter=Q(contributions__textanswer_set__state=TextAnswer.NOT_REVIEWED), distinct=True)) \
-        .filter(num_unreviewed_textanswers__gt=0) \
+        .filter(contributions__textanswer_set__state=TextAnswer.NOT_REVIEWED) \
+        .annotate(num_unreviewed_textanswers=Count("contributions__textanswer_set")) \
```
pixunil: I have no idea whether the `distinct` is necessary here, as I don't know what it does exactly

karyon: in my small example it was necessary, otherwise i got duplicated results. the reason i cannot explain exactly to you, i think it's because of the join that produces one line (and thus one object in the resulting querySet) per unreviewed textanswer, and not per course.

hm i see that you still need the annotate for the ordering, i thought that can be removed as well. well then the simpler filter is not that much of a simplification but oh well.

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,368 @@
...
+            to = Math.min(Math.max(to, 0), items.length);
+            let direction = to >= index? "right" : "left";
+            index = to;
+
+            if(index === items.length) {
+                let reviewed = items.filter(":not([data-review])").length === 0;
+                let alert = slider.find(".alert");
+                alert.toggleClass("alert-warning", !reviewed);
+                alert.toggleClass("alert-success", reviewed && items.length !== 0);
+                alert.toggleClass("alert-secondary", reviewed && items.length === 0);
```
karyon: to me it looks like these three classes correspond to the three spans in the alert. is there a way to put these classes directly into/around these spans?

evap/staff/templates/staff_course_comments_quick.html
pixunil: Yes, I think that `.startsWith("[data=action")` is not really clear about what it is doing. I just want to hear a "yes, this approach is more sound" or "please revert to the old one, that looks not so awesome i imagined" (Yes, still not a fan of your mental model) After that, I will remove some other minor flaws in the code

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,368 @@
...
+
+            function resetTouchSliding() {
+                touchX = null;
+                findActive(2).css({transform: "", opacity: ""});
+            }
+
+            repeat("unreviewed");
+        });
+
+        function slide(to) {
```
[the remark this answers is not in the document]
pixunil: `slideTo()` is also an interface for carousel, so I would use that as a convention pivot element

evap/staff/templates/staff_course_comments_quick.html
pixunil: No, they change the color of the alert. As I pointed out in another comment some lines below this (which is still relevant), custom classes could be used to toggle the spans (or misuse the `.alert-*` classes to have some visibility logic for their children). Also the `.alert-secondary` case is a bit of convenience to have not the template decide whether to apply it. It felt strange in both locations, so it ended up here.

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -0,0 +1,368 @@
...
+                {% endif %}
+                {% if next_course %}
+                    <span>
+                    {% blocktrans with name=next_course.name count answers=next_course.num_unreviewed_textanswers %}
+                        The next course {{ name }} has got one unreviewed text answer.
+                    {% plural %}
+                        The next course {{ name }} has got {{ answers }} unreviewed text answers.
+                    {% endblocktrans %}
+                    </span>
+                {% endif %}
```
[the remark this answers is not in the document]
pixunil: I can't deduce what the normal formatting for django templates is. Sometimes, control tags are indented after a html tag and sometimes they have they are on the same level. I preferred the later, as there is much nesting here

evap/staff/templates/staff_course_comments_quick.html
karyon: i'm not sure why the [0] is needed, but if it is necessary, can't you put a `:first` or something similar into the selector and remove the branching here?

evap/staff/templates/staff_course_comments_quick.html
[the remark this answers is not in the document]
karyon: yeah, it might very well be not consistent in other templates, but the rule for new code (and when changing existing code) is to indent everything.

evap/staff/templates/staff_course_comments_quick.html
pixunil: The `[0]` would be more equivalent to `.eq(0)`, wrapping the jQuery object into a DOM object, to fake a real click

karyon: so, can you put a :first into the selector and remove the branching?

also, in the previous review i proposed to include in the selector that the button must be active for it to be selected, so that the hotkey for a pressed button actually does nothing and does not pass on a toggle variable which is later used to determine that nothing should be done. did you try that?

evap/static/css/evap.scss
```diff
@@ -44,6 +45,26 @@ div.container {
...
+    }
+    > :nth-child(2) {
+        flex: 5;
+        text-align: center;
+    }
+    > :last-child {
+        flex: 1;
+        text-align: right;
+    }
+}
```
karyon: what does this do?

evap/staff/views.py
```diff
@@ -815,9 +816,24 @@ def course_comments(request, semester_id, course_id):
...
             section_list.append(CommentSection(questionnaire, contribution.contributor, contribution.label, contribution.responsible, text_results))
 
-    template_data = dict(semester=semester, course=course, course_sections=course_sections,
-            contributor_sections=contributor_sections, filter_comments=filter_comments)
-    return render(request, "staff_course_comments.html", template_data)
+    template_data = dict(semester=semester, course=course, view=view)
+
+    if view == 'quick':
+        visited = request.session.get('review-visited', set())
+        visited = set(visited)
```
karyon: this shouldn't be necessary afaics

evap/staff/views.py
```diff
@@ -815,9 +816,24 @@ def course_comments(request, semester_id, course_id):
...
+        visited.add(course.pk)
+        next_course = find_next_unreviewed_course(semester, visited)
+        if not next_course and visited:
```
karyon: visited is always true-ish

evap/static/css/evap.scss
janno42: it makes an element with three children placing the first with left-aligned content, the last with right-aligned content and the middle one (the second) with center-aligned content - this being five times as large as the outer ones. it's used for the quick review's card footer (and i think also for the page footer, but i didn't review the last changes yet).

pixunil: page footer and navigation bar was identified as further applications of this style class, was also tested on my local machine, but should be in a separate PR.
Also, I am open for better names. My alternatives: `three-splitted-row` `left-center-right` `centered-row`

evap/staff/views.py
pixunil: Great eye! This remained as a in-dev-hack while migrating from lists to sets

pixunil:
> please add a test that calls the view with view=quick (no need to actually test anything) for coverage.

Huh? The default test is going to quick, but full is not covered. Anyway, where should these tests live?

karyon:
probably right beside the current tests for that view.

evap/staff/templates/quick-review.js
```diff
@@ -0,0 +1,184 @@
...
+            return;
+        }
+
+        let active = items.eq(index);
+        let parameters = {"id": active.data("id"), "action": action, "course_id": {{ course.id }}};
+        $.ajax({
+            type: "POST",
+            url: "{% url 'staff:course_comments_update_publish' %}",
+            data: parameters,
+            error: function(){ window.alert("{% trans 'The server is not responding.' %}"); }
```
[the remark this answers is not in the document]
pixunil: The ajax error would be asynchronous to the function. If i is important, I could only slide in the success case, which could have a slight delay

evap/static/css/_quick-review.scss
```diff
@@ -0,0 +1,94 @@
...
+        .fas {
+            cursor: pointer;
+        }
+    }
+    .fas {
+        margin: 0.2rem 0;
+        font-size: 2rem;
+        color: grey;
+        transition: opacity;
+        transition-duration: 0.2s;
```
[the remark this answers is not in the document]
pixunil: I am inconsistent, but I used
```
transition: transform opacity;
transition-duration: 0.2s;
```
in another place to only have one duration setting and not two for each animated property. Which method do you prefer for single properties?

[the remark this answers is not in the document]
janno42: for multiple properties i would use the one you mentioned here and for a single one the short one-line version.

evap/staff/templates/quick-review.js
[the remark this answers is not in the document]
janno42: i think we should keep it as is. if an error occurs there is something wrong anyway and the user can invest some time thinking about it :)

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -101,7 +101,7 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
                         {% trans 'yes' %}
                     </button>
                     <button role="button" data-action="make_private" class="btn btn-sm btn-outline-secondary" disabled=""
-                        title="{% trans 'This text answer is about the course in general and can not refer to a single contributor.' %}" data-toggle="tooltip">
+                        title="{% trans "This answer is for a course question and can't be made private." %}" data-toggle="tooltip">
```
janno42: please use blocktrans to avoid using `"`

evap/staff/templates/staff_course_comments_quick.html
```diff
@@ -93,36 +93,34 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
...
-                    </button>
-                </div>
+        <div class="card-footer d-flex">
+            <div class="lcr-left"></div>
+            <div class="lcr-center" data-action-set="reviewing">
+                <button role="button" data-action="publish" class="btn btn-sm btn-outline-secondary">
+                    {% trans 'yes' %}
+                </button>
+                <button role="button" data-action="make_private" class="btn btn-sm btn-outline-secondary" disabled=""
+                    title="{% trans "This answer is for a course question and can't be made private." %}" data-toggle="tooltip">
```
janno42: please use blocktrans to avoid using `"`

PR #1157, opened 2018-04-23
Solve #1151 -- proportional reward

### LINKED ISSUE #1151: Proportional rewards
Students should get a reward even if they missed an evaluation to still motivate them to finish the other evaluations.

- Once students have evaluated all courses they participated in, they receive `REWARD_POINTS_FULLY_EVALUATED` (default: 3) reward points (already implemented, setting is currently called `REWARD_POINTS_PER_SEMESTER`).
- If students didn't evaluate all their courses, they can still receive reward points once for all courses they participated in either the evaluation period ended or they already submitted their evaluation for the course.
    - They receive `REWARD_POINTS_TWO_THIRDS` (default: 2) reward points if they evaluated at least 66% of their courses.
    - They receive `REWARD_POINTS_ONE_THIRD` (default: 1) reward points if they evaluated at least 33% of their courses.

felixrindt:
fixes #1151

### DIFF
--- a/evap/rewards/models.py
+++ b/evap/rewards/models.py
@@ -20,12 +20,6 @@ class RedemptionEventExpired(Exception):
     pass
 
 
-"""
-All of the following objects handle reward points. As reward points might be connected to monetary transactions,
-these objects may not be altered or deleted after creation.
-"""
-
-
 class RewardPointRedemptionEvent(models.Model):
     name = models.CharField(max_length=1024, verbose_name=_("event name"))
     date = models.DateField(verbose_name=_("event date"))
@@ -46,6 +40,12 @@ class RewardPointRedemptionEvent(models.Model):
             redemptions_dict[redemption.user_profile] += redemption.value
         return redemptions_dict
 
+
+
+"""
+The two following objects handle reward point amounts. As reward points might be connected to monetary transactions,
+these objects may not be altered or deleted after creation.
+"""
 
 class RewardPointGranting(models.Model):
     user_profile = models.ForeignKey('evaluation.UserProfile', models.CASCADE, related_name="reward_point_grantings")
--- a/evap/rewards/models.py
+++ b/evap/rewards/models.py
@@ -20,6 +20,12 @@ class RedemptionEventExpired(Exception):
     pass
 
 
+"""
+All of the following objects handle reward points. As reward points might be connected to monetary transactions,
+these objects may not be altered or deleted after creation.
+"""
+
+
 class RewardPointRedemptionEvent(models.Model):
     name = models.CharField(max_length=1024, verbose_name=_("event name"))
     date = models.DateField(verbose_name=_("event date"))
@@ -60,3 +66,4 @@ class RewardPointRedemption(models.Model):
 class SemesterActivation(models.Model):
     semester = models.OneToOneField('evaluation.Semester', models.CASCADE, related_name='rewards_active')
     is_active = models.BooleanField(default=False)
+
--- a/evap/rewards/tests/test_tools.py
+++ b/evap/rewards/tests/test_tools.py
@@ -1,5 +1,5 @@
 from django.conf import settings
-from django.test import TestCase
+from django.test import TestCase, override_settings
 from django.urls import reverse
 
 from model_mommy import mommy
@@ -58,16 +58,15 @@ class TestGrantRewardPoints(WebTest):
         self.form.submit()
         self.assertEqual(target_points(1.0), reward_points_of_user(self.student))
 
+    @override_settings(REWARD_POINTS=[
+        (0.5, 2),
+        (1.0, 5),
+    ])
     def test_target_points(self):
-        thresholds = [
-            (0.5, 2),
-            (1.0, 5),
-        ]
-        self.assertEqual(target_points(0.0, thresholds=thresholds), 0)
-        self.assertEqual(target_points(0.6, thresholds=thresholds), 2)
-        self.assertEqual(target_points(0.9, thresholds=thresholds), 2)
-        self.assertEqual(target_points(1.0, thresholds=thresholds), 5)
-        self.assertEqual(target_points(1.0, thresholds=[]), 0)
+        self.assertEqual(target_points(0.0), 0)
+        self.assertEqual(target_points(0.6), 2)
+        self.assertEqual(target_points(0.9), 2)
+        self.assertEqual(target_points(1.0), 5)
 
 
 class TestGrantRewardPointsParticipationChange(TestCase):
--- a/evap/rewards/tests/test_tools.py
+++ b/evap/rewards/tests/test_tools.py
@@ -8,7 +8,7 @@ from evap.evaluation.models import Course, Questionnaire, Question
 from evap.evaluation.models import UserProfile
 from evap.evaluation.tests.tools import WebTest
 from evap.rewards.models import SemesterActivation, RewardPointGranting
-from evap.rewards.tools import reward_points_of_user
+from evap.rewards.tools import reward_points_of_user, target_points
 
 
 class TestGrantRewardPoints(WebTest):
@@ -38,19 +38,31 @@ class TestGrantRewardPoints(WebTest):
     def test_everything_works(self):
         SemesterActivation.objects.create(semester=self.course.semester, is_active=True)
         self.form.submit()
-        self.assertEqual(reward_points_of_user(self.student), settings.REWARD_POINTS_PER_SEMESTER)
+        self.assertEqual(reward_points_of_user(self.student), target_points(1.0))
 
     def test_semester_activated_not_all_courses(self):
         SemesterActivation.objects.create(semester=self.course.semester, is_active=True)
         mommy.make(Course, semester=self.course.semester, participants=[self.student])
         self.form.submit()
-        self.assertEqual(0, reward_points_of_user(self.student))
+        self.assertEqual(target_points(0.5), reward_points_of_user(self.student))
 
     def test_already_got_points(self):
         SemesterActivation.objects.create(semester=self.course.semester, is_active=True)
         mommy.make(RewardPointGranting, user_profile=self.student, value=0, semester=self.course.semester)
         self.form.submit()
-        self.assertEqual(0, reward_points_of_user(self.student))
+        self.assertEqual(target_points(1.0), reward_points_of_user(self.student))
+
+    def test_target_points(self):
+        thresholds = [
+            (0.5, 2),
+            (1.0, 5),
+        ]
+        self.assertEqual(target_points(0.0, thresholds=thresholds), 0)
+        self.assertEqual(target_points(0.6, thresholds=thresholds), 2)
+        self.assertEqual(target_points(0.9, thresholds=thresholds), 2)
+        self.assertEqual(target_points(1.0, thresholds=thresholds), 5)
+        self.assertEqual(target_points(1.0, thresholds=[]), 0)
+
 
 class TestGrantRewardPointsParticipationChange(TestCase):
     @classmethod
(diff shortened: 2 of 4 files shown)


felixrindt:
A thing I noticed is that the `required_for_review` field of courses now behaves more like a `contribute_to_review` field, as we now can also provide reward if not all courses were evaluated. This historisch gewachsene naming issue might be confusing in the future.

felixrindt:
A thing I noticed is that the `required_for_reward` field of courses now behaves more like a `contribute_to_reward` field, as we now can also provide reward if not all courses were evaluated. This historisch gewachsene naming issue might be confusing in the future.

evap/rewards/tools.py
```diff
@@ -59,6 +60,21 @@ def reward_points_of_user(user):
...
+def target_points(progress, thresholds=None):
+    """
+    How many points should a user have based on their evaluation progress?
+    Progress should be 0.0 when none and 1.0 when all courses are evaluated.
+    thresholds is a list of tuples of form (<progress threshold>, <target reward value>).
+    If thresholds is None the REWARD_POINTS setting is used.
+    """
+    thresholds = thresholds if thresholds is not None else settings.REWARD_POINTS
+    # Filter reward point targets we have enough progress for
+    threshold_passed_reward_amount = map(lambda a: a[1], filter(lambda v: v[0] <= progress, thresholds))
```
He3lixxx: I think this
```
reached_thresholds = filter(lambda threshold: threshold[0] <= progress, thresholds)
threshold_passed_reward_amount = map(lamdba threshold: threshold[1], reached_thresholds)
```
drastically increases readability. Since I personally work with list comprehensions much more often than with map and filter, I'd even go for
```
passed_reward_amounts = [threshold[1] for threshold if threshold[0] <= progress]
```
This may even be short enough to directly wrap into the max:
```
reward_points = max([threshold[1] for threshold if threshold[0] <= progress], default=0)
```

@janno42, @karyon?

evap/rewards/tools.py
```diff
@@ -71,14 +87,20 @@ def grant_reward_points(user, semester):
...
-        return False
-    # did the user not already get reward points for this semester?
-    if RewardPointGranting.objects.filter(user_profile=user, semester=semester).exists():
+
+    voted_courses_count = required_courses.filter(voters=user).count()
+    # full evaluation progress from 0.0 to 1.0
+    progress = float(voted_courses_count) / float(required_courses.count())
+
+    # How many points have been granted to this user this semester?
+    granted_points = RewardPointGranting.objects.filter(user_profile=user, semester=semester).aggregate(Sum('value'))['value__sum'] or 0
```
He3lixxx: Why always create a new object and not just update the old one?

felixrindt: As rewards are tied to financial operation at events, these objects shall not be modified. Ask @janno42 for a more detailed explanation.

He3lixxx: Alrighty, that's a point

evap/rewards/tools.py
felixrindt: I prefer the second option. Has the right amount of "things going on" per line at my opinion.

felixrindt: Changed to unpacking in list comprehension in the newest commit.

evap/rewards/tools.py
```diff
@@ -71,14 +84,21 @@ def grant_reward_points(user, semester):
...
+    points_missing = target_points(progress) - granted_points
```
karyon: to be honest, i would inline the whole thing and simply do
`
target_points = max(points for threshold, points in settings.REWARD_POINTS if threshold <= progress, default=0)
`
target points are the maximum of all point values whose threshold was passed, or zero if none was passed. in my opinion that is clear enough, and adding a 10 line method for this might add more overhead when reading than it's helping.

evap/rewards/tools.py
```diff
@@ -71,14 +84,21 @@ def grant_reward_points(user, semester):
...
+
+    # is the user actually missing points?
+    if points_missing < 1:
         return False
-    # grant reward points
-    RewardPointGranting.objects.create(user_profile=user, semester=semester, value=settings.REWARD_POINTS_PER_SEMESTER)
+
+    # grant missing reward points
+    RewardPointGranting.objects.create(user_profile=user, semester=semester, value=points_missing)
     return True
```
karyon: also, i would switch that if around (and maybe even inline points_missing):
```
if target_points > granted_points:
    RewardPointGranting.objects.create(..., value=target_points-granted_points)
    return True
return False
```
for me that expresses exactly what needs to be done, no further comments needed. but if you want to keep the early `return False` to be in line with the rest, that's fine as well.

evap/rewards/tools.py
```diff
@@ -71,14 +84,21 @@ def grant_reward_points(user, semester):
     required_courses = Course.objects.filter(participants=user, semester=semester, is_required_for_reward=True)
     if not required_courses.exists():
         return False
-    # does the user not participate in any more required courses in this semester?
-    if required_courses.exclude(voters=user).exists():
-        return False
-    # did the user not already get reward points for this semester?
-    if RewardPointGranting.objects.filter(user_profile=user, semester=semester).exists():
+
+    voted_courses_count = required_courses.filter(voters=user).count()
```
karyon: [optional] i would inline that variable, it doesn't add much value for me.

evap/rewards/tools.py
karyon: ...of course this method cannot be tested then anymore, but i think testing the whole view is just fine.

evap/rewards/tools.py
felixrindt: sure

evap/rewards/tests/test_tools.py
```diff
@@ -38,19 +38,37 @@ def test_semester_not_activated(self):
...
+    def test_already_got_grant_objects(self):
         SemesterActivation.objects.create(semester=self.course.semester, is_active=True)
         mommy.make(RewardPointGranting, user_profile=self.student, value=0, semester=self.course.semester)
         self.form.submit()
-        self.assertEqual(0, reward_points_of_user(self.student))
+        self.assertEqual(target_points(1.0), reward_points_of_user(self.student))
+
+    def test_already_got_points(self):
+        SemesterActivation.objects.create(semester=self.course.semester, is_active=True)
+        mommy.make(RewardPointGranting, user_profile=self.student, value=target_points(1.0), semester=self.course.semester)
```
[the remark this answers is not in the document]
felixrindt: `already_got_grant_object` is the only test testing multiple grant objects per semester. This is a regression test, because this would not work before this PR.

`already_got_enough_points` is the only test testing that we dont grant points if the user has enough already.

I think both are important, but can't be combined.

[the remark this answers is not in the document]
karyon: okay, maybe call the first test `test_already_got_grant_object_but_points_missing` and the second `test_already_got_grant_object_and_enough_points` or something like that?
maybe add assertions that there is (or there is not, respectively) two RewardPointGranting objects for that user and semester.

felixrindt:
Ok. I tried to resolve your review comments. Let me know if there's more to improve!

evap/rewards/models.py
```diff
@@ -20,6 +20,12 @@ class RedemptionEventExpired(Exception):
     pass
 
 
+"""
+All of the following objects handle reward points. As reward points might be connected to monetary transactions,
+these objects may not be altered or deleted after creation.
+"""
+
+
```
karyon: can you put that above the granting model? that's the most relevant thing because there we're giving people money, and that's also the one with the weird append-only behavior. (also i think editing events is not that bad)

karyon:
@He3lixxx your move ;)

evap/rewards/tools.py
```diff
@@ -71,15 +73,16 @@ def grant_reward_points(user, semester):
...
+    granted_points = RewardPointGranting.objects.filter(user_profile=user, semester=semester).aggregate(Sum('value'))['value__sum'] or 0
+    progress = float(required_courses.filter(voters=user).count()) / float(required_courses.count())
+    target_points = max([points for threshold, points in settings.REWARD_POINTS if threshold <= progress], default=0)
+
+    if target_points > granted_points:
+        RewardPointGranting.objects.create(user_profile=user, semester=semester, value=target_points-granted_points)
+        return True
+    return False
 
 # Signal handlers
```
janno42: the message in `grant_reward_points_after_evaluate` needs to be changed, because `grant_reward_points` can return `True` before all courses have been evaluated.

janno42: unfortunately this was below the area where i could add comments - it's some lines down from here ;)

felixrindt: We could do all kinds of stuff like showing the amount, wether there are more courses to evaluate and so on. But I think keeping it simple is fine, as the user can just check out the rewards page when and if they're interested in the details.

janno42: "points" is misleading if it's just one ;)

janno42: but that's probably not too important. however, i would like the text to mention, that more points can be earned for motivational reasons.

felixrindt:
Lets do the renaming in another issue...

evap/rewards/tools.py
```diff
@@ -91,8 +97,14 @@ def grant_reward_points_after_evaluate(sender, **kwargs):
...
     semester = kwargs['semester']
 
-    if grant_reward_points(request.user, semester):
-        messages.success(request, _("You just have earned reward points for this semester because you evaluated all your courses. Thank you very much!"))
+    result = grant_reward_points(request.user, semester)
+    if result:
+        granted, remaining = result
+        reward = ngettext("{count} reward point", "{count} reward points", granted).format(count=granted)
+        if remaining:
+            messages.success(request, _("You just earned {reward} for this semester. More points can be earned by evaluating more courses.").format(reward=reward))
```
janno42: can you please also take into account whether the user can actually evaluate more courses? i.e. there are courses the user didn't evaluate and which have an evaluation period that ends in the future.
because the user might just have not participated in some already finished evaluations and would then have no chance to get more points.

felixrindt: Well right, that actually makes it much more difficult...

felixrindt:
So now we're showing the messages differently.

felixrindt:
I like your feedback. I'll wait for @janno42's comments before adressing the code style...

karyon:
oh wow \*facepalm\*. yeah you're right...

evap/rewards/tools.py
```diff
@@ -88,8 +97,18 @@ def grant_reward_points_after_evaluate(sender, **kwargs):
...
-        messages.success(request, _("You just have earned reward points for this semester because you evaluated all your courses. Thank you very much!"))
+    result = grant_reward_points(request.user, semester)
+    if result:
+        granted, progress = result
+        message = ngettext("You just earned {count} reward point for this semester.",
+                           "You just earned {count} reward points for this semester.", granted).format(count=granted)
+
+        if progress >= 1.0:
+            message += " " + _("Thank you very much for evaluating all your courses.")
+        elif Course.objects.filter(participants=request.user, semester=semester, is_required_for_reward=True, state="in_evaluation").exclude(voters=request.user).exists():
```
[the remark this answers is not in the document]
janno42: yes, please exclude these three states instead

evap/rewards/tools.py
```diff
@@ -88,8 +97,18 @@ def grant_reward_points_after_evaluate(sender, **kwargs):
...
+            message += " " + _("Please continue evaluating your courses.")
```
janno42: `"We're looking forward to receiving feedback for your other courses as well."` sounds nicer :)

karyon:
cool! thanks for staying with us!

(i pushed a typo fix to the last commit to not further delay this. i hope you don't mind :) )

felixrindt:
:tada:

PR #1159, opened 2018-04-24
Fixed redis server installation in provision_vagrant_vm.sh

He3lixxx:
In the current provision script, there is a -y ("assume yes") missing in the installation command for the redis server. Thus, apt aborts this installation and users have to manually install redis-server in the vm afterwards. This adds the -y parameter.


karyon:
thanks to you for spotting and thanks to our newest addition to the team, @LeaGe, for showing us all kinds of new error messages we haven't seen before! :D

Issue #1136, opened 2018-03-24
Simplify results tuples


karyon:
we are again and again confused by these [result sections](https://github.com/fsr-itse/EvaP/blob/e68c21c7e42f0623c67c91fa7d12c399f3938c99/evap/results/tools.py#L26). i want to attempt simplifying these by structuring and naming them along the structure and naming of course-contribution-questionnaire-question.

also investigate moving some models to the results app.

also investigate getting the data not from the cache, but directly from the database, and caching only the overall average and distribution.

also investigate removing total_count, average, and approval_count and instead calculate these on the fly (WIP commit somewhere)

PR #1164, opened 2018-04-26
[wip] Replace deviation with distribution bars, remove it from the exporter

karyon:
I'll resolve the remaining todos in the code.


evap/results/tools.py
```diff
@@ -170,43 +169,45 @@ def _calculate_results_impl(course):
...
 
         (avg_contribution_likert if contributor else avg_generic_likert).append(average_likert)
-        (dev_contribution_likert if contributor else dev_generic_likert).append(deviation_likert)
         (avg_contribution_grade if contributor else avg_generic_grade).append(average_grade)
-        (dev_contribution_grade if contributor else dev_generic_grade).append(deviation_grade)
 
     # the final total grade will be calculated by the following formula (GP = GRADE_PERCENTAGE, CP = CONTRIBUTION_PERCENTAGE):
     # final_likert = CP * likert_answers_about_persons + (1-CP) * likert_answers_about_courses
     # final_grade = CP * grade_answers_about_persons + (1-CP) * grade_answers_about_courses
     # final = GP * final_grade + (1-GP) * final_likert
```
karyon: do you think this comment helps? to me it it's actually harder to understand as the code below.

evap/contributor/views.py
```diff
@@ -29,7 +29,7 @@ def index(request):
 
     for course in all_courses:
         if course.state == 'published':
-            course.avg_grade, course.avg_deviation = calculate_average_grades_and_deviation(course)
+            course.avg_grade = calculate_average_grade(course)
```
karyon: you don't want to show the distribution here?

evap/results/tools.py
```diff
@@ -51,6 +61,37 @@ def mix(a, b, alpha):
...
+    return distribution
+
+
+def expand_to_100_percent(a, b):
+    if a == 0 and b == 0:
+        return 0, 0
+    if a == 0:
+        return 0, 1
+    if b == 0:
+        return 1, 0
```
felixrindt: The 2nd and 3rd if statements aren't needed. Is there a reason you'd want to keep it?

evap/results/tools.py
```diff
@@ -171,42 +211,91 @@ def _calculate_results_impl(course):
...
+        (contributors_distribution, contributions_percentage),
+        (course_grade_questions_distribution, course_grade_questions_percentage),
+        (course_non_grade_questions_distribution, course_non_grade_questions_percentage),
+    ]
+    average_distribution = {1: 0, 2: 0, 3: 0, 4: 0, 5: 0}
+    for distribution, percentage in distributions:
+        distribution_sum = sum(distribution.values())
+        if distribution_sum > 0:
+            for answer, count in distribution.items():
+                average_distribution[answer] += percentage * (count / distribution_sum)
```
karyon: right now it looks to me like the contributors_distribution contains percentages, and the course distributions contain counts. i think that should be changed, but that's kind of part of the other comments.

evap/results/tools.py
```diff
@@ -171,42 +211,91 @@ def _calculate_results_impl(course):
...
+    # calculate average distribution
+    distributions = [
+        (contributors_distribution, contributions_percentage),
+        (course_grade_questions_distribution, course_grade_questions_percentage),
+        (course_non_grade_questions_distribution, course_non_grade_questions_percentage),
+    ]
+    average_distribution = {1: 0, 2: 0, 3: 0, 4: 0, 5: 0}
+    for distribution, percentage in distributions:
+        distribution_sum = sum(distribution.values())
+        if distribution_sum > 0:
```
karyon: same here, i would leave that if out

evap/results/tools.py
```diff
@@ -40,6 +40,16 @@ def avg(iterable):
     return float(sum(items)) / len(items)
 
 
+def avg_distribution(distributions):
+    average_distribution = {1: 0, 2: 0, 3: 0, 4: 0, 5: 0}
+    for distribution in distributions:
+        distribution_sum = sum(distribution.values())
+        if distribution_sum > 0:
```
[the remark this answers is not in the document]
janno42: i think i had division by zero errors there, but will check again.

evap/results/tools.py
```diff
@@ -40,6 +40,16 @@ def avg(iterable):
...
+            for answer, count in distribution.items():
+                average_distribution[answer] += count / float(distribution_sum)
```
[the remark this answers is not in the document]
janno42: i'm fine with distinguishing between counts and distributions and making clear where the conversion happens

evap/results/tools.py
```diff
@@ -51,6 +61,37 @@ def mix(a, b, alpha):
...
+def mix_distributions(a, b, alpha):
+    sum_a = sum(a.values())
+    sum_b = sum(b.values())
+    distribution = {1: 0, 2: 0, 3: 0, 4: 0, 5: 0}
+    if sum_a == 0 and sum_b == 0:
+        return distribution
+    if sum_a == 0:
+        return b
+    if sum_b == 0:
+        return a
```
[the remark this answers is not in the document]
janno42: well, if we make sure the distribution adds up to 1 before returning it anywhere, the special cases are not important anymore.

evap/results/tests/test_tools.py
```diff
@@ -151,10 +157,17 @@ def test_average_deviation(self):
...
+        # contribution1: 0.4 * (1, 0, 1, 0, 0) + 0.6 * (0, 0, 4, 0, 4) = (0.4, 0, 2.8, 0, 2.4) = (0.0714285714, 0, 0.5, 0, 0.4285714286)
+        # contribution2: (0, 2, 0, 2, 0) = (0, 0.5, 0, 0.5, 0)
+        # contributions: (0.0714285714, 0.5, 0.5, 0.5, 0.4285714286) = (0.0357142857, 0.25, 0.25, 0.25, 0.2142857143)
+
+        # course_non_grade: (0, 0, 0, 0, 3) = (0, 0, 0, 0, 1)
 
-        total_likert_dev = settings.CONTRIBUTION_PERCENTAGE * 1 + (1 - settings.CONTRIBUTION_PERCENTAGE) * 0
-        total_grade_dev = 1
-        total_dev = settings.GRADE_PERCENTAGE * total_grade_dev + (1 - settings.GRADE_PERCENTAGE) * total_likert_dev
+        # total: 0.375 * (0.0357142857, 0.25, 0.25, 0.25, 0.2142857143) + 0.625 * (0, 0, 0, 0, 1) = (0.0133928571, 0.09375, 0.09375, 0.09375, 0.7053571429)
```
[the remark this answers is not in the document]
janno42: this is a comment to explain where the magic numbers in the test come from if something needs to be changed in the future, i can remove them - but that makes it really hard to follow.

evap/results/tools.py
```diff
@@ -171,42 +211,91 @@ def _calculate_results_impl(course):
...
-        average_likert = avg([result.average for result in results if result.question.is_likert_question])
-        deviation_likert = avg([result.deviation for result in results if result.question.is_likert_question])
-        average_grade = avg([result.average for result in results if result.question.is_grade_question])
-        deviation_grade = avg([result.deviation for result in results if result.question.is_grade_question])
+    if contributions_percentage == 0:
+        course_grade_questions_percentage, course_non_grade_questions_percentage = expand_to_100_percent(course_grade_questions_percentage, course_non_grade_questions_percentage)
+    if course_grade_questions_percentage == 0:
+        contributions_percentage, course_non_grade_questions_percentage = expand_to_100_percent(contributions_percentage, course_non_grade_questions_percentage)
+    if course_non_grade_questions_percentage == 0:
+        contributions_percentage, course_grade_questions_percentage = expand_to_100_percent(contributions_percentage, course_grade_questions_percentage)
```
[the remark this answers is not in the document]
janno42: inline what where? which other function? i can only guess what exactly you suggest ;)

evap/results/tests/test_tools.py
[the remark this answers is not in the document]
karyon: okay, i see it now. i think the comment is rather hard to follow. e.g. the equal signs are not correct, maybe use the counts_to_distribution function (or whatever it will be called) (in case you actually create that function)?

evap/results/tools.py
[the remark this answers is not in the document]
karyon: i proposed to inline `get_grade_percentages`.

evap/results/tests/test_tools.py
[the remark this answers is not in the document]
janno42: yes, this was based on the assumption, that looking at it from a final distribution perspective the actual numbers are irrelevant and a distribution is only defined by proportions, so all the values are equal. i can add the function there somehow.

evap/results/tools.py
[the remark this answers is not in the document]
janno42: although it calculates three values depending on the three bools or do you mean a more general "see if you can shorten all this to the bare minimum?" because i currently don't see one inline solution. but that might change once i'm actually on it.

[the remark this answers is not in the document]
janno42: nvm. i read "one line" instead of "inline".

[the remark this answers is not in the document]
karyon: by inlining i meant to just put the code of the function into the call site and remove the function call. the way i see it this part consists of seven rather trivial lines which are not really harder to parse as a two-line function call with three parameters that are calculated in there.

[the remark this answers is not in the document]
karyon: good timing :)

evap/results/templates/distribution_bar.html
```diff
@@ -0,0 +1,38 @@
...
+{% if average %}
+    {% if show_participant_count %}
+        <div class="grade-bg-result-bar-count text-center{% if participants_warning %} participants-warning{% endif %}">
+            <span class="fas fa-user small"></span> {{ result.total_count }}
+        </div>
+    {% endif %}
+    <div class="d-flex">
+        <div class="distribution-bar">
+            {% for answer, value in distribution.items %}
+                {% if answer != 0 and value != 0 %}
```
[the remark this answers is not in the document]
janno42: check

evap/results/tools.py
```diff
@@ -171,42 +211,91 @@ def _calculate_results_impl(course):
...
+        if contributor:
+            # collect the grade question average, non-grade question average and the distributions for each contributor
+            grade_questions_average, grade_questions_distribution, non_grade_questions_average, non_grade_questions_distribution = \
+                calculate_average_grade_and_distribution_for_contribution_results(grouped_results[contributor])
+            contributor_averages.append(mix(grade_questions_average, non_grade_questions_average, settings.CONTRIBUTOR_GRADE_QUESTIONS_PERCENTAGE))
+            contributor_distributions.append(mix_distributions(grade_questions_distribution, non_grade_questions_distribution, settings.CONTRIBUTOR_GRADE_QUESTIONS_PERCENTAGE))
+        else:
+            # calculate the grade question average, non-grade question average and the distributions for the course
+            course_grade_questions_average, course_grade_questions_distribution, course_non_grade_questions_average, course_non_grade_questions_distribution = \
+                calculate_average_grade_and_distribution_for_contribution_results(grouped_results[contributor])
```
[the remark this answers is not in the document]
janno42: check

evap/results/tools.py
```diff
@@ -171,42 +211,91 @@ def _calculate_results_impl(course):
...
-    final_likert_avg = mix(avg(avg_contribution_likert), avg(avg_generic_likert), settings.CONTRIBUTION_PERCENTAGE)
-    final_likert_dev = mix(avg(dev_contribution_likert), avg(dev_generic_likert), settings.CONTRIBUTION_PERCENTAGE)
-    final_grade_avg = mix(avg(avg_contribution_grade), avg(avg_generic_grade), settings.CONTRIBUTION_PERCENTAGE)
-    final_grade_dev = mix(avg(dev_contribution_grade), avg(dev_generic_grade), settings.CONTRIBUTION_PERCENTAGE)
+    non_grade_question_results = [result for result in results if result.question.is_non_grade_rating_question]
+    non_grade_questions_average = avg([result.average for result in non_grade_question_results])
+    non_grade_questions_distribution = {1: 0, 2: 0, 3: 0, 4: 0, 5: 0}
+    for result in non_grade_question_results:
+        for answer, count in result.distribution.items():
+            non_grade_questions_distribution[answer] += count
```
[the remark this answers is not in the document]
janno42: check

evap/results/tools.py
janno42: obsolete :)

evap/results/tools.py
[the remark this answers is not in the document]
janno42: check

evap/results/tools.py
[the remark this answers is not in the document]
janno42: removed

evap/results/tools.py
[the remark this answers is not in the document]
janno42: turns out we need the checks because a contributor might not have grade questions or non grade questions and then the other distribution is returned as is.

evap/results/tools.py
janno42: check

evap/results/tools.py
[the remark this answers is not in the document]
janno42: should be better now, please check. i can then change it in other places as well.

janno42:
i actually noticed an error in the results computation during refactoring where the number of questions per contributor influences the distribution. that's why the test was also changed to the now hopefully correct values.

evap/results/tools.py
```diff
@@ -171,42 +199,78 @@ def _calculate_results_impl(course):
...
+    for distribution in contributor_distributions:
+        for answer, value in distribution.items():
+            contributors_distribution[answer] += value
+    contributors_distribution = normalize_distribution(contributors_distribution)
+
+    # calculate to which percentage each group influences the total grade
+    contributions_percentage = settings.CONTRIBUTIONS_PERCENTAGE if contributors_average is not None else 0
+    course_grade_questions_percentage = settings.COURSE_GRADE_QUESTIONS_PERCENTAGE if course_grade_questions_average is not None else 0
+    course_non_grade_questions_percentage = 1 - settings.CONTRIBUTIONS_PERCENTAGE - settings.COURSE_GRADE_QUESTIONS_PERCENTAGE if course_non_grade_questions_average is not None else 0
+    if contributors_average or course_grade_questions_average or course_non_grade_questions_average:
```
karyon: how can all of them be zero?

janno42: if there are no answers yet for a course

evap/results/tools.py
```diff
@@ -171,42 +199,82 @@ def _calculate_results_impl(course):
...
+    # calculate average distribution
+    distributions = [
+        (contributors_distribution, contributions_percentage),
+        (course_grade_questions_distribution, course_grade_questions_percentage),
+        (course_non_grade_questions_distribution, course_non_grade_questions_percentage),
+    ]
+    average_distribution = {1: 0, 2: 0, 3: 0, 4: 0, 5: 0}
+    for distribution, percentage in distributions:
+        for answer, value in distribution.items():
+            average_distribution[answer] += percentage * value
```
karyon: ...all of this is computing a weighted average. how about we use a function called weighted_average like this:
```
def weighted_average_distribution(distributions, weights):
    sum_weights = sum(weights)
    scaled_weights = [weight / sum_weights for weight in weights]
    average_distribution = {1: 0, 2: 0, 3: 0, 4: 0, 5: 0}
    for distribution, weight in zip(distributions, scaled_weights):
        average_distribution = sum_distributions(average_distribution, multipy_distribution(distribution, weight))
    return average_distribution 
```
or don't scale the weights but normalize at the end, i'm not sure.

for consistency, i would also use this weighted average function instead of the mix (which would require calculating a second, kinda redundant, weight, but still), and also create such a function for floats instead of distributions.

the reason for this is that for once, the basic math stuff gets moved out of this function, and also that `mix` is kind of a special case of the weighted average of three values here. and here the averaging is performed manually while `mix` is a separate function, so why not create a function for the general case and use that everywhere...

evap/results/templates/distribution_bar.html
```diff
@@ -0,0 +1,11 @@
+{% load results_templatetags %}
+
+{% spaceless %}
+<div class="distribution-bar">
+    {% for answer, value in distribution.items %}
+        {% if value != 0 %}
+            <div class="vote-bg-{{ answer }}{% if participants_warning %} participants-warning{% endif %}" style="width: {% widthratio value distribution.values|sum 100 %}%;">&nbsp;</div>
```
karyon: that sum should now always be one, right?

karyon:
also, please replace the "change" in the last commitmessage with something more helpful :)

evap/results/tools.py
janno42: check

evap/results/tools.py
```diff
@@ -171,42 +199,82 @@ def _calculate_results_impl(course):
...
+    grouped_results = defaultdict(list)
+    for __, contributor, __, results, __ in calculate_results(course):
+        grouped_results[contributor].extend(results)
+
+    contributor_averages = []
+    contributor_distributions = []
+    course_grade_questions_average = None
+    course_grade_questions_distribution = {1: 0, 2: 0, 3: 0, 4: 0, 5: 0}
+    course_non_grade_questions_average = None
+    course_non_grade_questions_distribution = {1: 0, 2: 0, 3: 0, 4: 0, 5: 0}
```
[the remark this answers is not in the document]
janno42: removed them now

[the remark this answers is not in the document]
janno42: check

evap/results/tools.py
```diff
@@ -171,42 +199,82 @@ def _calculate_results_impl(course):
...
+            contributor_averages.append(mix(grade_questions_average, non_grade_questions_average, settings.CONTRIBUTOR_GRADE_QUESTIONS_PERCENTAGE))
+            grade_questions_distribution = normalize_distribution(grade_questions_counts)
+            non_grade_questions_distribution = normalize_distribution(non_grade_questions_counts)
+            contributor_distributions.append(mix_distributions(grade_questions_distribution, non_grade_questions_distribution, settings.CONTRIBUTOR_GRADE_QUESTIONS_PERCENTAGE))
+        else:
+            # calculate the grade question average, non-grade question average and the distributions for the course
+            course_grade_questions_average = grade_questions_average
+            course_grade_questions_distribution = normalize_distribution(grade_questions_counts)
+            course_non_grade_questions_average = non_grade_questions_average
+            course_non_grade_questions_distribution = normalize_distribution(non_grade_questions_counts)
```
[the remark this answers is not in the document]
janno42: worked out really nicely

evap/results/templates/distribution_bar.html
janno42: right

evap/results/tools.py
```diff
@@ -215,48 +215,42 @@ def calculate_average_grade_and_distribution(course):
...
     contributions_percentage = settings.CONTRIBUTIONS_PERCENTAGE if contributors_average is not None else 0
     course_grade_questions_percentage = settings.COURSE_GRADE_QUESTIONS_PERCENTAGE if course_grade_questions_average is not None else 0
     course_non_grade_questions_percentage = 1 - settings.CONTRIBUTIONS_PERCENTAGE - settings.COURSE_GRADE_QUESTIONS_PERCENTAGE if course_non_grade_questions_average is not None else 0
-    if contributors_average or course_grade_questions_average or course_non_grade_questions_average:
-        scale = 1 / sum([contributions_percentage, course_grade_questions_percentage, course_non_grade_questions_percentage])
-        contributions_percentage *= scale
-        course_grade_questions_percentage *= scale
-        course_non_grade_questions_percentage *= scale
+    if not contributors_average and not course_grade_questions_average and not course_non_grade_questions_average:
+        return None, None
```
karyon: this can be done earlier, maybe before the `# calculate the average grade` comment? also, [optional] please check whether something like `if not any(average for ..., average, .... in grouped_results.values()): return None, None` can be done right at the beginning of this function and if that looks better to you.

evap/results/tools.py
```diff
@@ -215,48 +215,42 @@ def calculate_average_grade_and_distribution(course):
...
         non_grade_questions_average = avg([result.average for result in non_grade_question_results])
-        non_grade_questions_counts = sum_up_result_counts(non_grade_question_results)
+        non_grade_questions_distribution = avg_distribution([result.distribution for result in non_grade_question_results])
         if contributor:
             # collect the grade question average, non-grade question average and the distributions for each contributor
             contributor_averages.append(mix(grade_questions_average, non_grade_questions_average, settings.CONTRIBUTOR_GRADE_QUESTIONS_PERCENTAGE))
-            grade_questions_distribution = normalize_distribution(grade_questions_counts)
-            non_grade_questions_distribution = normalize_distribution(non_grade_questions_counts)
+            grade_questions_distribution = grade_questions_distribution
+            non_grade_questions_distribution = non_grade_questions_distribution
```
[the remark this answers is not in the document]
janno42: :D

evap/results/tools.py
janno42: check

evap/results/tools.py
```diff
@@ -171,42 +170,102 @@ def _calculate_results_impl(course):
...
-    """Determines the final average grade and deviation for a course."""
-    avg_generic_likert = []
-    avg_contribution_likert = []
-    dev_generic_likert = []
-    dev_contribution_likert = []
-    avg_generic_grade = []
-    avg_contribution_grade = []
-    dev_generic_grade = []
-    dev_contribution_grade = []
+def normalize_distribution(distribution):
```
janno42: the methods `normalize_distribution`, `sum_distributions`and `multiply_distribution` are all only used once. do you want to keep them to have clean encapsulated math functions or do you want to inline them?

evap/results/tools.py
```diff
@@ -171,42 +170,109 @@ def _calculate_results_impl(course):
...
+    contributions_percentage *= scale
+    course_grade_questions_percentage *= scale
+    course_non_grade_questions_percentage *= scale
+
+    # calculate average grade
+    average_grade = contributions_percentage * (contributors_average or 0) + \
+                    course_grade_questions_percentage * (course_grade_questions_average or 0) + \
+                    course_non_grade_questions_percentage * (course_non_grade_questions_average or 0)
+
+    # calculate average distribution
```
karyon: this comment and the one above don't really add anything in my opinion since  the code quite clearly says `average distribution = some calculation`

evap/results/tools.py
```diff
@@ -171,42 +170,109 @@ def _calculate_results_impl(course):
...
+            course_non_grade_questions_average = non_grade_questions_average
+            course_non_grade_questions_distribution = non_grade_questions_distribution
+
+    if course.is_single_result:
+        course_grade_questions_average = None
+        course_non_grade_questions_average = None
+
+    # calculate the average grade and distribution over all contributors
+    contributors_average = avg(contributor_averages)
+    contributors_distribution = avg_distribution(contributor_distributions)
```
karyon: what about calling these `contributors_average_grade` and `contributors_average_distribution` and removing the comment?

evap/results/tools.py
```diff
@@ -171,42 +170,109 @@ def _calculate_results_impl(course):
...
+        non_grade_questions_distribution = avg_distribution([result.counts for result in non_grade_question_results])
+        if contributor:
+            # collect the grade question average, non-grade question average and the distributions for each contributor
+            contributor_averages.append(mix(grade_questions_average, non_grade_questions_average, settings.CONTRIBUTOR_GRADE_QUESTIONS_PERCENTAGE))
+            contributor_distributions.append(weighted_avg_distribution(
+                [grade_questions_distribution, non_grade_questions_distribution],
+                [settings.CONTRIBUTOR_GRADE_QUESTIONS_PERCENTAGE, 1 - settings.CONTRIBUTOR_GRADE_QUESTIONS_PERCENTAGE]
+            ))
+        else:
+            # calculate the grade question average, non-grade question average and the distributions for the course
```
karyon: this comment is not true anymore

evap/results/tools.py
```diff
@@ -171,42 +170,102 @@ def _calculate_results_impl(course):
...
+        grouped_results[contributor].extend(results)
+
+    # early return if no rating results exist for this course
+    if not any(result.average for result in grouped_results[contributor] for contributor in grouped_results.values() if result.question.is_rating_question):
+        return None, None
+
+    contributor_averages = []
+    contributor_distributions = []
+
+    for contributor in grouped_results:
```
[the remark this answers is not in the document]
janno42: check

evap/results/tools.py
```diff
@@ -171,42 +170,102 @@ def _calculate_results_impl(course):
...
-    final_avg = mix(final_grade_avg, final_likert_avg, settings.GRADE_PERCENTAGE)
-    final_dev = mix(final_grade_dev, final_likert_dev, settings.GRADE_PERCENTAGE)
+def calculate_average_grade_and_distribution(course):
+    # will contain a list of results for each contributor and one for the course (where contributor is None)
+    grouped_results = defaultdict(list)
+    for __, contributor, __, results, __ in calculate_results(course):
+        grouped_results[contributor].extend(results)
+
+    # early return if no rating results exist for this course
+    if not any(result.average for result in grouped_results[contributor] for contributor in grouped_results.values() if result.question.is_rating_question):
```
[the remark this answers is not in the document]
janno42: its lists of results grouped per contributor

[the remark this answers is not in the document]
janno42: ah, i know what you mean. funny that this did also work.

evap/results/tools.py
janno42: no they don't but the grouping would be off if some parts wouldn't have a comment. still remove them?

evap/results/tools.py
janno42: check

evap/results/tools.py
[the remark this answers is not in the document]
karyon: ....on a second look, i think the order of the loops is wrong. the current code iterates over `grouped_results[contributor]` first, and for each iteration, iterates over all elements in `grouped_results.values()`, which does not make sense.

evap/results/tools.py
karyon: it won't be off if we remove more comments ;) in general i would use text in comments primarily for commenting and not grouping.

evap/results/tools.py
```diff
@@ -171,42 +170,102 @@ def _calculate_results_impl(course):
...
+
+    scale = 1 / sum([contributions_percentage, course_grade_questions_percentage, course_non_grade_questions_percentage])
+    contributions_percentage *= scale
+    course_grade_questions_percentage *= scale
+    course_non_grade_questions_percentage *= scale
+
+    # calculate average grade
+    average_grade = contributions_percentage * (contributors_average or 0) + \
+                    course_grade_questions_percentage * (course_grade_questions_average or 0) + \
+                    course_non_grade_questions_percentage * (course_non_grade_questions_average or 0)
```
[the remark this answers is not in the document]
janno42: 1.+5. avg function with optional weights for floats and distributions it is.

2. i would say using `None`s in distributions makes things more difficult because we would have to replace them with `0`s if other values are valid before returning them.

3. percentage calculation is now removed.

4. initialization changed again.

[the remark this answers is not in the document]
karyon: 2. well, the same goes for the averages doesn't it? so whatever we do for the averages we could do for the distributions. or maybe i'm missing something.

evap/settings.py
```diff
@@ -38,18 +38,16 @@
...
-# the final total grade will be calculated by the following formula (GP = GRADE_PERCENTAGE, CP = CONTRIBUTION_PERCENTAGE):
-# final_likert = CP * likert_answers_about_persons + (1-CP) * likert_answers_about_courses
-# final_grade = CP * grade_answers_about_persons + (1-CP) * grade_answers_about_courses
-# final = GP * final_grade + (1-GP) * final_likert
-GRADE_PERCENTAGE = 0.8
-CONTRIBUTION_PERCENTAGE = 0.5
+# percentages for calculating a course's total average grade
+CONTRIBUTOR_GRADE_QUESTIONS_PERCENTAGE = 0.4  # grade questions make up this percentage of each contributor's average grade
+CONTRIBUTIONS_PERCENTAGE = 0.33  # the average contribution grade makes up this percentage of the course's average grade
+COURSE_GRADE_QUESTIONS_PERCENTAGE = 0.33  # the average grade of all grade questions about the course makes up this percentage of the course's average grade
```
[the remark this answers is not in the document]
janno42: 1 minus others might prevent some setting misconfigurations but we probably don't have to take these into account. i don't care about the naming that much. so just say what you prefer (i think i already know ;)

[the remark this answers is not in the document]
karyon: to clarify, this would also be a semantic change. you could specify 1, 1 and 10 as weights and it would just work. i would argue this even prevents misconfiguration compared to the current solution since it prevents the weight of the third group to become implicitly negative.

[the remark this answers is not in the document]
janno42: i thought you would say that.

evap/results/tools.py
```diff
@@ -171,42 +178,92 @@ def _calculate_results_impl(course):
...
+    course_non_grade_questions_percentage = 1 - settings.CONTRIBUTIONS_PERCENTAGE - settings.COURSE_GRADE_QUESTIONS_PERCENTAGE
+    average_grade = avg(
+        [contributors_average_grade, course_grade_questions_average, course_non_grade_questions_average],
+        [settings.CONTRIBUTIONS_PERCENTAGE, settings.COURSE_GRADE_QUESTIONS_PERCENTAGE, course_non_grade_questions_percentage]
+    )
+
+    average_distribution = avg_distribution(
+        [contributors_average_distribution, course_grade_questions_distribution, course_non_grade_questions_distribution],
+        [settings.CONTRIBUTIONS_PERCENTAGE, settings.COURSE_GRADE_QUESTIONS_PERCENTAGE, course_non_grade_questions_percentage]
+    )
```
karyon: now this breaks because avg_distribution is not checking whether there's actually anything inside the distribution.

evap/results/tools.py
```diff
@@ -171,42 +170,102 @@ def _calculate_results_impl(course):
...
+    if not any(result.average for result in grouped_results[contributor] for contributor in grouped_results.values() if result.question.is_rating_question):
+        return None, None
+
+    contributor_averages = []
+    contributor_distributions = []
+
+    for contributor in grouped_results:
+        grade_question_results = [result for result in grouped_results[contributor] if result.question.is_grade_question]
+        grade_questions_average = avg([result.average for result in grade_question_results])
+        grade_questions_distribution = avg_distribution([result.distribution for result in grade_question_results])
```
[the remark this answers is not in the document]
janno42: this was already changed to be counts

evap/results/tools.py
janno42: where does it break?

janno42: every parameter is a distribution (might be all zeros though)

evap/results/tools.py
```diff
@@ -25,19 +25,27 @@
...
+
+    average = 0
+    scale = 1 / sum([weight for value, weight in zip(values, weights) if value and weight])
+    for value, weight in zip(values, weights):
+        if value and weight:
+            average += scale * weight * value
+    return average
 
 
 def mix(a, b, alpha):
```
karyon: this one is unused now

evap/results/tools.py
karyon: yeah it does not break, sorry.

evap/results/tools.py
```diff
@@ -25,30 +25,30 @@
...
-        return b
-    if b is None:
-        return a
-
-    return alpha * a + (1 - alpha) * b
+    scale = 1 / weight_sum
+    for value, weight in zip(values, weights):
+        if value and weight:
+            average += scale * weight * value
+    return average
```
karyon: i tried to create a solution that sets the weights to one if the parameter is not present instead of a special case, iterates only once over the values, and crashes on weights that are None (because i think we don't need that do we?). here is what i came up with:
```
def avg(values, weights=itertools.repeat(1)):
    if all(x is None for x in values):
        return None
    
    sum_values = 0
    sum_weights = 0
    for value, weight in zip(values, weights):
        if value:
            sum_values += value * weight
            sum_weights += weight
    return sum_values / sum_weights
```

evap/results/tools.py
```diff
@@ -171,42 +170,101 @@ def _calculate_results_impl(course):
...
-    final_dev = mix(final_grade_dev, final_likert_dev, settings.GRADE_PERCENTAGE)
+def avg_distribution(distributions, weights=None):
+    if not weights:
+        return normalize_distribution(sum_distributions(distributions))
 
-    return final_avg, final_dev
+    weighted_distributions = []
+    for distribution, weight in zip(distributions, weights):
+        weighted_distributions.append(multiply_distribution(distribution, weight))
+    return normalize_distribution(sum_distributions(weighted_distributions))
```
karyon: i would make this here as analogous to the `avg` as possible. that would include skipping `None`s, and then you could remove the None handling from normalize, sum and multiply.

... actually, that would also include inlining sum and multiply, i think. although it's probably not really inlining, i think most of the code will just disappear. could you try that?

evap/results/tools.py
```diff
@@ -171,42 +170,101 @@ def _calculate_results_impl(course):
...
+
+
+def calculate_average_grade_and_distribution(course):
+    # will contain a list of results for each contributor and one for the course (where contributor is None)
+    grouped_results = defaultdict(list)
+    for __, contributor, __, results, __ in calculate_results(course):
+        grouped_results[contributor].extend(results)
+
+    # early return if no rating results exist for this course
+    if not any(result.average for result in grouped_results[contributor] for contributor in grouped_results.keys() if result.question.is_rating_question):
```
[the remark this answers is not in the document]
janno42: maybe now

evap/results/tools.py
```diff
@@ -171,42 +170,101 @@ def _calculate_results_impl(course):
...
+                [settings.CONTRIBUTOR_GRADE_QUESTIONS_WEIGHT, settings.CONTRIBUTOR_NON_GRADE_QUESTIONS_WEIGHT]
+            ))
+        else:
+            course_grade_questions_average = grade_questions_average
+            course_grade_questions_distribution = grade_questions_distribution
+            course_non_grade_questions_average = non_grade_questions_average
+            course_non_grade_questions_distribution = non_grade_questions_distribution
+
+    contributors_average_grade = avg(contributor_averages)
+    contributors_average_distribution = avg_distribution(contributor_distributions)
```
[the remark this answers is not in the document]
janno42: maybe `contributors_averages` for the other one, but i wouldn't call this `contributor_average_grade`  because its the average grade of all contributors

evap/results/tools.py
janno42: looks good on first sight. although i think `itertools.repeat(1)` doesn't make it easy to understand what we're doing here. i will just copy and paste it because i really can't keep up with all the changes anymore and have not really any idea what the results calculation is currently doing in general. i will review the final thing once you're happy, anything else currently feels like wasted time.

evap/results/tools.py
janno42: ah, so there's more stuff with avg. so then disregard the previous comment and i will right now review everything myself because i can't continue without getting an overview again :)

evap/results/tools.py
karyon: well, you can pass weights, and the default for that is a bunch of ones, i thought that was easy to understand...?

evap/results/tools.py
janno42: obsoleted. i'm happy that we don't need all of these anymore.

evap/results/tools.py
```diff
@@ -10,6 +10,8 @@
 from evap.evaluation.models import TextAnswer, Contribution, RatingAnswerCounter
 from evap.evaluation.tools import questionnaires_and_contributions
 
+import itertools
+
```
karyon: move that to the other stdlib import above please

evap/results/templates/distribution_bar.html
```diff
@@ -1,11 +1,38 @@
...
+            {% endif %}
+        </div>
+    </div>
+{% else %}
+    <div class="d-flex" data-toggle="tooltip" data-placement="left" title="{% trans 'Not enough answers were given.' %}">
+        {% if show_participant_count %}
+            <div class="grade-bg-result-bar-count text-center grade-bg-disabled">
+                <span class="fas fa-user small"></span> {{ result.total_count }}
+            </div>
         {% endif %}
```
karyon: same here. i think it's fair to not give the participant count the tooltip, since the participant count itself has no problem with the low number of answers.

evap/results/templates/distribution_bar.html
```diff
@@ -1,11 +1,38 @@
 {% load results_templatetags %}
+{% load evaluation_filters %}
 
 {% spaceless %}
-<div class="distribution-bar">
-    {% for answer, value in distribution.items %}
-        {% if value != 0 %}
-            <div class="vote-bg-{{ answer }}{% if participants_warning %} participants-warning{% endif %}" style="width: {% widthratio value distribution.values|sum 100 %}%;">&nbsp;</div>
+{% if average %}
```
karyon: this cannot happen if included from result_bar.html, right?

evap/results/views.py
```diff
@@ -124,14 +127,16 @@ def course_detail(request, semester_id, course_id):
...
+        course.avg_grade, course.distribution = calculate_average_grade_and_distribution(course)
+    else:
+        course.avg_grade, course.distribution = None, None
 
     template_data = dict(
             course=course,
             course_sections_top=course_sections_top,
             course_sections_bottom=course_sections_bottom,
             contributor_sections=contributor_sections,
-            show_grades=show_grades,
```
karyon: in this file, is that a functional change? if yes, could you explain what's the change?

evap/evaluation/tests/test_models.py
```diff
@@ -10,7 +10,8 @@
 
 from evap.evaluation.models import (Contribution, Course, CourseType, EmailTemplate, NotArchiveable, Questionnaire,
                                     RatingAnswerCounter, Semester, UserProfile)
-from evap.results.tools import calculate_average_grades_and_deviation
+from evap.results.tools import calculate_average_grade_and_distribution
+from evap.settings import EVALUATION_END_OFFSET_HOURS, EVALUATION_END_WARNING_PERIOD
```
karyon: that looks like a merge fail

evap/results/tests/test_tools.py
```diff
@@ -151,10 +157,17 @@ def test_average_deviation(self):
...
-        total_dev = settings.GRADE_PERCENTAGE * total_grade_dev + (1 - settings.GRADE_PERCENTAGE) * total_likert_dev
+        # total: 0.375 * (0.1, 0.25, 0.25, 0.25, 0.15) + 0.625 * (0, 0, 0, 0, 1) = (0.0375, 0.09375, 0.09375, 0.09375, 0.68125)
 
-        self.assertAlmostEqual(deviation, total_dev)
+        __, distribution = calculate_average_grade_and_distribution(course)
+        self.assertAlmostEqual(distribution[1], 0.0375)
+        self.assertAlmostEqual(distribution[2], 0.09375)
+        self.assertAlmostEqual(distribution[3], 0.09375)
+        self.assertAlmostEqual(distribution[4], 0.09375)
+        self.assertAlmostEqual(distribution[5], 0.68125)
```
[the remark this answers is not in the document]
janno42: okay, but i'll add an additional test, so that we also see whether not having all three components behaves as expected.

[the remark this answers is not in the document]
karyon: even better, thanks.

evap/results/tools.py
```diff
@@ -171,42 +165,83 @@ def _calculate_results_impl(course):
...
+        return None
 
-    for __, contributor, __, results, __ in calculate_results(course):
-        average_likert = avg([result.average for result in results if result.question.is_likert_question])
-        deviation_likert = avg([result.deviation for result in results if result.question.is_likert_question])
-        average_grade = avg([result.average for result in results if result.question.is_grade_question])
-        deviation_grade = avg([result.deviation for result in results if result.question.is_grade_question])
+    for answer, value in distribution.items():
+        distribution[answer] = value / float(distribution_sum)
+    return distribution
```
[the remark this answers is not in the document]
janno42: why would that be nicer? using the one liner now.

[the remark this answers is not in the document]
karyon: well, i don't find this way of creating dictionaries particularly intuitive, but your call.

evap/results/tests/test_tools.py
```diff
@@ -121,17 +121,23 @@ def test_average_grades(self):
...
-        self.assertAlmostEqual(deviation, 0)
+        contributors_percentage = settings.CONTRIBUTIONS_WEIGHT / (settings.CONTRIBUTIONS_WEIGHT + settings.COURSE_NON_GRADE_QUESTIONS_WEIGHT)  # 0.375
+        course_non_grade_percentage = settings.COURSE_NON_GRADE_QUESTIONS_WEIGHT / (settings.CONTRIBUTIONS_WEIGHT + settings.COURSE_NON_GRADE_QUESTIONS_WEIGHT)  # 0.625
 
-    @override_settings(CONTRIBUTION_PERCENTAGE=0.3, GRADE_PERCENTAGE=0.6)
-    def test_average_deviation(self):
+        total_grade = contributors_percentage * contributors_average + course_non_grade_percentage * course_non_grade_average  # 1.1625 + 3.125 = 4.2875
+
+        average_grade, __ = calculate_average_grade_and_distribution(course)
+        self.assertAlmostEqual(average_grade, total_grade)
```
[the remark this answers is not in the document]
janno42: check

evap/results/tools.py
```diff
@@ -144,7 +139,6 @@ def _calculate_results_impl(course):
 
                 total_count = len(answers)
                 average = avg(answers) if total_count > 0 else None
-                deviation = pstdev(answers, average) if total_count > 0 else None
                 counts = get_counts(question, answer_counters)
```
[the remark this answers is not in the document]
janno42: check

evap/results/tools.py
janno42: check. i'm really looking forward to deciding on an explicit order and using a linter for that :)

evap/results/templates/distribution_bar.html
janno42: what is the same as where?

evap/results/tools.py
janno42: i know! consistent import order is one of the few things i really liked better in 1327 so far.

evap/results/views.py
karyon: i saw that, the question was whether the templates now render something different since they don't get this boolean but instead None grades.

evap/results/templates/distribution_bar.html
janno42: i had some struggles when i was explicitly trying to get the tooltip around everything including the participant count so that it doesn't overlay the information. i agree that the participant count itself doesn't need the information but i don't want to hide it when the information is shown.

evap/results/templates/distribution_bar.html
janno42: result bar sets `average=result.average` when including. so `average` exists whenever there is data to display - otherwise the grey disabled bars are shown.

evap/results/templates/distribution_bar.html
karyon: what about any of the other three directions where the tooltip can be placed?

janno42: we don't display the tooltips to the right because we can't assume there is enough space. on the top and bottom other questions/courses would be overlapped which is also not really nice.
what's the problem with showing it left of the participation again?

evap/results/views.py
janno42: no, just the gray bar. before they also didn't show the grade and deviation.

evap/results/templates/distribution_bar.html
karyon: my primary problem is that large portions of this file are not used by all of it's callers, and there are too many code paths through this file, making this file hard to understand. moving some parts that are only used by one caller to that caller was a suggestion to help with that. if you have other solutions, i'm fine with that as well.

janno42: alright. the idea was to define all the elements only once and everywhere just using what is needed in the specific situation. i wanted to avoid defining similar elements that only differ in small details more than once. i'll see if i can rewrite the result_bar and distribution_bar in general.

evap/results/templates/distribution_bar.html
```diff
@@ -1,11 +1,38 @@
...
+    {% if show_participant_count %}
+        <div class="grade-bg-result-bar-count text-center{% if participants_warning %} participants-warning{% endif %}">
+            <span class="fas fa-user small"></span> {{ result.total_count }}
+        </div>
+    {% endif %}
```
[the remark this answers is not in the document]
janno42: will be checked together with the general bar rework

evap/results/tools.py
```diff
@@ -91,6 +90,9 @@ def get_counts(question, answer_counters):
 
     for answer_counter in answer_counters:
         counts[answer_counter.answer] = answer_counter.count
+
+    if sum(counts.values()) == 0:
+        return None
```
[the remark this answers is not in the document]
janno42: will do

evap/results/tools.py
```diff
@@ -168,13 +173,11 @@ def _calculate_results_impl(course):
 def normalize_distribution(distribution):
```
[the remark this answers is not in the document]
janno42: no. i thought you meant you want to have a different name if the parameter is changed. will change to `normalized_distribution`.

karyon:
the sum filter can probably be removed.

evap/results/templates/average_grade_badge.html
```diff
@@ -0,0 +1,16 @@
...
+        {% if result.question.is_yes_no_question %}
+            {{ result.approval_count|percentage:result.total_count }}
+        {% else %}
+            {{ average|floatformat:1 }}
+        {% endif %}
+    </div>
+{% else %}
+    <div class="grade-bg-result-bar text-center grade-bg-disabled"><span class="far fa-eye-slash"></span></div>
+{% endif %}
+{% endspaceless %}
```
karyon: perhaps consider making two versions of this. one for results index and the overview part of results course detail, because those don't have`participants_warning` and no `is_yes_no_question`, and the other one for the results bar, and in there you're then free to put the tooltip wherever necessary ;)

evap/results/tests/test_tools.py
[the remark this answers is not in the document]
karyon: ...the ultimate test whether we got this right would be to add up the distribution percentages multiplied with their index, and compare that with the average grade coming from the calculation function, wouldn't it? actually that would be kinda important to test, since we wanted the distribution to be accurate...

evap/contributor/templates/contributor_index.html
```diff
@@ -96,6 +96,10 @@
                                 {% else %}
                                     {% if course.avg_grade %}
                                         <div class="grade-bg text-center vertically-aligned" style="background-color: {{ course.avg_grade|gradecolor }}" data-toggle="tooltip" data-placement="top" title="{% trans 'Average grade' %}">{{ course.avg_grade|floatformat:1 }}</div>
+                                    {% else %}
+                                        <div class="grade-bg text-center vertically-aligned grade-bg-disabled" data-toggle="tooltip" data-placement="top" title="{% trans 'Not enough votes to publish results' %}">
+                                            <span class="far fa-eye-slash"></span>
+                                        </div>
```
karyon: this seems like a functional change, what happened here previously?

evap/results/tools.py
```diff
@@ -165,7 +165,7 @@ def _calculate_results_impl(course):
             elif question.is_heading_question:
                 results.append(HeadingResult(question=question))
 
-        section_warning = questionnaire_max_answers[(questionnaire, contribution)] < questionnaire_warning_thresholds[questionnaire]
+        section_warning = questionnaire_max_answers[(questionnaire, contribution)] > 0 and questionnaire_max_answers[(questionnaire, contribution)] < questionnaire_warning_thresholds[questionnaire]
```
karyon: why is this necessary?

if it stays, you can actually do `0 < questionnaire_max_answers[(questionnaire, contribution)] < questionnaire_warning_thresholds[questionnaire]`.

evap/results/views.py
```diff
@@ -86,8 +86,13 @@ def course_detail(request, semester_id, course_id):
                 answers = [answer for answer in result.answers if user_can_see_text_answer(request.user, represented_users, answer, public_view)]
                 if answers:
                     results.append(TextResult(question=result.question, answers=answers))
-            else:
+            elif show_grades or isinstance(result, HeadingResult):
                 results.append(result)
```
karyon: so why again do we show text answers but not rating answers if not show_grades?

...or rather why do we show a user the results page at all if they cannot see the results?

...can_user_see_results might need a rename because right now they can see results, even though only textanswers... `can_user_see_grades`?

evap/results/views.py
```diff
@@ -86,8 +86,13 @@ def course_detail(request, semester_id, course_id):
...
+            elif isinstance(result, RatingResult):
+                results.append(RatingResult(question=result.question, total_count=result.total_count, average=None, counts=None, warning=result.warning))
+            elif isinstance(result, YesNoResult):
+                results.append(YesNoResult(question=result.question, total_count=result.total_count, average=None, counts=None, warning=result.warning, approval_count=None))
```
karyon: previously, the template needed to remove the grades if the user was not allowed to see them, i'm missing a corresponding change in this commit...?

evap/results/tests/test_tools.py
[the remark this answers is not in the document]
janno42: check

evap/results/tools.py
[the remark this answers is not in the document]
janno42: sorry, merge fail.

evap/contributor/templates/contributor_index.html
janno42: if the course was published, the grade was shown even if the thresholds have not been met.

evap/results/tools.py
janno42: otherwise the warning "only a few ..." would be shown for sections where no answers exist at all. changed the condition.

evap/results/views.py
janno42: text answers are qualitative feedback, the rating answers are quantitative, so we decided not to have thresholds for publishing text answers.

the results page is shown to contributors because there might be text answers. we could also add additional checks and only show it if there really are text answers for the requesting user, but i wouldn't say that's necessary.

renaming to `can_user_see_grades`

evap/results/views.py
janno42: the previous commit changed the distribution bar and average grade badge rendering. they are not doing checks by themselves anymore and always display data if they receive some.

evap/contributor/templates/contributor_index.html
janno42: will be moved into the last commit

evap/results/templates/results_semester_detail.html
```diff
@@ -59,7 +59,7 @@ <h3>{{ semester.name }}</h3>
                                             {{ contributor.full_name }}{% if not forloop.last %}, {% endif %}
                                         {% endfor %}
                                     </td>
-                                    {% if course|can_user_see_results:request.user and course.avg_grade %}
+                                    {% if course|can_user_see_grades:request.user and course.avg_grade %}
```
karyon: the `can_user_see_grades` check (and thus the whole filter) shouldn't be necessary here anymore because the grade is set to None in the view, right?

evap/results/views.py
karyon: except for the first two hunks the changes in this file are missing their template changes, which seem to be in the next commit. ideally, they go into a separate commit along with their template changes, because the next commit is already a functional change and no refactoring.

evap/results/templates/result_bar.html
```diff
@@ -8,9 +8,13 @@
...
-        {% for answer, count in result.counts.items %}
-            {{ result.question|get_answer_name:answer }}: {{ count }}/{{ result.total_count }} ({{ count|percentage_one_decimal:result.total_count }})
-            {% if not forloop.last %}</br>{% endif %}
+        {% for count in result.counts %}
+            {% with answer_name=result.question|get_answer_name:forloop.counter %}
+                {% if answer_name %}
+                    {{ answer_name }}: {{ count }}/{{ result.total_count }} ({{ count|percentage_one_decimal:result.total_count }})
+                    {% if not forloop.last %}</br>{% endif %}
+                {% endif %}
+            {% endwith %}
```
karyon: bummer. i guess if we said that yesno questions used 1 and 2 instead of 1 and 5, that would have wider implications and complicate stuff in other places, right?

karyon: still, overall i think the tuples are a good thing.

evap/results/templates/results_semester_detail.html
janno42: yes, but i had to introduce it in the one commit to move that part around as requested - it is removed some commits later.

evap/results/templates/result_bar.html
janno42: i think it's not really important whether the options are 1 and 5 or anything else.

evap/results/templates/results_semester_detail.html
karyon: i don't quite understand, i meant that in this particular commit, you are setting the grades to None if `can_user_see_grades` in the view, but here you are still checking `can_user_see_grades`, which is redundant in my understanding.

evap/results/templates/result_bar.html
janno42: what i'm doing here is skipping all entries where no answer name exists, otherwise we would have 5 lines in the yes/no question tooltips, where the middle three are "None: 0/n (0%)" or something like that.

karyon: i know, and since that works around the zeroes in the counts tuple for yes/no questions, the question was whether it would make more stuff easier than it makes stuff harder if the counts tuple for yes/no questions contained only two elements instead of five so we don't need such workarounds.

evap/contributor/templates/contributor_index.html
```diff
@@ -96,6 +96,10 @@
                                 {% else %}
                                     {% if course.avg_grade %}
                                         <div class="grade-bg text-center vertically-aligned" style="background-color: {{ course.avg_grade|gradecolor }}" data-toggle="tooltip" data-placement="top" title="{% trans 'Average grade' %}">{{ course.avg_grade|floatformat:1 }}</div>
+                                    {% else %}
+                                        <div class="grade-bg text-center vertically-aligned grade-bg-disabled" data-toggle="tooltip" data-placement="top" title="{% trans 'Not enough votes to publish results' %}">
```
karyon: is it intentional that this view does not contain the distribution bar?

evap/results/templates/result_bar.html
```diff
@@ -1,34 +1,30 @@
...
+                    {% if answer_name %}
+                        {{ answer_name }}: {{ count }}/{{ result.total_count }} ({{ count|percentage_one_decimal:result.total_count }})
+                        {% if not forloop.last %}</br>{% endif %}
+                    {% endif %}
+                {% endwith %}
+            {% endfor %}"
+        {% endif %}
+    {% else %}
+        data-toggle="tooltip" data-placement="left" title="{% trans 'Not enough answers were given.' %}"
+    {% endif %}>
```
karyon: maybe the > on a new line so one has a chance to find where the div tag ends?

evap/staff/views.py
```diff
@@ -411,9 +411,9 @@ def semester_raw_export(request, semester_id):
         _('#Participants'), _('#Comments'), _('Average grade')])
     for course in semester.course_set.all():
         degrees = ", ".join([degree.name for degree in course.degrees.all()])
-        course.avg_grade, course.avg_deviation = calculate_average_grades_and_deviation(course)
-        if course.state in ['evaluated', 'reviewed', 'published'] and course.avg_grade is not None:
-            avg_grade = "{:.1f}".format(course.avg_grade)
+        distribution = calculate_average_distribution(course)
+        if course.state in ['evaluated', 'reviewed', 'published'] and distribution is not None:
```
He3lixxx: Sometimes we explicitly compare calculated values to None, sometimes we just say `if distribution` or `if course.avg_grade`. I think using `if a is not None` is the better way since it makes the intent more obvious (and also documents that None is a valid value here). I had a really hard time understanding why we use None as values for counts and distributions all over the results calculation and storing and think this might have helped. What do you guys think?

evap/results/views.py
```diff
@@ -74,7 +74,7 @@ def course_detail(request, semester_id, course_id):
     represented_users = list(request.user.represented_users.all())
     represented_users.append(request.user)
 
-    show_grades = request.user.is_reviewer or course.has_enough_voters_to_publish_grades
+    show_grades = course.can_user_see_grades(request.user)
 
     # filter text answers
```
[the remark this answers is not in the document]
karyon: changed the comment.

i agree regarding the readability, in fact I don't like both the result calculation that's assembling the mess of tuples, and this whole method here which kind of disassembles it again. i have this on my radar (#1136), but would not do it in this PR.

evap/evaluation/tests/test_auth.py
```diff
@@ -19,6 +19,7 @@ def setUpTestData(cls):
         course = mommy.make(Course, state='published')
         mommy.make(Contribution, course=course, contributor=cls.external_user, can_edit=True, responsible=True, comment_visibility=Contribution.ALL_COMMENTS)
         mommy.make(Contribution, course=course, contributor=cls.inactive_external_user, can_edit=True, responsible=False, comment_visibility=Contribution.ALL_COMMENTS)
+        course.general_contribution.questionnaires.set([mommy.make(Questionnaire)])
```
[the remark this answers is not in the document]
karyon: i'll remove the whole commit, the result calculation is robust enough by now to live without these questionnaires.

evap/staff/views.py
karyon: about understanding the Nones, it really just means "nothing to see here". we could also use tuples with all zeroes, but then we would still need to check whether the sum is zero e.g. to avoid divisions by zero when averaging, and before displaying. in case of the grade, using zero as "nothing to see here" is even weirder in my opinion.

besides being more explicit about the None, do you have any suggestions?

evap/results/tools.py
```diff
@@ -117,7 +99,7 @@ def _calculate_results_impl(course):
     """Calculates the result data for a single course. Returns a list of
     `ResultSection` tuples. Each of those tuples contains the questionnaire, the
     contributor (or None), a list of single result elements, the average grade and
-    deviation for that section (or None). The result elements are either
+    distribution for that section (or None). The result elements are either
     `RatingResult` or `TextResult` instances."""
```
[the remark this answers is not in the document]
karyon: yes.

karyon:
check out the last commit, it sets avg_grade and distribution to None for the templates, what do you say? i don't have a strong opinion about it.

karyon:
now using None instead of nonexistence.

evap/staff/views.py
He3lixxx: I agree that setting the value to None instead of to 0 is the better choice. I didnt't have any better idea.

He3lixxx:
I like it :)

evap/results/exporters.py
```diff
@@ -153,65 +129,55 @@ def export(self, response, course_types_list, include_not_enough_voters=False, i
...
                         approval_count = 0
 
                         for grade_result in qn_results:
                             if grade_result.question.id == question.id:
                                 if grade_result.average:
                                     values.append(grade_result.average * grade_result.total_count)
-                                    deviations.append(grade_result.deviation * grade_result.total_count)
                                     total_count += grade_result.total_count
                                 if grade_result.question.is_yes_no_question:
                                     approval_count += grade_result.approval_count
```
janno42: this throws `unsupported operand type(s) for +=: 'int' and 'NoneType'` when i tested exporting

evap/evaluation/models.py
```diff
@@ -348,6 +348,15 @@ def can_user_see_results(self, user):
             return self.can_user_see_course(user)
         return False
 
+    def can_user_see_grades(self, user):
+        if user.is_reviewer:
+            return True
+        if self.state == 'published':
```
janno42: maybe we should change the order and do a  `if self.state != 'published': return False`

evap/contributor/templates/contributor_index.html
janno42: it was intentional not to include the deviation when we added the grade. but that might have been because we were feeling the deviation doesn't help. we could add the distribution bar here, that would take up more space but would be consistent to the other views. what do you guys think?

karyon:
The "How are the grades calculated?" FAQ question needs to be updated.

evap/student/templates/student_index.html
```diff
@@ -81,34 +81,44 @@
...
+                                    {% if can_download_grades %}
+                                        <td class="text-center">
+                                            {% if course.grades_activated and course.grade_documents.count == 1 %}
+                                                <a class="btn btn-sm btn-sm-narrow btn-light" href="{% url 'grades:download_grades' course.grade_documents.all.first.id %}" role="button"
+                                                    data-toggle="tooltip" data-placement="top" title="{{ course.grade_documents.all.first.description }}">
+                                                    <span class="far fa-file"></span>
+                                                </a>
+                                            {% elif course.grades_activated and course.grade_documents.count > 1 %}
+                                                <div class="btn-group" role="group">
                                                     {% for grade_document in course.grade_documents.all %}
```
He3lixxx: Afaik, course.grade_documents.all is a function and should be called here, shouldn't it?
Same for the `count`s, `all`s and `first`s above. Also, some of the `.all().first()` could be shortened to just `first()`.

janno42: function calls are written without parentheses in django templates. you're right about the `.all.first`.

He3lixxx: Ah, well, forgot about that. I'm fine with the code, then :)

evap/student/templates/student_index.html
```diff
@@ -81,34 +81,44 @@
...
                                         {% endif %}
                                     </td>
-                                    <td class="text-right">
-                                        {% if course.grades_activated and course.grade_documents.count == 1 and can_download_grades %}
-                                            <a class="btn btn-sm btn-light" href="{% url 'grades:download_grades' course.grade_documents.all.first.id %}" role="button">{{ course.grade_documents.all.first.description }}</a>
-                                        {% elif course.grade_documents.count > 1 and can_download_grades %}
-                                            <div class="btn-group" role="group">
-                                                <a id="btnDownload{{ course.id }}" role="button" class="btn btn-sm btn-light dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">{% trans 'Grades' %}</a>
-                                                <div class="dropdown-menu" aria-labelledby="btnDownload{{ course.id }}">
+                                    {% if can_download_grades %}
```
karyon: `and semester|are_grades_activated`? because for these semesters that column will be empty for no apparent reason. the alternative would be to explain the reason somehow.

evap/student/templates/student_index.html
```diff
@@ -66,9 +69,6 @@
                                                 <span class="text text-primary" data-toggle="tooltip" data-placement="top" title="{% trans 'You did not evaluate this course.' %}"><span class="far fa-frown" aria-hidden="true"></span></span>
                                         {% endif %}
                                     </td>
-                                    <td>
-                                        {{ course.student_state|studentstatename }}
-                                    </td>
                                     <td>
                                         <span class="line">{{ course.vote_start_datetime }} &ndash;</span> <span class="line">{{ course.vote_end_date }}</span>
                                         {% if course.state == 'in_evaluation' and course in due_courses %}
```
janno42: add additional badges for starting

evap/student/templates/student_index.html
```diff
@@ -66,9 +69,6 @@
                                                 <span class="text text-primary" data-toggle="tooltip" data-placement="top" title="{% trans 'You did not evaluate this course.' %}"><span class="far fa-frown" aria-hidden="true"></span></span>
                                         {% endif %}
                                     </td>
-                                    <td>
-                                        {{ course.student_state|studentstatename }}
-                                    </td>
```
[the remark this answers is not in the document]
janno42: there is no "Evaluated" button. there is a disabled "Evaluate" button until the evaluation period starts. then it's enabled and removed once the student evaluated the course.

i can add the gray result bar for unpublished courses, but the result bar tooltip text would have to be changed (adds an additional variable for the template include), because "Not enough answers were given." does not apply here. also, students wouldn't have a visual separation between published courses where they can't see the results and courses they already voted for where results might become available in the future. they could only figure this out by parsing the evaluation period.
it would be better to have a clear visual difference between these two states "you cannot see results because we only got little feedback" and "you cannot see results because the evaluation is still running" - we can't really use the exact same ui element. maybe another icon would work.
what's your opinion?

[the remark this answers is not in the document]
janno42: we could also make a clear separation between running evaluations and all the other courses by, e.g.,  highlighting them.

there are multiple options for ordering the courses on the student index - each will probably have slightly different requirements for the results bar/highlighting/whatever. so we should decide that first:
- currently due evaluations (courses in the evaluation period where the user can still vote) should probably be on top.
- published results could be on the bottom.
- the remaining courses in between can be divided into three groups: 1. courses in preparation, 2. courses in evaluation for which the user already voted and 3. courses where the evaluation ended but which are not yet published. which order should we use for these? 1, 2, 3?

we could also use a strictly time-linear order: courses in preparation, due courses in evaluation, others courses in evaluation, unpublished, published.

what would you prefer?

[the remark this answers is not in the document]
janno42: and another question: in which states should we display the "will start/end in n days/tomorrow/..." badges?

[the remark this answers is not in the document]
janno42: and in which states do we want to show the participation bar?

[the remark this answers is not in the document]
janno42: rabbit hole. this is more an overall student index redesign. new issue for that?

evap/student/templates/student_index.html
janno42: check

evap/student/templates/student_index.html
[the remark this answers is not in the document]
karyon: yeah, okay. in that case i would however prefer to have no changes at all to the student pages in this branch.

PR #1165, opened 2018-04-30
Solve #1142 -- Fix tests failing at night

### LINKED ISSUE #1142: Test failure depending on time of day
We currently have 4 tests that fail if your system time is between 00:00 and 00:00 + EVALUATION_END_OFFSET_HOURS:

- evap.evaluation.tests.test_models.TestCourses:test_evaluation_ended
- evap.evaluation.tests.test_models.TestCourses:test_in_evaluation_to_evalued
- evap.evaluation.tests.test_models.TestCourses:test_in_evaluation_to_published
- evap.evaluation.tests.test_models.TestCourses:test_in_evaluation_to_reviewed

felixrindt:
Fixes #1142

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -21,7 +21,6 @@ from django_fsm.signals import post_transition
 # see evaluation.meta for the use of Translate in this file
 from evap.evaluation.meta import LocalizeModelBase, Translate
 from evap.evaluation.tools import date_to_datetime, get_due_courses_for_user
-from evap.settings import EVALUATION_END_OFFSET_HOURS, EVALUATION_END_WARNING_PERIOD
 
 logger = logging.getLogger(__name__)
 
@@ -272,7 +271,7 @@ class Course(models.Model, metaclass=LocalizeModelBase):
     @property
     def vote_end_datetime(self):
         # The evaluation ends at EVALUATION_END_OFFSET_HOURS:00 of the day AFTER self.vote_end_date.
-        return date_to_datetime(self.vote_end_date) + timedelta(hours=24 + EVALUATION_END_OFFSET_HOURS)
+        return date_to_datetime(self.vote_end_date) + timedelta(hours=24 + settings.EVALUATION_END_OFFSET_HOURS)
 
     @property
     def is_in_evaluation_period(self):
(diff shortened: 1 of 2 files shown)


He3lixxx:
The changes you made look good. I notices the following things in the existing codebase though:

In `evaluation/test/test_models.py`:
- `test_evaluation_doesnt_end_soon` has no assertion in it, it basically does nothing. Thus, the variable `course` is assigned but never used.
- In `test_evaluation_ends_soon`: `evaluation_end = date.today() - timedelta(hours=24)` with `EVALUATION_END_WARNING_PERIOD = 24` produces exactly the edge case where the test could possibly fail if executed at around 0:00 if the course is created before the date change and the assertion is made after the date change.
- settings is still imported from django.conf but never used.

I guess this should be put into another Issue. @karyon, @janno42 what do you think?

janno42:
if the second point is a real problem it should probably be addressed in this pr, shouldn't it?

felixrindt:
What would you do about that? It's behaving exactly as expected. One could override the actual time, but I don't know wether that's easy to do using e.g. the testing framework.

He3lixxx:
Actually, thinking about it more and trying to come up with a fix, I think this test is somewhat broken and only tests the current, unintuitive semantics of `evaluation_ends_soon` where it always returns `True` after a course's evaluation has finished.
In the test class, `EVALUATION_END_OFFSET_HOURS` is overridden with value `0`. This means, for any course, the evaluation ends as soon as the `end_date` day ends (which equals 0:00 on the day after `end_date`).
For our test course, we set the `end_date` to `date.today() - timedelta(hours=24)`, which is the day before the test is running. So, the evaluation ended the day the test is running at 0:00. So, whenever the test is running, the course is not in evaluation anymore. Thus, we're only testing that `evaluation_ends_soon` returns `True` after a courses evaluation has finished.

To make this really test what it's supposed to and also allow the semantics of `evaluation_ends_soon` we agreed on, we should remove the timedelta subtraction. This would mean that today is the last day of evaluation and the user has <24h left to evaluate. Thus, the warning should be displayed, since we override `EVALUATION_END_WARNING_PERIOD` with `24`. In theory, this is another issue, practically we could do it in this pr.

I feel like this whole system (end_date is a date, where the 0:00 mark of the next day is used as delimiter, except we have a offset, then the offset is added) is unnecessary complicated and error prone, maybe we should change that?

karyon:
just a quick note, I'll change evaluation_ends_soon to `return 0 < self.time_left_for_evaluation.total_seconds() < EVALUATION_END_WARNING_PERIOD * 3600` in my next PR. feel free to do this here already. I'll read through the discussion and think about it later today.

He3lixxx:
In this case, @karyon will have to look into the tests anyway, so we could merge this pr and handle the tests when he changes the behvaviour of evaluation_ends_soon.

PR #1169, opened 2018-05-05
Refactor vote and vote_preview student views 

### LINKED ISSUE #1144: Refactor vote and vote_preview student views
Replication of https://github.com/fsr-itse/EvaP/pull/1125#discussion_r177552984

Context:
```
course_form_group_top = [questions_form for questions_form in course_form_group if questions_form.questionnaire.is_above_contributors]
course_form_group_bottom = [questions_form for questions_form in course_form_group if questions_form.questionnaire.is_below_contributors]
```
Quote:
what if we let `helper_create_voting_form_groups` return `(course_form_group_top, contributor_form_groups, course_form_group_bottom)`? the transformation of `contributor_form_groups` to that other format with the label etc. would stay here i think. for the `if not all...` form validation we would need a list along the lines of `course_form_group_top + contributor_form_groups.values() + course_form_group_bottom`.

karyon:
among other stuff. fixes #1144. best reviewed per commit.

### DIFF
--- a/evap/evaluation/tests/test_models.py
+++ b/evap/evaluation/tests/test_models.py
@@ -66,16 +66,29 @@ class TestCourses(TestCase):
 
     @override_settings(EVALUATION_END_WARNING_PERIOD=24)
     def test_evaluation_ends_soon(self):
-        course = mommy.make(Course, state='in_evaluation', vote_start_datetime=datetime.now() - timedelta(days=2),
-                            vote_end_date=date.today() - timedelta(hours=24), is_graded=False)
+        course = mommy.make(Course, vote_start_datetime=datetime.now() - timedelta(days=2),
+                            vote_end_date=date.today() + timedelta(hours=24))
 
-        Course.update_courses()
+        self.assertFalse(course.evaluation_ends_soon())
+
+        course.vote_end_date = date.today()
         self.assertTrue(course.evaluation_ends_soon())
 
-    @override_settings(EVALUATION_END_WARNING_PERIOD=24)
-    def test_evaluation_doesnt_end_soon(self):
-        course = mommy.make(Course, state='in_evaluation', vote_start_datetime=datetime.now() - timedelta(days=5),
-                            vote_end_date=date.today() - timedelta(hours=24), is_graded=False)
+        course.vote_end_date = date.today() - timedelta(hours=48)
+        self.assertFalse(course.evaluation_ends_soon())
+
+    @override_settings(EVALUATION_END_WARNING_PERIOD=24, EVALUATION_END_OFFSET_HOURS=24)
+    def test_evaluation_ends_soon(self):
+        course = mommy.make(Course, vote_start_datetime=datetime.now() - timedelta(days=2),
+                            vote_end_date=date.today())
+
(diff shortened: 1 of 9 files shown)


evap/evaluation/tests/test_models.py
```diff
@@ -66,16 +66,16 @@ def test_in_evaluation_to_published(self):
...
-        course = mommy.make(Course, state='in_evaluation', vote_start_datetime=datetime.now() - timedelta(days=2),
-                            vote_end_date=date.today() - timedelta(hours=24), is_graded=False)
+        course = mommy.make(Course, vote_start_datetime=datetime.now() - timedelta(days=2),
+                            vote_end_date=date.today() + timedelta(hours=24))
 
-        Course.update_courses()
+        self.assertFalse(course.evaluation_ends_soon())
+
+        course.vote_end_date = date.today()
         self.assertTrue(course.evaluation_ends_soon())
```
He3lixxx: Two thoughts:
- This test might possibly fail if executed around midnight and the `date.today()` calls are on another day than the `evaluation_ends_soon()` calls. It's highly improbably that this happens, though, and we don't have an easy way to mock the time (as e.g.  freezegun could do), so I'm okay with leaving it like this. What do you think?
- Since we're overriding the `EVALUATION_END_OFFSET_HOURS` with 0 for this whole test class, we don't really test whether `evaluation_ends_soon` takes that setting into account. Do we want it to?

evap/student/views.py
```diff
@@ -65,42 +65,45 @@ def vote_preview(request, course, for_rendering_in_modal=False):
     return render(request, "student_vote.html", template_data)
 
 
+def render_vote_page(request, course):
```
He3lixxx: What do you think about a name like `get_valid_form_groups_or_render_vote_page`? It took me a while to understand the `if all(all(...))` line, I think this name change could help.

evap/student/views.py
```diff
@@ -81,16 +64,17 @@ def render_vote_page(request, course):
         errors_exist=any(any(form.errors for form in form_group) for form_group in form_groups.values()),
         course_form_group_top=[questions_form for questions_form in course_form_group if questions_form.questionnaire.is_above_contributors],
         course_form_group_bottom=[questions_form for questions_form in course_form_group if questions_form.questionnaire.is_below_contributors],
-        contributor_form_groups=list((contribution.contributor, contribution.label, form_group, helper_has_errors(form_group)) for contribution, form_group in form_groups.items()),
+        contributor_form_groups=list((contribution.contributor, contribution.label, form_group, any(form.errors for form in form_group)) for contribution, form_group in form_groups.items()),
```
He3lixxx: Is there a reason here to prefer list() over []?

evap/student/views.py
```diff
@@ -41,38 +41,21 @@ def index(request):
     return render(request, "student_index.html", template_data)
 
 
-def vote_preview(request, course, for_rendering_in_modal=False):
```
He3lixxx: What do you think about chanting the default value for preview or removing it? This would force all calls to explicitly state whether we want a preview rendering where no submission can be made or a no-preview rendering which can be submitted which could help identifying bugs.

PR #1171, opened 2018-05-07
applied #1163 Rename rewarded course attribute

### LINKED ISSUE #1163: Rename rewarded course attribute
The course attribute `is_required_for_reward` should be renamed to `is_rewarded`.
https://github.com/fsr-itse/EvaP/pull/1157 adds proportional rewards, so an individual course is not required for getting a reward, but can contribute to get a reward at some point. The new name better represents the actual meaning - whether evaluating this course is rewarded (possibly not directly but in combination with others) or not at all.

phoeinx:
fixes #1163


deployment/apache.template.conf
```diff
@@ -1,28 +0,0 @@
-<VirtualHost *:80>
```
janno42: umm, you accidentally removed some files here ;)

janno42:
please reset the changes to the bootstrap submodule and i think then we're ready to go

karyon:
lgtm

PR #1172, opened 2018-05-07
Personal result export. Fixes #1160

### LINKED ISSUE #1160: Personal result export
On each user profile page the list "Courses contributing to (active semester)" should be extended to hold all courses the user is contributing to, grouped by semester (current semester on top). Each entry has an "Personal export" button that links to this course's results page where
- individual cards for all other contributors are not shown (only collapsing is not sufficient because in some situations we have more than 20 contributors and the card headers would still use up too much space) and
- only text answers which are visible for this user are shown.

This helps to export results for a user as PDF on request. The buttons linking to the personal export view could also be added on the contributor index page.

LeaGe:
Fixes #1160

### DIFF
--- a/evap/results/templates/results_course_detail.html
+++ b/evap/results/templates/results_course_detail.html
@@ -21,7 +21,7 @@
         <div class="alert alert-warning">{% trans 'This is a preview. The course is still in the process of evaluation.' %}</div>
     {% endif %}
 
-    {% if reviewer or contributor %}
+    {% if is_reviewer or is_contributor_or_delegate %}
         {% if course.is_private %}
             <div class="alert alert-info d-print-none">
                 <span class="fas fa-lock"></span> {% trans 'This course is private. Only contributors and participants can see the results.' %}
@@ -31,14 +31,19 @@
             <h3 class="mr-auto">{{ course.name }} ({{ course.semester.name }})</h3>
 
             <div class="btn-switch btn-switch-light my-auto d-print-none">
-                <div class="btn-switch-label">{% trans 'View as' %}</div>
+                <div class="btn-switch-label">{% trans 'View' %}</div>
                 <div class="btn-group">
-                    <a href="{% url 'results:course_detail' course.semester.id course.id %}?public_view=false" role="button" class="btn btn-sm btn-light{% if not public_view %} active{% endif %}" data-toggle="tooltip" data-placement="top"
+                    {% if user.is_staff and view == 'export' or is_contributor %}
+                        <a href="{% if is_contributor %}{% url 'results:course_detail' course.semester.id course.id %}?view=export{% endif %}" role="button" class="btn btn-sm btn-light{% if view == 'export' %} active{% endif %}" data-toggle="tooltip" data-placement="bottom"
+                        title="{% blocktrans %}Shows filtered view meant for personal export. Only your personal results and those about the course. Other contributors' results and private answers are hidden.{% endblocktrans %}">
+                        {% trans 'Export' context 'view mode' %}
+                    {% endif %}
+                    <a href="{% url 'results:course_detail' course.semester.id course.id %}?view=full" role="button" class="btn btn-sm btn-light{% if view == 'full' %} active{% endif %}" data-toggle="tooltip" data-placement="bottom"
                     title="{% trans 'Shows all results available for you.' %}">
-                    {% trans 'Myself' %}
+                    {% trans 'Full' %}
                     </a>
                     {% if not course.can_publish_rating_results %}
-                        <div class="disabled-tooltip" data-toggle="tooltip" data-placement="top" title="{% blocktrans %}The results of this course have not been published because it didn't get enough votes.{% endblocktrans %}">
+                        <div class="disabled-tooltip" data-toggle="tooltip" data-placement="bottom" title="{% blocktrans %}The results of this course have not been published because it didn't get enough votes.{% endblocktrans %}">
                             <a href="#" role="button" class="btn btn-sm btn-light disabled disabled-right">
                                 {% if course.is_private %}
                                     {% trans 'Participant' %}
@@ -48,7 +53,7 @@
                             </a>
                         </div>
                     {% else %}
-                        <a href="{% url 'results:course_detail' course.semester.id course.id %}?public_view=true" role="button" class="btn btn-sm btn-light{% if public_view %} active{% endif %}" data-toggle="tooltip" data-placement="top"
+                        <a href="{% url 'results:course_detail' course.semester.id course.id %}?view=public" role="button" class="btn btn-sm btn-light{% if view == 'public' %} active{% endif %}" data-toggle="tooltip" data-placement="bottom"
                            title="
                         {% if course.is_private %}
                             {% trans 'Shows results available for the participants.' %}">
@@ -162,6 +167,20 @@
                         </div>
                     </div>
                 {% endfor %}
+                {% if view == 'export' and other_contributors %}
+                    <div class="card mt-3">
+                        <div class="card-header">
+                            {% trans 'Other contributors' %}
+                        </div>
+                        <div class="card-body">
+                            <ul>
+                                {% for contributor in other_contributors %}
+                                    <li>{{ contributor.full_name }}</li>
+                                {% endfor %}
+                            </ul>
+                        </div>
+                    </div>
+                {% endif %}
             </div>
         </div>
     {% endif %}
--- a/evap/results/tests/test_views.py
+++ b/evap/results/tests/test_views.py
@@ -2,7 +2,8 @@ from django.contrib.auth.models import Group
 from django.test.testcases import TestCase
 from model_mommy import mommy
 
-from evap.evaluation.models import Semester, UserProfile, Course, Contribution, Questionnaire, Degree, Question, RatingAnswerCounter
+from evap.evaluation.models import Semester, UserProfile, Course, Contribution, Questionnaire, Degree, Question, \
+    RatingAnswerCounter, TextAnswer
 from evap.evaluation.tests.tools import ViewTest, WebTest
 from evap.results.views import get_courses_with_prefetched_data
 
@@ -525,7 +526,7 @@ class TestResultsTextanswerVisibilityForExportView(WebTest):
     @classmethod
     def setUpTestData(cls):
         manager_group = Group.objects.get(name="Manager")
-        mommy.make(UserProfile, username="manager", groups=[manager_group])
+        cls.manager = mommy.make(UserProfile, username="manager", groups=[manager_group])
 
     def test_textanswer_visibility_for_responsible(self):
         page = self.app.get("/results/semester/1/course/1?view=export", user='responsible')
@@ -659,7 +660,7 @@ class TestResultsTextanswerVisibilityForExportView(WebTest):
         self.assertNotIn(".other_responsible_orig_private.", page)
         self.assertNotIn(".other_responsible_orig_notreviewed.", page)
 
-    def test_textanswer_visibility_for_staff(self):
+    def test_textanswer_visibility_for_manager(self):
         contributor_id = UserProfile.objects.get(username="responsible").id
         page = self.app.get("/results/semester/1/course/1?view=export&contributor_id={}".format(contributor_id), user='manager')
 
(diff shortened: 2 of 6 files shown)


evap/results/views.py
```diff
@@ -63,13 +63,17 @@ def course_detail(request, semester_id, course_id):
...
+        contributor_id = int(request.GET.get('contributor_id'))
 
     # If grades are not published, there is no public view
-    if not course.can_publish_grades:
-        public_view = False
+    if not course.can_publish_grades and view == 'public':
+        view = 'full'
 
     represented_users = list(request.user.represented_users.all())
     represented_users.append(request.user)
```
karyon: `request.user` is used in various places where i think it would be correct to use the user passed via the contributor_id parameter.

evap/results/views.py
```diff
@@ -81,7 +85,11 @@ def course_detail(request, semester_id, course_id):
         results = []
         for result in section.results:
             if isinstance(result, TextResult):
-                answers = [answer for answer in result.answers if user_can_see_text_answer(request.user, represented_users, answer, public_view)]
+                if view == 'filter':
+                    answers = [answer for answer in result.answers if user_can_see_text_answer(UserProfile.objects.get(id=contributor_id), [], answer, view)]
```
karyon: the `.get` executes a database query, which is rather expensive, so this should be done repeatedly in a loop.

evap/results/views.py
```diff
@@ -81,7 +85,11 @@ def course_detail(request, semester_id, course_id):
...
+                else:
+                    answers = [answer for answer in result.answers if user_can_see_text_answer(request.user, represented_users, answer, view)]
```
karyon: above is some code collecting the represented users. if you set that to `[]` in case of the filtered view, then you don't need the if else in here

evap/results/views.py
```diff
@@ -143,12 +151,16 @@ def course_detail(request, semester_id, course_id):
...
     return render(request, "results_course_detail.html", template_data)
 
 
-def user_can_see_text_answer(user, represented_users, text_answer, public_view=False):
-    if public_view:
+def user_can_see_text_answer(user, represented_users, text_answer, view='full'):
+    if view == 'public':
+        return False
+    if view == 'filter' and text_answer.is_private:
         return False
```
karyon: shouldn't comments of other contributors `return False` as well?

evap/staff/templates/staff_user_form.html
```diff
@@ -40,11 +40,22 @@ <h5 class="card-title">{% trans 'CC-User for' %}</h5>
...
             <div class="card mb-3">
                 <div class="card-body">
-                    <h5 class="card-title">{% trans 'Courses contributing to (active semester)' %}</h5>
+                    <h5 class="card-title">{% trans 'Personal result pages' %}</h5>
                     {% if courses_contributing_to %}
                         <ul>
-                            {% for course in courses_contributing_to %}
-                                <li>{{ course.name }}</li>
+                            {% for semester, courses in courses_contributing_to %}
+                            {% if courses %}
```
karyon: indent plz

evap/staff/views.py
```diff
@@ -1173,7 +1173,7 @@ def notify_reward_points(users, **kwargs):
     user = get_object_or_404(UserProfile, id=user_id)
     form = UserForm(request.POST or None, request.FILES or None, instance=user)
 
-    courses_contributing_to = Course.objects.filter(semester=Semester.active_semester(), contributions__contributor=user)
+    courses_contributing_to = [(semester, Course.objects.filter(semester=semester, contributions__contributor=user)) for semester in Semester.objects.all()]
```
karyon: this includes all the semesters, even if the contributor had to courses in it. the result could be rather ugly in the UI

He3lixxx:
Nice :D. I reviewed the code too and found nothing except the stuff karyon mentioned, so if his comments are adressed, regard this as approved by me :)

evap/staff/templates/staff_user_form.html
```diff
@@ -40,11 +40,22 @@ <h5 class="card-title">{% trans 'CC-User for' %}</h5>
             </div>
             <div class="card mb-3">
                 <div class="card-body">
-                    <h5 class="card-title">{% trans 'Courses contributing to (active semester)' %}</h5>
+                    <h5 class="card-title">{% trans 'Personal result pages' %}</h5>
                     {% if courses_contributing_to %}
```
janno42: the if should be around the card so that it is only shown if there are courses the user contributed to

evap/results/views.py
```diff
@@ -63,13 +63,17 @@ def course_detail(request, semester_id, course_id):
...
 
     if request.user.is_reviewer:
-        public_view = request.GET.get('public_view') != 'false'  # if parameter is not given, show public view.
+        view = request.GET.get('view', 'public')  # if parameter is not given, show public view.
     else:
-        public_view = request.GET.get('public_view') == 'true'  # if parameter is not given, show own view.
+        view = request.GET.get('view', 'full')  # if parameter is not given, show own view.
+
+    contributor_id = request.user.id
+    if view == 'filter' and user.is_staff:
```
janno42: `user` is not defined here (should be `request.user`)

evap/staff/views.py
janno42: it's actually filtered in the ui. however it would be better to not add them at all.

evap/results/templates/results_course_detail.html
```diff
@@ -35,9 +35,14 @@ <h3 class="mr-auto">{{ course.name }} ({{ course.semester.name }})</h3>
             <div class="btn-switch my-auto d-print-none">
                 <div class="btn-switch-label">{% trans 'View as' %}</div>
                 <div class="btn-group">
-                    <a href="{% url 'results:course_detail' course.semester.id course.id %}?public_view=false" role="button" class="btn btn-sm btn-light{% if not public_view %} active{% endif %}" data-toggle="tooltip" data-placement="top"
+                    {% if user.is_staff and view == 'filter' or contributor %}
+                        <a href="{% url 'results:course_detail' course.semester.id course.id %}?view=filter" role="button" class="btn btn-sm btn-light{% if view == 'filter' %} active{% endif %}" data-toggle="tooltip" data-placement="top"
+                        title="{% trans 'Shows filtered view for personal export.' %}">
```
janno42: please make this more verbose, like `Shows filtered view meant for personal export. Only your personal results and those about the course. Other contributors' results and private answers are hidden.`

evap/results/templates/results_course_detail.html
```diff
@@ -143,39 +148,41 @@ <h3>{{ course.name }} ({{ course.semester.name }})</h3>
             </div>
             <div class="card-body">
                 {% for contributor, contributor_data in contributor_sections.items %}
-                    <div class="card{% if not forloop.last %} mb-3{% endif %}">
-                        {# Logic for collapsing Contributor Sections with no votes. #}
-                        {% if contributor_data.total_votes == 0 %}
-                            <div class="card-header d-flex">
-                                <div class="mr-auto">
-                                    <a class="collapse-toggle collapsed" data-toggle="collapse" href="#contributor-{{ contributor.id }}" aria-expanded="false" aria-controls="contributor-{{ contributor.id }}">
+                    {% if not view == 'filter' or contributor_id == contributor.id %}
```
janno42: please add a list of other contributors as stated in the issue :)

evap/results/views.py
janno42: other contributors are completely removed, but for consistency: probably yes.

evap/results/views.py
He3lixxx: Maybe we should add a test? Something like this should trigger a test failure imo.

evap/staff/views.py
karyon: to me it looks like the filtering needs to be done only once, so it can actually be moved from the template to here. and since it's not more complexity doing it here than there, i don't see why we should carry data around that's never used anyways and not filter it right at the place where it's produced.

evap/results/views.py
karyon: if i understand correctly, you're referring to `{% if not view == 'filter' or view_as_user.id == contributor.id %}`. to me that specific check is the same case as below, technically it does not matter where we do the filtering, so i would do it at the earliest possible time, because why give data to the template where we know in the view already that it's not gonna be used?

regarding `user_can_see_text_answer`, i'd say a method is broken or not regardless of whether there is other code covering up the error ;) so yeah, i'm strongly in favor of fixing this.

evap/results/templates/results_course_detail.html
```diff
@@ -134,32 +139,48 @@ <h3>{{ course.name }} ({{ course.semester.name }})</h3>
...
-                                    {% if contribution_result.label %}
-                                        &ndash; <i>{{ contribution_result.label }}</i>
-                                    {% endif %}
-                                </a>
-                            </div>
-                            {% if not contribution_result.has_answers %}
-                                <div class="participants-warning">
-                                    <span class="fas fa-info-circle"></span>
-                                    {% trans 'There are no results for this person.' %}
+                    {% if not view == 'filter' or view_as_user.id == contribution_result.contributor.id %}
```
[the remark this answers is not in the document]
janno42: users should see their own card, so that will also be shown in personal view.

evap/results/views.py
```diff
@@ -104,21 +104,27 @@ def course_detail(request, semester_id, course_id):
...
 
-    # redirect to non-public view if there is none because the results have not been published
-    if not course.can_publish_rating_results:
-        public_view = False
+    view_as_user = UserProfile.objects.get(id=request.user.id)
+    if view == 'filter' and request.user.is_staff:
+        view_as_user = UserProfile.objects.get(id=int(request.GET.get('contributor_id')))
 
-    represented_users = list(request.user.represented_users.all()) + [request.user]
+    if view == 'filter':
```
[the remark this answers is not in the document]
janno42: rename to personal was done. i would stick with that in the UI to not make the name longer. we can rename it to personal_export in the code if you like.

evap/results/templates/results_course_detail.html
```diff
@@ -33,9 +33,14 @@ <h3 class="mr-auto">{{ course.name }} ({{ course.semester.name }})</h3>
             <div class="btn-switch my-auto d-print-none">
                 <div class="btn-switch-label">{% trans 'View as' %}</div>
                 <div class="btn-group">
-                    <a href="{% url 'results:course_detail' course.semester.id course.id %}?public_view=false" role="button" class="btn btn-sm btn-light{% if not public_view %} active{% endif %}" data-toggle="tooltip" data-placement="top"
+                    {% if user.is_staff and view == 'filter' or contributor %}
+                        <a href="{% url 'results:course_detail' course.semester.id course.id %}?view=filter" role="button" class="btn btn-sm btn-light{% if view == 'filter' %} active{% endif %}" data-toggle="tooltip" data-placement="top"
+                        title="{% blocktrans %}Shows filtered view meant for personal export. Only your personal results and those about the course. Other contributors' results and private answers are hidden.{% endblocktrans %}">
```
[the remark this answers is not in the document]
janno42: was changed, the button is now disabled for managers.

evap/results/templates/results_course_detail.html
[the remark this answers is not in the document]
karyon: as far as i can see, the view does not care whose comments it's processing, could you check whether that actually works?

in any case, i think this decision what data to show should be done in the view. actually, that even includes the `view == 'filter'` part.

evap/results/views.py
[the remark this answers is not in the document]
karyon: how about we call it export in the ui? (and perhaps also in the code?)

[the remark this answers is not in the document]
janno42: export it is

janno42:
i addressed the review comments, @karyon. the only remaining thing are the text answer visibility tests in `results/tests/test_views.py`.

evap/results/views.py
```diff
@@ -213,6 +228,9 @@ def user_can_see_text_answer(user, represented_users, text_answer, public_view=F
     if text_answer.is_private:
         return contributor == user
 
+    if view == 'export' and not text_answer.contribution.is_general and contributor != user:
+        return False
```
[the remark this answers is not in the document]
janno42: changed visibility check and added test.

evap/results/tests/test_views.py
```diff
@@ -682,6 +683,28 @@ def test_textanswer_visibility_for_staff(self):
...
+        mommy.make(TextAnswer, question=question, contribution=course.general_contribution, answer="general_contribution", state=TextAnswer.PUBLISHED)
+        mommy.make(TextAnswer, question=question, contribution=responsible_contribution, answer="responsible_contribution", state=TextAnswer.PUBLISHED)
+        mommy.make(TextAnswer, question=question, contribution=manager_contribution, answer="manager_contribution", state=TextAnswer.PUBLISHED)
+        mommy.make(TextAnswer, question=question, contribution=manager_contribution, answer="manager_contribution_private", state=TextAnswer.PRIVATE)
+
+        page = self.app.get("/results/semester/{}/course/{}?view=export&contributor_id={}".format(course.semester.id, course.id, self.manager.id), user='manager')
+        self.assertNotIn("general_contribution", page)
+        self.assertNotIn("responsible_contribution", page)
+        self.assertIn("manager_contribution", page)
+        self.assertNotIn("manager_contribution_private", page)
```
karyon: can you keep the structure of the other tests?

karyon: or was there a reason for not doing that?

janno42: we don't have that specific situation in the fixture data so i created everything we need here.

karyon: would it work to add a responsible/contributor to the staff group, or is something missing then?

janno42: that should be it. would you add an additional user or add the group to an existing one and remove it at the end of the test?
i thought you'd appreciate not using the fixture here as a start for a brighter future ;)

janno42: yeah, but i didn't want to do that here because that seems like really a lot of stuff needed. so we could open an issue for replacing the fixture and here i'm changing the group memberships for this one test?

PR #1174, opened 2018-05-08
Renames can_publish_grades to has_enough_votes_to_publish_grades, some related renames and cleanups

karyon:
This pull request renames can_publish_grades to has_enough_votes_to_publish_grades and does some related renames and cleanups.


evap/evaluation/models.py
```diff
@@ -362,7 +362,7 @@ def can_staff_delete(self):
         return self.can_staff_edit and (not self.num_voters > 0 or self.is_single_result)
 
     @property
-    def can_publish_grades(self):
+    def has_enough_votes_to_publish_grades(self):
```
janno42: i just noticed that the method does not actually check votes but voters. so we should either name it `has_enough_voters_to_publish_grades` or really count the votes. i'm wondering whether counting votes (would be the maximum value of the votes per question) might actually be a better thing to base this decision on.

karyon: yeah, a "vote" as used right now is more or less the set of answers a person submitted. 

i think the primary argument for this was anonymity in case of MIN_ANSWER_COUNT and significance/validity of the results in case of MIN_ANSWER_PERCENTAGE? for anonymity it does not matter whether they actually gave any answers or not. for validity, it kind of does, but i think it's a bit weird to not publish results because there were people who submitted but decided to not answer some questions...

janno42: maybe if students voted in such a way that there is not a single questions that received MIN_ANSWER_PERCENTAGE answers, we shouldn't publish the results to follow the original idea of the significance. evaluating at the results it does not really make a difference whether one person answered the questions or two people answered disjunctive subsets :)

karyon: but then one question that meets the MIN_ANSWER_PERCENTAGE is sufficient to publish all the other questions which might have a much lower number of answers, that doesn't make much sense either.

janno42: i would say that's at least one question meeting our demand which is more than none. i'm thinking about cases where students each only vote for parts of the course and think that it might be better to only publish the results if at least for some questions (or in a very extreme case one single question) the minimum number of people answered.

janno42: currently we're publishing even if the MIN_ANSWER_PERCENTAGE was not met for any question because it's really a MIN_VOTERS_PERCENTAGE now.so if we want to keep that we will need additional renaming.

karyon: i would rather not move from one wrong solution to another similarly but differently wrong solution. if we go down this rabbit hole, lets find a proper solution. but actually i'm against going down this rabbit hole because we'll never get the time back we spend on this :) maybe open a ticket? :D

you're right about the options. i'll rename them if you agree to not do any behavior changes in this PR.

karyon:
please review again (and consider opening a ticket for changing the logic)

janno42:
created issue, approved pr :)

karyon:
@He3lixxx please review since you brought this up :)

PR #1178, opened 2018-05-10
Remove text answers for grading questions in test data

janno42:


karyon:
can you please find out how this happened?

karyon:
that commit should replace the id column, nothing more. the answers got reordered in the json because of that, maybe that's why you got the impression.

PR #1179, opened 2018-05-10
Clear navbar caches when needed

janno42:


evap/staff/tools.py
```diff
@@ -82,13 +82,23 @@ def custom_redirect(url_name, *args, **kwargs):
...
+
+
+def delete_navbar_cache_for_special_users():
     for user in UserProfile.objects.all():
-        key = make_template_fragment_key('navbar', [user.username, 'de'])
-        cache.delete(key)
-        key = make_template_fragment_key('navbar', [user.username, 'en'])
-        cache.delete(key)
+        if user.is_reviewer or user.is_grade_publisher:
+            delete_navbar_cache_for_user(user)
```
karyon: following this structure, you would also need another `delete navbar_cache_*` method for the `helper_course_edit` case. how about a `delete_navbar_cache_for_users` which takes a list/queryset, and use that everywhere? the two callers of `delete_navbar_cache_for_special_users`, i don't know, maybe create a `users_that_need_their_navbar_cache_deleted_on_semester_creation_or_deletion()` method?

karyon: ...maybe `users_with_semesters_in_their_navbar`? not sure...

janno42: where would you place this method?

janno42: maybe let's just call the `delete_navbar_cache_for_users` method for reviewers and grade publishers there.

karyon: yeah, that was the idea. my problem was that when calling `delete_navbar_cache_for_users`, there will be two places that need to know that these two groups need their caches cleared, so i thought about creating one function that return a list containing those two groups. but it doesn't really matter, you can also duplicate the code.

karyon: if you create that method, i would place it where it's used.

janno42: k, please review again.

evap/staff/views.py
```diff
@@ -364,6 +365,7 @@ def semester_import(request, semester_id):
                 success_messages, warnings, __ = EnrollmentImporter.process(file_content, semester, vote_start_datetime, vote_end_date, test_run=False)
                 forward_messages(request, success_messages, warnings)
                 delete_import_file(request.user.id, import_type)
+                delete_navbar_cache_for_all_users()
```
karyon: derp

evap/staff/views.py
```diff
@@ -1182,6 +1186,7 @@ def notify_reward_points(users, **kwargs):
 
     if form.is_valid():
         form.save()
+        delete_navbar_cache_for_user(user)
```
karyon: i think i would just `delete_navbar_cache_for_users([user]). it's a tiny bit silly but so is having an extra method for that. your call.

janno42: actually already changed that :)

evap/staff/views.py
janno42: hrmpf

evap/staff/views.py
```diff
@@ -623,6 +624,8 @@ def notify_reward_points(users, **kwargs):
             messages.success(request, _("Successfully updated and approved course."))
         else:
             messages.success(request, _("Successfully updated course."))
+            delete_navbar_cache_for_users(course.participants.all())
+            delete_navbar_cache_for_users(UserProfile.objects.filter(contributions__course=course))
```
karyon: this should probably happen outside the if/else.

PR #1180, opened 2018-05-10
Requirements update and cleanup

janno42:
gemnasium is going to shut down on May 15th, 2018 and can be removed from the readme. the requirements are updated to the latest versions and an old landscape file is removed.

- do we want to add a new dependency checker?
- is `psycopg2-binary` fine? (http://initd.org/psycopg/articles/2018/02/08/psycopg-274-released/)


.landscape.yaml
```diff
@@ -1,22 +0,0 @@
...
-pep8:
-  disable:
-    - E501  # line too long - we don't enforce some fixed line length but rather complain about long lines in reviews.
-    - W503  # line break before operator - looks better and PEP8 allows it
-    - E128  # hanging indents not allowed - we indent differently
-    - E126  # overindentation - we indent differently
-    - N802  # function name should be lowercase - lots of false positives in tests that pylint doesn't show
-    - N806  # variable in function should be lowercase - lots of false positives, mostly Formset classes made from factories
-    - N804  # first argument of a classmethod should be named 'cls' - one false positive and pylint does it better
-    - E266  # too many leading '#' for block comment - that's used in settings on purpose
```
karyon: this file is used by the prospector linter, which is currently still enabled in codacy. i'd appreciate if you could rename it to prospector.yaml, though.

janno42:
now using requires.io instead of gemnasium

karyon:
could you also set model mommy to 1.5.1 and check whether the tests run? approved with that.

janno42:
doesn't run with the current 1.5.1 ;)

karyon:
any details?

janno42:
see on travis. same problem we had before.

karyon:
oh, right.  1.5.1 wasn't actually released "13 days ago" as github states, but is older. nevermind then

Issue #1181, opened 2018-05-12
Prevent removing participant who already voted


karyon:
It is possible to delete a participation either through the `CourseForm` and the `UserForm`. Both forms allow to do that even if the user already voted for the course, which is plain incorrect as far as i can see. @janno42 what do you think?

A simple solution would be to silently force all voters to be part of the participants on save. A nicer solution would be to figure out which user got removed but shouldn't be, and display a warning. Making the participants who voted already read-only in the form isn't feasible, i guess. @janno42?

janno42:
We should probably restrict removing them, yes.
It doesn't look like there's a read-only option for individual items in the select2 multiple select widget.

karyon:
warning or no warning?

janno42:
I would say at least a warning that some people haven't been removed. A message with individual names isn't really necessary (but is also not too hard to create, is it?).

PR #1189, opened 2018-05-20
Bugfixes

janno42:
A large number of bugfixes. See individual commit messages for details.


evap/results/templates/results_course_detail.html
```diff
@@ -179,14 +178,27 @@ <h3>{{ course.name }} ({{ course.semester.name }})</h3>
...
+        {% endif %}
+    {% else %}
         <div class="card card-outline-primary">
             <div class="card-header">
                 {% trans 'Course' %}
             </div>
             <div class="card-body">
+                {% for section in course_sections_top %}
+                    {% include 'results_course_detail_questionnaires.html' %}
+                {% endfor %}
```
He3lixxx: I would have merged course_secions_top and course_sections_bottom in the view if there are no contributor_sections. Have you thought about that? I guess it would result in less code duplication, but it's possibly harder to understand.

evap/staff/templates/staff_semester_view_course.html
```diff
@@ -87,11 +87,13 @@
         <span class="badge badge-info">{% if course.is_private %}{% trans 'private' %}{% endif %}</span>
     {% endif %}
 
-    {% if state == 'new' or state == 'prepared' or state == 'editor_approved' and not course.all_contributions_have_questionnaires %}
-        {% if not course.general_contribution_has_questionnaires %}
-            <span class="badge badge-warning">{% trans 'Course has no questionnaires' %}</span>
-        {% else %}
-            <span class="badge badge-warning">{% trans 'Not all contributors have questionnaires' %}</span>
+    {% if not course.all_contributions_have_questionnaires %}
+        {% if state == 'new' or state == 'prepared' or state == 'editor_approved' %}
```
karyon: [optional] i would switch the two ifs

evap/staff/forms.py
```diff
@@ -181,6 +181,16 @@ def validate_unique(self):
                     name_field = e.params["unique_check"][1]
                     self.add_error(name_field, e)
 
+    def clean_participants(self):
+        participants = self.cleaned_data.get('participants')
+        if self.instance.pk:
+            voters = self.instance.voters.all()
+            removed_voters = [voter.username for voter in voters if voter not in participants]
+            if removed_voters:
+                self.add_error("participants", _("Participants who already voted for the course can't be removed: %s") % ", ".join(removed_voters))
```
karyon: ..but now the user has to manually add that specific person again, right? wouldn't it be nice to save the form and show a warning "these people here haven't been removed"?

evap/staff/forms.py
```diff
@@ -181,6 +181,16 @@ def validate_unique(self):
                     name_field = e.params["unique_check"][1]
                     self.add_error(name_field, e)
 
+    def clean_participants(self):
+        participants = self.cleaned_data.get('participants')
+        if self.instance.pk:
+            voters = self.instance.voters.all()
+            removed_voters = [voter.username for voter in voters if voter not in participants]
```
karyon: `removed_voters = set(voters) - set(participants)`? might be a bit clearer and also isn't quadratic in runtime. if that complicates the string formatting too much, that's probably not worth it.

evap/staff/forms.py
```diff
@@ -238,7 +238,9 @@ def __init__(self, *args, **kwargs):
             disable_all_fields(self)
 
         if self.instance.pk:
-            self.fields['responsible'].initial = self.instance.responsible_contributors[0]
+            responsible = self.instance.responsible_contributors[0]
+            self.fields['responsible'].queryset = UserProfile.objects.filter(Q(pk=responsible.pk) | Q(is_active=True))
```
karyon: maybe `self.fields['responsible'].queryset = self.fields['responsible'].queryset | UserProfile.objects.filter(pk=responsible.pk)` if that works? i think that communicates the intent (use the existing queryset but also include the current choice) more clearly.

maybe the |= operator works?

evap/results/tools.py
```diff
@@ -164,7 +164,9 @@ def _calculate_results_impl(course):
             elif question.is_heading_question:
                 results.append(HeadingResult(question=question))
 
-        section_warning = questionnaire_max_answers[(questionnaire, contribution)] < questionnaire_warning_thresholds[questionnaire]
+        section_warning = False
+        if any(question.is_rating_question for question in questionnaire.question_set.all()):
```
karyon: i think it would make more sense to check whether there is any rating question in `results`, because that's what's shown in the UI and not `questionnaire.question_set.all()`.

evap/settings.py
```diff
@@ -121,7 +121,16 @@
             'CLIENT_CLASS': 'django_redis.client.DefaultClient',
             'MAX_ENTRIES': 100000
         }
-    }
+    },
+    'sessions': {
+        'BACKEND': 'django_redis.cache.RedisCache',
+        'LOCATION': 'redis://127.0.0.1:6379/2',
+        'TIMEOUT': None,  # is always invalidated manually
```
karyon: specifying `TIMEOUT` doesn't make sense here. we don't want to cache anything forever in this cache, and the timeout is probably set by django according to the session timeout.

evap/grades/views.py
```diff
@@ -134,7 +134,7 @@ def toggle_no_grades(request):
 @grade_downloader_required
 def download_grades(request, grade_document_id):
     grade_document = get_object_or_404(GradeDocument, id=grade_document_id)
-    if not grade_document.course.grades_activated:
+    if not grade_document.course.grades_activated and not request.user.is_grade_publisher:
         return HttpResponseForbidden()
```
karyon: as per our discussion, this commit is not the right solution anymore, or is it?

evap/staff/forms.py
janno42: i tried that before but it didn't work as expected. will try again, though.

evap/staff/forms.py
janno42: correct. but i really didn't care ;)
should i re-add the user?

evap/grades/views.py
janno42: we didn't decide on anything yet, so there currently is no solution.

janno42: i will probably remove this commit from the pr

evap/staff/forms.py
karyon: since it's probably just one more line, i would say so, yes.

karyon:
regarding the separate session cache, it should probably be cleared in `reload_testdata`, since that is kind of our "nuke all the things" command.

evap/results/templates/results_course_detail.html
janno42: i'm not sure about that. in the end it's just a display thing and semantically having two lists is what we do, but both solutions are fine to me. can i have one more opinion from @karyon, please?

evap/staff/templates/staff_semester_view_course.html
janno42: done

evap/results/tools.py
janno42: check

evap/staff/forms.py
janno42: not too much ;)

evap/staff/forms.py
janno42: shouldn't adding the users to the return value change what's displayed in the form afterwards? i can return a modified list of users, however the removed user stays removed in the form.

evap/results/templates/results_course_detail.html
karyon: right now we are splitting into the three lists in the view, which is a display thing already. splitting into three lists in the view and (if necessary) merging into one list in the template does not make *that* much sense, and personally i find the current solution to have too many branches/code paths in the template, also doing this in the view is probably much shorter, so i would support @He3lixxx's suggestion.

evap/staff/forms.py
janno42: don't know what the problem was before, it works with `|=` ... just perfectly

janno42:
all comments except the one about re-adding removed voters (or courses participated in) have been addressed.

evap/staff/forms.py
karyon: i feel there's a misunderstanding, i was not suggesting saving the wrong data, but overwriting the user's choice on save, thus saving the correct data, and displaying a warning after the redirect that some of the users choices have been ignored. of course that's not 100% nice, but having to add them manually is not that nice either.

in the end it doesn't really matter as it probably happens once every few years. you make the call.

evap/results/tools.py
```diff
@@ -164,7 +166,7 @@ def _calculate_results_impl(course):
             elif question.is_heading_question:
                 results.append(HeadingResult(question=question))
 
-        section_warning = questionnaire_max_answers[(questionnaire, contribution)] < questionnaire_warning_thresholds[questionnaire]
+        section_warning = questionnaire_max_answers[(questionnaire, contribution)] < questionnaire_warning_thresholds[questionnaire] if results_contain_rating_questions else False
```
karyon: `...and results_contain_rating_questions`

evap/evaluation/management/commands/reload_testdata.py
```diff
@@ -36,4 +36,7 @@ def handle(self, *args, **options):
         self.stdout.write('Executing "python manage.py refresh_results_cache"')
         call_command("refresh_results_cache")
 
+        self.stdout.write('Executing "python manage.py clear_cache --cache="sessions""')
+        call_command('clear_cache --cache="sessions"')
```
karyon: you can leave out the quotes around `sessions` here and in the comment to make this less confusing.

evap/student/views.py
```diff
@@ -60,11 +60,20 @@ def get_valid_form_groups_or_render_vote_page(request, course, preview, for_rend
 
     course_form_group = form_groups.pop(course.general_contribution)
 
+    contributor_form_groups = [(contribution.contributor, contribution.label, form_group, any(form.errors for form in form_group)) for contribution, form_group in form_groups.items()]
+    if contributor_form_groups:
+        course_form_group_top = [questions_form for questions_form in course_form_group if questions_form.questionnaire.is_above_contributors]
+        course_form_group_bottom = [questions_form for questions_form in course_form_group if questions_form.questionnaire.is_below_contributors]
+    else:
+        course_form_group_top = [questions_form for questions_form in course_form_group if questions_form.questionnaire.is_above_contributors or questions_form.questionnaire.is_below_contributors]
+        course_form_group_bottom = []
```
karyon: i would prefer something like 
```
course_form_group_top = ...
course_form_group_bottom = ...
if not contributor_form_groups:
    course_form_group_top += course_form_group_bottom 
    course_form_group_bottom  = []
```
it's less code to parse, and is a bit more "here's the common case and here if need be some special handling on top" instead of two equal cases.

karyon:
also, please add a test at least for #1181 since that's about data correctness, and consider adding tests for the others as well.

janno42:
added tests for #1181

PR #1190, opened 2018-05-20
Ui fixes

janno42:
several ui fixes. commits can be squashed as desired.

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -470,7 +470,10 @@ class Course(models.Model, metaclass=LocalizeModelBase):
 
     @property
     def days_until_evaluation(self):
-        return (self.vote_start_datetime.date() - date.today()).days
+        days_left = (self.vote_start_datetime.date() - date.today()).days
+        if self.vote_start_datetime < datetime.now():
+            days_left -= 1
+        return days_left
 
     def is_user_editor_or_delegate(self, user):
         if self.contributions.filter(can_edit=True, contributor=user).exists():
--- a/evap/staff/views.py
+++ b/evap/staff/views.py
@@ -148,7 +148,10 @@ def semester_course_operation(request, semester_id):
         raise SuspiciousOperation("Unknown target state: " + target_state)
 
     course_ids = (request.GET if request.method == 'GET' else request.POST).getlist('course')
-    courses = Course.objects.filter(id__in=course_ids)
+    courses = Course.objects.filter(id__in=course_ids).annotate(
+        midterm_grade_documents_count=Count("grade_documents", filter=Q(grade_documents__type=GradeDocument.MIDTERM_GRADES), distinct=True),
+        final_grade_documents_count=Count("grade_documents", filter=Q(grade_documents__type=GradeDocument.FINAL_GRADES), distinct=True)
+    )
 
     if request.method == 'POST':
         template = None
--- a/evap/contributor/templates/contributor_index.html
+++ b/evap/contributor/templates/contributor_index.html
@@ -63,6 +63,9 @@
                                     <span class="badge badge-primary">{{ degree }}</span>
                                 {% endfor %}
                                 <span class="badge badge-secondary">{{ course.type }}</span>
+                                {% if course.is_single_result %}
+                                    <span class="badge badge-success">{% trans 'Single result' %}</span>
+                                {% endif %}
                                 {% if course in delegated_courses %}
                                     <span class="badge badge-info" data-toggle="tooltip" data-placement="right" title="{% trans 'You are seeing this course because you are a delegate of a lecturer who can edit the course.' %}">{% trans 'Delegate' %}</span>
                                 {% endif %}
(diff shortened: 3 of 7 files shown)


evap/evaluation/models.py
```diff
@@ -470,7 +470,10 @@ def evaluation_ends_soon(self):
 
     @property
     def days_until_evaluation(self):
-        return (self.vote_start_datetime.date() - date.today()).days
+        days_left = (self.vote_start_datetime.date() - date.today()).days
+        if days_left == 0 and self.vote_start_datetime < datetime.now():
+            return -1
+        return days_left
```
He3lixxx: Would `days_left = (self.vote_start_datetime - datetime.now()).days` also work? I'd like that more.

evap/staff/views.py
```diff
@@ -148,7 +148,10 @@ def semester_course_operation(request, semester_id):
         raise SuspiciousOperation("Unknown target state: " + target_state)
 
     course_ids = (request.GET if request.method == 'GET' else request.POST).getlist('course')
-    courses = Course.objects.filter(id__in=course_ids)
+    courses = Course.objects.filter(id__in=course_ids).annotate(
+        midterm_grade_documents_count=Count("grade_documents", filter=Q(grade_documents__type=GradeDocument.MIDTERM_GRADES), distinct=True),
+        final_grade_documents_count=Count("grade_documents", filter=Q(grade_documents__type=GradeDocument.FINAL_GRADES), distinct=True)
+    )
```
He3lixxx: This small change also displays the counts in the UI? Awesome :D. Sometimes I wish I could use django's easy way of doing things in other web applications...

evap/evaluation/models.py
karyon: that would be a semantic change, since right now the `days_until_evaluation` tell you whether the evaluation starts today or tomorrow, and not whether it starts in up to 24 hours or up to 48 hours.

evap/evaluation/models.py
```diff
@@ -470,7 +470,10 @@ def evaluation_ends_soon(self):
...
 
     def is_user_editor_or_delegate(self, user):
         if self.contributions.filter(can_edit=True, contributor=user).exists():
```
karyon: by the way is there a reason why students don't have a "starts today" or "starts tomorrow" badge on their index page?

evap/evaluation/models.py
karyon: the `days_left == 0` doesn't change the output, right? if yes, i would propose

```
if self.vote_start_datetime < datetime.now():
    return -1
return (self.vote_start_datetime.date() - date.today()).days
```

evap/staff/views.py
karyon: well, we already had code for displaying these counts in the template (in an included template to be precise). to me it looks like it previously had some icon plus an empty span.

@janno42, in that case shouldn't the commit message rather be "fix" instead of "add"?

janno42: yes, it's a fix. will change the message.

evap/evaluation/models.py
janno42: no. is not really important but can be added.

evap/evaluation/models.py
janno42: this would always return -1 if the evaluation start was in the past. currently, courses where the start was earlier are sorted on top of the staff course list. we need to know the actual number here.

He3lixxx: Hmm... I think its unintuitive to have this conditional return value. I guess I'd have split this into two methods, one returns calendar days left, one returns full days left. If you both think it's okay we can leave it like this, though.

evap/staff/views.py
He3lixxx: Ah well.

evap/evaluation/models.py
karyon: right.

maybe 
```
days_left = (self.vote_start_datetime.date() - date.today()).days
if self.vote_start_datetime < datetime.now():
    days_left -= 1
return days_left
```
this makes this less special-case-y, doesn't have the weirdness that an evaluation start today and yesterday both result in a -1, which also might make the sorting inconsistent, and avoids the "return -1" which looks a bit like an error code to me.

evap/evaluation/models.py
janno42: will do so in the distribution_bars branch where the student index will be changed anyways

evap/evaluation/models.py
janno42: used proposed solution from the other discussion about the method

janno42:
ready for next review

evap/evaluation/models.py
```diff
@@ -470,7 +470,10 @@ def evaluation_ends_soon(self):
 
     @property
     def days_until_evaluation(self):
-        return (self.vote_start_datetime.date() - date.today()).days
+        days_left = (self.vote_start_datetime.date() - date.today()).days
+        if self.vote_start_datetime < datetime.now():
+            days_left -= 1
+        return days_left
```
karyon: there's a `if course.days_until_evaluation == 0 and state != 'in_evaluation'` somewhere in the template, i think the second part can be removed, please check that.

Issue #1175, opened 2018-05-09
Change publishing threshold calculation


janno42:
In #1174 we started a discussion whether the current publishing threshold really makes sense.

The main question here is: Should a course's results be published if the number of voters in total is `VOTER_COUNT_NEEDED_FOR_PUBLISHING` or greater than that but no question received more than one answer?
Maybe if students voted in such a way that there is not a single questions that received this number (and also the required percentage) of answers, we shouldn't publish the results to follow the original idea to show only significant results. When evaluating the results it does not really make a difference whether one person answered the questions or two people answered disjunctive subsets.

An alternative might be to use the maximum value of votes per single question as a basis for the decision whether the results will be published or not.

janno42:
new threshold were implemented in #1188

PR #1192, opened 2018-05-22
Fix flaky test

karyon:
just in case #1175 turns out to be another deep rabbit hole ;)


karyon:
just in case #1175 turns out to be another deep rabbit hole ;)

karyon:
i would guess it's postgres returning results in different orders depending some non-obvious state that's persistent between tests.

PR #1196, opened 2018-05-27
Make messages about reward points granted by removal compatible with proportional rewards

### LINKED ISSUE #1195: Wrong reward point message when removing participants
`grant_reward_points_after_delete` checks for `if grant_reward_points_if_eligible` when collecting user names that should be displayed in a message. This method returns tuples since 2498f15693cd1fabe71e472f57fb57fa5c51fdd8, so the check is always true and all removed participants are shown in the message.

The actual granting is checked correctly, but the displayed message is not necessarily correct.

pixunil:
Fixes #1195.

Also modifies messages to include the amount, making them more verbose, but I remember that this was not a great deal.

The added test which shows the regression is a bit hacky, but imo straight to the point and short. (It also tests many things together, so I am open for writing more tests)

### DIFF
--- a/evap/rewards/tools.py
+++ b/evap/rewards/tools.py
@@ -64,13 +64,13 @@ def is_semester_activated(semester):
 
 def grant_reward_points_if_eligible(user, semester):
     if not can_user_use_reward_points(user):
-        return 0, False
+        return None, False
     if not is_semester_activated(semester):
-        return 0, False
+        return None, False
     # does the user have at least one required course in this semester?
     required_courses = Course.objects.filter(participants=user, semester=semester, is_rewarded=True)
     if not required_courses.exists():
-        return 0, False
+        return None, False
 
     # How many points have been granted to this user vs how many should they have (this semester)
     granted_points = RewardPointGranting.objects.filter(user_profile=user, semester=semester).aggregate(Sum('value'))['value__sum'] or 0
@@ -79,19 +79,19 @@ def grant_reward_points_if_eligible(user, semester):
     missing_points = target_points - granted_points
 
     if missing_points > 0:
-        RewardPointGranting.objects.create(user_profile=user, semester=semester, value=missing_points)
-        return missing_points, progress >= 1.0
-    return 0, False
+        granting = RewardPointGranting.objects.create(user_profile=user, semester=semester, value=missing_points)
+        return granting, progress >= 1.0
+    return None, False
 
 
 # Signal handlers
 
 @receiver(Course.course_evaluated)
 def grant_reward_points_after_evaluate(request, semester, **_kwargs):
-    points_granted, completed_evaluation = grant_reward_points_if_eligible(request.user, semester)
-    if points_granted:
+    granting, completed_evaluation = grant_reward_points_if_eligible(request.user, semester)
+    if granting:
         message = ngettext("You just earned {count} reward point for this semester.",
-                           "You just earned {count} reward points for this semester.", points_granted).format(count=points_granted)
+                           "You just earned {count} reward points for this semester.", granting.value).format(count=granting.value)
 
         if completed_evaluation:
             message += " " + _("Thank you very much for evaluating all your courses.")
@@ -106,22 +106,24 @@ def grant_reward_points_after_evaluate(request, semester, **_kwargs):
 def grant_reward_points_after_delete(instance, action, reverse, pk_set, **_kwargs):
     # if users do not need to evaluate a course anymore, they may have earned reward points
     if action == 'post_remove':
-        affected = []
+        grantings = []
 
         if reverse:
             # a course got removed from a participant
             user = instance
 
             for semester in Semester.objects.filter(course__pk__in=pk_set):
-                if grant_reward_points_if_eligible(user, semester):
-                    affected = [user]
+                granting, __ = grant_reward_points_if_eligible(user, semester)
+                if granting:
+                    grantings = [granting]
         else:
             # a participant got removed from a course
             course = instance
 
             for user in UserProfile.objects.filter(pk__in=pk_set):
-                if grant_reward_points_if_eligible(user, course.semester):
-                    affected.append(user)
+                granting, __ = grant_reward_points_if_eligible(user, course.semester)
+                if granting:
+                    grantings.append(granting)
 
-        if affected:
-            RewardPointGranting.granted_by_removal.send(sender=RewardPointGranting, users=affected)
+        if grantings:
+            RewardPointGranting.granted_by_removal.send(sender=RewardPointGranting, grantings=grantings)
--- a/evap/staff/tests/test_views.py
+++ b/evap/staff/tests/test_views.py
@@ -6,6 +6,7 @@ from django.conf import settings
 from django.contrib.auth.models import Group
 from django.core import mail
 from django.urls import reverse
+from django.test import override_settings
 from model_mommy import mommy
 import xlrd
 
@@ -124,6 +125,11 @@ class TestUserCreateView(ViewTest):
         self.assertEqual(UserProfile.objects.order_by("pk").last().username, "mflkd862xmnbo5")
 
 
+@override_settings(REWARD_POINTS=[
+    (1 / 3, 1),
+    (2 / 3, 2),
+    (3 / 3, 3),
+])
 class TestUserEditView(ViewTest):
     url = "/staff/user/3/edit"
     test_users = ['staff']
@@ -156,7 +162,7 @@ class TestUserEditView(ViewTest):
         student.refresh_from_db()
 
         self.assertIn("Successfully updated user.", page)
-        self.assertIn("The removal of courses has granted the user &quot;{}&quot; reward points for the active semester.".format(student.username), page)
+        self.assertIn("The removal of courses has granted the user &quot;{}&quot; 3 reward points for the active semester.".format(student.username), page)
 
 
 class TestUserMergeSelectionView(ViewTest):
@@ -872,6 +878,11 @@ class TestCourseCreateView(ViewTest):
         self.assertEqual(Course.objects.get().name_de, "lfo9e7bmxp1xi")
 
 
+@override_settings(REWARD_POINTS=[
+    (1 / 3, 1),
+    (2 / 3, 2),
+    (3 / 3, 3),
+])
 class TestCourseEditView(ViewTest):
     url = '/staff/semester/1/course/1/edit'
     test_users = ['staff']
@@ -925,7 +936,7 @@ class TestCourseEditView(ViewTest):
         form['participants'] = [other.pk]
         page = form.submit('operation', value='save').follow()
 
-        self.assertIn("The removal of participants has granted 1 user reward points: &quot;{}&quot;".format(student.username), page)
+        self.assertIn("The removal as participant has granted the user &quot;{}&quot; 3 reward points for the active semester.".format(student.username), page)
 
     def test_remove_participants(self):
         already_evaluated = mommy.make(Course, semester=self.course.semester)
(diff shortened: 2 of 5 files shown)


evap/staff/tests/test_views.py
```diff
@@ -156,7 +156,7 @@ def test_reward_points_granting_message(self):
         student.refresh_from_db()
 
         self.assertIn("Successfully updated user.", page)
-        self.assertIn("The removal of courses has granted the user &quot;{}&quot; reward points for the active semester.".format(student.username), page)
+        self.assertIn("The removal of courses has granted the user &quot;{}&quot; 3 reward points for the active semester.".format(student.username), page)
```
janno42: please either take the number from the settings file or overwrite settings for this test if you test against a specific value.

evap/staff/tests/test_views.py
```diff
@@ -921,7 +921,7 @@ def test_participant_removal_reward_point_granting_message(self):
         form['participants'] = [other.pk]
         page = form.submit('operation', value='save').follow()
 
-        self.assertIn("The removal of participants has granted 1 user reward points: &quot;{}&quot;".format(student.username), page)
+        self.assertIn("The removal of participants has granted 1 user reward points: 3 points to &quot;{}&quot;".format(student.username), page)
```
janno42: see above

evap/staff/tests/test_views.py
```diff
@@ -939,7 +939,27 @@ def test_remove_participants(self):
         form['participants'] = [student.pk]
         page = form.submit('operation', value='save').follow()
 
-        self.assertIn("The removal of participants has granted 5 users reward points: &quot;a&quot;, &quot;b&quot;, &quot;c&quot;, &quot;d&quot;, &quot;e&quot;.", page)
+        self.assertIn("The removal of participants has granted 5 users reward points: 3 points to &quot;a&quot;, 3 points to &quot;b&quot;, 3 points to &quot;c&quot;, 3 points to &quot;d&quot;, 3 points to &quot;e&quot;.", page)
```
janno42: see above

evap/staff/tests/test_views.py
```diff
@@ -939,7 +939,27 @@ def test_remove_participants(self):
...
+            RewardPointGranting.objects.create(user_profile=user, semester=self.course.semester, value=points_granted)
+
+        page = self.app.get(reverse('staff:course_edit', args=[self.course.semester.pk, self.course.pk]), user='staff')
+
+        # remove five participants
+        form = page.forms['course-form']
+        form['participants'] = [student.pk]
+        page = form.submit('operation', value='save').follow()
+
+        self.assertIn("The removal of participants has granted 3 users reward points: 3 points to &quot;a&quot;, 2 points to &quot;b&quot;, 1 point to &quot;c&quot;.", page)
```
janno42: see above

evap/staff/tests/test_views.py
```diff
@@ -939,7 +939,27 @@ def test_remove_participants(self):
...
+
+
+    def test_remove_participants_propotional_reward_points(self):
+        already_evaluated = mommy.make(Course, semester=self.course.semester)
+        SemesterActivation.objects.create(semester=self.course.semester, is_active=True)
+        student = mommy.make(UserProfile, courses_participating_in=[self.course])
+
+        for name, points_granted in [("a", 0), ("b", 1), ("c", 2), ("d", 3)]:
+            user = mommy.make(UserProfile, username=name, email="{}@institution.example.com".format(name),
+            courses_participating_in=[self.course, already_evaluated], courses_voted_for=[already_evaluated])
```
janno42: please further indent this line

evap/staff/tests/test_views.py
```diff
@@ -939,7 +939,27 @@ def test_remove_participants(self):
...
+            RewardPointGranting.objects.create(user_profile=user, semester=self.course.semester, value=points_granted)
+
+        page = self.app.get(reverse('staff:course_edit', args=[self.course.semester.pk, self.course.pk]), user='staff')
+
+        # remove five participants
```
janno42: you're only removing four, right?

evap/staff/views.py
```diff
@@ -1180,7 +1181,9 @@ def user_import(request):
 def user_edit(request, user_id):
     @receiver(RewardPointGranting.granted_by_removal)
     def notify_reward_points(users, **kwargs):
-        messages.info(request, _('The removal of courses has granted the user "{}" reward points for the active semester.'.format(users[0].username)))
+        messages.info(request, ungettext('The removal of courses has granted the user "{name}" {points} reward point for the active semester.',
+            'The removal of courses has granted the user "{name}" {points} reward points for the active semester.', users[0][1])
+            .format(name=users[0][0].username, points=users[0][1]))
```
janno42: maybe we shouldn't call the parameter `users` anymore because it's now more than that.
in this case `grant_reward_points_after_delete` could maybe also just return a tuple instead of a one element list with a tuple. @karyon, what do you think?

evap/staff/tests/test_views.py
```diff
@@ -939,7 +939,27 @@ def test_remove_participants(self):
         form['participants'] = [student.pk]
         page = form.submit('operation', value='save').follow()
 
-        self.assertIn("The removal of participants has granted 5 users reward points: &quot;a&quot;, &quot;b&quot;, &quot;c&quot;, &quot;d&quot;, &quot;e&quot;.", page)
+        self.assertIn("The removal of participants has granted 5 users reward points: 3 points to &quot;a&quot;, 3 points to &quot;b&quot;, 3 points to &quot;c&quot;, 3 points to &quot;d&quot;, 3 points to &quot;e&quot;.", page)
+
+
+    def test_remove_participants_propotional_reward_points(self):
```
janno42: typo `proportional`

evap/staff/views.py
pixunil: The reason `grant_reward_points_after_delete` fires an event with a list of tuples (which can be a namedtuple for improved readability) is that the removal of participants has got the same way through the code, but could have more affected users. We discussed in the earlier PR the possibiity of two events. Btw, I called `users` while rewriting it `affected` but it had already a collision and I don't want to confuse more than necessary.

He3lixxx:
Concerning the commit message of the last commit: We want our tests to be consistent, no matter whether you run them locally or on travis. Did the test pass on your machine without the include statement? I don't quite see how that could happen, but it would be something we should investigate.

pixunil:
Well, I guess that the virtual machine has not synced the changes or I was too dump to save my files - I think it is nothing serious here (Me being frustrated in evening times results in unhelpful commit messages)

evap/staff/views.py
```diff
@@ -592,11 +592,22 @@ def course_edit(request, semester_id, course_id):
...
+            ).format(transaction=transaction)
+            for transaction in granted
+        )
+        messages.info(request,
+            ngettext(
+                "The removal of participants has granted {affected} user reward points: {transactions}.",
+                "The removal of participants has granted {affected} users reward points: {transactions}.",
+                len(granted)
+            ).format(affected=len(granted), transactions=transactions)
+        )
```
karyon: we could keep this really simple and just emit one message per user. i suppose having multiple users here happens rarely and the simple version is good enough, therefore i would prefer that. @janno42?

janno42: one message per user is fine.

janno42:
@pixunil in case you have the time to continue working on this: we would be happy to have this on production better sooner than later ;)

evap/rewards/tools.py
```diff
@@ -106,22 +106,24 @@ def grant_reward_points_after_evaluate(request, semester, **_kwargs):
...
         if reverse:
             # a course got removed from a participant
             user = instance
 
             for semester in Semester.objects.filter(course__pk__in=pk_set):
-                if grant_reward_points_if_eligible(user, semester):
-                    affected = [user]
+                points_granted, __ = grant_reward_points_if_eligible(user, semester)
+                if points_granted:
+                    granted = [{"user": user, "points": points_granted}]
```
karyon: now that's the prime example for a namedtuple :) my suggestion was probably unclear, i meant a single dictionary mapping from users to points, not a list of dicts. on second thought, i think the list of namedtuples would actually be okay, so you can choose, but a list of dictionaries kind of combines the disadvantages of both options :)

sorry for the back and forth.

...what about letting grant_reward_points_if_eligible return the actual granting object (or None, plus that boolean)? i mean, that contains the user and the points, so it's pretty much what we want here.

evap/staff/views.py
```diff
@@ -1187,8 +1191,14 @@ def user_import(request):
 @staff_required
 def user_edit(request, user_id):
     @receiver(RewardPointGranting.granted_by_removal)
-    def notify_reward_points(users, **_kwargs):
-        messages.info(request, _('The removal of courses has granted the user "{}" reward points for the active semester.'.format(users[0].username)))
+    def notify_reward_points(granted, **kwargs):
+        messages.info(request,
```
karyon: `_kwargs`, please, since it's unused (that changed in a PR merged today, maybe it got lost during the merge?).

also, could you `assert len(granted) == 1`? because it's not immediately obvious that this is assumption you're making is always true here.

also, if you call the elements of `granted` `transaction`, then maybe the list should be called `transactions`. although, maybe `granting` and `grantings` would be better, to not introduce any new names unnecessarily? especially if you choose to use the Granting objects instead of the tuples/dictionaries.

evap/staff/tests/test_views.py
```diff
@@ -872,6 +878,11 @@ def test_course_create(self):
         self.assertEqual(Course.objects.get().name_de, "lfo9e7bmxp1xi")
 
 
+@override_settings(REWARD_POINTS=[
+    (1.0 / 3.0, 1),
+    (2.0 / 3.0, 2),
+    (3.0 / 3.0, 3),
```
karyon: just 1 / 3, without the decimals. integer divisions result in floats in python 3. feel free to change that in the settings and in rewards/tests/test_tools.py as well, but you don't have to.

karyon:
i feel a bit bad for holding this back with those minor nits, sorry for that.

evap/staff/views.py
pixunil: Yes, I wasn't careful enough while rebasing stuff

evap/rewards/tools.py
```diff
@@ -79,19 +79,19 @@ def grant_reward_points_if_eligible(user, semester):
     missing_points = target_points - granted_points
 
     if missing_points > 0:
-        RewardPointGranting.objects.create(user_profile=user, semester=semester, value=missing_points)
-        return missing_points, progress >= 1.0
+        granting = RewardPointGranting.objects.create(user_profile=user, semester=semester, value=missing_points)
+        return granting, progress >= 1.0
     return 0, False
```
karyon: now i'd prefer to return None instead of all the zeroes to keep the types consistent.

evap/staff/views.py
```diff
@@ -593,11 +593,15 @@ def course_edit(request, semester_id, course_id):
...
-            "The removal of participants has granted {affected} users reward points: {names}.",
-            len(users)).format(affected=len(users), names=names))
+    def notify_reward_points(grantings, **_kwargs):
+        for granted in grantings:
+            messages.info(request,
+                ngettext(
+                    'The removal as participant has granted the user "{granted.user_profile.username}" {granted.value} reward point for the active semester.',
+                    'The removal as participant has granted the user "{granted.user_profile.username}" {granted.value} reward points for the active semester.',
+                    granted.value
+                ).format(granted=granted)
```
karyon: `granting` instead of granted?

evap/staff/views.py
```diff
@@ -1187,8 +1191,16 @@ def user_import(request):
...
-        messages.info(request, _('The removal of courses has granted the user "{}" reward points for the active semester.'.format(users[0].username)))
+    def notify_reward_points(grantings, **_kwargs):
+        assert len(grantings) == 1
+
+        messages.info(request,
+            ngettext(
+                'The removal of courses has granted the user "{granted.user_profile.username}" {granted.value} reward point for the active semester.',
+                'The removal of courses has granted the user "{granted.user_profile.username}" {granted.value} reward points for the active semester.',
+                grantings[0].value
+            ).format(granted=grantings[0])
```
karyon: same

karyon:
lgtm!

PR #1198, opened 2018-06-04
Give voters a hint when a text answer has no content 

### LINKED ISSUE #1197: Give voters a hint when a text answer has no content
When a voter enters a comment like "-" or "/", we should display some hint that we'd prefer the textbox to be empty instead.

The reason is that these kinds of answers don't add any value and are routinely filtered out during review, which is avoidable work for the staff users.

Suggestions for the hint: `Hey, please leave the textbox empty (or consider elaborating)` or `This comment will be filtered out. You can simply leave the textbox empty.`

felixrindt:
Closes #1197.

### DIFF
--- a/evap/static/css/evap.scss
+++ b/evap/static/css/evap.scss
@@ -696,19 +696,16 @@ select[multiple] {
     min-height: 40px;
 }
 
+.warning-label {
+    color: $evap-dark-yellow;
+}
 .error-label {
     color: $evap-dark-red;
-}
-
-.warning-label-bottom {
-    margin-top: auto;
-    color: $evap-dark-yellow;
 }
 
 .warning-border {
     border: 1px solid $evap-yellow;
 }
-
 
 .card-submit-area {
     background-color: $lighter-gray;
--- a/evap/static/css/evap.scss
+++ b/evap/static/css/evap.scss
@@ -705,6 +705,7 @@ select[multiple] {
 
 .warning-border {
     border: 1px solid $evap-yellow;
+    outline-color: $evap-yellow;
 }
 
 .card-submit-area {
--- a/evap/static/css/evap.scss
+++ b/evap/static/css/evap.scss
@@ -700,6 +700,16 @@ select[multiple] {
     color: $evap-dark-red;
 }
 
+.warning-label-bottom {
+    margin-top: auto;
+    color: $evap-dark-yellow;
+}
+
+.warning-border {
+    border: 1px solid $evap-yellow;
+}
+
+
 .card-submit-area {
     background-color: $lighter-gray;
     border-color: rgba($evap-dark-blue, 0.5);;
--- a/evap/student/templates/student_vote.html
+++ b/evap/student/templates/student_vote.html
@@ -175,27 +175,28 @@
             // (the data gets deleted every time the form is submitted, i.e. also when the form had errors and is displayed again)
             sisyphus.saveAllData();
 
-            // Show a warning on textareas if they only contain obsolete text (punctuation, k.A., n.a., ...)
+            // Show a warning on textareas if they only contain obsolete text (punctuation, k.A., n/a, ...)
             function isTextareaInvalid(textarea) {
                 let val = textarea.val();
                 return /^\W+$/g.test(val) || ["ka", "na"].indexOf(val.toLowerCase().replace(/\W/g,'')) != -1;
             }
-            function textareaWarning(textarea) {
+            function updateTextareaWarning(textarea) {
                 let showWarning = isTextareaInvalid(textarea);
-                textarea.parents('tr').find('.warning-label-bottom').toggleClass('d-none', !showWarning);
+                textarea.parents('tr').find('.warning-label').toggleClass('d-none', !showWarning);
                 textarea.toggleClass('warning-border', showWarning);
             }
-            $('textarea').on('input', function() { // Run the check on input, not if user might be starting a list
+            $('textarea').on('input', function() { // Run the check on input
                 let textarea = $(this);
+                let isDelayed = !textarea.hasClass('warning-border') && textarea.val().length < 3 && isTextareaInvalid(textarea)
                 setTimeout(function() {
-                    textareaWarning(textarea);
-                }, !$(this).hasClass('warning-border') && isTextareaInvalid(textarea) ? 1000:1);
+                    updateTextareaWarning(textarea);
+                }, isDelayed ? 1000:0);
             });
             $('textarea').blur(function() { // Run the check when focus is lost
-                textareaWarning($(this));
+                updateTextareaWarning($(this));
             });
             $('textarea').each(function() { // Run the check after page load
-                textareaWarning($(this));
+                updateTextareaWarning($(this));
             });
 
             var form = $('#student-vote-form');
--- a/evap/student/templates/student_vote.html
+++ b/evap/student/templates/student_vote.html
@@ -184,12 +184,11 @@
                 textarea.parents('tr').find('.warning-label').toggleClass('d-none', !showWarning);
                 textarea.toggleClass('warning-border', showWarning);
             }
+            var warningdelayTimer;
             $('textarea').on('input', function() { // Run the check on input
                 let textarea = $(this);
-                if(textarea.hasClass('warning-border') || textarea.val().length >= 3 || !isTextareaInvalid(textarea))
-                    updateTextareaWarning(textarea);
-                else
-                    setTimeout(() => updateTextareaWarning(textarea), 1000);
+                clearTimeout(warningdelayTimer);
+                warningdelayTimer = setTimeout(() => updateTextareaWarning(textarea), 300);
             });
             $('textarea').blur(function() { // Run the check when focus is lost
                 updateTextareaWarning($(this));
--- a/evap/student/templates/student_vote.html
+++ b/evap/student/templates/student_vote.html
@@ -186,9 +186,9 @@
             }
             var warningdelayTimer;
             $('textarea').on('input', function() { // Run the check on input
-                let textarea = $(this);
+                //let textarea = $(this);
                 clearTimeout(warningdelayTimer);
-                warningdelayTimer = setTimeout(() => updateTextareaWarning(textarea), 300);
+                warningdelayTimer = setTimeout(() => updateTextareaWarning($(this)), 300);
             });
             $('textarea').blur(function() { // Run the check when focus is lost
                 updateTextareaWarning($(this));
--- a/evap/student/templates/student_vote.html
+++ b/evap/student/templates/student_vote.html
@@ -175,6 +175,29 @@
             // (the data gets deleted every time the form is submitted, i.e. also when the form had errors and is displayed again)
             sisyphus.saveAllData();
 
+            // Show a warning on textareas if they only contain obsolete text (punctuation, k.A., n.a., ...)
+            function isTextareaInvalid(textarea) {
+                let val = textarea.val();
+                return /^\W+$/g.test(val) || ["ka", "na"].indexOf(val.toLowerCase().replace(/\W/g,'')) != -1;
+            }
+            function textareaWarning(textarea) {
+                let showWarning = isTextareaInvalid(textarea);
+                textarea.parents('tr').find('.warning-label-bottom').toggleClass('d-none', !showWarning);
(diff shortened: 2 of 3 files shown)


He3lixxx:
You mean #1197, not #1159, right? Also, this fixes #1197, right? `Refs` is not one of the [github keywords ](https://help.github.com/articles/closing-issues-using-keywords/) afaik.

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -20,10 +20,15 @@
...
                         {% if field|is_choice_field %}
                             <td class="vote-question vote-question-choice my-auto{% if field.errors %} choice-error{% endif %}" name="{{ field.name }}">
+                            {{ field.label }}
                         {% else %}
-                            <td class="vote-question vote-question-text">
+                            <td class="vote-question vote-question-text d-flex flex-column">
+                            {{ field.label }}
+                            <div class="warning-label mt-auto d-none">
+                                <span class="fas fa-exclamation-triangle"></span>
+                                {% trans "This answer will be filtered out. Please leave the textbox empty if you don't have any feedback for this particular question." %}
```
He3lixxx: This message does not clearly state that a human will process the answer. It could also mean that it's automatically filtered. Do we want this?

felixrindt: What about "This answer seems obsolete. Please leave [...]"? I was using what @janno42 suggested at #1197.

janno42: "This answer seems obsolete" sounds good. I would also not put the human processing into focus here.

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -26,7 +26,7 @@
                             {{ field.label }}
                             <div class="warning-label mt-auto d-none">
                                 <span class="fas fa-exclamation-triangle"></span>
-                                {% trans "This answer will be filtered out. Please leave the textbox empty if you don't have any feedback for this particular question." %}
+                                {% trans "Your answer seems meaningless. Please leave the textbox empty if you don't have any feedback for this question." %}
```
felixrindt: What about this? I'd also prefer to not mention the filtering.

janno42: maybe "unhelpful"?
we could even keep only the second sentence. it still explains what the user should do and is shorter.

felixrindt: A/B testing! :rofl: ... I'd like to hear more opinions...

evap/student/templates/student_vote_questionnaire_group.html
karyon: obsolete means outdated and i didn't find any other meanings that would match...?

personally i'm fine with the current solution in this PR, though i would prefer my shorter proposal `This comment will be filtered out. You can simply leave the textbox empty.` I think that's nice than "Please leave this empty" which imo is a little strong, and without the human work aspect it's not clear why we would care that much.

I think the mere appearance of some text that suggests a different behavior will be enough for most users to change behavior (of course not generally but in this particular case, since i'd guess that users have no particular desire to enter something in there).

janno42: Yes, obsolete is the wrong word here. I personally don't like "This comment will be filtered out" because i would say that this sounds too strong. So what about "You can simply leave the textbox empty if you don't have any feedback for this question."?

karyon: sorry, i realized i commented on an outdated diff and deleted the comment to re-post it below...

evap/student/templates/student_vote_questionnaire_group.html
janno42: then i'd say we should at least make it "will probably be filtered out" - that's something i could live with.

karyon: yeah, i like the "probably". i'd like to have a tooltip on it that says "using advanced tool-assisted filtering procedures".

evap/student/templates/student_vote.html
```diff
@@ -132,6 +132,30 @@ <h3 class="mb-3">{{ course.name }} ({{ course.semester.name }})</h3>
             // (the data gets deleted every time the form is submitted, i.e. also when the form had errors and is displayed again)
             sisyphus.saveAllData();
 
+            // Show a warning on textareas if they only contain obsolete text (punctuation, k.A., n/a, ...)
```
karyon: obsolete -> meaningless

evap/student/templates/student_vote.html
```diff
@@ -132,6 +132,30 @@ <h3 class="mb-3">{{ course.name }} ({{ course.semester.name }})</h3>
...
+            function isTextareaInvalid(textarea) {
+                let val = textarea.val();
+                return /^\W+$/g.test(val) || ["ka", "na"].indexOf(val.toLowerCase().replace(/\W/g,'')) != -1;
```
karyon: instead of the first part of the or you can add an empty string to the array, this way there's not a second regex to parse for the reader. then `textarea.val()` can be inlined. also there's a nicer [includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) method, and you could put the `toLowerCase` behind the `replace` because there's no need to lowercase punctuation, can you try that?

evap/student/templates/student_vote.html
```diff
@@ -132,6 +132,30 @@ <h3 class="mb-3">{{ course.name }} ({{ course.semester.name }})</h3>
...
+                let showWarning = isTextareaInvalid(textarea);
+                textarea.parents('tr').find('.warning-label').toggleClass('d-none', !showWarning);
+                textarea.toggleClass('warning-border', showWarning);
+            }
+            $('textarea').on('input', function() { // Run the check on input
+                let textarea = $(this);
+                let isDelayed = !textarea.hasClass('warning-border') && textarea.val().length < 3 && isTextareaInvalid(textarea)
+                setTimeout(function() {
+                    updateTextareaWarning(textarea);
+                }, isDelayed ? 1000:0);
```
karyon: i don't quite get the delay logic. what would make sense to me is to immediately do the update if there is already a warning on the textarea, and delay the update otherwise. what do the length and invalid check here do?

also, i think i would do
```
if (theCondition) {
    updateTextareaWarning(textarea);
else {
    setTimeout(... updateTextareaWarning(textarea) ...);
}
```
which matches my mental model of "if that is the case, do it now, otherwise do it a second" more closely, compared to "do this later, and to be precise if that is the case do it in a second and otherwise do it now".

also, you can do `setTimeout( () => updateTextareaWarning(textarea), 1000);` which looks nicer imo.

evap/student/templates/student_vote.html
```diff
@@ -132,6 +132,30 @@ <h3 class="mb-3">{{ course.name }} ({{ course.semester.name }})</h3>
...
             sisyphus.saveAllData();
 
+            // Show a warning on textareas if they only contain obsolete text (punctuation, k.A., n/a, ...)
+            function isTextareaInvalid(textarea) {
+                let val = textarea.val();
+                return /^\W+$/g.test(val) || ["ka", "na"].indexOf(val.toLowerCase().replace(/\W/g,'')) != -1;
+            }
+            function updateTextareaWarning(textarea) {
+                let showWarning = isTextareaInvalid(textarea);
+                textarea.parents('tr').find('.warning-label').toggleClass('d-none', !showWarning);
```
karyon: i have the feeling `.warning-label` is a weird selector here, because that's a styling thing and not some identification thing, but i don't really know what i'm talking about, @janno42?

evap/student/templates/student_vote.html
karyon: what do you mean by empty val? an empty string? that should just work...

evap/student/templates/student_vote.html
felixrindt: We want to show the warning as early as possible, but not if the user might be just starting or correcting the first entry of a list.

A list is typically started with `- `, so the third character is a letter. If the first 3 letters are punctuation, it's probably not a list and we show the warning.
If the user leaves it at `-`, we show the warning after a delay, because they might be starting a list.
The invalid check is there to prevent that the warning is shown without the right delay when changing the value like the follwing (`|` denoting the users cursor)  `- foo|` --> `- |` --> `- bar|`. The middle state here is invalid and the warning would be almost instantly shown in normal us due to the delayed call being triggered when removing the Os.
As you figured, removing the warning should also be without delay.

evap/student/templates/student_vote.html
karyon: ah, *exclude*, now i understand.

evap/student/templates/student_vote_questionnaire_group.html
He3lixxx: So we're talking about "This answer will probably be filtered out. You can simply leave the textbox empty" now, right? I like that.

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -20,10 +20,15 @@
...
                         {% if field|is_choice_field %}
                             <td class="vote-question vote-question-choice my-auto{% if field.errors %} choice-error{% endif %}" name="{{ field.name }}">
+                            {{ field.label }}
                         {% else %}
-                            <td class="vote-question vote-question-text">
+                            <td class="vote-question vote-question-text d-flex flex-column">
+                            {{ field.label }}
+                            <div class="warning-label mt-auto d-none">
+                                <span class="fas fa-exclamation-triangle"></span>
+                                {% trans "Your answer seems meaningless. Please leave the textbox empty if you don't have any feedback for this question." %}
```
janno42: please change to `This answer will probably be filtered out. You can simply leave the textbox empty.`

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -20,10 +20,15 @@
                     <tr class="vote-row d-flex">
                         {% if field|is_choice_field %}
                             <td class="vote-question vote-question-choice my-auto{% if field.errors %} choice-error{% endif %}" name="{{ field.name }}">
+                            {{ field.label }}
```
janno42: the indentation will always be somewhat strange here because the opening tag is inside the if and the closing tag isn't. but could you please still indent this line and those below the opening `<td>` below one level further?

evap/student/templates/student_vote.html
janno42: that's fine for me.

evap/static/css/evap.scss
```diff
@@ -696,10 +696,17 @@ select[multiple] {
...
+.warning-label {
+    color: $evap-dark-yellow;
+}
 .error-label {
     color: $evap-dark-red;
 }
 
+.warning-border {
+    border: 1px solid $evap-yellow;
+}
```
janno42: please add an additional `outline-color: $evap-yellow;` for chrome

felixrindt: :+1:

evap/student/templates/student_vote_questionnaire_group.html
felixrindt: :+1:

evap/student/templates/student_vote_questionnaire_group.html
felixrindt: :+1:

















⁣

evap/student/templates/student_vote.html
felixrindt: Now using debounce which should solve all of these concerns...

evap/student/templates/student_vote.html
```diff
@@ -132,6 +132,28 @@ <h3 class="mb-3">{{ course.name }} ({{ course.semester.name }})</h3>
...
+                textarea.toggleClass('warning-border', showWarning);
+            }
+            var warningdelayTimer;
+            $('textarea').on('input', function() { // Run the check on input
+                let textarea = $(this);
+                clearTimeout(warningdelayTimer);
+                warningdelayTimer = setTimeout(() => updateTextareaWarning(textarea), 300);
```
karyon: inline `textarea`?

felixrindt: Oh I originally thought that wouldn't work, because it did not with an anonymous function, but it does work with lambda expressions :+1:

PR #1200, opened 2018-06-08
Stuff

karyon:
see commit messages.

In the last commit i wasn't sure that's an improvement. By now i like it better than the previous version, but i'm open to different opinions.


evap/results/tests/test_views.py
```diff
@@ -40,6 +40,7 @@ def setUpTestData(cls):
         student2 = mommy.make(UserProfile)
         cls.course = mommy.make(Course, id=21, state='published', semester=cls.semester, participants=[student1, student2], voters=[student1, student2])
         questionnaire = mommy.make(Questionnaire)
+        cls.contribution = mommy.make(Contribution, course=cls.course, questionnaires=[questionnaire], contributor=None)
```
janno42: as @He3lixxx mentioned in a recent pr: shouldn't every course that has been saved already have a general contribution? can't you add the questionnaire to that?

evap/results/tests/test_views.py
```diff
@@ -225,6 +226,7 @@ def test_private_course(self):
         mommy.make(UserProfile, username="random", email="random@institution.example.com")
         degree = mommy.make(Degree)
         private_course = mommy.make(Course, state='published', is_private=True, semester=semester, participants=[student, student_external, test1, test2], voters=[test1, test2], degrees=[degree])
+        mommy.make(Contribution, course=private_course, contributor=None, questionnaires=[mommy.make(Questionnaire)])
```
janno42: same here

evap/results/tools.py
```diff
@@ -173,27 +181,18 @@ def calculate_average_distribution(course):
...
-        else:
-            average_course_grade_questions_distribution = average_grade_questions_distribution
-            average_course_non_grade_questions_distribution = average_non_grade_questions_distribution
+    course_results = grouped_results.pop(None)
 
-    average_contributor_distribution = avg_distribution(average_contributor_distributions)
+    average_contributor_distribution = avg_distribution([
+        avg_distribution(
+            [average_grade_questions_distribution(results), average_non_grade_rating_questions_distribution(results)],
+            [settings.CONTRIBUTOR_GRADE_QUESTIONS_WEIGHT, settings.CONTRIBUTOR_NON_GRADE_QUESTIONS_WEIGHT]
```
janno42: since you're at it: could you rename this to `CONTRIBUTOR_NON_GRADE_RATING_QUESTIONS_WEIGHT` to be consistent?

evap/results/tests/test_views.py
karyon: yeah, the code is currently plain wrong, since it adds a second contribution with contributor=None to the course.

PR #1203, opened 2018-06-13
New result tuples

### LINKED ISSUE #1136: Simplify results tuples

karyon:
fixes #1136 

definitely review per commit.

- the tuples are now proper classes and are called CourseResult, ContributionResult, QuestionnaireResult, [Heading|Text]Result.
- no more sections
- ̀ calculate_results` is now really trivial, the course_detail view also got a bit simpler
- unfortunately, some parts got more verbose
- it's faster: previously, the code in `calculate_results` that computed the warnings fetched everything from the database a second time, now that's gone. that and some added prefetches made `refresh_results_cache` twice as fast
- the second, last, and `Remove has_no_rating_answers` commits include behavioral fixes

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -421,8 +421,8 @@ class Course(models.Model, metaclass=LocalizeModelBase):
 
     @transition(field=state, source='published', target='reviewed')
     def unpublish(self):
-        from evap.results.tools import get_results_cache_key
-        caches['results'].delete(get_results_cache_key(self))
+        from evap.results.tools import get_collect_results_cache_key
+        caches['results'].delete(get_collect_results_cache_key(self))
 
     @cached_property
     def general_contribution(self):
--- a/evap/results/exporters.py
+++ b/evap/results/exporters.py
@@ -94,7 +94,7 @@ class ExcelExporter(object):
                 if not course.can_publish_rating_results and not include_not_enough_voters:
                     continue
                 results = OrderedDict()
-                for questionnaire_result in calculate_results(course):
+                for questionnaire_result in calculate_results(course).questionnaire_results:
                     if all(not question_result.question.is_rating_question or question_result.counts is None for question_result in questionnaire_result.question_results):
                         continue
                     results.setdefault(questionnaire_result.questionnaire.id, []).extend(questionnaire_result.question_results)
--- a/evap/results/exporters.py
+++ b/evap/results/exporters.py
@@ -138,7 +138,7 @@ class ExcelExporter(object):
 
                         for grade_result in qn_results:
                             if grade_result.question.id == question.id:
-                                if grade_result.average is not None:
+                                if grade_result.has_answers:
                                     values.append(grade_result.average * grade_result.total_count)
                                     total_count += grade_result.total_count
                                     if grade_result.question.is_yes_no_question:
@@ -147,7 +147,7 @@ class ExcelExporter(object):
                             avg = sum(values) / total_count
 
                             if question.is_yes_no_question:
-                                percent_approval = float(approval_count) / float(total_count) if total_count > 0 else 0
+                                percent_approval = approval_count / total_count if total_count > 0 else 0
                                 writec(self, "{:.0%}".format(percent_approval), self.grade_to_style(avg))
                             else:
                                 writec(self, avg, self.grade_to_style(avg))
--- a/evap/results/exporters.py
+++ b/evap/results/exporters.py
@@ -5,7 +5,7 @@ from django.utils.translation import ugettext as _
 import xlwt
 
 from evap.evaluation.models import CourseType
-from evap.results.tools import calculate_results, calculate_average_distribution, get_grade_color, distribution_to_grade
+from evap.results.tools import collect_results, calculate_average_distribution, get_grade_color, distribution_to_grade
 
 
 class ExcelExporter(object):
@@ -94,7 +94,7 @@ class ExcelExporter(object):
                 if not course.can_publish_rating_results and not include_not_enough_voters:
                     continue
                 results = OrderedDict()
-                for questionnaire_result in calculate_results(course).questionnaire_results:
+                for questionnaire_result in collect_results(course).questionnaire_results:
                     if all(not question_result.question.is_rating_question or question_result.counts is None for question_result in questionnaire_result.question_results):
                         continue
                     results.setdefault(questionnaire_result.questionnaire.id, []).extend(questionnaire_result.question_results)
--- a/evap/results/exporters.py
+++ b/evap/results/exporters.py
@@ -5,7 +5,7 @@ from django.utils.translation import ugettext as _
 import xlwt
 
 from evap.evaluation.models import CourseType
-from evap.results.tools import calculate_results, calculate_average_distribution, get_grade_color, has_no_rating_answers, distribution_to_grade
+from evap.results.tools import calculate_results, calculate_average_distribution, get_grade_color, distribution_to_grade
 
 
 class ExcelExporter(object):
@@ -95,7 +95,7 @@ class ExcelExporter(object):
                     continue
                 results = OrderedDict()
                 for section in calculate_results(course):
-                    if has_no_rating_answers(course, section.contributor, section.questionnaire):
+                    if all(not result.question.is_rating_question or result.counts is None for result in section.results):
                         continue
                     results.setdefault(section.questionnaire.id, []).extend(section.results)
                     used_questionnaires.add(section.questionnaire)
--- a/evap/results/exporters.py
+++ b/evap/results/exporters.py
@@ -94,11 +94,11 @@ class ExcelExporter(object):
                 if not course.can_publish_rating_results and not include_not_enough_voters:
                     continue
                 results = OrderedDict()
-                for section in calculate_results(course):
-                    if all(not result.question.is_rating_question or result.counts is None for result in section.results):
+                for questionnaire_result in calculate_results(course):
+                    if all(not question_result.question.is_rating_question or question_result.counts is None for question_result in questionnaire_result.question_results):
                         continue
-                    results.setdefault(section.questionnaire.id, []).extend(section.results)
-                    used_questionnaires.add(section.questionnaire)
+                    results.setdefault(questionnaire_result.questionnaire.id, []).extend(questionnaire_result.question_results)
+                    used_questionnaires.add(questionnaire_result.questionnaire)
                 courses_with_results.append((course, results))
 
             courses_with_results.sort(key=lambda cr: (cr[0].type, cr[0].name))
--- a/evap/results/exporters.py
+++ b/evap/results/exporters.py
@@ -94,11 +94,11 @@ class ExcelExporter(object):
                 if not course.can_publish_rating_results and not include_not_enough_voters:
                     continue
                 results = OrderedDict()
-                for questionnaire, contributor, __, data, __ in calculate_results(course):
-                    if has_no_rating_answers(course, contributor, questionnaire):
+                for section in calculate_results(course):
+                    if has_no_rating_answers(course, section.contributor, section.questionnaire):
                         continue
-                    results.setdefault(questionnaire.id, []).extend(data)
-                    used_questionnaires.add(questionnaire)
+                    results.setdefault(section.questionnaire.id, []).extend(section.results)
+                    used_questionnaires.add(section.questionnaire)
                 courses_with_results.append((course, results))
 
             courses_with_results.sort(key=lambda cr: (cr[0].type, cr[0].name))
--- a/evap/results/templates/result_bar.html
+++ b/evap/results/templates/result_bar.html
@@ -6,17 +6,15 @@
         {% if participants_warning %}
             class="participants-warning"
         {% endif %}
-        {% if result.total_count %}
-            data-toggle="tooltip" data-placement="left" title="{% if participants_warning %}{% trans 'Only a few participants answered this question.' %}</br></br>{% endif %}
-            {% for count in result.counts %}
-                {% with answer_name=result.question|get_answer_name:forloop.counter %}
-                    {% if answer_name %}
-                        {{ answer_name }}: {{ count }}/{{ result.total_count }} ({{ count|percentage_one_decimal:result.total_count }})
-                        {% if not forloop.last %}</br>{% endif %}
-                    {% endif %}
-                {% endwith %}
-            {% endfor %}"
-        {% endif %}
+        data-toggle="tooltip" data-placement="left" title="{% if participants_warning %}{% trans 'Only a few participants answered this question.' %}</br></br>{% endif %}
+        {% for count in result.counts %}
+            {% with answer_name=result.question|get_answer_name:forloop.counter %}
+                {% if answer_name %}
+                    {{ answer_name }}: {{ count }}/{{ result.total_count }} ({{ count|percentage_one_decimal:result.total_count }})
+                    {% if not forloop.last %}</br>{% endif %}
+                {% endif %}
+            {% endwith %}
+        {% endfor %}"
     {% else %}
         data-toggle="tooltip" data-placement="left" title="{% trans 'Not enough answers were given.' %}"
     {% endif %}
--- a/evap/results/templates/result_bar.html
+++ b/evap/results/templates/result_bar.html
@@ -21,7 +21,11 @@
 >
 
     <div class="grade-bg-result-bar-count text-center{% if not question_result.has_answers %} grade-bg-disabled{% endif %}">
-        <span class="fas fa-user small"></span> {{ question_result.total_count }}
+        {% if question_result.is_published %}
+            <span class="fas fa-user small"></span> {{ question_result.total_count }}
+        {% else %}
+            <span class="far fa-eye-slash"></span>
+        {% endif %}
     </div>
 
     {% include 'distribution_bar.html' with question_result=question_result distribution=question_result.counts|normalized_distribution average=question_result.average %}
--- a/evap/results/templates/result_bar.html
+++ b/evap/results/templates/result_bar.html
@@ -2,15 +2,15 @@
 {% load evaluation_filters %}
 
 {% spaceless %}
-<div {% if result.has_answers %}
-        {% if result.warning %}
+<div {% if question_result.has_answers %}
+        {% if question_result.warning %}
             class="participants-warning"
         {% endif %}
-        data-toggle="tooltip" data-placement="left" title="{% if result.warning %}{% trans 'Only a few participants answered this question.' %}</br></br>{% endif %}
-        {% for count in result.counts %}
-            {% with answer_name=result.question|get_answer_name:forloop.counter %}
+        data-toggle="tooltip" data-placement="left" title="{% if question_result.warning %}{% trans 'Only a few participants answered this question.' %}</br></br>{% endif %}
+        {% for count in question_result.counts %}
+            {% with answer_name=question_result.question|get_answer_name:forloop.counter %}
                 {% if answer_name %}
-                    {{ answer_name }}: {{ count }}/{{ result.total_count }} ({{ count|percentage_one_decimal:result.total_count }})
+                    {{ answer_name }}: {{ count }}/{{ question_result.total_count }} ({{ count|percentage_one_decimal:question_result.total_count }})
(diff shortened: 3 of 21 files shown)


evap/results/views.py
```diff
@@ -95,26 +94,24 @@ def course_detail(request, semester_id, course_id):
         questionnaire_result.question_results = filtered_question_results
 
     # remove empty questionnaire_results
-    questionnaire_results = [questionnaire_result for questionnaire_result in questionnaire_results if questionnaire_result.question_results]
+    for contribution_result in course_result.contribution_results:
+        contribution_result.questionnaire_results = [q for q in contribution_result.questionnaire_results if q.question_results]
```
He3lixxx: maybe make this a method? That would keep the verbosity while removing the comment and making this method shorter

He3lixxx: I think even more code in this method could be seperate methods in CourseResult:
- filter empty headings
- split into top, contributor and bottom

What do you think?

evap/results/templates/result_bar.html
```diff
@@ -6,17 +6,15 @@
...
-            {% for count in result.counts %}
-                {% with answer_name=result.question|get_answer_name:forloop.counter %}
-                    {% if answer_name %}
-                        {{ answer_name }}: {{ count }}/{{ result.total_count }} ({{ count|percentage_one_decimal:result.total_count }})
-                        {% if not forloop.last %}</br>{% endif %}
-                    {% endif %}
-                {% endwith %}
-            {% endfor %}"
-        {% endif %}
+        data-toggle="tooltip" data-placement="left" title="{% if participants_warning %}{% trans 'Only a few participants answered this question.' %}</br></br>{% endif %}
```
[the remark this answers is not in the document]
karyon: yeah, it's still possible in this commit, although not anymore in one of the later commits. @janno42 needs to tell me what this check was supposed to do.

evap/results/tools.py
```diff
@@ -84,7 +100,7 @@ def calculate_results(course, force_recalculation=False):
 def _calculate_results_impl(course):
     """Calculates the result data for a single course. Returns a list of
     `ResultSection` tuples. Each of those tuples contains the questionnaire, the
-    contributor (or None), a list of (Rating|YesNo|Text|Heading)Result tuples,
+    contributor (or None), a list of (Rating|Text|Heading)Result tuples,
```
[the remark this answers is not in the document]
karyon: yes. please excuse me if i don't bother in this case, this is removed altogether in a later commit.

evap/results/tools.py
```diff
@@ -24,31 +24,47 @@
...
+    def approval_count(self):
+        assert self.question.is_yes_no_question
+        if not self.counts:
+            return None
+        return self.counts[0] if self.question.is_positive_yes_no_question else self.counts[4]
+
+    @property
+    def average(self):
+        assert not self.question.is_yes_no_question
+        if self.counts is None:
```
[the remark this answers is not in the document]
karyon: no, i'll use `is None` everywhere.

evap/results/tools.py
```diff
@@ -24,31 +24,47 @@
...
-TextResult = namedtuple('TextResult', ('question', 'answers'))
 HeadingResult = namedtuple('HeadingResult', ('question'))
 
-
-def avg(iterable):
-    items = [item for item in iterable if item is not None]
-    if not items:
-        return None
-    return sum(items) / len(items)
+class RatingResult:
```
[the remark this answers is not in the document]
karyon: what do you suggest? i'd rather not test whether the method that sums some numbers actually sums those numbers ;)

evap/results/views.py
```diff
@@ -127,6 +131,29 @@ def course_detail(request, semester_id, course_id):
     return render(request, "results_course_detail.html", template_data)
 
 
+def add_warnings(course, result_sections):
+    if not course.can_publish_rating_results:
+        return
+
+    # calculate the median values of how many people answered a questionnaire type (lecturer, tutor, ...)
```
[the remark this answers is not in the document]
karyon: yeah, it's even more misleading now that we actually have questionnaire types. i'll change it to "how many people answered a questionnaire across all contributions".

evap/results/views.py
```diff
@@ -127,6 +131,29 @@ def course_detail(request, semester_id, course_id):
...
+    questionnaire_med_answers = defaultdict(list)
+    for section in result_sections:
+        max_answers = max((result.total_count or 0 for result in section.results if result.question.is_rating_question), default=0)
+        questionnaire_med_answers[section.questionnaire].append(max_answers)
+
+    questionnaire_warning_thresholds = {}
+    for questionnaire, max_answers in questionnaire_med_answers.items():
```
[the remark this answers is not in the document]
karyon: des

evap/results/exporters.py
```diff
@@ -95,7 +95,7 @@ def export(self, response, course_types_list, include_not_enough_voters=False, i
                     continue
                 results = OrderedDict()
                 for section in calculate_results(course):
-                    if has_no_rating_answers(course, section.contributor, section.questionnaire):
+                    if all(not result.question.is_rating_question or result.counts is None for result in section.results):
```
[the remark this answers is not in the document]
karyon: previously, has_no_rating_answers checked whether there were any RatingAnswerCounters, that is, whether any voter gave any answer to any rating question. however, these counters might not be written into result.counts if not course.can_publish_rating_results. i think it's correct to check whether there are any publishable counters, not any counters at all.

evap/results/tools.py
```diff
@@ -19,11 +18,34 @@
...
+
+
+class ContributionResult:
+    def __init__(self, contributor, label, questionnaire_results):
         self.contributor = contributor
         self.label = label
+        self.questionnaire_results = questionnaire_results
+
+    @property
+    def has_votes(self):
```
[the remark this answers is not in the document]
karyon: has_answers?

evap/results/templates/result_bar.html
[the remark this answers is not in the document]
janno42: i think this was introduced to prevent the case @He3lixxx described. if total_count is 0, no tooltip or warning "only a few..." should be displayed at any point. i don't know why we didn't check for if result.counts, though.
anyway - it's probably easier to just remove this check and find out during the review that something should be changed than trying to figure out what exactly led to this specific line of code. looks like we should have added a comment there ;)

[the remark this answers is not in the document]
karyon: this is not about questions having no answers, this case is handled by the surrounding `if result.average is not None`. we already talked about this and you said this might be for single results, but didn't remember in which case they would have zero answers.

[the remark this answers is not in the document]
karyon: upon reviewing this again, i don't see how this check can have any effect. before the changes in this PR, if total_count was 0, then average and counts are both None. this should be handled by the surrounding if. after this PR, total_count, counts and average are either all three None or all three larger than zero. @He3lixxx could you confirm?

evap/results/views.py
karyon: that would work and it would make this method here shorter, but i'm not a big fan of methods that are called in exactly one place. `add_warnings` i put into a method because it was lots of code that didn't do very much. actually i'm thinking about moving the has_votes part back into this method...

I mean, of course we can put every section into a separate method, but what's the point? then everything is in a method, what gives? it gives me warm fuzzy feelings as well but i can't find a rational reasoning why that would help. maybe you can.

in the last fixup commit, i tried to condense the split part, but if anything it just got more horizontal. @He3lixxx could you tell me whether you would prefer that in case we don't use methods for the sections?

evap/results/views.py
```diff
@@ -127,6 +131,29 @@ def course_detail(request, semester_id, course_id):
...
+        questionnaire_warning_thresholds[questionnaire] = max(settings.RESULTS_WARNING_PERCENTAGE * median(max_answers), settings.RESULTS_WARNING_COUNT)
```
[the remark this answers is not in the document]
janno42: though the wiki is outdated in many points (i will do some updates soon), this part explaining the warning thresholds is still correct: https://github.com/fsr-itse/EvaP/wiki/Publishing-Results#calculation-of-the-warning-thresholds
the reasoning is that there can be many, e.g., tutors for a course with one of them being the "main tutor". this contributor might get a number of votes close to the total number of voters while others who are maybe just working with subgroups of the participants are receiving way less votes. we don't want that one tutor to cause warnings for all other tutors. so by choosing the median we select the number of voters that is "normal" for the current course as a basis for the warning threshold.

evap/results/views.py
```diff
@@ -127,6 +131,29 @@ def course_detail(request, semester_id, course_id):
     return render(request, "results_course_detail.html", template_data)
 
 
+def add_warnings(course, result_sections):
+    if not course.can_publish_rating_results:
+        return
+
+    # calculate the median values of how many people answered a questionnaire type (lecturer, tutor, ...)
+    questionnaire_med_answers = defaultdict(list)
```
janno42: why is this called `med_answers`?

evap/results/templates/results_course_detail_questionnaires.html
```diff
@@ -23,9 +21,7 @@ <h4>{{ section.questionnaire.public_name }}</h4>
                 <tr class="no-break d-flex">
                     <td class="result-rating-question my-auto">{{ result.question.text }}</td>
                     <td class="result-rating-answer my-auto">
-                        {% with participants_warning=section.warning|or:result.warning %}
-                            {% include 'result_bar.html' with result=result participants_warning=participants_warning %}
-                        {% endwith %}
+                        {% include 'result_bar.html' with result=result participants_warning=result.warning %}
```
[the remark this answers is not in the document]
janno42: +1

evap/evaluation/models.py
```diff
@@ -421,8 +421,8 @@ def publish(self):
 
     @transition(field=state, source='published', target='reviewed')
     def unpublish(self):
-        from evap.results.tools import get_results_cache_key
-        caches['results'].delete(get_results_cache_key(self))
+        from evap.results.tools import get_collect_results_cache_key
```
janno42: maybe `collected_results`?

evap/results/tools.py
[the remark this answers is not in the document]
janno42: has_answers sounds good

evap/results/views.py
janno42: here you're introducing single letter variable names to save space - is this the only case where we're using that? maybe we should stick to the verbose names although the lines are getting longer. if not - this should be at least `qr`.

evap/results/views.py
```diff
@@ -100,18 +100,12 @@ def course_detail(request, semester_id, course_id):
     add_warnings(course, course_result)
 
     # split course_result into different lists
+    contributor_contribution_results = [contribution_result for contribution_result in course_result.contribution_results if contribution_result.contributor is not None]
+    general_contribution_result = next(contribution_result for contribution_result in course_result.contribution_results if contribution_result.contributor is None)
```
janno42: is this the same as `[questionnaire_result for contribution_result in course_result.contribution_results if contribution_result.contributor is None for questionnaire_result in contribution_result.questionnaire_results]`?

janno42: although much longer i might prefer this one because one can see that we're collecting the questionnaire results.

evap/results/templates/result_bar.html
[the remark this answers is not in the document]
He3lixxx: As I see it, yes, that is correct.

evap/results/tools.py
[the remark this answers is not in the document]
He3lixxx: results/tests/test_results.py:TestCalculateResults.test_calculation_results does some testing on these methods, I think that's sufficient.

evap/results/exporters.py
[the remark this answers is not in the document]
He3lixxx: commit message -> "with no publishable votes" then?

evap/results/views.py
He3lixxx: I looked at it again and I think it doesn't really become better, so I agree with you and would leave it like that.

About the last thing, condensing the split part: I don't know to be honest. I don't think I've ever seen people use an if expression to determine the object they want to call a method on in-place, but I don't see any reason to not do it, so idk.

He3lixxx:
Approved from by side with the one commit message changed

evap/results/views.py
```diff
@@ -114,15 +112,10 @@ def course_detail(request, semester_id, course_id):
...
-            for result in section.results:
-                if isinstance(result, TextResult):
-                    contributor_sections[section.contributor]['total_votes'] += 1
-                elif isinstance(result, RatingResult) or isinstance(result, YesNoResult):
-                    # Only count rating results if we show the grades.
-                    if course.can_publish_rating_results:
-                        contributor_sections[section.contributor]['total_votes'] += result.total_count
+                                            {'has_votes': False, 'sections': []})['sections'].append(section)
+
+            if any(result.question.is_rating_question and result.total_count or result.question.is_text_question for result in section.results):
```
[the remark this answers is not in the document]
karyon: i'm not 100% sure i understand, but rating answers that cannot be published never make it out of `collect_results`, so `has_votes` should be `False` in your case. could you verify that?

evap/evaluation/models.py
karyon: i was trying to say "this is the cache key of the method `collect_results`". not sure whether that makes sense, what do you think?

evap/results/views.py
karyon: please check again, i reverted the commit that made this part more confusing.

evap/results/views.py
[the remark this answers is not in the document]
janno42: i see. now we have the following situation: in case only one user votes for a course, the rating results are not published. staff and contributors can still access the results page and for every question the number of people that voted for it is shown to be 0 even if the one person actually voted for it.
so this changed compared to what we had before where the shown number of voters was always the actual number of voters and only the distribution and average bar didn't show values when the rating results couldn't be published. can we get that back working like that?

evap/evaluation/models.py
janno42: kind of :)
i think naming is weird both ways so don't bother.

evap/results/views.py
janno42: i don't remember exactly but i think that this replaced about 10 other lines of code. so you didn't like two two-line solution or why did you revert the one commit?

karyon: check out the last commit, this is what it would look like. it's not shorter, it's just less lines of code.

janno42: okay, that probably doesn't really make it easier to understand. i'm fine with either option.

evap/results/views.py
```diff
@@ -73,18 +73,13 @@ def course_detail(request, semester_id, course_id):
...
-                answers = [answer for answer in result.answers if user_can_see_text_answer(request.user, represented_users, answer, public_view)]
-                if answers:
-                    results.append(TextResult(question=result.question, answers=answers))
-            else:
-                results.append(result)
-
-        section.results[:] = results
+                result.answers = [answer for answer in result.answers if user_can_see_text_answer(request.user, represented_users, answer, public_view)]
+        # remove empty TextResults
+        section.results[:] = [result for result in section.results if not isinstance(result, TextResult) or len(result.answers) > 0]
```
janno42: why `[:]`?

evap/results/templates/result_bar.html
```diff
@@ -21,7 +21,7 @@
 >
 
     <div class="grade-bg-result-bar-count text-center{% if question_result.counts is None %} grade-bg-disabled{% endif %}">
-        <span class="fas fa-user small"></span> {{ question_result.total_count }}
+        <span class="fas fa-user small"></span> {{ question_result.total_count|default:0 }}
```
janno42: this commit is obsoleted by the next one

evap/results/templates/result_bar.html
```diff
@@ -21,7 +21,11 @@
 >
 
     <div class="grade-bg-result-bar-count text-center{% if question_result.counts is None %} grade-bg-disabled{% endif %}">
-        <span class="fas fa-user small"></span> {{ question_result.total_count|default:0 }}
+        {% if question_result.total_count %}
+            <span class="fas fa-user small"></span> {{ question_result.total_count }}
+        {% else %}
+            strich goes here
```
janno42: please use `&mdash;` for the dash

janno42: actually this will now also show a dash for questions where the number of voters is 0 even if other questions received enough answers to publish the results. i think in this case we should show the 0, don't you think? so use the dash only if `can_publish_rating_results` is `False`.

example:
novoter

evap/results/views.py
karyon: in this commit, `section.results` is still a immutable tuple.

janno42:
i have no further remarks after the one remaining issue is resolved :)

evap/results/views.py
```diff
@@ -139,7 +139,7 @@ def add_warnings(course, course_result):
     # calculate the median values of how many people answered a questionnaire across all contributions
     questionnaire_max_answers = defaultdict(list)
     for questionnaire_result in course_result.questionnaire_results:
-        max_answers = max((question_result.total_count or 0 for question_result in questionnaire_result.question_results if question_result.question.is_rating_question), default=0)
+        max_answers = max((question_result.total_count for question_result in questionnaire_result.question_results if question_result.question.is_rating_question), default=0)
```
janno42: removing the `or 0` leads to errors (e.g. results/semester/21/course/1577 in the test data)

evap/results/views.py
```diff
@@ -148,11 +148,11 @@ def add_warnings(course, course_result):
 
     for questionnaire_result in course_result.questionnaire_results:
         rating_results = [question_result for question_result in questionnaire_result.question_results if question_result.question.is_rating_question]
-        max_answers = max((rating_result.total_count or 0 for rating_result in rating_results), default=0)
+        max_answers = max((rating_result.total_count for rating_result in rating_results), default=0)
```
janno42: same here

evap/results/tools.py
```diff
@@ -52,27 +52,35 @@ def __init__(self, questionnaire, question_results):
...
-        return sum(answer * count for answer, count in enumerate(self.counts, start=1)) / sum(self.counts)
+        return sum(answer * count for answer, count in enumerate(self.counts, start=1)) / self.total_count
+
+    @property
+    def has_answers(self):
+        return self.is_published and any(count != 0 for count in self.counts)
+
+    @property
+    def is_published(self):
+        return self.counts is None
```
janno42: `is not None`

evap/results/templates/result_bar.html
```diff
@@ -21,7 +21,11 @@
 >
 
     <div class="grade-bg-result-bar-count text-center{% if question_result.counts is None %} grade-bg-disabled{% endif %}">
-        <span class="fas fa-user small"></span> {{ question_result.total_count|default:0 }}
+        {% if course.can_publish_rating_results %}
+            <span class="fas fa-user small"></span> {{ question_result.total_count|default:0 }}
+        {% else %}
+            &mdash;
```
janno42: sorry about this, but maybe we could use `<span class="far fa-eye-slash"></span>` here after all.

janno42:
contributors who only have text questions end up having an empty card on the results page in public view, the same goes for the "myself" view if no answers were given for these questions (see, e.g., results/semester/19/course/1625). before, they weren't listed at all.
the entry should be removed - or better: it should remain there but it should not be possible to open the card.

evap/results/views.py
karyon: this shouldn't be the case, as total_count is only None if the rating results cannot be published, and that has been ruled out at the beginning of this method. have you refreshed the results cache? can you give me more info?

karyon:
i filtered out empty contributors for now. i agree that showing them is the better solution, but right now i'm really kind of tired of this PR and want to get it done.

PR #1212, opened 2018-06-19
Check whether text results can be published

### LINKED ISSUE #1211: Text results are never published
In `student/views.py:vote` voters are added by `course.voters.through.objects.get_or_create`. This doesn't call the `m2m_changed` signal in `evaluation/models.py:voters_changed`. As a consequence, `can_publish_text_results` is not set to `True` by adding the second voter to a course as it should be.

Our tests didn't show the issue because we're adding voters by `course.voters.add` there.

janno42:
fix #1211

### DIFF
--- a/evap/evaluation/tests/tools.py
+++ b/evap/evaluation/tests/tools.py
@@ -5,6 +5,7 @@ from model_mommy import mommy
 from webtest import AppError
 
 from evap.evaluation.models import Contribution, Course, UserProfile, Questionnaire, Degree
+from evap.student.tools import question_id
 
 
 def to_querydict(dictionary):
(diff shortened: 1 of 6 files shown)


evap/results/tests/test_views.py
```diff
@@ -203,13 +196,34 @@ def helper_test_answer_visibility(self, username, expect_page_not_visible_first=
...
+        self.helper_test_answer_visibility_one_voter("staff")
+        self.course = Course.objects.get(id=self.course.id)
+        self.helper_test_answer_visibility_one_voter("responsible")
+        self.helper_test_answer_visibility_one_voter("student", expect_page_not_visible=True)
+
+        # add second voter
+        self.course.unpublish()
+        self.course.reopen_evaluation()
+        self.course.save()
+        self.let_user_vote_for_course(self.student2, self.course)
```
karyon: put this into a separate test? the unpublishing and cache clearing shouldn't be necessary then.

evap/evaluation/tests/tools.py
```diff
@@ -53,6 +54,19 @@ def get_submit_assert_200(self, url, user):
         self.assertEqual(response.status_code, 200, 'url "{}" failed with user "{}"'.format(url, user))
         return response
 
+    def let_user_vote_for_course(self, user, course):
+        url = '/student/vote/{}'.format(course.id)
+        page = self.get_assert_200(url, user=user)
+        form = page.forms["student-vote-form"]
+        for contribution in course.contributions.all():
```
karyon: please add a `.prefetch_related("questionnaires", "questionnaires__question_set")` for likely no performance gain but still.

evap/results/tests/test_views.py
janno42: cache clearing is indeed not necessary but we need some kind of reset for the course because "Modifications to in-memory objects from setup work done at the class level will persist between test methods."
so the course would still be in state `published` in the second test.

karyon: yeah, that's why i suggested to put the second part into a separate test :)

karyon: yeah, you probably need a `self.course = Course.objects.get(pk=self.course.pk)`, but you shouldn't need the unpublishing part.

karyon: ...and please put that into a setUp method, otherwise it will be dependent on the execution order.

janno42: okay

PR #1213, opened 2018-06-21
Cleanups, fix reload_testdata

karyon:
See commit messages. Includes fixes to weirdnesses discovered when scrolling through the diff of #1122 by @PFischbeck :)

### DIFF
--- a/evap/contributor/tests/test_forms.py
+++ b/evap/contributor/tests/test_forms.py
@@ -71,17 +71,17 @@ class ContributionFormsetTests(TestCase):
         InlineContributionFormset = inlineformset_factory(Course, Contribution, formset=ContributionFormSet, form=EditorContributionForm, extra=1)
 
         data = {
-            "contributions-TOTAL_FORMS": "1",
-            "contributions-INITIAL_FORMS": "1",
-            "contributions-MIN_NUM_FORMS": "0",
-            "contributions-MAX_NUM_FORMS": "1000",
-            "contributions-0-course": "{}".format(course.pk),
-            "contributions-0-order": "1",
-            "contributions-0-id": "{}".format(contribution.pk),
-            "contributions-0-contributor": "{}".format(user.pk),
+            "contributions-TOTAL_FORMS": 1,
+            "contributions-INITIAL_FORMS": 1,
+            "contributions-MIN_NUM_FORMS": 0,
+            "contributions-MAX_NUM_FORMS": 1000,
+            "contributions-0-course": course.pk,
+            "contributions-0-order": 1,
+            "contributions-0-id": contribution.pk,
+            "contributions-0-contributor": user.pk,
             "contributions-0-does_not_contribute": "on",
-            "contributions-0-responsibility": "EDITOR",
-            "contributions-0-comment_visibility": "OWN",
+            "contributions-0-responsibility": Contribution.IS_EDITOR,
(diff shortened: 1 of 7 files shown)


evap/student/views.py
```diff
@@ -111,31 +111,26 @@ def vote(request, course_id):
     # all forms are valid, begin vote operation
     with transaction.atomic():
         # add user to course.voters
-        # not using course.voters.add(request.user) since it fails silently when done twice.
-        # manually inserting like this gives us the 'created' return value and ensures at the database level that nobody votes twice.
-        __, created = course.voters.through.objects.get_or_create(userprofile_id=request.user.pk, course_id=course.pk)
-        if not created:  # vote already got recorded, bail out
-            raise SuspiciousOperation("A second vote has been received shortly after the first one.")
+        # not using course.voters.add(request.user) since that fails silently when done twice.
+        course.voters.through.objects.create(userprofile_id=request.user.pk, course_id=course.pk)
```
janno42: why don't we have to check for a second vote anymore?

evap/contributor/tests/test_forms.py
```diff
@@ -147,7 +147,7 @@ def test_editors_cannot_delete_responsibles(self):
             "contributions-0-id": "{}".format(contribution.pk),
             "contributions-0-contributor": "{}".format(user.pk),
             "contributions-0-does_not_contribute": "on",
-            "contributions-0-responsibility": "RESPONSBILE",
+            "contributions-0-responsibility": "RESPONSIBLE",
```
janno42: the changes in the second commit are completely obsoleted by the fourth commit as far as i can see. could you remove it?

evap/student/views.py
karyon: because create throws an exception when such an object already exists. using get_or_create and then manually throwing seemed unnecessary to me.

evap/contributor/tests/test_forms.py
karyon: well, they do different things. the commit message would be something like "lots of cleanups and somewhere in there are also behavioral fixes", which to me would look like that should be two separate commits. what do you think?

evap/student/views.py
janno42: 👍

evap/contributor/tests/test_forms.py
karyon: yeah, fixing stuff that's going to be removed doesn't make much sense. but, here it got refactored, not removed. if you are moving a method somewhere else, i would totally expect you to not include some subtle behavioral changes in that move just because "it got removed from the original place, so no need to fix it there".

I prefer having behavioral changes and refactorings separated. in case something breaks, and bisecting shows it broke during a commit that says "rename foo to bar" and touches a few hundred lines, and for some reason there is some behavioral change in there because the rename would have made a separate behavior-change-commit obsolete, that would suck. likewise, if during review some commit says "rename foo to bar" i would like to trust the author that the commit does exactly that, i don't want to actually closely inspect every line because the author might have put behavioral changes in there as well.

janno42: i don't see it as a behavioral change in this case because you didn't rename anything, but fixed a typo that can't happen after the refactoring anymore. i see your point though and i could imagine a situation where we want to revert some refactoring which of course shouldn't add wrong code.
maybe we should then pay more attention to such strict separations in the future? because it doesn't feel like we thoroughly did that so far.

i still don't really like the second commit but guess it's okay to do it like that.

Issue #1092, opened 2018-01-30
Linting rules bikeshedding


karyon:
**THIS IS OBSOLETE, check the latest comment**

right now, we only have test coverage via coveralls.

some sort of static analysis would be nice. landscape.io was nice but very unreliable.

options are:
* use https://www.codacy.com, they support prospector: https://support.codacy.com/hc/en-us/articles/213632009-Engines
* use any other service. i checked out codeclimate, but it seems they don't even support pylint. however, if we find a service that supports a set of analysis tools that seems "good enough" we can use that.
* integrate e.g. prospector/any set of tools into our tests. this would kind of require a zero-tolerance-policy since the test can only give us a failed/succeeded, and that might be annoying. maybe @janno42 can tell us how this works out in 1327.
* ???

Once we have anything like that set up, i would suggest a pass through the reports to fix the easy stuff, and maybe update our existing .pylintrc (generate a new one, transplant the changes applied to ours) and review the .landscape.yaml.

and maybe, while at it, rename .landscape.yaml to .prospector.yml.

janno42:
in contrast to evap we wanted to enforce the linter checks in 1327, so failing tests because of linter checks (or missing migrations) are no problem there. we can discuss whether we change that for evap as well.
the repeatedly discussed downside is that new contributors have to take care of this additional thing - but in the end if we want to have nice code that might still be the way to go. i would say linter checks are common in larger projects and the additional hurdle is worth it.

but in the end i'll agree with every solution you'll propose here in this case because i don't have any strong preferences.

He3lixxx:
I'm finde with a zero-tolerance policy for linter errors.

janno42:
- migrations: ✔️
- protected members: ✔️
- assertions: ✔️

janno42:
- [ ] discuss indentation
- [ ] alphabetical imports

He3lixxx:
We ran into errors with codacy evaluating the import order because it does not set up the requirements for static code analysis, thus it can not detect third party modules - see https://github.com/fsr-de/EvaP/pull/1403#discussion_r364379207

I think this is even more a reason to run the linter on travis (where requirements are installed) and fail if anything goes wrong. We could then disable pylint on codacy and just look at the other issues they find.

I think alphabetical imports would be a good rule to have. I don't care about the indentation rules.

He3lixxx:
Black as a code formatter was introduced in #1600
isort was introduced in #1629

[...]

PR #1214, opened 2018-06-23
Lint fixes and linter config tweaks

karyon:
Boring stuff, mostly. review per commit.

After this we can have a look at Codacy and decide whether the current state is enough to consider #1092 fixed.


.pylintrc
```diff
@@ -218,7 +219,7 @@ class-attribute-naming-style=any
 
 # Regular expression matching correct class attribute names. Overrides class-
 # attribute-naming-style
-class-attribute-rgx=[a-z_][a-z0-9_]{1,30}$
+#class-attribute-rgx=
```
janno42: i don't understand why this commit changes the settings in contrast to the one before?

evap/evaluation/migrations/0010_fill_textanswer_state.py
```diff
@@ -28,6 +30,7 @@ def backward_update_states(apps, schema_editor):
             answer.checked = False
         answer.save()
 
+
 class Migration(migrations.Migration):
```
janno42: could you also make these empty lines at the beginning of a class consistent? i'd prefer having none there.

.pylintrc
```diff
@@ -254,7 +254,6 @@ good-names=i,
            _,
            __,
            e,
-           app_name,
```
janno42: why was this introduced a few commits before and is removed here?

evap/evaluation/models.py
```diff
@@ -583,7 +583,7 @@ def update_courses(cls):
 
 
 @receiver(post_transition, sender=Course)
-def log_state_transition(sender, instance, name, source, target, **kwargs):
+def log_state_transition(instance, name, source, target, **_kwargs):
```
janno42: using `**_kwargs` makes copying methods from other places in the code or from the documentation a little bit more work. i think i'd prefer listing `kwargs` in the `dummy-variables-rgx` and keeping the name the same everywhere regardless whether it's actually used in the method or not.

.pylintrc
```diff
@@ -135,7 +135,6 @@ disable=print-statement,
         exception-escape,
         comprehension-escape,
         line-too-long,               # we don't enforce some fixed line length but rather complain about long lines in reviews.
-        cyclic-import,               # landscape re-reports them every time, https://github.com/PyCQA/pylint/issues/59
```
janno42: this is another thing i don't really understand. we're not using landscape anymore, you introduced this a few commits before and remove it here.
so do we want to check for cyclic imports? could you then not add it in the other commit or did i miss something?

.pylintrc
```diff
@@ -278,7 +278,8 @@ module-naming-style=snake_case
 
 # Regular expression matching correct module names. Overrides module-naming-
 # style
-#module-rgx=
+# allowing exactly four digits at the beginning for migrations
+module-rgx=([0-9]{4})?([a-z_][a-z0-9_]*)$
```
janno42: sorry that i'm bringing this up again: couldn't this just be put in the commit "Apply our changes to .pylintrc"? i mean there are no other changes related to this that would make a separate commit necessary.

evap/evaluation/migrations/0055_reviewer_group.py
```diff
@@ -2,11 +2,11 @@
 from django.db import migrations
 
 
-def add_group(apps, schema_editor):
+def add_group(_apps, _schema_editor):
```
janno42: as this is not only about `schema_editor` could you please rename the commit to `Mark unused arguments in migrations`?

karyon:
the `new default .pylintrc` and `apply our changes` were supposed to just update the `.pylintrc` with the new defaults without any further changes, and in the following commits additional changes come in. i can make the second commit `Apply a completely new set of rules` instead of `apply [previous] changes`, i don't really care.

evap/evaluation/models.py
karyon: by "a little bit more work" you mean adding an underscore? :)

i think there should be good reasons for ignoring parameters, and i don't see why that shouldn't apply to keyword arguments. additionally, in our codebase it's really just the receivers that are not using kwargs.

.pylintrc
```diff
@@ -127,13 +127,23 @@ disable=print-statement,
...
         comprehension-escape,
+        line-too-long,               # we don't enforce some fixed line length but rather complain about long lines in reviews.
+        cyclic-import,               # landscape re-reports them every time, https://github.com/PyCQA/pylint/issues/59
+        too-many-public-methods,     # reported for some models, that won't change
+        too-few-public-methods,      # noisy, e.g. exception classes, mixins etc
+        no-member, no-name-in-module, # false positives, deals badly with django classes
+        pointless-string-statement,  # used for documentation in some test files
+        logging-format-interpolation, # decided against the proposed change in favor of .format()
+        too-many-ancestors,          # noisy, reported for all test classes
+        bad-continuation,            # Over a hundred reports, we seem to indent differently, but at least consistently.
```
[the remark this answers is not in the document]
karyon: this commit applies the rules we had before. we can totally discuss this, and if we decide to follow whatever pylint wants, we can remove this exception and apply that style everywhere, but i really wouldn't do that in this PR. maybe you can open a new "style bikeshedding" issue? or put this into the static analysis one?

evap/evaluation/migrations/0010_fill_textanswer_state.py
karyon: i don't see us enforcing that during review, or would you do that? and i cannot find a lint that would enforce this. then i would rather not care about this altogether...

janno42: i just thought since you're at it. sure, we definitely can't enforce that during review when there is no lint checking it because we will overlook it eventually. i mean in most cases i would notice it if someone adds it because it annoys me, but that's not guaranteed. okay then, look like we keep it.

evap/evaluation/models.py
janno42: especially i mean thinking about anything in addition to the actual code. it is no problem at all for you and me, but a small one for less experienced contributors. the linter won't explain them what to do here and they might wonder why something they read in the docs doesn't fulfill the requirements. so it will always be a comment in the reviews explaining them what to do.
but as it makes the code more correct i'm on your side that we shouldn't handle them differently, so i'm okay with keeping the proposed solution.

karyon:
sorted imports are already part of the lints. i have not, in fact, cleaned up everything, just some of the ones that were reported in large numbers, there are now about 300 lints reported instead of 600.

PR #1215, opened 2018-06-23
include number of reward points in participation export

### LINKED ISSUE #1209: Participation export should include number of reward points
Instead of stating `True` or `False` whether a user received reward points in the `semester_participation_export`, the respective column should instead show the number of reward points the user received.

michaelgruenewald:
fixes #1209

### DIFF
--- a/evap/staff/views.py
+++ b/evap/staff/views.py
@@ -442,7 +442,7 @@ def semester_participation_export(request, semester_id):
         number_of_required_courses_voted_for = semester.course_set.filter(voters=participant, is_rewarded=True).count()
         number_of_optional_courses = semester.course_set.filter(participants=participant, is_rewarded=False).count()
         number_of_optional_courses_voted_for = semester.course_set.filter(voters=participant, is_rewarded=False).count()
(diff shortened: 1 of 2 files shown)


evap/staff/views.py
```diff
@@ -442,7 +442,7 @@ def semester_participation_export(request, semester_id):
         number_of_required_courses_voted_for = semester.course_set.filter(voters=participant, is_rewarded=True).count()
         number_of_optional_courses = semester.course_set.filter(participants=participant, is_rewarded=False).count()
         number_of_optional_courses_voted_for = semester.course_set.filter(voters=participant, is_rewarded=False).count()
-        earned_reward_points = RewardPointGranting.objects.filter(semester=semester, user_profile=participant).exists()
+        earned_reward_points = RewardPointGranting.objects.filter(semester=semester, user_profile=participant).aggregate(Sum('value'))['value__sum']
```
janno42: this is currently not consistent with the other columns which show a `0` if there are no courses. can you please add an `or 0` so that also in this column every line has an entry?

karyon:
thanks!

PR #1225, opened 2018-07-02
navbar cache uses username and language as key. 

### LINKED ISSUE #1208: Navbar cache refresh is broken
Since e0beea96a39419422a6c20621f07ba15aa288e1a the navbar cache is based on `session.session_key` but the refresh operation in `delete_navbar_cache_for_users` in `staff/tools.py` tries to delete a key based on `user.username`.

Since this is not working correctly for quite some time now we definitely need to have a test for that as well.

He3lixxx:
Test for delete_navbar_for_user helper method

fixes #1208

### DIFF
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -38,7 +38,7 @@
 
 {% get_current_language as LANGUAGE_CODE %}
 <div class="fixed-top d-print-none">
-    {% cache 3600 navbar request.session.session_key LANGUAGE_CODE %}
+    {% cache 3600 navbar request.user.username LANGUAGE_CODE %}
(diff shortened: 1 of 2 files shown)


evap/evaluation/templates/base.html
```diff
@@ -38,7 +38,7 @@
 
 {% get_current_language as LANGUAGE_CODE %}
 <div class="fixed-top d-print-none">
-    {% cache 3600 navbar request.session.session_key LANGUAGE_CODE %}
+    {% cache 3600 navbar request.user.username LANGUAGE_CODE %}
```
felixrindt: Is this for security reasons or does it actually fix some functionality?

janno42: It fixes a bug, see #1208

PR #1226, opened 2018-07-04
Weight questions by number of answers for results calculation

### LINKED ISSUE #1222: Weigh results of questions by number of answers
Currently, we calculate a grade distribution for each question, and then calculate an average distribution for groups of questions. If one question has 1 vote, and another question has 10 votes, that one vote will have roughly 10 times the weight of the other votes in that average distribution, which doesn't seem right.

janno42:
fix #1222 

this changes how the results are calculated. the number of answers for a question now defines how much this question influences the average grade.
it also leads to contributors with more votes having a stronger influence on the total grade which seems very reasonable.

### DIFF
--- a/evap/results/tools.py
+++ b/evap/results/tools.py
@@ -160,11 +160,17 @@ def avg_distribution(distributions, weights=itertools.repeat(1)):
 
 
 def average_grade_questions_distribution(results):
-    return avg_distribution([normalized_distribution(result.counts) for result in results if result.question.is_grade_question])
+    return avg_distribution(
+        [normalized_distribution(result.counts) for result in results if result.question.is_grade_question],
+        [result.count_sum for result in results if result.question.is_grade_question]
+    )
 
 
 def average_non_grade_rating_questions_distribution(results):
-    return avg_distribution([normalized_distribution(result.counts) for result in results if result.question.is_non_grade_rating_question])
+    return avg_distribution(
+        [normalized_distribution(result.counts) for result in results if result.question.is_non_grade_rating_question],
+        [result.count_sum for result in results if result.question.is_non_grade_rating_question]
+    )
 
 
 def calculate_average_distribution(course):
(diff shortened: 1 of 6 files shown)


evap/results/tools.py
```diff
@@ -179,12 +185,17 @@ def calculate_average_distribution(course):
...
-    ])
+    average_contributor_distribution = avg_distribution(
+        [
+            avg_distribution(
+                [average_grade_questions_distribution(contributor_results), average_non_grade_rating_questions_distribution(contributor_results)],
+                [settings.CONTRIBUTOR_GRADE_QUESTIONS_WEIGHT, settings.CONTRIBUTOR_NON_GRADE_RATING_QUESTIONS_WEIGHT]
+            ) for contributor_results in grouped_results.values()
+        ],
+        [
+            sum(sum(result.counts) for result in contributor_results if result.question.is_rating_question) for contributor_results in grouped_results.values()
```
karyon: this seems to be the wrong order of the loops. please find out why this got through the tests.

evap/results/tools.py
```diff
@@ -160,11 +160,17 @@ def avg_distribution(distributions, weights=itertools.repeat(1)):
 
 
 def average_grade_questions_distribution(results):
-    return avg_distribution([normalized_distribution(result.counts) for result in results if result.question.is_grade_question])
+    return avg_distribution(
+        [normalized_distribution(result.counts) for result in results if result.question.is_grade_question],
+        [sum(result.counts) for result in results if result.question.is_grade_question]
```
karyon: there is `result.total_count`, which is admittedly weird since it does barely more than `return sum(counts)`, but we need it for the templates and i would suggest using it everywhere.

...maybe `sum_of_counts` would be a better name, if you think so as well please rename it.

evap/results/tools.py
```diff
@@ -160,11 +160,17 @@ def avg_distribution(distributions, weights=itertools.repeat(1)):
...
+    )
 
 
 def average_non_grade_rating_questions_distribution(results):
-    return avg_distribution([normalized_distribution(result.counts) for result in results if result.question.is_non_grade_rating_question])
+    return avg_distribution(
+        [normalized_distribution(result.counts) for result in results if result.question.is_non_grade_rating_question],
+        [sum(result.counts) for result in results if result.question.is_non_grade_rating_question]
+    )
```
karyon: it looks like all our calls to `avg_distribution` are now using weights. you can remove the default weights parameter of that function. could you please also consider whether it would make things easier if it took a list of `(distribution, weight)` tuples instead of two separate lists? i'm not sure myself, but it would at least remove the duplicate loops. in the call sites.

karyon:
we thought so much about the weighing and then it was still at this level of wonkiness...

evap/results/tools.py
karyon: i did not properly recognize the parentheses, this is fine as it is.

evap/results/tools.py
janno42: i now called it `count_sum` because we already have `*_sum` at other places

karyon:
well, both solutions aren't exactly pretty... but i think the current solution is fine.

Issue #1109, opened 2018-02-22
Switchable staff mode


janno42:
We could add a staff and/or reviewer mode that can be activated/deactivated individually for these users.
When activated, the platform behaves as currently for staff and reviewer users. When deactivated, they use the page with their respective roles as students and/or contributors.

This improves the following:
- results pages for the current semester can optionally only show regularly published courses for these users
- default view for staff users on results pages could be "own" again, but would still be "public" when the mode is deactivated

karyon:
More details:

* If that mode, working title: sudo mode, is enabled, the page looks and behaves *exactly* as if that user weren't a staff user/reviewer, except for some (toggle) button that enables sudo mode.

* with sudo mode enabled, the default view of result pages would be the public view. (with sudo mode disabled, there is simply no option to change views)

* the results semester view/ overview page needs a "public view" switch as well, but that's a separate issue and probably part of #571 

* sudo mode is off by default. since staff users/reviewers might just enable sudo mode and have little reason to ever disable it again, it should expire e.g. an hour after the last page load.

* implementation-wise, one possibility would be to let `is_staff` and `is_reviewer` return false if sudo mode is off. the (toggle) button would then need a new function e.g. `can_enter_sudo_mode`

* to store the activation somewhere, options would be to link that to the session (especially when we renew it on every request, see #1101), add a timestamp to the user model, or (ab)use the caching framework and put a bool in there with a timeout of one hour.

janno42:
i don't like the name "sudo mode". superusers as defined in django have *all* permissions, so reviewers are definitely no superusers. i also wouldn't expect sudo stuff to change what are default pages for the results.

i'd say we should do the following:
- reviewers stay `reviewers`.
- current staff gets renamed to `managers`. managers also exist in the django terminology, but we aren't using them and even if we would, they would be exactly these people. we split off reviewers from the staff group, although one could argue that they are still working on the platform, so both groups are still "staff" in some way. "managers" better differentiates that.
- the extended mode can then be called `staff mode`.

i think this better represents what we're doing with it: when deactivated the users are regular students and once they activate the staff mode they are acting as staff members who are doing their work tasks on the platform.

janno42:
We could also make the mode switch go as far as possible by letting `is_student` and `is_contributor` return false when activated. This would then remove all voting and "my courses" pages for the staff users and make clear in while role they are currently working on the platform. The mode could also be visualized with a different colored navbar or similar.
[...]

PR #1242, opened 2018-07-30
Rename staff to manager

PFischbeck:
This changes `staff` to `manager`, which hopefully makes #1109 (staff mode) easier.


evap/rewards/views.py
```diff
@@ -1,127 +1,127 @@
...
+
+
+@manager_required
+def semester_activation(request, semester_id, active):
+    semester = get_object_or_404(Semester, id=semester_id)
+    active = active == 'on'
+
+    SemesterActivation.objects.update_or_create(semester=semester, defaults={'is_active': active})
+
+    return semester_view(request=request, semester_id=semester_id)
```
karyon: line endings got screwed up

evap/evaluation/auth.py
```diff
@@ -136,6 +136,15 @@ def check_user(user):
     return user_passes_test(check_user)(view_func)
```
karyon: is `staff_required` used anywhere after applying this PR?

evap/evaluation/migrations/0079_rename_staff_to_manager.py
```diff
@@ -0,0 +1,38 @@
+# Generated by Django 2.0.7 on 2018-07-30 19:25
+
+from django.db import migrations, models
+
+
+def rename_group(apps, schema_editor, old_name, new_name):
+    Group = apps.get_model('auth', 'group')
+    db_alias = schema_editor.connection.alias
+    Group.objects.using(db_alias).filter(name=old_name).update(name=new_name)
```
karyon: why is this `using` part necessary?

evap/rewards/views.py
PFischbeck: Fixed

evap/evaluation/auth.py
PFischbeck: This is not used yet, so I removed it.

evap/evaluation/migrations/0079_rename_staff_to_manager.py
PFischbeck: This is the way it is suggested in the official documentation
https://docs.djangoproject.com/en/2.0/ref/migration-operations/#runpython

PFischbeck:
@karyon Ready for another review. The Codacy checks are false positives, as far as I can tell.

karyon:
i can take over any merge conflicts in other PRs.

PR #1250, opened 2018-08-13
Fix participant count on results page when participants were deleted

### LINKED ISSUE #1248: Participant count incorrect on results index
E.g. for "Lehrerworkshop bei der Endrunde des Bundeswettbewerb Informatik (Sommersemester 2017)" the overview shows "0/0", or for MINT-Camps it says "1/1" or "3/3" participants - the correct numbers are 12, 24 and 22.

capture

karyon:
fixes #1248

### DIFF
--- a/evap/results/tests/test_views.py
+++ b/evap/results/tests/test_views.py
@@ -21,7 +21,14 @@ class TestResultsView(ViewTest):
 class TestGetCoursesWithPrefetchedData(TestCase):
     def test_returns_correct_participant_count(self):
         """ Regression test for #1248 """
-        course = mommy.make(Course, state='published', _participant_count=2, _voter_count=2)
+        participants = mommy.make(UserProfile, _quantity=2)
+        course = mommy.make(Course,
+            state='published', _participant_count=2, _voter_count=2,
+            participants=participants, voters=participants
+        )
+        participants[0].delete()
+        course = Course.objects.get(pk=course.pk)
+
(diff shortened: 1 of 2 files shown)


evap/results/tests/test_views.py
```diff
@@ -16,6 +18,18 @@ def setUpTestData(cls):
...
+    def test_returns_correct_participant_count(self):
+        """ Regression test for #1248 """
+        course = mommy.make(Course, state='published', _participant_count=2, _voter_count=2)
+        courses = get_courses_with_prefetched_data([course])
+        self.assertEqual(courses[0].num_participants, 2)
+        self.assertEqual(courses[0].num_voters, 2)
+        courses = get_courses_with_prefetched_data(Course.objects.filter(pk=course.pk))
+        self.assertEqual(courses[0].num_participants, 2)
+        self.assertEqual(courses[0].num_voters, 2)
+
```
janno42: how do you feel about adding another test that is deleting a user who voted for a course and checking that the numbers are still correct?

karyon: if i understand correctly, that would produce the exactly the situation this test sets up: `_participation_count` and `_voter_count` are set, and the actual number of participant and voters differs from that (it's zero). or wouldn't it?

janno42: yes, but on a more basic level without assuming what we're doing with the counters when that happens. it's not necessary for me, it just felt like a more thorough test.

PR #1252, opened 2018-08-19
Ui changes

janno42:
see commit messages for changes


karyon:
regarding the last commit, isn't that a bootstrap bug?

janno42:
yes, one could classify it as such. the [bootstrap accordion example](http://getbootstrap.com/docs/4.1/components/collapse/#accordion-example) has the same problem. i have put it on my list of issues to report.

PR #1254, opened 2018-08-29
Collapsible improvements

janno42:
this adds additional `div`s around collapsible content to make the animation smooth.
additionally, the toggling chevron is now also animated.


karyon:
that's *a lot* smoother indeed

Issue #1243, opened 2018-08-03
Don't sort by name_de by default


karyon:
Several models have `name_de` as their default `ordering`. That's almost always the wrong choice, since either we want to sort by `name` (this is the localized name, either `name_de` or `name_en`, and must be sorted in the views), or we want to sort by `pk` or `created_at` or something similar.

We should probably remove that. Maybe do some spot tests (search for e.g. `Course.objects`, `.courses` and `.course_set`) whether there are any places that actually need sorting, but don't explicitly specify any.

PR #1255, opened 2018-08-29
Sortable course types

janno42:
enables explicitly sorting course types. they were sorted by their german name before.


karyon:
why would staff users want to manually sort course types? wouldn't sorting by name (of course respecting the current locale) or by usage (number of courses that have that type) be sufficient?

karyon:
ref #1243

janno42:
I want to sort them by importance. Lectures and seminars come first, then some projects, "Other" courses are last. Reason is the filter list on the results page which should be ordered in this way.

karyon:
what about automatically sorting them by the number of courses that have that type? that does not require any user interaction and is less arbitrary than the subjective opinion of a staff member.

janno42:
i would argue it's more or less the same as with degrees. there is kind of a "natural" order of the course types that can be set manually. E.g., "Soft skills colloquium" is used quite often but still less important than every other "regular" course type.

evap/staff/forms.py
```diff
@@ -102,6 +102,11 @@ def clean(self):
 
 
 class CourseTypeForm(forms.ModelForm):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.fields["order"].widget = forms.HiddenInput()
```
karyon: i see we use these `HiddenInput`s a lot, but is there a reason why we can't just exclude the field from `fields`?

janno42: the field must be included in the form to post the new value after reordering and saving the form, doesn't it?

karyon: oh, i didn't realize that this is the form used on the course type index page.

PR #1256, opened 2018-08-29
Ui improvements

janno42:
A couple of random UI improvements. See commit messages for details.

### DIFF
--- a/evap/contributor/views.py
+++ b/evap/contributor/views.py
@@ -8,7 +8,7 @@ from django.db import IntegrityError, transaction
 from evap.contributor.forms import CourseForm, DelegatesForm, EditorContributionForm
 from evap.evaluation.auth import contributor_or_delegate_required, editor_or_delegate_required, editor_required
 from evap.evaluation.models import Contribution, Course, Semester
-from evap.evaluation.tools import STATES_ORDERED, sort_formset
+from evap.evaluation.tools import get_parameter_from_url_or_session, STATES_ORDERED, sort_formset
 from evap.results.tools import calculate_average_distribution, distribution_to_grade
 from evap.staff.forms import ContributionFormSet
 from evap.student.views import get_valid_form_groups_or_render_vote_page
@@ -17,17 +17,21 @@ from evap.student.views import get_valid_form_groups_or_render_vote_page
 @contributor_or_delegate_required
 def index(request):
     user = request.user
+    show_delegated = get_parameter_from_url_or_session(request, "show_delegated", True)
 
     contributor_visible_states = ['prepared', 'editor_approved', 'approved', 'in_evaluation', 'evaluated', 'reviewed', 'published']
     own_courses = Course.objects.filter(contributions__contributor=user, state__in=contributor_visible_states)
 
-    represented_users = user.represented_users.all()
-    delegated_courses = Course.objects.exclude(id__in=own_courses).filter(contributions__can_edit=True, contributions__contributor__in=represented_users, state__in=contributor_visible_states)
+    displayed_courses = list(own_courses)
+    if show_delegated:
+        represented_users = user.represented_users.all()
+        delegated_courses = Course.objects.exclude(id__in=own_courses).filter(contributions__can_edit=True, contributions__contributor__in=represented_users, state__in=contributor_visible_states)
+        for course in delegated_courses:
+            course.delegated_course = True
+        displayed_courses += list(delegated_courses)
+    displayed_courses.sort(key=lambda course: list(STATES_ORDERED.keys()).index(course.state))
 
-    all_courses = list(own_courses) + list(delegated_courses)
-    all_courses.sort(key=lambda course: list(STATES_ORDERED.keys()).index(course.state))
-
-    for course in all_courses:
+    for course in displayed_courses:
         course.distribution = calculate_average_distribution(course)
         course.avg_grade = distribution_to_grade(course.distribution)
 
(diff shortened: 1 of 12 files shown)


evap/results/templates/results_index_course.html
```diff
@@ -22,7 +22,7 @@
                 <span data-toggle="tooltip" data-placement="top" class="fas fa-chart-bar icon-yellow" title="{% trans 'Results not yet published' %}"></span>
             {% endif %}
         {% endif %}
-        <span class="course-name">{{ course.name }}</span>
+        <span class="course-name">{{ course.name }}{% if course.is_single_result %} ({{ course.vote_start_datetime|date:"Y-m-d" }}){% endif %}</span>
```
karyon: you should be able to leave out the parameter to `date`, the default is `DATE_FORMAT`, which is exactly what you wrote there.

evap/evaluation/templates/base.html
```diff
@@ -97,13 +97,17 @@
...
             language: "{{ LANGUAGE_CODE }}",
             placeholder: "{% trans 'Please select...' %}",
-            minimumResultsForSearch: 15,});
+            minimumResultsForSearch: 15,
+            minimumInputLength: minimumInput ? 3 : 0
+        });
     };
-    applySelect2($('select').not('.form-template select').not('[disabled]'));
+    $('select').not('.form-template select').not('[disabled]').each(function() {
+        applySelect2($(this), $(this).children('option').length > 50);
```
karyon: since the parameter is set only here, i would inline that.

evap/evaluation/tools.py
```diff
@@ -134,3 +134,13 @@ def get_due_courses_for_user(user):
...
+
+
+def get_parameter_from_url_or_session(request, parameter):
+    result = request.GET.get(parameter, None)
+    if result is None:  # if no parameter is given take session value
+        result = request.session.get(parameter, False)  # defaults to False if no session value exists
+    else:
+        result = {'true': True, 'false': False}.get(result.lower())  # convert parameter to boolean
+    request.session[parameter] = result  # store value for session
+    return result
```
karyon: you can squash that into the next commit, it's a really trivial refactoring.

karyon: ...disregard that, i missed the changes in the next commit.

janno42:
About hiding delegated courses: Some users (like me) have a huge list of courses on the contributor index page because they are set as delegates. That is perfectly acceptable when preparing the evaluation for all these courses. But when I'm just looking for my own evaluation results, a filtered list is quite nice because I can then get a fast overview and much faster find which courses are really my "Own courses".

We could also introduce a "delegation mode" similar to what we're planning for staff, but that's probably way too much overhead.

evap/results/templates/results_index_course.html
janno42: leaving that out returns `yyyy-MM-dd HH:mm`

evap/evaluation/templates/base.html
janno42: like how?

janno42: ah, i think i know what you mean.

karyon: i'm not quite sure what you're asking, i was thinking something like `minimumInputLength: elem..children('option').length > 50 ? 3 : 0`

actually, if i understand correctly, without that you'd be introducing a bug. `applySelect2` is called e.g. when adding a form to the contribution formset on the course edit page, and the added forms would have a `minimumInputLength` of 0 while the initially visible forms would have set that to 3.

evap/results/templates/results_index_course.html
karyon: that is weird. https://docs.djangoproject.com/en/2.1/ref/templates/builtins/#date says `When used without a format string, the DATE_FORMAT format specifier is used`, and from a quick glance at the code, it does just that. could you investigate?

janno42: it's not a date, but a datetime ;)

karyon: yes, but the `date` filter does not care as far as i can see.

evap/evaluation/templates/base.html
karyon: does that still work when calling `applySelect2` with a single element like when adding a form?

janno42: yes. e.g., on the course edit page i can add additional contributors and the selection field there still requires three characters to be typed before showing options.

evap/contributor/views.py
```diff
@@ -17,17 +17,20 @@
...
     own_courses = Course.objects.filter(contributions__contributor=user, state__in=contributor_visible_states)
 
     represented_users = user.represented_users.all()
     delegated_courses = Course.objects.exclude(id__in=own_courses).filter(contributions__can_edit=True, contributions__contributor__in=represented_users, state__in=contributor_visible_states)
 
-    all_courses = list(own_courses) + list(delegated_courses)
-    all_courses.sort(key=lambda course: list(STATES_ORDERED.keys()).index(course.state))
+    displayed_courses = list(own_courses)
+    if show_delegated:
+        displayed_courses += list(delegated_courses)
```
karyon: right now, if the user chose to not show delegated courses, you are still getting them only to check in the template whether any such course exists.
i'd move the `delegated_courses` here and check whether the user `is_delegate` in the template.

janno42: done and i completely removed the delegated courses from the template data by adding an attribute to the delegated courses.

PR #1257, opened 2018-09-03
Simpler translate and stuff

karyon:
see commit messages.


janno42:
look good (apart from the test failures 😉)

Issue #1246, opened 2018-08-03
Investigate making the general contribution optional


karyon:
Currently, each `Course` must have a general contribution, enforced in it's `save` method.

This is unnecessary at least for single results, where it never contains any questionnaires. IIRC it also makes some tests more verbose.

We don't remember exactly we we have that constraint. We should find out, and either remove it, or document it.

karyon:
not requiring this for normal courses would require us to make e.g. templates handle that case. @janno42 couldn't think of a scenario where we want to use that "feature" and it would be some work, so we'll leave it as it is.

PR #1259, opened 2018-09-17
Make single results not have general contributions

karyon:
I tried removing the part in Course.save that always adds a general contribution, but there were several dozen tests that depend on that. besides, if removing that constraint, we would need to make e.g. templates handle that case, and we don't see much value that would justify the effort. so this PR is as good as it will get.

ref #1246


janno42:
please check the one codacy issue

PR #1263, opened 2018-10-06
Remove ViewTest where possible

### LINKED ISSUE #1245: Remove ViewTest where possible
Right now we have a `class ViewTest`, for which there is one subclass for each view that we have.

For views that we have tested properly, it provides no additional value and I I propose to replace it with the original `WebTest`. 

Originally I proposed to remove it altogether and copypaste its test to all the test cases that wouldn't have any valuable test otherwise. @janno42 convinced me to leave it there and rename it to `WebTestWith200Check` instead.

karyon:
fixes #1245. review per commit.

### DIFF
--- a/evap/grades/tests.py
+++ b/evap/grades/tests.py
@@ -3,10 +3,11 @@ from datetime import date, datetime, timedelta
 from django.core import mail
 from django.contrib.auth.models import Group
 
+from django_webtest import WebTest
 from model_mommy import mommy
 
 from evap.evaluation.models import UserProfile, Course, Questionnaire, Contribution, Semester
-from evap.evaluation.tests.tools import WebTestWith200Check, WebTest
+from evap.evaluation.tests.tools import WebTestWith200Check
 
 
 class GradeUploadTests(WebTest):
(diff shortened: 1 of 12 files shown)


evap/grades/tests.py
```diff
@@ -183,34 +182,37 @@ def test_visible_semesters(self):
...
-        self.app.get('/grades/semester/2', user="grade_publisher", status=403)
+        self.assertIn(semester_course.name, page)
 
+    def test_403_on_deleted(self):
+        mommy.make(Semester, pk=1, grade_documents_are_deleted=True)
+        self.app.get('/grades/semester/1', user="grade_publisher", status=403)
 
-class GradeDocumentCourseWebTestWith200Check(WebTestWith200Check):
+
+class GradeCourseView(WebTest):
```
He3lixxx: GradeCourseViewTest?

evap/grades/tests.py
```diff
@@ -183,34 +182,37 @@ def test_visible_semesters(self):
         self.assertNotIn(self.archived_semester.name, page)
 
 
-class GradeDocumentSemesterWebTestWith200Check(WebTestWith200Check):
+class GradeSemesterView(WebTest):
```
He3lixxx: GradeSemesterViewTest?

PR #1264, opened 2018-10-08
update_production.sh: stop apache at the beginning, start it at the end

### LINKED ISSUE #1237: Loading results page during cache preparation results in broken page
When a user requests the results page after `clear_cache` and while running `refresh_results_cache`, the results page is rendered with missing data (resulting in missing distribution bars) and is then cached as such and returned on further page loads.
The cache should be overwritten with the correct data on `refresh_results_cache`.

karyon:
fixes #1237

we noticed that anything else would be unsafe.

what's missing is putting apache into some kind of maintenance mode, but right now i don't want to do that and it's kind of out of scope for evap.

### DIFF
--- a/deployment/update_production.sh
+++ b/deployment/update_production.sh
@@ -4,6 +4,9 @@ cd `dirname $0`/.. # change to root directory
 
 sudo -H -u evap git fetch
 
+# Note that apache should not be running during most of the upgrade,
+# since then e.g. the backup might be incomplete or the code does not
+# match the database layout, or https://github.com/fsr-itse/EvaP/issues/1237.
 sudo service apache2 stop
 
 # argument 1 is the filename for the backupfile.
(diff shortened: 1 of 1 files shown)


deployment/update_production.sh
```diff
@@ -14,20 +18,17 @@ if [ $# -eq 1 ] # if there is exactly one argument
...
 sudo -H -u evap pip3 install --user -r requirements.txt
 sudo -H -u evap ./manage.py compilemessages
 sudo -H -u evap ./manage.py collectstatic --noinput
 sudo -H -u evap ./manage.py compress --verbosity=0
 sudo -H -u evap ./manage.py migrate
-# reload only after static files are updated, so the new code finds all the files it expects.
-# also, reload after migrations happened. see https://github.com/fsr-itse/EvaP/pull/817 for a discussion.
-sudo service apache2 reload
-# update caches. this can take minutes but doesn't need a reload.
 sudo -H -u evap ./manage.py clear_cache
```
He3lixxx: Maybe link the issue so we will never fall for this again?

karyon: like this?

PR #1266, opened 2018-10-15
Don't scroll to page start on language change, fixes #1265

### LINKED ISSUE #1265: Don't scroll to page start on language change
In `evap/evaluation/templates/navbar.html:76`, we have the language buttons (html anchors). Removing the `href="#"` attribute is legal in html5 and will prevent the browser from scrolling to the top of the page. Also, modern browsers will then scroll back to the same position after reloading the page, which increases usability. Thus, we should remove the `href` attributes there.

BenBals:


karyon:
hey, the author of your commit is lars kappert...? you might want to fix that before have that in our master for all eternity :)

BenBals:
I have looked into the problem and even though I did not find it's cause, I was able to change the author information.

PR #1269, opened 2018-10-15
Set Course.last_modified_user only when saving the course form (#1097)

### LINKED ISSUE #1097: Set Course.last_modified_user only when saving the course form
Course.`last_modified_user` should only be set if some user edits the course through the course edit form. At the same time, the `last_modified_time` should be set manually to the current time.

Currently, `helper_semester_course_operation_start` sets the last_modified_time but not the user, but it should set neither. `update_courses` sets both, but should set neither. The `cronjob_user` is used for that, it can be removed completely (basically, create a migration that reverts `0053_create_cronjob_user.py`). `last_modified_time` is currently set automatically by django (`auto_now=True`), this would need to be changed.

This will partly revert https://github.com/fsr-itse/EvaP/pull/845. Rationale: According to @janno42, the `last_modified_user` is used to determine who to talk to in case of questions. If some staff user changes the state of a course that's of no interest, and the cronjob user is not helping at all in that regard.

jkhlr:
Fixes #1097

- `course.last_modified_time` and `course.last_modified_user` are now only updated if changes are made through `staff.CourseForm` or `contributor.CourseForm`
- if neither the values of the form, nor the values of the contributor formset are changed, `course.last_modified_time` and `course.last_modified_user` are not updated
- the user `cronjob` has been removed

I hope I didn't break anything! ;)


evap/evaluation/migrations/0085_auto_20181015_1908.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 2.0.9 on 2018-10-15 19:08
+
+from django.db import migrations, models
```
karyon: can you give the migration a more descriptive filename?

evap/evaluation/migrations/0085_auto_20181015_1908.py
```diff
@@ -0,0 +1,28 @@
...
+import django.utils.timezone
+
+
+def delete_cronjob_user(apps, _schema_editor):
+    UserProfile = apps.get_model('evaluation', 'UserProfile')
+    try:
+        UserProfile.objects.get(username='cronjob').delete()
+    except UserProfile.DoesNotExist:
+        pass
```
karyon: i would expect the user to always exist, so no need for the error handling.

please add a reverse part to the migration (for examples, search for `reverse_code` in the migrations). that is, when the migration is undone (e.g. in order to go back to an older revision to debug things), the cronjob user would need to be created again.

evap/staff/forms.py
```diff
@@ -134,19 +135,13 @@ class CourseForm(forms.ModelForm):
         label=_("General questions")
     )
     semester = forms.ModelChoiceField(Semester.objects.all(), disabled=True, required=False, widget=forms.HiddenInput())
-
-    # the following field is needed, because the auto_now=True for last_modified_time makes the corresponding field
-    # uneditable and so it can't be displayed in the model form
-    # see https://docs.djangoproject.com/en/dev/ref/models/fields/#datefield for details
-    last_modified_time_2 = forms.DateTimeField(label=_("Last modified"), required=False, localize=True, disabled=True)
-    # last_modified_user would usually get a select widget but should here be displayed as a readonly CharField instead
-    last_modified_user_2 = forms.CharField(label=_("Last modified by"), required=False, disabled=True)
```
karyon: nice!

evap/evaluation/migrations/0085_auto_20181015_1908.py
jkhlr: sure!

evap/evaluation/migrations/0085_auto_20181015_1908.py
jkhlr: makes sense... i added a reverse_code function

karyon:
you lost the migration again :) (and in case you didn't change that in the meantime, you need to reference the create method in the RunPython line)

jkhlr:
> you lost the migration again :) (and in case you didn't change that in the meantime, you need to reference the create method in the RunPython line)

oh, sorry :o 
fixed!

evap/contributor/views.py
```diff
@@ -135,8 +134,12 @@ def course_edit(request, course_id):
         if post_operation not in ('save', 'approve'):
             raise SuspiciousOperation("Invalid POST operation")
 
-        course_form.save(user=user)
-        formset.save()
+        if course_form.has_changed():
+            course_form.save(user=request.user)
+        elif formset.has_changed():
```
He3lixxx: This `elif` will break if a user changes both, the `formset` and the `course_form`. You probably want something like:

```
if course_form.has_changed() or formset.has_changed():
    # Save form, even if only formset has changed, to update last_modified_user
    course_form.save(user=request.user)

if formset.has_changed():
    formset.save()
```

Also, since we are obivously missing a tests for this case, would you mind adding one? Otherwise, I would add one, if you don't want to.

evap/staff/views.py
```diff
@@ -651,9 +651,12 @@ def notify_reward_points(grantings, **_kwargs):
 
         if course.state in ['evaluated', 'reviewed'] and course.is_in_evaluation_period:
             course.reopen_evaluation()
+
         if form.has_changed():
             form.save(user=request.user)
-        if formset.has_changed():
+        elif formset.has_changed():
```
He3lixxx: Same as above.

evap/contributor/views.py
jkhlr: Hey, thanks for the review! you are totally right, I flipped around the order of the conditions here 😬 I changed it to 
```
if formset.has_changed():
    formset.save()
    # Save form, even if only formset has changed, to update last_modified_user
    form.save(user=request.user)
elif form.has_changed():
    form.save(user=request.user)
```
which should be equivalent to your proposal, right?

I also added tests for both cases (formset changed, nothing changed)

evap/evaluation/migrations/0085_delete_cronjob_user.py
```diff
@@ -0,0 +1,32 @@
...
+def create_cronjob_user(apps, _schema_editor):
+    UserProfile = apps.get_model('evaluation', 'UserProfile')
+
+    UserProfile.objects.create(username="cronjob")
+
+
+def delete_cronjob_user(apps, _schema_editor):
+    UserProfile = apps.get_model('evaluation', 'UserProfile')
+
+    UserProfile.objects.get(username="cronjob").delete()
```
janno42: this migration fails because the `UserProfileManager` excludes the cronjob user by default (see https://github.com/fsr-itse/EvaP/blob/master/evap/evaluation/models.py#L882).
please make sure to remove all occurrences in the `UserProfileManager` as well.

jkhlr: @janno42 I think I took care of that: https://github.com/fsr-itse/EvaP/pull/1269/files#diff-293b09da50fa9fc3f6f832c04020b4c9L881

jkhlr:
@janno42 
> `last_modified_user` and `last_modified_time` are currently also updated if no data was changed, but staff only used the form to approve the course.
> as stated in the issue, the information should not be changed if the "save and approve" operation only consists of the course approval.
I am not sure if I implemented this behavior correctly, I added a test to check the functionality: https://github.com/fsr-itse/EvaP/pull/1269/files#diff-75573cf7a4c2b0667ee437d73897ac1aR1084
Is that correct?

evap/evaluation/migrations/0085_delete_cronjob_user.py
janno42: can you run the migration locally? it fails on my testing system.

jkhlr: I deleted my local database, recreated it and ran all migrations successfully. Does a user with username `cronjob` exist on your testing system before this migration is applied?

janno42:
i need to check that again but in my short test yesterday, opening a course form and clicking "save and approve" without any further changes did update the values.

jkhlr:
> i need to check that again but in my short test yesterday, opening a course form and clicking "save and approve" without any further changes did update the values.

@janno42 I can reproduce this. The reason seems to be that the `order` attribute of a `contribution` is not properly set (it is 0 for all instances) in the sample data. In the frontend, the hidden `order` field is then populated with the right value (through JS, I guess), which triggers a change in the formset and updates the timestamp the first time a course is saved (approval does not make a difference). Any subsequent save/approval does not update the timestamp. When creating a new course through the frontend, this does not happen (because the `order` fields are then already populated correctly by the frontend). 

This means that the problem is not really existent, it just occurs because of the incorrect sample data. If you want, I can update the sample data, so that this problem is also solved on the dev instances.

evap/evaluation/migrations/0085_delete_cronjob_user.py
He3lixxx: Interesting. I had this failure too, though it was because I messed up my local database somewhere in between. reload_testdata worked fine. I think the issue is that (before this pr) we don't have a cronjob user in our test_data.json (I didn't find it, at least). Thus, when we do reload testdata, we get:

- reset
- migrate (creates cronjob user)
- flush (deletes cronjob user)
- loaddata (doesn't create cronjob user, since it is not in the testdata.json)
- ...

Thus, the migration fails as the user simply does not exist. Can @karyon or @janno42 verify?

evap/evaluation/migrations/0085_auto_20181015_1908.py
janno42: just saw this - we can either add the few lines again to prevent the migration fail on testing systems or (as it is not important for production) we could argue that testing instances need to be reset after applying this pr.

janno42:
Another remark: Please remove the merge commit from this pr and instead rebase on the current master if you depend on any changes that have been made in between.

jkhlr:
> Another remark: Please remove the merge commit from this pr and instead rebase on the current master if you depend on any changes that have been made in between.

done. I hope this PR is now ready to be merged :)

jkhlr:
@janno42 there's another thing I noticed: in the `test_data`, there are some contributions which have `null` as a contributor (e.g. pk=3994). These contributions apparently are not rendered in the form. Do you know why they exist? they make it hard to order the `test_data` correctly...

karyon:
that's the `general contribution`, used for questions that are not linked to a individual contributor (we call them course questions)

evap/contributor/views.py
```diff
@@ -130,13 +129,16 @@ def course_edit(request, course_id):
...
             raise SuspiciousOperation("Invalid POST operation")
 
-        course_form.save(user=user)
-        formset.save()
+        if formset.has_changed():
+            formset.save()
+            # Save form, even if only formset has changed, to update last_modified_user
+            course_form.save(user=request.user)
+        elif course_form.has_changed():
+            course_form.save(user=request.user)
```
karyon: so the comment you added made me think: we're not doing anything complicated. if we need to explain this with a comment, maybe it's not the right way.

what about this:
```
if course_form.has_changed() or formset.has_changed():
    course.form.instance.set_last_modified_user(request.user)
course_form.save()
formset.save()

```
no need to explain anything here i think. `set_last_modified_user` would also set the time, and output the log message. in that message, we would lose the information that it's been a contributor or a staff member, but i don't think that's important. calling the save methods unconditionally is kind of unnecessary, but well, we do that everywhere else as well.

i also thought about
```
set_last_modified_user =  course_form.has_changed() or formset.has_changed():
course_form.save(set_last_modified_user, request.user)
formset.save()
```
to document that a course form save should probably also set the last_modified_user, but then again that's misleading since a formset save should also update it, right?

janno42: i prefer the first suggestion. having just one message without information whether the current user is contributor or staff is perfectly fine - we still have the username in there.

evap/contributor/views.py
```diff
@@ -130,13 +129,16 @@ def course_edit(request, course_id):
...
 
         if post_operation == 'approve':
```
janno42: since you're already on this: maybe you could also improve the following lines so when approving the course without any further changes it shows a message `"Successfully approved course."`? we would then have three different messages for update / approve / update and approve. it would be less misleading than the current message.

evap/staff/forms.py
```diff
@@ -134,19 +135,13 @@ class CourseForm(forms.ModelForm):
...
+    last_modified_user_name = forms.CharField(disabled=True, required=False)
```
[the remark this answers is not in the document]
janno42: true. please make sure the labels are translated.

evap/staff/views.py
```diff
@@ -651,10 +651,13 @@ def notify_reward_points(grantings, **_kwargs):
...
             course.reopen_evaluation()
-        if form.has_changed():
-            form.save(user=request.user)
+
         if formset.has_changed():
             formset.save()
+            # Save form, even if only formset has changed, to update last_modified_user
+            form.save(user=request.user)
+        elif form.has_changed():
+            form.save(user=request.user)
```
janno42: remember to apply all changes resulting from https://github.com/fsr-itse/EvaP/pull/1269/files#r226777525 here as well.

evap/evaluation/tests/tools.py
```diff
@@ -63,10 +66,22 @@ def create_course_with_responsible_and_editor(course_id=None):
     contributor = mommy.make(UserProfile, username='responsible')
     editor = mommy.make(UserProfile, username='editor')
 
+    in_one_hour = (timezone.now() + timedelta(hours=1)).replace(second=0, microsecond=0)
+    tomorrow = (timezone.now() + timedelta(days=1)).date
+    course_params = dict(
+        state='prepared',
+        degrees=[mommy.make(Degree)],
+        vote_start_datetime=in_one_hour,
+        vote_end_date=tomorrow
```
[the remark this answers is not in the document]
jkhlr: @karyon My idea here was to make sure that the data generated for tests is realistic: It is not possible to create a Course with with a `vote_start_datetime` in the future through the app, so why should a factory by default violate this validation logic? There are indeed some other tests (e.g. `test_contributor_course_edit) that faced the same problem. Is there any case in which `timezone.now()` is actually a useful value for `vote_start_datetime`?
The same goes for second and microsecond: it is not possible to set them to a value different than `0` through the app.
If you still feel like it would be better to revert these changes and put the logic into the TestCase, I would also be fine with that.

evap/evaluation/tests/tools.py
```diff
@@ -63,10 +66,22 @@ def create_course_with_responsible_and_editor(course_id=None):
...
     if course_id:
-        course = mommy.make(Course, state='prepared', degrees=[mommy.make(Degree)], id=course_id)
-    else:
-        course = mommy.make(Course, state='prepared', degrees=[mommy.make(Degree)])
+        course_params['id'] = course_id
+
+    course = mommy.make(
+        Course,
+        **course_params
+    )
```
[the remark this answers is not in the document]
jkhlr: ok ;)

evap/staff/tests/test_views.py
```diff
@@ -1053,6 +1053,60 @@ def test_last_modified_user(self):
...
+            user=test_user.username,
+            status=200
+        )
+        form = page.forms["course-form"]
+
+        # Change label of the first contribution
+        form['contributions-0-label'] = 'test_label'
+        form.submit(name="operation", value="approve")
+
+        self.course = Course.objects.get(pk=self.course.pk)
```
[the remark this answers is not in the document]
jkhlr: I fixed that, thanks for pointing it out!

evap/staff/tests/test_views.py
```diff
@@ -1053,6 +1053,60 @@ def test_last_modified_user(self):
...
+        last_modified_time_before = self.course.last_modified_time
+
+        test_user = mommy.make(
+            UserProfile,
+            username='approve_test_user',
+            groups=[Group.objects.get(name='Manager')]
+        )
+
+        page = self.app.get(
+            '/staff/semester/{}/course/{}/edit'.format(self.course.semester.pk, self.course.pk),
```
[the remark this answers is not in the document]
jkhlr: fixed

evap/staff/forms.py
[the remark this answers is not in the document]
jkhlr: Oh I'm sorry, I totally forgot that! 😬 fixed!

evap/contributor/views.py
jkhlr: I think a lot of complexity here is caused by the fact that the view deals with both the form and the formset. I don't really see a reason why the formset should not be handled by the form itself. This way, the form would always know if it has to update the `last_modified` timestamp or not....
But since this PR already got big enough, I implemented @karyon's first suggestion (with a little modification to better deal with the "only approve" requirement).

evap/contributor/views.py
jkhlr: done!

evap/staff/views.py
jkhlr: done!

evap/evaluation/tests/tools.py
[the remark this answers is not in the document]
karyon: thanks. my idea was to not specify anything that the tests do not rely on. if we start adding code that in the end is not needed for the tests to do their thing, that's noise that makes the signal harder to see and understand. e.g. the microsecond thing, our code has no reason to not handle that correctly. if in the future it does not handle that anymore, then it's 100% reasonable to add it, but now it's just additional code that a programmer new to the project might wonder whether it has any significance.

but your point about realistic data is valid, too. i guess we need someone more senior to decide on that :D on a more serious note, i think i still prefer "my" approach, but yours is fine as well, so it's your call. 

note that while contributors must set the start date to the future, and the end date after the start date, staff members do not have the former restriction, only the latter. so they can set both the start and end date into the past. also, when anyone sets the dates, then waits for these dates to pass, and then opens the course again, they would have a form with both dates in the past. the only thing that should really not be possible is an end date that's smaller than the start date. `timezone.now()` might be useful if you want to create an evaluation that should start immediately.

evap/staff/tests/test_views.py
```diff
@@ -915,6 +913,10 @@ def setUpTestData(cls):
         # This is necessary so that the call to is_single_result does not fail.
         responsible = mommy.make(UserProfile)
         cls.contribution = mommy.make(Contribution, course=cls.course, contributor=responsible, responsible=True, can_edit=True, comment_visibility=Contribution.ALL_COMMENTS)
+        cls.url = reverse('staff:course_edit', args=[cls.course.semester.pk, cls.course.pk])
```
karyon: that's not a bad idea in general, but right now every view testcase has this `url` class variable by convention. i'm not 100% sure that's the best we can do, but would stick to it for consistency until we make a different decision.

evap/contributor/tests/test_views.py
```diff
@@ -220,3 +220,37 @@ def test_information_message(self):
...
+        """
+        self.assertIsNone(self.course.last_modified_user)
+        last_modified_time_before = self.course.last_modified_time
+
+        page = self.app.get(self.url, user='responsible', status=200)
+        form = page.forms["course-form"]
+        form.submit(name="operation", value="approve")
+
+        self.course = Course.objects.get(pk=self.course.pk)
+        self.assertIsNone(self.course.last_modified_user)
```
karyon: perhaps assert that the state actually changed? otherwise the form might have thrown an error and we wouldn't notice.

evap/staff/views.py
```diff
@@ -651,16 +653,21 @@ def notify_reward_points(grantings, **_kwargs):
...
         if course.state in ['evaluated', 'reviewed'] and course.is_in_evaluation_period:
             course.reopen_evaluation()
-        if form.has_changed():
-            form.save(user=request.user)
-        if formset.has_changed():
-            formset.save()
+
+        form_has_changed = form.has_changed() or formset.has_changed()
+
+        if form_has_changed:
```
karyon: i would prefer this variable to be inlined, because it doesn't add much value and is more text to parse, but you don't need to change it.

evap/staff/tests/test_views.py
```diff
@@ -1038,6 +1040,52 @@ def test_last_modified_user(self):
...
+            username='approve_test_user',
+            groups=[Group.objects.get(name='Manager')]
+        )
+
+        page = self.app.get(self.url, user=test_user, status=200)
+        form = page.forms["course-form"]
+        form.submit(name="operation", value="approve")
+
+        self.course = Course.objects.get(pk=self.course.pk)
+        self.assertEqual(self.course.last_modified_user, last_modified_user_before)
```
karyon: same here

karyon:
regarding the form handling the formset, that sounds like an idea worth investigating. no need to do that in this PR (or you doing that at all), but i'd be interested in details of your suggestion, let's talk about that the next time we meet.

evap/staff/views.py
jkhlr: `form_has_changed` is reused later (l.667), that's why I did not inline it

evap/contributor/tests/test_views.py
jkhlr: good point! I will change that...

evap/staff/views.py
karyon: oh, right.

evap/evaluation/tests/tools.py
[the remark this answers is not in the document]
jkhlr: just to put it into context: my motivation for this change was that [this test](https://github.com/fsr-itse/EvaP/pull/1269/files/99cf84e234e63647664b3fea84c9a2a69747a17e#diff-ac9ab9135a427edac9993bc57dd13dceR243) failed: the form always submits a datetime with seconds=0 and microseconds=0, so the form submission would always trigger a change, even though we are only approving the course (and the form submission would generally fail, because `mommy`s default `vote_start_datetime = timezone.now()` is already in the past when we do the request)

karyon:
please check the codacy report. feel free to ignore the setUp thing and "too many branches" warning.

jkhlr:
juhuuu 🎉

karyon:
thanks a lot! by the way, you got from zero to a top ten PR (when looking at the numbers of comments) within a week! that certainly doesn't mean anything helpful, but it's still cool.

He3lixxx:
:tada:

PR #1278, opened 2018-10-29
refreshes the results cache after a change

### LINKED ISSUE #1236: Refresh results page cache on changes
When these (and possibly a few other) things are changed, the results page cache has to be refreshed:
- course name
- semester name
- degree name
- course type name
- course's responsible

Tardigrade00:
fixes #1236 
The results cache now updates if the semester name, course type name or degree name is changed, so that it is also updated in the display of the courses.
A problem that arose is that there is a race condition in the function update_template_cache.


He3lixxx:
Looks good so far :)
To circumvent the race condition we could probably use django's cache.set or cache.set_many methods, see the documentation here: https://docs.djangoproject.com/en/2.1/topics/cache/#basic-usage

This would be a refactoring that is quite a bit bigger I think (because we don't have a single template for just one row afaik), though, and I didn't see anything about atomic updates.

karyon:
we do have a template for one row! `results_index_course.html`. it's actually rendered separately in the warmup method.

we might be able to do this race-free when handling the template caching completely in python-code and not in the template (the template would then insert a string per line that was rendered in the view previously), but that makes all of this even uglier.

another option would be to make the cache optional, so that the template can be rendered without calling warmup first. (actually, only then it would actually deserve the name "cache", right? :) )

but that's something for another PR, i guess. i think we'll live with the race condition for now.

evap/staff/forms.py
```diff
@@ -116,6 +131,13 @@ def clean(self):
         if self.cleaned_data.get('DELETE') and not self.instance.can_manager_delete:
             raise SuspiciousOperation("Deleting course type not allowed")
 
+    def save(self, *args, **kwargs):
+        course_type = super().save(*args, **kwargs)
+        if "name_en" in self.changed_data or "name_de" in self.changed_data:
+            for course in course_type.courses.filter(state="published"):
+                update_template_cache(course)
```
karyon: i noticed we have this for-loop in all three `save`-methods, only to call the same method inside each of these loops. i would suggest to move the loop into that method. like this:
```
if "name_en" in ... <same as before>:
    update_template_cache(course_type.courses.filter(state="published"))
```

and then in `update_template_cache`
```
def update_template_cache(courses):
    for course in courses:
        <do what you did previously in update_template_cache>
```

and then, since `warm_up_template_cache` can already be called with multiple courses, we could call it with all the courses at once instead of with each individual one.

evap/results/views.py
```diff
@@ -23,6 +23,10 @@ def get_course_result_template_fragment_cache_key(course_id, language, can_user_
 
 def delete_template_cache(course):
     assert course.state != 'published'
+    delete_template_cache_impl(course)
+
+
+def delete_template_cache_impl(course):
     caches['results'].delete(get_course_result_template_fragment_cache_key(course.id, 'en', True))
```
karyon: i get the intent that you wanted to keep the assert for the other calls to this method, but now we have two similar methods, where the only difference is an added assert. i wouldn't do that and just remove the assert. @He3lixxx if you disagree, lets talk about that.

He3lixxx: (All below is afaik, there might be something wrong in there, please correct me)

Keeping these as two methods way mainly my request. My idea is that this function is dangerous. I (as a non-evap guy) wouldn't expect that you can break stuff if you delete cache entries, I would rather expect the system to fix itself on the fly when it encounters missing cache entries (in the end, it's a cache that doesn't give integrity guarantees). However, we're doing something quite different. Published courses must always have valid cache entries, they will be created on publishing and removed on unpublishing. Non-published courses must not have cache entries. These are some very strict constraints, and if they are violated, we will probably end up in a state where every page rendering will be incorrect.

In favor of keeping this seperated:
- I favor embedding these constraints in the code in general rather than just having the knowledge in our heads. (for the sake of anyone new or anyone who forgets how exactly things were)
- I think that the whole code around caching the results is rather complex. I expect us (or atleast myself) to miss some situations where we will try to call these methods where they shouldn't be called in the future. The asserts (somewhat) guarantee no harm is done - the single call will fail, but we will not get a permanent invalid state.
- I like the division in a "safe public method" that other modules may call and a designated "unsafe private method" that may not be called from outside.

In favor of removing the assert:
- As long as every reviewer knows that the cache is a fragile thing here, we will probably not get code through code review that breaks stuff because we know we have to be careful about that.
- We can not assume that any call to the "safe" method is indeed safe as it has this assert and thus might raise an uncaught exception, so even with the "safe" method, we still need to be careful.

I'd say the last point makes merging these methods and removing the assert reasonable, so I don't have a strong opinion, but I'm still slightly in favor of keeping the assert. Anything I don't see so far?

karyon: i think you understood the problem perfectly, and i agree with all you said.

what about this solution:

```python
def do_the_right_thing(courses): # TODO better name ;)
    assert (all_courses_are_published) or (all_courses_are_not_published)
    for course in courses:
         delete_template_cache(courses)
    if all_courses_are_published:
        warm_up_template_cache(courses)
```
this would be the only public interface. any code that wants to remove, add, or update a course in/from the cache would call this method.  this would always be safe, because it always does the right thing. however, it would include unnecessary work: in some cases we know a course has not been in the cache, the code would still attempt to remove it. also, on the call site, it's less clear what is actually the desired action, because it would always call `do_the_right_thing` and not `warm_up`, `delete`, or `update`.

...actually, having described the downsides, i think i like your solution better.

lets prefix `delete_template_cache_impl` with an underscore to fulfill the convention, and otherwise leave it as it is.

He3lixxx:
I think we still want to add tests for this.

karyon:
lets merge this and talk about the test later.

He3lixxx:
Nice :+1:

karyon:
thanks @Tardigrade00!

Issue #1272, opened 2018-10-16
configure TravisCI and Codacy to support python 3.5


jkhlr:
Currently, tests are failing on TravisCI if python `3.5` features (like f-strings) are used, since it currently seems to execute the tests with python `3.4.6`. Same for Codacy (f-strings are considered a syntax error there)
On the production instance, python `3.5` seems to be available (https://github.com/fsr-itse/EvaP/issues/1244#issuecomment-417714795), and `3.5` is also the python version in the vagrant environment, so I guess it would make sense to update the CI pipeline accordingly.

jkhlr:
@karyon You are right, f-strings came with python `3.6`. I checked again the version of the python interpreter in the vagrant VM and it is actually `3.6.3`. So if we could enable python `3.6`/`3.7` on the production instance, it would be even better :)

janno42:
You can assume 3.7 to be available.

jkhlr:
feel free to assign me here :)

janno42:
we're now using python 3.7 in travis (#1282) and are waiting for codacy to support it as well.

PR #1282, opened 2018-11-12
#1272 update python version to 3.7

jkhlr:
- Updated python version in vagrant VM
- Updated python version for TravisCI
- Added some tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
 language: python
-python:
-  - "3.4"
+python: "3.7"
+dist: xenial
+sudo: false
+cache:
+  directories:
+  - $HOME/.cache/pip
+
 services:
 - postgresql
 before_script:
@@ -14,8 +19,4 @@ script:
   - python manage.py test --keepdb --reverse
   - python manage.py test --keepdb --debug-mode
 after_success:
-  - coveralls
-sudo: false
-cache:
-  directories:
-  - $HOME/.cache/pip
+  - coveralls\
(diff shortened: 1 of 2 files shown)


jkhlr:
Codacy can apparently not be upgraded to support new python features (e.g. dictionary unpacking) :(

karyon:
codacy only uses some other linters, e.g. prospector, which throws that error. prospector in turn runs multiple other linters and merges their results. could you run prospector locally and find out where the syntax error comes from? then we could make a bug report/feature request to the respective project.

jkhlr:
> codacy only uses some other linters, e.g. prospector, which throws that error. prospector in turn runs multiple other linters and merges their results. could you run prospector locally and find out where the syntax error comes from? then we could make a bug report/feature request to the respective project.

Codacy uses PyLint to lint python code. To define which python version it should lint for, you simply run pylint with the corresponding python version (https://stackoverflow.com/questions/23052637/specify-which-python-version-pylint-should-evaluate-for). Unfortunately, we don't have control over the python runtime Codacy uses, the only option we have is to choose between python2 and python3 (https://support.codacy.com/hc/en-us/articles/115002330985-Specifying-your-Python-version). So this is definitely a problem with Codacy. Probably they are running all linters on the same, not-so-recent python version :(

karyon:
as far as i can see, it's prospector that complains about the syntax. if you expand the "invalid syntax" error in codacy, it says `Prospector_F999` at the bottom.

i just checked and we can enable prospector and pylint separately in codacy. no idea why their pylint does not complain but their prospector does...

karyon:
I've reported the problem to codacy. In the meantime, we decided to go ahead and merge this. For the time being we'll ignore the syntax errors on codacy. In case they won't fix this or don't reply, we can think about building a workaround or disabling prospector.

karyon:
thanks @jkhlr!

PR #1283, opened 2018-11-12
1234 login key list creation

### LINKED ISSUE #1234: Login key list creation
For on-site courses with external users we want to be able to give them access to the evaluation of the course without having to login to their email accounts first.

The proposed solution is a login key export function which allows staff users to (generate and) export login keys for all external users of a selected course as an Excel sheet.

He3lixxx:
fixes #1234
- Login key list export for courses
- Refactoring: Login key authentication logic is now done by an evaluation view.

Do we want / need tests on the export? @karyon? We're testing `UserProfile.login_url`, but not the exporter itself.
Maybe we want to test that only external users are included? Maybe we want to test that the button is only shown for courses where external users exist?


evap/staff/importers.py
```diff
@@ -48,7 +48,7 @@ def store_in_database(self):
                                                                  'title': self.title,
                                                                  'is_active': True})
         if user.needs_login_key:
-            user.refresh_login_key()
+            user.reset_login_key_validity()
```
karyon: isn't that exactly the wrong method in this case?

evap/staff/templates/staff_course_person_management.html
```diff
@@ -72,17 +72,39 @@ <h6 class="card-subtitle mb-2 text-muted">{% trans 'From Excel file' %}</h6>
...
+                        <h4 class="card-title">{% trans 'Export Login Keys' %}</h4>
+                        <h6 class="card-subtitle mb-2 text-muted">{% trans 'To Excel File' %}</h6>
+                        <p class="card-text">{% trans 'This will create an Excel file containing login keys for all external users.' %}</p>
+                    </div>
+                    <div class="card-footer text-center">
+                        {% if course.has_external_user %}
+                            <button name="operation" value="export-login-keys" type="submit" class="btn btn-sm btn-primary form-submit-btn">{% trans 'Export login keys' %}</button>
+                        {% else %}
+                            {% trans 'This course has no external users.' %}
+                        {% endif %}
```
karyon: @janno42 please check, i think the "no external users" case might look weird.

evap/evaluation/views.py
```diff
@@ -80,10 +82,45 @@ def index(request):
...
-    response['Location'] += '?loginkey=' + str(key)
-    return response
+def login_key_authentication(request, key):
+    # AuthenticationMiddleware is required so that request.user exists.
+    if not hasattr(request, 'user'):
+        raise ImproperlyConfigured(
+            "This view requires the"
+            " authentication middleware to be installed.  Edit your"
+            " MIDDLEWARE_CLASSES setting to insert"
+            " 'django.contrib.auth.middleware.AuthenticationMiddleware'")
```
karyon: i think that part has outlived its usefulness. if that middleware is missing, i'm sure there's a massive bunch of stuff failing loudly, i don't know why of all things we should test for this particular middleware to exist.

evap/staff/templates/staff_course_person_management.html
```diff
@@ -72,17 +72,39 @@ <h6 class="card-subtitle mb-2 text-muted">{% trans 'From Excel file' %}</h6>
...
+    </div>
+
+    <div class="row">
+        <div class="col">
+            <form id="login-key-export-form" method="POST" class="form-vertical" action="{% url 'staff:course_login_key_export' semester.id course.id %}">
+                {% csrf_token %}
+                <div class="card">
+                    <div class="card-body import-card">
+                        <h4 class="card-title">{% trans 'Export Login Keys' %}</h4>
+                        <h6 class="card-subtitle mb-2 text-muted">{% trans 'To Excel File' %}</h6>
```
janno42: it's actually not an excel file, but a csv file. can you please change that here and below?

evap/staff/templates/staff_course_person_management.html
janno42: please add a disabled button with an explanatory tooltip instead.

evap/evaluation/models.py
```diff
@@ -616,6 +617,10 @@ def update_courses(cls):
         send_publish_notifications(evaluation_results_courses)
         logger.info("update_courses finished.")
 
+    @property
+    def has_external_user(self):
```
janno42: `has_external_participant`

evap/staff/importers.py
He3lixxx: Hmm. Actually, I don't think the call makes any sense here - login keys should be generated when needed, right?

I thought this was so one could also import login keys and they will be valid from the date imported on, but that doesn't happen, so yeah, I'd remove it. What do you and @janno42 think?
Do we want a test for this?

He3lixxx: Alright, @karyon, @janno42 and me agreed that we don't need this here. Will remove.

He3lixxx:
Test on the exporter is now included - for free!

evap/evaluation/tests/test_auth.py
```diff
@@ -20,29 +21,38 @@ def setUpTestData(cls):
         mommy.make(Contribution, course=course, contributor=cls.external_user, can_edit=True, responsible=True, textanswer_visibility=Contribution.GENERAL_TEXTANSWERS)
         mommy.make(Contribution, course=course, contributor=cls.inactive_external_user, can_edit=True, responsible=False, textanswer_visibility=Contribution.GENERAL_TEXTANSWERS)
 
+    @override_settings(PAGE_URL='https://testdomain.com')
```
karyon: example.com?

evap/evaluation/tests/test_models.py
```diff
@@ -467,32 +467,32 @@ def test_no_login_url_when_delegates_in_cc(self):
         self.user.delegates.add(self.other_user)
         EmailTemplate.send_to_users_in_courses(self.template, [self.course], EmailTemplate.CONTRIBUTORS, use_cc=True, request=None)
         self.assertEqual(len(mail.outbox), 2)
-        self.assertFalse("loginkey" in mail.outbox[0].body)  # message does not contain the login url
-        self.assertTrue("loginkey" in mail.outbox[1].body)  # separate email with login url was sent
+        self.assertFalse("key" in mail.outbox[0].body)  # message does not contain the login url
```
karyon: i would check for a longer string, this substring could also appear in the default email templates.

evap/evaluation/auth.py
```diff
@@ -1,83 +1,11 @@
...
-            # A user exists, but the login key is not valid anymore. Send the user a new one.
-            user.ensure_valid_login_key()
-            EmailTemplate.send_login_url_to_user(user)
-            messages.warning(request, _("The login URL is not valid anymore. We sent you a new one to your email address."))
-        else:
-            messages.warning(request, _("Invalid login URL. Please request a new one below."))
+from evap.evaluation.models import UserProfile
 
 
 class RequestAuthUserBackend(ModelBackend):
```
karyon: the comment in the backend below has a reference to the middleware that you removed.

evap/staff/templates/staff_course_person_management.html
```diff
@@ -72,17 +72,38 @@ <h6 class="card-subtitle mb-2 text-muted">{% trans 'From Excel file' %}</h6>
...
+    </div>
+
+    <div class="row">
+        <div class="col">
+            <form id="login-key-export-form" method="POST" class="form-vertical" action="{% url 'staff:course_login_key_export' semester.id course.id %}">
+                {% csrf_token %}
+                <div class="card">
+                    <div class="card-body import-card">
+                        <h4 class="card-title">{% trans 'Export Login Keys' %}</h4>
+                        <h6 class="card-subtitle mb-2 text-muted">{% trans 'To CSV File' %}</h6>
```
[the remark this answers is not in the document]
He3lixxx: What do you think about "Sample File" which is on the page a few times? Do we want "Sample file" there?

evap/evaluation/tests/test_models.py
He3lixxx: The template is generated above in the test class
`        cls.template = mommy.make(EmailTemplate, body="{{ login_url }}")`

However, i see your point. Essentially, one body is empty, the other is something like "https://example.com/key/144875930" if overriding the PAGE_URL setting. Would you rather test for emptiness, for a part of the url or for the full url?

karyon: i think the complete url, but i have no explanation for that other than that everything else feels even more weird to me. we would need to specify the domain setting for that, but i think that's fine.

He3lixxx:
Will try to fix the commit history. Initially thought I would just squash in the end.

He3lixxx:
About the unused variable `request`: I thought about that and explicitly wanted to leave that there. @karyon, what do you think?

He3lixxx:
Alright I completely rearranged the changes into 4 commits, @karyon maybe take a small look at it. Tests pass in all commits, I think it makes sense this way.

unused variable request and maybe the "Sample File" thing todo

karyon:
could it be that `staff_course_person_import.html` reappeared?

karyon:
can you use `_request`?

He3lixxx:
Hmm, interesting, will fix. Thanks for the hint

He3lixxx:
`_request` would be fine for me. @janno42?

evap/staff/templates/staff_course_person_management.html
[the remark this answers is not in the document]
janno42: yes, we use uppercase just for the first word

janno42:
sure, `_request` is perfectly fine

He3lixxx:
`request` renamed to `_request` on the 4 staff views that don't use it. `staff_course_person_import.html` deleted.

He3lixxx:
Actually, thinking about it, I think "sample file" would be even better - why would we start capitalizing in the parentheses?

image

janno42:
you are perfectly right

He3lixxx:
Alright, we should then be set if the tests pass :)

karyon:
nice! please give the second commit a different message, because it does actually fix things and change the behavior in another place, and then we're good to go i think.

karyon:
👍

PR #1286, opened 2018-11-26
added the display of the username in the contributor form

Tardigrade00:
fixes issue #900

### DIFF
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -367,7 +367,7 @@ class SingleResultForm(forms.ModelForm):
 
 
 class ContributionForm(forms.ModelForm):
-    contributor = forms.ModelChoiceField(queryset=UserProfile.objects.exclude_inactive_users())
+    contributor = UserModelChoiceField(queryset=UserProfile.objects.exclude_inactive_users())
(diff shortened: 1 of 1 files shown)


evap/staff/forms.py
```diff
@@ -282,7 +282,7 @@ def save(self, *args, **kw):
 
 
 class ContributionForm(forms.ModelForm):
-    contributor = forms.ModelChoiceField(queryset=UserProfile.objects.exclude_inactive_users())
+    contributor = UserModelChoiceField(queryset=UserProfile.objects.exclude_inactive_users())
```
He3lixxx: Nice, I just saw that there was code in there before that was supposed to overwrite which field class is used, if you check a few lines below in the Meta class (line 300), you can see that the UserModelChoiceField was set there in a `field_classes` dictionary. Looks like someone forgot to use the right class here, probably when they added the "exclude inactive users" part. Right now, this part in the meta class is not used anymore. Could you remove these lines? (300-302). Sorry that I didn't see that before.

He3lixxx: Also, since this was correct and then broke, we should write a regression test for this. We can either do this together the next time or you look into our tests. We would have to have one test method for every place where we have a user selection, according to @janno42 this should be about 4 places.

janno42:
@He3lixxx please review again

Issue #1091, opened 2018-01-28
Extend anonymize script


janno42:
The anonymize script should optimally (not all of that needs to be done for a PR):
- randomly select degrees for courses
- randomly select course types for courses
- replace email template texts with default values (there might be names, urls and emails in there)
- regenerate course names (maybe take them from list and not just shuffle them around)
- regenerate reward point redemption event names (maybe take them from list and not just shuffle them around)
- regenerate questionnaire names (maybe take them from list)
- regenerate questions (maybe take them from list)
- remove all login keys
- have at least one test. maybe use test_data.json?

karyon:
It's not clear to me why shuffling course names is not enough. Course names are public information. Same for questionnaire names which are rather generic. Reward point redemption event names are likewise nothing that needs to be protected. Alright, maybe there are some questions that are specific enough to necessitate some filtering. Or can you give examples for the other categories?

janno42:
Private courses and their questionnaires aren't public information.

karyon:
True. Then i would like to have the lists generated and filtered from production data. I can do that, if I don't have the clearance for that you'll have to do it ;) you think we should have multiple lists of course names? Single results come to mind since they are titled differently, bachelors projects as well, but the effort is probably not worth it... Or we don't use the lists and simply exclude private courses and questionnaires used exclusively in private courses.

janno42:
By excluding do you mean completely removing private courses from the data? I think it would be better to have some private courses in the test data as well.
I'm fine with having lists generated from production data. And I don't think we have to take course types into account when choosing names.

karyon:
no, i meant excluding private course's names when shuffling the names around and duplicate some other names instead. same for the questionnaires, private courses could simply get some of the "normal" questionnaires.

marcelgarus:
What's still left to do:
* anonymize email templates
* randomize questionnaire names
* randomize questionnaire questions

janno42:
the current state is fine for now.

PR #1287, opened 2018-12-03
Extended anonymize script

marcelgarus:
See issue #1091.

* courses' degrees are randomly selected
* courses' types are randomly selected
* email template subject and body are replaced with sample text and lorem ipsum
* public course names remain, but private ones get assigned some sample names
* all login keys are replaced by random ones

### DIFF
--- a/evap/locale/de/LC_MESSAGES/django.po
+++ b/evap/locale/de/LC_MESSAGES/django.po
@@ -5,8 +5,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: EvaP\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2019-02-19 14:09+0100\n"
-"PO-Revision-Date: 2019-02-19 22:02+0100\n"
+"POT-Creation-Date: 2019-03-18 23:10+0100\n"
+"PO-Revision-Date: 2019-03-18 23:12+0100\n"
 "Last-Translator: Johannes Wolf <johannes-wolf@posteo.de>\n"
 "Language-Team: Johannes Wolf (janno42)\n"
 "Language: de\n"
@@ -38,15 +38,11 @@ msgstr "Name (Englisch)"
 msgid "The first day of evaluation must be before the last one."
 msgstr "Der erste Evaluierungstag muss vor dem letzten liegen."
 
-#: evap/contributor/forms.py:55
-msgid "The first day of evaluation must be in the future."
-msgstr "Der erste Tag des Evaluierungszeitraums muss in der Zukunft liegen."
-
-#: evap/contributor/forms.py:64
+#: evap/contributor/forms.py:58
 msgid "The last day of evaluation must be in the future."
 msgstr "Das Ende des Evaluierungszeitraums muss in der Zukunft liegen."
 
-#: evap/contributor/forms.py:100
+#: evap/contributor/forms.py:94
 msgid "Delegate to"
 msgstr "Delegieren an"
 
@@ -191,7 +187,9 @@ msgstr "Informationen für Mitwirkende"
 
 #: evap/contributor/templates/contributor_index.html:19
 #: evap/contributor/templates/contributor_settings.html:66
-#: evap/evaluation/models.py:1183 evap/staff/templates/staff_user_merge.html:75
+#: evap/evaluation/models.py:1183
+#: evap/staff/templates/staff_semester_preparation_reminder.html:17
+#: evap/staff/templates/staff_user_merge.html:75
 msgid "Delegates"
 msgstr "Stellvertreter"
 
@@ -292,7 +290,7 @@ msgstr "Verbergen"
 #: evap/contributor/templates/contributor_index.html:63
 #: evap/evaluation/models.py:1381
 #: evap/grades/templates/grades_semester_view.html:19
-#: evap/staff/templates/staff_semester_todo.html:32
+#: evap/staff/templates/staff_semester_preparation_reminder.html:32
 #: evap/staff/templates/staff_semester_view.html:227
 #: evap/staff/templates/staff_user_index.html:35 evap/staff/views.py:421
 #: evap/student/templates/student_index.html:41
@@ -300,7 +298,8 @@ msgid "Name"
 msgstr "Name"
 
 #: evap/contributor/templates/contributor_index.html:64
-#: evap/staff/templates/staff_semester_todo.html:33 evap/staff/views.py:421
+#: evap/staff/templates/staff_semester_preparation_reminder.html:33
+#: evap/staff/views.py:421
 msgid "State"
 msgstr "Zustand"
 
@@ -481,7 +480,7 @@ msgstr ""
 "Das Formular wurde nicht gespeichert. Bitte beheben Sie die unten "
 "angezeigten Fehler."
 
-#: evap/contributor/views.py:224
+#: evap/contributor/views.py:227
 msgid ""
 "{} was added as a contributor for evaluation \"{}\" and was sent an email "
 "with further information."
@@ -1839,7 +1838,8 @@ msgid "Invalid login URL. Please request a new one below."
 msgstr "Ungültiger Anmeldelink. Bitte fordern Sie unten einen neuen an."
 
 #: evap/grades/forms.py:15 evap/staff/forms.py:153 evap/staff/forms.py:196
-#: evap/staff/forms.py:285 evap/staff/templates/staff_semester_todo.html:35
+#: evap/staff/forms.py:285
+#: evap/staff/templates/staff_semester_preparation_reminder.html:35
 msgid "Last modified by"
 msgstr "Zuletzt verändert von"
 
@@ -2535,7 +2535,8 @@ msgstr "Veranstaltung erfolgreich geändert."
 msgid "RewardPoints"
 msgstr "Belohnungspunkte"
 
-#: evap/staff/forms.py:29 evap/staff/templates/staff_semester_todo.html:34
+#: evap/staff/forms.py:29
+#: evap/staff/templates/staff_semester_preparation_reminder.html:34
 msgid "Start of evaluation"
 msgstr "Beginn der Evaluierung"
 
@@ -3629,6 +3630,16 @@ msgstr ""
 "Veranstaltungen werden erstellt und verbunden. Außerdem werden die "
 "angegebenen Standardwerte für alle Evaluierungen eingetragen."
 
+#: evap/staff/templates/staff_semester_preparation_reminder.html:7
+#: evap/staff/templates/staff_semester_send_reminder.html:7
+#: evap/staff/templates/staff_semester_view.html:52
+msgid "Preparation reminder"
+msgstr "Erinnerung zur Vorbereitung"
+
+#: evap/staff/templates/staff_semester_preparation_reminder.html:25
+msgid "Send Reminder"
+msgstr "Erinnerung senden"
+
 #: evap/staff/templates/staff_semester_questionnaire_assign_form.html:5
 #: evap/staff/templates/staff_semester_questionnaire_assign_form.html:10
 #: evap/staff/templates/staff_semester_questionnaire_assign_form.html:24
@@ -3647,20 +3658,6 @@ msgstr ""
 "verantwortlichen Mitwirkenden zugewiesen werden sollen. Die Fragebögen "
 "werden nur Evaluierungen in Vorbereitung zugewiesen. Wenn nichts ausgewählt "
 "wird, bleibt die aktuelle Zuweisung für diesen Veranstaltungstyp unverändert."
-
-#: evap/staff/templates/staff_semester_send_reminder.html:7
-#: evap/staff/templates/staff_semester_todo.html:7
-#: evap/staff/templates/staff_semester_view.html:52
-msgid "Todo"
-msgstr "Todo"
-
-#: evap/staff/templates/staff_semester_todo.html:17
-msgid "Delegate"
-msgstr "Stellvertreter"
-
-#: evap/staff/templates/staff_semester_todo.html:25
-msgid "Send Reminder"
-msgstr "Erinnerung senden"
 
 #: evap/staff/templates/staff_semester_view.html:20
 msgid "Archiving"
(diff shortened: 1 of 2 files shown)


evap/evaluation/management/commands/anonymize.py
```diff
@@ -111,22 +137,100 @@ def replace_usernames_emails(random_usernames):
...
+
+            # Randomize the degrees.
+            # Either, this is a Bachelor course or a Master course or a
+            # Bachelor/Master course or something else
+            self.stdout.write("Randomizing degrees...")
+            for course in courses:
+                # Build a set of degrees and add some degrees to it (at least
+                # one).
+                new_degrees = { random.choice(all_degrees) }
+                for _ in range(len(all_degrees)):
```
janno42: please use `__` (two underscores)

evap/evaluation/management/commands/anonymize.py
```diff
@@ -111,22 +137,100 @@ def replace_usernames_emails(random_usernames):
...
+            # random courses get shuffled and their names then get applied to
+            # the private ones in order. If there are more private courses than
+            # public ones, the remaining private courses just get some default
+            # names.
+            self.stdout.write("Replacing private course names...")
+            random.shuffle(public_courses)
+
+            for i, course in enumerate(private_courses):
+                # Give courses unique temporary names to counter identity errs.
+                course.name_de = f"<privater Kurs #{i}>"
```
janno42: please call it `Veranstaltung` in german

evap/evaluation/management/commands/anonymize.py
```diff
@@ -76,33 +72,63 @@ def anonymize_data(self):
...
-        while len(random_usernames) != user_count:
-            random_usernames.add((random.choice(first_names), random.choice(last_names)))
-        return random_usernames
-
-    @staticmethod
-    def replace_usernames_emails(random_usernames):
-        for user, new_username in zip(UserProfile.objects.all(), random_usernames):
+    # Anonymizes the email templates by replacing the actual subject with a sample one
+    # and replacing the body with lorem ipsum.
+    def anonymize_email_templates(self, lorem_ipsum):
```
janno42: as discussed: please leave this empty for now - the anonymization should be added once we have extended the EmailTemplates by a list of available variables.

marcelgarus:
I applied all the proposed changes.

evap/evaluation/management/commands/anonymize.py
```diff
@@ -111,22 +129,100 @@ def replace_usernames_emails(random_usernames):
                 new_domain = Command.new_institution_domain if is_institution_domain else Command.new_external_domain
                 user.email = user.username + '@' + new_domain
 
-            user.save()
+            user.login_key = key
+            user.login_key_valid_until = valid_until
```
karyon: if i understand correctly, you're giving every user a login key (plus valid until date). on production, only some of the users have login keys, and i would suggest to give only those users a new one. actually, i would use `ensure_valid_login_key` (so, first delete the existing key, and then call that method) for this, to not duplicate the code that generates the key.

additionally, in the past we had problems that the test data went stale, i.e. that after some months the states of the courses changed. to avoid the same happening here, i would suggest to set the validity to the year 2100 (and perhaps half of them to some past date to reflect that not all login keys are valid on production)

evap/evaluation/management/commands/anonymize.py
```diff
@@ -111,22 +129,100 @@ def replace_usernames_emails(random_usernames):
...
-                course.name_en = shuffled_courses[i].name_en + " "
+            courses = semester.courses.all()
+            public_courses = [course for course in courses if not course.is_private]
+            private_courses = [course for course in courses if course.is_private]
+
+            num_public, num_private = len(public_courses), len(private_courses)
+            self.stdout.write(f"Anonymizing courses of semester {semester} ("
+                + ( f"{num_public} public" if num_private == 0 else 
+                    f"{num_private} private" if num_public == 0 else
+                    f"{num_public} public, {num_private} private, {len(courses)} in total" )
```
karyon: lets keep it simple and always print out the three numbers. it's an admin tool, no need to impress anyone ;)

evap/evaluation/management/commands/anonymize.py
```diff
@@ -111,22 +129,100 @@ def replace_usernames_emails(random_usernames):
...
+            self.stdout.write("Randomizing degrees...")
+            for course in courses:
+                # Build a set of degrees and add some degrees to it (at least
+                # one).
+                new_degrees = { random.choice(all_degrees) }
+                for __ in range(len(all_degrees)):
+                    if random.random() < 0.5:
+                        new_degrees.add(random.choice(all_degrees))
+                course.degrees.set(new_degrees)
+                course.save()
```
karyon: weeeell now every course gets (roughly) half the amount of all existing degrees. i would rather give every course as many degrees as it had before. (you could use `random.sample` for that btw)

evap/evaluation/management/commands/anonymize.py
```diff
@@ -111,22 +129,100 @@ def replace_usernames_emails(random_usernames):
...
+
+            # Assign public courses' names to private ones as their names may
+            # be confidential.
+            # In order to ensure the same course name doesn't appear twice, the
+            # random courses get shuffled and their names then get applied to
+            # the private ones in order. If there are more private courses than
+            # public ones, the remaining private courses just get some default
+            # names.
+            self.stdout.write("Replacing private course names...")
+            random.shuffle(public_courses)
```
karyon: we need to apply the shuffling to the database to break the link between course names and the results.

evap/evaluation/tests/test_commands.py
```diff
@@ -10,14 +11,38 @@
 
 from model_mommy import mommy
 
-from evap.evaluation.models import UserProfile, Course, Semester
+from evap.evaluation.models import UserProfile, Course, Semester, EmailTemplate
 
 
 class TestAnonymizeCommand(TestCase):
     @patch('builtins.input')
     def test_anonymize_does_not_crash(self, mock_input):
```
karyon: that name would need to be changed ;)

karyon:
ah, and i need to ask you to not use f-strings until we have sorted out https://github.com/fsr-itse/EvaP/pull/1282 :(

evap/evaluation/management/commands/anonymize.py
marcelgarus: In the current version, I use `refresh_login_key`, which automatically overwrites the last key.
Then, I change the valid until date by 100 years in the past or future.

evap/evaluation/tests/test_commands.py
marcelgarus: We talked about that.

evap/evaluation/management/commands/anonymize.py
He3lixxx: Jumping in here since I just changed some of the code that does the login key generation:
`refresh_login_key` was just renamed to `reset_login_key_validity` and does exactly that - it resets the validity, but does not touch the actual key. The old name was misleading. You will need `ensure_valid_login_key` as @karyon proposed. Also, you will need to delete the old key before the call as the method keeps a key if it is still valid.

evap/evaluation/management/commands/anonymize.py
```diff
@@ -76,33 +72,41 @@ def anonymize_data(self):
...
-        for user, new_username in zip(UserProfile.objects.all(), random_usernames):
+    # Replaces names, usernames, email addresses, login keys and valid until
+    # dates with fake ones.
+    def anonymize_users(self, first_names, last_names):
+        user_profiles = UserProfile.objects.all()
+
+        # Generate as many fake usernames as real ones exist. Use the provided
+        # first/last names for that.
+        self.stdout.write("Generating fake usernames...")
+        fake_usernames = set()
```
He3lixxx: @karyon have we ever thought about simplifying this? I don't see why we need this build-a-set-set-and-zip procedure, I think looping over all users and just giving him a random first and last name should do exactly the same, be faster, less code and simpler to understand.

He3lixxx: Ahh I see - no duplicates cause it's a set. Funny, I've thought about this, but somehow my mind did completely ignore the fact that sets can not have duplicates.

evap/evaluation/management/commands/anonymize.py
marcelgarus: The intention was to provide diverse test data with
* some users with valid keys,
* some users with invalid keys and
* some users without any key at all.
So should we just provide data for the first case?

marcelgarus: Okay, fixed that.

evap/evaluation/management/commands/anonymize.py
```diff
@@ -76,33 +70,39 @@ def anonymize_data(self):
...
-            random_usernames.add((random.choice(first_names), random.choice(last_names)))
-        return random_usernames
-
-    @staticmethod
-    def replace_usernames_emails(random_usernames):
-        for user, new_username in zip(UserProfile.objects.all(), random_usernames):
+    # Anonymizes the email templates by replacing the actual subject with a sample one
+    # and replacing the body with lorem ipsum.
+    def anonymize_email_templates(self):
+        self.stdout.write("TODO: Anonymize email templates...")
```
karyon: same here, i would prefer to not have any stubs in the code, they have the tendency to be unused for years or be removed because the plans changed :) the next person looking into this will figure out what to do by themselves.

evap/evaluation/management/commands/anonymize.py
```diff
@@ -111,22 +111,109 @@ def replace_usernames_emails(random_usernames):
                 new_domain = Command.new_institution_domain if is_institution_domain else Command.new_external_domain
                 user.email = user.username + '@' + new_domain
 
+            if user.login_key is not None:
+                # Create a new login key.
+                user.login_key = None
+                user.valid_until = None
+                user.ensure_valid_login_key()
+                # Also create some invalid keys.
+                user.valid_until = date.today() + random.choice([1, -1]) * timedelta(356 * 100)
```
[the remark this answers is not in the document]
marcelgarus: What exactly is wrong with the code?
Some keys' valid until date is set 100 years into the future, the other keys' date to the past.

evap/evaluation/management/commands/anonymize.py
```diff
@@ -111,22 +111,109 @@ def replace_usernames_emails(random_usernames):
...
+
             user.save()
 
-    @staticmethod
-    def shuffle_courses():
-        # do this per semester to avoid problems e.g. with archived semesters
+    # Comprehensively anonymizes courses.
+    def anonymize_courses(self):
+        all_degrees = Degree.objects.all()
+        all_course_types = CourseType.objects.all()
```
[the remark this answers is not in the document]
marcelgarus: By "take the data from the semester" you mean choosing the amount of Bachelor, Master etc. degrees based on how many of those degrees were in the semester before?

evap/evaluation/management/commands/anonymize.py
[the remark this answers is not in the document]
marcelgarus: Done. Year length fixed

marcelgarus:
All the requested changes were implemented. This is ready to merge.

karyon:
@janno42 is waiting for #1294 since it'll be a pain to resolve merge conflicts in that one. we agreed that he'll adjust and merge your PR as soon as #1294 is merged. thanks for your contribution!

evap/evaluation/management/commands/anonymize.py
```diff
@@ -205,8 +216,36 @@ def anonymize_questionnaires(self, lorem_ipsum):
...
+                    )
+                    answers = [rating_answer_counter.answer for rating_answer_counter in rating_answer_counters]
+                    for answer in answers:
+                        counter = RatingAnswerCounter.objects.get(
+                            contribution=contribution,
+                            question=question,
+                            answer=answer
+                        )
+                        if answer == answers[-1]:
+                            new_count = answer_count
```
karyon: why?

evap/evaluation/management/commands/anonymize.py
```diff
@@ -205,8 +216,36 @@ def anonymize_questionnaires(self, lorem_ipsum):
...
+                        else:
+                            new_count = random.randint(0, answer_count)
```
karyon: this will result in a much larger number of answers than there was before, is that intended?

janno42: it doesn't, the answer_count is decreased by the number of "assigned" answers every time

evap/evaluation/management/commands/anonymize.py
janno42: the last option gets all "remaining" answers that have not been assigned to other options.

evap/evaluation/management/commands/anonymize.py
```diff
@@ -125,78 +130,84 @@ def anonymize_courses(self):
...
+                evaluation.save()
+
+            for i, evaluation in enumerate(evaluations):
+                if evaluation.course.evaluations.count() == 1:
+                    evaluation.name_de = ""
+                    evaluation.name_en = ""
+                elif names:
+                    name = names.pop()
+                    evaluation.name_de = name[0]
+                    evaluation.name_en = name[1]
```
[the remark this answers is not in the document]
janno42: true, an empty name would be added once, that's not what i wanted here. will be changed.

evap/evaluation/management/commands/anonymize.py
karyon: ah, right, cool. this will make the first answer have a larger value on average than later answers, and there might be some zeroes at the end, but i guess that doesn't matter.

janno42:
"i wondered whether shuffling the answer counters is really necessary but didn't come to any conclusion." - we also thought about it for a while, didn't consider it necessary but an interesting thing to do and so we added it.

karyon:
cool. this might actually take a few minutes, there are probably a few hundred thousand rating answer counters on production, which produce one update query each.

evap/evaluation/management/commands/anonymize.py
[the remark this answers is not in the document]
He3lixxx: resolved in line 180

janno42:
we thought randomizing answer counters sounded like a good thing to do, so we added it yesterday. then we thought it should be removed if it makes things slower and we don't really need it, so we removed it again. then @karyon and i discussed that it is better to randomize answers because there might be the following case: users who have access to production might find unpublished courses in the anonymized data and can find results there based on answer distribution of other courses and de-anonymizing contributor's names. this is a very special case and there wouldn't be much data available as a course is only then not published if less than two participants voted for it - but nevertheless there might be data leaked.
so we decided that answer counters should be set randomly on anonymization. the answer count per question should (roughly) stay the same during the process and it must be made sure that the answer count never exceeds the number of voters in the course.

He3lixxx:
Uhh, i messed this up (tried to push my branch to marcelgarus/master, it did push my master branch) :/ Will try to fix...

He3lixxx:
Merged manually in https://github.com/fsr-de/EvaP/commit/f0c763548496f1cfdb25f1ae6cb54831d86b855d

PR #1288, opened 2018-12-07
Importer fixes

janno42:
- enrollment importer tests do actually what they are expected to do
- courses where the german name is the same but only the english name differs don't throw an internal server error on import but are correctly identified
- the card with importer warnings about many enrollments is created with valid html and can be collapsed


evap/staff/tests/test_importers.py
```diff
@@ -175,30 +175,38 @@ def test_enrollment_importer_high_enrollment_warning(self):
...
         with open(self.filename_invalid, "rb") as excel_file:
             excel_content = excel_file.read()
 
-        __, __, __, errors_test = UserImporter.process(excel_content, test_run=True)
-        __, __, __, errors_no_test = UserImporter.process(excel_content, test_run=False)
+        __, __, errors_test = EnrollmentImporter.process(excel_content, semester, None, None, test_run=True)
+        __, __, errors_no_test = EnrollmentImporter.process(excel_content, semester, None, None, test_run=False)
 
         self.assertEqual(errors_test, errors_no_test)
-        self.assertIn('Sheet "Sheet1", row 2: Email address is missing.', errors_test)
```
He3lixxx: We may want to merge this with staff/test/test_views.py:573 (TestSemesterImportView:test_error_handling) since they are now basically testing the same thing? Would require some kind of import mocking though, don't know if it would make the code any easier.

janno42: i didn't even know the other tests exist. yes, we should somehow deduplicate this and i'm open for suggestions. maybe it would be sufficient for the view tests to assert that some error is shown without checking the actual text while the importer tests make sure the correct errors are returned from the importer?

evap/staff/importers.py
```diff
@@ -271,10 +271,13 @@ def read_one_enrollment(self, data):
     def process_course(self, course_data, sheet, row):
         course_id = course_data.name_en
         if course_id not in self.courses:
-            self.courses[course_id] = course_data
+            if any(course.name_de == course_data.name_de for course in self.courses.values()):
```
karyon: that's O(enrolments * courses), can you perhaps put the german names into a set?

janno42:
the generated html was broken. so keep the commit message or change it anyway? and what about the tests - should we open an issue?

karyon:
sure, keep the commit message.

regarding the tests... i think both of them are valid. perhaps the view test should actually mock the importer, but i don't really mind. feel free to open an issue if you want, but I'd merge this PR here regardless.

PR #1291, opened 2018-12-10
Update to django2.1. This will break one migration in sqlite, we can not

### LINKED ISSUE #1244: Django 2.1 upgrade
https://docs.djangoproject.com/en/2.1/releases/2.1/

There is a guide for upgrading: https://docs.djangoproject.com/en/2.1/howto/upgrade-version/

Basically
* Read the release notes
* update dependencies
* run tests with `python -Wa` and solve deprecation warnings
* put the new django into the requirements
* run tests, fix failures if any
* run tests with `python -Wa` and solve deprecation warnings again
* if there was any new feature in the release notes that might help us, use it

also, we need to check the installed python version on production, django 2.1 supports python 3.5 and newer.

He3lixxx:
get it to work right now. Fixes #1244

### DIFF
--- a/evap/evaluation/migrations/0062_replace_textanswer_id_with_uuid.py
+++ b/evap/evaluation/migrations/0062_replace_textanswer_id_with_uuid.py
@@ -33,13 +33,12 @@ class Migration(migrations.Migration):
             name='uuid',
             field=models.UUIDField(primary_key=False, default=uuid.uuid4, serialize=False, editable=False),
         ),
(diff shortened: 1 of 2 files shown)


He3lixxx:
Looks like everything works fine. We need to tell travis to use python >=3.5, then we can merge this.

He3lixxx:
Tests pass, I manually clicked through the page for a few minutes and didn't notice anything strange. @janno42, can I merge or do you have any objections?

PR #1293, opened 2018-12-31
Freeze redis version

janno42:
freeze redis version because 3.0 causes troubles (`manage.py refresh_results_cache` on `vagrant up` already throws errors about the `CacheKey` being invalid)


janno42:
using newest versions

janno42:
of course

Issue #1047, opened 2017-12-11
Multiple evaluations for one course


He3lixxx:
For each course, we should be able to have multiple evaluations. This feature should be done after the results rework (#571) because it's not possible with the current UI.

Use cases:
- Evaluation during the semester so the lecturer can directly react to feedback
- Evaluation of the exam after the main evaluation finished so the lecturers can get feedback on the exam. This required knowledge of the exam dates, which we don't currently have

janno42:
see also https://github.com/fsr-itse/EvaP/wiki/Feature-Ideas

janno42:
The proposed approach is as follows:
Current `Course`s will be split into `Course`s which will only hold meta-information about the actual course (like name, degree, participants, ...) and `Evaluation`s which hold the course questionnaires and contributions as well as the evaluation periods. Each `Course` can have multiple `Evaluation`s.

Most views for staff, contributors and students will have to be updated. Only staff will be able to edit `Course` information, editors will only be able to change `Evaluation` data during their review.
Results pages will be per `Evaluation`. On an `Evaluation`'s result page other results pages belonging to the same `Course` should be linked (e.g., by adding navigation arrows on the left and right side of the page).

We still have to decide how multiple `Evaluation`s of one `Course` will be displayed on the staff, contributor and student index pages and on the results overview page.

janno42:
done in #1300

PR #1294, opened 2019-01-02
Evaluations per course

janno42:
this is a huge preparation step for #1047. it basically changes nothing apart from decoupling: the thing students can vote for is now called an `Evaluation` and each `Evaluation` references exactly one `Course` with some basic information. vice versa each `Course` references exactly one `Evaluation`.

- the first commit removes all admin pages. those weren't used for several years now and it's one thing less to struggle with.
- the second commit is a "simple" renaming from the previous "Course" to "Evaluation"
- the third commit then re-introduces a `Course` but this time it only holds part of the data it had before. it currently references exactly one `Evaluation` which holds the rest of the data. this is the basis for further changes which will allow a `Course` to have multiple `Evaluations`.

not necessarily required for #1047 be we discussed that it's logical to do it:
- the fourth commit moves responsibilities from `Contribution` to `Course`

i'm sorry for touching almost all relevant files in the project. all automated tests pass and i manually tested a lot of things in the ui but i can't guarantee that i made no mistakes.

any remarks, wishes and suggestions are welcome :)


janno42:
i'm not sure why loading the test data fails. i didn't make any changes related to groups and `reload_testdata` succeeds.

evap/evaluation/migrations/0092_add_course.py
```diff
@@ -0,0 +1,60 @@
...
+                ('last_modified_user', models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='courses_last_modified+', to=settings.AUTH_USER_MODEL)),
+                ('semester', models.ForeignKey(on_delete=django.db.models.deletion.PROTECT, related_name='courses', to='evaluation.Semester', verbose_name='semester')),
+                ('type', models.ForeignKey(on_delete=django.db.models.deletion.PROTECT, related_name='courses', to='evaluation.CourseType', verbose_name='course type')),
+            ],
+            options={
+                'verbose_name': 'course',
+                'verbose_name_plural': 'courses',
+                'ordering': ('name_de',),
+                # this is required to prevent database errors about already existing relations and will be changed back in migration 0094
+                'db_table': 'evaluation_course_temp',
```
karyon: can you add in the comment a source that explains this in more detail?

evap/evaluation/models.py
```diff
@@ -1275,16 +1269,15 @@ def is_student(self):
 
     @property
     def is_contributor(self):
-        return self.contributions.exists()
+        return self.contributions.exists() or self.is_responsible
```
He3lixxx: Open discussion: Do we want this or do we want the definition like "Contributors are all the persons that have contributions".

karyon:
for the record, the `load_test_data` was failing because the execution order of the migrations changed with this PR. The migration adding the grade publisher group was now executed before the migration adding the reviewer group, leading to pks in the db which did not match the pks in the json.

evap/evaluation/models.py
```diff
@@ -240,37 +245,93 @@ def __str__(self):
...
+
+    def __str__(self):
+        return self.name
+
+    def set_last_modified(self, modifying_user):
+        self.last_modified_user = modifying_user
+        self.last_modified_time = timezone.now()
+        logger.info('Course "{}" (id {}) was edited by user {}.'.format(self, self.id, modifying_user.username))
+
+    def can_user_see_course(self, user):
```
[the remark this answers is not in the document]
janno42: sure. we have inconsistent naming for the `can_see` stuff anyway. do you want to open an issue?

evap/staff/forms.py
```diff
@@ -291,7 +298,7 @@ def __init__(self, *args, **kwargs):
             disable_all_fields(self)
 
         if self.instance.pk:
-            responsible = self.instance.responsible_contributors[0]
+            responsible = self.instance.course.responsibles.first()
             self.fields['responsible'].queryset |= UserProfile.objects.filter(pk=responsible.pk)
             self.fields['responsible'].initial = responsible
```
He3lixxx: Can a single result have multiple responsibles?

evap/staff/forms.py
```diff
@@ -336,14 +343,17 @@ def save(self, *args, **kw):
         single_result_questionnaire = Questionnaire.single_result_questionnaire()
         single_result_question = single_result_questionnaire.questions.first()
 
-        contribution, created = Contribution.objects.get_or_create(evaluation=self.instance, responsible=True, can_edit=True, textanswer_visibility=Contribution.GENERAL_TEXTANSWERS)
+        contribution, created = Contribution.objects.get_or_create(evaluation=self.instance, can_edit=True, textanswer_visibility=Contribution.GENERAL_TEXTANSWERS)
         contribution.contributor = self.cleaned_data['responsible']
```
He3lixxx: So for single results, we will have a contribution where the responsible is the contributor. Is this what you intended?

evap/staff/importers.py
```diff
@@ -92,6 +94,8 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
             semester=semester,
         )
         course.save()
```
He3lixxx: Why two times course.save()?

evap/staff/importers.py
```diff
@@ -100,9 +104,7 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
             course=course,
         )
         evaluation.save()
-        # This is safe because the user's email address is checked before in the importer (see #953)
-        responsible_dbobj = UserProfile.objects.get(email=self.responsible_email)
-        evaluation.contributions.create(contributor=responsible_dbobj, evaluation=evaluation, responsible=True, can_edit=True, textanswer_visibility=Contribution.GENERAL_TEXTANSWERS)
+        evaluation.contributions.create(contributor=responsible_dbobj, evaluation=evaluation, can_edit=True, textanswer_visibility=Contribution.GENERAL_TEXTANSWERS)
```
He3lixxx: Same as for the single results here - we are unconditionally generating a contribution where the responsible is the contributor. Why?

evap/staff/templates/staff_user_merge.html
```diff
@@ -143,6 +143,42 @@ <h3>{% trans 'Merge users' %}</h3>
...
+                                <ul>
+                                    {% for course in courses.list %}
+                                        <li>{{ course.name }}</li>
+                                    {% endfor %}
+                                </ul>
+                            {% endfor %}
+                        </td>
+                        {% regroup merged_user.courses_responsible_for by semester as course_list %}
+                        <td{% if course_list|length > 0 %} class="success"{% endif %}>
+                            {% for courses in course_list %}
```
He3lixxx: we have this duplicated here two times - maybe include from another file?

evap/staff/tools.py
```diff
@@ -125,6 +125,8 @@ def merge_users(main_user, other_user, preview=False):
 
     errors = []
     warnings = []
+    if any(course in main_user.get_sorted_courses_responsible_for() for course in other_user.get_sorted_courses_responsible_for()):
```
He3lixxx: is main_user.get_sorted_courses_responsible_for() called n times here?

evap/staff/forms.py
janno42: yes, every course can have multiple responsibles, we make no restrictions for single results.

janno42: ah, no i see what you mean. this form here saves only one responsible person, because it is behaving like the form we had before. it drives me crazy because it will be changed in the next commit again - probably going for the 1:1 relation first and then making this 1:n wasn't the best way of doing this.
so once courses will be edited independent from evaluations (which will happen in the mentioned next commit) the course editing is separated from evaluations or single results and thus multiple responsibles can be set. not sure whether changing that here makes any sense at all.

evap/grades/migrations/0015_add_grade_documents_on_course.py
```diff
@@ -0,0 +1,32 @@
...
+        ('grades', '0014_rename_course_to_evaluation'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='gradedocument',
+            name='course',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.PROTECT, related_name='grade_documents', to='evaluation.Course', verbose_name='course'),
+        ),
+        migrations.AlterField(
```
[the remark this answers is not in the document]
janno42: it's not necessary - will be removed.

evap/evaluation/migrations/0092_add_course.py
```diff
@@ -0,0 +1,60 @@
...
+            },
+        ),
+        migrations.AddField(
+            model_name='evaluation',
+            name='course',
+            field=models.OneToOneField(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, related_name='evaluation', to='evaluation.Course', verbose_name='course'),
```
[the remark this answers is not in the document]
janno42: yes. migration 0094 then alters this to `field=models.OneToOneField(on_delete=django.db.models.deletion.CASCADE, related_name='evaluation', to='evaluation.Course', verbose_name='course'),`

evap/evaluation/models.py
```diff
@@ -240,37 +245,93 @@ def __str__(self):
...
+            return self.is_user_contributor_or_delegate(user) or self.evaluations.filter(participants__in=user).exists()
+        return True
+
+    @property
+    def can_manager_edit(self):
+        return not self.semester.participations_are_archived
+
+    @property
+    def can_manager_delete(self):
+        return self.evaluation.can_manager_delete
```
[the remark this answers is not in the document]
janno42: no. will be removed (and re-introduced in the next commit).

evap/grades/migrations/0015_add_grade_documents_on_course.py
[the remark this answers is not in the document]
janno42: it is required for the backward migration - when reverting migration 0015 "evaluation" must be able to hold null values before it's filled with actual data.

evap/grades/migrations/0015_add_grade_documents_on_course.py
```diff
@@ -0,0 +1,32 @@
...
+            model_name='gradedocument',
+            name='evaluation',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.PROTECT, related_name='grade_documents', to='evaluation.Evaluation', verbose_name='evaluation'),
+        ),
+        migrations.AlterUniqueTogether(
+            name='gradedocument',
+            unique_together=set(),
```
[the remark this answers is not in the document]
janno42: removing the unique constraint. which actually seems not to be necessary here, so i'll remove it.

evap/staff/importers.py
janno42: the course instance must have an id before setting the many to many relation works.

evap/evaluation/models.py
```diff
@@ -267,6 +266,9 @@ class Course(models.Model):
     # defines whether results can only be seen by contributors and participants
     is_private = models.BooleanField(verbose_name=_("is private"), default=False)
 
+    # persons responsible for the course; their names will be shown next to course, they can edit the course and see general text answers
+    responsibles = models.ManyToManyField(settings.AUTH_USER_MODEL, verbose_name=_("responsibles"), related_name="courses_responsible_for")
```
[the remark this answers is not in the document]
janno42: we decided that's okay.

evap/staff/forms.py
janno42: yes, that's what we had so far. as it doesn't make too much sense now that we don't need a contribution to have a responsible, this will be changed. i already have this prepared somewhere and will add it to this commit.

evap/staff/importers.py
janno42: here it makes sense - we assume the responsible person is also contributing to the course, so we can add a contribution during import. in case these people do not actually contribute and shall not be part of the evaluation, they will be removed later by managers or editors. but that's not usually the case.

evap/evaluation/tests/test_models.py
```diff
@@ -154,26 +154,11 @@ def test_deleting_last_modified_user_does_not_delete_evaluation(self):
         user.delete()
         self.assertTrue(Evaluation.objects.filter(pk=evaluation.pk).exists())
 
-    def test_responsible_contributors_ordering(self):
```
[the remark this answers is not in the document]
janno42: we have `contributor.tests.test_forms.ContributionFormsetWebTests.test_form_ordering`. do you want something in addition?

evap/evaluation/models.py
```diff
@@ -1396,12 +1392,12 @@ class EmailTemplate(models.Model):
...
-        elif cls.EDITORS in recipient_groups:
-            recipients += UserProfile.objects.filter(contributions__evaluation=evaluation, contributions__can_edit=True)
-        elif cls.RESPONSIBLE in recipient_groups:
-            recipients += evaluation.responsible_contributors
+        if cls.CONTRIBUTORS in recipient_groups or cls.EDITORS in recipient_groups or cls.RESPONSIBLE in recipient_groups:
+            recipients += evaluation.course.responsibles.all()
+            if cls.CONTRIBUTORS in recipient_groups:
+                recipients += UserProfile.objects.filter(contributions__evaluation=evaluation)
+            elif cls.EDITORS in recipient_groups:
+                recipients += UserProfile.objects.filter(contributions__evaluation=evaluation, contributions__can_edit=True)
```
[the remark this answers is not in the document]
janno42: i'm now using a set all the way through - please check.

evap/evaluation/templates/responsibility_buttons.html
```diff
@@ -2,17 +2,11 @@
...
             {% trans 'Able to edit the evaluation for preparation.' as tooltip %}
         {% elif choice.data.value == 'CONTRIBUTOR' %}
             {% trans 'Default value. No special rights.' as tooltip %}
         {% endif %}
-        {% if choice.data.value == 'RESPONSIBLE' and not can_change_responsible %}
-            {% include 'choice_button.html' with formelement=form.responsibility choice=choice tooltip=tooltip disabled=True %}
-        {% else %}
-            {% include 'choice_button.html' with formelement=form.responsibility choice=choice tooltip=tooltip disabled=False %}
-        {% endif %}
+        {% include 'choice_button.html' with formelement=form.responsibility choice=choice tooltip=tooltip %}
```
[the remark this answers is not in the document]
janno42: did remove.

evap/evaluation/templatetags/evaluation_filters.py
```diff
@@ -120,4 +120,6 @@ def hours_and_minutes(time_left_for_evaluation):
 
 @register.filter
 def has_nonresponsible_editor(evaluation):
-    return evaluation.contributions.filter(responsible=False, can_edit=True).exists()
+    editors = [contribution.contributor for contribution in evaluation.contributions.filter(can_edit=True)]
+    responsibles = evaluation.course.responsibles.all()
+    return any(editor not in responsibles for editor in editors)
```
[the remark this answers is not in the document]
janno42: works perfectly.

evap/evaluation/models.py
janno42: we want an exact definition. so i'll change that.

evap/staff/templates/staff_user_merge.html
janno42: most things in this file exist three times. so that might be a refactoring idea for a different issue?

evap/staff/tools.py
janno42: no, just once because it's independent from `course`.

evap/staff/tests/test_tools.py
```diff
@@ -42,8 +42,8 @@ def setUpTestData(cls):
         cls.user1 = mommy.make(UserProfile, username="test1")
         cls.user2 = mommy.make(UserProfile, username="test2")
         cls.user3 = mommy.make(UserProfile, username="test3")
-        cls.group1 = mommy.make(Group, name="group1")
-        cls.group2 = mommy.make(Group, name="group2")
+        cls.group1 = mommy.make(Group, name="group1", pk=4)
+        cls.group2 = mommy.make(Group, name="group2", pk=5)
```
janno42: any idea why after explicitly setting pks for groups in the migrations, mommy tries to create a group with `pk=1` when not specifying a pk here? this would result in a duplicate key error when running this test.

evap/staff/forms.py
janno42: added a new commit for single result contribution changes

evap/evaluation/migrations/0092_add_course.py
janno42: i think there was no source that explicitly stated a solution for this problem. i just thought it might be helpful and it turned out to work.
what happens is that in migration 0092 we're getting this error if the table wasn't manually renamed before:
`django.db.utils.ProgrammingError: relation "evaluation_course_semester_id_name_en_9ebe41fb_uniq" already exists`
which is strange because `Course` was renamed to `Evaluation` in 0091 so i assumed everything related in the database would also have been renamed in this step. but adding a new model named `Course` then results in this error. maybe you have some additional ideas.

evap/staff/tests/test_tools.py
karyon: perhaps because django uses some database sequence for choosing the next PK and we're not updating that. bummer. so we can either specify the migration dependencies, instead, or make this test use existing groups (in which case we choose to explicitly specify the pks for all groups in the future, (which doesn't sound too bad to me))

janno42: okay, then i'll use existing groups for the test.

karyon:
Well, you could of course leave it as it is right now, since it works, I guess. But then again, creating new groups is kind of a non-use-case... Your call.

evap/evaluation/models.py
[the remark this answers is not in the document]
janno42: see #1298

evap/evaluation/tests/test_models.py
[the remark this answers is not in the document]
He3lixxx: No, I think that's fine.

evap/staff/forms.py
He3lixxx: So, did I understand you correctly:
- We already had multiple responsibles for one course
- The single result form did not respect that and this change just keeps this as it was before.
?
If that's the case, please resolve, otherwise, I'm a bit confused :D

evap/staff/templates/staff_user_merge.html
He3lixxx: Fair enough, created #1299

evap/staff/tools.py
He3lixxx: I'd be impressed if python could derive that (function call could modify the object -> next call could be different!) but I'm pretty sure it does not do that :/
```
Python 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> def test():
...     print("test called")
...     return [1, 2, 3]
... 
>>> any(number in test() for number in [4, 5, 6, 3])
test called
test called
test called
test called
True
>>> 
```

janno42: ah, i didn't pay attention and though you meant the second part :)
`other_user.get_sorted_courses_responsible_for()` is only called once.
any suggestions for changes? i actually don't care that much about runtime in user merging.

evap/staff/forms.py
janno42: yes, multiple responsibles per course were possible, but not in a single result.

evap/staff/tools.py
He3lixxx: I'd introduce a new variable and store the result of the call there, if that's ok for you. (maybe `courses_where_main_user_is_responsible`?)

janno42: done

evap/evaluation/migrations/0098_remove_contributors_from_single_results.py
```diff
@@ -0,0 +1,49 @@
...
+    Evaluation = apps.get_model('evaluation', 'Evaluation')
+    Contribution = apps.get_model('evaluation', 'Contribution')
+
+    questionnaire = Questionnaire.objects.get(name_en='Single result')
+    questionnaire.is_for_contributors = False
+    questionnaire.save()
+
+    for evaluation in Evaluation.objects.filter(is_single_result=True):
+        contribution = Contribution.objects.get(evaluation=evaluation)
+        contribution.contributor = None
```
He3lixxx: So we're not using information here since the responsible was also already stored in course.responsibles, correct?

He3lixxx:
Looks good to me for all commit up to the current (remove contributors from single results)

evap/evaluation/migrations/0098_remove_contributors_from_single_results.py
janno42: not also, but only. after migration 0096 contributions don't hold responsibility information anymore.
and yes, the person from this responsible contribution was then saved as a course responsible.

karyon:
oh and please for the next large PR, make your changes after reviews reviewable somehow, immediately squashing the changes into old commits makes it really hard to review again.

janno42:
- at first it was only small changes which i though could nicely be reviewed by looking at the force push diffs. i see that in the meantime it got quite a lot and will do it differently next time.
- coveralls threshold was updated to 0.05.
- i will look into the mentioned coverage issues.

janno42:
- `can_user_see_course` is indeed unused. everything is based on lists of evaluations so it is not needed and i removed it.
- responsibilities have been added to the user merge test.
- the validate_unique hack in the EvaluationForm was not covered, but the validation was actually not needed yet. i commented it out for later use. the test for that was already commented out :)

evap/evaluation/migrations/0096_move_responsibilities_into_course.py
```diff
@@ -0,0 +1,35 @@
+from django.db import migrations
+
+
+def move_data_to_course(apps, _schema_editor):
+    Course = apps.get_model('evaluation', 'Course')
+    Contribution = apps.get_model('evaluation', 'Contribution')
+    for course in Course.objects.all():
+        responsibles = [contribution.contributor for contribution in Contribution.objects.filter(evaluation=course.evaluation, responsible=True)]
+        course.responsibles.set(responsibles)
+        course.save()
```
He3lixxx: In this migration, you do a `modelInstance.save()` after using `modelInstance.manyToManyField.set`, in some other migrations (especially in the commits before this), you do not. I did not find any documentation, but a bit of googling suggests that it is not required. I guess we want this unified?

(For all spots to look at, just Ctrl+F for ".set(" )

evap/evaluation/models.py
```diff
@@ -606,13 +608,13 @@ def days_until_evaluation(self):
...
     def is_user_editor_or_delegate(self, user):
-        return self.contributions.filter(Q(contributor=user) | Q(contributor__in=user.represented_users.all()), can_edit=True).exists()
+        return self.contributions.filter(Q(contributor=user) | Q(contributor__in=user.represented_users.all()), can_edit=True).exists() or self.course.responsibles.filter(pk=user.pk).exists()
 
     def is_user_contributor_or_delegate(self, user):
         # early out that saves database hits since is_contributor_or_delegate is a cached_property
         if not user.is_contributor_or_delegate:
             return False
-        return self.contributions.filter(Q(contributor=user) | Q(contributor__in=user.represented_users.all())).exists()
+        return self.contributions.filter(Q(contributor=user) | Q(contributor__in=user.represented_users.all())).exists() or self.course.responsibles.filter(pk=user.pk).exists()
```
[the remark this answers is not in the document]
He3lixxx: This could only be slow if the user is a responsible or a contributor or a delegate at all, otherwise, the early-out would be taken.

This method is called if:
- Course is private or user is external and we want to determine if the user can see the evaluation (`can_user_see_evaluation`)
- Course is published but can not publish rating results or semester results are archived (`can_user_see_results_page`)
- Evaluation preview permissions check (`evaluation_preview`)
- In the evaluation detail view

For the first two, I could only find one scenario that could be slow:
In the second case, we might have multiple executions on the results_index.html template, since for each evaluation this is checked. It could be bad if we had a lot of published courses where `can_publish_rating_results` is false or results are archived for this semester - usually, that's not the case, though, right?. Still, this could be optimized by selecting the evaluations that should be shown in the view. We might want to test this.

The last two cases are just one call (one user, one evaluation). I think the query itself is rather simple, so one execution should not take that long.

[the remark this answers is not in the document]
janno42: there was one more thing missing: this method (which in the meantime is called `is_user_responsible_or_contributor_or_delegate`) should also return `True` if the user is a delegate of one of the responsibles for this course even if the responsible isn't a contributor. please have a look at the implementation in the latest commit.

evap/evaluation/models.py
```diff
@@ -589,13 +589,17 @@ def days_until_evaluation(self):
...
+        represented_user_pks = [user.pk for user in user.represented_users.all()]
+        represented_user_pks.append(user.pk)
+        return self.contributions.filter(contributor__pk__in=represented_user_pks, can_edit=True).exists() or self.course.responsibles.filter(pk__in=represented_user_pks).exists()
 
     def is_user_responsible_or_contributor_or_delegate(self, user):
         # early out that saves database hits since is_responsible_or_contributor_or_delegate is a cached_property
         if not user.is_responsible_or_contributor_or_delegate:
             return False
-        return self.contributions.filter(Q(contributor=user) | Q(contributor__in=user.represented_users.all())).exists() or self.course.responsibles.filter(pk=user.pk).exists()
+        represented_user_pks = [user.pk for user in user.represented_users.all()]
```
karyon: codacy suggests to not shadow the `user` variable, i think i agree

PR #1295, opened 2019-01-07
1289 merge degrees

### LINKED ISSUE #1289: Auto merge degrees on import
Currently the enrollment importer shows an error message if the same course has a different degree than in a previous line of the enrollment file.
Instead, it should display a warning message if a course's data differs only in its degree from the previous lines for this course and add all degrees to the imported course data.

Irliung:
Fixes #1289 and fixed issue with ExcelImporter.process not returning the error dictionaries in case of an unhandled exception.


evap/staff/importers.py
```diff
@@ -271,14 +273,19 @@ def read_one_enrollment(self, data):
...
+
         if course_id not in self.courses:
             if course_data.name_de in self.names_de:
                 self.errors.append(_('Sheet "{}", row {}: The German name for course "{}" already exists for another course.').format(sheet, row + 1, course_data.name_en))
             else:
                 self.courses[course_id] = course_data
                 self.names_de.add(course_data.name_de)
         else:
-            if not course_data == self.courses[course_id]:
+            if set(course_data.degree_names) != set(self.courses[course_id].degree_names) and \
```
He3lixxx: I'd prefer one comparison per line here, binary operators at line start. Idc about line endings (with \ or implicit with parentheses) @karyon, @janno42  what do you think?

evap/staff/importers.py
```diff
@@ -271,14 +273,19 @@ def read_one_enrollment(self, data):
...
+                course_data.name_de == self.courses[course_id].name_de and course_data.name_en == self.courses[course_id].name_en and course_data.type_name == self.courses[course_id].type_name and course_data.is_graded == self.courses[course_id].is_graded and course_data.responsible_email == self.courses[course_id].responsible_email:
+                self.warnings[self.W_DEGREE].append(_('Sheet "{}", row {}: The course\'s "{}" degree differs from it\'s degree in a previous row. Courses were automatically joined.').format(sheet, row+1, course_data.name_en))
+                self.courses[course_id].degree_names.extend(course_data.degree_names)
+            elif not course_data == self.courses[course_id]:
```
He3lixxx: I know you just copied it, but I find "not ==" strange here. @karyon @janno42 remember any reason why we did it this way? I think != would be the way to go.

He3lixxx:
@Irliung do you want to continue this or should we take over?

Irliung:
This issue should be fixed (I already committed the fix) and just needs
review before it can be marked as completed.

He3lixxx:
Did you push your latest commit (`git push`)? The changes are not (yet) reflected here.

evap/staff/tests/test_importers.py
```diff
@@ -125,43 +126,73 @@ def test_import_makes_inactive_user_active(self):
...
+        success_messages, warnings, errors = EnrollmentImporter.process(excel_content, self.semester, self.vote_start_datetime, self.vote_end_date, test_run=False)
+        self.assertIn("Successfully created 1 courses/evaluations, 2 students and 1 contributors", "".join(success_messages))
+        self.assertEqual(errors, [])
+        self.assertEqual(warnings, expected_warnings)
+
+        self.assertEqual(Course.objects.all().count(), 1)
+        self.assertEqual(Evaluation.objects.all().count(), 1)
+
+        course = Course.objects.get(name_de="Bauen")
+        self.assertEqual("Bachelor, Master", ", ".join(degree.name_de for degree in course.degrees.all().order_by("name_de")))
```
karyon: compare `set(course.degrees.all())` with `set(Degrees.objects.filter(name__in=...)` or something like that? the string assembly and comparison seems weird.

PR #1297, opened 2019-01-10
Use virtualenv

janno42:
the instructions in the wiki (https://github.com/fsr-de/EvaP/wiki/Installation) have already been updated to explain the virtualenv setup

### DIFF
--- a/deployment/apache.template.conf
+++ b/deployment/apache.template.conf
@@ -1,11 +1,11 @@
 <VirtualHost *:80>
     ServerName evap
-    
+
     DocumentRoot /var/www/
-    
-    WSGIDaemonProcess wsgi processes=2 threads=15 display-name=%{GROUP} user=vagrant
-    WSGIProcessGroup wsgi
+
     WSGIScriptAlias / /vagrant/evap/wsgi.py
+    WSGIDaemonProcess evap processes=2 threads=15 display-name=%{GROUP} user=vagrant python-home=/home/vagrant/venvs/env
+    WSGIProcessGroup evap
 
     Alias /static /vagrant/evap/static_collected
     <Location /static>
--- a/deployment/provision_vagrant_vm.sh
+++ b/deployment/provision_vagrant_vm.sh
@@ -2,7 +2,7 @@ set -x # print executed commands
 
 # install python stuff
 apt-get -q update
-apt-get -q install -y python3-dev python3-pip gettext
+apt-get -q install -y python3.7 python3.7-dev python3.7-venv python3-pip gettext
 
 # install sass
 apt-get -q install -y sassc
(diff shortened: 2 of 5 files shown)


karyon:
should the vagrant VM do the same?

janno42:
It actually could also use a virtualenv, good catch. We just have to check how to activate it automatically on start so users don't have to do that after sshing to the machine, but that shouldn't be too hard.

janno42:
unfortunately the first run of `vagrant up` requires to be run as administrator on windows to be able to create the virtualenv. i didn't find a way around that.

janno42:
fixed, works without administrator privileges

karyon:
@He3lixxx we need to talk and decide about the virtualbox 6 slowness

.gitignore
```diff
@@ -35,3 +35,6 @@ htmlcov
 
 # pip puts editable packages here
 src
+
+# virtualenv
```
He3lixxx: I think you can remove this again, right? Or do you want to keep the venv on production in this spot?

deployment/apache.template.conf
```diff
@@ -1,11 +1,11 @@
...
     ServerName evap
-    
+
     DocumentRoot /var/www/
-    
-    WSGIDaemonProcess wsgi processes=2 threads=15 display-name=%{GROUP} user=vagrant
-    WSGIProcessGroup wsgi
+
     WSGIScriptAlias / /vagrant/evap/wsgi.py
+    WSGIDaemonProcess evap processes=2 threads=15 display-name=%{GROUP} user=vagrant python-home=/vagrant/env
```
He3lixxx: Me no understand. We are not running vagrant in production, right? Why is this prepared for a venv in /vagrant/?

He3lixxx: Or is this for the development vagrant machine and should be changed to /home/vagrant/...?

janno42: i forgot to change that and didn't test the apache server in the vm.

.gitignore
janno42: right :)

karyon:
virtualbox 6.0.2 was released. the changelog doesn't look promising, but we should still try it out. if it doesn't work and the internet does not have a workaround, maybe we can redirect the logfile to the deployment directory so there's not another file in the root dir? but i don't really mind either way. it should be gitignored of course.

He3lixxx:
Issue is not solved with virtualbox 6.0.4 and vagrant 2.2.3. I suppose the following fix in the vagrantfile instead of our current logic to disable the serial port:
```
  config.vm.provider :virtualbox do |v, override|
    # disable logfile
    if Vagrant::Util::Platform.windows? then
      v.customize [ "modifyvm", :id, "--uartmode1", "file", "nul" ]
    else
      v.customize [ "modifyvm", :id, "--uartmode1", "file", "/dev/null" ]
    end

    # show virtualbox gui, uncomment this to debug startup problems
    #v.gui = true
```
I haven't tested the windows "nul" file, maybe someone with a windows installed can confirm that this indeed works? (taken from [here (nul)](https://stackoverflow.com/questions/313111/is-there-a-dev-null-on-windows) and [here (windows check)](https://stackoverflow.com/questions/26811089/vagrant-how-to-have-host-platform-specific-provisioning-steps))

The linux setup works, boots quickly enough and does not create the logfile. Everything unix-like should be fine if there is /dev/null (-> MacOS should be fine)

deployment/provision_vagrant_vm.sh
```diff
@@ -18,9 +18,18 @@ sudo -u postgres createdb -O evap evap
 # setup redis
 apt-get -q install -y redis-server
 
+# install apache
+apt-get -q install -y apache2 apache2-dev
+
+# setup virtualenv and mod_wsgi
+sudo -H -u vagrant python3.7 -m venv /home/vagrant/venvs/env
+sudo -H -u vagrant /home/vagrant/venvs/env/bin/pip install wheel
```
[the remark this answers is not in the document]
karyon: okay, please add a comment that that's the reason for installing wheel.

PR #1300, opened 2019-01-27
Multiple evaluations per course

janno42:
based on #1294

there are multiple things still missing. besides smaller bugs that's at least:
- [x] useful test data with multiple evaluations per course, different weights and so on
- [x] update results export
- [x] lots of tests
- [x] check displayed names in breadcrumbs/forms/lists/headings/emails


janno42:
some tests fail in debug mode. looks like it's always the same problem: `grade` is an empty string when the color is calculated. maybe some caching problem - do you have any advice, @karyon?

evap/evaluation/migrations/0100_clear_evaluation_names.py
```diff
@@ -0,0 +1,59 @@
...
+            evaluation = Evaluation.objects.get(pk=course.evaluations.first().pk)
+            evaluation.name_de = course.name_de
+            evaluation.name_en = course.name_en
+            evaluation.save()
+        else:
+            for i in range(0, course.evaluations.count()):
+                evaluation = Evaluation.objects.get(pk=course.evaluations.all()[i].pk)
+                evaluation.name_de = "{} ({})".format(course.name_de, i)
+                evaluation.name_en = "{} ({})".format(course.name_en, i)
+                evaluation.save()
```
He3lixxx: Why not use a object iteration, and use that for both cases (1 and many)?
```
for evaluation in course.evaluations.all():
    # set name, save
```

evap/evaluation/models.py
```diff
@@ -312,16 +316,20 @@ def midterm_grade_documents(self):
     def responsibles_names(self):
         return ", ".join([responsible.full_name for responsible in self.responsibles.all().order_by("last_name")])
 
+    @property
+    def all_evaluations_finished(self):
+        return all(evaluation.state in ['evaluationFinished', 'reviewed', 'published'] for evaluation in self.evaluations.all())
```
He3lixxx: Have you thought about writing this so we ask the database whether this is true or false? Something like `not self.evaluations.exclude(-allowed_states-).exists()`? Guess it would be faster.

evap/results/tests/test_exporters.py
```diff
@@ -108,12 +108,12 @@ def test_view_excel_file_sorted(self):
 
         # Load responses as Excel files and check for correct sorting
         workbook = xlrd.open_workbook(file_contents=content_de.read())
-        self.assertEqual(workbook.sheets()[0].row_values(0)[1], "A - Evaluation1")
-        self.assertEqual(workbook.sheets()[0].row_values(0)[2], "B - Evaluation2")
+        self.assertEqual(workbook.sheets()[0].row_values(0)[1], "A – Evaluation1")
```
He3lixxx: I already know that I will use a "-" at some point in the future. I'm sorry, in advance :laughing:

evap/staff/forms.py
```diff
@@ -193,21 +192,22 @@ class EvaluationForm(forms.ModelForm):
...
+                  'last_modified_user_name')
         localized_fields = ('vote_start_datetime', 'vote_end_date')
         field_classes = {
             'participants': UserModelMultipleChoiceField,
         }
 
     def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+        semester = kwargs.pop('semester', None)
+        super(EvaluationForm, self).__init__(*args, **kwargs)
```
He3lixxx: why the `(EvaluationForm, self)`?

evap/staff/forms.py
```diff
@@ -271,22 +269,21 @@ class SingleResultForm(forms.ModelForm):
...
     class Meta:
         model = Evaluation
-        fields = ('name_de', 'name_en', 'type', 'degrees', 'event_date', 'responsible', 'answer_1', 'answer_2', 'answer_3', 'answer_4',
+        fields = ('course', 'name_de', 'name_en', 'event_date', 'answer_1', 'answer_2', 'answer_3', 'answer_4',
                   'answer_5', 'last_modified_time_2', 'last_modified_user_2')
 
     def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+        semester = kwargs.pop('semester', None)
+        super(SingleResultForm, self).__init__(*args, **kwargs)
```
He3lixxx: why `(SingleResultForm, self)`?

evap/contributor/views.py
```diff
@@ -21,22 +23,42 @@ def index(request):
...
+            evaluation.distribution = calculate_average_distribution(evaluation)
+            evaluation.avg_grade = distribution_to_grade(evaluation.distribution)
+        else:
+            evaluation.single_result_rating_result = get_single_result_rating_result(evaluation)
+        if evaluation.course.evaluations.count() > 1:
+            if any(evaluation.state != "published" for evaluation in evaluation.course.evaluations.all()):
+                evaluation.course.not_all_evaluations_are_published = True
+            evaluation.course.distribution = calculate_average_course_distribution(evaluation.course)
+            evaluation.course.avg_grade = distribution_to_grade(evaluation.course.distribution)
+            evaluation.course_weight_sum = sum(evaluation.weight for evaluation in evaluation.course.evaluations.all())
```
He3lixxx: We're having some kind of independent logic to determine what can be seen by this contributor and to calculate grades here. I think we should rather try to rely on the functions we already have that test whether people can see stuff or not - otherwise, semantics may drift apart. What do you think? (This might be something for another issue.)

evap/evaluation/models.py
```diff
@@ -676,19 +692,13 @@ def update_evaluations(cls):
 @receiver(post_transition, sender=Evaluation)
 def warmup_cache_on_publish(instance, target, **_kwargs):
```
He3lixxx: I like the way this is handled now, but `warmup_cache_on_publish` and `delete_cache_on_unpublish` do not actually do what you would think from their names now. Also, comments indicating where the actual cache generation / invalidation happens might be good.

evap/results/tools.py
```diff
@@ -202,6 +202,19 @@ def average_non_grade_rating_questions_distribution(results):
     )
 
 
+def calculate_average_course_distribution(course):
+    if not all(evaluation.state == 'published' for evaluation in course.evaluations.all()):
```
He3lixxx: "if any not equal" instead of "if not all equal"?
Also, maybe delegate to database again?

evap/results/views.py
```diff
@@ -27,24 +32,55 @@ def delete_template_cache(evaluation):
 
 
 def _delete_template_cache_impl(evaluation):
+    _delete_evaluation_template_cache_impl(evaluation)
+    if evaluation.course.evaluations.count() > 1:
```
He3lixxx: why only delete the course template cache when there are more evaluations?

evap/results/views.py
```diff
@@ -27,24 +32,55 @@ def delete_template_cache(evaluation):
...
+    courses_to_render = set([evaluation.course for evaluation in evaluations if evaluation.course.evaluations.count() > 1])
+    course_avg_grades = {}
+    course_weight_sums = {}
     try:
+        for course in courses_to_render:
+            course.distribution = calculate_average_course_distribution(course)
+            course.avg_grade = distribution_to_grade(course.distribution)
+            course_avg_grades[course.pk] = course.avg_grade
+            course_weight_sums[course.pk] = sum(evaluation.weight for evaluation in course.evaluations.all())
+            if any(evaluation.state != "published" for evaluation in course.evaluations.all()):
```
He3lixxx: If you're repeatedly using "course.evaluations.all()", this is generating a new QuerySet each time, so the query is not cached, right? I guess we want the caching here?

evap/results/views.py
```diff
@@ -92,8 +138,21 @@ def index(request):
...
+                course__semester__in=semesters,
+                state__in=['in_evaluation', 'evaluated', 'reviewed']
+            )
+        )
+        for evaluation in additional_evaluations:
+            # the value is required in the template if the evaluation's course has multiple evaluations.
+            # as this evaluation isn't published, the course can have no average grade yet.
+            evaluation.course_avg_grade = None
+            evaluation.course_weight_sum = sum(evaluation.weight for evaluation in evaluation.course.evaluations.all())
+            evaluation.course.not_all_evaluations_are_published = True
```
He3lixxx: Does this not require a .save()? Same exists in some other places, too.

evap/results/views.py
```diff
@@ -179,15 +238,31 @@ def evaluation_detail(request, semester_id, evaluation_id):
...
+            course_evaluations += evaluation.course.evaluations.filter(state__in=['in_evaluation', 'evaluated', 'reviewed'])
+        course_weight_sum = sum(evaluation.weight for evaluation in course_evaluations)
+        for course_evaluation in course_evaluations:
+            course_evaluation.course_weight_sum = course_weight_sum
+            if course_evaluation.is_single_result:
+                course_evaluation.single_result_rating_result = get_single_result_rating_result(course_evaluation)
+            else:
+                course_evaluation.distribution = calculate_average_distribution(course_evaluation)
+                course_evaluation.avg_grade = distribution_to_grade(course_evaluation.distribution)
+            all_evaluations_of_course.append(course_evaluation)
```
He3lixxx: This is one more place where grade calculation is happening now, I feel like we want this in one central method somehow.

evap/staff/forms.py
```diff
@@ -227,6 +228,9 @@ def __init__(self, *args, **kwargs):
         if self.instance.pk and not self.instance.can_manager_edit:
             disable_all_fields(self)
 
+        if self.instance.pk:
+            self.instance.old_course = self.instance.course
```
He3lixxx: Attaching additional information like this works, the attached information will still be there when the form is constructed again from request.POST? I'd be really surprised.

evap/student/views.py
```diff
@@ -10,38 +10,41 @@
...
+            evaluation.single_result_rating_result = get_single_result_rating_result(evaluation)
+        evaluation.participates_in = request.user in evaluation.participants.all()
+        evaluation.voted_for = request.user in evaluation.voters.all()
+        if evaluation.course.evaluations.count() > 1:
+            if any(evaluation.state != "published" for evaluation in evaluation.course.evaluations.all()):
+                evaluation.course.not_all_evaluations_are_published = True
+            evaluation.course.distribution = calculate_average_course_distribution(evaluation.course)
+            evaluation.course.avg_grade = distribution_to_grade(evaluation.course.distribution)
+            evaluation.course_weight_sum = sum(evaluation.weight for evaluation in evaluation.course.evaluations.all())
+    evaluations.sort(key=lambda evaluation: evaluation.full_name)  # evaluations must be sorted for regrouping them in the template
```
He3lixxx: Feels like duplication here again.

evap/contributor/forms.py
```diff
@@ -17,14 +17,25 @@
 class EvaluationForm(forms.ModelForm):
     general_questionnaires = forms.ModelMultipleChoiceField(queryset=None, widget=CheckboxSelectMultiple, label=_("General questionnaires"))
     course = forms.ModelChoiceField(Course.objects.all(), disabled=True, required=False, widget=forms.HiddenInput())
+    name_de_ = forms.CharField(label=_("Name (German)"), disabled=True, required=False)
+    name_en_ = forms.CharField(label=_("Name (English)"), disabled=True, required=False)
 
     class Meta:
         model = Evaluation
-        fields = ('name_de', 'name_en', 'vote_start_datetime', 'vote_end_date', 'general_questionnaires', 'course')
+        fields = ('name_de_', 'name_en_', 'vote_start_datetime', 'vote_end_date', 'general_questionnaires', 'course')
```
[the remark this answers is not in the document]
janno42: hm. i thought we were using the same for fields like `last_modified_*_` when they were not actual form fields but rather used for displaying related information, but i can't find any occurrence. i don't know where this thought came from. no problem to change that.

evap/contributor/views.py
janno42: i will change it for this pr. we should have a (single) clear way of determining the visibility, otherwise it will be very hard to maintain. so i'll think about a more centralized way.

evap/results/views.py
janno42: the thought was that there is none if the course has only one evaluation but probably just trying to delete it anyway although it might not exist is faster so we don't need to count that first.

evap/staff/forms.py
janno42: uhm. very good question that i can't answer right now. i don't remember why i put it there - tests seem to be okay when removing it again - so i'll look into it.

evap/results/views.py
janno42: the values are not stored in the db and are just passed on with this current instance

evap/results/views.py
janno42: true. similar to the visibility checks this should somehow be centralized. there could be a method filtering a list of evaluations to those a certain user can see and adding the appropriate grading infos to them or something like that.

evap/staff/forms.py
janno42: i'm actually a bit confused, but the `if hasattr(self.instance, 'old_course'):` in line 269 is doing stuff when the course changed so yeah. magic.
this was influenced by https://stackoverflow.com/questions/23361057/django-comparing-old-and-new-field-value-before-saving

He3lixxx: Impressive. Using [changed_data](https://docs.djangoproject.com/en/2.1/ref/forms/api/#checking-which-form-data-has-changed) was not possible here? If it was, I would really prefer that.

janno42: changed_data returns only the name of fields that changed. i need to know what the old value was.

He3lixxx: Hmm make sense. In this case, I'd suggest to add a test that this works, then I'm fine with it (or do we already have one? I just quickly checked and did not find anything)

janno42: I didn't add any tests yet

evap/contributor/views.py
janno42: please check again. grade calculation is now done in the results tools for all views.
collecting the evaluations which can be seen depends on the view (contributor index, student index, results index) so probably we'll keep having that in three places. what do you think?

evap/student/views.py
janno42: only the grade calculation or the sorting as well? i now moved the grade calculation.

janno42:
from the travis log:
```
$ coveralls\
No command 'coveralls ' found, did you mean:
 Command 'coveralls' from package 'ruby-coveralls' (universe)
coveralls : command not found
```

janno42:
some teasers:
teaser1

teaser2

teaser3

evap/staff/tests/test_forms.py
```diff
@@ -597,3 +594,31 @@ def test_voter_cannot_be_removed_from_evaluation(self):
...
+        self.assertEqual(Evaluation.objects.get(pk=evaluation.pk).course, course1)
+        self.assertEqual(delete_call.call_count, 0)
+        self.assertEqual(warmup_call.call_count, 0)
+
+        # change course and save
+        form_data = get_form_data_from_instance(EvaluationForm, evaluation)
+        form_data["course"] = course2.pk
+        form = EvaluationForm(form_data, instance=evaluation, semester=semester)
+        self.assertTrue(form.is_valid())
+        with patch('evap.results.views._delete_course_template_cache_impl') as delete_call, patch('evap.results.views.warm_up_template_cache') as warmup_call:
```
He3lixxx: Unify this with the patch above to prevent the first assertions just being true due to typing errors, please.

He3lixxx:
Code looks good to me when my last two annotations have been resolved. Will manually test a bit.

He3lixxx:
Looks really nice! :) Two small things:
- Staff semester index: "Click to add contributors" is a link to a page that does not only add contributors, but also participants and export login keys. Maybe re-text?
- `fa_caret_right` gives me the feeling that there is something to expand. Maybe we can use something else as bullet points? (results index, results evaluation, contributor index, student index, ...). Doesn't bother me too much, though, so if you want the caret, I'll be happy too.

evap/results/exporters.py
```diff
@@ -175,11 +184,43 @@ def export(self, response, course_types_list, include_not_enough_voters=False, i
...
+                        writec(self, evaluation.course.avg_grade, self.grade_to_style(evaluation.course.avg_grade))
+                    else:
+                        self.write_empty_cell()
+
+                writen(self)
+                for evaluation, results in evaluations_with_results:
+                    if evaluation.course.evaluations.count() > 1:
+                        writec(self, None, "border_top")
+                    else:
+                        self.write_empty_cell()
```
karyon: i wonder whether we should simply not show individual evaluations but only whole courses in the export. i mean it's too much data for humans anyway and now it's even more

evap/results/exporters.py
```diff
@@ -175,11 +184,43 @@ def export(self, response, course_types_list, include_not_enough_voters=False, i
...
+                writen(self, _("Course Grade"), "bold")
+                for evaluation, results in evaluations_with_results:
+                    if evaluation.course.evaluations.count() > 1:
+                        writec(self, evaluation.course.avg_grade, self.grade_to_style(evaluation.course.avg_grade))
+                    else:
+                        self.write_empty_cell()
+
+                writen(self)
+                for evaluation, results in evaluations_with_results:
+                    if evaluation.course.evaluations.count() > 1:
```
karyon: that query is done five times for each evaluation. you can leave that if you want, i mean it's the export and we don't really care how fast it is, but if it does get too slow I won't be the one optimizing queries again.

evap/results/views.py
```diff
@@ -93,8 +126,22 @@ def index(request):
...
+                course__semester__in=semesters,
+                state__in=['in_evaluation', 'evaluated', 'reviewed']
+            )
+        )
+        for evaluation in additional_evaluations:
+            # these values are required in the template if the evaluation's course has multiple evaluations.
+            # as this evaluation isn't published, no cache entries exist and we define them manually.
+            evaluation.course.evaluation_count = evaluation.course.evaluations.count()
+            evaluation.course.avg_grade = None
+            evaluation.course.evaluation_weight_sum = evaluation.course.evaluations.all().aggregate(Sum('weight'))["weight__sum"]
```
karyon: why not use `get_evaluations_with_course_result_attributes` here?

evap/evaluation/models.py
```diff
@@ -312,16 +316,20 @@ def midterm_grade_documents(self):
...
 
     state = FSMField(default='new', protected=True)
 
-    course = models.OneToOneField(Course, models.CASCADE, verbose_name=_("course"), related_name="evaluation")
+    course = models.ForeignKey(Course, models.PROTECT, verbose_name=_("course"), related_name="evaluations")
 
-    name_de = models.CharField(max_length=1024, verbose_name=_("name (german)"))
-    name_en = models.CharField(max_length=1024, verbose_name=_("name (english)"))
+    name_de = models.CharField(max_length=1024, verbose_name=_("name (german)"), blank=True)
+    name_en = models.CharField(max_length=1024, verbose_name=_("name (english)"), blank=True)
```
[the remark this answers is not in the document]
janno42: they will be almost always blank. if there is just one evaluation in a course there is no need to name it. just for those courses where we have additional evaluations they will get a name (e.g. "exam") - the "main" evaluation will probably still have no name.

evap/evaluation/migrations/0100_clear_evaluation_names.py
```diff
@@ -0,0 +1,59 @@
...
+from django.db import migrations, models
+
+
+def clear_evaluation_names(apps, _schema_editor):
+    Evaluation = apps.get_model('evaluation', 'Evaluation')
+    for evaluation in Evaluation.objects.all():
+        if evaluation.course.evaluations.count() == 1:
+            evaluation.name_de = ""
+            evaluation.name_en = ""
+            evaluation.save()
```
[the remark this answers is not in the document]
janno42: if the evaluation is the only one for a course it doesn't need an individual name

evap/evaluation/models.py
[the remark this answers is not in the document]
karyon: makes sense. could you add a comment?

evap/staff/forms.py
```diff
@@ -271,22 +276,21 @@ class SingleResultForm(forms.ModelForm):
...
     class Meta:
         model = Evaluation
-        fields = ('name_de', 'name_en', 'type', 'degrees', 'event_date', 'responsible', 'answer_1', 'answer_2', 'answer_3', 'answer_4',
+        fields = ('course', 'name_de', 'name_en', 'event_date', 'answer_1', 'answer_2', 'answer_3', 'answer_4',
                   'answer_5', 'last_modified_time_2', 'last_modified_user_2')
 
     def __init__(self, *args, **kwargs):
+        semester = kwargs.pop('semester', None)
         super().__init__(*args, **kwargs)
+        self.fields['course'].queryset = Course.objects.filter(semester=semester)
```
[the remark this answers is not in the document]
janno42: yes, multiple single results can be part of a course. i told you i would go for this as the least restricting option.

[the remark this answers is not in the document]
karyon: hmkay i must have forgotten that :)

evap/contributor/views.py
```diff
@@ -21,22 +23,40 @@ def index(request):
...
+            )
+        )
+        delegated_evaluations = [evaluation for course in delegated_courses for evaluation in course.evaluations.all() if evaluation.state != "new" and evaluation.can_user_see_evaluation(user)]
         for evaluation in delegated_evaluations:
             evaluation.delegated_evaluation = True
-        displayed_evaluations += list(delegated_evaluations)
-    displayed_evaluations.sort(key=lambda evaluation: list(STATES_ORDERED.keys()).index(evaluation.state))
-
-    delegate_selection_form = DelegateSelectionForm()
+        displayed_evaluations.update(delegated_evaluations)
```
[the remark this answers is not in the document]
janno42: will check the overwriting.
i'm not sure what you mean by the second part. one cannot contribute to a course, only to an evaluation. if users have own courses in which they aren't contributing to any evaluation they must be responsible so there is no delegation.

evap/results/tools.py
```diff
@@ -202,6 +202,30 @@ def average_non_grade_rating_questions_distribution(results):
...
+
+
+def get_evaluations_with_course_result_attributes(evaluations):
+    for evaluation in evaluations:
+        if evaluation.course.evaluations.exclude(state="published").exists():
+            evaluation.course.not_all_evaluations_are_published = True
+        evaluation.course.evaluation_count = evaluation.course.evaluations.count()
+        evaluation.course.distribution = calculate_average_course_distribution(evaluation.course)
+        evaluation.course.avg_grade = distribution_to_grade(evaluation.course.distribution)
+        evaluation.course.evaluation_weight_sum = evaluation.course.evaluations.all().aggregate(Sum('weight'))["weight__sum"]
```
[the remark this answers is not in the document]
janno42: i could collect all courses from the given `evaluations`, calculate the values for them and then iterate over the evaluations again and assign their annotated course to them. or do you have other ideas?

evap/results/tools.py
```diff
@@ -202,6 +202,30 @@ def average_non_grade_rating_questions_distribution(results):
     )
 
 
+def calculate_average_course_distribution(course):
+    if course.evaluations.exclude(state="published").exists():
+        return None
```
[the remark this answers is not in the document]
janno42: the if checks whether any unpublished evaluation exists. if a course has unpublished evaluations, no result will be calculated.

evap/contributor/views.py
```diff
@@ -21,22 +23,40 @@ def index(request):
...
+    displayed_evaluations = list(displayed_evaluations)
+    displayed_evaluations.sort(key=lambda evaluation: evaluation.full_name)  # evaluations must be sorted for regrouping them in the template
 
     for evaluation in displayed_evaluations:
-        evaluation.distribution = calculate_average_distribution(evaluation)
-        evaluation.avg_grade = distribution_to_grade(evaluation.distribution)
+        if evaluation.state == "published":
+            if not evaluation.is_single_result:
+                evaluation.distribution = calculate_average_distribution(evaluation)
```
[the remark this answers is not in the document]
janno42: this calculates the results for the evaluations, `get_evaluations_with_course_result_attributes` calculates the courses' results. is this still a problem after addressing what you commented on `get_evaluations_with_course_result_attributes` above?

evap/results/tools.py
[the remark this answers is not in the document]
karyon: i see that, but why do you need the special case? shouldn't the code below return some empty result in that case all by itself?
also, isn't it inconsistent that if all evaluations are unpublished, nothing is calculated, but as soon as there is one published evaluation, the unpublished ones are included in the calculation as well?

evap/results/views.py
```diff
@@ -54,17 +77,26 @@ def warm_up_template_cache(evaluations):
...
         assert evaluation.state == "published"
         _delete_template_cache_impl(evaluation)
         warm_up_template_cache([evaluation])
 
 
+def update_template_cache_of_published_evaluations_in_course(course):
+    course_evaluations = course.evaluations.filter(state="published")
+    for course_evaluation in course_evaluations:
+        _delete_evaluation_template_cache_impl(course_evaluation)
+    _delete_course_template_cache_impl(course)
```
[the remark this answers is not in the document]
janno42: `warm_up_template_cache` recreates it

evap/contributor/views.py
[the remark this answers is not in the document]
karyon: i understood that, i took issue with doing that work multiple times for the same evaluation. as far as i can see, if you restructure `get_evaluations_with_course_result_attributes` to do that work per course, you're still doing it twice per evaluation. see the top comment of the review for some input on how this might be restructured.

evap/results/views.py
```diff
@@ -54,17 +77,26 @@ def warm_up_template_cache(evaluations):
         translation.activate(current_language)  # reset to previously set language to prevent unwanted side effects
 
 
-def update_template_cache(evaluations):
+def update_evaluation_template_cache(evaluations):
     for evaluation in evaluations:
         assert evaluation.state == "published"
         _delete_template_cache_impl(evaluation)
```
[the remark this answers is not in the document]
janno42: it calls `_delete_template_cache_impl` which calls `_delete_evaluation_template_cache_impl` and `_delete_course_template_cache_impl`

evap/results/views.py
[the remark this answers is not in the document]
janno42: do you have any suggestions how it could be improved?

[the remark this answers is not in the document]
janno42: yes. i'm counting the absolute number of evaluations the course of the current evaluation has as well as the weight_sum for all of these evaluations.

evap/results/views.py
```diff
@@ -180,15 +227,27 @@ def evaluation_detail(request, semester_id, evaluation_id):
         general_questionnaire_results_top += general_questionnaire_results_bottom
         general_questionnaire_results_bottom = []
 
-    evaluation.distribution = calculate_average_distribution(evaluation)
-    evaluation.avg_grade = distribution_to_grade(evaluation.distribution)
+    course_evaluations = []
+    if evaluation.course.evaluations.count() > 1:
+        course_evaluations = [evaluation for evaluation in evaluation.course.evaluations.filter(state="published") if evaluation.can_user_see_evaluation(request.user)]
```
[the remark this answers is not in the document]
janno42: the list holds other evaluations for the course that are shown in an overview on top of the page. only published evaluations should be listed there. `can_user_see_evaluation` doesn't tell anything about results access, merely whether the user "may know that this evaluation exists". so `can_user_see_results_page` could be something to look at here but that wouldn't work for single results as there is no results page to view but they should be included in the list.

evap/student/views.py
```diff
@@ -10,38 +10,38 @@
...
 def index(request):
-    # retrieve all evaluations, where the user is a participant and that are not new
-    evaluations = list(set(Evaluation.objects.filter(participants=request.user).exclude(state="new")))
+    # retrieve all courses which have evaluations that are not in state "new" and in which the user participates
+    courses = Course.objects.filter(
+        evaluations__participants=request.user,
+        evaluations__state__in=['prepared', 'editor_approved', 'approved', 'in_evaluation', 'evaluated', 'reviewed', 'published']
+    )
+    # retrieve all evaluations which the user can see that are not new
+    evaluations = list(set([evaluation for course in courses for evaluation in course.evaluations.all() if evaluation.state != "new" and evaluation.can_user_see_evaluation(request.user)]))
```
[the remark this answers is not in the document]
janno42: i want to show all evaluations of a course even if the user doesn't participate in all of them. so if a course has at least one evaluation in which the user participates that is in one of the allowed states, i want all evaluations of this course in the allowed states, regardless of participation. i'm not sure how to query that but will think about it.
putting the "new" check into `can_user_see_evaluation` sound great - i have no ide why we didn't do this before.

evap/results/views.py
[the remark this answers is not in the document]
karyon: yes, prefetching. in case of `evaluation_count` it seems you're already setting `course_evaluations_count` in `get_evaluations_with_prefetched_data`.

evap/staff/forms.py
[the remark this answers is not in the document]
janno42: hm. i added it during testing because it made creating courses with many evaluations easier. but i can't think of real use cases right now, so we can probably remove everything related to it.

evap/results/views.py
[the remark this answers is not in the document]
karyon: yes, i can see that in the code :) the question would be *why* you're taking all evaluations into account while only displaying some of them.

evap/staff/forms.py
```diff
@@ -133,7 +134,7 @@ def clean(self):
     def save(self, *args, **kwargs):
         course_type = super().save(*args, **kwargs)
         if "name_en" in self.changed_data or "name_de" in self.changed_data:
-            update_template_cache(Evaluation.objects.filter(state="published", course__type=course_type))
+            update_evaluation_template_cache(Evaluation.objects.filter(state="published", course__type=course_type))
```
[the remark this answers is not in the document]
janno42: because of the overview list on each evaluation detail page that includes the course's information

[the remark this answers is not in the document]
karyon: i don't understand, does the evaluation line in that overview list change when changing the course's type?

evap/results/exporters.py
janno42: what would a course be in the export? it's more or less just one grade. all the relevant information is in the evaluations, isn't it?

evap/results/views.py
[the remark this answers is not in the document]
karyon: i don't really see a consistent structure how the method names indicate whether they should receive a course or an evaluation and whether they touch just the evaluation, just the course, a course's evaluations or all of that, please check whether you can make that consistent.

evap/results/exporters.py
karyon: right... i don't know. i had the impression that the only remaining use case of the export is to see whether any course is making problems. if people have to aggregate multiple columns of numbers in their heads to answer that question i think that's making that use case less user friendly. if you think the export serves its purpose just fine with multiple columns per course, then that's fine with me.

evap/results/tools.py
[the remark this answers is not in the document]
janno42: i think we're talking at cross purposes, because i don't get what you're asking at all.
if one evaluation is published, nothing is done, only if all evaluations are published, results are calculated.

evap/results/views.py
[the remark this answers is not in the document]
janno42: once there is more than one evaluation i want to show how large the weight of each individual one is. if only one of the evaluations is shown and has a weight of, e.g., 1, and another one with weight 2 is not yet displayed, the first one should already say that it's grade will make 33% of the final grade. so i need to now the absolute number and the weight sum.

evap/staff/forms.py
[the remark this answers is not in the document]
janno42: ah, i think i confused the evaluation line and evaluation page here. you're right, then only the course template cache should need to be updated.
only if the course name changes and the evaluation has no name it will have an effect on the evaluation line as well.

evap/results/exporters.py
janno42: it's basically scanning the sheet for orange and red fields when talking about the last semester - so i think everything is fine.

evap/contributor/views.py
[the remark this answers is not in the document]
janno42: now using `displayed_evaluations |= delegated_evaluations - displayed_evaluations`

evap/staff/forms.py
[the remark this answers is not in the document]
janno42: should be removed in refactoring

evap/contributor/views.py
```diff
@@ -29,7 +29,7 @@ def index(request):
             Q(evaluations__contributions__contributor=user)
         )
     )
-    own_evaluations = [evaluation for course in own_courses for evaluation in course.evaluations.all() if evaluation.state != "new" and evaluation.can_user_see_evaluation(user)]
+    own_evaluations = [evaluation for course in own_courses for evaluation in course.evaluations.all() if evaluation.state in evaluation.can_user_see_evaluation(user)]
```
He3lixxx: Quick comment for others: This is broken here but fixed again in 5b7940c, another fixup commit.

janno42: sorry :(

evap/results/views.py
```diff
@@ -86,10 +86,7 @@ def update_evaluation_template_cache(evaluations):
 
 def update_template_cache_of_published_evaluations_in_course(course):
     course_evaluations = course.evaluations.filter(state="published")
-    for course_evaluation in course_evaluations:
-        _delete_evaluation_template_cache_impl(course_evaluation)
-    _delete_course_template_cache_impl(course)
-    warm_up_template_cache(course_evaluations)
+    update_template_cache(course_evaluations)
```
He3lixxx: This is why the test fails - Sorry, I messed this up because we run into a strange scenario here: If the course does not have evaluations anymore, we still need to do `_delete_course_template_cache_impl` - otherwise, course_evaluations will be empty -> the cache for the course will not be cleared. You could do:
```
if course_evaluations:
    update_template_cache(course_evaluations)
else:
    _delete_course_template_cache_impl(course)
    warm_up_template_cache(???) # or something else - do we need some cache here? Or will a course without evaluations always be filtered from results? This is a struggle with the old as well as the new version of this method.
```
_if_ we want it to be possible to have a course without evaluations. In this case, we'd expect two calls on `_delete_course_template_cache_impl` and 0 calls on `warm_up_template_cache`, since the evalulation is not published and therefore filtered. I get that with the modifications. It may be easier to just revert these three lines if we plan on changing the caching anyway (only cache per course)

PR #1307, opened 2019-02-11
translate warning descriptors lazily

### LINKED ISSUE #1206: Translation of import warning card titles missing
When importing course data and receiving warnings, they are grouped in cards. The card titles are not translated although they should be.

Pottiman:
This fixes #1206 by using `ugettext_lazy` to translate `WARNING_DESCRIPTORS` of the importer.

### DIFF
--- a/evap/staff/importers.py
+++ b/evap/staff/importers.py
@@ -3,7 +3,7 @@ import xlrd
 
 from django.conf import settings
 from django.db import transaction
-from django.utils.translation import ugettext as _
+from django.utils.translation import ugettext as _, ugettext_lazy as _lazy
 from django.utils.safestring import mark_safe
 from django.core.exceptions import ValidationError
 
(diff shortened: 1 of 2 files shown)


evap/staff/importers.py
```diff
@@ -3,7 +3,7 @@
 
 from django.conf import settings
 from django.db import transaction
-from django.utils.translation import ugettext as _
+from django.utils.translation import ugettext as _, ugettext_lazy as _lazy
```
He3lixxx: Thanks for contributing! :)

This is just a minor thing, but I'd rather think of `_` as the only shorthand and not create the similar shorthand `_lazy`. Thus, I'd ask you to just
```
from django.utils.translation import ugettext_lazy, ugettext as _
```
and use `ugettext_lazy` below.

I've talked with @janno42 and he agreed.
@karyon what do you think?

karyon: 👍

Pottiman: done

PR #1308, opened 2019-02-11
Css fixes

### LINKED ISSUE #1302: Repair print view for bipolar distributions bars
The print view of results pages does not explicitly handle bipolar distribution bars. In the current implementation the icon can barely be seen. Instead, the adjective should be shown in black.

pixunil:
- Resolve #1302 by always showing the adjectives and hiding the icons for bipolar results
- Footer now has the colors back which were stolen by the bipolar renames


pixunil:
Should be now resolved. I had to change the unit for `line-height` but it should work like the same before

PR #1309, opened 2019-02-13
Results page degree filter

### LINKED ISSUE #1301: Fix degree filter on results page
When selecting a degree for filtering on the results page, only courses which have solely this degree are shown. Instead, all courses which have at least this degree and any number of additional other degrees should remain visible.

janno42:
fix #1301; correctly filter degrees on results page

### DIFF
--- a/evap/results/templates/results_index.html
+++ b/evap/results/templates/results_index.html
@@ -119,7 +119,7 @@
             $('#results-table tr:not(.table-header)').each(function() {
                 var showRow =
                     (semesters.length == 0 || semesters.includes($(this).find('.semester-name').text())) &&
-                    (degrees.length == 0 || degrees.includes($(this).find('.badge-degree').text())) &&
+                    (degrees.length == 0 || includesAny($(this).find('.badge-degree').text(), degrees)) &&
                     (courseTypes.length == 0 || courseTypes.includes($(this).find('.badge-course-type').text())) &&
                     (searchValue.length == 0 ||
                         $(this).find('.evaluation-name').html().toLowerCase().includes(searchValue) ||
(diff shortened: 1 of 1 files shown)


evap/results/templates/results_index.html
```diff
@@ -128,6 +128,15 @@ <h6 class="card-subtitle mb-1">{% trans 'Semester' %}</h6>
             });
         };
 
+        function includesAny(string, array){
+            for(var i = 0; i < array.length; i++){
+                if(string.includes(array[i])){
+                    return true;
+                }
+            }
+            return false;
```
karyon: this would be equivalent to `return array.some(s => searchstring.includes(s))`

but actually, don't you want `s.includes(searchstring)`?

janno42: i was looking for something like `some`, thanks :)

the searchstring consists of all selected degrees. so i want to check whether the array contains any element that is included in the string.

karyon: ah, right. i assumed this was something the user entered manually.

janno42: please review again

evap/results/templates/results_index.html
```diff
@@ -117,9 +117,10 @@ <h6 class="card-subtitle mb-1">{% trans 'Semester' %}</h6>
 
             // find which rows to show
             $('#results-table tr:not(.table-header)').each(function() {
+                var evaluationDegrees = $(this).find('.badge-degree').text();
                 var showRow =
                     (semesters.length == 0 || semesters.includes($(this).find('.semester-name').text())) &&
-                    (degrees.length == 0 || degrees.includes($(this).find('.badge-degree').text())) &&
+                    (degrees.length == 0 || degrees.some(s => evaluationDegrees.includes(s))) &&
```
karyon: i'd rather inline this, or use the function as before, because right now the code is doing work for every line even if it's not required.

PR #1310, opened 2019-02-18
Refactor evaluation bulk operation

felixrindt:
This refactors the evaluation bulk operation.

In a separate commit, it removes the start evaluation operation, because it is not needed anymore. Evaluations can be started by setting the start date accordingly and waiting for the cronjob to trigger the state change.

### DIFF
--- a/evap/staff/views.py
+++ b/evap/staff/views.py
@@ -169,10 +169,7 @@ class EvaluationOperation:
 
 class RevertToNewOperation(EvaluationOperation):
     confirmation_message = ugettext_lazy("Do you want to revert the following evaluations to preparation?")
-
-    @staticmethod
-    def applicable_to(evaluation):
-        return evaluation.state in ['prepared', 'editor_approved', 'approved']
+    applicable_to = lambda evaluation: evaluation.state in ['prepared', 'editor_approved', 'approved']
 
     @staticmethod
     def warning_for_inapplicables(amount):
@@ -187,10 +184,7 @@ class RevertToNewOperation(EvaluationOperation):
 class RevertToPreparedOperation(EvaluationOperation):
     email_template_name = EmailTemplate.EDITOR_REVIEW_NOTICE
     confirmation_message = ugettext_lazy("Do you want to revert the following evaluations to preparation?")
-
-    @staticmethod
-    def applicable_to(evaluation):
-        return evaluation.state in ['new', 'editor_approved']
+    applicable_to = lambda evaluation: evaluation.state in ['new', 'editor_approved']
 
     @staticmethod
     def warning_for_inapplicables(amount):
@@ -204,10 +198,7 @@ class RevertToPreparedOperation(EvaluationOperation):
 
 class RevertToReviewedOperation(EvaluationOperation):
     confirmation_message = ugettext_lazy("Do you want to unpublish the following evaluations?")
-
-    @staticmethod
-    def applicable_to(evaluation):
-        return evaluation.state == 'published'
+    applicable_to = lambda evaluation: evaluation.state == 'published'
 
     @staticmethod
     def warning_for_inapplicables(amount):
@@ -222,10 +213,7 @@ class RevertToReviewedOperation(EvaluationOperation):
 class PublishOperation(EvaluationOperation):
     email_template_name = EmailTemplate.PUBLISHING_NOTICE
     confirmation_message = ugettext_lazy("Do you want to publish the following evaluations?")
-
-    @staticmethod
-    def applicable_to(evaluation):
-        return evaluation.state == 'reviewed'
+    applicable_to = lambda evaluation: evaluation.state == 'reviewed'
 
     @staticmethod
     def warning_for_inapplicables(amount):
--- a/evap/staff/views.py
+++ b/evap/staff/views.py
@@ -159,8 +159,7 @@ class EvaluationOperation:
 
     @staticmethod
     def warning_for_inapplicables(amount):
-        return ungettext("{} evaluation can not be updated. It was removed from the selection.",
-            "{} evaluations can not be updated. They were removed from the selection.", amount).format(amount)
+        raise NotImplementedError
 
     @staticmethod
     def apply(request, evaluations, email_template=None):
@@ -315,9 +314,8 @@ def semester_evaluation_operation(request, semester_id):
     applicable_evaluations = list(filter(operation.applicable_to, evaluations))
     difference = len(evaluations) - len(applicable_evaluations)
     if difference:
-        evaluations = applicable_evaluations
         messages.warning(request, operation.warning_for_inapplicables(difference))
-    if not evaluations:  # no evaluations where applicable or none were selected
+    if not applicable_evaluations:  # no evaluations where applicable or none were selected
         messages.warning(request, _("Please select at least one evaluation."))
         return custom_redirect('staff:semester_view', semester_id)
 
@@ -327,7 +325,7 @@ def semester_evaluation_operation(request, semester_id):
 
     template_data = dict(
         semester=semester,
-        evaluations=evaluations,
+        evaluations=applicable_evaluations,
         target_state=target_state,
         confirmation_message=operation.confirmation_message,
         email_template=email_template,
--- a/evap/staff/views.py
+++ b/evap/staff/views.py
@@ -158,7 +158,7 @@ class EvaluationOperation:
         raise NotImplementedError
 
     @staticmethod
-    def warning_for_unapplicables(amount):
+    def warning_for_inapplicables(amount):
         return ungettext("{} evaluation can not be updated. It was removed from the selection.",
             "{} evaluations can not be updated. They were removed from the selection.", amount).format(amount)
 
@@ -173,7 +173,7 @@ class RevertToNewOperation(EvaluationOperation):
         return evaluation.state in ['prepared', 'editor_approved', 'approved']
 
     @staticmethod
-    def warning_for_unapplicables(amount):
+    def warning_for_inapplicables(amount):
         return ungettext("{} evaluation can not be reverted, because it already started. It was removed from the selection.",
             "{} evaluations can not be reverted, because they already started. They were removed from the selection.", amount).format(amount)
 
@@ -194,7 +194,7 @@ class RevertToPreparedOperation(EvaluationOperation):
         return evaluation.state in ['new', 'editor_approved']
 
     @staticmethod
-    def warning_for_unapplicables(amount):
+    def warning_for_inapplicables(amount):
         return ungettext("{} evaluation can not be reverted, because it already started. It was removed from the selection.",
             "{} evaluations can not be reverted, because they already started. They were removed from the selection.", amount).format(amount)
 
@@ -213,7 +213,7 @@ class RevertToReviewedOperation(EvaluationOperation):
         return evaluation.state == 'published'
 
     @staticmethod
-    def warning_for_unapplicables(amount):
+    def warning_for_inapplicables(amount):
         return ungettext("{} evaluation can not be unpublished, because it's results have not been published. It was removed from the selection.",
             "{} evaluations can not be unpublished because their results have not been published. They were removed from the selection.", amount).format(amount)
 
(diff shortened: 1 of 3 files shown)


evap/staff/views.py
```diff
@@ -51,6 +51,13 @@ def index(request):
     return render(request, "staff_index.html", template_data)
 
 
+def annotate_evaluations_with_grade_document_counts(evaluations):
+    return evaluations.annotate(
+            midterm_grade_documents_count=Count("course__grade_documents", filter=Q(course__grade_documents__type=GradeDocument.MIDTERM_GRADES), distinct=True),
+            final_grade_documents_count=Count("course__grade_documents", filter=Q(course__grade_documents__type=GradeDocument.FINAL_GRADES), distinct=True))
+
+
+
```
felixrindt: There should be only 2 empty lines here.

evap/staff/views.py
```diff
@@ -139,97 +145,142 @@ def __init__(self):
...
+                    amount) % {'evaluations': amount}
+
+    def confirmation_message(self):
+        return _("Do you want to update the following evaluations?")
+
+    def apply(self, request, evaluations, email_template=None):
+        raise NotImplementedError
+
+    def email_template_name(self):
+        return None
```
felixrindt: I don't really like the use of `None` here yet.

evap/staff/views.py
```diff
@@ -139,97 +145,142 @@ def __init__(self):
...
+    def confirmation_message(self):
+        return _("Do you want to publish the following evaluations?")
+
+    def apply(self, request, evaluations, email_template=None):
+        helper_semester_evaluation_operation_publish(request, evaluations, email_template)
+
+    def email_template_name(self):
+        return EmailTemplate.PUBLISHING_NOTICE
+
+
```
felixrindt: There should be only 2 empty lines here.

He3lixxx:
Just from a quick glanze:
- Code formatting is inconsistent at many places. Please do two empty lines between classes, indent 4 spaces - also when indenting because you are listing function parameters. I'm not sure about empty lines at the start of classes, but I think we agreed on not having one there the last time we discussed (@janno42, @karyon?)
- You could make things that act like properties look like properties with the @property decorator - e.g. `email_template_name` and `confirmation_message`
- Have you thought about making all the methods class methods? I think we do that in some other places - it would save you the need of instantiating something that does not actually need any state.
- String formatting: Using `%` is discouraged by the python documentation, we probably don't want to introduce new code using it (see [here](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting) )
- Last but most important: I really think this makes it better. Thanks!

evap/staff/views.py
```diff
@@ -139,105 +144,152 @@ def __init__(self):
...
+        return evaluation.state in ['prepared', 'editor_approved', 'approved']
+
+    @staticmethod
+    def warning_for_unapplicables(amount):
+        return ungettext("{} evaluation can not be reverted, because it already started. It was removed from the selection.",
+            "{} evaluations can not be reverted, because they already started. They were removed from the selection.", amount).format(amount)
+
+    @staticmethod
+    def confirmation_message():
+        return _("Do you want to revert the following evaluations to preparation?")
```
felixrindt: Although I would like this to be a property, I'm not sure how to do that on class level. As it is a translated string, either there needs to be a function behind this property or it's a property like `email_template_name`, but with a lazy ugettext. In my testing, even with how it is now, the strings were not translated, although there should have been a translation available, as these weren't new strings.

He3lixxx: I guess the `_lazy` translation methods are what you want.

janno42:
"I'm not sure about empty lines at the start of classes, but I think we agreed on not having one there the last time we discussed" - yes, please don't put empty lines at the beginning of classes.

evap/staff/views.py
felixrindt: I'll try on Monday.

evap/staff/views.py
```diff
@@ -139,105 +144,152 @@ def __init__(self):
     return render(request, "staff_semester_view.html", template_data)
 
 
+class EvaluationOperation:
+    email_template_name = None
+    confirmation_message = None
```
karyon: this one is unused.

speaking of which, i actually think it's a good idea, how about making the first three methods *actual* properties? just like `email_template_name`. the first one would be a list of strings, then a string and a lazy string. i also thought about making `apply` a lambda and all the `Operations` tuples, but i'm not sure that makes it better. if you want you could try that as well.

evap/staff/views.py
```diff
@@ -139,105 +144,152 @@ def __init__(self):
...
+        raise NotImplementedError
+
+
+class RevertToNewOperation(EvaluationOperation):
+    @staticmethod
+    def applicable_to(evaluation):
+        return evaluation.state in ['prepared', 'editor_approved', 'approved']
+
+    @staticmethod
+    def warning_for_unapplicables(amount):
```
karyon: inapplicable or nonapplicable

evap/staff/views.py
```diff
@@ -144,159 +149,194 @@ def __init__(self):
...
-            email_template = EmailTemplate.objects.get(name=EmailTemplate.PUBLISHING_NOTICE)
-            confirmation_message = _("Do you want to publish the following evaluations?")
-
-    if not evaluations:
+    applicable_evaluations = list(filter(operation.applicable_to, evaluations))
+    difference = len(evaluations) - len(applicable_evaluations)
+    if difference:
+        evaluations = applicable_evaluations
+        messages.warning(request, operation.warning_for_inapplicables(difference))
+    if not evaluations:  # no evaluations where applicable or none were selected
```
karyon: i think this should be `applicable_evaluations`, could you check?

evap/staff/views.py
```diff
@@ -144,159 +149,194 @@ def __init__(self):
...
+    confirmation_message = ugettext_lazy("Do you want to immediately start the following evaluations?")
+
+    @staticmethod
+    def applicable_to(evaluation):
+        return evaluation.state == 'approved' and evaluation.vote_end_date >= date.today()
+
+    @staticmethod
+    def warning_for_inapplicables(amount):
+        return ungettext("{} evaluation can not be started, because it was not approved, was already evaluated or its evaluation end date lies in the past. It was removed from the selection.",
+            "{} evaluations can not be started, because they were not approved, were already evaluated or their evaluation end dates lie in the past. They were removed from the selection.", amount).format(amount)
```
karyon: i'd suggest to make `warning_for_inapplicables` a property just like `confirmation_message` is.

evap/staff/views.py
```diff
@@ -144,159 +149,194 @@ def __init__(self):
...
+    confirmation_message = None
+
+    @staticmethod
+    def applicable_to(evaluation):
+        raise NotImplementedError
+
+    @staticmethod
+    def warning_for_inapplicables(amount):
+        return ungettext("{} evaluation can not be updated. It was removed from the selection.",
+            "{} evaluations can not be updated. They were removed from the selection.", amount).format(amount)
```
karyon: `raise NotImplementedError`?

evap/staff/views.py
felixrindt: Should be correct, as `evaluations` points to `applicable_evaluations` if there was a difference between the two. `evaluations` is also used in the template context. 

A way to make this clearer is to have `evaluations` be the filtered list from the beginning, and keeping a `selected_evaluations` list for the difference check. What do you think about that?

evap/staff/views.py
felixrindt: I wasn't sure how to do this, as this is using the pluralized form. How do I make a property out of that?

evap/staff/views.py
felixrindt: Fits the rest of the class more - I agree.

evap/staff/views.py
karyon: i think `evaluations` and `applicable_evaluations` are fine, i would simply drop the re-purposing two lines above, use `applicable_evaluations` here, and do `evaluations=applicable_evaluations` for the template.

evap/staff/views.py
karyon: um, yeah, i didn't realize this was two strings in a method call. i can't think of something that doesn't look ugly.

evap/staff/views.py
```diff
@@ -144,159 +149,192 @@ def __init__(self):
...
+        for evaluation in evaluations:
+            evaluation.revert_to_new()
+            evaluation.save()
+        messages.success(request, ungettext("Successfully reverted {} evaluation to in preparation.",
+            "Successfully reverted {} evaluations to in preparation.", len(evaluations)).format(len(evaluations)))
+
+
+class RevertToPreparedOperation(EvaluationOperation):
+    email_template_name = EmailTemplate.EDITOR_REVIEW_NOTICE
+    confirmation_message = ugettext_lazy("Do you want to revert the following evaluations to preparation?")
```
janno42: wrong message

evap/staff/views.py
```diff
@@ -144,159 +149,192 @@ def __init__(self):
...
-            helper_semester_evaluation_operation_prepare(request, evaluations, template)
-        elif target_state == 'in_evaluation':
-            helper_semester_evaluation_operation_start(request, evaluations, template)
-        elif target_state == 'reviewed':
-            helper_semester_evaluation_operation_unpublish(request, evaluations)
-        elif target_state == 'published':
-            helper_semester_evaluation_operation_publish(request, evaluations, template)
-
+            email_template = EmailTemplate(subject=request.POST['email_subject'], body=request.POST['email_body'])
+        operation.apply(request, evaluations, email_template)
```
janno42: remember this needs to be updated when #1317 is merged

evap/staff/views.py
```diff
@@ -144,159 +149,192 @@ def __init__(self):
...
+    @staticmethod
+    def apply(request, evaluations, email_template=None):
+        for evaluation in evaluations:
+            evaluation.revert_to_new()
+            evaluation.save()
+        messages.success(request, ungettext("Successfully reverted {} evaluation to in preparation.",
+            "Successfully reverted {} evaluations to in preparation.", len(evaluations)).format(len(evaluations)))
+
+
+class RevertToPreparedOperation(EvaluationOperation):
```
janno42: you could rename all these to `MoveToPreparedOperation` and similar, as the action is not always actually "reverting" the state (e.g. a change from "new" to "prepared")

evap/staff/views.py
```diff
@@ -144,159 +149,192 @@ def __init__(self):
...
+
+    @staticmethod
+    def warning_for_inapplicables(amount):
+        return ungettext("{} evaluation can not be reverted, because it already started. It was removed from the selection.",
+            "{} evaluations can not be reverted, because they already started. They were removed from the selection.", amount).format(amount)
+
+    @staticmethod
+    def apply(request, evaluations, email_template=None):
+        for evaluation in evaluations:
+            evaluation.ready_for_editors()
```
janno42: the methods should optimally use the same names as the states. it is a bit confusing that `ready_for_editors` moves the evaluation to `prepared`. maybe `move_to_...` (`move_from_..._to_...` if the direction is important).

evap/staff/views.py
felixrindt: I prefer to change only this one to `Move`. The rest was fine to @janno42 and me.

evap/staff/views.py
felixrindt: There is a name mismatch with

* `MoveToPreparedOperation` and `evaluation.ready_for_editors`
* `StartEvaluationOperation` and `evaluation.evaluation_begin`
* `RevertToReviewedOperation` and `evaluation.unpublish()`

Do you agree that there should be another refactoring for the state transition names in the model to make it more consistent among themselves and throughout the whole project.

janno42: yes. would you open an issue for that?

evap/staff/views.py
felixrindt: this will be merged first

He3lixxx: We agreed to change 1317 afterwards (as it makes more sense for me to do the merging)

PR #1290, opened 2018-12-10
Fixes #1253 Individual publish notifications for participants and con…

LeaGe:
…tributors

PR #1317, opened 2019-02-25
Individual publish notifications - Continuation of #1290

### LINKED ISSUE #1253: Individual publish notifications for participants and contributors
The Publishing Notice when evaluation results of a course are published should be split into two individual email templates: one sent to the course's participants (with more informal wording) and one for the contributors.

He3lixxx:
Continues #1290
Fixes #1253 

Add individual publishing notifications for participants and contributors.

I think I've implemented all changes that were requested there, probably @janno42 should still review again.

@LeaGe: Thanks for doing most of this :)


evap/staff/tests/test_views.py
```diff
@@ -808,8 +808,13 @@ def helper_semester_state_views(self, evaluation, old_state, new_state):
...
     """
     def test_semester_publish(self):
-        evaluation = mommy.make(Evaluation, course=mommy.make(Course, semester=self.semester), state='reviewed')
+        participant1 = mommy.make(UserProfile, email="foo@example.com")
+        participant2 = mommy.make(UserProfile, email="bar@example.com")
+        course = mommy.make(Course, semester=self.semester, state='reviewed', participants=[participant1, participant2], voters=[participant1, participant2])
+        evaluation = mommy.make(Evaluation, course=course, semester=self.semester, state='reviewed')
+
         self.helper_semester_state_views(evaluation, "reviewed", "published")
+        self.assertEqual(len(mail.outbox), 2)
```
janno42: could you please also add a test counting the emails if one or two of the respective checkboxes are not selected?

evap/staff/views.py
```diff
@@ -162,8 +160,14 @@ def semester_evaluation_operation(request, semester_id):
 
     if request.method == 'POST':
         template = None
+        template_contributor = None
+        template_participant = None
         if request.POST.get('send_email') == 'on':
             template = EmailTemplate(subject=request.POST['email_subject'], body=request.POST['email_body'])
```
janno42: ['email_subject'] results in a `MultiValueDictKeyError`. did you even test this? :)

He3lixxx: That's a good point, forgot to set to WIP. Yes, I haven't tested yet, so far it's been code only. Sorry for the confusion.

He3lixxx:
Reminder to self: Also change the templates in the test data (difference: "Du" vs "Sie")

He3lixxx:
I don't understand why the "load testdata" test fails - or at least I don't understand why it didn't fail up to this point.
`migrate` will create an empty email template in migration 102, just like migration 81, 61 and 4 do.
`flush` will remove these again. After that, `loaddata` can be executed (-> `reload_testdata` does this and works fine)

However, obviously, when using loaddata on a migrated database without flush, the import will fail, just as it does on travis. What I don't understand is what's different about the email templates created by 4, 61 and 81. They are also created in the migrations and they are also contained in the test_data.json. What am I missing? Is this some kind of weird django bug caused by the rename? Maybe @karyon can help me here.

karyon:
previously, the json overwrote the values in the database (created by the migrations) with identical values (in terms of pk and name), which works. i'd guess that with your changes, after the migrations are executed, the `Login Key Created` template still has pk 4 in the database, because that's the order the templates were created in the migrations, but the json now says it should create `Publishing Notice Participant` with pk 4, which fails. 

i think in other, similar cases we hard-coded the pks in the migrations so that json pks and migration pks match.

He3lixxx:
Sounds reasonable. I reordered the template in the test_data.json so that the pks should match what comes out of the migrations. If you prefer, I can also change the migrations, but I felt like that was more history rewriting.

PR #1321, opened 2019-03-11
button only gets displayed if the user has the right permissions

### LINKED ISSUE #1314: Evaluation preview button visibility
As a teaching assistant, I might be a contributor to a given course and therefore get my own feedback in the main evaluation. If that course also has an exam evaluation, I see that listed on my "own evaluations" page with the option to preview the questionnaire. However, as not being responsible, I miss the access rights to preview the linked page, resulting in an error.

I would like to either don't have the preview button (it already knows while rendering that page that I am not a contributor, shown through the corresponding icon next to the exam evaluation title) or to give me the rights to preview the questionnaire.

Mupico:
fix #1314


evap/contributor/templates/contributor_index.html
```diff
@@ -179,11 +179,11 @@
                                                         <span class="fas fa-pencil-alt"></span>
                                                     </a>
                                                 {% endif %}
+                                                <a href="{% url 'contributor:evaluation_preview' evaluation.id %}" class="btn btn-sm btn-light"
```
He3lixxx: This is now inside the `if user is editor or delegate` part. The preview view, however, has the decorator `@responsible_or_contributor_or_delegate_required` (see https://github.com/fsr-de/EvaP/blob/c0eada344ece2bc6bb9750502cb5f9b217bb3792/evap/contributor/views.py#L172) and also checks for that relationship between user and evaluation (whether the user `is_user_responsible_or_contributor_or_delegate`)
This doesn't really match, does it? When a user is contributor in that evaluation, even though they are not an editor, we still want them to get that link (and they should see the page), right?

He3lixxx: Uhh fancy inline preview. Seriously github, you got some nice stuff going on there xD

He3lixxx:
(Random thought: This could be a candidate for a test - check whether links exist where they should and don't exist where they shouldn't)

evap/contributor/templates/contributor_index.html
Mupico: changed that

Issue #1312, opened 2019-02-20
Remove vote_start_datetime error for contributors


tzwenn:
When approving an exam evaluation as lecturer, the "Beginn der Evaluierung" field is prefilled with `<current day> 08:00`.
As I was (wt 2018/19) notified of the pending approval just after this time, this default value gives me a form validation error when not manually adjusted.

![screen shot 2019-02-20 at 10 24 32](https://user-images.githubusercontent.com/495690/53081345-f649a280-34fa-11e9-9751-7d2cdc359985.png)

janno42:
thanks for informing us. that actually was a user error by the evaluation team - the beginning time should have been set later that day. the problem should not occur in non-experimental same day rush starts :)
it is intended that contributors can't set the beginning date to the past.

tzwenn:
👍

tzwenn:
Maybe this can still be fixed in code: When a lecture/exam is not approved, it can't be evaluated. (Can it?) Why then offer any date in the past? Maybe just prefill this field with "now + X"?

janno42:
good point. automatically setting values feels a bit weird, but in this case it shouldn't cause any problems.
`vote_start_datetime` should be set to `now` on saving the contributor course edit form when the entered value is in the past.

janno42:
We decided to just remove the form error for contributors. Dates can then be saved also if they are in the past.

He3lixxx:
Fixed in #1324

Issue #228, opened 2014-01-11
broken layout of /results/semester/x


karyon:
"voters" and "contributors" are too narrow. line breaks should be avoided for usual names with a "prof. dr."

alternatively, the names could be shown in italic letters below the courses name, as in the fsr/semester view.

![results](https://f.cloud.github.com/assets/1891915/1895170/f5f6c14c-7b18-11e3-912f-434364259737.png)

PR #1322, opened 2019-03-13
Solve #1312

SilvanVerhoeven:
solves #1312

### DIFF
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -42,11 +42,12 @@ class EvaluationForm(forms.ModelForm):
     def clean(self):
         super().clean()
 
-        if self.cleaned_data['vote_start_datetime'] < datetime.now():
-            self.cleaned_data['vote_start_datetime'] = round_datetime_up(round_to=5 * 60)
-
         vote_start_datetime = self.cleaned_data.get('vote_start_datetime')
         vote_end_date = self.cleaned_data.get('vote_end_date')
+
+        if vote_start_datetime and vote_start_datetime < datetime.now():
+            self.cleaned_data['vote_start_datetime'] = round_datetime_up(round_to=5 * 60)
+
         if vote_start_datetime and vote_end_date:
             if vote_start_datetime.date() > vote_end_date:
                 self.add_error("vote_start_datetime", "")
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -8,7 +8,7 @@ from django.forms.widgets import CheckboxSelectMultiple
 from django.utils.translation import ugettext_lazy as _
 from evap.evaluation.forms import UserModelMultipleChoiceField, UserModelChoiceField
 from evap.evaluation.models import Course, Evaluation, Questionnaire, UserProfile
-from evap.evaluation.tools import date_to_datetime
+from evap.evaluation.tools import date_to_datetime, round_datetime_up
 from evap.staff.forms import ContributionForm
 
 logger = logging.getLogger(__name__)
@@ -42,6 +42,9 @@ class EvaluationForm(forms.ModelForm):
     def clean(self):
         super().clean()
 
+        if self.cleaned_data['vote_start_datetime'] < datetime.now():
+            self.cleaned_data['vote_start_datetime'] = round_datetime_up(round_to=5*60)
+
         vote_start_datetime = self.cleaned_data.get('vote_start_datetime')
         vote_end_date = self.cleaned_data.get('vote_end_date')
         if vote_start_datetime and vote_end_date:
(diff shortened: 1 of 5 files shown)


He3lixxx:
#228 seems to be the wrong issue number. You probably meant #1312?

SilvanVerhoeven:
Yes, no idea how I could be mixing this up. Thanks for correcting!

evap/contributor/forms.py
```diff
@@ -44,23 +44,26 @@ def clean(self):
 
         vote_start_datetime = self.cleaned_data.get('vote_start_datetime')
         vote_end_date = self.cleaned_data.get('vote_end_date')
+
+        # IMPORTANT: when removing following feature (auto-set a vote_start_datetime of the past to the future),
+        # make sure to alter the according test ("start date in the past -> form invalid?")!
+        # Location of test: evap/staff/tests/test_forms.py
+        #                   EvaluationFormTests.test_contributor_evaluation_form_date_validation(self):
+        if vote_start_datetime and vote_start_datetime < datetime.now():
+            self.cleaned_data['vote_start_datetime'] = round_datetime_up(round_to=5 * 60)
```
He3lixxx: I'd that we are overwriting user input here, so we at least need some kind of message on the next page telling the user about it. @janno42 opinions?

evap/contributor/forms.py
```diff
@@ -44,23 +44,26 @@ def clean(self):
...
+
         if vote_start_datetime and vote_end_date:
             if vote_start_datetime.date() > vote_end_date:
                 self.add_error("vote_start_datetime", "")
                 self.add_error("vote_end_date", _("The first day of evaluation must be before the last one."))
 
-    def clean_vote_start_datetime(self):
```
He3lixxx: Have you thought about putting the cleaning logic for the vote start date in here? I mean, instead of throwing the error, you could add the fix here and just use that. Then, the form's clean function would not have to deal with this, at all.

evap/contributor/forms.py
```diff
@@ -44,23 +44,26 @@ def clean(self):
...
-
     def clean_vote_end_date(self):
-        vote_end_date = self.cleaned_data.get('vote_end_date')
+        vote_end_date = self.cleaned_data['vote_end_date']
 
         # The actual deadline is EVALUATION_END_OFFSET_HOURS:00 AM of the day after vote_end_date.
         # Therefore an evaluation date 24h + EVALUATION_END_OFFSET_HOURS in the past would technically still be in the future.
-        if vote_end_date and date_to_datetime(vote_end_date) + timedelta(hours=24 + settings.EVALUATION_END_OFFSET_HOURS) < datetime.now():
+        if vote_end_date and date_to_datetime(vote_end_date) + timedelta(hours=24 +
+           settings.EVALUATION_END_OFFSET_HOURS) < datetime.now():
```
He3lixxx: I dislike the three spaces indent, tbh. I'd make it an early out
```
if not vote_end_date:
    return vote_end_date
```
directly at the start and then below the condition would be shorter:
```
if date_to_datetime(vote_end_date) + timedelta(...) < datetime.now():
```
Maybe put the calculation of the timedelta in a new line. Actually, I think there should be one central place where vote_end datetimes are calculated from vote_end_dates anyway, so I'd rather like this
```
end_datetime = Evaluation.calculate_end_datetime_from_end_date(end_date) # naming to be done
if end_datetime < datetime.now():
   raise ...
```

evap/evaluation/tools.py
```diff
@@ -116,3 +116,47 @@ def get_parameter_from_url_or_session(request, parameter, default=False):
 def translate(**kwargs):
     # get_language may return None if there is no session (e.g. during management commands)
     return property(lambda self: getattr(self, kwargs[get_language() or 'en']))
+
+
+def round_datetime(dt=None, round_to=60):
+    """Round a datetime object to closest time lapse (hours/minutes/seconds)
```
He3lixxx: @janno42 is "lapse" the correct word here? I only know it as "Zeitraffer".

evap/evaluation/tools.py
```diff
@@ -116,3 +116,47 @@ def get_parameter_from_url_or_session(request, parameter, default=False):
 def translate(**kwargs):
     # get_language may return None if there is no session (e.g. during management commands)
     return property(lambda self: getattr(self, kwargs[get_language() or 'en']))
+
+
+def round_datetime(dt=None, round_to=60):
```
He3lixxx: I think, since we're not usually doing this in a closed loop a lot of times, I'd prefer having `round_to` taking a string that would either be "seconds", "minutes" or "hours" and you would then do a dictionary lookup to find the corresponding number you need. This would make calls much more readable (imo)

evap/evaluation/tools.py
```diff
@@ -116,3 +116,47 @@ def get_parameter_from_url_or_session(request, parameter, default=False):
...
+
+    if dt is None:
+        dt = datetime.datetime.now()
+    seconds = (dt.replace(tzinfo=None) - dt.min).seconds
+    rounding = (seconds + round_to / 2) // round_to * round_to
+
+    return dt + datetime.timedelta(0, rounding - seconds, -dt.microsecond)
+
+
+def round_datetime_down(dt=None, round_to=60):
```
He3lixxx: python naming would be floor and ceil. I feel like having 3 function here is kind of overkill since basically, all three (floor, ceil, round) are just a different calculation in the original function when you are doing
`rounding = (seconds + round_to / 2) // round_to * round_to`. One would add `round_to`, one would add `round_to / 2`, one would not add anything.

He3lixxx: Also, the _down variant is unused currently, as well as the round variant itself, right? I vote to just make the one function we need, call it `ceil_datetime` and be done with it.

evap/staff/tests/test_forms.py
```diff
@@ -546,7 +547,10 @@ def test_contributor_evaluation_form_date_validation(self):
         evaluation.general_contribution.questionnaires.set([mommy.make(Questionnaire)])
 
         # contributors: start date must be in the future
-        self.helper_date_validation(ContributorEvaluationForm, "1999-01-01", "2099-01-01", False)
+        # since issue #1312, a past vote_start_datetime is automatically set to datetime.now() (rounded up)
+        # -> test wrong, instead test whether form is still valid
+        # self.helper_date_validation(ContributorEvaluationForm, "1999-01-01", "2099-01-01", False)
```
He3lixxx: Code comments are not the place to keep the whole history, we have git for that. I vote to remove the original comment and just replace it by
"since issue #1312, a past vote_start_datetime is automatically set to datetime.now() (ceiled). Thus, this form should be valid"

He3lixxx: (Especially do we not need or want commented code)

evap/contributor/forms.py
SilvanVerhoeven: 👍

PR #1323, opened 2019-03-15
Random UI fixes

janno42:


evap/static/css/_quick-review.scss
```diff
@@ -18,13 +18,19 @@
         .fas {
             cursor: pointer;
         }
+        .badge {
```
pixunil: I would suggest to extend `.slider-side` so that it will manipulate `visibility`, that way the badges will be disabled without touching `display`. If you can sacrifice the opacity transition, then things are very easy, if you want to keep them, then you [could use animations](https://stackoverflow.com/questions/30064932/hide-element-after-fade-out-using-only-css) (you don't need vendor prefixes anymore).

janno42: what is the advantage of your suggested solution compared to setting `display` to `none`?

pixunil: The disadvantage here is that you try to fix a bug by overwriting the child, instead of changing the parent which has the show / hide behaviour. But the children are independent of this behaviour and should only be adjusted when it's really necessary.

Also, if I set the transition duration to a higher value, I see that you already ditched the fade-out and fade-in effects for the badges. Thus, I think something like this would do the same job without child selector: (locally tested)

```css
.slider-item {
    visibility: hidden;
    &.shown {
        visibility: visible;
    }
}
```

(If you apply this, please also remove the `opacity` stuff as it won't be in effect after this)

evap/staff/views.py
```diff
@@ -505,7 +505,7 @@ def semester_todo(request, semester_id):
 def semester_grade_reminder(request, semester_id):
     semester = get_object_or_404(Semester, id=semester_id)
 
-    courses = semester.courses.filter(evaluations__state__in=['evaluated', 'reviewed', 'published'], is_graded=True, gets_no_grade_documents=False).all()
+    courses = semester.courses.filter(evaluations__state__in=['evaluated', 'reviewed', 'published'], is_graded=True, gets_no_grade_documents=False).all().distinct()
```
He3lixxx: You could remove the `all()` after the filter, right?

evap/static/css/_quick-review.scss
```diff
@@ -67,6 +67,9 @@
     &:focus {
         outline: none;
     }
+    &[data-layer="0"] {
```
He3lixxx: what does this do? Github renders it red as if it was completely broken xD

evap/staff/views.py
janno42: yes

evap/static/css/_quick-review.scss
janno42: its a selector for all elements with the `data-layer` attribute set to `0`. it's correct, don't mind github.

PR #1325, opened 2019-03-25
Apply missing renames. Fixes #1298

### LINKED ISSUE #1298: Consistent permission method naming
we currently have multiple methods checking permissions which don't use consistent naming. e.g.,
- `user_can_see_textanswer(user, textanswer, ...)` (which is defined in `results.views` although it should probably rather be located in `results.tools`)
- `Course.can_user_see_course(user)`
- `Course.can_manager_edit()`

these should be refactored to something like
- `can_user_see_textanswer(user, textanswer, ...)`
- `Course.can_be_seen_by(user)`
- `Course.can_be_edited_by_managers()`

the methods above are just some examples. all models, views and tools should be checked for similar occurrences.

_Originally posted by @karyon in https://github.com/fsr-de/EvaP/pull/1294#discussion_r250199820_

He3lixxx:
fixes #1298. Follow up to #1319

### DIFF
--- a/evap/contributor/templates/contributor_index.html
+++ b/evap/contributor/templates/contributor_index.html
@@ -98,7 +98,7 @@
                             {% for evaluation in evaluations|dictsort:"name" %}
                                 <tr {% if evaluation|is_user_editor_or_delegate:user and evaluation.state == 'prepared' %}
                                         class="{% if course.evaluation_count > 1 %}evaluation-row{% else %}heading-row{% endif %} hover-edit" data-url="{% url 'contributor:evaluation_edit' evaluation.id %}"
-                                    {% elif evaluation.state == 'published' and evaluation|can_user_see_results_page:user %}
+                                    {% elif evaluation.state == 'published' and evaluation|can_results_page_be_seen_by:user %}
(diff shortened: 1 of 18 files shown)


He3lixxx:
@karyon Do we want to add tests for this? At some points buttons were disabled where they should be enabled, at some points some buttons were missing. The second one could probably be tested rather easily the first one would be rather fragile when using only `assertContains`..

karyon:
not sure... if we started adding tests whether all our buttons do actually exist i think we wouldn't get anywhere anytime soon. perhaps it's worth it to do that for "critical" buttons, i don't know :) we could look into making the templates stricter. however there doesn't seem to be much we can do, i found only [this](https://stackoverflow.com/questions/4300442/show-undefined-variable-errors-in-django-templates) (first two answers) and variants thereof. jinja2 seems to have more options, but well, we're probably not going that far...

karyon:
i just noticed we could've catched that if we had looked at the coverage. it went down, which shouldn't happen when renaming things.

janno42:
there is another occurrence missing: `can_user_see_textanswer` in `evap\results\tools.py`

PR #1329, opened 2019-03-29
1327 anonymize answer counters

### LINKED ISSUE #1327: Anonymization script: scramble evaluation results
This is a follow up to #1287. We came up with the idea of scrambling evaluation results as a part of the anonymization. Quote from @janno42:

"then @karyon and i discussed that it is better to randomize answers because there might be the following case: users who have access to production might find unpublished courses in the anonymized data and can find results there based on answer distribution of other courses and de-anonymizing contributor's names. this is a very special case and there wouldn't be much data available as a course is only then not published if less than two participants voted for it - but nevertheless there might be data leaked.
so we decided that answer counters should be set randomly on anonymization. the answer count per question should (roughly) stay the same during the process and it must be made sure that the answer count never exceeds the number of voters in the course."

We had two implementations flying around in that debate:
- One summed up the answer count, iterated over the RatingAnswerCounters and then took a random number between 0 and the remaining answers and assigned it to the current counter. This can be found [here](https://github.com/fsr-de/EvaP/pull/1287/commits/2c20342a61409f560aeea93a0d77e2bec9a5ee42#diff-8966c64d41a29b22e8ee43d34675dc33R220) but is biased towards giving higher vote counts to the answer counters we are looking at earlier
- The other one was given [here](https://github.com/fsr-de/EvaP/pull/1287#pullrequestreview-218545677). Basically, give every counter a random number between 0 and the count of votes, then divide by a factor so that the vote count matches the earlier result again. In order to keep the count of votes equal, one would have to then distribute the error evenly to the counters. This will give more uniform results but has more clean up to do.

He3lixxx:
Well, if you really want to make sure everything is correct in the end, this takes a bit more than what we initially required thought.

I'd understand if we said that this is too much code so we do what @janno42 wanted to do initally (set to random integer between 0 and `answers_left`, then decrease `answers_left`).

We could remove the single result check. I left it in for now because I ran in exactly that problem (giving answers to NO_ANSWER which then is a invalid database state. Our asserts really saved lifes here :D)

fixes #1327

Edit: We should also note that this generates very small variance, almost every evaluation will end up between 2.5 and 3.5

### DIFF
--- a/evap/evaluation/management/commands/anonymize.py
+++ b/evap/evaluation/management/commands/anonymize.py
@@ -95,7 +95,7 @@ class Command(BaseCommand):
             # Give users unique temporary names to counter identity errors due to the names being unique
             if user.username in Command.ignore_usernames:
                 continue
-            user.username = "<User #" + str(i) + ">"
+            user.username = f"<User #{i}>"
             user.save()
 
         # Actually replace all the real user data
@@ -145,7 +145,7 @@ class Command(BaseCommand):
             random.shuffle(courses)
             public_courses = [course for course in courses if not course.is_private]
 
-            self.stdout.write("Anonymizing " + str(len(courses)) + " courses of semester " + str(semester) + "...")
+            self.stdout.write(f"Anonymizing {len(courses)} courses of semester {semester}...")
 
             # Shuffle public courses' names in order to decouple them from the results.
             # Also, assign public courses' names to private ones as their names may be confidential.
@@ -155,8 +155,8 @@ class Command(BaseCommand):
 
             for i, course in enumerate(courses):
                 # Give courses unique temporary names to counter identity errors due to the names being unique
-                course.name_de = "<Veranstaltung #" + str(i) + ">"
-                course.name_en = "<Course #" + str(i) + ">"
+                course.name_de = f"<Veranstaltung #{i}>"
+                course.name_en = f"<Course #{i}>"
                 course.save()
 
             for i, course in enumerate(courses):
@@ -165,16 +165,15 @@ class Command(BaseCommand):
                     course.name_de = name[0]
                     course.name_en = name[1]
                 else:
-                    course.name_de = "Veranstaltung #" + str(i + 1)
-                    course.name_en = "Course #" + str(i + 1)
+                    course.name_de = f"Veranstaltung #{i + 1}"
+                    course.name_en = f"Course #{i + 1}"
                 course.save()
 
     def anonymize_evaluations(self):
         for semester in Semester.objects.all():
             evaluations = list(semester.evaluations.all())
             random.shuffle(evaluations)
-            self.stdout.write("Anonymizing " + str(len(evaluations)) + " evaluations of semester " + str(semester) +
-                "...")
+            self.stdout.write(f"Anonymizing {len(evaluations)} evaluations of semester {semester}...")
 
             self.stdout.write("Shuffling evaluation names...")
             named_evaluations = (evaluation for evaluation in evaluations if evaluation.name_de and evaluation.name_en)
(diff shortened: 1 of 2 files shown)


evap/evaluation/management/commands/anonymize.py
```diff
@@ -165,16 +165,15 @@ def anonymize_courses(self):
...
+                    course.name_en = f"Course #{i + 1}"
                 course.save()
 
     def anonymize_evaluations(self):
         for semester in Semester.objects.all():
             evaluations = list(semester.evaluations.all())
             random.shuffle(evaluations)
-            self.stdout.write("Anonymizing " + str(len(evaluations)) + " evaluations of semester " + str(semester) +
-                "...")
+            self.stdout.write(f"Anonymizing {len(evaluations)} evaluations of semester {semester}....")
```
janno42: three dots would be sufficient :)

evap/evaluation/management/commands/anonymize.py
```diff
@@ -197,23 +201,75 @@ def anonymize_evaluations(self):
...
+                    original_sum = sum(counter.count for counter in counters)
+
+                    missing_values = set(CHOICES[question.type].values).difference(set(c.answer for c in counters))
+                    missing_values.discard(NO_ANSWER)  # don't add NO_ANSWER counter if it didn't exist before
+                    for value in missing_values:
+                        counters.append(RatingAnswerCounter(question=question, contribution=contribution, answer=value, count=0))
+
+                    generated_counts = [random.random() for c in counters]
+                    generated_sum = sum(generated_counts)
+                    generated_counts = [round(count / generated_sum * original_sum) for count in generated_counts]
```
He3lixxx: I think floow + adding up is less code and easier to understand, so will probably change that.

evap/evaluation/management/commands/anonymize.py
```diff
@@ -197,23 +201,75 @@ def anonymize_evaluations(self):
...
+                    generated_sum = sum(generated_counts)
+
+                    while generated_sum != original_sum:
+                        to_add = original_sum - generated_sum
+                        index = random.randint(0, len(generated_counts) - 1)
+
+                        if to_add < 0:  # don't go below zero
```
He3lixxx: This if is unneeded, but will get removed anyway

He3lixxx:
@karyon @janno42 @felixrindt
- Changed the `round` part to a `floor` as distribution the remaining answers is then easier
- Added tests

I'd appreciate a review (or a short info if you don't want to review)

evap/evaluation/management/commands/anonymize.py
```diff
@@ -197,23 +202,68 @@ def anonymize_evaluations(self):
...
+                        counter.count = generated_count
+
+                        if counter.count:
+                            counter.save()
+                        elif counter.id:
+                            counter.delete()
+
+                    assert original_sum == sum(counter.count for counter in counters)
+
+                if contribution.evaluation.is_single_result:
```
janno42: do we need that?

janno42: just asking because it adds database queries :)

He3lixxx: Ah fair enough. Well, I guess if the test covers it, we can remove it?

evap/evaluation/tests/test_commands.py
```diff
@@ -1,21 +1,30 @@
...
+from evap.evaluation.models import (CHOICES, Contribution, Course, Evaluation, EmailTemplate, NO_ANSWER,
+    Question, Questionnaire, RatingAnswerCounter, Semester, UserProfile)
 
 
 class TestAnonymizeCommand(TestCase):
-    @patch('builtins.input')
-    def test_anonymize_does_not_crash(self, mock_input):
+    @staticmethod
+    def get_random_rating_question_type():
+        return random.choice(list(CHOICES.keys()))
```
karyon: it seems uncritical here but generally i would avoid randomness to avoid random failures. not sure what mommy.make allows, but perhaps you can use https://docs.python.org/3/library/itertools.html#itertools.cycle somehow? this would also make sure we're using all the types.

edit: nevermind. the code does not have any conditions based on the question type, it really doesn't matter. same for the other randomness in the other tests.

He3lixxx: I think it's a fair point. It's probably way easier to debug if every execution runs down the same path. I made sure that we always do a `random.seed` before using random so every test run should not always be the same. I also exchanged the question type generation with that `cycle` method, seems like a much better idea there.

PR #1334, opened 2019-04-09
Added auto-completion for manage.py

SilvanVerhoeven:


He3lixxx:
- you have one empty file committed there, guess we don't need that?
- the hash bang in  deployment/manage_autocompletion.sh is wrong (should be `#!`)
- the deployment/manage_autocompletion.sh does not have a newline at the end of the file. Please add one.
- @janno42 usually does `python manage.py` or `python3 manage.py`. Will it work for these commands, too?

janno42:
@SilvanVerhoeven do you want to continue working on this?

SilvanVerhoeven:
Yeah, can finish that

________________________________
Von: Johannes Wolf <notifications@github.com>
Gesendet: Montag, 13. Mai 2019 18:31:11
An: fsr-de/EvaP
Cc: Verhoeven, Silvan; Mention
Betreff: Re: [fsr-de/EvaP] Added auto-completion for manage.py (#1334)


@SilvanVerhoeven<https://github.com/SilvanVerhoeven> do you want to continue working on this?

SilvanVerhoeven:
Better early than late, still better than never...
Dealt with the first three annotations. Added support for the commands python and python3, too, those are pretty hacky though. I don't know of another way

He3lixxx:
Looks good, thanks for updating this :)

Have you tried passing `"python3 manage.py"` as the second parameter to the `complete` command instead of just `python3`? I could imagine that to work.

SilvanVerhoeven:
Yes, I tried "python manage.py". It didn't work out, unfortunately

He3lixxx:
I don't really like the fact that we give autocompletion for python independent of whether manage.py is run or not, but a proper solution probably has much more overhead that has to be maintained, so I'd merge and try this. If it's too bad, I'd improve again.

pixunil:
@SilvanVerhoeven Unfortunately, my tabbing for `./manage.py r` completes to the file `./manage.py requirements`. Am I missing an installation step?

He3lixxx:
@pixunil you will need to either run `vagrant provision` or manually copy the file `deployment/manage_autocompletion.sh` to `/etc/bash_completion.d/` inside the vm - using the provisioning script command for that should work: 
```sudo cp /vagrant/deployment/manage_autocompletion.sh /etc/bash_completion.d/```

PR #1336, opened 2019-05-13
Implement the issue Proxy users #1304

### LINKED ISSUE #1304: Proxy users
UserProfiles should get an additional flag `Proxy user` which, if set
- prevents them from being deleted on user bulk deletion
- removes them from user selection fields for all users but staff (i.e., in the user profile delegate selection, delegate course form and contribution form)

velramiir:
fixes #1304


evap/evaluation/models.py
```diff
@@ -1185,6 +1188,9 @@ class UserProfile(AbstractBaseUser, PermissionsMixin):
     # users to which all emails should be sent in cc without giving them delegate rights
     cc_users = models.ManyToManyField("UserProfile", verbose_name=_("CC Users"), related_name="ccing_users", blank=True)
 
+    # flag for proxy users like FSR or student office which are not actual users themselves, but represent a group of users
```
He3lixxx: "FSR" and "student office" are rather HPI specific. My last information is that in general, we want to use more wide range wording. @janno42?

evap/evaluation/templates/textanswer_visibility_info.html
```diff
@@ -1,7 +1,7 @@
 <span class="far fa-eye ml-1 icon-gray" data-toggle="tooltip" data-placement="left" title="
     {{ intro_text }}
     {% for contributor in visible_by_contribution %}
-        {{ contributor.full_name }}{% if not forloop.last or visible_by_delegation_count > 0 %}, {% endif %}
+        {{ contributor.full_name }}{% if contributor.is_proxy_user %}{% blocktrans count delegate_count=contributor.delegates.count %} ( {{ delegate_count }} person ){% plural %} ( {{ delegate_count }} persons ){% endblocktrans %}{% endif %}{% if not forloop.last or visible_by_delegation_count > 0 %}, {% endif %}
```
He3lixxx: @janno42 says this is not one of the one correct versions to whitespace this (should be `(n persons)` without inner spaces)

evap/evaluation/tests/test_models.py
```diff
@@ -363,6 +367,27 @@ def test_inactive_users_shown(self):
...
+    def test_can_be_marked_inactive_by_manager(self):
+        user = mommy.make(UserProfile)
+        mommy.make(Evaluation, participants=[user], state="new")
+        self.assertFalse(user.can_be_marked_inactive_by_manager)
+
+        contributor = mommy.make(UserProfile)
+        mommy.make(Contribution, contributor=contributor)
+        self.assertFalse(contributor.can_be_marked_inactive_by_manager)
+
+        user2 = mommy.make(UserProfile, groups=[Group.objects.get(name="Reviewer")])
```
He3lixxx: I'd name him `reviewer` and make a new user for the other group (`grade_publisher`)

He3lixxx:
tests still todo

evap/evaluation/templates/textanswer_visibility_info.html
```diff
@@ -1,7 +1,7 @@
 <span class="far fa-eye ml-1 icon-gray" data-toggle="tooltip" data-placement="left" title="
     {{ intro_text }}
     {% for contributor in visible_by_contribution %}
-        {{ contributor.full_name }}{% if not forloop.last or visible_by_delegation_count > 0 %}, {% endif %}
+        {{ contributor.full_name }}{% if contributor.is_proxy_user %}{% blocktrans count delegate_count=contributor.delegates.count %} ({{ delegate_count }} person){% plural %} ({{ delegate_count }} persons){% endblocktrans %}{% endif %}{% if not forloop.last or visible_by_delegation_count > 0 %}, {% endif %}
```
karyon: one or two newlines plz

evap/contributor/forms.py
```diff
@@ -91,4 +93,5 @@ def save(self, *args, **kw):
 
 
 class DelegateSelectionForm(forms.Form):
-    delegate_to = UserModelChoiceField(label=_("Delegate to"), queryset=UserProfile.objects.exclude_inactive_users())
+    delegate_to = UserModelChoiceField(label=_("Delegate to"),
+                                       queryset=UserProfile.objects.exclude_inactive_users().exclude(is_proxy_user=True))
```
karyon: exclude_proxy_users()? i think you might actually inline and remove that method (not sure why we have `exclude_inactive_users` actually, right now i don't see much of a benefit), but if you keep it, lets be consistent in using it :)

evap/evaluation/templates/textanswer_visibility_info.html
karyon: also, `people` is the more common plural of person i think.

evap/staff/forms.py
```diff
@@ -368,7 +368,7 @@ def save(self, *args, **kw):
 
 
 class ContributionForm(forms.ModelForm):
-    contributor = UserModelChoiceField(queryset=UserProfile.objects.exclude_inactive_users())
+    contributor = UserModelChoiceField(queryset=UserProfile.objects.exclude_inactive_users().exclude(is_proxy_user=True))
```
karyon: same here

evap/contributor/forms.py
janno42: we want to exclude inactive users from selection lists - making users inactive has more or less the only effect of not including them into lists anymore.

evap/evaluation/tests/test_models.py
```diff
@@ -363,6 +367,26 @@ def test_inactive_users_shown(self):
         self.assertIn(active_user, user_list)
         self.assertIn(inactive_user, user_list)
 
+    def test_can_be_marked_inactive_by_manager(self):
+        user = mommy.make(UserProfile)
+        mommy.make(Evaluation, participants=[user], state="new")
```
janno42: there should be an assert where `can_be_marked_inactive_by_manager` is `True` in the beginning before adding the user as an participant, to make sure a different outcome is possible :)

evap/contributor/forms.py
karyon: sure. i was wondering why we have that exclude in a separate method, but it's fine.

evap/contributor/forms.py
```diff
@@ -83,12 +85,13 @@ class Meta:
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        self.fields['delegates'].queryset = UserProfile.objects.exclude_inactive_users()
+        self.fields['delegates'].queryset = UserProfile.objects.exclude(is_active=False)
```
He3lixxx: changing the queryset here conflicts with the change above (line 76)

evap/evaluation/tests/test_models.py
```diff
@@ -349,11 +350,14 @@ def test_can_be_deleted_by_manager(self):
...
 
+        proxy_user = mommy.make(UserProfile, is_proxy_user=True)
+        self.assertFalse(proxy_user.can_be_deleted_by_manager)
+
     def test_inactive_users_hidden(self):
         active_user = mommy.make(UserProfile)
         mommy.make(UserProfile, is_active=False)
 
-        self.assertEqual(list(UserProfile.objects.exclude_inactive_users().all()), [active_user])
+        self.assertEqual(list(UserProfile.objects.exclude(is_active=False).all()), [active_user])
```
He3lixxx: When you are already changing this line, could you also remove the .all() call? It should not be required.

evap/results/tools.py
```diff
@@ -296,7 +296,8 @@ def textanswers_visible_to(contribution):
         ).distinct().order_by('last_name', 'first_name')
     else:
         contributors = [contribution.contributor]
-    num_delegates = len(set(UserProfile.objects.filter(represented_users__in=contributors).distinct()) - set(contributors))
+    non_proxy_contributors = [contributor for contributor in contributors if not contributor.is_proxy_user]
+    num_delegates = len(set(UserProfile.objects.filter(represented_users__in=non_proxy_contributors).distinct()) - set(contributors))
```
He3lixxx: I think this changes the character of the method so much that it should be renamed (or copied, in case the old method is still needed somewhere) -- and subsequently also the names of the variables where the results are stored should be renamed...

velramiir: Technically all users are still included in the result, because the proxy users are 'composed' of their delegates. So even though they are deducted from the delegate count one still knows they can see the text answer, because the proxy user they represent is listed as contributor. So  no copying or renaming should be necessary

evap/evaluation/tests/test_models.py
```diff
@@ -363,6 +367,29 @@ def test_inactive_users_shown(self):
...
+        evaluation.save()
+        self.assertFalse(user.can_be_marked_inactive_by_manager)
+
+        contributor = mommy.make(UserProfile)
+        mommy.make(Contribution, contributor=contributor)
+        self.assertFalse(contributor.can_be_marked_inactive_by_manager)
+
+        reviewer = mommy.make(UserProfile, groups=[Group.objects.get(name="Reviewer")])
+        self.assertFalse(reviewer.can_be_marked_inactive_by_manager)
+        grade_publisher = mommy.make(UserProfile, groups=[Group.objects.get(name="Grade publisher")])
```
janno42: please add an empty line

PR #1340, opened 2019-05-27
UI changes for bipolar scales

janno42:
i suggest changing the icons in bipolar scales from plus/minus to carets pointing up and down to avoid associations with plus = good and minus = bad.
bipolar

also, the print layout needs improvements: the icons are barely visible because the shadow is missing and the adjectives are never shown because one can not hover on elements printed on paper.
so we should discuss whether the adjectives should be always shown in print view (or maybe also in the standard view if there is sufficient space?).

### DIFF
--- a/evap/results/templates/distribution_with_grade.html
+++ b/evap/results/templates/distribution_with_grade.html
@@ -14,10 +14,10 @@
         {% endwith %}
     </div>
     {% if question_result.question.is_bipolar_likert_question %}
-        <span class="left pole-icon fas fa-minus"></span>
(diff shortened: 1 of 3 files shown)


He3lixxx:
+/- to up/down is a nice idea :+1:

I'd rather not always have the texts in standard view (weak opinion).

pixunil:
That the adjectives are not visible in print view has a simple reason: a certain someone has forgotten a `@import "print"`... :see_no_evil: I'm so sorry...
Btw, the background is a bit strange to remove in print, one solution is to move the css into a new file and import this file before print.

PR #1342, opened 2019-05-27
Speed up results index

karyon:
By bringing it back to O(1) queries again

This speeds up the request time for students from 8s to 1.5s on my laptop and from 23 to 15s for staff users

Note that this is not the complete loading time for this page. my browser (firefox) needs another six seconds to finish processing/rendering the page.

### DIFF
--- a/evap/results/templates/results_index.html
+++ b/evap/results/templates/results_index.html
@@ -102,9 +102,9 @@
     </div>
     <div class="card card-noflex">
         <div class="card-body pt-0 pb-2">
-            {% if courses %}
+            {% if courses_and_evaluations %}
                 <div id="results-grid">
-                    {% for course, evaluations in courses %}
+                    {% for course, evaluations in courses_and_evaluations %}
                         {# this is the reason why results_index_evaluation does not need to include is_subentry in the cache key: an evaluation is a subentry or no subentry regardless of which user asks #}
                         {% if course.num_evaluations > 1 %}
                             <div>
--- a/evap/results/templates/results_index.html
+++ b/evap/results/templates/results_index.html
@@ -102,19 +102,19 @@
     </div>
     <div class="card card-noflex">
         <div class="card-body pt-0 pb-2">
-            {% if evaluations %}
+            {% if courses %}
                 <div id="results-grid">
-                    {% regroup evaluations by course as course_evaluations %}
-                    {% for course, evaluations in course_evaluations|dictsort:"grouper.name" %}
-                        {% if course.evaluations.count > 1 %}
+                    {% for course, evaluations in courses %}
+                        {# this is the reason why results_index_evaluation does not need to include is_subentry in the cache key: an evaluation is a subentry or no subentry regardless of which user asks #}
+                        {% if course.num_evaluations > 1 %}
                             <div>
                                 {% include 'results_index_course.html' %}
-                                {% for evaluation in evaluations|dictsort:"name" %}
+                                {% for evaluation in evaluations %}
                                     {% include 'results_index_evaluation.html' with links_to_results_page=evaluation|can_results_page_be_seen_by:request.user is_subentry=True %}
                                 {% endfor %}
                             </div>
                         {% else %}
-                            {% for evaluation in evaluations|dictsort:"name" %}
+                            {% for evaluation in evaluations %}
                                 {% include 'results_index_evaluation.html' with links_to_results_page=evaluation|can_results_page_be_seen_by:request.user is_subentry=False %}
                             {% endfor %}
                         {% endif %}
--- a/evap/results/tests/test_views.py
+++ b/evap/results/tests/test_views.py
@@ -1,26 +1,93 @@
 from unittest.mock import patch
 
 from django.contrib.auth.models import Group
+from django.core.cache import caches
+from django.core.management import call_command
 from django.test.testcases import TestCase
+from django.db import connection
+from django.test import override_settings
+from django.test.utils import CaptureQueriesContext
 
 from django_webtest import WebTest
 from model_mommy import mommy
 
 from evap.evaluation.models import (Contribution, Course, Degree, Evaluation, Question, Questionnaire, RatingAnswerCounter,
                                     Semester, UserProfile)
-from evap.evaluation.tests.tools import WebTestWith200Check, let_user_vote_for_evaluation
-from evap.results.views import get_evaluations_with_prefetched_data
+from evap.evaluation.tests.tools import WebTestWith200Check, let_user_vote_for_evaluation, FuzzyInt
+from evap.results.views import get_evaluations_with_prefetched_data, get_course_result_template_fragment_cache_key
 
-import random
-
-
-class TestResultsView(WebTestWith200Check):
+class TestResultsView(WebTest):
     url = '/results/'
-    test_users = ['manager']
 
-    @classmethod
-    def setUpTestData(cls):
-        mommy.make(UserProfile, username='manager', email="manager@institution.example.com")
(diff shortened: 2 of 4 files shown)


evap/results/templates/results_index.html
```diff
@@ -83,7 +83,8 @@ <h6 class="card-subtitle mb-1">{% trans 'Semester' %}</h6>
                     <tbody>
                     {% regroup evaluations by course as course_evaluations %}
                     {% for course, evaluations in course_evaluations|dictsort:"grouper.name" %}
-                        {% if course.evaluations.count > 1 %}
+                        {# we're actually using length here instead of count to avoid a query (`evaluations` is accessed below anyway #}
+                        {% if course.evaluations|length > 1 %}
```
karyon: this is broken, it should be `course.evaluations.all`. i'll probably prefetch the count instead.

evap/results/templates/results_index.html
```diff
@@ -81,15 +81,15 @@ <h6 class="card-subtitle mb-1">{% trans 'Semester' %}</h6>
...
+                    {% for course, evaluations in courses %}
+                        {# this is the reason why results_index_evaluation does not need to include is_subentry in the cache key: an evaluation is a subentry or no subentry regardless of which user asks #}
+                        {% if course.num_evaluations > 1 %}
                             {% include 'results_index_course.html' %}
-                            {% for evaluation in evaluations|dictsort:"name" %}
+                            {% for evaluation in evaluations %}
                                 {% include 'results_index_evaluation.html' with links_to_results_page=evaluation|can_results_page_be_seen_by:request.user is_subentry=True %}
                             {% endfor %}
                         {% else %}
-                            {% for evaluation in evaluations|dictsort:"name" %}
```
karyon: @janno42 can you tell me why those sorts were here? perhaps to match what JS will sort by default so the user sees no change?

evap/results/tests/test_views.py
```diff
@@ -14,13 +19,78 @@
...
 
-class TestResultsView(WebTestWith200Check):
+class TestResultsView(WebTest):
     url = '/results/'
-    test_users = ['manager']
 
-    @classmethod
-    def setUpTestData(cls):
-        mommy.make(UserProfile, username='manager', email="manager@institution.example.com")
+    @patch('evap.evaluation.models.Evaluation.can_be_seen_by', new=lambda self, user: True)
```
He3lixxx: I found that lambda part really hard to parse. The comma looks like it's separating the `new` argument from the next argument (which I thought was called `user`). Can we put parentheses around the `self, user` or do something else to improve this?

evap/results/views.py
```diff
@@ -135,13 +136,26 @@ def index(request):
         additional_evaluations = get_evaluations_with_course_result_attributes(additional_evaluations)
         evaluations += additional_evaluations
 
-    evaluations.sort(key=lambda evaluation: (evaluation.course.semester.pk, evaluation.full_name))  # evaluations must be sorted for regrouping them in the template
+    # put evaluations into a dict that maps from course to a list of evaluations.
+    # this dict is sorted by course.pk (important for the zip below)
+    # (this relies on python 3.7's guarantee that the insertion order of the dict is preserved)
```
He3lixxx: @janno42 do we have python3.7 running in production?

evap/results/views.py
```diff
@@ -135,13 +136,26 @@ def index(request):
...
+    # annotate each course in courses with num_evaluations
+    annotated_courses = Course.objects.filter(pk__in=course_pks).annotate(num_evaluations=Count('evaluations')).order_by('pk').defer()
+    for course, annotated_course in zip(courses.keys(), annotated_courses):
+        course.num_evaluations = annotated_course.num_evaluations
+
+    degrees = Degree.objects.filter(courses__pk__in=course_pks).distinct()
+    course_types = CourseType.objects.filter(courses__pk__in=course_pks).distinct()
     template_data = dict(
-        evaluations=evaluations,
+        courses=courses.items(),
```
He3lixxx: Waybe name the template variable `course_evaluation_tuples` or something? I was confused in the template that you didn't access it with `.items` and only saw later that you do it here.

A rather weak opinion of mine, though, so I'm also fine with leaving it like this.

evap/results/views.py
```diff
@@ -135,13 +136,26 @@ def index(request):
...
+    # this dict is sorted by course.pk (important for the zip below)
+    # (this relies on python 3.7's guarantee that the insertion order of the dict is preserved)
+    evaluations.sort(key=lambda evaluation: evaluation.course.pk)
 
-    evaluation_pks = [evaluation.pk for evaluation in evaluations]
-    degrees = Degree.objects.filter(courses__evaluations__pk__in=evaluation_pks).distinct()
-    course_types = CourseType.objects.filter(courses__evaluations__pk__in=evaluation_pks).distinct()
+    courses = defaultdict(list)
+    for evaluation in evaluations:
+        courses[evaluation.course].append(evaluation)
```
He3lixxx: Have you thought about doing something like ```
evaluation.course.num_evaluations = evaluation.course.num_evaluations + 1 if evaluation.course.num_evaluations else 0 ```
here instead of the query logic below?

evap/results/tests/test_views.py
```diff
@@ -14,13 +19,78 @@
...
+        # then measure the number of queries with one more course and compare
+        make_course_with_evaluations('eggs')
+        call_command("refresh_results_cache")
+        with CaptureQueriesContext(connection) as context:
+            self.app.get(self.url, user="student")
+        num_queries_after = context.final_queries - context.initial_queries
+
+        self.assertEqual(num_queries_before, num_queries_after)
+
+        # django does not clear the LocMemCache in between tests. clear it hear just to be safe.
```
karyon: typo

evap/results/views.py
karyon: yes you do :)

evap/results/views.py
karyon: I don't quite understand, where does `num_evaluations` come from if not from the query logic?

evap/results/views.py
karyon: I called this courses_and_evaluations, but it's still the same in template and view. actually I wanted to do the `.items()` in the template but that does not work, seems like i still haven't understood the rules how method calls work there...

He3lixxx: Hmm, strange. A bit of debugging indicates that using `.items` on a `defaultdict` in a template breaks (returns an empty list), if you convert it to a normal dict first, it works fine.

I guess I'm fine with the current code.

evap/results/views.py
He3lixxx: We can simply count up the count of evaluations from 0 in this loop, it runs once for every evaluation anyway.

Doing it that way instead of querying the database for the number of evaluations would save that query -  but I'm not sure about whether it's better.

karyon: `evaluations` does not contain all evaluations, but is filtered. here i actually need all evaluations. should i add a comment in that regard?

He3lixxx: Ah, I see. No, I think it's fine, I should have just read more context.

He3lixxx:
🎉

He3lixxx:
You removed the `stdout=StringIO()` parameter from the `management.call_command()` call - if was like this before.
```
management.call_command('refresh_results_cache', stdout=StringIO())
```
Did you _want_ to do this? I think I tested it and everything was fine - but with the current master, the stdout output is printed to the console again.

karyon:
No, that was for debugging, i guess. I opened https://github.com/fsr-de/EvaP/pull/1399.

karyon:
thanks for spotting!

PR #1344, opened 2019-05-27
Ignore duplicate users during imports

### LINKED ISSUE #1337: Clean duplicates on user import
When importing a list of users that includes the same person multiple times, these duplicates should be ignored during the import (a message for each ignored user could be shown). Currently the importer throws an exception.

felixrindt:
Now again as a draft. This ist supposed to fix #1337.


karyon:
as discussed, i would prefer a solution earlier in the process, like not even creating the association.

also, if you want to experiment with changing those datastructures to make this easier, feel free! :)

testing the behavior of duplicate lines would be useful i think. i'd prefer to have this in a separate test to not pile edge cases into an undocumented and binary file, which becomes hard to change later.

if you're really cool you could create tests for the other importers as well that assert that the behavior is sane :)

evap/staff/importers.py
```diff
@@ -411,9 +408,10 @@ def process(cls, excel_content, semester, vote_start_datetime, vote_end_date, te
 
 
 class UserImporter(ExcelImporter):
-    def read_one_user(self, data):
+    def read_one_user(self, data, sheet, row):
         user_data = UserData(username=data[0], title=data[1], first_name=data[2], last_name=data[3], email=data[4], is_responsible=False)
-        return user_data
+        if user_data not in self.associations.values():
```
felixrindt: The check is now earlier. What do you think? The lookup in `.values()` might still be too slow...

karyon:
ummm... could you check `process_user`? it seems like that is using a `users` dict to prevent duplication. What does #1337 request that this code does not do?

felixrindt:
> ummm... could you check `process_user`? it seems like that is using a `users` dict to prevent duplication. What does #1337 request that this code does not do?

The check in `process_user` creates an error when there is conflicting data for a single mail. 
What #1337 wants is that completely identical entries get ignored.

felixrindt:
Ok @karyon.
So while `process_users` checks values in `self.users` and the Enrollment importer also works on `self.users`, the `UserImporter` works on `self.associations`, where this additional check is needed.

felixrindt:
This is a way to do it by not using associations on save. I kept the set for warnings at the moment.
https://github.com/felixrindt/EvaP/pull/1

evap/staff/importers.py
```diff
@@ -390,9 +388,19 @@ def process(cls, excel_content, semester, vote_start_datetime, vote_end_date, te
...
+        self._read_emails = set()
+
+    def read_one_user(self, data, sheet, row):
         user_data = UserData(title=data[0], first_name=data[1], last_name=data[2], email=data[3], is_responsible=False)
-        return user_data
+        if user_data.email not in self._read_emails:
+            self.associations[(sheet.name, row)] = user_data
+            self._read_emails.add(user_data.email)
+        else:
+            warningstring = _("A duplicated entry was ignored: {}.".format(user_data.email))
```
felixrindt: I must move the format outside the translation right? :D

evap/staff/importers.py
```diff
@@ -390,9 +388,26 @@ def process(cls, excel_content, semester, vote_start_datetime, vote_end_date, te
...
+        else:
+            orig_sheet, orig_row = self._read_emails[user_data.email]
+            warningstring = _("A duplicated entry in sheet '{sheet}' on row {row} was ignored: {email}. It was first found in sheet '{orig_sheet}' on row {orig_row}.").format(
+                    sheet=sheet.name,
+                    row=row+1,
+                    email=user_data.email,
+                    orig_sheet=orig_sheet,
+                    orig_row=orig_row+1,
+            )
+            self.warnings[self.W_IGNORED].append(mark_safe(warningstring))
```
felixrindt: Looks like this: 
image

evap/staff/importers.py
```diff
@@ -406,7 +424,7 @@ def save_users_to_db(self):
         new_participants = []
         created_users = []
         with transaction.atomic():
-            for (sheet, row), (user_data) in self.associations.items():
+            for user_data in self.users.values():
```
felixrindt: This fixes the crash.

evap/staff/importers.py
```diff
@@ -390,9 +391,26 @@ def process(cls, excel_content, semester, vote_start_datetime, vote_end_date, te
 
 
 class UserImporter(ExcelImporter):
-    def read_one_user(self, data):
+
+    def __init__(self):
+        super().__init__()
+        self._read_user_data_hashes = dict()
```
felixrindt: This is used to report on duplicated lines. @karyon preferred to warn on duplicate lines and not already on duplicate emails. 
We might have preferred to report on duplicate lines in the ExcelImporter Superclass, but that would be a bigger rewrite, I thought.

evap/staff/importers.py
```diff
@@ -141,14 +145,11 @@ def check_column_count(self, expected_column_count):
...
-    def for_each_row_in_excel_file_do(self, parse_row_function):
+    def for_each_row_in_excel_file_do(self, row_function):
         for sheet in self.book.sheets():
             try:
                 for row in range(self.skip_first_n_rows, sheet.nrows):
-                    line_data = parse_row_function(sheet.row_values(row))
-                    # store data objects together with the data source location for problem tracking
-                    self.associations[(sheet.name, row)] = line_data
-
+                    row_function(sheet.row_values(row), sheet, row)
```
felixrindt: I could reverse this change in interface at the moment, it is not needed with the current implementation.

evap/staff/importers.py
```diff
@@ -26,6 +26,9 @@ def __eq__(self, other):
         return (isinstance(other, self.__class__)
             and self.__dict__ == other.__dict__)
 
+    def __hash__(self):
+        return hash(tuple(sorted(self.__dict__.items())))
+
```
felixrindt: needed to be able to hash user data objects.

evap/staff/importers.py
karyon: I don't mind either way.

janno42: ok, it's your choice then, @felixrindt

felixrindt: I'm gonna reverse it then to save on an argument...

felixrindt: then lets keep it

PR #1348, opened 2019-06-08
[WIP] Adding evaluation approval fields

janno42:
this adds additional fields for evaluations to log the approval of editors and managers with timestamp and name of the person.
the information can be used to clarify confusions when having multiple people with the respective permissions and is meant to be the legal basis for any regulations asking for explicit approval by lecturers for the evaluation of their courses.


evap/staff/tests/test_views.py
```diff
@@ -1263,6 +1263,20 @@ def test_last_modified_unchanged(self):
...
+        """
+            Tests if approving the evaluation changes the state and updates the manager approval field
+        """
+        self.assertIsNone(self.evaluation.manager_approval_user)
+
+        page = self.app.get(self.url, user=self.user, status=200)
+        form = page.forms["evaluation-form"]
+        form.submit(name="operation", value="approve")
+
+        self.evaluation = Evaluation.objects.get(pk=self.evaluation.pk)
```
He3lixxx: I'd prefer `self.evaluation.reload_from_db()` here (for me it's faster to understand what's happening)

janno42: might be a problem: https://github.com/fsr-de/EvaP/blob/9a268798a55c9e15b9ffaed6158a33ffc628fc54/evap/evaluation/tests/test_models.py#L378-L380

He3lixxx: hmm interesting. Someone will have to explain to me why exactly it does not work next time we need, but the code is fine for me then.

karyon: it does not work because of the `state` field which is an `FSMField(...., protected=True)`, which forbids any direct (i.e. not using the transitions) modification of the state field. refresh_from_db would perform such a direct modification.

we could probably write our own reload_from_db for that, but it might not be worth it :)

karyon:
the approach seems to be very limited, for example if an evaluation is sent back to the editors for review, the data of their first change will be overwritten, and also any new type of message would imply an additional database field.

I would propose a more generic logging thing, like an EvaluationLogMessage with a timestamp, user, and message (and FK to Evaluation). this could be displayed to the similar to how log files look like. if we want to have this localized, we could either save a string that somehow marks parts of the string as translatable, or save an enum value (one for each reason that a log message is saved) that gets expanded to a string.

also i thought we can't be the first ones to build such a thing, but my google searches haven't turned anything up. also i thought about using somethling like django-simple-history, but that's probably overkill.

janno42:
you're right, it doesn't scale and there are workflows which might break the concept. for the use case i had in mind that's okay though. if we send an evaluation back to editors, they make changes and approve again, i don't care about the first approval anymore. if they didn't approve a second time, managers can see that it isn't approved at this point and probably know why.
when looking for a complete solution, the history thing actually sounds like a good idea to me. managers can now change everything after the editor's approval so it doesn't really tell anything about the configuration of the evaluation that they approved. we would need a complete log about all changes (so really everything that can be changed in an evaluation (and probably also the course)) that were made and which user made them to have that clear. i thought about such a log some time ago and would still consider it to be a good idea and probably the most complete solution, but its also a lot of things to add and lots of data we would produce. what do you say?

janno42:
We decided to implement a log solution, tracking all important changes (like state changes) with timestamp and user. This log will be accessible on the edit page.

janno42:
i will not continue to work on this.

PR #1349, opened 2019-06-08
Random stuff

janno42:

### DIFF
--- a/evap/staff/templates/staff_evaluation_form.html
+++ b/evap/staff/templates/staff_evaluation_form.html
@@ -1,5 +1,7 @@
 {% extends 'staff_evaluation_base.html' %}
 
+{% load evaluation_filters %}
+
 {% block content %}
     {{ block.super }}
     <h3>
@@ -11,7 +13,10 @@
     </h3>
     {% if evaluation_form.instance.id %}
         <div class="mb-2">
-            {% include 'course_badges.html' %}
+            {% with approval_state_values=evaluation.state|approval_state_values %}
+                <span class="badge badge-light"><span class="{{ approval_state_values.icon }}"></span> {{ approval_state_values.description }}</span>
+            {% endwith %}
+            {% include 'course_badges.html' with course=evaluation.course %}
         </div>
     {% endif %}
     <form id="evaluation-form" method="POST" class="form-horizontal multiselect-form select2form">
--- a/evap/staff/templates/staff_semester_view.html
+++ b/evap/staff/templates/staff_semester_view.html
@@ -47,7 +47,7 @@
 
     <div class="card collapsible mb-3">
         <div class="card-header d-flex">
-            <a class="collapse-toggle" data-toggle="collapse" href="#overviewCard" aria-expanded="false" aria-controls="overviewCard" onclick="saveCollapseState()">{% trans 'Overview' %}</a>
+            <a class="collapse-toggle collapsed" data-toggle="collapse" href="#overviewCard" aria-expanded="false" aria-controls="overviewCard" onclick="saveCollapseState()">{% trans 'Overview' %}</a>
             {% if request.user.is_manager %}
                 <div class="ml-auto">
                     {% if not semester.participations_are_archived %}
@@ -64,7 +64,7 @@
                 </div>
             {% endif %}
         </div>
-        <div class="collapse show" id="overviewCard">
+        <div class="collapse" id="overviewCard">
             <div class="card-body">
                 <table class="table">
                     <thead>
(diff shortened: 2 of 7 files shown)


evap/staff/templates/staff_evaluation_form.html
```diff
@@ -11,7 +11,16 @@ <h3>
...
     {% if evaluation_form.instance.id %}
         <div class="mb-2">
-            {% include 'course_badges.html' %}
+            {% if evaluation.state == 'new' %}
+                <span class="badge badge-light"><span class="fas fa-circle icon-yellow"></span> {% trans 'In preparation' %}</span>
+            {% elif evaluation.state == 'prepared' %}
+                <span class="badge badge-light"><span class="far fa-square icon-gray"></span> {% trans 'Awaiting editor review' %}</span>
+            {% elif evaluation.state == 'editor_approved' %}
+                <span class="badge badge-light"><span class="far fa-check-square icon-yellow"></span> {% trans 'Approved by editor' %}</span>
+            {% elif evaluation.state == 'approved' or state == 'in_evaluation' or state == 'evaluated' or state == 'reviewed' or state == 'published' %}
```
He3lixxx: is using only `state` instead of `evaluation.state` correct here?

evap/staff/templates/staff_semester_view.html
```diff
@@ -444,16 +444,16 @@ <h3>
...
         }
         else {
-            $('#overviewCard').collapse('show');
+            $('#overviewCard').collapse('hide');
         }
 
         function saveCollapseState() {
-            var state = window.localStorage['collapse_overview'] === 'true';
-            localStorage['collapse_overview'] = (!state).toString();
+            var state = window.localStorage['show_overview'] === 'true';
```
He3lixxx: Is there a reason to use `window.localStorage` here when we just use `localStorage` everywhere else? Confused me quite a bit.

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -78,6 +78,9 @@
         {% endif %}
     </div>
     {% include 'evaluation_badges.html' with mode='manager' %}
+    {% if evaluation.course.evaluations.count > 1 %}
```
He3lixxx: If performance could be an issue here we might want to think about annotating that instead of manually traversing the relations here -- @karyon?

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -149,13 +149,13 @@
     {% endif %}
 {% elif info_only or not request.user.is_manager %}
     <td style="width: 20%">
-        <span data-toggle="tooltip" data-placement="top" title="{% trans 'Number of participants' %}">
+        <span {% if evaluation.num_participants == 0 %}class="text-danger" {% endif %}data-toggle="tooltip" data-placement="top" title="{% trans 'Number of participants' %}">
```
He3lixxx: (same performance consideration might apply here and below)

evap/staff/templates/staff_evaluation_form.html
janno42: oh boy. it's not - thanks :)

evap/staff/templates/staff_semester_view.html
janno42: i can't see any reason, i'll remove it.

evap/staff/templates/staff_semester_view_evaluation.html
janno42: `num_participants` is already annotated

evap/staff/templates/staff_semester_view_evaluation.html
janno42: annotation done

evap/staff/templates/staff_evaluation_form.html
```diff
@@ -11,7 +11,16 @@ <h3>
...
+            {% if evaluation.state == 'new' %}
+                <span class="badge badge-light"><span class="fas fa-circle icon-yellow"></span> {% trans 'In preparation' %}</span>
+            {% elif evaluation.state == 'prepared' %}
+                <span class="badge badge-light"><span class="far fa-square icon-gray"></span> {% trans 'Awaiting editor review' %}</span>
+            {% elif evaluation.state == 'editor_approved' %}
+                <span class="badge badge-light"><span class="far fa-check-square icon-yellow"></span> {% trans 'Approved by editor' %}</span>
+            {% elif evaluation.state == 'approved' or evaluation.state == 'in_evaluation' or evaluation.state == 'evaluated' or evaluation.state == 'reviewed' or evaluation.state == 'published' %}
+                <span class="badge badge-light"><span class="far fa-check-square icon-green"></span> {% trans 'Approved by manager' %}</span>
+            {% endif %}
+            {% include 'course_badges.html' with course=evaluation.course %}
```
karyon: there's the `statename` filter which produces different strings which are also shown to contributors. please check whether that should be unified.

janno42: hm, good question. the strings used here are the ones from the staff semester page where we're not really using states as a concept for display in the ui anymore but the four categories (preparation, evaluation phase, text answer review, results publishing) in which the evaluations proceed.
i'm not sure and open for suggestions on general progression in this matter.

janno42: one list for staff, one for contributors, or one for both if possible

Issue #1274, opened 2018-10-24
Add degrees to result export


janno42:
The result export should include the degree and course type for each course in separate lines in the excel sheet right below the title of the course.

Furthermore, it should be possible to select for which degrees the results shall be exported. In addition to the current course type selection a degree selection needs to be added. A course will then be included in a specific sheet if it is in one of the selected degrees *and* in one of the selected course types.

karyon:
regarding the export part, the exporter is kind of hard to read, but the actual change should be small. in the exporter there's a small loop that iterates over all courses and outputs their names (the exporter works line by line!), and below that, another loop would have to be added that outputs the degree.

regarding the degree selection, that should work analogously to the course type selection.

janno42:
@Irliung do you still want to work on this or shall i remove your assignment?

PR #1350, opened 2019-06-09
Add degree selection for results export

janno42:
Implements #1274

### DIFF
--- a/evap/results/exporters.py
+++ b/evap/results/exporters.py
@@ -4,7 +4,7 @@ from django.utils.translation import ugettext as _
 
 import xlwt
 
-from evap.evaluation.models import CourseType
+from evap.evaluation.models import CourseType, Degree
 from evap.results.tools import (collect_results, calculate_average_course_distribution, calculate_average_distribution,
                                 distribution_to_grade, get_grade_color)
 
@@ -76,13 +76,13 @@ class ExcelExporter(object):
 
         return filtered_questions
 
-    def export(self, response, course_types_list, include_not_enough_voters=False, include_unpublished=False):
+    def export(self, response, selection_list, include_not_enough_voters=False, include_unpublished=False):
         self.workbook = xlwt.Workbook()
         self.init_styles(self.workbook)
         counter = 1
         course_results_exist = False
 
-        for course_types in course_types_list:
+        for degrees, course_types in selection_list:
             self.sheet = self.workbook.add_sheet("Sheet " + str(counter))
             counter += 1
             self.row = 0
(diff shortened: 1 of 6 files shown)


He3lixxx:
(Check our the codacy issues - I'm not currently sure in what scope list comprehensions can redefine variables, but in the worst case we might create nasty bugs there)

evap/staff/views.py
```diff
@@ -466,15 +466,20 @@ def semester_export(request, semester_id):
     if formset.is_valid():
         include_not_enough_voters = request.POST.get('include_not_enough_voters') == 'on'
         include_unpublished = request.POST.get('include_unpublished') == 'on'
+        degrees_list = []
         course_types_list = []
         for form in formset:
+            if 'selected_degrees' in form.cleaned_data:
+                degrees_list.append(form.cleaned_data['selected_degrees'])
             if 'selected_course_types' in form.cleaned_data:
                 course_types_list.append(form.cleaned_data['selected_course_types'])
```
karyon: when do those ifs return false? can one of the ifs return false and the other true?

evap/staff/templates/staff_semester_export.html
```diff
@@ -18,11 +18,12 @@ <h3>{% trans 'Export' %} {{ semester.name }}</h3>
 
         <div class="card mb-3">
             <div class="card-body">
-                <p>{% trans 'Select the course types you want to export. Add multiple lines to create an export file with multiple sheets.' %}</p>
+                <p>{% trans 'Select the degrees and course types you want to export. An export will include all evaluations of courses that have at least one of the selected degrees and at least one of the selected course types. Add multiple lines to create an export file with multiple sheets.' %}</p>
```
karyon: courses have only one type, so the "at least" doesn't apply there.

evap/results/exporters.py
```diff
@@ -76,13 +76,13 @@ def filter_text_and_heading_questions(self, questions):
 
         return filtered_questions
 
-    def export(self, response, course_types_list, include_not_enough_voters=False, include_unpublished=False):
+    def export(self, response, degrees_list, course_types_list, include_not_enough_voters=False, include_unpublished=False):
```
karyon: i'd suggest to put the expectation that the two lists are of equal length somewhere in the code. perhaps an assert, or using a list of tuples.

evap/staff/templates/staff_semester_export.html
janno42: right

evap/results/exporters.py
janno42: k

evap/staff/views.py
janno42: they would probably return false if nothing was selected but the form is not valid then so this part of the code wouldn't be reached. so it can be removed.

PR #1351, opened 2019-06-09
Semester deletion

### LINKED ISSUE #1201: Fix semester deletion, add additional confirmation
* right now, deleting a semester that has any courses in it fails with a 500. the delete-button should not be enabled in this case in the first place. also, right now archived semesters cannot be deleted, that should be changed. check out `Semester.can_staff_delete`.

* To be precise, a semester should be deletable if it is either archived or has no courses. For archived semesters, their courses should be deleted one by one in `staff.views.semester_delete`

* when deleting archived semesters, there should be some additional confirmation, e.g. `enter "WT 12/13" in the text box below to confirm` since deleting semesters potentially means a rather large amount of data to be deleted at once.

janno42:
fix #1201; allow deleting archived semesters and ask for explicit confirmation

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -58,7 +58,7 @@ class Semester(models.Model):
 
     @property
     def can_be_deleted_by_manager(self):
-        return all(evaluation.can_be_deleted_by_manager for evaluation in self.evaluations.all())
+        return self.evaluations.count() == 0 or (self.participations_are_archived and self.grade_documents_are_deleted and self.results_are_archived)
 
     @property
     def participations_can_be_archived(self):
--- a/evap/evaluation/templates/confirmation_text_modal.html
+++ b/evap/evaluation/templates/confirmation_text_modal.html
@@ -1,0 +1,46 @@
+{% load evaluation_filters %}
+
+<div class="modal fade" id="{{ modal_id }}" tabindex="-1" role="dialog" aria-labelledby="{{ modal_id }}Label" aria-hidden="true">
+    <div class="modal-dialog" role="document">
+        <div class="modal-content">
+            <div class="modal-header">
+                <h5 class="modal-title" id="{{ modal_id }}Label">{{ title }}</h5>
+                <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                    <span class="fas fa-times"></span>
+                </button>
+            </div>
+            <form id="{{ modal_id }}Form">
+                <div class="modal-body">
+                    {{ question|safe }}
+                    <br /><br />
+                    <div class="form-group">
+                        <input type="text" class="form-control" id="{{ modal_id }}ConfirmationText" oninput="{{ modal_id }}CheckValue();" />
+                    </div>
+                </div>
+                <div class="modal-footer">
+                    <button type="button" id="{{ modal_id }}ActionButton" class="btn btn-{{ btn_type }}" data-dismiss="modal" disabled>{{ action_text }}</button>
+                    <button type="button" class="btn btn-light" data-dismiss="modal">{% trans 'Cancel' %}</button>
+                </div>
+            </form>
+        </div>
+    </div>
+</div>
+
+<script type="text/javascript">
+    function {{ modal_id }}Show(dataId, dataLabel) {
+        // call the modal's action function when action button was pressed and give dataId as parameter
+        $('#{{ modal_id }} #{{ modal_id }}ActionButton').unbind().click(function(){ {{ modal_id }}Action(dataId); });
+        // put the dataLabel text in the right place
+        $('#{{ modal_id }} #data-label').text(dataLabel);
+        // show modal
+        $('#{{ modal_id }}').modal();
+    }
+
+    function {{ modal_id }}CheckValue() {
+        if ($('input#{{ modal_id }}ConfirmationText').val() == $('#{{ modal_id }} #data-label').text()) {
+            $('#{{ modal_id }}ActionButton').prop('disabled', false);
+        } else {
+            $('#{{ modal_id }}ActionButton').prop('disabled', true);
+        }
+    }
+</script>
--- a/evap/staff/templates/staff_semester_view.html
+++ b/evap/staff/templates/staff_semester_view.html
@@ -10,7 +10,7 @@
         {% if request.user.is_manager %}
             &nbsp;
             <a href="{% url 'staff:semester_edit' semester.id %}" class="btn btn-sm btn-secondary">{% trans 'Edit' %}</a>
-            <button type="button"{% if not semester.can_be_deleted_by_manager %} disabled{% endif %} onclick="deleteSemesterModalShow({{ semester.id }}, '{{ semester.name|escapejs }}');" class="btn btn-sm btn-danger">{% trans 'Delete' %}</button>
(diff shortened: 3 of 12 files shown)


evap/evaluation/templates/confirmation_text_modal.html
```diff
@@ -0,0 +1,46 @@
...
+            <div class="modal-header">
+                <h5 class="modal-title" id="{{ modal_id }}Label">{{ title }}</h5>
+                <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                    <span class="fas fa-times"></span>
+                </button>
+            </div>
+            <form id="{{ modal_id }}Form">
+                <div class="modal-body">
+                    {{ question|safe }}
+                    <br /><br />
```
He3lixxx: Personally, I'd prefer using just `<br>`. Do we have a policy for this?

Check out https://stackoverflow.com/questions/1946426/html-5-is-it-br-br-or-br

evap/staff/views.py
```diff
@@ -404,6 +404,13 @@ def semester_delete(request):
 
     if not semester.can_be_deleted_by_manager:
         raise SuspiciousOperation("Deleting semester not allowed")
+    for course in semester.courses.all():
```
He3lixxx: Considering performance - can't we remove these two loops and make this O(1) database queries?
Like `RatingAnswerCounter.objects.filter(contribution__evaluation__course__semester=semester).delete()` and so on.

evap/evaluation/templates/confirmation_text_modal.html
```diff
@@ -0,0 +1,46 @@
...
+        // call the modal's action function when action button was pressed and give dataId as parameter
+        $('#{{ modal_id }} #{{ modal_id }}ActionButton').unbind().click(function(){ {{ modal_id }}Action(dataId); });
+        // put the dataLabel text in the right place
+        $('#{{ modal_id }} #data-label').text(dataLabel);
+        // show modal
+        $('#{{ modal_id }}').modal();
+    }
+
+    function {{ modal_id }}CheckValue() {
+        if ($('input#{{ modal_id }}ConfirmationText').val() == $('#{{ modal_id }} #data-label').text()) {
```
He3lixxx: Maybe make this case insensitive? For people actually typing the name that might be easier.

evap/staff/templates/staff_semester_view.html
```diff
@@ -342,9 +342,9 @@ <h3>
...
     {{ block.super }}
     {% trans 'Delete semester' as title %}
-    {% trans 'Do you really want to delete the semester <strong id="data-label"></strong>?' as question %}
+    {% trans 'Do you really want to delete the semester <strong id="data-label"></strong>? All courses and evaluations will be deleted as well as all results. If you are sure, enter the name of the semester below.' as question %}
     {% trans 'Delete semester' as action_text %}
-    {% include 'confirmation_modal.html' with modal_id='deleteSemesterModal' title=title question=question action_text=action_text btn_type='danger' %}
+    {% include 'confirmation_text_modal.html' with modal_id='deleteSemesterModal' title=title question=question action_text=action_text btn_type='danger' %}
     <script type="text/javascript">
         function deleteSemesterModalAction(dataId) {
             $.ajax({
```
He3lixxx: The problem with this ajax thing is that the user will not see that anything is "working" until the request gets completed. I just tested and was quite confused until the redirect happened. Anything we can do about it? Maybe make it a form that is submitted instead of an ajax request?

evap/evaluation/templates/confirmation_text_modal.html
janno42: i prefer `<br />`, because i expect every opening tag to be closed. same with, e.g., `<input />`. we're using the self closing version more often (57 vs 15 times) - we should probably make it consistent everywhere.

evap/evaluation/templates/confirmation_text_modal.html
janno42: it's all about making it not easy :)
i thought it would be better to ask for the exact same name but i can change that if you don't like it.

evap/staff/templates/staff_semester_view.html
janno42: we could make it behave like the contact modal where the action button is disabled and says "sending". i thought about that but didn't really want to add more code for something that barely (if ever) happens. would be a more complete version, though.

evap/evaluation/templates/confirmation_text_modal.html
He3lixxx: I just thought that forcing the correct capitalization will probably make people copy and paste the name from the page. Over time, this would circumvent this check, but since you would not do this every day, it's probably not that much of an issue.

(Its done differently in popular places anyway - Github security check for deleting a repository forces you to enter its name but allows different capitalization. The apt package manager will make you type "Yes, do as I say!" in that exact capitalization when doing something that will probably break the system)

evap/staff/templates/staff_semester_view.html
He3lixxx: Changing the text to "Sending..." and back shouldn't be that much code, I'd definitely vote in favor of doing that.

evap/evaluation/templates/confirmation_text_modal.html
He3lixxx: Fair enough. Follow up issue?

evap/evaluation/models.py
```diff
@@ -58,7 +58,7 @@ def __str__(self):
 
     @property
     def can_be_deleted_by_manager(self):
-        return all(evaluation.can_be_deleted_by_manager for evaluation in self.evaluations.all())
+        return self.evaluations.count() == 0 or (self.participations_are_archived and self.grade_documents_are_deleted and self.results_are_archived)
```
karyon: now there's the slight inconsistency that can_be_deleted_by_manager returns false on an evaluation in a semester that can be deleting, and by deleting the semester a manager could in fact delete the evaluation. not sure what to do about that. or in other words, why (on a conceptual level) does Semester.can_be_deleted_by_manager not check Evaluation.can_be_deleted_by_manager of every evaluation?

evap/evaluation/templates/confirmation_text_modal.html
janno42: general cleanup or just br-tags? i could add a commit for just those right here.

evap/evaluation/models.py
janno42: we could check that every evaluation can be deleted and that the archiving stuff is done, no problem. so just adding the removed part in addition?

karyon: that would need further changes, since Evaluation.can_be_deleted_by_manager  currently checks can_be_edited which is false for archived evaluations (not sure atm whether we want that and i need to go to bed now :) )

janno42: if we want to delete a complete semester and all of its evaluations we probably don't care about the current state of each individual evaluation. there are two "real" use cases i can think of:
1. a manager wants to delete a semester because creating it was a mistake, something went wrong, it was just a test, ...
2. all the data from a semester shall be deleted because of data privacy reasons, deleting regulations after a certain time, or similar.

deleting a semester is made "hard" in the ui because all archiving steps must have been applied (which already delete some data, so managers should be aware what they are doing) and then this explicit confirmation thing must be dealt with - so mistakenly deleting data is more or less not a problem.
that means the managers really want to have the data removed from the platform and in this situation they don't really care whether some evaluations are not finished or whether some results are not published and so on. so i would argue it's okay to just delete everything without further detailed checks.

evap/evaluation/templates/confirmation_text_modal.html
He3lixxx: So far I've only noticed the br tags, might be the same for some of the other void tags though (https://www.w3.org/TR/html5/syntax.html#void-elements)

PR #1352, opened 2019-06-24
Responsive views for login and results

pixunil:
Uses a grid-based layout which will adjust for smaller screen sizes. See #883.

#### Done
- [x] Login works for `xs`
- [x] Result index wraps rows intelligently
- [x] Result index has ordering feature without datatable
- [x] Result index has search  input at top / left
- [x] Result detail works for `xs`
- [x] Improved tooltips with changing orientation and color / icon legend

<p float="left">
login-en
login-de
result-filter
result-order
result-order-column
result-detail
</p>


He3lixxx:
Cool! Code looks good to me, I tested around on the webpage a bit and noticed two things:
- For me (Firefox 68 on linux), the ordering on the results page does not work.
- On the login page, the two buttons on the external users card have no padding / margin when they are stacked vertically.

He3lixxx:
We agreed on changing the sorting to be case insensitive.

pixunil:
@janno42 I tweaked some little things, so this PR is now feature-complete and fully reviewable.

evap/results/templates/results_index_course.html
```diff
@@ -24,7 +24,7 @@
     <div data-col="responsible" data-order="{{ course.responsibles.first.last_name|lower }}">
         {{ course.responsibles_names }}
     </div>
-    <div data-col="result" data-order="{{ course.avg_grade }}">
+    <div data-col="result" data-order="{{ course.avg_grade|default:7 }}">
```
pixunil: As I fetch the key via `.data("order")`, jQuery converts the values into an integer. But, a `"None"` keeps a `"None"`. To have consistent sorting behavior, I convert `None` to an imaginary grade. Possibly this should not be done in the template but in the tools.

Yes, in Js both `"None" >= 4` and `"None" <= 4` are `false`…

evap/results/templates/results_index_evaluation.html
```diff
@@ -57,14 +57,14 @@
         </div>
     {% endif %}
     {% if evaluation.is_single_result %}
-        <div data-col="participants" class="text-center" data-order="100"><span class="fas fa-user"></span>&nbsp;{{ evaluation.single_result_rating_result.count_sum }}</div>
+        <div data-col="participants" class="text-center"><span class="fas fa-user"></span>&nbsp;{{ evaluation.single_result_rating_result.count_sum }}</div>
     {% else %}
         {% with num_participants=evaluation.num_participants num_voters=evaluation.num_voters %}
-            <div data-col="participants" data-order="{{ num_voters|percentage_value:num_participants|default:'0' }}">{% include 'progress_bar.html' with done=num_voters total=num_participants %}</div>
```
pixunil: As we don't know how to sort the participants column, I removed the attributes. However, `percentage_value` now seams unused. Should I remove?

janno42: yes please

evap/results/templates/results_index_course.html
janno42: to have the default there is fine for me

evap/results/templates/results_index.html
```diff
@@ -51,71 +60,77 @@ <h6 class="card-subtitle mb-1">{% trans 'Semester' %}</h6>
...
+                    <label class="btn btn-light btn-lg active">
+                        <input type="radio" name="result-sorting-order" value="asc" />
+                        <span class="fas fa-sort-alpha-down"></span>
+                    </label>
+                    <label class="btn btn-light btn-lg">
+                        <input type="radio" name="result-sorting-order" value="desc" />
+                        <span class="fas fa-sort-alpha-up-alt"></span>
+                    </label>
                 </div>
+                <select name="result-sorting-column" class="no-select2">
```
janno42: please add `form-control`

PR #1353, opened 2019-06-25
Change vagrant user to evap, test update_production backup.

### LINKED ISSUE #1240: Importing a backup made by update_production.sh does not work flawlessly.
Last week we wanted to do a production update. The json dump file created during that update could not be imported without issues:
- The dump does not contain the cronjob user, but foreign key references to it. This can not be imported
- The dump contains data included by django by default (auth, permission, ...). These need to be excluded when importing.

There should be some kind of documentation on what needs to be executed to import this dump back into the database. We should also add some test (could probably just run on travis) that ensures this always works (dump, flush database, migrate, load dump).

He3lixxx:
- [x] fixes #1240 (creates a restore_backup script and lets the process run once in travis)
- [x] for testing, creates an "evap" user in the vagrant vm and uses that. This is more similar to the production server (and necessary for testing update_production)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,14 @@
 language: python
 python: "3.7"
 dist: xenial
-sudo: false
+sudo: required
 cache:
   directories:
   - $HOME/.cache/pip
 
 services:
-- postgresql
+  - postgresql
+  - redis-server
 before_script:
   - psql -c 'create database evap;' -U postgres
 install:
@@ -18,5 +19,10 @@ script:
   - python manage.py test --keepdb evap.evaluation.tests.test_misc.TestDataTest.load_test_data
   - python manage.py test --keepdb --reverse
   - python manage.py test --keepdb --debug-mode
+  # create a backup and load it again
+  - python manage.py migrate
+  - python manage.py loaddata test_data
+  - EVAP_OVERRIDE_BACKUP_FILENAME=true EVAP_SKIP_CHECKOUT=true EVAP_RUNNING_INSIDE_TRAVIS=true deployment/update_production.sh backup.json
+  - echo "yy" | EVAP_RUNNING_INSIDE_TRAVIS=true deployment/load_production_backup.sh backup.json
 after_success:
   - coveralls
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -5,7 +5,7 @@ Vagrant.require_version ">= 1.8.1"
 
 Vagrant.configure("2") do |config|
   config.vm.box = "ubuntu/bionic64"
-  config.vm.box_version = "= 20190112.0.0 "
+  config.vm.box_version = "= 20190621.0.0 "
 
   # port forwarding
   config.vm.network :forwarded_port, guest: 8000, host: 8000 # django server
@@ -22,6 +22,17 @@ Vagrant.configure("2") do |config|
 
     # show virtualbox gui, uncomment this to debug startup problems
     #v.gui = true
+  end
+
+  # override username to be evap instead of vagrant, just as it is on production.
+  # This is necessary so management script can assume evap is the correct user to
+  # execute stuff as.
+  # Mounting with uid and gid is necessary as the provision script will create
+  # this user, so mounting before does not work with an owner specified by name.
+  # Also, provision needs to use vagrant as ssh user (since evap does not exist yet)
+  config.vm.synced_folder ".", "/evap", mount_options: ["uid=1042", "gid=1042"]
+  if ARGV[0] == "ssh"
+    config.ssh.username = 'evap'
   end
 
   config.vm.provision "shell", path: "deployment/provision_vagrant_vm.sh"
--- a/deployment/provision_vagrant_vm.sh
+++ b/deployment/provision_vagrant_vm.sh
@@ -1,5 +1,11 @@
 set -x # print executed commands
 
+USER="evap"
+REPO_FOLDER="/home/$USER/evap"
+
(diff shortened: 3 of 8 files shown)


deployment/provision_vagrant_vm.sh
```diff
@@ -21,16 +26,24 @@ apt-get -q install -y redis-server
 # install apache
 apt-get -q install -y apache2 apache2-dev
 
+# make user, create home folder, set uid 1000 (required for becoming the synced folder owner), set default shell to bash
+sudo useradd -m -u 1042 -s /bin/bash evap
+# allow ssh login
+sudo cp -r /home/vagrant/.ssh /home/$USER/.ssh
+sudo chown -R $USER:$USER /home/$USER/.ssh
+# allow sudo without password
+echo "evap ALL=(ALL) NOPASSWD:ALL" | sudo tee /etc/sudoers.d/evap
```
He3lixxx: use $USER here

deployment/provision_vagrant_vm.sh
```diff
@@ -21,16 +26,24 @@ apt-get -q install -y redis-server
 # install apache
 apt-get -q install -y apache2 apache2-dev
 
+# make user, create home folder, set uid 1000 (required for becoming the synced folder owner), set default shell to bash
```
karyon: replace "uid 1000" by "the same uid that's specified in Vagrantfile"?

He3lixxx:
Alrighty, so I really really do not want to put any logic into the script that allows executing anything that might damage data without checking back with the user. For the reset_db and flush calls, this did not work out (tried to get it to for several hours), so now I fell back onto doing that by selectively passing `--noinput` to these calls. I don't like it, but I think it's better than not having a test.
@janno42 @karyon what do you think?

Test "logic" is in the .travis.yml, the part I don't like is the `ADDITIONAL_ARGUMENT` stuff in deployment/load_production_backup.sh

He3lixxx:
Three points to think about:
- Currently we are backing up and importing back an empty database - do we want to have some meaningful data in it (At least the test finished very quick, although the TestDataTest is executed before)
- Currently, we are not asserting that the data that is imported back is actually what we expect, we only check whether the import script has exit code 0.
- We probably want to move the test "logic" into a proper test class, where it relies on subprocess.open or something similar, especially when we add more logic.

Any thoughts on this? For now, I'd ignore points 1 and 2 and only fix 3 (since this is the same we do with the load_data test and adding better assertions seems to be complex to me).

.travis.yml
```diff
@@ -18,5 +19,9 @@ script:
   - python manage.py test --keepdb evap.evaluation.tests.test_misc.TestDataTest.load_test_data
   - python manage.py test --keepdb --reverse
   - python manage.py test --keepdb --debug-mode
+  # create a backup and load it again
+  - python manage.py migrate
+  - EVAP_OVERRIDE_BACKUP_FILENAME=true EVAP_SKIP_CHECKOUT=true EVAP_RUNNING_INSIDE_TRAVIS=true deployment/update_production.sh backup.json
```
karyon: loaddata here?

Vagrantfile
```diff
@@ -24,5 +24,13 @@ Vagrant.configure("2") do |config|
     #v.gui = true
   end
 
+  # override username to be evap instead of vagrant. Mounting with uid and gid is necessary as the provision script
```
karyon: and some comments on the "why" would be nice.

deployment/update_production.sh
```diff
@@ -1,37 +1,56 @@
...
-if [ $# -eq 1 ] # if there is exactly one argument
-    then
-        echo Creating backup.
-        set -x # print executed commands. enable this here to not print the if above.
-        sudo -H -u evap env/bin/python manage.py dumpdata --indent 2 --output $1
-    else
-        echo No backup file specified, skipping backup creation.
-        set -x # print executed commands
-fi
+sudo -H -u $USERNAME $ENVDIR/bin/python manage.py dumpdata --natural-foreign --natural-primary --all -e contenttypes -e auth.Permission --indent 2 --output $FILENAME
```
karyon: could you try putting that into dump_testdata?

He3lixxx: It changes ID handling for the users: The users will not be referenced by id anymore but by username. The dump will not contain the old user id.

karyon: hm, is that a problem?

He3lixxx: Nope. Will add these parameters to the dump_testdata command.

.travis.yml
He3lixxx: added

Vagrantfile
He3lixxx: added something, please check again.

He3lixxx:
We thought about making these commands into management commands a bit more (as we thought it might be much better to test then), but we think there are two main points of failure:
- If these are managements commands, they are executed as the `evap` user. However, this user does not have sudo privileges. So, stuff like stopping and starting apache would not work. Additionally, checking out another commit and doing things like `pip install` while the manage.py file is running is somewhat smelly.
- If these are python files out of the management cycle, transactions in test cases get in our way. If, in a test case, we populate the database and then call any of these scripts, the test case transaction is still not committed, so the script will still see an empty database. This stops up from efficiently testing the scripts.

karyon:
otherwise, lgtm!

He3lixxx:
Does anyone else want to review this (@janno42?, @felixrindt?) Otherwise, I'd need an approval to be able to merge @karyon

Issue #1229, opened 2018-07-21
Tests for results cache


janno42:
we should add tests for the results page cache introduced in #1149.
these test ideas are taken from the pr and haven't been implemented yet:

- [ ] test performance for all user groups
- [ ] test performance on crappy laptop
- [ ] any correctness tests?
- [ ] write performance tests

we had ideas for some tests but it seems like testing for different languages and different users requires the tests to be in a specific order and we'd like to understand this situation first, before adding the respective tests.

He3lixxx:
More specifically, we tried these tests, but they failed at the annotated positions:
<details>
<summary>Click to expand the code</summary>

results/tests/test_views.py:TestResultsView:
```
def test_cache_entry_created_on_page_load(self):
    self.assertIsNone(caches['results'].get(get_course_result_template_fragment_cache_key(self.course.id, "en", False)))
    self.assertIsNone(caches['results'].get(get_course_result_template_fragment_cache_key(self.course.id, "en", True)))
    self.assertIsNone(caches['results'].get(get_course_result_template_fragment_cache_key(self.course.id, "de", False)))
    self.assertIsNone(caches['results'].get(get_course_result_template_fragment_cache_key(self.course.id, "de", True)))

    response = self.app.get("/results/", user="student", extra_environ={'HTTP_ACCEPT_LANGUAGE': 'en'}, status=200)
    self.assertEqual(response.status_code, 200)

    response = self.app.get("/results/", user="staff", extra_environ={'HTTP_ACCEPT_LANGUAGE': 'en'}, status=200)
    self.assertEqual(response.status_code, 200)

    response = self.app.get("/results/", user="student", extra_environ={'HTTP_ACCEPT_LANGUAGE': 'de'}, status=200)
    self.assertEqual(response.status_code, 200)

    response = self.app.get("/results/", user="staff", extra_environ={'HTTP_ACCEPT_LANGUAGE': 'de'}, status=200)
    self.assertEqual(response.status_code, 200)

    self.assertIsNotNone(caches['results'].get(get_course_result_template_fragment_cache_key(self.course.id, "en", False)))
    self.assertIsNotNone(caches['results'].get(get_course_result_template_fragment_cache_key(self.course.id, "en", True)))

    # breaks here, looks like the HTTP_ACCEPT_LANGUAGE is ignored
    self.assertIsNotNone(caches['results'].get(get_course_result_template_fragment_cache_key(self.course.id, "de", False)))
[...]

PR #1355, opened 2019-07-01
Save the language in the user's session on first visit

karyon:
This resolves the following inconsistency: If a user first visits the page, the brower's language gets written to user.language, but is not saved in the session. If the user now changes the browser's language, EvaP's language will also change. However, if the user logs out and logs in again, changing the browser's language will not change EvaP's language, because on the second login, the user.language will be written to the session.

@He3lixxx tripped over this while writing tests for the results cache, see #1229.

### DIFF
--- a/evap/staff/tests/test_views.py
+++ b/evap/staff/tests/test_views.py
@@ -352,12 +352,15 @@ class TestSemesterView(WebTest):
             course=mommy.make(Course, name_de="B", name_en="A", semester=cls.semester))
 
     def test_view_list_sorting(self):
-        page = self.app.get(self.url, user='manager', extra_environ={'HTTP_ACCEPT_LANGUAGE': 'en'}).body.decode("utf-8")
+        UserProfile.objects.filter(username='manager').update(language='en')
+        page = self.app.get(self.url, user='manager').body.decode("utf-8")
         position_evaluation1 = page.find("Evaluation 1")
         position_evaluation2 = page.find("Evaluation 2")
         self.assertGreater(position_evaluation1, position_evaluation2)
+        self.app.reset() # language is only loaded on login, so we're forcing a re-login here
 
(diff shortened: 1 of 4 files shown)


evap/staff/tests/test_views.py
```diff
@@ -352,15 +353,19 @@ def setUpTestData(cls):
...
         self.assertGreater(position_evaluation1, position_evaluation2)
+        self.app.reset() # language is only loaded on login, so we're forcing a re-login here
 
-        page = self.app.get(self.url, user='manager', extra_environ={'HTTP_ACCEPT_LANGUAGE': 'de'}).body.decode("utf-8")
+        UserProfile.objects.filter(username='manager').update(language='de')
+        page = self.app.get(self.url, user='manager').body.decode("utf-8")
         position_evaluation1 = page.find("Evaluation 1")
         position_evaluation2 = page.find("Evaluation 2")
         self.assertLess(position_evaluation1, position_evaluation2)
+        translation.deactivate()
```
He3lixxx: Why is this necessary?

karyon: good point. I added a comment and added the same line to the other tests where it's needed.

He3lixxx: The comments are good, but why do we need to reset the translation? I mean, shouldn't the app we are querying only keep the language based on the user (so sending another request with another user resets it)? Or is the issue that if the next request is not made with a specific user, it will keep this language activated? That would be quite strange, wouldn't it? I wouldn't expect the actual server to behave that way.

In general, this feels wrong to me, I would expect it work as if the translation is automatically reset after each test...

karyon: You are right, this was weird and there was a fix for this: the LocaleMiddleware was after the AuthenticationMiddleware, therefore the set_or_get_language picked up whatever language was active in the thread before the LocaleMiddleware had any chance to set a new language according to the request (and, if the request does not give any hints, it would choose the default setting). Fixing the middleware order made the manual deactivations unnecessary.

PR #1356, opened 2019-07-08
Update to django version 2.2

He3lixxx:
In 3.1, all `group_by` queries will _not_ use Meta.ordering anymore in the returned result. This is probably what we mostly want, but this produces a lot of warnings from 2.2 on when running the tests with `-Wa`. We probably should check all these warnings some time between merging this and updating to 3.1 (should I create an issue? Probably easiest to deal with this when doing the actual update...)


karyon:
have you seen the test failures already? perhaps it becomes necessary to use a current version of model mommy (this is what has been holding us back: https://github.com/berinhard/model_mommy/issues/338)

PR #1360, opened 2019-09-16
 Remove AtLeastOne contributor constraint

### LINKED ISSUE #1347: Remove AtLeastOne contributor constraint
We have a constraint that every evaluation must have at least one contributor. This was useful to define at least one person with access to the results.
Now, evaluations belong to courses which must have at least one responsible. This responsible can access the results also when not explicitly being added as a contributor. When creating exam evaluations we don't want to add additional questions and use contributions only for visibility permissions. If only the responsible is involved, we currently unnecessarily have to add the responsible as a contributor. So the constraint should be removed.

felixrindt:
fixes #1347

### DIFF
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -511,7 +511,7 @@ class AtLeastOneFormSet(BaseInlineFormSet):
             raise forms.ValidationError(_('You must have at least one of these.'))
 
 
-class ContributionFormSet(AtLeastOneFormSet):
+class ContributionFormSet(BaseInlineFormSet):
     def __init__(self, data=None, *args, **kwargs):
         data = self.handle_moved_contributors(data, **kwargs)
         super().__init__(data, *args, **kwargs)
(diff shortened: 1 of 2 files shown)


evap/staff/views.py
```diff
@@ -1458,8 +1460,6 @@ def user_import(request):
         success_messages=success_messages, warnings=dict(warnings), errors=errors, test_passed=test_passed))
 
 
-@manager_required
-def user_edit(request, user_id):
     @receiver(RewardPointGranting.granted_by_removal)
```
He3lixxx: pls outdent

felixrindt: resolved

evap/staff/views.py
```diff
@@ -793,19 +793,21 @@ def evaluation_edit(request, semester_id, evaluation_id):
     return helper_evaluation_edit(request, semester, evaluation)
 
 
+
+@receiver(RewardPointGranting.granted_by_removal)
+def notify_reward_points(grantings, **_kwargs):
+    for granting in grantings:
+        messages.info(request,
```
felixrindt: ah we have no request here

He3lixxx:
Code is fine for me

@janno42 do we want to have a test for creating an evaluation without a contributor.

felixrindt:
there's a test now

PR #1366, opened 2019-10-14
OpenID login

janno42:
this implements the option to activate an openid login, replacing the current kerberos solution.
the documentation will be updated after merging.

### DIFF
--- a/evap/evaluation/auth.py
+++ b/evap/evaluation/auth.py
@@ -1,11 +1,15 @@
 from functools import wraps
+import unicodedata
 
+from django.conf import settings
 from django.core.exceptions import PermissionDenied
 from django.contrib.auth.backends import ModelBackend
 from django.contrib.auth.views import redirect_to_login
 from django.utils.decorators import available_attrs
+from mozilla_django_oidc.auth import OIDCAuthenticationBackend
 
 from evap.evaluation.models import UserProfile
+from evap.evaluation.tools import clean_email
 
 
 class RequestAuthUserBackend(ModelBackend):
@@ -149,3 +153,36 @@ def reward_user_required(view_func):
         from evap.rewards.tools import can_reward_points_be_used_by
         return can_reward_points_be_used_by(user)
     return user_passes_test(check_user)(view_func)
+
+
+# see https://mozilla-django-oidc.readthedocs.io/en/stable/
+class OpenIDAuthenticationBackend(OIDCAuthenticationBackend):
+    def filter_users_by_claims(self, claims):
+        email = claims.get('email')
+        if not email:
+            return []
+
+        try:
+            return [UserProfile.objects.get(email=clean_email(email))]
+        except UserProfile.DoesNotExist:
+            return []
+
+    def create_user(self, claims):
+        user = UserProfile.objects.create(
+            username = generate_username_from_email(claims.get('email')),
+            email = claims.get('email'),
+            first_name = claims.get('given_name', ''),
+            last_name = claims.get('family_name', ''),
+        )
+        return user
+
+    def update_user(self, user, claims):
+        user.email = claims.get('email')
+        user.first_name = claims.get('given_name', '')
+        user.last_name = claims.get('family_name', '')
+        user.save()
+        return user
+
+
+def generate_username_from_email(email):
+    return unicodedata.normalize('NFKC', email).split('@')[0].lower()
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -21,7 +21,7 @@ from django.utils.translation import ugettext_lazy as _
 from django.urls import reverse
 from django_fsm import FSMField, transition
 from django_fsm.signals import post_transition
-from evap.evaluation.tools import date_to_datetime, get_due_evaluations_for_user, translate
+from evap.evaluation.tools import clean_email, date_to_datetime, get_due_evaluations_for_user, translate
 
 logger = logging.getLogger(__name__)
 
@@ -1164,7 +1164,7 @@ class UserProfileManager(BaseUserManager):
         user = self.create_user(
             username=username,
             password=password,
-            email=email,
+            email=self.normalize_email(email),
             first_name=first_name,
             last_name=last_name
         )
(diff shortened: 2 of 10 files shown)


evap/settings.py
```diff
@@ -178,6 +183,10 @@
             'level': 'DEBUG',
             'propagate': True,
         },
+        'mozilla_django_oidc': {
+            'handlers': ['console'],
+            'level': 'DEBUG'
```
karyon: I'm not sure anymore why the other loggers have `propagate: True`, but perhaps we want this here as well?

evap/evaluation/auth.py
```diff
@@ -149,3 +152,44 @@ def check_user(user):
...
+
+# see https://mozilla-django-oidc.readthedocs.io/en/stable/
+class OpenIDAuthenticationBackend(OIDCAuthenticationBackend):
+    def filter_users_by_claims(self, claims):
+        email = claims.get('email')
+        if not email:
+            return self.UserModel.objects.none()
+
+        user_email = email
+        for original_domain, replaced_domain in settings.INSTITUTION_EMAIL_REPLACEMENTS:
```
karyon: perhaps add a comment on why the replacement is necessary?

evap/settings.py
```diff
@@ -355,6 +367,16 @@
     "Enhances vibrant academic programs",
 ]
 
+
+### OpenID Login
+ACTIVATE_OPEN_ID_LOGIN = True
```
karyon: false by default? because otherwise every developer might have to reprovision to get the new localsettings template?

evap/evaluation/models.py
```diff
@@ -1177,7 +1177,7 @@ def create_superuser(self, username, password, email=None, first_name=None, last
 class UserProfile(AbstractBaseUser, PermissionsMixin):
     username = models.CharField(max_length=255, unique=True, verbose_name=_('username'))
 
-    # null=True because users created through kerberos logins and certain external users don't have an address.
+    # null=True because certain external users don't have an address
```
karyon: perhaps keep that comment because getting the email is not guaranteed?

evap/evaluation/auth.py
```diff
@@ -149,3 +152,44 @@ def check_user(user):
...
+            if email.endswith(original_domain):
+                user_email = email[:-len(original_domain)] + replaced_domain
+                break
+
+        try:
+            return UserProfile.objects.filter(email=user_email)
+        except UserProfile.DoesNotExist:
```
karyon: I think `filter` never throws that (`get` does)

evap/settings.py
karyon: also, have you decided to not log this to file and mail_admins, and if yes why?

evap/evaluation/auth.py
```diff
@@ -149,3 +152,44 @@ def check_user(user):
...
+            return self.UserModel.objects.none()
+
+    def create_user(self, claims):
+        user = super(OpenIDAuthenticationBackend, self).create_user(claims)
+
+        user.email = claims.get('email')
```
He3lixxx: We want to set the username from the part of the email before the "@" here afaik.

evap/settings.py
janno42: i updated the configuration to propagate and use file and mail_admins as well.

evap/evaluation/models.py
janno42: we discussed yesterday that the username is created from the email, so if we don't get the email from openid, we can't create the userprofile at all.
i think it should be safe to assume to get the email. if that should turn out to be wrong, we have to make some further changes and handle this case explicitly.

evap/evaluation/auth.py
janno42: i don't know why i put a `filter` there, should be `get`

janno42: now i know: the return is expected to be a list (or QuerySet). so i made it a one element list now.

He3lixxx: Strange: In the documentation [here](https://mozilla-django-oidc.readthedocs.io/en/stable/installation.html#connecting-oidc-user-identities-to-django-users) they show an example where the user object is returned. Also, wouldn't it make sense to then also do this at the other returns (`return []`)?

He3lixxx: Hmm, relevant code is here: https://github.com/mozilla/mozilla-django-oidc/blob/3c3c4a4e3b991c4113c5424716e79c001d5b88e9/mozilla_django_oidc/auth.py#L315
That definitely expects something list-like (even though it throws an error when there is more than one element in the list :D)
I guess I'll open an issue on mozilla-django-oidc about this. I'd still argue we should make all the returns consistent, so the others should return empty lists.

janno42: i'm returning empty lists now

janno42:
i addressed all of the above mentioned points:
1 and 2) the email is now processed on creating the user profile so in the database we'll always have the version with replaced domains.
3) the `UserProfile` is now created manually without calling the super method to avoid any possible mistakes. probably we don't need the definition of `OIDC_USERNAME_ALGO` now anymore, should i still keep it in the settings?

evap/settings.py
```diff
@@ -59,6 +59,11 @@
 
 INTERNAL_USERNAMES_MAX_LENGTH = 20
 
+# List of tuples defining email domains that should be replaced when using OpenID login.
```
He3lixxx: The "when using OpenID" part might be misleading - if I understood it correctly, the current code will replace the email domain on a model level, so this will happen whenever a UserProfile is saved, no matter whether OpenID login is used or not.

evap/settings.py
```diff
@@ -355,6 +368,25 @@
     "Enhances vibrant academic programs",
 ]
 
+
+### OpenID Login
+# replace 'example.com', OIDC_RP_CLIENT_ID and OIDC_RP_CLIENT_SECRET with real values in localsettings when activating
+ACTIVATE_OPEN_ID_LOGIN = False
+OIDC_RENEW_ID_TOKEN_EXPIRY_SECONDS = 60 * 60 * 24 * 7  # one week
+OIDC_USERNAME_ALGO = 'evap.evaluation.auth.generate_username_from_email'
```
He3lixxx: You asked whether to keep this - personally, I'd prefer to remove it. Right now, if user creation code of the OIDC plugin runs unexpectedly, it would probably add a user without cleaning the email correctly. If we make it an empty string, it should fail verbosely.

PR #1367, opened 2019-10-16
Replace DataTables dependency with unified implementation

### LINKED ISSUE #1103: Inline datatables localization files to speed up first paint
Right now, datatables gets the localization file in form of a URL (see [datatables.html](https://github.com/fsr-itse/EvaP/blob/master/evap/evaluation/templates/datatables.html)), that is, it starts an ajax request when it starts processing the tables, and waits with processing until the result has been received.

both locales should be included into the compressed javascript or inlined into the html template so they are loaded earlier.

we do something similar for the [bootstrap datetimepicker](https://github.com/fsr-itse/EvaP/blob/028b6301e3eed446d93ae8675030d82c68d46886/evap/evaluation/templates/bootstrap_datetimepicker.html). unfortunately, it's not that easy in this case, since the localization files are json files, not javascript files.

one approach would be to turn the json files to js files, and simply putting the datastructure inside into a variable with the name of the corresponding locale.

pixunil:
We used DataTables at four places:

1. Results index (with custom filtering and search)
2. Semester view (with custom filtering)
3. User view
4. Grades view

As the Results index wasn't a table after #1352 anymore, the sorting was implemented on our side. The logic is pretty much the same like DataTable's and combined with the fact that we already had to implement custom filtering at two places, this is an attempt to remove DataTables in favor for a custom, but unified implementation.

Little overview of the architecture in `data-grid.js` which contains the logic:

- The base class `DataGrid` is an abstract view of a grid-like container. It does not know how to do certain things (like finding the header or look up a cell by column name). These operations are implemented via a template method.
- `TableGrid` and `EvaluationGrid` are `<table>`-based: They will name columns after their index and look for a `<thead>` for header information and a `tbody` for the rows.
- `ResultGrid` is grid based: It will name the columns after their name. It separates head and container, so both need to be supplied via `attachTo`
- Everything which represents the current sorting / filtering state is put into `this.state`. It will also be used for persistence in `localStorage`.
- Public methods like `search`, `filter` and `order` are meant as an entry point to modify the state. Protected methods `applyFilter` (which also does searching) and `applyOrder` apply the state to the `rows` collection. The changes will be reflected to the DOM with `update`.

fixes #1103


evap/static/js/datagrid.js
```diff
@@ -0,0 +1,284 @@
...
+        const searchWords = this.searchWordsOf(this.state.search);
+        for (const row of this.rows) {
+            const isDisplayedBySearch = searchWords.every(searchWord => {
+                return row.search.some(rowWord => rowWord.includes(searchWord));
+            });
+            const isDisplayedByFilters = Object.entries(this.state.filter).every(([name, filterValues]) => {
+                const matchesAtLeastOneValue = filterValues.some(filterValue => {
+                    return row.filter.get(name).some(rowValue => rowValue === filterValue);
+                });
+                return filterValues.length === 0 || matchesAtLeastOneValue;
```
karyon: I'd guess that the JS engines are not that smart and would check `filterValues.length === 0` before computing `matchesAtLeastOneValue`.

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,284 @@
...
+                    button.addClass("active");
+                    this.filterButtons.not(button).removeClass("active");
+                    this.filter({icon: [button.data("filter")]});
+                }
+            });
+        });
+    }
+
+    addFilterRowData(row, data) {
+        // Parsing the icon classes for filtering may not be robust
```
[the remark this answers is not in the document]
pixunil: Well, there are two problems that could arise in the future:
1. The position in the class attribute changes.
    Currently, the classes are `"fas fa-icon-name icon-color"`. Changing the `fas` portion or adding more classes in front of it breaks the `.slice(4)` method.
2. The order in the class attribute changes.
    If the attribute evolves to `"fas icon-color fa-icon-name"` or even more evil to `"icon-color fas fa-icon-name"`, the `===` check later one will fail. (This already existed before this PR)

Alternative would be to make a data attribute independent of the icon classes. But I was wary whether it is worth the effort.

evap/results/templates/results_index.html
```diff
@@ -126,174 +127,53 @@ <h5 class="card-title mb-lg-0">
...
-                selectedSemesters.forEach(function(value){
-                    $(`input[name=semester][value="${value}"]`).prop('checked', true);
+            grid.attachTo($("#results-grid"), $(".grid-header .col-sorting"));
+
+            $("[data-reset=filter]").click(() => {
+                grid.search("");
+                grid.filter({
+                    semester: [],
+                    degree: [],
+                    courseType: [],
```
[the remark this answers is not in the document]
pixunil: I didn't planned to move that much stuff actually into `data-grid`, but suddenly the design drove itself to put event handling of inputs in to the classes.

So, when you think that more code should live in the class, I'm fine with that. When you think more code should be at the client site, there is probably a reason why it was in the .js

evap/static/js/datagrid.js
pixunil: Do you ask for performance considerations?
The reason why it is written like that: `some` will be `false` on an empty collection – but if nothing is selected in a filter category, we want to show everything

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,284 @@
...
+        this.applyOrder();
+        this.update();
+    }
+
+    saveToStorage() {
+        localStorage.setItem(this.storageKey, JSON.stringify(this.state));
+    }
+
+    syncFilterStateWithInputs() {
+        this.searchInput.val(this.state.search);
```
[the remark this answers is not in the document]
pixunil: It gets overwritten in [`EvaluationGrid`](https://github.com/fsr-de/EvaP/blob/3ed29d19f30a94027f185a326c55de2a0f5a6218/evap/static/js/datagrid.js#L218-L223) and [`ResultGrid`](https://github.com/fsr-de/EvaP/blob/3ed29d19f30a94027f185a326c55de2a0f5a6218/evap/static/js/datagrid.js#L271-L283) – there they have more than the search field as filter inputs.

evap/static/js/locale/select2_en.js
```diff
@@ -1,3 +0,0 @@
-/*! Select2 4.0.0 | https://github.com/select2/select2/blob/master/LICENSE.md */
-
-(function(){if(jQuery&&jQuery.fn&&jQuery.fn.select2&&jQuery.fn.select2.amd)var e=jQuery.fn.select2.amd;return e.define("select2/i18n/en",[],function(){return{errorLoading:function(){return"The results could not be loaded."},inputTooLong:function(e){var t=e.input.length-e.maximum,n="Please delete "+t+" character";return t!=1&&(n+="s"),n},inputTooShort:function(e){var t=e.minimum-e.input.length,n="Please enter "+t+" or more characters";return n},loadingMore:function(){return"Loading more results…"},maximumSelected:function(e){var t="You can only select "+e.maximum+" item";return e.maximum!=1&&(t+="s"),t},noResults:function(){return"No results found"},searching:function(){return"Searching…"}}}),{define:e.define,require:e.require}})();
```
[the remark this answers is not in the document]
pixunil: Oopsie! I got the wrong directory there…

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,284 @@
...
+
+    bindEvents() {
+        this.delayTimer = null;
+        this.searchInput.on("change paste input", () => {
+            clearTimeout(this.delayTimer);
+            this.delayTimer = setTimeout(() => {
+                this.search(this.searchInput.val());
+            }, 200);
+        }).keypress(event => {
+            if (event.keyCode === 13) {
```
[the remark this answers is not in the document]
pixunil: As a bonus, I explained the "why" for this line

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,285 @@
...
+        this.searchInput = searchInput;
+    }
+
+    attachTo(container) {
+        this.container = container;
+        this.columns = Array.from(this.headers.keys());
+        this.rows = this.fetchRowData();
+        this.loadFromStorage();
+        this.bindEvents();
+    }
```
pixunil: The situation with `constructor` and `attachTo` is bad.
The problem I have encountered was that there is no `this` before `super`. As you can see, I exclusively set properties in the constructor and do the logic stuff in `attachTo`.

What I would have wanted in subclasses:
```javascript
constructor({filterCheckboxes, ...options}) {
    this.filterCheckboxes = filterCheckboxes;
    super(options); // would also set up row data, bind events, etc. which uses above defined properties
}
```

Adding to the confusion, the subclasses of `DataGrid` must populate `this.headers` before the call to `super.attachTo`. Note that in `attachTo`, as we have an already constructed object, there is `this`-life before `super`.

I didn't come up with alternatives, do you have?

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,284 @@
...
+        }).keypress(event => {
+            if (event.keyCode === 13) {
+                event.target.blur();
+            }
+        });
+
+        for (const [column, header] of this.headers) {
+            header.click(() => {
+                const ordering = header.hasClass("col-sorting-asc") ? "desc" : "asc";
+                this.order([[column, ordering]]);
```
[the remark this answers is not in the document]
pixunil: Because `order` calls `applyOrder`?

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    addFilterRowData(row, data) {}
+
+    findCellByColumn(row, column) {
+        return row.children().eq(column);
+    }
+
+    get defaultSorting() {
+        if (this.columns.length > 0) {
```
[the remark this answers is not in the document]
pixunil: That's in there because grids without sorting (like the user grid) will still be processed like other grids: `applyOrder` will be called. Passing `sorting = []` makes this a method a no-op. But passing a value like `[[undefined, "asc"]]` will throw an error :(

It is more a guard to access `this.columns` within bounds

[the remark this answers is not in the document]
karyon: why would we pass `[[undefined, "asc"]]`?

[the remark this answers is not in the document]
karyon: ....to expand on that, I understand the idea with guarding the access with a bounds check. however, I'd rather make the code throw some error in that case instead of silently eating our programming errors. But that's probably a personal preference?

evap/static/js/datagrid.js
karyon: I think I've understood what the code does and i think it is correct. Yes, I was purely looking at performance. Reordering such boolean expressions is what i did in 433dcf28b and it made a rather large difference.

To avoid misunderstandings, I propose to early-out on `filterValues.length === 0` before executing the `some` call. whether that's by inlining `matchesAtLeastOneValue` or adding an additional `if` i don't care. I also don't care if you show that all of this doesn't make any practical difference :)

evap/results/templates/results_index.html
```diff
@@ -126,174 +127,53 @@ <h5 class="card-title mb-lg-0">
...
-                delayTimer = setTimeout(function() {
-                    updateTable();
-                }, 200);
-            }).keypress(event => {
-                if (event.keyCode == 13) {
-                    event.target.blur();
-                }
+
+            $("[data-reset=ordering]").click(() => {
+                grid.order(grid.defaultSorting);
```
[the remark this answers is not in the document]
karyon: jakob: this is the only table that has a separate "default sort" button, the others can be sorted via column headers

evap/static/js/datagrid.js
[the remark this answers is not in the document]
karyon: jakob: lets add another attribute

evap/static/js/datagrid.js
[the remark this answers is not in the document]
karyon: misunderstanding. `columns` are actually only the columns that are used for sorting, so with a better name this makes 100% sense.

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,284 @@
...
+        for (const header of this.headers.values()) {
+            header.removeClass("col-sorting-asc col-sorting-desc");
+        }
+        for (const [column, ordering] of this.state.sorting) {
+            this.headers.get(column).addClass(`col-sorting-${ordering}`);
+        }
+
+        this.rows.sort((a, b) => {
+            for (const [column, ordering] of this.state.sorting) {
+                if (a.order.get(column) < b.order.get(column)) {
```
karyon: as discussed, lets call a.order orderValues instead and a.filter filterValues

evap/static/js/datagrid.js
karyon: jakob: said he might just as well put everything into an init method.

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,314 @@
+class DataGrid {
+    // Use init instead of constructor to enable access to this before calling super
+    init({storageKey, container, searchInput}) {
+        this.storageKey = storageKey;
+        this.container = container;
+        this.sortableColumns = Array.from(this.sortableHeaders.keys());
```
karyon: Is there a benefit to keeping sortableColumns and sortableHeaders separate? I think I'd prefer to give each column a header, and not having two seemingly-but-in-fact-not independent datastructures.

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,314 @@
...
+
+    fetchRowData() {
+        return this.container.children().get()
+            .map(row => {
+                let data = {
+                    element: row,
+                    orderValues: new Map(),
+                    filterValues: new Map(),
+                };
+                data.searchWords = this.findSearchableCells($(row))
```
karyon: why not put that into the initialization above?

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,314 @@
...
+                    button.addClass("active");
+                    this.filterButtons.not(button).removeClass("active");
+                    this.filter({state: [button.data("filter")]});
+                }
+            });
+        });
+    }
+
+    addFilterRowData(row, data) {
+        const state = row.find(".icon").get()
```
karyon: I thought I had seen you replacing the icon matching by something else, or do i misremember?

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,314 @@
...
+
+    findSearchableCells(row) {
+        return row.find(".evaluation-name, [data-col=responsible]").get();
+    }
+
+    addFilterRowData(row, data) {
+        for (const [name, {selector, checkboxes}] of Object.entries(this.filterCheckboxes)) {
+            const filterValues = $(row).find(selector).get()
+                .map(element => $(element).text().trim())
+                .map(filterName => checkboxes.filter(`[data-filter="${filterName}"]`).val());
```
karyon: what do these three lines do?

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,314 @@
...
+        data.filterValues.set("state", state);
+    }
+
+    get defaultOrder() {
+        return [[5, "asc"]];
+    }
+
+    syncFilterStateWithInputs() {
+        super.syncFilterStateWithInputs();
+        if (this.state.filter.state) {
```
karyon: perhaps rename the second state to evaluationState?

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,314 @@
...
+        this.resetOrder = resetOrder;
+        super.init(options);
+    }
+
+    bindEvents() {
+        super.bindEvents();
+        for (const [name, {checkboxes}] of Object.entries(this.filterCheckboxes)) {
+            checkboxes.on("change", () => {
+                const values = checkboxes.filter(":checked").get().map(elem => elem.value);
+                this.filter({[name]: values});
```
karyon: why is `name` in a list here? that doesn't seem to be the case for the other calls to `filter`

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,314 @@
...
+
+                this.addFilterRowData($(row), data);
+
+                for (const column of this.sortableColumns) {
+                    const cell = this.findCellByColumn($(row), column);
+                    if (cell.is("[data-order]")) {
+                        data.orderValues.set(column, cell.data("order"));
+                    } else {
+                        data.orderValues.set(column, cell.html().trim());
+                    }
```
karyon: [nit] I'd do something like 
```
const orderValue = cell.data("order");
if (orderValue === undefined)
    orderValue = cell.html().trim();
data.orderValues.set(column, orderValues);
```
only one access to the order attribute, and only one call to set.

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,314 @@
...
+    filterRows() {
+        const searchWords = this.searchWordsOf(this.state.search);
+        for (const row of this.rows) {
+            const isDisplayedBySearch = searchWords.every(searchWord => {
+                return row.searchWords.some(rowWord => rowWord.includes(searchWord));
+            });
+            const isDisplayedByFilters = Object.entries(this.state.filter).every(([name, filterValues]) => {
+                if (filterValues.length === 0) {
+                    return true;
+                } else {
```
karyon: because of the return i'd remove the else and de-indent the code below

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,314 @@
...
+            if (column === "name-semester") {
+                this.sort([["name", order], ["semester", order]]);
+            } else {
+                this.sort([[column, order]]);
+            }
+        });
+
+        this.resetFilter.click(() => {
+            this.state.filter = {};
+            this.search("");
```
karyon: :)
I was about to suggest to inline `filter` and `search`, and for those it would work fine, but then `sort` should probably also be inlined and that one is called in three places...

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,314 @@
...
+    update() {
+        this.container.children().detach();
+        const elements = this.rows
+            .filter(row => row.isDisplayed)
+            .map(row => row.element);
+        this.container.append($(elements));
+        this.saveToStorage();
+    }
+
+    loadFromStorage() {
```
karyon: maybe call this restoreStateFromStorage? so it's clear that this is about the state and that it will restore things, and not just load them (at least to me "restore" sounds like it would do more things (like applying the state) than just "load")

(i'd call the other method `saveStateToStorage` then accordingly)

evap/static/scss/components/_grid.scss
```diff
@@ -159,6 +159,20 @@ div.container {
     opacity: 1;
 }
 
+th.col-sorting {
```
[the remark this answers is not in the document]
He3lixxx: I lol'd

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,314 @@
...
+        }).keypress(event => {
+            // after enter, unfocus the search input to collapse the screen keyboard
+            if (event.keyCode === 13) {
+                event.target.blur();
+            }
+        });
+
+        for (const [column, header] of this.sortableHeaders) {
+            header.click(() => {
+                const ordering = header.hasClass("col-sorting-asc") ? "desc" : "asc";
```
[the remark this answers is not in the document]
pixunil: The first click should order the column ascending. All following clicks should toggle the order.
This condenses into this expression. Do you prefer more explicit code or a comment?

evap/static/js/datagrid.js
pixunil: I changed `filter` to be a map, not a object. This removes some quirky spots in the code, like the `Object.assign` and the `[key]` syntax here (the difference between `{key: value}` and `{[key]: value}` is that the key of the latter will be dynamically determined by a variable)

evap/static/js/datagrid.js
pixunil: Primarly to reduce bloat in the object literal. I now set a variable before it, so you see all keys in the object literal

evap/static/js/datagrid.js
pixunil: Actually, the `filter` method couldn't suffice all use-cases of filtering (reset, update one filter, clear one filter) and it is not really logical to have a separate `search` method.

Currently, everything related to filtering / searching is inlined, which may not be quite what we want. As in my empirical profilings the updates to the DOM are outweighing the filter / sort logic, I could imagine to call `filterRows` and `sortRows` in `update`.
That way, only one call would be needed at the expense of doing potentially unneeded work. What do you think?

evap/static/js/datagrid.js
pixunil: Resolved as filters now are never empty. This moves this logic away from `filterRows`. Also, having only "active" filters in the map is imo a simpler mental model.

evap/evaluation/templatetags/evaluation_filters.py
```diff
@@ -38,10 +38,10 @@
...
 StateValues = namedtuple('StateValues', ('order', 'icon', 'filter', 'description'))
 APPROVAL_STATES = {
-    'new': StateValues(0, 'fas fa-circle icon-yellow', 'fa-circle icon-yellow', _('In preparation')),
-    'prepared': StateValues(2, 'far fa-square icon-gray', 'fa-square icon-gray', _('Awaiting editor review')),
-    'editor_approved': StateValues(1, 'far fa-check-square icon-yellow', 'fa-check-square icon-yellow', _('Approved by editor, awaiting manager review')),
-    'approved': StateValues(3, 'far fa-check-square icon-green', 'fa-check-square icon-green', _('Approved by manager')),
+    'new': StateValues(0, 'fas fa-circle icon-yellow', 'new', _('In preparation')),
+    'prepared': StateValues(2, 'far fa-square icon-gray', 'prepared', _('Awaiting editor review')),
+    'editor_approved': StateValues(1, 'far fa-check-square icon-yellow', 'editor_approved', _('Approved by editor, awaiting manager review')),
+    'approved': StateValues(3, 'far fa-check-square icon-green', 'approved', _('Approved by manager')),
```
karyon: I can't find the place where those are used, so: why does it help to have the key as part of the value tuple?

evap/static/js/datagrid.js
```diff
@@ -0,0 +1,323 @@
...
+    saveStateToStorage() {
+        const stored = {
+            search: this.state.search,
+            filter: [...this.state.filter],
+            order: this.state.order,
+        };
+        localStorage.setItem(this.storageKey, JSON.stringify(stored));
+    }
+
+    syncFilterStateWithInputs() {
```
karyon: you're changing the inputs to match the state, right? I'd suggest to reflect that in the name somehow. `sync` goes both ways for me, or if it should be one way, you're syncing the inputs, not the state. at least I would say so :)

evap/evaluation/templatetags/evaluation_filters.py
pixunil: Actually, it is well-hidden behind the `|approval_state_values` template filter. The property `filter` is used here:
https://github.com/fsr-de/EvaP/blob/f713938529166613866996222d657cead2ff3322/evap/staff/templates/staff_semester_view_evaluation.html#L18
One might ask, why not use the state name as-is? Well, we want to clamp all higher states to `approved`. This could be implemented differently, but I just kept the structure like I found it

PR #1370, opened 2019-11-04
delay unreview button activation

### LINKED ISSUE #1364: Delay unreview button activation
The unreview button in the quick view for reviewing text answers should not immediately be activated when selecting a publishing option, because the view immediately moves forward to the next text answer and the button is deactivated again. Instead, it should only be active if the current text answer was already reviewed.

niklasmohrin:
moved frontend button enabling / disabling around to fix "unreview" button flickering when reviewing text answers

fix #1364

### DIFF
--- a/evap/staff/templates/quick-review.js
+++ b/evap/staff/templates/quick-review.js
@@ -136,11 +136,11 @@ $(document).ready(() => {
 
         if(action === "unreview") {
             active.removeAttr("data-review");
+            updateButtons();
         } else {
             active.attr("data-review", action);
+            updateButtonActive();
         }
-
-        updateButtonActive();
 
         if(action !== "unreview") {
             slideTo(index + 1);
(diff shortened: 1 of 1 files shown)


evap/staff/templates/quick-review.js
```diff
@@ -179,6 +185,5 @@ $(document).ready(() => {
             btn.toggleClass(actions[action], review === action);
             btn.toggleClass("btn-outline-secondary", review !== action);
         }
-        slider.find("[data-action=unreview]").attr("disabled", !review);
     }
-});
+});
```
janno42: please add a newline at the end of the file

evap/staff/templates/quick-review.js
```diff
@@ -136,6 +136,7 @@ $(document).ready(() => {
 
         if(action === "unreview") {
             active.removeAttr("data-review");
+            updateButtons();
         } else {
             active.attr("data-review", action);
```
janno42: please move the method call into the else part

janno42:
thanks :)

PR #1376, opened 2019-11-18
Explicitely set `weak=True` on `@receiver`

pixunil:


evap/staff/views.py
```diff
@@ -795,7 +795,7 @@ def evaluation_edit(request, semester_id, evaluation_id):
 
 @manager_required
 def helper_evaluation_edit(request, semester, evaluation):
-    @receiver(RewardPointGranting.granted_by_removal)
+    @receiver(RewardPointGranting.granted_by_removal, weak=True)
```
He3lixxx: Personally, I'd like a short comment explaining what the problem is here (
- This receiver is added once per request for this view -- this looks fishy. It's okay though, as we use weak=True, so whenever this handler ends, no references to this function will be kept, so the receiver will be removed again
- Capturing the request is not a problem. If the receiver is removed and this function ends, the whole inner function will be garbage collected, so there should be no references to the request left and it should also be gced

)

PR #1377, opened 2019-11-18
Remind all responsibles

### LINKED ISSUE #1081: Remind all
On the Todo list a button `Remind all` should be added, that allows to send reminders to all responsibles at once. This button opens a modal with a list of the responsibles, asking for confirmation. Confirming will then send all emails that would be sent by clicking the `Send reminder` buttons for all responsibles individually.

niklasmohrin:
Fixes #1081

### DIFF
--- a/evap/staff/templates/staff_semester_preparation_reminder.html
+++ b/evap/staff/templates/staff_semester_preparation_reminder.html
@@ -9,6 +9,14 @@
 
 {% block content %}
     {{ block.super }}
+    
+    {% if responsible_list %}
+        <div class="d-flex mb-3">
+            <div class="mr-auto">
+                <button type="button" id="remindAllButton" onClick="remindAllModalShow()" class="btn btn-sm btn-light">{% trans 'Remind All' %}</button>
+            </div>
+        </div>
+    {% endif %}
 
     {% for responsible, evaluations, delegates in responsible_list %}
         <div class="card{% if not forloop.last %} mb-3{% endif %}">
@@ -60,4 +68,30 @@
         </div>
     {% endfor %}
 
+{% endblock %}
+
+{% block modals %}
+    {{ block.super }}
+    {% trans 'Remind all' as title %}
+    {% trans 'Do you really want to remind everyone?' as question %}
+    {% trans 'Remind all' as action_text %}
+    {% include 'confirmation_modal.html' with modal_id='remindAllModal' title=title question=question action_text=action_text btn_type='primary' %}
+    <script type="text/javascript">
+        function remindAllModalAction() {
+            var remindAllButton = $('#remindAllButton');
+            remindAllButton.prop('disabled', true);
+            remindAllButton.text('{% trans 'Reminding everyone' %}');
+
+            $.ajax({
+                type: "POST",
+                url: "{% url 'staff:semester_preparation_reminder' semester_id=semester.id %}",
+                success: function(){
+                    location.reload();
+                },
+                error: function(){
+                    window.alert("{% trans 'The server is not responding.' %}");
+                }
+            })
+        }
+    </script>
 {% endblock %}
--- a/evap/staff/tests/test_views.py
+++ b/evap/staff/tests/test_views.py
@@ -1,6 +1,7 @@
 import datetime
 import os
 import glob
+from unittest.mock import patch
 
 from django.conf import settings
 from django.contrib.auth.models import Group
@@ -512,6 +513,7 @@ class TestSemesterAssignView(WebTest):
 class TestSemesterPreparationReminderView(WebTestWith200Check):
     url = '/staff/semester/1/preparation_reminder'
     test_users = ['manager']
+    csrf_checks = False
 
     @classmethod
     def setUpTestData(cls):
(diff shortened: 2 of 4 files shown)


evap/staff/templates/staff_semester_preparation_reminder.html
```diff
@@ -10,6 +10,16 @@
 {% block content %}
     {{ block.super }}
 
+    {% csrf_token %}
```
He3lixxx: Hmm, I think I told you that you need to handle the csrf token here - guess I was wrong. Django has some doku entries about how to do this [here](https://docs.djangoproject.com/en/2.2/ref/csrf/#setting-the-token-on-the-ajax-request).
We already have some places where we do something similar. Our implementation is in `static/js/crsf_utils.js`. From a quick view, it looks like you don't need to do any additional handling, it should work automatically. Can you verify that?

evap/staff/templates/staff_semester_preparation_reminder.html
```diff
@@ -10,6 +10,16 @@
...
+
+    {% if responsible_list %}
+        <div class="card card-submit-area text-center mb-3">
+            <div class="card-body">
+                <button id="remindAllButton" onClick="remindAllModalShow()" class="btn btn-primary">{% trans 'Remind All' %}</button>
+            </div>
+        </div>
```
He3lixxx: Just about the looks: Up to this point, I think we always had buttons stack up from the left without a card (compare e.g. the user list or the course type list). This looks a bit different. @janno42 do we want it this way or should it be similar to the other buttons we have?

evap/staff/templates/staff_semester_preparation_reminder.html
```diff
@@ -61,3 +71,27 @@
     {% endfor %}
 
 {% endblock %}
+
+{% block modals %}
+    {{ block.super }}
+    {% trans 'Remind all' as title %}
+    {% trans 'Do you really want to remind everyone?' as question %}
```
He3lixxx: In the issue (#1081), @janno42 wrote that the list of responsibles that will be reminded should be shown. I like that idea. Putting together the list here is probably a bit annoying with django's template language, so I'd suggest we build up the question text in the view. What do you think?

evap/staff/tests/test_views.py
```diff
@@ -527,6 +529,23 @@ def test_preparation_reminder(self):
         self.assertContains(response, 'user_to_find')
         self.assertContains(response, 'name_to_find')
 
+    @patch("evap.staff.views.EmailTemplate.send_to_user")
+    def test_remind_all(self, send_to_user_mock):
+        user = mommy.make(UserProfile)
+        evaluation = mommy.make(Evaluation, course=mommy.make(Course, semester=self.semester, responsibles=[user]), state='prepared')
+
+        response = self.app.post(self.url, user='manager')
+        self.assertEqual(response.status_int, 200)
```
He3lixxx: So far in our code, we always used `status_code`. Is there a difference between the two? If not, I'd like to keep it unified and use `status_code` here as well.

evap/staff/tests/test_views.py
```diff
@@ -527,6 +529,23 @@ def test_preparation_reminder(self):
...
+
+        recipient = UserProfile.objects.get(pk=user.pk)
```
He3lixxx: You should be able to just use `user` here.

evap/staff/views.py
```diff
@@ -576,8 +576,18 @@ def semester_preparation_reminder(request, semester_id):
     responsible_list = [(responsible, [evaluation for evaluation in evaluations if responsible in evaluation.course.responsibles.all()],
                          responsible.delegates.all()) for responsible in responsibles]
 
-    template_data = dict(semester=semester, responsible_list=responsible_list)
-    return render(request, "staff_semester_preparation_reminder.html", template_data)
+    if request.method == "POST":
+        template = EmailTemplate.objects.get(name=EmailTemplate.EDITOR_REVIEW_REMINDER)
+        subject_params = {}
+        for responsible, evaluations, __ in responsible_list:
+            recipient = UserProfile.objects.get(pk=responsible.pk)
```
He3lixxx: Same here, you should be able to just use `responsible`

evap/staff/views.py
```diff
@@ -576,8 +576,18 @@ def semester_preparation_reminder(request, semester_id):
...
+            body_params = {"user": recipient, "evaluations": evaluations}
+            EmailTemplate.send_to_user(recipient, template, subject_params, body_params, use_cc=True, request=request)
+        messages.success(request, _("Successfully sent reminders to everyone."))
+        return HttpResponse("")
```
He3lixxx: So far, we've always used
```
return HttpResponse()  # 200 OK
```
Again, your code is fine, but I'd like to keep it consistent.

evap/staff/templates/staff_semester_preparation_reminder.html
janno42: I think it's fine if we don't have the list because it would be too much work for very little benefit.

evap/staff/templates/staff_semester_preparation_reminder.html
janno42: yes, please put the button left-aligned on top of the page (like, e.g., on `staff/user/`).

Issue #1374, opened 2019-11-12
Make emails main identification for users


janno42:
After the change to OpenID for the login we rarely require the usernames anymore. Therefore we should make emails the main identification for users.

That means:

- [x] The importers shouldn't expect username columns anymore.
- [x] User selection fields should show the leading part of the email address (everything before the `@`) instead of the username.
- [x] The user profile UI needs to be updated in several places.
- [x] ~User merging must be updated.~ (nothing to do right now)
- [x] User bulk deletion must be updated.
- [x] The development login option with username and password should be changed to email and password (where we could allow invalid email addresses, so that `evap` would stay a valid login name).

After this the username can completely be removed from the `UserProfile`.

felixrindt:
This might lead to a lot of hickups because of the `@hpi.de` and `@hpi.uni-potsdam.de` identity.

janno42:
That's no problem, we're already replacing these on saving. So all addresses have the same domain.

PR #1382, opened 2019-11-26
Remove usernames in importers

janno42:
This is a first step for #1374 (which will be a lot more work, so this is separated as we need it soon). The importers can now handle files without usernames which is the only form we're processing from now on.


evap/staff/tests/test_importers.py
```diff
@@ -60,32 +60,17 @@ def test_created_users(self):
...
+        mommy.make(UserProfile, first_name='Lucilia', last_name="Manilium")
 
         __, __, warnings_test, __ = UserImporter.process(self.valid_excel_content, test_run=True)
         __, __, warnings_no_test, __ = UserImporter.process(self.valid_excel_content, test_run=False)
 
         self.assertEqual(warnings_test, warnings_no_test)
         self.assertIn("An existing user has the same first and last name as a new user:<br />"
-                " - lucilia.manilium2 ( Lucilia Manilium, ) (existing)<br />"
-                " - lucilia.manilium ( Lucilia Manilium, lucilia.manilium@institution.example.com) (new)",
+                " -  Lucilia Manilium,  (existing)<br />"
```
janno42: joko: i'd add an email address so it's clear what happens here.

PR #1394, opened 2019-12-13
Contributor results export

### LINKED ISSUE #1378: Excel export for contributors
There should be a functionality for exporting evaluation results per contributor. Contributors should be able to download an automatically created Excel sheet containing all the evaluations they contributed to or were responsible for.
Similar to the semester export this should include a column for each evaluation (with added information about the semester where it took place) which contains the rating results for general questions and for questions about this contributor.

janno42:
Fix #1378 
This PR adds the functionality to export an Excel sheet with the results for all evaluations a specific user has contributed to or is responsible for. This export can be accessed from each contributor's index page and for managers from the user's UserProfile edit page.

Open todos:
- [x] Write tests

### DIFF
--- a/evap/contributor/templates/contributor_index.html
+++ b/evap/contributor/templates/contributor_index.html
@@ -37,6 +37,9 @@
                 <em>{% trans 'More details:' %} <a href="/faq#3-s">{% trans 'FAQ/Results' %}</a></em>
             </div>
         </div>
+        <div class="mb-auto ml-2 d-print-none">
+            <a href="{% url 'contributor:export' %}" class="btn btn-sm btn-light">{% trans 'Export results' %}</a>
+        </div>
         {% if user.is_delegate %}
             <div class="btn-switch btn-switch-light mb-auto ml-2 d-print-none">
                 <div class="btn-switch-label"><span class="fas fa-user-circle"></span> {% trans 'Delegated evaluations' %}</div>
--- a/evap/results/exporters.py
+++ b/evap/results/exporters.py
@@ -131,7 +131,12 @@ class ExcelExporter(object):
             ), "headline")
 
             for evaluation, results in evaluations_with_results:
-                writec(self, evaluation.full_name, "evaluation")
+                title = evaluation.full_name
+                if len(semesters) > 1:
+                    title += "\n{}".format(evaluation.course.semester.name)
+                responsible_names = [responsible.full_name for responsible in evaluation.course.responsibles.all()]
+                title += "\n{}".format(", ".join(responsible_names))
+                writec(self, title, "evaluation")
 
             writen(self, _("Degrees"), "bold")
             for evaluation, results in evaluations_with_results:
--- a/evap/results/exporters.py
+++ b/evap/results/exporters.py
@@ -1,5 +1,6 @@
 from collections import OrderedDict
 
+from django.db.models import Q
 from django.utils.translation import ugettext as _
 
 import xlwt
@@ -75,7 +76,7 @@ class ExcelExporter(object):
 
         return filtered_questions
 
-    def export(self, response, semesters, selection_list, include_not_enough_voters=False, include_unpublished=False):
+    def export(self, response, semesters, selection_list, include_not_enough_voters=False, include_unpublished=False, contributor=None):
         self.workbook = xlwt.Workbook()
         self.init_styles(self.workbook)
         counter = 1
@@ -93,9 +94,11 @@ class ExcelExporter(object):
                 evaluation_states.extend(['evaluated', 'reviewed'])
 
             used_questionnaires = set()
-            for evaluation in Evaluation.objects.filter(
-                course__semester__in=semesters, state__in=evaluation_states, course__degrees__in=degrees, course__type__in=course_types
-            ).distinct():
+            evaluations_filter = Q(course__semester__in=semesters, state__in=evaluation_states, course__degrees__in=degrees, course__type__in=course_types)
+            if contributor:
+                evaluations_filter = evaluations_filter & (Q(course__responsibles__in=[contributor]) | Q(contributions__contributor__in=[contributor]))
+            evaluations = Evaluation.objects.filter(evaluations_filter).distinct()
+            for evaluation in evaluations:
                 if evaluation.is_single_result:
                     continue
                 if not evaluation.can_publish_rating_results and not include_not_enough_voters:
@@ -117,7 +120,9 @@ class ExcelExporter(object):
             used_questionnaires = sorted(used_questionnaires)
 
             export_name = "Evaluation"
-            if len(semesters) == 1:
+            if contributor:
+                export_name += "\n{}".format(contributor.full_name)
+            elif len(semesters) == 1:
                 export_name += "\n{}".format(semesters[0].name)
             degree_names = [degree.name for degree in Degree.objects.filter(pk__in=degrees)]
             course_type_names = [course_type.name for course_type in CourseType.objects.filter(pk__in=course_types)]
--- a/evap/results/exporters.py
+++ b/evap/results/exporters.py
@@ -4,7 +4,7 @@ from django.utils.translation import ugettext as _
 
 import xlwt
 
-from evap.evaluation.models import CourseType, Degree
+from evap.evaluation.models import CourseType, Degree, Evaluation
 from evap.results.tools import (collect_results, calculate_average_course_distribution, calculate_average_distribution,
                                 distribution_to_grade, get_grade_color)
 
@@ -15,8 +15,7 @@ class ExcelExporter(object):
     NUM_GRADE_COLORS = 21  # 1.0 to 5.0 in 0.2 steps
     STEP = 0.2  # we only have a limited number of custom colors
 
-    def __init__(self, semester):
-        self.semester = semester
+    def __init__(self):
         self.styles = dict()
 
     def normalize_number(self, number):
@@ -76,7 +75,7 @@ class ExcelExporter(object):
 
         return filtered_questions
 
-    def export(self, response, selection_list, include_not_enough_voters=False, include_unpublished=False):
+    def export(self, response, semesters, selection_list, include_not_enough_voters=False, include_unpublished=False):
         self.workbook = xlwt.Workbook()
         self.init_styles(self.workbook)
         counter = 1
@@ -94,8 +93,8 @@ class ExcelExporter(object):
                 evaluation_states.extend(['evaluated', 'reviewed'])
 
             used_questionnaires = set()
-            for evaluation in self.semester.evaluations.filter(
-                state__in=evaluation_states, course__degrees__in=degrees, course__type__in=course_types
+            for evaluation in Evaluation.objects.filter(
+                course__semester__in=semesters, state__in=evaluation_states, course__degrees__in=degrees, course__type__in=course_types
             ).distinct():
                 if evaluation.is_single_result:
                     continue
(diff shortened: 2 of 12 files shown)


evap/evaluation/tools.py
```diff
@@ -122,3 +123,21 @@ def clean_email(email):
...
+    from evap.results.exporters import ExcelExporter
+
+    filename = "Evaluation_{}.xls".format(contributor.full_name)
+    response = HttpResponse(content_type="application/vnd.ms-excel")
+    response["Content-Disposition"] = "attachment; filename=\"{}\"".format(filename)
+    ExcelExporter().export(
+        response,
+        Semester.objects.all(),
+        [(Degree.objects.all(), CourseType.objects.all())],
+        include_not_enough_voters=True,
```
karyon: that's surprising, can contributors usually access those?

evap/results/exporters.py
```diff
@@ -105,26 +112,37 @@ def export(self, response, selection_list, include_not_enough_voters=False, incl
                 for questionnaire_result in collect_results(evaluation).questionnaire_results:
                     if all(not question_result.question.is_rating_question or question_result.counts is None for question_result in questionnaire_result.question_results):
                         continue
-                    results.setdefault(questionnaire_result.questionnaire.id, []).extend(questionnaire_result.question_results)
-                    used_questionnaires.add(questionnaire_result.questionnaire)
+                    if not contributor or (questionnaire_result.contributor is None or questionnaire_result.contributor == contributor):
```
karyon: parentheses are unneeded

evap/staff/views.py
```diff
@@ -1642,3 +1643,11 @@ def development_components(request):
         'theme_colors': theme_colors
     }
     return render(request, "staff_development_components.html", template_data)
+
+
+@manager_required
+def export_contributor_results_view(request, contributor_id):
+    print(contributor_id)
+    contributor = get_object_or_404(UserProfile, id=contributor_id)
+    print(contributor)
```
karyon: leftover prints

evap/staff/views.py
```diff
@@ -1491,8 +1491,18 @@ def notify_reward_points(grantings, **_kwargs):
...
+    ).distinct()
+    evaluations_contributing_to = [
+        (
+            semester,
+            Evaluation.objects.filter(
+                Q(course__semester=semester) & (Q(contributions__contributor=user) | Q(course__responsibles__in=[user]))
+            ).distinct()
+        )
+        for semester in semesters_with_evaluations
+    ]
```
karyon: I would suggest to only do the evaluations query, and then group by semesters manually.

Also, this does not match the commit message.

evap/evaluation/tools.py
janno42: when collecting results we always check for `can_publish_rating_results`. if it is false, the results will not be returned. so if an evaluation does not have enough voters, it will here be included into the export, but with an empty column.
i would expect this behavior as a contributor in contrast to not finding the evaluation at all.

evap/contributor/templates/contributor_index.html
```diff
@@ -37,6 +37,9 @@
                 <em>{% trans 'More details:' %} <a href="/faq#3-s">{% trans 'FAQ/Results' %}</a></em>
             </div>
         </div>
+        <div class="mb-auto ml-2 d-print-none">
+            <a href="{% url 'contributor:export' %}" class="btn btn-sm btn-light">{% trans 'Export results' %}</a>
+        </div>
```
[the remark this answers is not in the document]
janno42: i don't like it as well but it doesn't belong here :)

evap/staff/templates/staff_user_form.html
```diff
@@ -43,25 +43,29 @@ <h5 class="card-title">{% trans 'CC-User for' %}</h5>
...
-                                                        {{ evaluation.full_name }}
-                                                    </a>
-                                                {% else %}
+                            {% regroup evaluations_contributing_to by course.semester as evaluation_list %}
+                            {% for semester_evaluations in evaluation_list %}
+                                <li>{{ semester_evaluations.grouper.name }}</li>
+                                <ul>
+                                    {% for evaluation in semester_evaluations.list %}
+                                        <li>
+                                            {% if evaluation|can_results_page_be_seen_by:form.instance %}
```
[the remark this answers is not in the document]
janno42: we only want to have the export for a contributor if this contributor can actually see the results (so only after publishing etc.)

evap/results/tests/test_exporters.py
```diff
@@ -131,23 +134,23 @@ def test_view_excel_file_sorted(self):
...
-        self.assertEqual(workbook.sheets()[0].row_values(0)[1], "A – Evaluation1")
-        self.assertEqual(workbook.sheets()[0].row_values(0)[2], "B – Evaluation2")
+        self.assertEqual(workbook.sheets()[0].row_values(0)[1], "A – Evaluation1\n")
+        self.assertEqual(workbook.sheets()[0].row_values(0)[2], "B – Evaluation2\n")
 
         workbook = xlrd.open_workbook(file_contents=content_en.read())
-        self.assertEqual(workbook.sheets()[0].row_values(0)[1], "A – Evaluation2")
-        self.assertEqual(workbook.sheets()[0].row_values(0)[2], "B – Evaluation1")
+        self.assertEqual(workbook.sheets()[0].row_values(0)[1], "A – Evaluation2\n")
+        self.assertEqual(workbook.sheets()[0].row_values(0)[2], "B – Evaluation1\n")
```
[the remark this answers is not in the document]
janno42: `results/exporters.py:144`. the responsibles' names are added to a second line and there are none here.

evap/results/exporters.py
```diff
@@ -105,26 +112,37 @@ def export(self, response, selection_list, include_not_enough_voters=False, incl
                 for questionnaire_result in collect_results(evaluation).questionnaire_results:
                     if all(not question_result.question.is_rating_question or question_result.counts is None for question_result in questionnaire_result.question_results):
                         continue
-                    results.setdefault(questionnaire_result.questionnaire.id, []).extend(questionnaire_result.question_results)
-                    used_questionnaires.add(questionnaire_result.questionnaire)
```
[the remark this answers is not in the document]
janno42: if we export for a contributor then we want to include only general and personal results. so we add results if `not contributor` (a "normal" semester export), `questionnaire_result.contributor is None` (result for a general question) or `questionnaire_result.contributor == contributor` (the result is for the current contributor).
otherwise we skip this result completely.

karyon:
I don't think we ever considered type annotations. I'd be interested in those as well. Although I'd guess they don't help that much when trying to understand the intent behind the code, but more the correctness of the code. We probably need to try harder to convey the intent through the code or (gah!) comments :)

evap/results/templates/results_evaluation_detail.html
```diff
@@ -125,7 +125,7 @@ <h3>{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</h3>
         </div>
     {% endif %}
 
-    {% if contributor_contribution_results %}
+    {% if contributor_contribution_results or other_contributors %}
```
karyon: the condition below says `view == 'export' and other_contributors `, shouldn't both parts be in here? (An empty card body might be rendered otherwise?)

He3lixxx: I agree that the condition here and below should be identical, I'd say we should drop the view == 'export' below though. In the view, the `other_contributors` is filled only when view == 'export'

He3lixxx: Speaking of `other_contributors`: What do you think about renaming the variable? Re-reading the code, it took me quite a while to understand how "other contributors" are different from "non other contributors"^^

I'd maybe `suggest contributors_without_displayed_results` or something similar

evap/staff/templates/staff_user_form.html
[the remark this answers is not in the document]
He3lixxx: So this is not a security feature, but rather a friendly way of keeping staff users of accidentally "leaking" this information? I'm fine with it then.

evap/results/exporters.py
```diff
@@ -15,8 +16,7 @@ class ExcelExporter(object):
     NUM_GRADE_COLORS = 21  # 1.0 to 5.0 in 0.2 steps
     STEP = 0.2  # we only have a limited number of custom colors
 
-    def __init__(self, semester):
-        self.semester = semester
+    def __init__(self):
```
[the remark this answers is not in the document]
janno42: see #1404

evap/results/templates/results_evaluation_detail.html
janno42: i removed the `view == 'export'` below.
naming the variable is hard, though. it only makes sense when looking at it from the export view perspective which is always for a dedicated contributor and all other contributors are then "other contributors". "contributors_without_displayed_results" is also not optimal, because there might be a number of contributors without displayed results (because they might not have any or no results are published at all) independent from the export view stuff.

evap/results/exporters.py
```diff
@@ -112,8 +112,9 @@ def export(self, response, semesters, selection_list, include_not_enough_voters=
                 for questionnaire_result in collect_results(evaluation).questionnaire_results:
                     if all(not question_result.question.is_rating_question or question_result.counts is None for question_result in questionnaire_result.question_results):
                         continue
-                    results.setdefault(questionnaire_result.questionnaire.id, []).extend(questionnaire_result.question_results)
-                    used_questionnaires.add(questionnaire_result.questionnaire)
+                    if not contributor or questionnaire_result.contributor is None or questionnaire_result.contributor == contributor:
```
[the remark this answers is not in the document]
janno42: sorry, i think i was blind there. when i wrote this i thought it should be made differently and it seems that i forgot to do that.

evap/results/templates/results_evaluation_detail.html
He3lixxx: I see your point. How about `contributors_with_omitted_results` or variations with `collapsed`, `left_out`, or `skipped`?

If these also have issues, I'm also fine with leaving it as is.

evap/results/exporters.py
```diff
@@ -94,37 +94,51 @@ def export(self, response, selection_list, include_not_enough_voters=False, incl
...
-                        continue
-                    results.setdefault(questionnaire_result.questionnaire.id, []).extend(questionnaire_result.question_results)
-                    used_questionnaires.add(questionnaire_result.questionnaire)
+                for contribution_result in collect_results(evaluation).contribution_results:
+                    for questionnaire_result in contribution_result.questionnaire_results:
+                        if all(not question_result.question.is_rating_question or question_result.counts is None for question_result in questionnaire_result.question_results):
+                            continue
+                        if not contributor or contribution_result.contributor is None or contribution_result.contributor == contributor:
+                            results.setdefault(questionnaire_result.questionnaire.id, []).extend(questionnaire_result.question_results)
+                            used_questionnaires.add(questionnaire_result.questionnaire)
```
He3lixxx: You could inverse the control flow here (just as is done above), it would allow shortening the condition:
```python
if contributor and contribution_result.contributor != contributor:
    continue

results.setdefault(...)
used_questionnaired.add(...)
```

evap/results/templates/results_evaluation_detail.html
janno42: `contributors_with_omitted_results` is fine

evap/results/exporters.py
janno42: we still need the check for `contribution_result.contributor is None` so that general contributions are not skipped.
the negation would be `contributor and contribution_result.contributor is not None and contribution_result.contributor != contributor`

He3lixxx: Ah true, sorry.

PR #1395, opened 2019-12-16
External user login confirm

### LINKED ISSUE #1393: Add confirmation for login with login key
External users can login by clicking on the login URL sent to them via email. Before users are actually authenticated on the platform after clicking on this link, they should have to confirm the login by, e.g., clicking a confirmation button in a modal or on a designated page.

This prevents cases in which the URL is already requested for, e.g., a preview of the page and thus invalidates the login key which results in a new link being sent when users actively click on the link (which is then already the second GET request).

niklasmohrin:
Fixes #1393

### DIFF
--- a/evap/evaluation/templates/external_user_confirm_login.html
+++ b/evap/evaluation/templates/external_user_confirm_login.html
@@ -1,0 +1,28 @@
+{% extends 'base.html' %}
+
+{% block content %}
+{{ block.super }}
+
+<div class="jumbotron jumbotron-evap">
+    <h1 class="display-4">{% trans 'Welcome to the evaluation platform!' %}</h1>
+    <hr class="my-4" />
+    <div class="card-deck-external-login-confirm">
+        <div class="card border-primary mx-auto">
+            <div class="card-body">
+                <h4 class="card-title text-primary">
+                    {% trans 'Confirm your login' %}
+                </h4>
+                <p class="card-text">
+                    {% trans 'By clicking this button, you will be logged in and your login link will be invalidated.' %}
+                </p>
+                <form class="d-flex flex-column" role="form" method="POST">
+                    {% csrf_token %}
+                    <div class="d-flex justify-content-center">
+                        <button type="submit" class="btn btn-primary login-button">{% blocktrans %}Login as {{ username }}{% endblocktrans %}</button>
+                    </div>
+                </form>
+            </div>
+        </div>
+    </div>
+</div>
+{% endblock %}
--- a/evap/evaluation/tests/test_auth.py
+++ b/evap/evaluation/tests/test_auth.py
@@ -33,22 +33,36 @@ class LoginTests(WebTest):
         self.assertRedirects(self.app.get(reverse("contributor:index")), "/?next=/contributor/")
 
         url_with_key = reverse("evaluation:login_key_authentication", args=[self.external_user.login_key])
-        page = self.app.get(url_with_key).follow().follow()
+        old_login_key = self.external_user.login_key
+        old_login_key_valid_until = self.external_user.login_key_valid_until
+        page = self.app.get(url_with_key)
+        self.external_user.refresh_from_db()
+        self.assertEqual(old_login_key, self.external_user.login_key)
+        self.assertEqual(old_login_key_valid_until, self.external_user.login_key_valid_until)
+        self.assertContains(page, 'Login')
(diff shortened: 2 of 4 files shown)


evap/evaluation/templates/external_user_confirm_login.html
```diff
@@ -0,0 +1,28 @@
...
+                <h4 class="card-title text-primary">
+                    {% trans 'Confirm your login' %}
+                </h4>
+                <p class="card-text">
+                    {% trans 'By clicking this button, you will be logged in and your login link will be invalidated.' %}
+                </p>
+                <form class="d-flex flex-column" role="form" method="POST">
+                    {% csrf_token %}
+                    <div class="d-flex justify-content-center">
+                        <button type="submit" class="btn btn-primary login-button">{% trans 'Login as' %} {{ username }}</button>
```
janno42: can you please make the username part of the translation? we probably want to translate it as `Als {{ username }} anmelden` in german.

evap/evaluation/tests/test_auth.py
```diff
@@ -30,25 +30,29 @@ def test_login_url_generation(self):
         self.assertEqual(reversed_url, '/key/{}'.format(self.external_user.login_key))
 
     def test_login_url_works(self):
-        self.assertRedirects(self.app.get(reverse("contributor:index")), "/?next=/contributor/")
```
He3lixxx: Why did you remove this?

evap/evaluation/tests/test_auth.py
```diff
@@ -30,25 +30,29 @@ def test_login_url_generation(self):
...
+        self.assertContains(page, 'Login')
+        self.assertContains(page, self.external_user.full_name)
+
+        page = self.app.post(url_with_key).follow().follow()
         self.assertContains(page, 'Logout')
+        self.assertContains(page, self.external_user.full_name)
 
     def test_login_key_valid_only_once(self):
-        page = self.app.get(reverse("evaluation:login_key_authentication", args=[self.external_user.login_key])).follow().follow()
-        self.assertContains(page, self.external_user.full_name)
```
He3lixxx: I'd keep this here (looks like you moved it to the other test? I guess both tests should have it)

evap/evaluation/tests/test_auth.py
```diff
@@ -30,25 +30,29 @@ def test_login_url_generation(self):
...
     def test_login_url_works(self):
-        self.assertRedirects(self.app.get(reverse("contributor:index")), "/?next=/contributor/")
-
         url_with_key = reverse("evaluation:login_key_authentication", args=[self.external_user.login_key])
-        page = self.app.get(url_with_key).follow().follow()
+        page = self.app.get(url_with_key)
+        self.assertContains(page, 'Login')
+        self.assertContains(page, self.external_user.full_name)
+
+        page = self.app.post(url_with_key).follow().follow()
```
He3lixxx: While this tests the typical login flow, it does not test whether the special case we are doing this for works (email client issues get, browser issues get, browser issues post).
I'd add
```
old_key = self.external_user.login_key
old_valid_until = self.external_user.login_key_valid_until
```
before the get and
```
self.external_user.refresh_from_db()
self.assertEqual(old_login_key, self.external_user.login_key)
self.assertEqual(old_login_key_valid_until, self.external_user.login_key_valid_until)
```
afterwards.

Alternatively, we could add another `self.app.get` between the `get` and the `post` calls, however, I think this might work just because the app remembers something about the user. Not sure about this, though (how much state does `self.app` have in the tests?)

evap/evaluation/templates/external_user_confirm_login.html
```diff
@@ -0,0 +1,28 @@
...
+                <h4 class="card-title text-primary">
+                    {% trans 'Confirm your login' %}
+                </h4>
+                <p class="card-text">
+                    {% trans 'By clicking this button, you will be logged in and your login link will be invalidated.' %}
+                </p>
+                <form class="d-flex flex-column" role="form" method="POST">
+                    {% csrf_token %}
+                    <div class="d-flex justify-content-center">
+                        <button type="submit" class="btn btn-primary login-button">{% blocktrans %}Login as {{ username }} {% endblocktrans %}</button>
```
janno42: there is a trailing space in the string which shouldn't be there ;)

janno42:
thanks!

PR #1403, opened 2020-01-06
Fixes duplicate course name detection when importing semester

### LINKED ISSUE #1390: Error message if course already exists on import
When importing enrollment data including at least one course with a name that already exists for this semester in the database, the importer fails silently. This shouldn't be the case.
The test import should display an error message if a course with the same (German or English) name already exists in the current semester.

There should also be a test for that.

niklasmohrin:
Fixes #1390 

Duplicates were checked in the evaluations, but since these do not have names, a duplicate could never be detected.
The check is now performed on the courses instead.

@janno42 can you confirm?

### DIFF
--- a/evap/staff/importers.py
+++ b/evap/staff/importers.py
@@ -278,9 +278,10 @@ class EnrollmentImporter(ExcelImporter):
 
     def check_evaluation_data_correctness(self, semester):
         for evaluation_data in self.evaluations.values():
-            already_exists = Evaluation.objects.filter(course__semester=semester, name_de=evaluation_data.name_de).exists()
-            if already_exists:
+            if Course.objects.filter(semester=semester, name_en=evaluation_data.name_en).exists():
                 self.errors.append(_("Course {} does already exist in this semester.").format(evaluation_data.name_en))
+            if Course.objects.filter(semester=semester, name_de=evaluation_data.name_de).exists():
+                self.errors.append(_("Course {} does already exist in this semester.").format(evaluation_data.name_de))
 
         degree_names = set()
         for evaluation_data in self.evaluations.values():
--- a/evap/staff/tests/test_importers.py
+++ b/evap/staff/tests/test_importers.py
@@ -220,6 +220,19 @@ class TestEnrollmentImporter(TestCase):
         self.assertIn('Errors occurred while parsing the input data. No data was imported.', errors_test)
         self.assertEqual(UserProfile.objects.count(), original_user_count)
 
+    def test_duplicate_course_error(self):
+        with open(self.filename_valid, "rb") as excel_file:
+            excel_content = excel_file.read()
+
+        semester = baker.make(Semester)
+        baker.make(Course, name_de="Stehlen", name_en="Stehlen", semester=semester)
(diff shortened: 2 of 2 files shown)


evap/staff/importers.py
```diff
@@ -278,7 +280,7 @@ def consolidate_enrollment_data(self):
 
     def check_evaluation_data_correctness(self, semester):
         for evaluation_data in self.evaluations.values():
-            already_exists = Evaluation.objects.filter(course__semester=semester, name_de=evaluation_data.name_de).exists()
+            already_exists = Course.objects.filter(Q(semester=semester) & (Q(name_en=evaluation_data.name_en) | Q(name_de=evaluation_data.name_de))).exists()
             if already_exists:
                 self.errors.append(_("Course {} does already exist in this semester.").format(evaluation_data.name_en))
```
He3lixxx: I suggest doing the following so that a user does not get confused because we're printing back the name in a language that is not actually the problem:
```python
if Course.objects.filter(semester=semester, name_en=evaluation_data.name_en).exists():
    self.errors.append(_("Course {} does already exist in this semester.").format(evaluation_data.name_en))
if Course.objects.filter(semester=semester, name_de=evaluation_data.name_de).exists():
    self.errors.append(_("Course {} does already exist in this semester.").format(evaluation_data.name_de))
```

I think one more line of code for this is reasonable.

evap/staff/importers.py
```diff
@@ -1,15 +1,17 @@
...
 from django.db import transaction
+from django.db.models import Q
 from django.utils.translation import ugettext_lazy, ugettext as _
 from django.utils.safestring import mark_safe
 from django.core.exceptions import ValidationError
 
 from evap.evaluation.models import Contribution, Course, CourseType, Degree, Evaluation, UserProfile
 from evap.evaluation.tools import clean_email
 
+import xlrd
```
janno42: this should actually stay on top (see https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#imports)

He3lixxx: Ah, my fault then - codacy complained about it (just as it does with other files, check this out: https://app.codacy.com/manual/evap/EvaP/file/41013863394/issues/source?bid=7257767&fileBranchId=7257767)
Can we fix this? From the message, it seems to expect the same order as django does, but it probably can't detect xlrd as being third-party. Codacy does not seem to have an option to install the requirements before running pylint. If we can't fix it, I'd suggest turning off the import order linting rule. I observe myself ignoring codacy more and more due to false negatives.

janno42: i'm fine with turning it off. we could instead use the same approach as 1327 to enforce correct linting (running https://github.com/fsr-de/1327/blob/master/_1327/main/management/commands/lint.py in https://github.com/fsr-de/1327/blob/master/.travis.yml)

He3lixxx: Added a comment on #1092

Problem is that the rule itself as it is configured in the .pylintrc is probably a good thing (running a linter locally should tell you this), it's just that codacy doesn't install the requirements thus pylint can not detect xlrd as a third party module.

He3lixxx:
Thanks for contributing! :)

PR #1413, opened 2020-01-24
Various code quality improvements / linting fixes

He3lixxx:
Can we get codacy to 0 issues? Let's find out!

Opening pull request already so it's easier to filter out codacy issues that are already solved.

### DIFF
--- a/evap/evaluation/tests/test_commands.py
+++ b/evap/evaluation/tests/test_commands.py
@@ -108,7 +108,7 @@ class TestAnonymizeCommand(TestCase):
         for question in chain(self.contributor_questions, self.general_questions):
             choices = [choice for choice in CHOICES[question.type].values if choice != NO_ANSWER]
             for answer in choices:
-                count = random.randint(10, 100)
+                count = random.randint(10, 100)  # nosec
                 baker.make(RatingAnswerCounter, question=question, contribution=self.contribution, count=count, answer=answer)
                 answers_per_question[question] += count
 
@@ -128,7 +128,7 @@ class TestAnonymizeCommand(TestCase):
         choices = [choice for choice in CHOICES[question.type].values if choice != NO_ANSWER]
         random.seed(0)
         for answer in choices:
-            count = random.randint(50, 100)
+            count = random.randint(50, 100)  # nosec
             baker.make(RatingAnswerCounter, question=question, contribution=single_result.general_contribution, count=count, answer=answer)
             answer_count_before += count
 
--- a/evap/evaluation/tests/test_tools.py
+++ b/evap/evaluation/tests/test_tools.py
@@ -40,8 +40,12 @@ class TestLanguageSignalReceiver(WebTest):
         self.assertEqual(user.language, 'de')
 
 
+class SaboteurException(Exception):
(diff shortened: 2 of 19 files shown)


He3lixxx:
Meh, codacy does not give you a view of issues that are still present on a specific branch / pr - so I can't really check, but I think this will solve all issues except for the prospector MC0001 (Complexity too high) ones. I couldn't come up with a good solution to them.

evap/evaluation/tests/test_tools.py
```diff
@@ -50,10 +50,13 @@ def test_log_exceptions_decorator(self, mock_logger, __):
             One could create a mock management command and call its handle method manually,
             but to me it seemed safer to use a real one.
         """
+        exception_was_reraised = False
         try:
             management.call_command('update_evaluation_states')
         except Exception:  # pylint: disable=broad-except
-            pass
+            exception_was_reraised = True
```
pixunil: As I understand this test, you transform `Evaluation.update_evaluations()` into a saboteur which raises a generic `Exception`.
What do you think about using a specific `class SaboteurException(Exception)` subclass for the saboteur and using `assertRaises`?

evap/staff/tools.py
```diff
@@ -68,6 +68,13 @@ def delete_navbar_cache_for_users(users):
         cache.delete(key)
 
 
+def create_user_list_html_string_for_message(users):
+    msg = format_html("")
+    for user in users:
+        msg += format_html("<br />{} {} ({})", user.first_name, user.last_name, user.email)
+    return msg
```
pixunil: There exists a [`format_html_join`](https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.html.format_html_join) which could make this more compact

evap/evaluation/tests/test_commands.py
```diff
@@ -128,7 +128,7 @@ def test_single_result_anonymization(self):
         choices = [choice for choice in CHOICES[question.type].values if choice != NO_ANSWER]
         random.seed(0)
         for answer in choices:
-            count = random.randint(50, 100)
```
karyon: can we perhaps exclude anonymize.py and test_*.py from the security checks for less `#nosec` (but i'd do that only if it doesn't require another config file)

evap/evaluation/tests/test_tools.py
He3lixxx: fancy idea, I like it :+1:

evap/staff/tools.py
He3lixxx: Cool, didn't see that. I think this is the only occurence where it really makes sense to prefer this over `format_html` -- do you agree?

evap/evaluation/tests/test_commands.py
He3lixxx: Sadly, the really really great tool codacy does not provide an option to exclude folders / modules / files from their web frontend, so we would need another file.

I guess this resolves this discussion? Please reopen otherwise.

evap/staff/tools.py
pixunil: Yes, I think that’ll be the only occurrence

PR #1415, opened 2020-01-27
Replace user bulk deletion with user bulk update

janno42:
contributes to #1374


evap/staff/tools.py
```diff
@@ -68,31 +69,95 @@ def delete_navbar_cache_for_users(users):
...
+    for imported_email in imported_emails:
+        if is_external_email(imported_email):
+            skipped_external_emails_counter += 1
+            continue
+        matching_user = find_matching_internal_user_for_email(request, imported_email)
+        if not matching_user:
+            emails_of_users_to_be_created.append(imported_email)
+        elif matching_user.email != imported_email:
+            users_to_be_updated.append((matching_user, imported_email))
+            emails_of_non_obsolete_users.append(matching_user.email)  # add old address so this user won't be deleted
```
pixunil: Could you move the `emails_of_non_obsolete_users` variable just before `obsolete_users`?
It could be expressed with something similar to:
```python
emails_of_non_obsolete_users = set(imported_emails) | {user.email for user, _imported_email in users_to_be_updated}
```

evap/staff/tools.py
```diff
@@ -68,31 +69,95 @@ def delete_navbar_cache_for_users(users):
...
+        )
+    if deletable_users:
+        messages.info(request, mark_safe(_('Users to be deleted are:<br />{}')
+            .format('<br />'.join([user.email or user.username for user in deletable_users]))))
+    if users_to_mark_inactive:
+        messages.info(request, mark_safe(_('Users to be marked inactive are:<br />{}')
+            .format('<br />'.join([user.email or user.username for user in users_to_mark_inactive]))))
+    if emails_of_users_to_be_created:
+        messages.info(request, mark_safe(_('Users to be created are:<br />{}')
+            .format('<br />'.join(emails_of_users_to_be_created))))
```
pixunil: Those `mark_safe` should be replaced by [`format_html_join`](https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.html.format_html_join). If it works, haven’t tested it yet :D

evap/staff/tools.py
```diff
@@ -68,31 +69,95 @@ def delete_navbar_cache_for_users(users):
...
             user.is_active = False
             user.save()
-
-        messages.info(request, _('{} users have been deleted').format(len(deletable_users)))
-        messages.info(request, _('{} users have been marked inactive').format(len(users_to_mark_inactive)))
+        for user, email in users_to_be_updated:
+            user.email = email
+            user.save()
+        for email in emails_of_users_to_be_created:
+            UserProfile.objects.create(email=email, username=email)
```
pixunil: You wanted to use a transaction here ^^

evap/staff/tools.py
```diff
@@ -68,31 +69,95 @@ def delete_navbar_cache_for_users(users):
...
-
-    messages.info(request, _('The uploaded text file contains {} usernames. {} other users have been found in the database. '
-                           'Of those, {} will be deleted and {} will be marked inactive.')
-                  .format(len(usernames), len(users), len(deletable_users), len(users_to_mark_inactive)))
-    messages.info(request, mark_safe(_('Users to be deleted are:<br />{}')
-                  .format('<br />'.join([u.username for u in deletable_users]))))
-    messages.info(request, mark_safe(_('Users to be marked inactive are:<br />{}')
-                  .format('<br />'.join([u.username for u in users_to_mark_inactive]))))
+def find_matching_internal_user_for_email(request, email):
+    # for internal users only the part before the @ must be the same to match a user to an email
```
karyon: is that "fuzzy match" something the importers should also do?

karyon:
perhaps check whether all of the input is valid email addresses?

karyon:
also, consider putting the test-run-file on the server and using that on the actual import instead of letting the user upload another file without dry-run, to prevent user errors.

evap/staff/tools.py
```diff
@@ -68,31 +69,95 @@ def delete_navbar_cache_for_users(users):
...
+            _('For {} multiple matching existing users have been found. {} will be updated with the new data.')
+            .format(email, matching_users[0].email)
+        )
+    return matching_users[0]
+
+
+def bulk_update_users(request, user_file, test_run):
+    # pylint: disable=too-many-branches
+    # user_file must have one user per line in the format "{username},{email}"
+    imported_emails = {clean_email(line.decode().split(',')[1]) for line in user_file.readlines()}
```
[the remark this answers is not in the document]
janno42: yes, the username won't be used anymore

evap/staff/tools.py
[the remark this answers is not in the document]
janno42: in most cases there should be an import of some kind which will add the complete information before they receive an email (this creation here is basically just important for editors who can then add new users as contributors who haven't logged in before). but even if not it should be okay, managers should see that during the confirmation. and if that also passes unseen then it's no big deal as well.

evap/staff/tools.py
He3lixxx: I'd put it in the importers too.

evap/staff/tools.py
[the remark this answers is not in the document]
He3lixxx: Do we have some reason to keep the username then?

evap/staff/tools.py
He3lixxx: Or use [bulk_create](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#bulk-create)?

evap/staff/tools.py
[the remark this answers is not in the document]
janno42: this is the file format we have to work with on production.

evap/staff/tests/test_views.py
```diff
@@ -209,40 +209,58 @@ def test_testrun_deletes_no_users(self):
...
+        # testupdate should have been renamed
+        self.assertFalse(UserProfile.objects.filter(email='testupdate@institution.example.com').exists())
+        self.assertTrue(UserProfile.objects.filter(email='testupdate@internal.example.com').exists())
+
+        # contributor1 should still be active, contributor2 should have been set to inactive
+        self.assertTrue(UserProfile.objects.get(email='contributor1@institution.example.com').is_active)
+        self.assertFalse(UserProfile.objects.get(email='contributor2@institution.example.com').is_active)
+        self.assertEqual(UserProfile.objects.filter(is_active=True).count(), user_count_before - 1)
+
+        self.assertEqual(UserProfile.objects.count(), user_count_before)  # one deletion, one addition
```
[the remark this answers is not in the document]
janno42: email addresses change as well

evap/staff/tools.py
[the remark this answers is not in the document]
janno42: it actually can happen if "shadow users" are created to have multiple contributions in the same evaluation. when testing the file, the admins will see a message which of the userprofiles will be used for changes if multiple are matching.

evap/staff/tools.py
```diff
@@ -68,31 +69,95 @@ def delete_navbar_cache_for_users(users):
...
+            emails_of_users_to_be_created.append(imported_email)
+        elif matching_user.email != imported_email:
+            users_to_be_updated.append((matching_user, imported_email))
+            emails_of_non_obsolete_users.append(matching_user.email)  # add old address so this user won't be deleted
+
+    obsolete_users = UserProfile.objects.exclude(email__in=emails_of_non_obsolete_users)
+    deletable_users = [user for user in obsolete_users if user.can_be_deleted_by_manager]
+    users_to_mark_inactive = [
+        user for user in obsolete_users
+        if user not in deletable_users and user.is_active and user.can_be_marked_inactive_by_manager
```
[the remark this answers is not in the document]
janno42: properties aren't callable. i partition it in place now.

evap/staff/tools.py
janno42: i find it actually strange to do the fuzzy match at all and would like to keep it only where absolutely needed. in the long run i hope we can completely remove it from the code when the users in our productive environment will get unified email addresses (which hopefully happens soon).

evap/staff/tools.py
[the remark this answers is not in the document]
karyon: I didn't quite understand what you saying but perhaps `split_on_condition(obsolete_users, lambda user: user.can_be_deleted_by_manager):`? though I see that this is a little wordy.

janno42:
i addressed some of the comments.
however, i didn't do any validity checks on the emails (we are not doing those in the importers either, do we?) and also didn't implement the forced test run.
i'd be grateful if someone else wants to do that or if we decide to just leave it be because i don't really feel like continue working on it.

evap/staff/tools.py
He3lixxx: Fair enough, I think that is reasonable. (Email-issue: #1374)

evap/staff/tests/test_views.py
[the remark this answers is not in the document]
He3lixxx: I guess best we can do then is compare the set of IDs?

evap/staff/tools.py
```diff
@@ -68,31 +69,95 @@ def delete_navbar_cache_for_users(users):
...
+    if users_to_be_updated:
+        messages.info(
+            request,
+            mark_safe(
+                _('Users to be updated are:<br />{}')
+                .format('<br />'.join(
+                    [_("{} > {}").format(user.email, email) for user, email in users_to_be_updated]
+                ))
+            )
+        )
```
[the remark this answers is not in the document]
janno42: is this okay for you in the current version?

evap/staff/tools.py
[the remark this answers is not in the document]
janno42: is this still a problem?

evap/staff/tools.py
```diff
@@ -68,31 +69,95 @@ def delete_navbar_cache_for_users(users):
...
+
+    emails_of_users_to_be_created = []
+    users_to_be_updated = []
+    emails_of_non_obsolete_users = list(imported_emails)
+    skipped_external_emails_counter = 0
+
+    for imported_email in imported_emails:
+        if is_external_email(imported_email):
+            skipped_external_emails_counter += 1
+            continue
```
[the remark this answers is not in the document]
janno42: is this still a problem (see also below)?

He3lixxx:
> perhaps check whether all of the input is valid email addresses?

@karyon I'd say if we want to enforce correct emails, we'd have to do it everywhere (otherwise it might happen that a user has an invalid email and we can't reference them in the bulk update). However, checking it everywhere disallows us from having a development-user "evap" since that obviously isn't a valid email. Thus, I'd say it's fine to not check the email validity here. What do you think?

evap/staff/tools.py
```diff
@@ -68,31 +69,95 @@ def delete_navbar_cache_for_users(users):
...
+            UserProfile.objects.count(), len(users_to_be_updated), len(deletable_users), len(users_to_mark_inactive),
+            len(emails_of_users_to_be_created))
+    )
+    if users_to_be_updated:
+        messages.info(
+            request,
+            mark_safe(
+                _('Users to be updated are:<br />{}')
+                .format('<br />'.join(
+                    [_("{} > {}").format(user.email, email) for user, email in users_to_be_updated]
```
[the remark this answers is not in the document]
He3lixxx: I think this is resolved in the current version?

evap/staff/tools.py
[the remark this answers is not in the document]
janno42: thinking about it again: i can't find a valid usecase in which we have multiple matching userprofiles anymore. so we can refuse to process the file as suggested.

evap/staff/tests/test_views.py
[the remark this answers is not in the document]
He3lixxx: It now compares the set of UserProfile objects, which boils down to comparing the IDs.

evap/staff/tools.py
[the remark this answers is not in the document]
He3lixxx: It now gives an error when more than one user match a given mail address.

He3lixxx:
> also, consider putting the test-run-file on the server and using that on the actual import instead of letting the user upload another file without dry-run, to prevent user errors.

implemented.

evap/staff/views.py
```diff
@@ -1539,20 +1539,30 @@ def user_delete(request):
...
+            form.file_required = True
+            if form.is_valid():
+                username_file = form.cleaned_data['username_file']
+                file_content = username_file.read()
+                success = bulk_update_users(request, file_content, test_run)
+                if success:
+                    save_import_file(username_file, request.user.id, import_type)
+        else:
+            file_content = get_import_file_content_or_raise(request.user.id, import_type)
+            success = bulk_update_users(request, file_content, test_run)
```
janno42: you don't need that variable, right?

evap/staff/tools.py
```diff
@@ -25,8 +27,16 @@ def forward_messages(request, success_messages, warnings):
...
+    User = 'user'
+    Contributor = 'contributor'
+    Participant = 'participant'
+    Semester = 'semester'
+    UserBulkUpdate = 'user_bulk_update'
+
+
 def generate_import_filename(user_id, import_type):
-    return settings.MEDIA_ROOT + '/temp_import_files/' + str(user_id) + '.xls' + '.' + import_type
+    return settings.MEDIA_ROOT + '/temp_import_files/' + str(user_id) + '.xls' + '.' + import_type.value
```
pixunil, line 39: Wouldn’t this be a place to use `os.path.join` or a f-string? Also, it feels a bit counter-intuitive to put the import type after the `.xls` extension rather than between the filename and the “real” extension. But if this is just how conventions are, I’m fine like it is

evap/staff/tests/test_views.py
```diff
@@ -24,7 +24,10 @@
 
 
 def helper_delete_all_import_files(user_id):
-    file_filter = generate_import_filename(user_id, "*")
+    class AllFiles:
+        value = "*"  # used for globbing all existing files in tests.
+
+    file_filter = generate_import_filename(user_id, AllFiles)
```
pixunil, line 30: I like this clever approach, but to propose a more “traditional” approach: You could iterate over the whole `ImportType` enum and remove the corresponding file if it exists. You probably would also get ride of passing it to `glob`

evap/staff/tests/test_views.py
```diff
@@ -185,64 +188,136 @@ def setUpTestData(cls):
...
+            'Multiple users match the email testuser1@institution.example.com:'
+            + '<br />None None (testuser1@institution.example.com)'
+            + '<br />None None (testuser1@internal.example.com)'
+        )
+        button_substring = 'value="bulk_update"'
 
-    def test_deletes_users(self):
-        baker.make(UserProfile, username='testuser1')
-        baker.make(UserProfile, username='testuser2')
+        expected_users = set(UserProfile.objects.all())
```
pixunil, line 230: This variable seams not to be checked after `new_user` is added to it. Did you forgot to check it after the test run (which wouldn’t add that much value imo)? In the other case, can you move it down right before the assertion?

evap/staff/tests/test_views.py
```diff
@@ -185,64 +188,136 @@ def setUpTestData(cls):
...
+
+    @override_settings(INSTITUTION_EMAIL_DOMAINS=["institution.example.com", "internal.example.com"])
+    def test_multiple_email_matches_trigger_error(self):
+        baker.make(UserProfile, email='testremove@institution.example.com')
+        baker.make(UserProfile, email='testuser1@institution.example.com')
+
+        error_string = (
+            'Multiple users match the email testuser1@institution.example.com:'
+            + '<br />None None (testuser1@institution.example.com)'
+            + '<br />None None (testuser1@internal.example.com)'
```
pixunil, line 226: This is more my personal taste: Could we give testuser1 a first and last name? I know, it is not really neccessary for this test, but the None looks a bit odd to me

evap/staff/forms.py
```diff
@@ -78,8 +78,16 @@ def clean(self):
...
 
 
-class UserBulkDeleteForm(forms.Form):
-    username_file = forms.FileField(label=_("Username file"))
+class UserBulkUpdateForm(forms.Form):
+    username_file = forms.FileField(label=_("Username file"), required=False)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.file_required = False
```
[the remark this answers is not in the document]
He3lixxx: Nice idea, added a commit that replaces this additional variable with the existing `required` attribute for all the cases where this mechanism is currently used.

evap/staff/tools.py
He3lixxx: Good point. I think I never really spend a thought on the order here.

evap/staff/tests/test_views.py
He3lixxx: Yup, makes sense, now that we can actually iterate the enum.

evap/staff/tests/test_views.py
He3lixxx: Yup, I thought that testing after both test runs are done also implies that no user was lost in the middle, but you're right, making it explicit is a bit nicer. Added.

evap/staff/tests/test_views.py
```diff
@@ -185,64 +187,137 @@ def setUpTestData(cls):
...
+    @override_settings(INSTITUTION_EMAIL_DOMAINS=["institution.example.com", "internal.example.com"])
+    def test_handles_users(self):
+        baker.make(UserProfile, email='testuser1@institution.example.com')
+        baker.make(UserProfile, email='testuser2@institution.example.com')
+        baker.make(UserProfile, email='testupdate@institution.example.com')
         contribution1 = baker.make(Contribution)
         semester = baker.make(Semester, participations_are_archived=True)
-        evaluation = baker.make(Evaluation, course=baker.make(Course, semester=semester), _participant_count=0, _voter_count=0)
+        evaluation = baker.make(Evaluation,
+            course=baker.make(Course, semester=semester), _participant_count=0, _voter_count=0
```
pixunil, line 260: Only a code style hint: I would give each keyword argument an own line. As I first read this line, I was confused that `course` accepts a `_participant_count` argument

evap/staff/tests/test_views.py
```diff
@@ -185,64 +187,137 @@ def setUpTestData(cls):
...
         contribution2 = baker.make(Contribution, evaluation=evaluation)
-        baker.make(UserProfile, username='contributor1', contributions=[contribution1])
-        baker.make(UserProfile, username='contributor2', contributions=[contribution2])
+        baker.make(UserProfile, email='contributor1@institution.example.com', contributions=[contribution1])
+        baker.make(UserProfile, email='contributor2@institution.example.com', contributions=[contribution2])
 
-        page = self.app.get(self.url, user='manager')
-        form = page.forms["user-bulk-delete-form"]
+        expected_users = set(UserProfile.objects.all())
+        expected_users.discard(UserProfile.objects.get(email='testuser2@institution.example.com'))
```
pixunil, line 267: What about `UserProfile.objects.exclude`? Or did you want to raise an error if `testuser2@institutation.example.com` does not exists?

evap/staff/tests/test_views.py
```diff
@@ -185,64 +187,137 @@ def setUpTestData(cls):
...
-        self.assertFalse(UserProfile.objects.filter(username='testuser2').exists())
+        # testuser1 is in the file and must not be deleted
+        self.assertTrue(UserProfile.objects.filter(email='testuser1@institution.example.com').exists())
+        # testuser2 is not in the file and must be deleted
+        self.assertFalse(UserProfile.objects.filter(email='testuser2@institution.example.com').exists())
+        # manager is not in the file but still must not be deleted
         self.assertTrue(UserProfile.objects.filter(username='manager').exists())
+        # testusernewinternal is a new internal user and should be created
+        expected_users.add(UserProfile.objects.get(email='testusernewinternal@institution.example.com'))
+        self.assertTrue(UserProfile.objects.filter(email='testusernewinternal@institution.example.com').exists())
```
pixunil, line 291: This assert should come before the `get` one line above, otherwise the test will error via the get instead of fail via the assert

evap/staff/tests/test_views.py
```diff
@@ -327,7 +402,7 @@ def test_invalid_upload_operation(self):
         form = page.forms["user-import-form"]
         page = form.submit(name="operation", value="test")
 
-        self.assertContains(page, 'Please select an Excel file')
+        self.assertContains(page, 'This field is required.')
```
pixunil, line 405: I did not check myself, but would it be possible to check `form.errors`, so that we also check that the message is also on the correct field?
(Also applies to later occurrences)

PR #1416, opened 2020-01-31
Send review notice to responsibles with editors in cc

### LINKED ISSUE #1338: Send editor review notice to responsible with others in CC
When sending an editor review notice email to all editors of an evaluation, this message should be addressed to the responsible(s) and CC'ed to all other editors.
Currently the message is sent to each editor individually.

pixunil:
Closes #1338.

### DIFF
--- a/evap/staff/tests/test_views.py
+++ b/evap/staff/tests/test_views.py
@@ -860,6 +860,8 @@ class TestEvaluationOperationView(WebTest):
     def setUpTestData(cls):
         baker.make(UserProfile, username='manager', groups=[Group.objects.get(name='Manager')])
         cls.semester = baker.make(Semester, pk=1)
+        cls.responsible = baker.make(UserProfile, email='responsible@example.com')
+        cls.course = baker.make(Course, semester=cls.semester, responsibles=[cls.responsible])
 
     def helper_publish_evaluation_with_publish_notifications_for(self, evaluation, contributors=True, participants=True):
         page = self.app.get("/staff/semester/1", user="manager")
@@ -881,8 +883,7 @@ class TestEvaluationOperationView(WebTest):
         participant2 = baker.make(UserProfile, email="bar@example.com")
         contributor1 = baker.make(UserProfile, email="contributor@example.com")
 
-        course = baker.make(Course, semester=self.semester)
-        evaluation = baker.make(Evaluation, course=course, state='reviewed',
+        evaluation = baker.make(Evaluation, course=self.course, state='reviewed',
                                 participants=[participant1, participant2], voters=[participant1, participant2])
         baker.make(Contribution, contributor=contributor1, evaluation=evaluation)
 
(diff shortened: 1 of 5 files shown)


evap/staff/views.py
```diff
@@ -220,7 +220,18 @@ def apply(request, evaluations, email_template=None, email_template_contributor=
         messages.success(request, ungettext("Successfully enabled {} evaluation for editor review.",
             "Successfully enabled {} evaluations for editor review.", len(evaluations)).format(len(evaluations)))
         if email_template:
-            email_template.send_to_users_in_evaluations(evaluations, [EmailTemplate.EDITORS], use_cc=True, request=request)
+            evaluations_by_responsible = {}
```
karyon: have you considered moving that logic into send_to_users_in_evaluations?

evap/staff/tests/test_views.py
```diff
@@ -966,18 +969,116 @@ def test_operation_start_evaluation(self):
...
         response = self.app.get(self.url + urloptions, user='manager')
         self.assertEqual(response.status_code, 200, 'url "{}" failed with user "manager"'.format(self.url))
-
         form = response.forms['evaluation-operation-form']
         form.submit()
 
         evaluation = Evaluation.objects.get(pk=evaluation.pk)
         self.assertEqual(evaluation.state, 'prepared')
 
+    def submit_operation_prepare_form(self, url_options):
```
karyon: this method could return the emails directly instead of going through self.actual_emails.

evap/staff/tests/test_views.py
```diff
@@ -967,18 +967,119 @@ def test_operation_start_evaluation(self):
...
+    def test_operation_prepare_does_not_put_responsible_into_cc(self):
+        evaluation = baker.make(Evaluation, state='new', course=self.course)
+        baker.make(Contribution, evaluation=evaluation, contributor=self.responsible, can_edit=True)
+        url_options = '?evaluation={}&target_state=prepared'.format(evaluation.pk)
+        actual_emails = self.submit_operation_prepare_form(url_options)
+
+        self.assertEqual(len(actual_emails), 1)
+        self.assertEqual(actual_emails[0]['additional_cc_users'], set())
+
+    def test_operation_prepare_does_not_sends_email_to_contributors(self):
```
He3lixxx: `sends` should be `send`

evap/staff/views.py
pixunil: Yes, I considered it, but I currently do not see a great benefit here as this case currentlyexists only once.
If we want to keep the abstraction level of `send_to_users_in_evaluations`, a new parameter would be necessary, like `cc_groups` which controls which groups should be CC’ed.
For editors, we decided for some trade-off here as some editors may receive mails whereas only one displayed evaluation is relevant to them. It may be fine in this particular case, but it should be reconsidered for other groups.
Of course, we don’t have to choose this rather abstract level, but then we have a parameter which is mostly unused and makes `send_to_users_in_evaluations` more verbose.
Or do you have another option in your mind?

PR #1421, opened 2020-02-03
Evaluation text answers export

### LINKED ISSUE #1383: Evaluation TextAnswer export
There should be an option on an Evaluation's result page to export all currently visible TextAnswers (in "full" and "export" view).
This should download an Excel file including name, semester and responsible of the evaluation followed by a two-column table with the question's text on the left and all answers in separate lines on the right.
Top general questionnaires, Contributor questionnaires, and Bottom general questionnaires should be separated by horizontal lines.

niklasmohrin:
Fixes #1383

### DIFF
--- a/evap/results/exporters.py
+++ b/evap/results/exporters.py
@@ -1,4 +1,4 @@
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 
 from django.db.models import Q
 from django.utils.translation import gettext as _
@@ -284,3 +284,68 @@ def _write(exporter, label, style, rows, cols):
         exporter.col += cols - 1
     else:
         exporter.sheet.write(exporter.row, exporter.col, label, style)
+
+
+class TextAnswerExcelExporter:
+    # pylint: disable=too-many-instance-attributes
+
+    class InputData:
+        def __init__(self, contribution_results):
+            self.questionnaires = defaultdict(list)
+
+            for contribution_result in contribution_results:
+                contributor_name = contribution_result.contributor.full_name if contribution_result.contributor is not None else ""
+                for questionnaire_result in contribution_result.questionnaire_results:
+                    for result in questionnaire_result.question_results:
+                        q_type = result.question.questionnaire.type
+                        answers = [answer.answer for answer in result.answers]
+                        self.questionnaires[q_type].append((contributor_name, result.question, answers))
+
+    def __init__(self, evaluation_name, semester_name, responsibles, results, contributor_name):
+        self.evaluation_name = evaluation_name
+        self.semester_name = semester_name
+        self.responsibles = responsibles
+        assert isinstance(results, TextAnswerExcelExporter.InputData)
+        self.results = results
+        self.contributor_name = contributor_name
+
+        self.styles = {
+            "default": xlwt.Style.default_style,
+            "border_top": xlwt.easyxf("borders: top medium"),
+        }
+
+        self.workbook = xlwt.Workbook()
+        self.sheet = self.workbook.add_sheet(_("Text Answers"))
+        self.row = 0
+        self.col = 0
+        self.sheet.col(0).width = 10000
+        self.sheet.col(1).width = 40000
+
+    def export(self, response):
+        writec(self, self.evaluation_name, "default")
+        writen(self, self.semester_name, "default")
+        writen(self, self.responsibles, "default")
+        if self.contributor_name is not None:
+            writen(self, _("Export for {}").format(self.contributor_name), "default")
+
+        for questionnaire_type in Questionnaire.Type.values:
+            # The first line of every questionnaire type should have an overline.
+            line_style = "border_top"
+
+            for (contributor_name, question, answers) in self.results.questionnaires[questionnaire_type]:
+                # The first line of every question should contain the
+                # question text and contributor name (if present).
+                first_cell = ""
+                if contributor_name:
+                    first_cell += f"{contributor_name}: "
+                first_cell += question.text
+
+                for answer in answers:
+                    writen(self, first_cell, line_style)
+                    writec(self, answer, line_style)
+
+                    # after the first line, these should reset to their defaults
+                    first_cell = ""
+                    line_style = "default"
+
+        self.workbook.save(response)
--- a/evap/results/templates/results_evaluation_detail.html
+++ b/evap/results/templates/results_evaluation_detail.html
@@ -72,10 +72,15 @@
     <div class="card card-outline-primary mb-3">
         <div class="card-header d-flex">
             <div class="mr-auto">{% trans 'Overview' %}</div>
+            {% if can_export_text_answers %}
+                <a class="btn btn-sm btn-light d-print-none" href="{% url 'results:evaluation_text_answers_export' evaluation.id %}?view={{ view }}{% if contributor_id is not None %}&contributor_id={{ contributor_id }}{% endif %}" type="button">
+                    {% trans 'Export text answers' %}
+                </a>
+            {% endif %}
             {% if evaluation.course.grade_documents.count == 1 and can_download_grades %}
-                <a class="btn btn-sm btn-light d-print-none" href="{% url 'grades:download_grades' evaluation.course.grade_documents.first.id %}">{{ evaluation.course.grade_documents.first.description }}</a>
+                <a class="btn btn-sm btn-light d-print-none ml-2" href="{% url 'grades:download_grades' evaluation.course.grade_documents.first.id %}">{{ evaluation.course.grade_documents.first.description }}</a>
             {% elif evaluation.course.grade_documents.count > 1 and can_download_grades %}
-                <div class="btn-group d-print-none" role="group">
+                <div class="btn-group d-print-none ml-2" role="group">
                     <button type="button" id="btnDownload" role="button" class="btn btn-sm btn-light dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">{% trans 'Grades' %}</button>
                     <div class="dropdown-menu" aria-labelledby="btnDownload">
                         {% for grade_document in evaluation.course.grade_documents.all %}
--- a/evap/results/tests/test_exporters.py
+++ b/evap/results/tests/test_exporters.py
@@ -1,4 +1,5 @@
 from io import BytesIO
+
 from model_bakery import baker
 from django.test import TestCase
 from django.utils import translation
@@ -7,8 +8,10 @@ import xlrd
 
 from evap.contributor.views import export_contributor_results
 from evap.evaluation.models import (Contribution, Course, CourseType, Degree, Evaluation, Question, Questionnaire,
-                                    RatingAnswerCounter, Semester, UserProfile)
-from evap.results.exporters import ExcelExporter
+                                    RatingAnswerCounter, Semester, UserProfile, TextAnswer)
+from evap.results.exporters import ExcelExporter, TextAnswerExcelExporter
+from evap.results.tools import collect_results
+from evap.results.views import filter_text_answers
 
 
 class TestExporters(TestCase):
@@ -254,3 +257,39 @@ class TestExporters(TestCase):
         self.assertEqual(workbook.sheets()[0].row_values(8)[2], 3.0)
         self.assertEqual(workbook.sheets()[0].row_values(10)[0], "Overall Average Grade")
         self.assertEqual(workbook.sheets()[0].row_values(10)[2], 3.25)
+
+    def test_text_answer_export(self):
+        evaluation = baker.make(Evaluation, can_publish_text_results=True)
+        questions = [baker.make(Question, questionnaire__type=t, type=Question.TEXT) for t in Questionnaire.Type.values]
+
+        for idx in [0, 1, 2, 2, 0]:
+            baker.make(
+                TextAnswer,
+                question=questions[idx],
+                contribution__evaluation=evaluation,
+                contribution__questionnaires=[questions[idx].questionnaire],
+                state=TextAnswer.State.PUBLISHED
+            )
+
+        evaluation_result = collect_results(evaluation)
+        filter_text_answers(evaluation_result)
+
+        results = TextAnswerExcelExporter.InputData(evaluation_result.contribution_results)
+
+        binary_content = BytesIO()
+        TextAnswerExcelExporter(evaluation.name, evaluation.course.semester.name,
+                                evaluation.course.responsibles_names,
+                                results, None).export(binary_content)
+        binary_content.seek(0)
+        workbook = xlrd.open_workbook(file_contents=binary_content.read())
+        sheet = workbook.sheets()[0]
+
+        # Sheet headline
+        self.assertEqual(sheet.row_values(0)[0], evaluation.name)
+        self.assertEqual(sheet.row_values(1)[0], evaluation.course.semester.name)
+        self.assertEqual(sheet.row_values(2)[0], evaluation.course.responsibles_names)
+
+        # Questions are ordered by questionnaire type, answers keep their order respectively
+        self.assertEqual(sheet.row_values(3)[0], questions[0].text)
+        self.assertEqual(sheet.row_values(5)[0], questions[1].text)
+        self.assertEqual(sheet.row_values(6)[0], questions[2].text)
--- a/evap/results/tests/test_views.py
+++ b/evap/results/tests/test_views.py
@@ -13,9 +13,11 @@ from django.test.utils import CaptureQueriesContext
 from django_webtest import WebTest
 from model_bakery import baker
 
-from evap.evaluation.models import (Contribution, Course, Degree, Evaluation, Question, Questionnaire, RatingAnswerCounter,
+from evap.evaluation.models import (Contribution, Course, Degree, Evaluation, Question, Questionnaire,
+                                    RatingAnswerCounter,
                                     Semester, UserProfile)
 from evap.evaluation.tests.tools import WebTestWith200Check, let_user_vote_for_evaluation
+from evap.results.exporters import TextAnswerExcelExporter
 from evap.results.views import get_evaluations_with_prefetched_data
 
 
@@ -717,3 +719,36 @@ class TestArchivedResults(WebTest):
         self.app.get(url, user='manager', status=200)
         self.app.get(url, user='reviewer', status=403)
         self.app.get(url, user='student_external', status=403)
+
+
+class TestTextAnswerExportView(WebTest):
+
+    @classmethod
+    def setUpTestData(cls):
+        baker.make(UserProfile, username="reviewer", groups=[Group.objects.get(name="Reviewer")])
+        evaluation = baker.make(Evaluation)
+        cls.url = f"/results/evaluation/{evaluation.id}/text_answers_export"
+
+    def test_file_sent(self):
+        def mock(_self, res):
+            res.write(b"1337")
+
+        with patch.object(TextAnswerExcelExporter, "export", mock):
+            response = self.app.get(self.url, user="reviewer", status=200)
+            self.assertEqual(response.headers["Content-Type"], "application/vnd.ms-excel")
+            self.assertEqual(response.content, b"1337")
+
+    @patch("evap.results.exporters.TextAnswerExcelExporter.export")
+    def test_permission_denied(self, export_method):
+        baker.make(UserProfile, username="student")
+        baker.make(UserProfile, username="manager", groups=[Group.objects.get(name="Manager")])
+
+        self.app.get(self.url, user="student", status=403)
+        export_method.assert_not_called()
+
+        self.app.get(self.url, user="reviewer", status=200)
+        export_method.assert_called_once()
+
+        export_method.reset_mock()
+        self.app.get(self.url, user="manager", status=200)
+        export_method.assert_called_once()
--- a/evap/results/views.py
+++ b/evap/results/views.py
@@ -179,7 +179,7 @@ def evaluation_detail(request, semester_id, evaluation_id):
 
     evaluation_result = collect_results(evaluation)
     remove_textanswers_that_the_user_must_not_see(evaluation_result, view_as_user, represented_users, view)
-    filter_empty_headings(evaluation_result)
+    exclude_empty_headings(evaluation_result)
     remove_empty_questionnaire_and_contribution_results(evaluation_result)
     add_warnings(evaluation, evaluation_result)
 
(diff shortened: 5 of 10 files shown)


niklasmohrin:
View tests still work in progress

evap/results/templates/results_evaluation_detail.html
```diff
@@ -28,8 +28,16 @@
         {% endif %}
         <div class="d-flex">
             <h3 class="mr-auto">{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</h3>
-
-            <div class="btn-switch btn-switch-light my-auto d-print-none">
+        </div>
+        <div class="d-flex mb-3">
```
janno42: please place the button next to the grade download buttons in the overview card div

evap/results/views.py
```diff
@@ -336,3 +324,61 @@ def add_warnings(evaluation, evaluation_result):
...
+
+    return results
+
+
+def evaluation_text_answers_export(request, evaluation_id):
+    evaluation = get_object_or_404(Evaluation, id=evaluation_id)
+
+    results = evaluation_to_question_answer_dict(request, evaluation)
+
+    filename = "Evaluation-Text-Answers-{}-{}.xls".format(evaluation.name, get_language())
```
janno42: the file name should include the semester short name

evap/results/exporters.py
```diff
@@ -284,3 +284,45 @@ def _write(exporter, label, style, rows, cols):
...
+
+        self.workbook = xlwt.Workbook()
+        self.sheet = self.workbook.add_sheet(_("Text Answers"))
+        self.row = 0
+        self.col = 0
+        self.sheet.col(0).width = 10000
+        self.sheet.col(1).width = 40000
+
+    def export(self, response):
+        writec(self, self.evaluation_name, "default")
```
janno42: this should be the `full_name` of the evaluation

evap/results/views.py
janno42: and should use the `full_name` of the evaluation as well

evap/results/exporters.py
niklasmohrin: The view calls this with
```python
TextAnswerExcelExporter(evaluation.name, evaluation.course.semester.name,
                            evaluation.course.responsibles_names, results).export(response)
```
and Semester only has `name` and `short_name` , right?

niklasmohrin: Nevermind, caller provides the `full_name` now

evap/results/views.py
```diff
@@ -339,3 +330,61 @@ def add_warnings(evaluation, evaluation_result):
...
+
+    results = dict()
+    for questionnaire_result in evaluation_result.questionnaire_results:
+        for result in questionnaire_result.question_results:
+            results[result.question] = [answer.answer for answer in result.answers]
+
+    return results
+
+
+def evaluation_text_answers_export(request, evaluation_id):
```
He3lixxx: What about access control? I guess not all users should be able to access this view?

evap/results/tests/test_exporters.py
```diff
@@ -254,3 +255,41 @@ def test_contributor_result_export(self):
         self.assertEqual(workbook.sheets()[0].row_values(8)[2], 3.0)
         self.assertEqual(workbook.sheets()[0].row_values(10)[0], "Overall Average Grade")
         self.assertEqual(workbook.sheets()[0].row_values(10)[2], 3.25)
+
+    def test_text_answer_export(self):
+        evaluation = baker.make(Evaluation, name_en="eval_en", name_de="eval_de")
+        questions = [
```
He3lixxx: Have you considered
```python
questions = [baker.make(Question, questionnaire__type=type_id) for type_id in [10, 20, 30]]
```
?

evap/results/tests/test_exporters.py
```diff
@@ -254,3 +255,41 @@ def test_contributor_result_export(self):
...
+            baker.make(Question, questionnaire__type=20),
+            baker.make(Question, questionnaire__type=30),
+        ]
+        text_answers = [
+            baker.make(TextAnswer, question=questions[0], contribution__evaluation=evaluation),
+            baker.make(TextAnswer, question=questions[1], contribution__evaluation=evaluation),
+            baker.make(TextAnswer, question=questions[2], contribution__evaluation=evaluation),
+            baker.make(TextAnswer, question=questions[2], contribution__evaluation=evaluation),
+            baker.make(TextAnswer, question=questions[0], contribution__evaluation=evaluation),
+        ]
```
He3lixxx: similarly here:
```python
text_answers = [
    baker.make(TextAnswer, question=questions[question_idx], contribution__evaluation=evaluation)
    for question_idx in [0, 0, 1, 2, 2]
]
```

evap/results/tests/test_exporters.py
```diff
@@ -254,3 +255,41 @@ def test_contributor_result_export(self):
...
+
+        results = defaultdict(list)
+        for text_answer in text_answers:
+            question_text = text_answer.question
```
He3lixxx: This is smelly, question is an actual database object, not the text of the question.

evap/results/tests/test_exporters.py
```diff
@@ -254,3 +255,41 @@ def test_contributor_result_export(self):
...
+        binary_content.seek(0)
+        workbook = xlrd.open_workbook(file_contents=binary_content.read())
+        sheet = workbook.sheets()[0]
+
+        # Sheet headline
+        self.assertEqual(sheet.row_values(0)[0], evaluation.name)
+        self.assertEqual(sheet.row_values(1)[0], evaluation.course.semester.name)
+        self.assertEqual(sheet.row_values(2)[0], evaluation.course.responsibles_names)
+
+        # Questions, ordered by questionnaire type (answers must not be ordered)
```
He3lixxx: "(answers must not be ordered)" is supposed to mean "The excel-importer should use the order given by the results list", right? If so, I'd change this to:
```python
# Questions, ordered by questionnaire type. Answers keep their order from results list
```
or similar. What do you think?

evap/results/tests/test_views.py
```diff
@@ -699,3 +701,34 @@ def test_archived_results(self):
...
         self.app.get(url, user='student_external', status=403)
+
+
+class TestTextAnswerExportView(WebTest):
+
+    @classmethod
+    def setUpTestData(cls):
+        baker.make(UserProfile, username="student")
+        baker.make(UserProfile, username="reviewer", groups=[Group.objects.get(name="Reviewer")])
+        baker.make(UserProfile, username="manager", groups=[Group.objects.get(name="Manager")])
```
He3lixxx: student and manager are only used in one test - what do you think about moving their creation in that test method? From my experience, we usually struggle with too much set up being done in the set up methods, not the other way around.

evap/results/views.py
```diff
@@ -50,7 +53,7 @@ def _delete_course_template_cache_impl(course):
 
 def warm_up_template_cache(evaluations):
     evaluations = get_evaluations_with_course_result_attributes(get_evaluations_with_prefetched_data(evaluations))
-    current_language = translation.get_language()
+    current_language = get_language()
```
He3lixxx: What's the reason for this change? We need one additional import this way, right?

niklasmohrin: I also used `get_language` down in `evaluation_text_answers_export`:
```python
filename = "Evaluation-Text-Answers-{}-{}-{}.xls".format(evaluation.course.semester.short_name, evaluation.full_name, get_language())
```
to make the line shorter. Should I just expand both to `translation.get_language()` or leave it as is?

evap/results/tests/test_exporters.py
niklasmohrin: This is how `TextAnswerExcelExporter` expects the input as of now.
Should I rather change it to a dict mapping from the tuple (question_text, question_type) to the answers, or a dict mapping from the type to (dict from question text to answers)? Or something completely different?

evap/results/views.py
niklasmohrin: A `PermissionDenied` would be thrown later down the road:
```python
def evaluation_detail_parse_get_parameters(request, evaluation):
    if not evaluation.can_results_page_be_seen_by(request.user):
        raise PermissionDenied
```

 I still added `@reviewer_required`.

evap/results/tests/test_exporters.py
niklasmohrin: Since the questions are fetched once when rendering the site and again when generating the Excel File, I am not sure if the answers actually are in the same order. Further, the order of the answers is not tested here (or anywhere else).
Still, this comment is more accurate, since the exporter does not change the order, so I will adapt it.

evap/results/tests/test_exporters.py
He3lixxx: I'm fine with the way you handle the data and the way the exporter want the data. My annotation was just about the local variable name `question_text`, which is misleading, since `text_answer.question` does not give you the question text, but a Question object. Since you're just using the variable in this for loop, renaming it would solve this... I guess we could just omit the variable:

```python
for text_answer in text_answers:
   results[text_answer.question].append(text_answer.answer)
```

Or am I missing something?

evap/results/views.py
He3lixxx: Ah okay. I'd leave the import as it was and use `translation.get_language()`. The line length could be reduced by just adding some newlines
```python
filename = "Evaluation-Text-Answers-{}-{}-{}.xls".format(
    evaluation.course.semester.short_name,
    evaluation.full_name,
    translation.get_language(),
)
```

What do you think about that?

evap/results/tests/test_exporters.py
niklasmohrin: You are right, this makes little sense 🤔

evap/results/views.py
niklasmohrin: Fair enough

evap/results/tests/test_views.py
```diff
@@ -699,3 +701,35 @@ def test_archived_results(self):
...
+
+
+class TestTextAnswerExportView(WebTest):
+
+    @classmethod
+    def setUpTestData(cls):
+        baker.make(UserProfile, username="reviewer", groups=[Group.objects.get(name="Reviewer")])
+        evaluation = baker.make(Evaluation)
+
+        baker.make(TextAnswer, _quantity=3, __contribution__evaluation=evaluation)
```
He3lixxx: The two leading underscores in `__contribution__evaluation` are smelly, I quickly checked and it looks like baker does not use that argument. (This is really strange as it also doesn't give an error, I would expect that it somehow gives an error if you pass it some kwarg that it does not understand -- will look into that a bit more later).

Do you need these TextAnswers? From what I see, you patch / mock the export method in both tests, so they shouldn't be necessary - is that correct?

evap/results/tests/test_views.py
```diff
@@ -699,3 +701,35 @@ def test_archived_results(self):
...
+    def test_permission_denied(self, export_method):
+        baker.make(UserProfile, username="student")
+        baker.make(UserProfile, username="manager", groups=[Group.objects.get(name="Manager")])
+
+        self.app.get(self.url, user="student", status=403)
+        export_method.assert_not_called()
+        self.app.get(self.url, user="reviewer", status=200)
+        export_method.assert_called_once()
+        self.app.get(self.url, user="manager", status=200)
+        self.assertEqual(export_method.call_count, 2)
```
He3lixxx: I think this is fine here - just as a future reference: The mocks have a `reset_mock` method, allowing you to reset the recorded calls.

evap/results/templates/results_evaluation_detail.html
```diff
@@ -72,6 +72,13 @@ <h3>{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</h3>
     <div class="card card-outline-primary mb-3">
         <div class="card-header d-flex">
             <div class="mr-auto">{% trans 'Overview' %}</div>
+            {% if view == "export" or view == "full" %}
+                <div class="ml-auto mr-3 my-auto">
```
janno42: please pay attention that the button is currently changing the card-header's size which it shouldn't. also, please include `d-print-none`, as the button should not be visible when printing the results page.

niklasmohrin:
I came across a problem, where
```response["Content-Disposition"] = "attachment; filename=\"{}\"".format(filename)```
results in something like this:
```
Content-Disposition: =?utf-8?q?attachment=3B_filename=3D=22Evaluation-Text-Answers-ST_14-3D-Computer_Graphic_I_=E2=80=93_Software_Architecture-en=2Exls=22?=
```
I fixed it by turning the filename into bytes (see last commit; will be squashed later). Is this normal behaviour (I remember it working without the fix)?

**Comment by niklasmohrin** [comment]
Also, this feature is not ready yet.

**Comment by niklasmohrin** [comment]
@janno42 
I added three more commits that should add contributor names, but the code is very cluttered now. Maybe I should add a class to hold the data that `TextAnswerExcelExporter` expects?

Do I need to test the contributor names appearance in the file?

File: evap/results/views.py [hunk 14]
```diff
@@ -339,3 +329,71 @@ def add_warnings(evaluation, evaluation_result):
...
+    contributor_name = UserProfile.objects.get(id=contributor_id).full_name if contributor_id is not None else None
+
+    filename = "Evaluation-Text-Answers-{}-{}-{}.xls".format(
+        evaluation.course.semester.short_name,
+        evaluation.full_name,
+        translation.get_language()
+    )
+
+    response = HttpResponse(content_type="application/vnd.ms-excel")
+    response["Content-Disposition"] = bytes("attachment; filename=\"{}\"".format(filename), "utf-8")
```
He3lixxx: I think this is smelly, especially if we can't really explain the behaviour with the old code or why the new code is any better. I'll try to find out why your problem occurred and whether this is a proper solution some time this week.

(I read your comment)

He3lixxx: Alrighty, so the issue is that [RFC 2183](https://tools.ietf.org/html/rfc2183), which defines the content-disposition header, only allows for us-ascii encoded strings in the header. To comply to this, django will format it the way it did for you if it contains non-ascii characters. Manually encoding it to a bytestring before will circumvent django's logic, but we're relying on non-standard behavior in the browsers then.

Django handles it [like this](https://github.com/django/django/blob/master/django/http/response.py#L445) in their FileResponse - sadly, this class is not compatible with having a byte stream in memory, it only supports real files in the file system.
```python
try:
     filename.encode('ascii')
     file_expr = 'filename="{}"'.format(filename)
except UnicodeEncodeError:
     file_expr = "filename*=utf-8''{}".format(quote(filename))

self['Content-Disposition'] = '{}; {}'.format(disposition, file_expr))
```
as defined in RFC 6266. [All modern browsers should support this](http://test.greenbytes.de/tech/tc2231/#attwithfn2231utf8). Thus, I'd say we should also do it like this.

How often do we currently have the case that we manually set the content disposition? Could you put this code snipped into a method (`make_file_download_response` or whatever) and use it in these places? That would be really nice.

evap/results/views.py
```diff
@@ -339,3 +329,71 @@ def add_warnings(evaluation, evaluation_result):
...
+
+    view, view_as_user, represented_users, contributor_id = evaluation_detail_parse_get_parameters(request, evaluation)
+
+    evaluation_result = collect_results(evaluation)
+    filter_text_answers(evaluation_result)
+    remove_textanswers_that_the_user_must_not_see(evaluation_result, view_as_user, represented_users, view)
+
+    results = dict()
+    for contribution_result in evaluation_result.contribution_results:
+        contributor_name = contribution_result.contributor.full_name if contribution_result.contributor is not None else ""
```
He3lixxx: Would  `contributor_name = contribution_result.contributor.full_name  or ""` also do what you need here?

He3lixxx: Ah i see. Never mind. I guess you could still remove the "is not None" part, but the rest is necessary.

evap/results/templates/results_evaluation_detail.html
janno42: the button is now placed too close to the grade document download button. can you please add some spacing?

evap/results/views.py
niklasmohrin: The pattern
```python
response = HttpResponse(content_type=...)
response["Content-Disposition"] = ...
```
is now handled by evaluation.tools.FileResponse (do we use Django's FileResponse somewhere?):
```python
response = FileResponse(filename, content_type=...)
```
This sets the header according to the code above.

I replaced this everywhere I found the word "Content-Disposition".

evap/results/templates/results_evaluation_detail.html
```diff
@@ -73,12 +73,12 @@ <h3>{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</h3>
...
                     {% trans 'Export text answers' %}
                 </a>
             {% endif %}
             {% if evaluation.course.grade_documents.count == 1 and can_download_grades %}
-                <a class="btn btn-sm btn-light d-print-none" href="{% url 'grades:download_grades' evaluation.course.grade_documents.first.id %}">{{ evaluation.course.grade_documents.first.description }}</a>
+                <a class="btn btn-sm btn-light d-print-none ml-2" href="{% url 'grades:download_grades' evaluation.course.grade_documents.first.id %}">{{ evaluation.course.grade_documents.first.description }}</a>
             {% elif evaluation.course.grade_documents.count > 1 and can_download_grades %}
                 <div class="btn-group d-print-none" role="group">
                     <button type="button" id="btnDownload" role="button" class="btn btn-sm btn-light dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">{% trans 'Grades' %}</button>
```
janno42: you should add the margin for this element as well in case there are multiple documents available

niklasmohrin: Right, I did not read this properly.

niklasmohrin: `57fa7f0983fe9ff31253aab286fdc6d3a25d5dfc` onward:

I added a class `InputData` to the exporter that makes the exporting a bit more readable, since it got really really awful after fixing the first error.
I had to handcraft an instance of `InputData` in the tests (there is a comment that explains why), should I add a test for the `InputData.__init__` method?
Or maybe this is not the optimal approach anyway and I should change it again? Let me know what you think?

Again, I can squash all these fixups after we settled everything.

evap/results/tests/test_exporters.py
```diff
@@ -254,3 +255,39 @@ def test_contributor_result_export(self):
...
+
+        questions = [baker.make(Question, questionnaire__type=t) for t in [10, 20, 30]]
+        for idx in [0, 1, 2, 2, 0]:
+            baker.make(TextAnswer, question=questions[idx], contribution__evaluation=evaluation)
+
+        # The InputData object needs to be hand crafted here, because the
+        # items returned from collect_data(evaluation).contribution_results
+        # did not link back to the questions.
+        # This is probably because baker.make does not fully cover the steps
+        # taken when a user submits something on the website.
```
He3lixxx, line 270: The implementation of `collect_results` iterates like this:
```python
for contribution in evaluation.contributions.all().prefetch_related("questionnaires", "questionnaires__questions"):
    for questionnaire in contribution.questionnaires.all():
        do_stuff™()
```

You use baker to make TextAnswer objects. baker will create a Contribution for each of these. However, you only specified that these newly created contributions should have `evaluation=evaluation`, so their `questionnaires` Many-To-Many-Field will still be empty.

So, in order to fix this, simply change
```python
baker.make(TextAnswer, question=questions[idx], contribution__evaluation=evaluation)
```
to
```python
baker.make(TextAnswer, question=questions[idx], contribution__evaluation=evaluation,
           contribution__questionnaires=[questions[idx].questionnaire])
```

Then, `collect_results(evaluation).contribution_results[0].questionnaire_results` is not empty anymore.

He3lixxx:
(Rest of the code looks good to me as it currently is - have not tested it yet)

evap/results/templates/results_evaluation_detail.html
```diff
@@ -72,10 +72,15 @@ <h3>{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</h3>
     <div class="card card-outline-primary mb-3">
         <div class="card-header d-flex">
             <div class="mr-auto">{% trans 'Overview' %}</div>
+            {% if view == "export" or view == "full" %}
+                <a class="btn btn-sm btn-light d-print-none" href="{% url 'results:evaluation_text_answers_export' evaluation.id %}?view={{ view }}{% if contributor_id is not None %}&contributor_id={{ contributor_id }}{% endif %}" type="button">
+                    {% trans 'Export text answers' %}
+                </a>
+            {% endif %}
```
pixunil, line 79: This button is also visible for students (you can confirm it by logging in as “student” with password “evap”). At least they get a 403. Could you change the conditions for the button here?

evap/results/views.py
```diff
@@ -229,9 +213,11 @@ def evaluation_detail(request, semester_id, evaluation_id):
         is_contributor=evaluation.is_user_contributor(view_as_user),
         is_responsible_or_contributor_or_delegate=evaluation.is_user_responsible_or_contributor_or_delegate(view_as_user),
         can_download_grades=view_as_user.can_download_grades,
+        can_export_text_answers=(not view_as_user.is_student and view in ("export", "full")),
         view=view,
         view_as_user=view_as_user,
         contributors_with_omitted_results=contributors_with_omitted_results,
+        contributor_id=contributor_id
```
janno42, line 220: please add a trailing `,`

evap/results/views.py
```diff
@@ -340,3 +331,65 @@ def add_warnings(evaluation, evaluation_result):
...
+    evaluation_result = collect_results(evaluation)
+    filter_text_answers(evaluation_result)
+    remove_textanswers_that_the_user_must_not_see(evaluation_result, view_as_user, represented_users, view)
+
+    results = TextAnswerExcelExporter.InputData(evaluation_result.contribution_results)
+
+    return results, contributor_id
+
+
+@reviewer_required
```
janno42, line 377: please remove this, we decided the correct test (`reviewer_or_responsible_or_contributor_or_delegate_required`) which does not exist yet might be too much.

evap/results/views.py
```diff
@@ -229,9 +213,11 @@ def evaluation_detail(request, semester_id, evaluation_id):
         is_contributor=evaluation.is_user_contributor(view_as_user),
         is_responsible_or_contributor_or_delegate=evaluation.is_user_responsible_or_contributor_or_delegate(view_as_user),
         can_download_grades=view_as_user.can_download_grades,
+        can_export_text_answers=(not view_as_user.is_student and view in ("export", "full")),
```
janno42, line 216: instead of using `is_student` (which is true for any user who participated in any course - we can't assume students and contributors are disjoint) this should probably be something like `is_user_responsible_or_contributor_or_delegate` for this evaluation (see method in `Evaluation`).

PR #1423, opened 2020-02-03
Get some order_bys straight for django 3.1

### LINKED ISSUE #1359: Prepare Django 3.1 Update: group_by queries using Meta.ordering
In django 3.1, all `GROUP BY` queries will not use `Meta.ordering` anymore in the returned result.

See https://docs.djangoproject.com/en/2.2/topics/db/aggregation/#interaction-with-default-ordering-or-order-by

Thus, running the tests with `-Wa` currently gives a deprecation warning for each group_by query whose behaviour will change.

In most cases, we think that it will not require us to change any of these queries (and hence, no action needs to be taken) - still, it needs to be checked.

karyon:
fixes #1359 

The first commit isn't really necessary, it's just so we're warning free and it can be removed when we have 3.1. Or i remove it again and the warning stays until 3.1.

the last commit is for you @janno42


evap/student/views.py
```diff
@@ -41,7 +41,7 @@ def index(request):
         evaluation.participates_in = request.user in evaluation.participants.all()
         evaluation.voted_for = request.user in evaluation.voters.all()
     evaluations = get_evaluations_with_course_result_attributes(evaluations)
-    evaluations.sort(key=lambda evaluation: evaluation.full_name)  # evaluations must be sorted for regrouping them in the template
+    evaluations.sort(key=lambda evaluation: (evaluation.course.name, evaluation.name))  # evaluations must be sorted for regrouping them in the template
```
He3lixxx: why do we not want to use `full_name` here?

karyon: it has the exact same result in the end. however, the regroup in the template requires the evaluations to be sorted by course, and i thought this would make it more explicit, instead of using the human-readable name for a technical requirement that just happens to work. I don't have strong feelings here though :)

evap/results/views.py
```diff
@@ -94,9 +94,9 @@ def update_template_cache_of_published_evaluations_in_course(course):
 
 def get_evaluations_with_prefetched_data(evaluations):
     if isinstance(evaluations, QuerySet):
-        participant_counts = evaluations.annotate(num_participants=Count("participants")).values_list("num_participants", flat=True)
-        voter_counts = evaluations.annotate(num_voters=Count("voters")).values_list("num_voters", flat=True)
-        course_evaluations_counts = evaluations.annotate(num_course_evaluations=Count("course__evaluations")).values_list("num_course_evaluations", flat=True)
+        participant_counts = evaluations.annotate(num_participants=Count("participants")).order_by('pk').values_list("num_participants", flat=True)
+        voter_counts = evaluations.annotate(num_voters=Count("voters")).order_by('pk').values_list("num_voters", flat=True)
+        course_evaluations_counts = evaluations.annotate(num_course_evaluations=Count("course__evaluations")).order_by('pk').values_list("num_course_evaluations", flat=True)
```
pixunil: I wonder whether these annotations could be grouped into one. When I tried it locally, I discovered that [one must use `Count(…, distinct=True)` when annotating with multiple counts](https://docs.djangoproject.com/en/3.0/topics/db/aggregation/#combining-multiple-aggregations).
I could also imagine that the the later `for in zip()` loop could be expressed with a single annotation with a `Coalesce` for participants and voters.
But these are just wild thoughts with unprofiled guesses :D

evap/evaluation/models.py
```diff
@@ -467,7 +467,7 @@ def general_contribution_has_questionnaires(self):
 
     @property
     def all_contributions_have_questionnaires(self):
-        return self.general_contribution and (all(self.contributions.annotate(Count('questionnaires')).values_list("questionnaires__count", flat=True)))
+        return self.general_contribution and (all(self.contributions.annotate(Count('questionnaires')).order_by().values_list("questionnaires__count", flat=True))) # the order_by silences a warning and can be removed with django 3.1
```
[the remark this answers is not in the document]
karyon: nice, thanks! i used `not ....exists()`.

evap/results/views.py
karyon: it seemed to work! unfortunately, it was about 2x slower than doing individual queries...

pixunil: Ah, bummer. Properly it is because Django uses joins instead of subqueries which the database might not be able to optimize correctly…

evap/grades/views.py
```diff
@@ -95,6 +95,7 @@ def upload_grades(request, semester_id, course_id):
 
     if form.is_valid():
         form.save(modifying_user=request.user)
+        # specific ordering?
```
janno42: not important, can be unordered

evap/grades/views.py
```diff
@@ -127,6 +128,7 @@ def toggle_no_grades(request):
 
     course.gets_no_grade_documents = not course.gets_no_grade_documents
     course.save()
+    # specific ordering?
```
janno42: ordering would only be shown in the emails, we have another pr working on this (will probably done in the templates).

evap/staff/templates/staff_course_form.html
```diff
@@ -24,6 +24,7 @@ <h3>
                     <h5 class="card-title">{% trans 'Evaluations' %}</h5>
                     {% if course.evaluations.count > 0 %}
                         <ul>
+                            {# specific ordering? #}
```
janno42: alphabetically by evaluation name, please

evap/staff/views.py
```diff
@@ -325,7 +325,7 @@ def semester_evaluation_operation(request, semester_id):
         raise SuspiciousOperation("Unknown target state: " + str(target_state))
 
     evaluation_ids = (request.GET if request.method == 'GET' else request.POST).getlist('evaluation')
-    evaluations = annotate_evaluations_with_grade_document_counts(Evaluation.objects.filter(id__in=evaluation_ids))
+    evaluations = annotate_evaluations_with_grade_document_counts(Evaluation.objects.filter(id__in=evaluation_ids)) # I guess we want some specific ordering here...?
```
janno42: alphabetically by evaluation name, please

evap/staff/views.py
```diff
@@ -616,6 +616,7 @@ def send_reminder(request, semester_id, responsible_id):
 
     form = RemindResponsibleForm(request.POST or None, responsible=responsible)
 
+    # i guess we want some specific ordering here...?
```
janno42: only shown in emails, so doesn't need to be ordered here

PR #1424, opened 2020-02-03
Logging changes for Courses and Evaluations

### LINKED ISSUE #1401: Logging changes for Courses and Evaluations
When making changes in Courses and Evaluations, these should be logged. This should be the case for all users who can make changes (editors and managers) and should include status changes as well as changed fields with timestamp and user.
The log should be accessible on the manager's edit page.
See also #1348.

felixrindt:
For now a draft to fix #1401. Don't need a full review yet ;)


felixrindt:
image

Logging for m2m attributes is possible :)

felixrindt:
image

now supports simple attributes, many-to-one and many-to-many. Next working on one-to-one and then one-to-many.

felixrindt:
image
I was able to hack together basic one_to_many logs, but thats really WIP ^^

felixrindt:
I worked on deletion today:
image

Deletion was hard because we can't reference the object that was deleted, because it got deleted. But because we should have information about the type to display it, a workaround was to safe the content type object id of the object. At the moment, I'm just popping it of the field_data when generating the log entry message.

I think I now have every possible object modifications covered. I should work on

* getting the user object on more log entries, without making it too ugly I hope...
* Improving the look, esp. by implementing proper `__str__` and `__repr__` for our models and model instances. That should make it look much more finished. Also every model itself could define black/whitelists for fields to track.

felixrindt:
I actually got it working in a sensible manner now :)

felixrindt:
image

evap/evaluation/models.py
```diff
@@ -812,6 +1083,15 @@ class Meta:
     def is_general(self):
         return self.contributor_id is None
 
+    @property
+    def object_to_attach_logentries_to(self):
+        return Evaluation, self.evaluation_id
+
+    def __str__(self):
+        if self.contributor:
+            return _("Contribution by") + " " + self.contributor.full_name
```
He3lixxx, line 1092: Using one f-string here would allow localization to also change word order (which we usually want to allow?)

evap/evaluation/models.py
```diff
@@ -812,6 +1083,15 @@ class Meta:
...
+        return str(_("General Contribution"))
```
He3lixxx, line 1093: Why the str()?

evap/evaluation/templates/log/changed_fields_entry.html
```diff
@@ -0,0 +1,35 @@
...
+
+<ul>
+    {% for field_name, actions in fields.items %}
+        {% for action in actions %}
+            <li>
+                {% if action.type == "add" %}
+                    {{ action.label}} {% trans "added" %}: {{ action.items|join:", " }}
+                {% elif action.type == "remove" %}
+                    {{ action.label}} {% trans "removed" %}: {{ action.items|join:", " }}
+                {% elif action.type == "clear" %}
```
He3lixxx, line 14: This mode is not in evaluation/models.py:204 and I didn't read anything about a "clear" action anywhere so far... Did I miss something?

felixrindt: 'clear' is used when m2m sets are cleared. A m2m signal exists, but I have not seen it happen in evap yet.

evap/evaluation/models.py
```diff
@@ -1217,17 +1497,17 @@ def create_superuser(self, email, password, first_name=None, last_name=None):
             password=password,
             email=self.normalize_email(email),
             first_name=first_name,
-            last_name=last_name
+            last_name=last_name,
         )
         user.is_superuser = True
         user.save()
-        user.groups.add(Group.objects.get(name="Manager"))
+        user.groups.add(Group.objects.get(name='Manager'))
```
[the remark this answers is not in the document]
felixrindt: Happend while autoformatting. You want me to go back and revert it?

evap/evaluation/models.py
felixrindt: image

felixrindt: we're using gettext_lazy

evap/evaluation/tests/test_model_logging.py
```diff
@@ -0,0 +1,47 @@
...
+        self.assertEqual(
+            self.logentry._evaluation_log_template_context(json.loads(self.logentry.data))["vote_start_datetime"],
+            [
+                FieldAction(
+                    "Start of evaluation",
+                    "change",
+                    [log_serialize(self.old_start_date), log_serialize(self.new_start_date)],
+                )
+            ],
+        )
```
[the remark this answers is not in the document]
felixrindt: sure :+1:

evap/evaluation/models.py
```diff
@@ -1,34 +1,332 @@
...
+
+    def _change_data(self, action_type, include_none_values=False):
+        """
+        Return a dict mapping field names to changes that happened in this model instance,
+        depending on the action that is being done to the instance.
+        """
+        self_dict = self._as_dict()
+        if action_type == "create":
+            changes = {
+                field_name: {'change': [None, created_value]}
```
karyon, line 198: 'create'?

evap/evaluation/models.py
```diff
@@ -1,34 +1,332 @@
...
+            changes = {}
+            for field_name, old_value in old_dict.items():
+                if old_value is None and not include_none_values:
+                    continue
+                field = self._meta.get_field(field_name)
+                if field.many_to_many:
+                    action_items = [obj.pk for obj in old_value]
+                else:
+                    action_items = [old_value]
+                changes[field_name] = {'delete': action_items}
```
karyon, line 220: I don't quite understand what's going on, I'd appreciate a comment on why `delete` needs so much extra work.

evap/evaluation/models.py
```diff
@@ -1,34 +1,332 @@
...
+
+        changes.update(self._m2m_changes)
+        return changes
+
+    def update_log(self, mode):
+        action = {
+            'delete': 'delete',
+            'create': 'create',
+            'change': 'change',
+            'm2m': 'change',
```
karyon, line 232: right now there are these three sets of strings as far as i can see: delete|create|change|m2m (modes?), delete|create|change (actions?), and add|remove|clear (some other actions?). clearly distinguishing between these three sets with some prefixes would be nice, and using constants instead of magic strings too. enums might be useful here.

karyon (commented):
i know this comes rather late, sorry!

evap/evaluation/models.py
```diff
@@ -1,34 +1,332 @@
...
+        self._logentry = None
+        for field in type(self)._meta.many_to_many:
+            self.register_logged_m2m_field(field)
+
+    def register_logged_m2m_field(self, field):
+        through = getattr(type(self), field.name).through  # converting from field to its descriptor
+        m2m_changed.connect(
+            LoggedModel.m2m_changed,
+            sender=through,
+            dispatch_uid="m2m_log-{!r}-{!r}".format(type(self), field)
```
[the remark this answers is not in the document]
felixrindt: that worked out nicer, yes :)

evap/evaluation/models.py
```diff
@@ -1,34 +1,332 @@
...
+        self.all_logentries().delete()
+        super().delete(*args, **kw)
+
+    def all_logentries(self):
+        """
+        Return a queryset with all logentries that should be shown with this model.
+        """
+        return LogEntry.objects.filter(
+            attached_to_object_type=ContentType.objects.get_for_model(type(self)), attached_to_object_id=self.pk,
+        ).select_related("user")
```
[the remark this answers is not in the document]
felixrindt: the `select_related` or the whole query? I'd keep the query but would understand if you'd want to move the `select_related`.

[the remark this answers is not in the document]
karyon: the `select_related`.

evap/evaluation/models.py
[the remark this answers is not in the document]
karyon: i realise i broke the second link, it should be https://docs.djangoproject.com/en/3.1/topics/signals/#connecting-to-signals-sent-by-specific-senders. would that make your manual check unnecessary?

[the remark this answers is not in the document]
felixrindt: Well we actually need to connect to every m2m field of a model, not only the model itself. I don't see how that's possible in the base class using the sender kwarg on the decorator.

evap/evaluation/models.py
```diff
@@ -1,34 +1,332 @@
...
+            else:
+                time_matches = abs(group[0].datetime - entry.datetime) < timedelta(seconds=10)
+                if entry.request_id is None and group[0].request_id is None and time_matches:
+                    group.append(entry)
+                else:
+                    groups.append(group)
+                    group = [entry]
+
+        if group:
+            groups.append(group)
```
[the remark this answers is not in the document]
felixrindt: I opted for dropping the time-based grouping, as that is kind of hacky anyway. 
https://docs.python.org/3/library/itertools.html#itertools.groupby to the rescue

evap/evaluation/models.py
```diff
@@ -1,34 +1,332 @@
...
+            'message': message,
+            'fields': self._evaluation_log_template_context(field_data),
+        })
+
+
+def log_serialize(obj):
+    if obj is None:
+        return ""
+    if isinstance(obj, (date, time, datetime)):
+        return localize(obj)
```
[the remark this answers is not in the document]
felixrindt: JSON doesn't have a datetime datatype, so it would store a datetime as an ISO-string. Now instead of building special serialization logic for that, I just store it as a appropriately formatted string here already. Afaik, we use the same formats for german and english, so this should be fine.

evap/evaluation/models.py
```diff
@@ -1,34 +1,332 @@
...
+        ordering = ("-datetime", "-id")
+
+    @staticmethod
+    def _pk_to_string_representation(key, field, related_objects):
+        if key is None:
+            return key
+        try:
+            return str(related_objects.get(pk=key))
+        except field.related_model.DoesNotExist:
+            return "�"
```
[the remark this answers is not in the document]
felixrindt: If the m2m set changes on an object, we store the primary keys of the objects added/deleted. For display, we fetch those objects and build a string representation from them. We can't do that if those objects got deleted, but we might still want to show that *an* object as deleted. I figured this symbol could represent an "unknown" object. 

With evap, I think this will rarely be seen, as we rarely delete objects used on m2m fields with Evaluation and Course (that's mostly Questionaires and users), but I still included this as a sensible last resort.

evap/evaluation/models.py
felixrindt: I created enums, but I'm unsure about how to access them in the template. Using a context processor providing the two enums looks like a valid option. Do you have another idea?

evap/evaluation/tests/test_models_logging.py
```diff
@@ -0,0 +1,88 @@
...
+from evap.evaluation.models_logging import FieldAction
+
+
+class TestLoggedModel(TestCase):
+    def setUp(self):
+        self.old_start_date = datetime.now()
+        self.new_start_date = self.old_start_date + timedelta(days=10)
+
+        self.course = baker.make(Course)
+        self.course.save()
```
He3lixxx, line 17: I think you don't need that, baker.make will create the object and store the row in the database (happens more often)

evap/evaluation/models_logging.py
```diff
@@ -0,0 +1,290 @@
...
+            message = _("A {cls} was deleted.")
+
+        message = message.format(
+            cls=self.content_type.model_class()._meta.verbose_name_raw,
+            obj=f"\"{self.content_object!s}\"" if self.content_object else "",
+        )
+
+        return render_to_string("log/changed_fields_entry.html", {
+            'message': message,
+            'fields': self._evaluation_log_template_context(self.data),
```
karyon, line 125: a method that's only called from a template and returns a rendered template smells to me like it should be template include. `message` and `fields` could be properties on `LogEntry` (perhaps with a more descriptive name for `fields`)

evap/evaluation/tests/test_models_logging.py
```diff
@@ -0,0 +1,88 @@
...
+            vote_end_date=date.today() + timedelta(days=20),
+        )
+        self.evaluation.save()  # first logentry
+
+        self.evaluation.vote_start_datetime = self.new_start_date
+        self.evaluation.save()  # second logentry
+
+        self.logentry = self.evaluation.related_logentries()[1]
+
+    def test_voters_not_in_evluation_data(self):
```
He3lixxx, line 33: evluation!

He3lixxx (commented):
This feels quite sophisticated but I'm afraid that's the price we have to pay...

evap/evaluation/models_logging.py
```diff
@@ -0,0 +1,290 @@
...
+@receiver(m2m_changed)
+def _m2m_changed(sender, instance, action, reverse, model, pk_set, **kwargs):  # pylint: disable=unused-argument
+    if reverse:
+        return
+    if not isinstance(instance, LoggedModel):
+        return
+
+    field_name = next((field.name for field in type(instance)._meta.many_to_many
+                       if getattr(type(instance), field.name).through == sender), None)
+    if field_name is None:
```
karyon, line 279: comment on why this might happen?

evap/evaluation/models_logging.py
```diff
@@ -0,0 +1,290 @@
...
+        label = getattr(field, "verbose_name", field_name).capitalize()
+    except FieldDoesNotExist:
+        label = field_name.capitalize()
+
+    for field_action_type, items in actions.items():
+        if field.many_to_many or field.many_to_one or field.one_to_one:
+            # convert item values from primary keys to string-representation for relation-based fields
+            related_objects = field.related_model.objects.filter(pk__in=items)
+            missing = len(items) - related_objects.count()
+            items = [str(obj) for obj in related_objects] + ["�"] * missing
```
karyon, line 68: how about <deleted object> instead of �? then you don't need to read the ~~manual~~code to know what that symbol means.

evap/evaluation/models_logging.py
```diff
@@ -0,0 +1,290 @@
...
+from django.db import models
+from django.db.models.signals import m2m_changed
+from django.dispatch import receiver
+from django.forms.models import model_to_dict
+from django.template.defaultfilters import yesno
+from django.template.loader import render_to_string
+from django.utils.formats import localize
+from django.utils.translation import gettext_lazy as _
+
+logger = logging.getLogger(__name__)
```
karyon, line 23: unused

evap/evaluation/models_logging.py
felixrindt: what instead?

evap/evaluation/models_logging.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+        Return a model class and primary key for the object for which this logentry should be shown.
+        By default, show it to the object described by the logentry itself.
+        """
+        return type(self), self.pk
+
+    @property
+    def unlogged_fields(self):
+        """Specify a list of field names so that these fields don't get logged."""
+        return ['id', 'last_modified_time', 'last_modified_user', 'order']
```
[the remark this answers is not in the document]
felixrindt: If upgrading between seasons, you likely don't need last_modified. I'd opt for removing it.

evap/evaluation/models_logging.py
```diff
@@ -0,0 +1,290 @@
...
+        if not self._logentry:
+            try:
+                user = self.thread.request.user
+                request_id = self.thread.request_id
+            except AttributeError:
+                user = None
+                request_id = None
+            attach_to_model, attached_to_object_id = self.object_to_attach_logentries_to
+            attached_to_object_type = ContentType.objects.get_for_model(attach_to_model)
+            self._logentry = LogEntry(
```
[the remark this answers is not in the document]
felixrindt: It would mean less logentries in rare cases I guess?

evap/evaluation/models_logging.py
```diff
@@ -0,0 +1,290 @@
...
+        return LogEntry.objects.filter(
+            attached_to_object_type=ContentType.objects.get_for_model(type(self)), attached_to_object_id=self.pk,
+        )
+
+    def grouped_logentries(self):
+        """
+        Returns a list of lists of logentries for display. The order is not changed.
+        Logentries are grouped if they have a matching request_id.
+        """
+        yield from (list(group) for key, group in itertools.groupby(
```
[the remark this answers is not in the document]
felixrindt: yielding is lazy, so one could islice or whatever.
but it could be a list as well.

evap/evaluation/models_logging.py
```diff
@@ -0,0 +1,290 @@
...
+    def _evaluation_log_template_context(self, data):
+        model = self.content_type.model_class()
+        return {
+            field_name: list(_field_actions_for_field(model, field_name, actions))
+            for field_name, actions in data.items()
+        }
+
+    def display(self):
+        if self.action_type == InstanceActionType.CHANGE:
+            if self.content_object:
```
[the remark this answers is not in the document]
felixrindt: sure :) 
Content object might have been deleted in the meantime.

evap/evaluation/models_logging.py
```diff
@@ -0,0 +1,290 @@
...
+def _choice_to_display(field, choice):  # does not support nested choices
+    return next(filter(lambda t: t[0] == choice, field.choices), (choice, choice))[1]
+
+
+def _field_actions_for_field(model, field_name, actions):
+    field = model._meta.get_field(field_name)
+    try:
+        label = getattr(field, "verbose_name", field_name).capitalize()
+    except FieldDoesNotExist:
+        label = field_name.capitalize()
```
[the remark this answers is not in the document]
felixrindt: good catch. don't even need that because of the default used with getattr.

evap/evaluation/models_logging.py
felixrindt: You are right indeed. I changed that.

evap/evaluation/models_logging.py
karyon: nice :D i meant `<deleted object>`, but github swallowed that.

felixrindt: Hmm would we want extra css for that?

karyon: if you ask me, no, in my opinion the `<>` are visual indicator enough.

evap/evaluation/models_logging.py
```diff
@@ -0,0 +1,286 @@
...
+            m2m_field_names = [field.name for field in type(self)._meta.many_to_many if field.name not in self.unlogged_fields]
+            for field_name, related_objects in model_to_dict(self, m2m_field_names).items():
+                changes[field_name] = {FieldActionType.INSTANCE_DELETE: [obj.pk for obj in related_objects]}
+        else:
+            raise ValueError("Unknown action type: '{}'".format(action_type))
+
+        return changes
+
+    def _update_log(self, action_type: InstanceActionType, m2m_changes=None):
+        changes = self._get_change_data(action_type)
```
karyon, line 179: it's still weird that when logging m2m changes, the code first starts comparing the instance with itself. can you find a nice way around that?

my suggestion would be to have

`def log_m2m_change(field_name, action_type, pk_set)`
`def log_instance_create/change/delete()`

and all of those call  `def _update_log(action_type, changes)`

evap/evaluation/models_logging.py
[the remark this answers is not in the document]
janno42: yes, please remove them

karyon (approved):

He3lixxx (approved):

janno42:
what's still missing (i'll work on that):
- [ ] handle merging users for which log entries exist (currently this fails because they are referenced through a protected foreign key)
- [ ] add some logs to the test data (we should have some log data within our test data for courses/evaluations in the active semester)

evap/grades/forms.py
```diff
@@ -10,7 +10,6 @@ class GradeDocumentFileWidget(forms.widgets.ClearableFileInput):
 
 
 class GradeDocumentForm(forms.ModelForm):
-    # see EvaluationForm (staff/forms.py) for details, why the following two fields are needed
     last_modified_time_2 = forms.DateTimeField(label=_("Last modified"), required=False, localize=True, disabled=True)
     last_modified_user_2 = forms.CharField(label=_("Last modified by"), required=False, disabled=True)
```
karyon, line 14: should these also replaced with a LoggedModel in the long run?

evap/staff/views.py
```diff
@@ -717,7 +717,6 @@ def course_create(request, semester_id):
             raise SuspiciousOperation("Invalid POST operation")
 
         course = course_form.save()
-        course.set_last_modified(request.user)
         course.save()
```
karyon, line 720: form.save already saves the model, so the model.save shouldn't be necessary anymore (there's quite a few instances of that pattern)

evap/grades/forms.py
janno42: we could do that, but there's no real need. actually this information is probably never used at all.
i'm fine with both keeping it and changing it. what would you prefer?

karyon: if it's not used lets not spend time on it.

karyon (approved):

janno42 (approved):

janno42:
@felixrindt thank you so much for this incredibly cool new feature! and of course thank you @karyon and @He3lixxx for your contributions as well!

PR #1429, opened 2020-02-10
Ui changes

janno42:
see individual commit messages

### DIFF
--- a/evap/evaluation/templatetags/navbar_templatetags.py
+++ b/evap/evaluation/templatetags/navbar_templatetags.py
@@ -1,5 +1,5 @@
 from django.template import Library
-from evap.settings import LANGUAGES
+from evap.settings import DEBUG, LANGUAGES
 from evap.evaluation.models import Semester
 
 register = Library()
@@ -14,4 +14,5 @@ def include_navbar(user, language):
         'published_result_semesters': Semester.get_all_with_published_unarchived_results(),
         'result_semesters': Semester.get_all_with_unarchived_results(),
         'grade_document_semesters': Semester.objects.filter(grade_documents_are_deleted=False),
+        'debug': DEBUG,
     }
--- a/evap/results/templatetags/results_templatetags.py
+++ b/evap/results/templatetags/results_templatetags.py
@@ -19,3 +19,8 @@ def evaluation_results_cache_timeout(evaluation):
     if evaluation.state == 'published':
         return None  # cache forever
     return 0  # don't cache at all
+
+
+@register.filter(name='participationclass')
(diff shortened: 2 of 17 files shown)


evap/results/templatetags/results_templatetags.py
```diff
@@ -19,3 +19,8 @@ def evaluation_results_cache_timeout(evaluation):
     if evaluation.state == 'published':
         return None  # cache forever
     return 0  # don't cache at all
+
+
+@register.filter(name='participationclass')
+def participationclass(number_of_voters, number_of_participants):
+    return round((number_of_voters/float(number_of_participants))*10)
```
He3lixxx: why the float? Python3 int division should return floats.

He3lixxx:
(also I like the "uiii" branch name - will you count it up in roman numerals, so we arrive at "uv" in 2 PRs?)

janno42:
wait for "ux" ;)

evap/results/templatetags/results_templatetags.py
janno42: old habit

evap/static/scss/components/_nav.scss
```diff
@@ -9,6 +9,16 @@
     line-height: 42px;
 }
 
+.bg-debug {
```
pixunil: `.vote-bg-*` live in `_utilities.scss`. I have no hard opinion here, but this may be the better place for it

evap/evaluation/templatetags/navbar_templatetags.py
```diff
@@ -14,4 +14,5 @@ def include_navbar(user, language):
         'published_result_semesters': Semester.get_all_with_published_unarchived_results(),
         'result_semesters': Semester.get_all_with_unarchived_results(),
         'grade_document_semesters': Semester.objects.filter(grade_documents_are_deleted=False),
+        'DEBUG': DEBUG,
```
pixunil: Hm, I’d prefer to keep template variables in lowercase; `languages` is also transformed even though it originates from settings

evap/static/scss/components/_nav.scss
janno42: really? i thought this is clearly part of the navbar.

pixunil: As I said, no hard opinions here. As the class _could_ be applied to other elements, it would be more a background utility, but otherwise it is fine for me to leave it in `_nav.scss`

janno42: keeping possible reuse in mind i also like having it in utilities

PR #1431, opened 2020-02-11
Fix disabled error codes being in the wrong parent in .prospector.yaml

He3lixxx:
In #1413 I put the new exceptions in the wrong linting tool parent, so these errors were still shown. 

Funnily enough, the very very great tool codacy doesn't even show you what linting tool inside prospector caused these, it just gives you a "Prospector_F999" (don't know whether this is their fault or an error in prospector).

On the [interblag](https://xkcd.com/181/), [some people put F999 inside the pyflakes section](https://github.com/readthedocs/pydoc.io/blob/master/prospector.yml). Let's see if this helps...

(Or maybe I'm just stupid - can you find out which tool causes this error?)
image

Related to #1092


He3lixxx:
Oh noes, looks like god's wrath is upon us, for I have sinned and accused the greatest static coda analysis platform of all time of being faulty.

(Codacy analysis wasn't done for this and #1430)

PR #1433, opened 2020-02-24
Vagrant stuff: Move evap repo in VM to /opt/evap. Fix autocompletion.

### LINKED ISSUE #1240: Importing a backup made by update_production.sh does not work flawlessly.

He3lixxx:
fixes #1240


He3lixxx:
> Have you considered making all the paths relative to the script location instead of hardcoding /opt/evap?

Seems to be not that easy :D https://stackoverflow.com/questions/59895/how-to-get-the-source-directory-of-a-bash-script-from-within-the-script-itself

The intention for giving full paths was to make sure that the script works no matter what directory you call it from (especially to allow you to go the location where you want the backup to be stored and call it from there). I'd totally agree to replace the variable used to build with full path with
```bash
REPO_DIR=get_current_file_location() + "../"
```
or whatever, but bash is bash.

PR #1435, opened 2020-03-09
UI changes

janno42:
UI improvements and a new feature (see commit messages)


evap/results/templates/distribution_widget.html
```diff
@@ -3,7 +3,9 @@
 
 {% spaceless %}
 {% if average is not None %}
-    <div class="d-flex">
+    <div class="d-flex" data-toggle="tooltip" data-placement="left" data-fallback-placement='["left", "bottom"]'
+        title="{% if weight_sum and evaluation.course.evaluation_count > 1 %}{% blocktrans with percentage=evaluation.weight|percentage:weight_sum %}This evaluation contributes {{ percentage }} to the final grade of the course.{% endblocktrans %}{% endif %}"
```
pixunil: This title shows also on individual questions.
Also (little nit): When the tooltip breaks to its bottom position, it looks weirdly positioned:
Screenshot from 2020-03-09 20-18-59

evap/context_processors.py
```diff
@@ -8,3 +8,7 @@ def slogan(request):
     if get_language() == "de":
         return {'SLOGAN': random.choice(settings.SLOGANS_DE)}  # nosec
     return {'SLOGAN': random.choice(settings.SLOGANS_EN)}  # nosec
+
+
+def debug(request):
+    return {'DEBUG': settings.DEBUG}
```
pixunil: If we lowercased `SLOGAN`, we could also lowercase `DEBUG` :)
Also, would it be possible to use this context processor in the nav bar for the striped background?

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -39,15 +39,21 @@
...
+    <td class="icon" data-col="state-textanswers" data-order="3" data-filter="no_textanswers">
         <span class="fas fa-comment icon-gray" data-toggle="tooltip" data-placement="top" title="{% trans 'No text answers available' %}"></span>
     </td>
 {% elif evaluation.num_textanswers > evaluation.num_reviewed_textanswers %}
-    <td class="icon" data-col="state-textanswers" data-order="1" data-filter="unreviewed_textanswers">
-        <span class="fas fa-comment icon-yellow" data-toggle="tooltip" data-placement="top" title="{% trans 'Text answers awaiting review' %}"></span>
-    </td>
+    {% if evaluation.course.final_grade_documents or evaluation.course.gets_no_grade_documents or not evaluation.course.is_graded %}
+        <td class="icon" data-col="state-textanswers" data-order="1" data-filter="unreviewed_textanswers_urgent">
+            <span class="fas fa-comment icon-red" data-toggle="tooltip" data-placement="top" title="{% trans 'Text answers awaiting urgent review' %}"></span>
```
pixunil: What about “Text answers awaiting urgent review as grading process was already completed”?
(Similar for the urgent publication state)

evap/results/templates/distribution_widget.html
janno42: i won't address the tooltip positioning (here)

evap/staff/templates/staff_semester_view_evaluation.html
janno42: unfortunately, the case that the course is ungraded would still be missing in this message. any further ideas?

pixunil: I am not quite satisfied with these, but maybe they give you some other inspirations:

“Text answers awaiting urgent review because necessary grade documents were uploaded”
“Results not yet published although all necessary grade documents were uploaded”

“Text answers awaiting urgent review because no missing grade document hinders publication”
“Results not yet published although no grade document is missing”

(but of course, if you think the string is fine, we can keep it)

evap/context_processors.py
janno42: it is possible to use the value from the context processor but we would have to forward it in the navbar inclusion, so i think there is not really an advantage to the current solution.

evap/staff/templates/staff_semester_view_evaluation.html
janno42: i will go with "Text answers awaiting urgent review because grading process is finished" as we can argue that if a course is ungraded, its grading process is immediately finished.

evap/results/templates/result_widget_tooltip.html
```diff
@@ -1,5 +1,10 @@
 {% load evaluation_filters %}
-
+{% if show_weight_information and evaluation.course.evaluation_weight_sum and evaluation.course.evaluation_count > 1 %}
```
He3lixxx: In `results_index_evaluation.html`, you include `result_widget.html` and pass `show_weight_information=True`. However, in `result_widget.html`, where `result_widget_tooltip.html` is included, the variable is not explicitly passed, but only captured implicitly. (In the end, reading this if statement made me think: Huh, where did `show_weight_information` come from?)

I guess we'd want to make it explicit?

evap/results/templates/results_index_evaluation.html
```diff
@@ -62,9 +62,9 @@
 
     <div data-col="result"{% if not is_subentry %} data-order="{% if evaluation.is_single_result %}{{ evaluation.single_result_rating_result.average }}{% else %}{{ evaluation.avg_grade|default:7 }}{% endif %}"{% endif %}>
         {% if evaluation.is_single_result %}
-            {% include 'result_widget.html' with question_result=evaluation.single_result_rating_result single_result=True weight=evaluation.weight weight_sum=evaluation.course.evaluation_weight_sum %}
+            {% include 'result_widget.html' with question_result=evaluation.single_result_rating_result single_result=True weight=evaluation.weight weight_sum=evaluation.course.evaluation_weight_sum evaluation=evaluation show_weight_information=True %}
```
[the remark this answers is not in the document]
He3lixxx: I agree about the passed variables
```
weight=evaluation.weight weight_sum=evaluation.course.evaluation_weight_sum evaluation=evaluation
```
is somewhat weird. If we give the `evaluation` instance, there is no need to pass the other two.

janno42:
okay, it's getting way too messy. the tooltip will be removed from this pr and has to be refactored completely in #1436.

PR #1437, opened 2020-03-14
Weighting info tooltip

### LINKED ISSUE #1436: Evaluation weight info tooltip
The percentage information of the weight of an evaluation for the course's total grade needs an explanatory tooltip.
See https://github.com/janno42/EvaP/tree/weighting_info for a commit implementing this and #1435 for some notes on required refactoring.

pixunil:
Includes a small refactoring of the templates.
Screenshot from 2020-03-14 13-07-50
Screenshot from 2020-03-14 13-08-16

Remember to run `manage.py refresh_results_cache` to see the changes in action.
Resolves #1436


evap/results/templates/result_widget_tooltip.html
```diff
@@ -1,13 +1,29 @@
...
-        {{ name }}: {{ count }}/{{ question_result.count_sum }} ({{ count|percentage_one_decimal:question_result.count_sum }})
-        {% if not forloop.last %}</br>{% endif %}
-    {% endfor %}
-{% endwith %}
+{% if weight_info %}
+    <p>
+        {# QUESTION: Do we want to call a single result “evaluation” (currently) or “singel result” here? #}
+        {% blocktrans with percentage=weight_info %}
+          This evaluation contributes {{ percentage }} to the final grade of the course.
+        {% endblocktrans %}
```
pixunil: Do we want to call a single result “evaluation” (currently) or “single result” here?
I am in favor with calling everything consistently, but here I was thinking whether we should include more information here. On the index pages, I can go to the details page if I click on the result of a “full” evaluation – whereas single results are unclickable. May be confusing for some users or maybe I am overthinking here again.

evap/results/templates/distribution_widget.html
```diff
@@ -2,13 +2,16 @@
...
+    <div class="d-flex"
+        data-toggle="tooltip" data-placement="left" data-fallback-placement='["left", "bottom"]'
+        title="{% spaceless %}{% include 'result_widget_tooltip.html' with question_result=course_or_evaluation.single_result_rating_result weight_info=course_or_evaluation|weight_info %}{% endspaceless %}"
+    >
+        {% include "distribution_with_grade.html" with distribution=course_or_evaluation.distribution average=course_or_evaluation.avg_grade weight_info=course_or_evaluation|weight_info %}
     </div>
 {% else %}
     <div class="d-flex"{% if not question_result %} data-toggle="tooltip" data-placement="left" title="{% trans 'Not enough answers were given.' %}"{% endif %}>
-        {% include "distribution_with_grade_disabled.html" with icon="far fa-eye-slash" %}
+        {% include "distribution_with_grade_disabled.html" with icon="far fa-eye-slash" %} {# QUESTION: Do we want weight_info here? #}
```
pixunil: I was not quite sure how to handle cases where some weighted evaluations are included in the result and some are excluded because answers are missing. In this case, I think the numbers are a bit wrong.
I remember that I discussed with @janno42 this problem for a minute, but I can’t remember possible solutions.

evap/results/templates/distribution_widget.html
```diff
@@ -2,13 +2,16 @@
 {% load evaluation_filters %}
 
 {% spaceless %}
-{% if average is not None %}
-    <div class="d-flex">
-        {% include "distribution_with_grade.html" with distribution=distribution average=average weight=weight weight_sum=weight_sum %}
+{% if course_or_evaluation.avg_grade is not None %}
+    <div class="d-flex"
+        data-toggle="tooltip" data-placement="left" data-fallback-placement='["left", "bottom"]'
+        title="{% spaceless %}{% include 'result_widget_tooltip.html' with question_result=course_or_evaluation.single_result_rating_result weight_info=course_or_evaluation|weight_info %}{% endspaceless %}"
```
pixunil: Any idea for a nicer solution than to put `{% spaceless %}` around the tooltip? The problem lies in courses or evaluations with no weighting information that the tooltip is considered empty, but the included file contains whitespaces (even if I wrap the entire file into `{% spaceless %}`, one line break after the closing tag persists).

evap/results/templates/results_index_evaluation.html
```diff
@@ -61,11 +61,7 @@
     {% endif %}
 
     <div data-col="result"{% if not is_subentry %} data-order="{% if evaluation.is_single_result %}{{ evaluation.single_result_rating_result.average }}{% else %}{{ evaluation.avg_grade|default:7 }}{% endif %}"{% endif %}>
-        {% if evaluation.is_single_result %}
-            {% include 'result_widget.html' with question_result=evaluation.single_result_rating_result single_result=True weight=evaluation.weight weight_sum=evaluation.course.evaluation_weight_sum %}
-        {% else %}
-            {% include 'distribution_widget.html' with distribution=evaluation.distribution average=evaluation.avg_grade weight=evaluation.weight weight_sum=evaluation.course.evaluation_weight_sum %}
-        {% endif %}
+        {% include 'distribution_widget.html' with course_or_evaluation=evaluation %}
```
pixunil: Commenting on an arbitrary line for some naming questions. As the widget templates now have some clearer use-cases, we could also rename those:

- `distribution_widget` → `evaluation_or_course_result_widget` or `result_widget_on_index`
- `result_widget` → `question_result_widget` or `result_widget_on_details`
- `result_widget_tooltip` also may need renaming

But none of the names I currently can think of are a great fit, do you have a better idea?

evap/results/templates/result_widget_tooltip.html
janno42: it can be called evaluation as it also is an evaluation and the information about being a single result is displayed in another place.

evap/results/templates/distribution_widget.html
janno42: we thought that just keeping the numbers although they don't add up to 100% is okay.

evap/results/templates/distribution_widget.html
janno42: i don't mind having `spaceless` here.

evap/results/templates/results_index_evaluation.html
janno42: i would go for `evaluation_result_widget` (which would also be used for course results, but having that in the names turns it into some very long name that doesn't really help if you ask me), `question_result_widget` and `result_widget_tooltip`.
is the tooltip specific to one of the two widgets? then the prefix can be added.

PR #1438, opened 2020-03-15
Remove default ordering by name_de. Fixes #1243

### LINKED ISSUE #1243: Don't sort by name_de by default

karyon:
For Semester and Questionnaire, i replaced it with `pk`. They are primarily ordered by other attributes and i figured if those are the same, any stable sorting would be enough.

For Course, I went through the places where the ordering is user-visible and added a manual sort() in the one place where I saw it was missing. In the EvaluationForm and SingleResultForm the course field is now completely unordered, but only when creating a new Evaluation/SingleResult, and I thought that would not be worth it.

(Yes, that's the lowest-prio refactoring issue that we have. Why do you ask?)

Fixes #1243


evap/evaluation/models.py
```diff
@@ -173,18 +173,18 @@ class Questionnaire(models.Model):
...
+        ordering = ('type', 'order', 'pk')
         verbose_name = _("questionnaire")
         verbose_name_plural = _("questionnaires")
 
     def __str__(self):
         return self.name
 
     def __lt__(self, other):
-        return (self.type, self.order, self.name_de) < (other.type, other.order, self.name_de)
+        return (self.type, self.order, self.pk) < (other.type, other.order, other.pk)
```
janno42: what did we do here? 🤦‍♂

karyon: 🤷‍♂ 

(no i definitely didn't git blame that)

PR #1440, opened 2020-03-24
Fix results preview for unpublished evaluations with rating answers

### LINKED ISSUE #1439: Crash when previewing results
Steps to reproduce:
1. Log in as evap
2. Select `Preview results` on "Subject Specific English 2" (breaks in German too)

Interestingly, I don't get any errors when viewing "Beautiful Data", which is also a preview, or "3d Computer Graphic - Software Architecture", which is public.

I am on `25401c018f8f304001d67dcd22f52861d01ddf3b` and reloaded the test data right before testing.

```
AttributeError at /results/semester/21/evaluation/347

'Course' object has no attribute 'evaluation_weight_sum'
```

Can anyone confirm this?

pixunil:
Fixes #1439

To verify that `test_preview_with_rating_answers` would catch possible regressions, run:
```sh
git checkout master
git checkout fix-preview-results evap/results/tests/test_views.py
./manage.py test
```
Only one of the new two tests is expected to fail. The other exists for symmetry, as a “greyed out” evaluation result won’t render weighting info, but this could change in the future.

I noticed that in production code, the return value of `get_evaluations_with_prefetched_data` is always passed to `get_evaluations_with_course_result_attributes`. So I could imagine that the former always calls the latter, as there exists afaik to render only a single evaluation without the knowledge of the extra course attributes.
Also (I couldn’t find it in the test data), could it be that the weighting info for previews was invalid now and before #1437? (course with multiple evaluations where the average grade is visible) Should we add this case to the test data then?


pixunil:
This was only a guess because I found it strange that the missing call hasn’t caused problems in other cases. And the case that I mentally constructed was a course which has multiple evaluations and one of them is not yet published, but has its results. In this case, I would expect that managers would also see the weighting info (as tooltip and inside the distribution bar), but I couldn’t explain to me how the template could access the necessary attributes from the course.
But as I said, I couldn’t find this setup in the test data, so I couldn’t check my theory. But I also think it is not practical to research the answer for it :)

PR #1441, opened 2020-03-30
Update to Django 3.0

### LINKED ISSUE #1387: Upgrade to django 3.0
https://docs.djangoproject.com/en/3.0/releases/3.0/

There is a guide for upgrading: https://docs.djangoproject.com/en/3.0/howto/upgrade-version/

Basically
* Read the release notes
* check dependencies for compatibility and update those
* run tests with `python -Wa` and solve deprecation warnings
  * Be aware of #1359
* put the new django into the requirements
* run tests, fix failures if any
* run tests with `python -Wa` and solve deprecation warnings again
* if there was any new feature in the release notes that might help us, use it
  * I spotted [Enumerations for model field choices](https://docs.djangoproject.com/en/3.0/releases/3.0/#enumerations-for-model-field-choices), anything else?

also, we need to check the installed python version on production, django 3.0 supports python 3.6 and newer. I think right now production runs python 3.5.

PFischbeck:
This fixes #1387.

I couldn't resist the urge to update the project to Django 3.0.
In particular:

- I updated all dependencies. Note that I replaced `django-sendfile` with `django-sendfile2`, because the former project is unmaintained and threw some warnings. The latter seems to be actively maintained.
- I updated Django to 3.0.
- I fixed the warnings, which mainly meant using `gettext` instead of `ugettext`.
- I made use of the new [enumerations for model field choices](https://docs.djangoproject.com/en/3.0/releases/3.0/#enumerations-for-model-field-choices) wherever possible. I quite like it, since the enums are now more descriptive, e.g., `TextAnswer.State.HIDDEN` instead of `TextAnswer.HIDDEN`.

### DIFF
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -5,7 +5,7 @@ from django import forms
 from django.conf import settings
 from django.db.models import Q
 from django.forms.widgets import CheckboxSelectMultiple
-from django.utils.translation import ugettext_lazy as _
+from django.utils.translation import gettext_lazy as _
 from evap.evaluation.forms import UserModelMultipleChoiceField, UserModelChoiceField
 from evap.evaluation.models import Course, Evaluation, Questionnaire, UserProfile
 from evap.evaluation.tools import date_to_datetime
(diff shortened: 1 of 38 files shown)


He3lixxx:
@janno42 do you want to review too or should I merge?

PR #1443, opened 2020-04-04
Import names for degrees and course types

### LINKED ISSUE #1388: Alternative names for degrees and course types on import
There should be fields on the staff pages for the definition of degrees and course types that allow to define a number of alternative names for each entry.
When importing the enrollment data, these are then also accepted by the importer and mapped accordingly. E.g., a course could then be defined as "B.Sc."/"L" in the import data and would be mapped to "Bachelor"/"Lecture".

pixunil:
image

Closes #1388

### Status

Import names are stored in an `ArrayField`. This has the downside that the uniqueness of import names can not be enforced with a database constraint, but is way more ergonomic than creating `DegreeImportName` models (caused a lot of duplication and confusion about the instances vs strings).

- The select2 handling is now getting a bit more messy than it was before as I now have a configuration that differs from the default configuration. May be we should use this as an opportunity to refactor this?
- There are some issues inherited from select2: You can not tab into the import names input and a import name not accepted by pressing <kbd>Enter</kbd> will get discarded when losing focus. But this is maybe something to fix for the whole site.


evap/staff/importers.py
```diff
@@ -243,11 +251,18 @@ def __init__(self):
...
+
+        self.degrees = {}
+        for degree_import_name in DegreeImportName.objects.all():
+            self.degrees[degree_import_name.name] = degree_import_name.degree
 
     def read_one_enrollment(self, data, sheet, row):
         student_data = UserData(first_name=data[2], last_name=data[1], email=data[3], title='', is_responsible=False)
         responsible_data = UserData(first_name=data[10], last_name=data[9], title=data[8], email=data[11], is_responsible=True)
-        evaluation_data = EvaluationData(name_de=data[6], name_en=data[7], type_name=data[4], is_graded=data[5], degree_names=data[0], responsible_email=responsible_data.email)
+        evaluation_data = EvaluationData(name_de=data[6], name_en=data[7], type_name=data[4], is_graded=data[5], degree_names=data[0], responsible_email=responsible_data.email, course_types=self.course_types, degrees=self.degrees)
```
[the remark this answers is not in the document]
pixunil: I now use some kind of factory which demotes `EvaluationData` to a dataclass; I hope that makes things easier to wrap ones head around

evap/staff/forms.py
```diff
@@ -25,6 +26,54 @@ def disable_all_fields(form):
...
+    def clean(self):
+        super().clean()
+        import_names = set()
+        for form in self.forms:
+            if self.can_delete and self._should_delete_form(form):
+                continue
+            for import_name in form.cleaned_data.get('import_names', []):
+                if import_name in import_names:
+                    form.add_error('import_names', _('Import name "{}" is duplicated.').format(import_name))
+                import_names.add(import_name)
```
He3lixxx, line 74: nice!

evap/staff/importers.py
```diff
@@ -65,47 +67,43 @@ def validate(self):
         user.clean_fields()
 
 
-class EvaluationData(CommonEqualityMixin):
+@dataclass
```
He3lixxx, line 70: nice²!

evap/staff/importers.py
```diff
@@ -243,11 +295,19 @@ def __init__(self):
         self.evaluations = {}
         self.enrollments = []
         self.names_de = set()
+        self.evaluation_data_factory = EvaluationDataFactory()
```
He3lixxx, line 298: nice³!

I feel like the class is quite long for what it does, but I don't see how it could be any shorter, so probably the problem is just a bit more complex than I'd intuitively assume...

evap/staff/tests/test_importers.py
```diff
@@ -181,6 +190,24 @@ def test_degrees_are_merged(self):
...
+        success_messages, warnings, errors = EnrollmentImporter.process(excel_content, self.semester, self.vote_start_datetime, self.vote_end_date, test_run=False)
+        self.assertIn("Successfully created 2 courses/evaluations, 4 students and 2 contributors:", "".join(success_messages))
+        self.assertEqual(errors, [])
+        self.assertEqual(warnings, {})
+
+        self.assertEqual(Course.objects.all().count(), 2)
+        course_shake = Course.objects.get(name_en="Shake")
+        self.assertEqual(course_shake.type.name_de, 'Vorlesung')
+        self.assertEqual(list(course_shake.degrees.values_list('name_en', flat=True)), ['Bachelor'])
+        course_speak = Course.objects.get(name_en="Build")
```
He3lixxx, line 206: I'm confused. Is it about building of speaking?

evap/staff/tests/test_importers.py
```diff
@@ -122,16 +122,25 @@ def test_import_makes_inactive_user_active(self):
 class TestEnrollmentImporter(TestCase):
     filename_valid = os.path.join(settings.BASE_DIR, "staff/fixtures/test_enrollment_data.xls")
     filename_valid_degree_merge = os.path.join(settings.BASE_DIR, "staff/fixtures/test_enrollment_data_degree_merge.xls")
+    filename_valid_import_names = os.path.join(settings.BASE_DIR, "staff/fixtures/test_enrollment_data_import_names.xls")
     filename_invalid = os.path.join(settings.BASE_DIR, "staff/fixtures/invalid_enrollment_data.xls")
     filename_random = os.path.join(settings.BASE_DIR, "staff/fixtures/random.random")
 
+    maxDiff = None
+
```
He3lixxx, line 130: ?

pixunil: This was useful as the diff generated by `assertCountEqual` got sometimes too long to display with the default settings. `None` always displays the full diff; if you rather want to have a fixed `maxDiff`, I can look which value is a good fit here

He3lixxx: ahh, i just didn't know that, looked like an unused variable for me.
Maybe add a comment? Idk.

pixunil: We decided against a comment as it is a property of the standard library [`TestCase.maxDiff`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.maxDiff).

evap/evaluation/migrations/0109_import_names.py
```diff
@@ -0,0 +1,51 @@
+from django.contrib.postgres.fields import ArrayField
+from django.db import migrations, models
+
+
+def fill_import_names(apps, schema_editor):
+    db_alias = schema_editor.connection.alias
+    Degree = apps.get_model('evaluation', 'Degree')
+    for degree in Degree.objects.using(db_alias).all():
```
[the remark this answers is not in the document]
pixunil: No, the `using(db_alias)` is not postgres-related. I think I added it because of the warning for [`RunPython`](https://docs.djangoproject.com/en/3.0/ref/migration-operations/#runpython), but I guess that most of the migrations don’t use that, so we should be fine without it…

To make a simple `update` work, I need an expression that creates an array based on a field. I tried `Degree.objects.update(import_names=[F('name_de')])`, but this sets all import_names to the string `name_de`, not the value of the field. Other quick option would be to write raw sql, but I think that is unnecessary here.

evap/staff/tests/test_importers.py
janno42: most evap developers probably don't know the library by heart and i also found it unexpected when first seeing it. adding a comment shouldn't hurt so i'd be happy if you could do that.

evap/evaluation/migrations/0109_import_names.py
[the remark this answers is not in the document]
He3lixxx: Ah, never mind, I just messed up there in the first comment. I thought you were giving each degree the same, constant value, but each degree gets a different one.

evap/staff/importers.py
```diff
@@ -151,6 +152,16 @@ def get_course_or_add_error(self, course_type_name, errors):
...
 
+    def parse_is_graded_or_add_error(self, is_graded, errors):
+        is_graded = is_graded.strip()
+        if is_graded == settings.IMPORTER_GRADED_YES:
+            return True
+        elif is_graded == settings.IMPORTER_GRADED_NO:
+            return False
+        else:
+            errors['is_graded'] = is_graded
+            return None
```
He3lixxx: Hmm, funny that pylint doesn't complain here, I thought we had a rule about if-return-else (will have to look into that, I changed about 50 occurences to be consistent, I was sure there was a rule that failed)

Could you make it either
```python
if is_graded == settings.IMPORTER_GRADED_YES:
    return True

if is_graded == settings.IMPORTER_GRADED_NO:
    return False

errors['is_graded'] = is_graded
return None
```


or

````python
if is_graded not in [settings.IMPORTER_GRADED_YES, settings.IMPORTER_GRADED_NO]:
    errors['is_graded'] = is_graded
    return None

return is_graded == settings.IMPORTER_GRADED_YES
````

for consistency with the rest of the code base?

Comment (He3lixxx) [reply evap/staff/importers.py]: Just noticed that you just moved it. In this case, I'm also fine with leaving it as it was.

Comment (He3lixxx) [reply evap/staff/importers.py]: Hmm, so pylint has this rule `no-else-return` (R1705), we haven't disabled it, yet it doesn't complain here...

Well, I guess if it doesn't complain, neither do it.

File: evap/staff/tests/test_importers.py (hunk 6 above)
Comment (pixunil) [reply evap/staff/tests/test_importers.py]: Now removed as the errors are categorized and the diff does not grow big enough anymore

File: evap/staff/importers.py (hunk 8 above)
Comment (pixunil) [reply evap/staff/importers.py]: Indeed, during development pylint has complained about this piece of code (one for `elif`, one for `else`). I assigned the fix for this to the wrong commit previously, sorry for confusing you

Comment (He3lixxx) [reply evap/staff/importers.py]: ah that makes sense and also explains why the comment was instantly outdated :D

File: evap/static/scss/components/_card.scss [hunk 9]
```diff
@@ -16,6 +16,14 @@
     background-color: rgba($evap-yellow, 0.5);
 }
 
+.card-outline-danger {
```
janno42, line 19: maybe you could update the other definitions as well so we have the same structure for all cards

PR #1444, opened 2020-04-06
Manually set semester to active

### LINKED ISSUE #1432: Manually making a semester active
Which semester is currently active (highlighted on student index and contributor index, starting page for staff and grade uploaders) should be manually configurable.

Currently the active semester is the one which was created last, but semesters can be created before they are actually running and in this case they don't necessarily should be handled as active.
Instead, staff users should be able to decide that a semester is currently active on the respective semester index page (next to the edit button) which will make this the active semester and whichever semester was active before, won't be anymore.

niklasmohrin:
Fixes #1432

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -61,7 +61,13 @@ class Semester(models.Model):
 
     @property
     def can_be_deleted_by_manager(self):
-        return self.evaluations.count() == 0 or (self.participations_are_archived and self.grade_documents_are_deleted and self.results_are_archived)
+        if self.is_active:
+            return False
+
+        if self.evaluations.count() == 0:
+            return True
+
+        return self.participations_are_archived and self.grade_documents_are_deleted and self.results_are_archived
 
     @property
     def participations_can_be_archived(self):
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -49,7 +49,7 @@ class Semester(models.Model):
 
     created_at = models.DateField(verbose_name=_("created at"), auto_now_add=True)
 
-    is_active_semester = models.BooleanField(default=None, unique=True, blank=True, null=True, verbose_name=_("semester is active"))
+    is_active = models.BooleanField(default=None, unique=True, blank=True, null=True, verbose_name=_("semester is active"))
 
     class Meta:
         ordering = ('-created_at', 'pk')
@@ -118,7 +118,7 @@ class Semester(models.Model):
 
     @classmethod
     def active_semester(cls):
-        return cls.objects.filter(is_active_semester=True).first()
+        return cls.objects.filter(is_active=True).first()
 
     @property
     def evaluations(self):
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -49,6 +49,8 @@ class Semester(models.Model):
 
     created_at = models.DateField(verbose_name=_("created at"), auto_now_add=True)
 
+    is_active_semester = models.BooleanField(default=None, unique=True, blank=True, null=True, verbose_name=_("semester is active"))
+
     class Meta:
         ordering = ('-created_at', 'pk')
         verbose_name = _("semester")
@@ -110,11 +112,7 @@ class Semester(models.Model):
 
     @classmethod
     def active_semester(cls):
-        return cls.objects.order_by("created_at").last()
-
-    @property
-    def is_active_semester(self):
-        return self == Semester.active_semester()
+        return cls.objects.filter(is_active_semester=True).first()
 
     @property
     def evaluations(self):
--- a/evap/staff/templates/staff_semester_view.html
+++ b/evap/staff/templates/staff_semester_view.html
@@ -11,6 +11,16 @@
         {% if request.user.is_manager %}
             &nbsp;
             <a href="{% url 'staff:semester_edit' semester.id %}" class="btn btn-sm btn-secondary">{% trans 'Edit' %}</a>
+            <button
+                onClick="makeActiveSemesterModalShow({{ semester.id }})" class="btn btn-sm btn-light"
+                id="makeActiveSemesterButton" type="button"
+                {% if semester.is_active_semester %}
+                    disabled
+                    data-toggle="tooltip" data-placement="top"
+                    title="{% blocktrans %}This is the active semester{% endblocktrans %}"
+                {% endif %}>
+                {% trans 'Make active' %}
+            </button>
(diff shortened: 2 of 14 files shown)


evap/staff/templates/staff_semester_view.html
```diff
@@ -11,7 +11,8 @@ <h3>
         {% if request.user.is_manager %}
             &nbsp;
             <a href="{% url 'staff:semester_edit' semester.id %}" class="btn btn-sm btn-secondary">{% trans 'Edit' %}</a>
-            <button type="button" id="deleteSemesterButton"{% if not semester.can_be_deleted_by_manager %} disabled{% endif %} onclick="deleteSemesterModalShow({{ semester.id }}, '{{ semester.name|escapejs }}');" class="btn btn-sm btn-danger">{% trans 'Delete' %}</button>
+            <button onClick="makeActiveSemesterModalShow({{ semester.id }})" class="btn btn-sm btn-danger" {% if semester.is_active %}disabled{% endif %} id="makeActiveSemesterButton" type="button">{% trans 'Make active' %}</button>
+            <button type="button" id="deleteSemesterButton"{% if not semester.can_be_deleted_by_manager or semester.is_active %} disabled{% endif %} onclick="deleteSemesterModalShow({{ semester.id }}, '{{ semester.name|escapejs }}');" class="btn btn-sm btn-danger">{% trans 'Delete' %}</button>
```
He3lixxx, line 15: So you made `is_active == False` a condition for a manager to delete a semester? This check should then be done in `can_be_deleted_by_manager`.

evap/staff/views.py
```diff
@@ -401,6 +401,22 @@ def semester_create(request):
...
 
+@require_POST
+@manager_required
+def semester_make_active(request):
+    semester_id = request.POST.get("semester_id")
+    semester = get_object_or_404(Semester, id=semester_id)
+
+    previous_active_semester = Semester.objects.filter(is_active=True).first()
+    previous_active_semester.is_active = None
+    previous_active_semester.save()
```
He3lixxx, line 412: `Semester.objects.update(is_active=None)`
is one line instead of three and it gives the advantage that this doesn't crash if no semester is currently active (which you will have when you set up a new instance of the software - I think we should allow users to at least make a semester active in this case)

evap/staff/views.py
```diff
@@ -422,6 +438,8 @@ def semester_delete(request):
 
     if not semester.can_be_deleted_by_manager:
         raise SuspiciousOperation("Deleting semester not allowed")
+    if semester.is_active:
```
He3lixxx, line 441: Should be contained in `can_be_deleted_by_manager`

(besides the semantic reasoning, here you also have the safety concern: you could have easily forgotten the check here)

niklasmohrin:
As you can see, I had to change some stuff in the tests when creating a semester sometimes. I think this is because no semester is flagged as active when just creating them in the tests. Here is what I am thinking:

The staff index page should not use `Semester.active_semester()` to decide where to redirect but rather something like `Semester.active_semester_or_last()` to avoid these issues.

Otherwise, one might have to think about flagging a semester as active when writing tests (which could also break something since the `is_active` field is unique) for something totally unrelated.

Another approach would be to just mark the semester that is created first as active, which would remove `None`s in places where there were weren't before. Still, this would (silently?) differ from the previous behavior and could break tests somewhere.

What do you guys think? Is this a real problem or am I overthinking this?

evap/staff/views.py
```diff
@@ -401,6 +401,20 @@ def semester_create(request):
...
 
+@require_POST
+@manager_required
+def semester_make_active(request):
+    semester_id = request.POST.get("semester_id")
+    semester = get_object_or_404(Semester, id=semester_id)
+
+    Semester.objects.update(is_active=None)
+
+    semester.is_active = True
```
karyon, line 412: I'd decorate the entire view with `@transaction.atomic` to eliminate race conditions (there's a brief period without an active semester)

evap/evaluation/models.py
```diff
@@ -110,11 +112,7 @@ def get_all_with_published_unarchived_results(cls):
 
     @classmethod
     def active_semester(cls):
-        return cls.objects.order_by("created_at").last()
-
-    @property
-    def is_active_semester(self):
-        return self == Semester.active_semester()
+        return cls.objects.filter(is_active=True).first()
```
karyon, line 115: You could do 
```suggestion
        try:
            return cls.objects.get(is_active=True)
        except Semester.DoesNotExist:
            return None
```
to express that there should be at most one active semester but not multiple, but it's rather wordy, maybe it's not worth it.

evap/staff/views.py
niklasmohrin: After a quick read of the documentation, I think we should just wrap the part after `get_object_or_404` into a `transaction.atomic` context, so that we don't need to get a lock for the database (or however `atomic` works here) if the request is invalid from the start.

What do you think?

evap/evaluation/models.py
```diff
@@ -59,7 +61,7 @@ def __str__(self):
 
     @property
     def can_be_deleted_by_manager(self):
-        return self.evaluations.count() == 0 or (self.participations_are_archived and self.grade_documents_are_deleted and self.results_are_archived)
+        return not self.is_active and (self.evaluations.count() == 0 or (self.participations_are_archived and self.grade_documents_are_deleted and self.results_are_archived))
```
He3lixxx, line 64: I think we can leave this like this for now, but in general, at some point it becomes more readable to split such statements like this:
```
if self.is_active:
    return False

return self.evaluations.count() == 0 or (self.participations_are_archived and self.grade_documents_are_deleted and self.results_are_archived)
```

evap/staff/views.py
He3lixxx: From the code perspective, the smaller context is a bit cleaner, so I'd go with that you can realize it with one additional line only (just as the decorator).

(I think it wouldn't really matter performance-wise, there's no real "locking" going on (as in: The whole db has to wait for this request). postgres is quite smart in being ACID while keeping good performance.)

@karyon good catch thinking about this, debugging might have been a nightmare here :D

evap/staff/templates/staff_semester_view.html
```diff
@@ -11,6 +11,7 @@ <h3>
         {% if request.user.is_manager %}
             &nbsp;
             <a href="{% url 'staff:semester_edit' semester.id %}" class="btn btn-sm btn-secondary">{% trans 'Edit' %}</a>
+            <button onClick="makeActiveSemesterModalShow({{ semester.id }})" class="btn btn-sm btn-danger" {% if semester.is_active %}disabled{% endif %} id="makeActiveSemesterButton" type="button">{% trans 'Make active' %}</button>
```
janno42, line 14: please use `btn-light` instead of `btn-danger` and add an explanatory tooltip if the semester is currently active ("This is the active semester")

janno42: Please remove the full stop, we usually don't have these in tooltips :)

PR #1452, opened 2020-04-27
Require login by default

### LINKED ISSUE #1450: Make login required for almost all views
We should make sure the default for all views is that they require a logged in user to be accessed. There are only a few exceptions of views available for everyone:
- Login index page
- FAQ
- Legal notice

niklasmohrin:
Fixes #1450


evap/middleware.py
```diff
@@ -0,0 +1,30 @@
...
+            return None
+
+        return redirect(f"{settings.LOGIN_URL}?next={request.path}")
+
+
+def no_login_required(func):
+    func.no_login_required = True
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        return func(*args, **kwargs)
```
He3lixxx, line 29: I wonder if these three lines are even required here - we don't want to add any additional code in the wrapper before or after the call, nor do we want to change arguments. Would simply `return func` also do it?

niklasmohrin: Wrapper is not needed. The decorator `evaluation.auth.user_passes_text` used to check for an authenticated user, this is gone now. I couldn't find any other redirects to the login page.

As for tests: Three auth tests failed initially, but I fixed that by adding `no_login_required` to `login_key_authentication`.

evap/evaluation/auth.py
```diff
@@ -33,16 +32,13 @@ def authenticate(self, request, key):  # pylint: disable=arguments-differ
 
 def user_passes_test(test_func):
     """
-    Decorator for views that checks whether users are authenticated
-    (redirecting to login if not) and pass a given test (raising 403
-    if not). The test should be a callable
-    that takes the user object and returns True if the user passes.
+    Decorator for views that checks whether pass a given test (raising 403
```
He3lixxx, line 35: You miss a subject here ("a user"?)

evap/middleware.py
```diff
@@ -0,0 +1,25 @@
+from django.contrib.auth.views import redirect_to_login
+
+
+class RequireLoginMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        return self.get_response(request)
```
He3lixxx, line 9: Kind of sad that django doesn't give us a base class that already implements these, but well...

PR #1453, opened 2020-04-27
Allow disabling the evaluation form for editors. Fixes #1417

### LINKED ISSUE #1417: Disabling evaluation form for editors
Managers should get the option to disable the editing of individual evaluations for editors, so that those cannot change the evaluation data (evaluation period and questionnaires). They should still be able to edit the contributor data. This will be used for special evaluations like exam evaluations, where the default values set by the managers should not be changed by editors.

Managers should get an additional checkbox field like `Disable evaluation data form for editors` or similar when editing an evaluation. If activated, editors will see a form where the respective fields are still visible but have been disabled. An additional button `Request changes` similar to the one for the list of participants should be added.

He3lixxx:
Fixes #1417

Should I change the new flag to false on some instances in the test data?


evap/staff/forms.py
```diff
@@ -183,9 +183,9 @@ class EvaluationForm(forms.ModelForm):
 
     class Meta:
         model = Evaluation
-        fields = ('course', 'name_de', 'name_en', 'weight', 'is_rewarded', 'is_midterm_evaluation',
-                  'vote_start_datetime', 'vote_end_date', 'participants', 'general_questionnaires',
-                  'last_modified_time', 'last_modified_user_name')
+        fields = ('course', 'name_de', 'name_en', 'weight', 'is_rewarded', 'allow_editors_to_edit',
```
janno42, line 186: please move `allow_editors_to_edit` before `is_rewarded`

He3lixxx: Alrighty. For self reference, in case I need to rebuild the test data:

```python
import random
e = list(Evaluation.objects.filter(course__semester=21))
chosen = random.choices(e, k=int(len(e)*0.2))
ids = [e.id for e in chosen]
Evaluation.objects.filter(id__in=ids).update(allow_editors_to_edit=False)
```

PR #1455, opened 2020-04-27
Fixes questionnaire filtering on evaluation creation

### LINKED ISSUE #1411: Questionnaire filtering on Evaluation creation broken
When creating a new Evaluation in the staff create evaluation form, in the general questionnaire list questionnaires which are set to be hidden but have never been used are listed. They should be hidden instead.

niklasmohrin:
Fixes #1411 

I believe this is a fix, but I could use a confirmation by someone.

### DIFF
--- a/evap/staff/tests/test_forms.py
+++ b/evap/staff/tests/test_forms.py
@@ -707,3 +707,15 @@ class EvaluationFormTests(TestCase):
         # Assert form is valid and locked questionnaire is removed
         form.save()
         self.assertEqual({questionnaire}, set(evaluation.general_contribution.questionnaires.all()))
+
+    def test_unused_questionnaire_visibility(self):
+        evaluation = baker.make(Evaluation)
+        questionnaire = baker.make(Questionnaire, visibility=Questionnaire.Visibility.HIDDEN, type=Questionnaire.Type.TOP)
+
+        form = EvaluationForm(instance=evaluation, semester=evaluation.course.semester)
+        self.assertNotIn(questionnaire, form.fields["general_questionnaires"].queryset)
+
+        evaluation.general_contribution.questionnaires.add(questionnaire)
+
+        form = EvaluationForm(instance=evaluation, semester=evaluation.course.semester)
+        self.assertIn(questionnaire, form.fields["general_questionnaires"].queryset)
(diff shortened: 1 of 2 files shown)


niklasmohrin:
Conclusion:
- `Questionnaire.objects.filter(contribution__evaluation=None).all()` is not empty, but contains the questionnaire "Master Thesis", a questionnaire with 0 uses
- in `EvaluationForm.__init__`, `self.instance` is not `None`, but a `Evaluation` with no `id` or `pk`
- somehow, the "Master thesis" questionnaire matches against the unsaved `Evaluation` (`self.instance`)

He3lixxx:
Hmm, best way to verify that you fixed something is simply writing a regression test. Then, there is also no doubt that you actually fixed the problem.

evap/staff/tests/test_forms.py
```diff
@@ -681,3 +681,30 @@ def test_locked_questionnaire(self):
...
+            form_data["general_questionnaires"]
+        )
+
+        form_data["general_questionnaires"] = [questionnaire2.pk]
+        form = EvaluationForm(form_data, instance=evaluation, semester=semester)
+        form.save()
+
+        form_data = get_form_data_from_instance(EvaluationForm, evaluation)
+
+        # Now, that no evaluation is using the first questionnaire, it should not be visible anymore
```
janno42, line 706: well. the questionnaire needs to be set to hidden if you want to test the case described in the issue :)
in form_data you will find only selected questionnaires, so you should check for the options or just the existence of the questionnaire's name on the webpage.

evap/staff/tests/test_forms.py
```diff
@@ -681,3 +681,29 @@ def test_locked_questionnaire(self):
...
+
+        # Delete the hidden questionnaire
+        form.save()
+
+        # Now, that no evaluation is using the first questionnaire, it should not be visible anymore
+        form = EvaluationForm(instance=evaluation, semester=evaluation.course.semester)
+        self.assertNotIn(
+            questionnaire,
+            form.fields["general_questionnaires"].queryset
+        )
```
He3lixxx, line 709:
```python
form = EvaluationForm(instance=evaluation, semester=evaluation.course.semester)
self.assertNotIn(questionnaire, form.fields["general_questionnaires"].queryset)

evaluation.general_contribution.questionnaires.add(questionnaire)
form = EvaluationForm(instance=evaluation, semester=evaluation.course.semester)
self.assertIn(questionnaire, form.fields["general_questionnaires"].queryset)
```

He3lixxx:
Check out codacy, it says you have a blank line at the end of a file

PR #1457, opened 2020-04-29
Fix wrong participant numbers of in-progress evaluations

### LINKED ISSUE #1456: Participant count wrong for unpublished results
On the results index page the number of voters and participants for unpublished evaluations is not shown correctly. Staff users can see these evaluations before they are published, but the numbers differ from the actual values (which are shown correctly on results detail pages).

karyon:
fixes #1456. Probably introduced in #1423.


He3lixxx:
emergency bug fix department saves the day with incredible response times ... again! :D

PR #1459, opened 2020-05-04
Refresh results cache when changing username

### LINKED ISSUE #1422: Changing a username won’t refresh results cache
When the name of a contributor is changed, the new name won’t be reflected on the results pages.

niklasmohrin:
Fixes #1422 

Done as on line 280 in staff/tools

### DIFF
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -15,6 +15,7 @@ from evap.evaluation.models import (Contribution, Course, CourseType, Degree, Em
                                     FaqSection, Question, Questionnaire, RatingAnswerCounter, Semester, TextAnswer,
                                     UserProfile)
 from evap.evaluation.tools import date_to_datetime
+from evap.results.tools import collect_results
 from evap.results.views import (update_template_cache,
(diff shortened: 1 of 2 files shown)


evap/staff/forms.py
```diff
@@ -702,6 +703,10 @@ def save(self, *args, **kw):
 
         self.instance.is_active = not self.cleaned_data.get('is_inactive')
 
+        # refresh results cache
+        for evaluation in Evaluation.objects.filter(contributions__contributor=self.instance).distinct():
```
janno42, line 707: this should only be done when the name actually changed. when, e.g., just updating the `email`, we don't have to refresh the results.

evap/staff/forms.py
```diff
@@ -702,6 +703,10 @@ def save(self, *args, **kw):
...
+            collect_results(evaluation, force_recalculation=True)
```
karyon, line 708: as this can take a while on some userprofiles, it might be worth it to restrict this to `if any(attribute in self.changed_data for attribute in ['first_name', 'last_name', 'title']` (untested).

PR #1460, opened 2020-05-04
Speed up travis by running everything in parallel

He3lixxx:
Do everything, but faster!


He3lixxx:
I we're lucky, it will now run in about 3 minutes instead of seven. If we're unlucky and travis decides to run everything linearly even though it could be run in parallel, we might get slowed down to about 12 minutes.

PR #1463, opened 2020-05-11
Tests for the results exporter

### LINKED ISSUE #999: Result exporter is not tested
there are some tests in staff/tests/test_views that test the exporter, but it seems like it is not creating any questionnaires. large portions of the results exporter are not covered therefore.

niklasmohrin:
Fixes #999

### DIFF
--- a/evap/results/tests/test_exporters.py
+++ b/evap/results/tests/test_exporters.py
@@ -313,6 +313,19 @@ class TestExporters(TestCase):
         self.assertNotIn(unused_questionnaire.name, sheet.col_values(0))
         self.assertNotIn(unused_question.text, sheet.col_values(0))
 
+    def test_multiple_evaluations(self):
+        semester = baker.make(Semester)
+        degree = baker.make(Degree)
+        evaluation1 = baker.make(Evaluation, course__semester=semester, course__degrees=[degree], state="published")
+        evaluation2 = baker.make(Evaluation, course__semester=semester, course__degrees=[degree], state="published")
+
+        sheet = self.get_export_sheet(semester, degree, [evaluation1.course.type.id, evaluation2.course.type.id])
+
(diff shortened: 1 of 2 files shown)


niklasmohrin:
Things to test:
- [ ] Multiple sheets
- [ ] `include_unpublished`
- [ ] `include_not_enough_voters`
- [ ] exclusion of single-result evaluations
- [ ] probably

niklasmohrin:
The tests are getting a bit lengthy, do you see any possibilities for shortcuts?

niklasmohrin:
Any more suggestions on what should be tested?

evap/results/tests/test_exporters.py
```diff
@@ -202,6 +202,165 @@ def test_course_type_ordering(self):
...
+
+        sheet = self.get_export_sheet(evaluation.course.semester, degree, [evaluation.course.type.id])
+        self.assertEqual(sheet.row_values(4)[0], used_questionnaire.name)
+        self.assertEqual(sheet.row_values(5)[0], used_question.text)
+        self.assertNotIn(unused_questionnaire.name, sheet.col_values(0))
+        self.assertNotIn(unused_question.text, sheet.col_values(0))
+
+    def test_degree_course_type_name(self):
+        degree = baker.make(Degree, name_en="Celsius")
+        course_type = baker.make(CourseType, name_en="LetsPlay")
```
He3lixxx, line 322: I think it's funny, however, I'm not sure whether we want to be more serious. @janno42?

janno42: I'm fine with that in tests. People writing tests do need to have some fun in between because the tests usually aren't :)
We just shouldn't use it in our test_data.json, as it might be confusing for some people because `LetsPlay` is not something users would usually expect for a `CourseType`.

PR #1464, opened 2020-05-11
Update dependencies

He3lixxx:


He3lixxx:
Currently we have the pylint and pylint-django version two times, once in requirements-dev.txt and once in requirements-travis.txt. I don't quite remember why I made it that way, seems a bit complicated...

.pylintrc
```diff
@@ -79,8 +79,7 @@ disable=locally-disabled,            # we allow locally disabling some checks if
 # it would be nice to have these enabled, but sadly pylint is currently buggy / unusable with these
         arguments-differ,            # currently, if a method takes *args and **kwargs and passes them to the super method,
                                      # pylint compains - see https://github.com/PyCQA/pylint/pull/3001
```
janno42, line 81: just seeing this now: "compains" :)

PR #1471, opened 2020-05-18
Excel Exporter refactor

### LINKED ISSUE #1404: ExcelExporter refactoring
Keep #999 in mind.

- Currently, we have two ExcelExporter classes (`evap.results.exporters.ExcelExporter` and `evap.rewards.exporters.ExcelExporter`). These should be named differently for readability - I suggest ResultsExcelExporter and RewardsExcelExporter.
- In `evap.results.exporters`, thee helper functions are defined: `writen`, `writec`, and `_write`. The excel exporter classes have member variables `col`, `row` and `sheet`. The helper methods use and change these members to write to the desired field. Thus, these functions should be methods of the class that has the `col`, `row` and `sheet` members - they currently are not so that `evap.rewards.exporters` can import them and use them. I'd suggest to put the essential features we often use to build the excel file in a small helper class that could be named `ExcelFile` or similar. This class then encapsulated the xlwt workbook and offers the interface that is convenient to use in the `ExcelExporter` classes - `writec`, `writen`, `save` and the `styles`. I suggest thinking of a more verbose name for `writec` and `writen`.
As a consequence, the `ExcelExporter` classes should have no non-business-logic left as all the technical details are hidden in the new class.
- In the results exporters, the `export` method should be split up (currently 157 lines).

Opinions? Agree / Disagree?

niklasmohrin:
Eventually fixes #1404

### DIFF
--- a/evap/evaluation/tools.py
+++ b/evap/evaluation/tools.py
@@ -1,5 +1,7 @@
+from abc import ABC, abstractmethod
 import datetime
 from urllib.parse import quote
+import xlwt
 
 from django.conf import settings
 from django.contrib.auth import user_logged_in
@@ -70,3 +72,67 @@ class FileResponse(HttpResponse):
             self["Content-Disposition"] = f"attachment; filename=\"{filename}\""
         except UnicodeEncodeError:
             self["Content-Disposition"] = f"attachment; filename*=utf-8''{quote(filename)}"
+
+
+class ExcelExporter(ABC):
+    styles = {
+        "default":                  xlwt.Style.default_style,
+        "headline":                 xlwt.easyxf("font: bold on, height 400; alignment: horiz centre, vert centre, wrap on; borders: bottom medium", num_format_str="0.0"),
+        "bold":                     xlwt.easyxf("font: bold on"),
+        "italic":                   xlwt.easyxf("font: italic on"),
+        "border_left_right":        xlwt.easyxf("borders: left medium, right medium"),
+        "border_top_bottom_right":  xlwt.easyxf("borders: top medium, bottom medium, right medium"),
+        "border_top":               xlwt.easyxf("borders: top medium"),
+    }
+
+    # Derived classes can set this to
+    # have a sheet added at initialization.
+    default_sheet_name = None
+
+    def __init__(self):
+        self.workbook = xlwt.Workbook()
+        self.cur_row = 0
+        self.cur_col = 0
+        if self.default_sheet_name is not None:
+            self.cur_sheet = self.workbook.add_sheet(self.default_sheet_name)
+        else:
+            self.cur_sheet = None
+
+    def write_cell(self, label="", style="default"):
+        """Write a single cell and move to the next column."""
+        self.cur_sheet.write(
+            self.cur_row,
+            self.cur_col,
+            label,
+            self.styles[style],
+        )
+        self.cur_col += 1
+
+    def next_row(self):
+        self.cur_col = 0
+        self.cur_row += 1
+
+    def write_row(self, vals, style="default"):
+        """
+        Write a cell for every value and go to the next row.
+        Styling can be chosen
+          - once for all cells by providing a string.
+          - separately for every value, based on a callable that maps the value to a style name.
+        """
+        for val in vals:
+            self.write_cell(val, style=style(val) if callable(style) else style)
+        self.next_row()
+
+    def write_empty_row_with_styles(self, styles):
+        for style in styles:
+            self.write_cell(None, style)
+        self.next_row()
+
+    @abstractmethod
+    def export_impl(self, *args, **kwargs):
+        """Specify the logic to insert the data into the sheet here."""
+
+    def export(self, response, *args, **kwargs):
+        """Convenience method to avoid some boilerplate."""
+        self.export_impl(*args, **kwargs)
+        self.workbook.save(response)
--- a/evap/results/exporters.py
+++ b/evap/results/exporters.py
@@ -281,7 +281,7 @@ class ResultsExporter(ExcelExporter):
 ResultsExporter.init_grade_styles()
 
 
-class TextAnswerExcelExporter(ExcelExporter):
+class TextAnswerExporter(ExcelExporter):
 
     class InputData:
         def __init__(self, contribution_results):
@@ -302,7 +302,7 @@ class TextAnswerExcelExporter(ExcelExporter):
         self.evaluation_name = evaluation_name
         self.semester_name = semester_name
         self.responsibles = responsibles
-        assert isinstance(results, TextAnswerExcelExporter.InputData)
+        assert isinstance(results, TextAnswerExporter.InputData)
         self.results = results
         self.contributor_name = contributor_name
 
--- a/evap/results/exporters.py
+++ b/evap/results/exporters.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict, defaultdict
+from itertools import chain, repeat
 
 from django.db.models import Q
 from django.utils.translation import gettext as _
@@ -291,8 +292,7 @@ def _write(exporter, label, style, rows, cols):
         exporter.sheet.write(exporter.row, exporter.col, label, style)
 
 
-class TextAnswerExcelExporter:
-    # pylint: disable=too-many-instance-attributes
+class TextAnswerExcelExporter(ExcelExporter):
 
     class InputData:
         def __init__(self, contribution_results):
@@ -306,7 +306,10 @@ class TextAnswerExcelExporter:
                         answers = [answer.answer for answer in result.answers]
                         self.questionnaires[q_type].append((contributor_name, result.question, answers))
 
+    default_sheet_name = _("Text Answers")
+
     def __init__(self, evaluation_name, semester_name, responsibles, results, contributor_name):
+        super().__init__()
         self.evaluation_name = evaluation_name
         self.semester_name = semester_name
         self.responsibles = responsibles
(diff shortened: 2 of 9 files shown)


evap/evaluation/tools.py
```diff
@@ -70,3 +73,100 @@ def set_content_disposition(self, filename):
...
+        "headline":                 xlwt.easyxf("font: bold on, height 400; alignment: horiz centre, vert centre, wrap on; borders: bottom medium", num_format_str="0.0"),
+        "bold":                     xlwt.easyxf("font: bold on"),
+        "italic":                   xlwt.easyxf("font: italic on"),
+        "border_left_right":        xlwt.easyxf("borders: left medium, right medium"),
+        "border_top_bottom_right":  xlwt.easyxf("borders: top medium, bottom medium, right medium"),
+        "border_top":               xlwt.easyxf("borders: top medium"),
+    }
+
+    add_default_sheet = True
+    default_sheet_name = "EvaP Export"
```
karyon, line 90: You could do `default_sheet_name = None` and then below check for `if self.default_sheet_name is not None:`. It's one setting less and the unused (?) default value becomes a feature.

evap/evaluation/tools.py
```diff
@@ -70,3 +73,100 @@ def set_content_disposition(self, filename):
...
+            self.write_cell(val, style_name=map_style(val))
+        self.next_row()
+
+    @classmethod
+    def init_styles(cls):
+        return cls.default_styles.copy()
+
+    def create_color(self, color_name, palette_index, color):
+        xlwt.add_palette_colour(color_name, palette_index)
+        self.workbook.set_colour_RGB(palette_index, *color)
```
karyon, line 160: since this relies on the CUSTOM_COLOR_START stuff to work, i would either move this method to the one exporter that needs it, or move the palette_index logic into this class here (keeping a counter of the already taken palette indices, starting with CUSTOM_COLOR_START )

niklasmohrin:
Apart from the inheritance stuff, there are two major changes in this pull request:

1. Styles are now a class level matter. All styles are managed in the `styles` class attribute. The base `ExcelExporter` already has a lot of styles, so that the `RewardsExporter` and `TextAnswerExcelExporter` (should this be renamed to `TextAnswerExporter`?) don't need any more styles. The `ResultsExporter` defines some additional styles at definition (on top of the base styles) and has some computed styles, that are added right after the class definition (for reasons).
2. Instead of `writec` (write a cell and move the cursor one to the right) and `writen` (go to the beginning of the next row and call `writec`), we now have `write_cell` (in combination with `next_row` if needed) and `write_row` (writes all values of a list and moves the cursor to the first cell in the next row). Before, there was this weird pattern that involved calling `writen` once and then repeatedly `writec`. This means, that when writing a row (of grades for example), you had to do all the writing in one place, but  were forced to carry out the line break in the next step of whatever you were doing. Now, all the logic for a single row can happen in one place. Furthermore, I think that this is also closer to what is intended and easier to read (take for example `RewardsExporter.export_impl`).

Some things I stumbled upon:

- When restructuring the `ResultExporter`, I was wondering what the intention of the `course_grade_exists` stuff is. This is a variable, that may be set while writing, but also stays set for the following sheets? Is it not possible, that none of the evaluations on the next sheet have course grades? Why?

- The row with the course grades can be drawn, even if all of the course grades are `None`. Originial code:
```python
writen(self, _("Course Grade"), "bold")
for evaluation, results in evaluations_with_results:
    if evaluation.course_evaluations_count > 1:
        if evaluation.course.avg_grade:
            writec(self, evaluation.course.avg_grade, self.grade_to_style(evaluation.course.avg_grade))
        else:
            self.write_empty_cell_with_borders()
    else:
        self.write_empty_cell()
```
Current version:
```python
if course_results_exist:
    # Only query the number of evaluations once and keep track of it here.
    count_gt_1 = list(map(lambda e: e.course_evaluations_count > 1, evaluations))

    # ...

    self.write_cell(_("Course Grade"), "bold")
    for evaluation, gt1 in zip(evaluations, count_gt_1):
        avg = evaluation.course.avg_grade
        if not gt1:
            style = "default"
        elif avg is not None:
            style = self.grade_to_style(avg)
        else:
            style = "border_left_right"
        self.write_cell(avg, style)
    self.next_row()

    # ...
```

niklasmohrin:
We think that the `course_results_exist` thing is not intended. We will change the variable to only live for one iteration. Therefore, following sheets will only have the course result row if there are course results on this sheet.

niklasmohrin:
Right, looks like I missed some stuff. The course average grade crash should be fixed, the renaming is done and the line is gone. I will look into the styling another day. About the other crash:

This is not caused by these changes. The error occurs before the exporter is called:

```python
@manager_required
def semester_export(request, semester_id):

    ExportSheetFormset = formset_factory(form=ExportSheetForm, can_delete=True, extra=0, min_num=1, validate_min=True)
    formset = ExportSheetFormset(request.POST or None, form_kwargs={'semester': semester})

    if formset.is_valid():
        include_not_enough_voters = request.POST.get('include_not_enough_voters') == 'on'
        include_unpublished = request.POST.get('include_unpublished') == 'on'
        selection_list = []
        for form in formset:
            selection_list.append((form.cleaned_data['selected_degrees'], form.cleaned_data['selected_course_types']))
            # ^ Line 515, error occurs here

        filename = "Evaluation-{}-{}.xls".format(semester.name, get_language())
        response = FileResponse(filename, content_type="application/vnd.ms-excel")

        ResultsExporter().export(
            response, [semester], selection_list, include_not_enough_voters, include_unpublished
        )
        return response

    return render(request, "staff_semester_export.html", dict(semester=semester, formset=formset))
```

I _was_ able to reproduce the bug **on master** (`a1dca61b7e700f8ffe8238268bc8682801d5a0f9`). At first, I could not reproduce the error, but I got it after some testing. I broke with `master` and `lecture` selected (also other additions threw). Strangely, selecting the two after a refresh (without resending form data; by reentering the url) was no problem and I got the file prompt.

In conclusion, I think this is a problem with the form and auth or something, I have no idea :smile: We should probably open another issue.

niklasmohrin:
Okay, I found the mistake that caused the weird styling. I looked at the xlwt reference and source and I don't think there is a good way to test for styling (except for comparing binary data ?).

Other than that, this should be all done here, I can squash the fixups after review or beforehand, if you want.

evap/results/exporters.py
```diff
@@ -79,220 +87,198 @@ def filter_text_and_heading_questions(questions):
...
+                    self.write_cell(avg, self.grade_to_style(avg))
+            self.next_row()
+
+        self.write_empty_row_with_styles(["border_left_right"] * len(evaluations_with_results))
+
+    def export_impl(self, semesters, selection_list, include_not_enough_voters=False, include_unpublished=False, contributor=None):
         # the excel file we're creating here is rather complex. However, from the nature of a single
         # file, it doesn't make much sense to split up the code into different methods as they will
         # always be tightly coupled based on the layout of the sheet. We thus think that one big method
         # containing the business logic is okay here
```
karyon, line 246: obsolete comment?

evap/results/exporters.py
```diff
@@ -79,220 +87,193 @@ def filter_text_and_heading_questions(questions):
...
+
+        self.write_cell(_("Evaluation rate"), "bold")
+        # round down like in progress bar
+        participant_percentages = map(lambda e: int((e.num_voters / e.num_participants) * 100) if e.num_participants > 0 else 0, evaluations)
+        participant_percentages = map(lambda p: f"{p}%", participant_percentages)
+        self.write_row(participant_percentages, style="evaluation_rate")
+
+        if course_results_exist:
+            # Only query the number of evaluations once and keep track of it here.
+            count_gt_1 = list(map(lambda e: e.course_evaluations_count > 1, evaluations))
```
He3lixxx, line 177: have you considered using list comprehensions here and in 180?

evap/results/exporters.py
```diff
@@ -79,220 +87,193 @@ def filter_text_and_heading_questions(questions):
...
-            writen(self)
-            for evaluation, results in evaluations_with_results:
-                self.write_empty_cell_with_borders()
+            evaluations_with_results, used_questionnaires, course_results_exist = self.filter_evaluations(
+                semesters,
+                evaluation_states,
+                degrees,
+                course_types,
+                contributor,
+                include_not_enough_voters
```
janno42, line 261: comma :)

evap/results/exporters.py
```diff
@@ -1,68 +1,76 @@
...
+        Adds the grade styles to cls.styles and as a xlwt identifier.
+        This also notes all registered colors in cls.COLOR_MAPPINGS for the instances.
+
+        This method should only be called once, right after the class definition.
+        Instances need the styles, but they should only be registered once for xlwt.
+        """
+
+        if len(cls.COLOR_MAPPINGS) > 0:
+            # Method has already been called (probably in another import of this file).
+            return
```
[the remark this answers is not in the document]
niklasmohrin: Didn't know about this, very cool. I don't know if there should be a warning, since it is okay, that the file is imported in multiple places, right? On the other hand, I am not sure, whether the file is "run" multiple times anyway (I tried with some `print` statements, but only got one line in the console output). I guess I can add an `ImportWarning`, since this is about importing and these warnings are ignored by default.
What do you think?

evap/results/exporters.py
niklasmohrin: I haven't, good idea. I removed all occurrences of `map` in the class and replaced them with list or generator expressions. :+1:

evap/results/exporters.py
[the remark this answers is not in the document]
He3lixxx: With the current code, even with multiple imports, this code should only be run once.
Afaik, all warnings are ignored by default.
`ImportWarning` sounds good.

PR #1472, opened 2020-05-19
Course evaluation creation redesign

### LINKED ISSUE #1405: Course and Evaluation creation redesign
The creation of Courses and Evaluations should get a different user flow.

- The current dropdown button for creating a Course, Evaluation or Single Result and the import of data should be replaced with a single (light) button just for the import in the staff semester view.
- On the "Courses" tab a new (dark) button for creating a Course should be placed on the upper left with the same functionality as before.
- In each row of the table a column with one (dark) button for adding an Evaluation and one (dark) button for adding a Single Result to this Course should be added on the right.
- When clicking on these buttons, the Course field in the Evaluation form/Single Result form should be pre-filled with the respective Course.
- Additionally, each Evaluation in the "Evaluations" tab should get a "Copy" button, which, similar to the questionnaire copying, will open an Evaluation creation form with all fields (including participants and contributors) pre-filled like in the original Evaluation.

pixunil:
Resolves #1405 

Some background about copying models:
Not only the evaluation object must be copied, also the related contribution objects should be copyable.
The current approach uses the form fields to build `initial` data from the original objects (some fields needed manual cleaning though).
Passing the original object as-is to the form and setting `pk = None` just before saving will result in possible locked forms. Setting `pk = None` before displaying it will still result in saving some unwanted attributes (like state).
For contributions, I first wanted to pass `initial = [initial_contribution_data(contribution) for contribution in evaluation.contributions.all()]` to the `InlineContributionFormset`. This behaves very quirky, one has to set `queryset = Contribution.objects.none()` `extra = evaluation.contributions.count()` – but this messes up ordering.
If you want to learn more about the details, you can look into my resolved self-review

### DIFF
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -1,4 +1,5 @@
 from collections.abc import Iterable
+from datetime import datetime
 import logging
 
 from django import forms
@@ -326,6 +327,15 @@ class EvaluationForm(forms.ModelForm):
         return evaluation
 
 
+class EvaluationCopyForm(EvaluationForm):
+    def __init__(self, data=None, instance=None):
+        opts = self._meta
+        initial = forms.models.model_to_dict(instance, opts.fields, opts.exclude)
+        initial['last_modified_time'] = datetime.now()
+        initial['general_questionnaires'] = instance.general_contribution.questionnaires.all()
+        super().__init__(data=data, initial=initial, semester=instance.course.semester)
+
+
 class SingleResultForm(forms.ModelForm):
     last_modified_time_2 = forms.DateTimeField(label=_("Last modified"), required=False, localize=True, disabled=True)
     last_modified_user_2 = forms.CharField(label=_("Last modified by"), required=False, disabled=True)
@@ -456,6 +466,18 @@ class ContributionForm(forms.ModelForm):
     def clean(self):
         if not self.cleaned_data.get('does_not_contribute') and not self.cleaned_data.get('questionnaires'):
             self.add_error('does_not_contribute', _("Select either this option or at least one questionnaire!"))
+
+
+class ContributionCopyForm(ContributionForm):
+    def __init__(self, data=None, instance=None, evaluation=None, **kwargs):
+        initial = None
+        copied_instance = Contribution(evaluation=evaluation)
+        if instance:
+            opts = self._meta
+            initial = forms.models.model_to_dict(instance, opts.fields, opts.exclude)
+            del initial['evaluation']
+            initial['does_not_contribute'] = not instance.questionnaires.exists()
+        super().__init__(data, initial=initial, instance=copied_instance, evaluation=evaluation, **kwargs)
 
 
 class EvaluationEmailForm(forms.Form):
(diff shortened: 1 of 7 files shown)


evap/staff/forms.py
```diff
@@ -646,6 +669,16 @@ def clean(self):
...
+class ContributionCopyFormSet(ContributionFormSet):
+    def __init__(self, data, instance, new_instance):
+        super().__init__(data, instance=instance, form_kwargs={'evaluation': new_instance})
+        self.instance = new_instance
+
+    def save(self, commit=True):
+        self.save_as_new = True
+        super().save(commit)
+
+
```
pixunil, line 681: This does not need to be a class – all calls can be inlined.
But to explain what’s going on:
While creating the forms, the old instance (= evaluation) has to be passed, as for each object a form will be created. Before saving, the instance must switch to the new evaluation (otherwise, we would copy the contributions to the original evaluation, which would generate unique errors in most cases). This can be done before displaying the form, but it can also happen in `save()`, like setting `save_as_new = True`.

On the other hand, `save_as_new = True` is needed for the side effect that it makes `initial_form_count()` return 0. Otherwise, the formset will skip the initial forms, as it thinks that they should be ignored (as their `instance.pk is None`). But, `save_as_new` must not be set to `True` when displaying the formset – otherwise, the user would see only one blank extra form and would be sad to create all contributions again.

I know that this explanation would be later expressed as comments, but for discussions, I think this more lengthy description would suit better.

evap/staff/tests/test_views.py
```diff
@@ -1219,6 +1219,11 @@ def setUpTestData(cls):
         cls.manager_user = baker.make(UserProfile, username='manager', groups=[Group.objects.get(name='Manager')])
         cls.course = baker.make(Course, semester=baker.make(Semester, pk=1))
 
+    def test_course_is_prefilled(self):
+        response = self.app.get(f'{self.url}/{self.course.pk}', user='manager', status=200)
+        form = response.context['form']
```
pixunil, line 1224: This is something I found today: You can retrieve the Django objects (like the form) via `response.context`. There even exists a [`self.assertFormError`](https://docs.djangoproject.com/en/3.0/topics/testing/tools/#django.test.SimpleTestCase.assertFormError), which may be the thing I looked for. Anyway, in this test case I queried the django form just for simple convenience as the webtest form would require to put `str(self.course.pk)` on the right side (so no big advantage here)

evap/staff/tests/test_forms.py
```diff
@@ -719,3 +762,48 @@ def test_unused_questionnaire_visibility(self):
...
+    def test_initial_from_original(self):
+        form = EvaluationCopyForm(None, self.evaluation)
+        self.assertEqual(form['course'].initial, self.course.pk)
+        self.assertCountEqual(form.fields['course'].queryset, self.semester.courses.all())
+        self.assertEqual(form['name_de'].initial, "Das Original")
+        self.assertEqual(form['name_en'].initial, "The Original")
+        self.assertCountEqual(form['participants'].initial, self.participants)
+        self.assertGreater(form['last_modified_time'].initial, self.evaluation.last_modified_time)
+        self.assertEqual(form['last_modified_user_name'].initial, None)
+        self.assertCountEqual(form['general_questionnaires'].initial, self.general_questionnaires)
```
pixunil, line 790: I tried to write this test as view test, but this was very inconvenient, as I had to stringify nearly everything :/
But, there is currently no guarantee that the `EvaluationCopyForm` will be used by the form – using `response.context['evaluation_form']` would combine both benefits (only thing that is not checked whether the form is rendered correctly, but I think this is django’s job)
What do you think?

evap/staff/forms.py
```diff
@@ -326,6 +327,15 @@ def save(self, *args, **kw):
         return evaluation
 
 
+class EvaluationCopyForm(EvaluationForm):
+    def __init__(self, data=None, instance=None):
+        opts = self._meta
```
He3lixxx, line 332: You only use this once, right? Just use self._meta in-place then?

evap/staff/views.py
```diff
@@ -806,6 +808,33 @@ def evaluation_create(request, semester_id, course_id=None):
...
+        copied_evaluation.set_last_modified(request.user)
+        copied_evaluation.save()
+        formset.save()
+        update_template_cache_of_published_evaluations_in_course(copied_evaluation.course)
+
+        messages.success(request, _("Successfully created evaluation."))
+        return redirect('staff:semester_view', semester_id)
+
+    return render(request, "staff_evaluation_form.html", dict(
+        semester=semester, evaluation_form=form, formset=formset, manager=True,
```
janno42, line 833: we usually put these in individual lines

PR #1474, opened 2020-05-23
Completely remove usernames

janno42:
This completes #1374


evap/evaluation/migrations/0114_remove_username.py
```diff
@@ -0,0 +1,21 @@
...
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='userprofile',
+            options={'ordering': ('last_name', 'first_name', 'email'), 'verbose_name': 'user', 'verbose_name_plural': 'users'},
+        ),
+        migrations.RemoveField(
+            model_name='userprofile',
+            name='username',
```
karyon, line 19: fwiw, this is probably not reversible. (since we never checked, we probably already have other migrations that are not reversible, and i know we didn't need a reversible migration history for a while, so maybe we shouldn't care).

evap/results/tests/test_views.py
```diff
@@ -272,8 +272,8 @@ def setUpTestData(cls):
     def setUp(self):
         self.evaluation = Evaluation.objects.get(pk=self.evaluation.pk)
 
-    def helper_test_answer_visibility_one_voter(self, username, expect_page_not_visible=False):
-        page = self.app.get("/results/semester/2/evaluation/22", user=username, expect_errors=expect_page_not_visible)
+    def helper_test_answer_visibility_one_voter(self, user, expect_page_not_visible=False):
+        page = self.app.get("/results/semester/2/evaluation/22", user=user, expect_errors=expect_page_not_visible)
```
karyon, line 276: user_email?

evap/results/tests/test_views.py
```diff
@@ -285,8 +285,8 @@ def helper_test_answer_visibility_one_voter(self, username, expect_page_not_visi
             number_of_disabled_grade_badges = str(page).count("badge-grade badge-disabled")
             self.assertEqual(number_of_disabled_grade_badges, 5)
 
-    def helper_test_answer_visibility_two_voters(self, username):
-        page = self.app.get("/results/semester/2/evaluation/22", user=username)
+    def helper_test_answer_visibility_two_voters(self, user):
+        page = self.app.get("/results/semester/2/evaluation/22", user=user)
```
karyon, line 289: user_email?

evap/evaluation/migrations/0114_remove_username.py
janno42: true, as the username was not nullable this migration is not reversible. in case we should run in serious troubles after the update it shouldn't be too hard to manually reverse the migration by filling the usernames with the email addresses.
but as you said, we most probably won't need that at all. so should i add some additional code for making it reversible or shall we just leave it as it is?

evap/results/tests/test_views.py
```diff
@@ -514,37 +514,37 @@ def test_textanswer_visibility_for_student(self):
...
+        cls.other_contributor_1 = baker.make(UserProfile, email='other_contributor_1@institution.example.com')
         baker.make(Contribution, evaluation=cls.evaluation, contributor=cls.other_contributor_1, questionnaires=[questionnaire], textanswer_visibility=Contribution.TextAnswerVisibility.OWN_TEXTANSWERS)
-        cls.other_contributor_2 = baker.make(UserProfile, username='other contributor 2')
+        cls.other_contributor_2 = baker.make(UserProfile, email='other_contributor_2@institution.example.com')
         baker.make(Contribution, evaluation=cls.evaluation, contributor=cls.other_contributor_2, questionnaires=[questionnaire], textanswer_visibility=Contribution.TextAnswerVisibility.OWN_TEXTANSWERS)
 
     def test_contributor_list(self):
         url = '/results/semester/{}/evaluation/{}?view=export'.format(self.semester.id, self.evaluation.id)
-        page = self.app.get(url, user='responsible')
+        page = self.app.get(url, user="responsible@institution.example.com")
```
He3lixxx, line 547: Not in the scope of this PR

Here, it would be much easier if we changed line 532 to
```python
cls.responsible = baker.make(UserProfile, email='responsible@institution.example.com')
```
so we can use
```python
page = self.app.get(url, user=self.responsible)
```
here.

It seems like we have this case a few times: The UserProfile instance is available or can easily be made available, yet we still use the username, which is rather unstable in my opinion. Should I add an issue about this?

evap/student/tests/test_views.py
```diff
@@ -26,10 +26,10 @@ class TestVoteView(WebTest):
...
     @classmethod
     def setUpTestData(cls):
-        cls.voting_user1 = baker.make(UserProfile)
-        cls.voting_user2 = baker.make(UserProfile)
-        cls.contributor1 = baker.make(UserProfile)
-        cls.contributor2 = baker.make(UserProfile)
+        cls.voting_user1 = baker.make(UserProfile, email="voting_user1@institution.example.com")
+        cls.voting_user2 = baker.make(UserProfile, email="voting_user2@institution.example.com")
+        cls.contributor1 = baker.make(UserProfile, email="contributor1@institution.example.com")
+        cls.contributor2 = baker.make(UserProfile, email="contributor2@institution.example.com")
```
He3lixxx, line 32: I think we don't need the email here and can just use `self.voting_user1` as the user argument for `self.app.get` calls below

evap/evaluation/models.py
```diff
@@ -1176,12 +1176,11 @@ class Meta:
 
 
 class UserProfileManager(BaseUserManager):
-    def create_user(self, username, email=None, password=None, first_name=None, last_name=None):
-        if not username:
-            raise ValueError(_('Users must have a username'))
+    def create_user(self, email, password=None, first_name=None, last_name=None):
+        if not email:
```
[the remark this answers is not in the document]
janno42: you're right, we can allow it here

evap/evaluation/models.py
```diff
@@ -1257,7 +1253,7 @@ def full_name(self):
                 name = self.title + " " + name
             return name
 
-        return self.username
+        return self.email.split('@')[0] if self.email else "<unnamed>"
```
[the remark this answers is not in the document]
janno42: yes 👍

evap/results/tests/test_views.py
janno42: see #1475

evap/student/tests/test_views.py
janno42: will be dealt with in #1475 (if we find a way to get it working).

PR #1476, opened 2020-05-25
Only mark textanswer review as urgent on closed evaluations. Fixes #1465

### LINKED ISSUE #1465: Urgent review only for closed evaluations
Text answer review should be marked as "urgent" on the staff semester page only for evaluations where the evaluation period already ended (state `evaluated`).

He3lixxx:
Fixes #1465

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -1,5 +1,6 @@
 from collections import namedtuple, defaultdict
 from datetime import datetime, date, timedelta
+from enum import Enum, auto
 import logging
 import secrets
 import uuid
@@ -393,6 +394,13 @@ class Evaluation(models.Model):
     # whether to wait for grade uploading before publishing results
     wait_for_grade_upload_before_publishing = models.BooleanField(verbose_name=_("wait for grade upload before publishing"), default=True)
 
+    class TextAnswerReviewState(Enum):
+        do_not_call_in_templates = True
+        NO_TEXTANSWERS = auto()
+        REVIEW_NEEDED = auto()
+        REVIEW_URGENT = auto()
+        REVIEWED = auto()
+
     class Meta:
         unique_together = (
             ('course', 'name_de'),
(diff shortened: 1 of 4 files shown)


evap/evaluation/models.py
```diff
@@ -700,6 +700,13 @@ def reviewed_textanswer_set(self):
     def num_reviewed_textanswers(self):
         return self.reviewed_textanswer_set.count()
 
+    @property
+    def text_answer_review_is_urgent(self):
+        if self.state != "evaluated":
+            return False
+
+        return self.course.final_grade_documents or self.course.gets_no_grade_documents or not self.wait_for_grade_upload_before_publishing
```
He3lixxx, line 708: Should I add some linebreaks?

evap/staff/tests/test_views.py
```diff
@@ -465,6 +465,28 @@ def test_badge_for_external_responsibles(self):
         response = self.app.get(self.url, user="manager@institution.example.com")
         self.assertContains(response, 'External responsible')
 
+    def test_text_answer_review_urgent(self):
```
He3lixxx, line 468: Any ideas for better names that will not be more than 120 chars? Technically, this only tests that when the evluation is still in progress, the urgent-indicator is not shown, but when it is finished, it is...

evap/evaluation/models.py
pixunil: I would prefer them :)
Also, while you are at breaking up the line, could you experiment with replacing the `if` with a `self.state == "evaluated" and …`? But I also didn’t come up with a nice solution

evap/staff/tests/test_views.py
pixunil: Actually, I’d expect the tests to look like this:

- In `test_models`, different evaluations would be created and tested on their `text_answer_review_is_urgent` property (may be also other properties, as those are more “scenario” tests)
- In `test_views`, you mock the `text_answer_review_is_urgent` property (which then could look like your test, but with different evaluation setup)

If you do not want to split the tests (which is really okay here, it is a rather small property), I would like a comment that the evaluation first is considered to be “not urgent” and after ending it to be “urgent”. Then the intend is clear and the name fits imo.

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -43,7 +43,7 @@
         <span class="fas fa-comment icon-gray" data-toggle="tooltip" data-placement="top" title="{% trans 'No text answers available' %}"></span>
     </td>
 {% elif evaluation.num_textanswers > evaluation.num_reviewed_textanswers %}
```
pixunil, line 45: The predicate in the template is semantically connected to `text_answer_review_is_urgent` – as far as I understand, it currently can be true in cases where there are no text answers.
How do you feel about adding some more properties to reflect the text answer review state? Or change the property to return an enum `{NO_TEXTANSWERS, REVIEW_NEEDED, REVIEW_URGENT, REVIEWED}`? This would also remove more logic from the template (yay)

evap/evaluation/models.py
He3lixxx: I thought about it a bit, and for me, using the guard-clause is easier to read than the big boolean expression. I'll assume leaving that as it is now is fine for you(?)

evap/staff/templates/staff_semester_view_evaluation.html
He3lixxx: Done. However, its 100 lines more code (as the testing is quite intensive now). I don't know, I'm still unsure about where the perfect balance between testing and productive code is - I think this one is too far on the testing side now...

PR #1477, opened 2020-05-25
Tests: Use UserProfile instances in self.page.get and post. Minor styling.

### LINKED ISSUE #1475: Use UserProfile objects as arguments in tests
In tests we should consistently use UserProfile objects as arguments for get and post calls instead of email addresses where possible.
See this comment by @He3lixxx in https://github.com/e-valuation/EvaP/pull/1474:

Here, it would be much easier if we changed line 532 to
```python
cls.responsible = baker.make(UserProfile, email='responsible@institution.example.com')
```
so we can use
```python
page = self.app.get(url, user=self.responsible)
```
here.

It seems like we have this case a few times: The UserProfile instance is available or can easily be made available, yet we still use the username, which is rather unstable in my opinion.

He3lixxx:
Fixes #1475.

I found it intriguing how often we had that same `baker.make` call for creating a manager. I considered making a baker recipe to shorten these calls, but I wasn't quite happy with how they work, so I decided to go for a helper function. I'm open for suggestions here, though.

Also, I came along a lot of very long lines (like >160 chars) during this, so I chopped some of them down. I think it significantly improved readability at these places, but I'm also open for suggestions here. Our tests kind of suffer from the fact that we often have to work through very verbose hierarchies to express ourselves.

### DIFF
--- a/evap/contributor/tests/test_views.py
+++ b/evap/contributor/tests/test_views.py
@@ -28,7 +28,11 @@ class TestContributorDirectDelegationView(WebTest):
 
     def test_direct_delegation_request(self):
         data = {"delegate_to": self.non_editor.id}
-        page = self.app.post('/contributor/evaluation/{}/direct_delegation'.format(self.evaluation.id), params=data, user=self.editor).follow()
+        page = self.app.post(
+            f'/contributor/evaluation/{self.evaluation.id}/direct_delegation',
+            params=data,
+            user=self.editor
+        ).follow()
 
         self.assertContains(
             page,
@@ -50,7 +54,11 @@ class TestContributorDirectDelegationView(WebTest):
         old_contribution_count = Contribution.objects.count()
 
         data = {"delegate_to": self.non_editor.id}
-        page = self.app.post('/contributor/evaluation/{}/direct_delegation'.format(self.evaluation.id), params=data, user=self.editor).follow()
+        page = self.app.post(
+            f'/contributor/evaluation/{self.evaluation.id}/direct_delegation',
+            params=data,
+            user=self.editor
+        ).follow()
 
         self.assertContains(
             page,
@@ -79,16 +87,18 @@ class TestContributorSettingsView(WebTest):
 
     @classmethod
     def setUpTestData(cls):
-        create_evaluation_with_responsible_and_editor()
+        result = create_evaluation_with_responsible_and_editor()
+        cls.responsible = result['responsible']
 
     def test_save_settings(self):
         user = baker.make(UserProfile)
-        page = self.app.get(self.url, user="responsible@institution.example.com", status=200)
+        page = self.app.get(self.url, user=self.responsible, status=200)
         form = page.forms["settings-form"]
         form["delegates"] = [user.pk]
         form.submit()
 
-        self.assertEqual(list(UserProfile.objects.get(email='responsible@institution.example.com').delegates.all()), [user])
+        self.responsible.refresh_from_db()
+        self.assertEqual(list(self.responsible.delegates.all()), [user])
 
 
 class TestContributorEvaluationView(WebTestWith200Check):
@@ -97,7 +107,9 @@ class TestContributorEvaluationView(WebTestWith200Check):
 
     @classmethod
     def setUpTestData(cls):
-        create_evaluation_with_responsible_and_editor(evaluation_id=TESTING_EVALUATION_ID)
+        result = create_evaluation_with_responsible_and_editor(evaluation_id=TESTING_EVALUATION_ID)
+        cls.responsible = result['responsible']
+        cls.editor = result['editor']
 
     def setUp(self):
         self.evaluation = Evaluation.objects.get(pk=TESTING_EVALUATION_ID)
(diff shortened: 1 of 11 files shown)


evap/evaluation/tests/test_misc.py
```diff
@@ -20,7 +20,7 @@ class SampleXlsTests(WebTest):
     @classmethod
     def setUpTestData(cls):
         cls.semester = baker.make(Semester)
-        baker.make(UserProfile, email="user@institution.example.com", groups=[Group.objects.get(name="Manager")])
+        cls.user = baker.make(UserProfile, email="user@institution.example.com", groups=[Group.objects.get(name="Manager")])
```
He3lixxx, line 23: could this be a `make_manager` too?

evap/results/tests/test_views.py
```diff
@@ -341,31 +358,30 @@ def test_private_evaluation(self):
...
-        self.app.get(url, user="contributor@institution.example.com", status=200)
-        self.app.get(url, user="manager@institution.example.com", status=200)
-        self.app.get(url, user="student_external@example.com", status=200)  # this external user participates in the evaluation and can see the results
+        self.app.get(url, user=random, status=403)
+        self.app.get(url, user=student, status=200)
+        self.app.get(url, user=responsible, status=200)
+        self.app.get(url, user=responsible_contributor, status=200)
+        self.app.get(url, user=contributor, status=200)
+        self.app.get(url, user=manager, status=200)
+        self.app.get(url, user=student, status=200)  # this external user participates in the evaluation and can see the results
```
He3lixxx, line 376: should be student_external

evap/staff/tests/test_views.py
```diff
@@ -100,22 +100,28 @@ def test_num_queries_is_constant(self):
         semester = baker.make(Semester, participations_are_archived=True)
 
         # this triggers more checks in UserProfile.can_be_deleted_by_manager
-        evaluation = baker.make(Evaluation, state="published", course=baker.make(Course, semester=semester), _participant_count=1, _voter_count=1)
+        evaluation = baker.make(
+            Evaluation,
+            state="published",
+            course=baker.make(Course, semester=semester),
```
He3lixxx, line 106: course__semester=semester

evap/staff/tests/test_views.py
```diff
@@ -2300,14 +2450,18 @@ class TestSemesterActiveStateBehaviour(WebTest):
     csrf_checks = False
 
     def test_make_other_semester_active(self):
-        baker.make(UserProfile, email='manager@institution.example.com', groups=[Group.objects.get(name='Manager')])
+        manager = baker.make(
```
He3lixxx, line 2453: make_manager

evap/contributor/tests/test_views.py
```diff
@@ -86,21 +96,23 @@ class TestContributorEvaluationView(WebTestWith200Check):
 
     @classmethod
     def setUpTestData(cls):
-        create_evaluation_with_responsible_and_editor(evaluation_id=TESTING_COURSE_ID)
+        result = create_evaluation_with_responsible_and_editor(evaluation_id=TESTING_COURSE_ID)
+        cls.responsible = result['responsible']
+        cls.editor = result['editor']
```
pixunil, line 101: `WebTestWith200Check` still works with the email addresses. I would remove `test_users` as class attribute and create it here in `setUpTestData`, like:
```python
cls.test_users = [cls.responsible, cls.editor]
```
I think this makes it clearer which users get tested by the simple 200 check

evap/rewards/tests/test_views.py
```diff
@@ -84,7 +84,7 @@ class TestEventsView(WebTestWith200Check):
 
     @classmethod
     def setUpTestData(cls):
-        baker.make(UserProfile, email='manager@institution.example.com', groups=[Group.objects.get(name='Manager')])
+        make_manager()
```
pixunil, line 87: In this case, I wish that this user gets assigned somewhere – else I first think that it is unused and don’t know where it is actually needed. Similar to my prior comment,
```python
cls.test_users = [make_manager()]
```
(And of course in all other subclasses of `WebTestWith200Check` where this is sensible)

evap/staff/tests/test_views.py
```diff
@@ -925,19 +976,24 @@ class TestLoginKeyExportView(WebTest):
...
         cls.internal_user = baker.make(UserProfile, email="user@institution.example.com")
 
         semester = baker.make(Semester, pk=1)
-        baker.make(Evaluation, pk=1, course=baker.make(Course, semester=semester), participants=[cls.external_user, cls.internal_user], voters=[cls.external_user, cls.internal_user])
+        baker.make(
+            Evaluation,
+            pk=1,
+            course__semester=semester,
+            participants=[cls.external_user, cls.internal_user],
+            voters=[cls.external_user, cls.internal_user]
```
pixunil, line 989:
```suggestion
            voters=[cls.external_user, cls.internal_user],
```
(Trailing comma is also missing in more occurrences)

evap/results/tests/test_views.py
```diff
@@ -335,15 +348,15 @@ class TestResultsSemesterEvaluationDetailViewPrivateEvaluation(WebTest):
...
+        manager = make_manager()
         student = baker.make(UserProfile, email="student@institution.example.com")
         student_external = baker.make(UserProfile, email="student_external@example.com")
         contributor = baker.make(UserProfile, email="contributor@institution.example.com")
         responsible = baker.make(UserProfile, email="responsible@institution.example.com")
         responsible_contributor = baker.make(UserProfile, email="responsible_contributor@institution.example.com")
         test1 = baker.make(UserProfile, email="test1@institution.example.com")
         test2 = baker.make(UserProfile, email="test2@institution.example.com")
-        baker.make(UserProfile, email="random@institution.example.com")
+        uninitialized = baker.make(UserProfile, email="uninitialized@institution.example.com")
```
pixunil, line 359: I got some questions about these names:

- `responsible_contributor` is not a responsible, it is an editor
- `test1` and `test2` are more like extra students imo
- Neither `uninitialized` nor `random` make a good naming for a non-participant imo. How about `third_party`, `other` or `non_participant` (maybe with `_user` suffix)?

PR #1478, opened 2020-05-26
Replace boolean Contribution.can_edit with Contribution.Role enum

pixunil:
- Makes the conversion in `ContributionForm` from `can_edit` to `responsibility` obsolete
- Removes the confusing “responsiblity” terminology (as `can_edit` or `role` isn’t about responsibles at all)
- Will remove one weird thing in the `ContributionCopyForm` of #1472
- Split some lines >120 characters

As the diff grew longer than expected, this got its own PR :)


evap/evaluation/migrations/0116_remove_contributon_can_edit.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('evaluation', '0115_add_contribution_role'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
janno42, line 11: could be added to the previous migration but like this it's fine for me as well

evap/evaluation/templates/choice_button.html
```diff
@@ -2,10 +2,10 @@
 
 <label class="btn btn-sm btn-option
     {% if formelement.field.disabled or disabled %} disabled{% endif %}
-    {% if formelement.value == choice.data.value %} active{% endif %}
+    {% if choice.data.selected %} active{% endif %}
     {% if formelement.errors %} choice-error{% endif %}"
-    name="{{ choice.data.name }}" data-toggle="tooltip" data-placement="top" title="{{ tooltip }}">
```
janno42, line 7: why do you remove the name?

evap/evaluation/migrations/0116_remove_contributon_can_edit.py
pixunil: Sadly, no. I get an error like this when trying to put everything in a single migration:
```
django.db.utils.OperationalError: cannot ALTER TABLE "evaluation_contribution" because it has pending trigger events

evap/evaluation/templates/choice_button.html
pixunil: It was part of a `<label>` and a `name` attribute wasn’t sensible for me. The `name` is on the contained checkbox and the reference between label and checkbox is provided through nesting (in other cases, one would use `id` and `for`, but not `name`)

janno42: i was just wondering why it was there in the first place but it can probably removed.

PR #1484, opened 2020-06-19
Handling locked questionnaires

### LINKED ISSUE #1479: Locked questionnaires failing in editor form
#1445 introduced locked questionnaires. However, they are not dealt with correctly in the evaluation editor form. When initially opening the form, the locked questionnaires are correctly selected but are not handled correctly when saving the form.

Steps to reproduce:
1. As manager, assign a locked questionnaire as the only general questionnaire for an evaluation.
2. Enable the evaluation for editor review.
3. As editor, open the evaluation form and try to save it. Saving will fail with an error for the field "General questionnaires" ("This field is required.").

The locked questionnaire should count as a selected questionnaire and the form should be saved.
A test should be added for this use case.

janno42:
fixes #1479

This fixes the problems of handling locked questionnaires in editor forms.
What kinds of tests do we need? @pixunil do you want to add yours?

### DIFF
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -15,7 +15,7 @@ logger = logging.getLogger(__name__)
 
 
 class EvaluationForm(forms.ModelForm):
-    general_questionnaires = forms.ModelMultipleChoiceField(queryset=None, widget=CheckboxSelectMultiple, label=_("General questionnaires"))
+    general_questionnaires = forms.ModelMultipleChoiceField(queryset=None, required=False, widget=CheckboxSelectMultiple, label=_("General questionnaires"))
     course = forms.ModelChoiceField(Course.objects.all(), disabled=True, required=False, widget=forms.HiddenInput())
     name_de_field = forms.CharField(label=_("Name (German)"), disabled=True, required=False)
     name_en_field = forms.CharField(label=_("Name (English)"), disabled=True, required=False)
@@ -69,6 +69,9 @@ class EvaluationForm(forms.ModelForm):
             not_locked = list(self.cleaned_data.get('general_questionnaires').filter(is_locked=False))
 
         locked = list(self.instance.general_contribution.questionnaires.filter(is_locked=True))
+
+        if not not_locked + locked:
+            self.add_error("general_questionnaires", _("At least one questionnaire must be selected."))
 
         return not_locked + locked
 
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -64,10 +64,11 @@ class EvaluationForm(forms.ModelForm):
 
     def clean_general_questionnaires(self):
         # Ensure all locked questionnaires still have the same status (included or not)
-        locked_qs = self.fields['general_questionnaires'].queryset.filter(is_locked=True)
+        not_locked = []
+        if self.cleaned_data.get('general_questionnaires'):
+            not_locked = list(self.cleaned_data.get('general_questionnaires').filter(is_locked=False))
 
-        not_locked = [q for q in self.cleaned_data.get('general_questionnaires') if q not in locked_qs]
-        locked = [q.pk for q in self.instance.general_contribution.questionnaires.filter(is_locked=True)]
+        locked = list(self.instance.general_contribution.questionnaires.filter(is_locked=True))
 
         return not_locked + locked
 
--- a/evap/contributor/tests/test_views.py
+++ b/evap/contributor/tests/test_views.py
@@ -1,16 +1,12 @@
-from unittest.mock import patch
-import operator
-
 from django.core import mail
 
 from django_webtest import WebTest
 from model_bakery import baker
-import webtest
 
 from evap.evaluation.models import Evaluation, UserProfile, Contribution, Questionnaire, Course
 from evap.evaluation.tests.tools import WebTestWith200Check, create_evaluation_with_responsible_and_editor
 
-TESTING_COURSE_ID = 2
+TESTING_EVALUATION_ID = 2
 
 
 class TestContributorDirectDelegationView(WebTest):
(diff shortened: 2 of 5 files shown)


evap/evaluation/templates/questionnaires_widget.html
```diff
@@ -27,7 +27,7 @@
             </ul>
             {% endspaceless %}">
             <input class="form-check-input" id="{{ choice.id_for_label }}" name="{{ choice.data.name }}" type="checkbox" value="{{ choice.data.value }}"
-                autocomplete="off"{% if choice.data.selected %} checked{% endif %}
+                autocomplete="off"{% if choice.data.selected or questionnaire.is_locked and not manager and choice.data.value in field.initial %} checked{% endif %}
```
pixunil, line 30: This change is necessary to keep locked questionnaires checked even when the form was reloaded with errors, right?
I do not know whether it is possible, but could the `choice.data.selected` part be modified by the form? Thinking this further, is it possible to include a `choice.data.is_locked` attribute (or similar)? Just thinking loud, this isn’t blocking

evap/contributor/forms.py
```diff
@@ -66,8 +69,11 @@ def clean_general_questionnaires(self):
         # Ensure all locked questionnaires still have the same status (included or not)
         locked_qs = self.fields['general_questionnaires'].queryset.filter(is_locked=True)
 
-        not_locked = [q for q in self.cleaned_data.get('general_questionnaires') if q not in locked_qs]
-        locked = [q.pk for q in self.instance.general_contribution.questionnaires.filter(is_locked=True)]
+        not_locked = []
+        if self.cleaned_data.get('general_questionnaires'):
+            not_locked = [q for q in self.cleaned_data.get('general_questionnaires') if q not in locked_qs]
```
pixunil, line 74: We now know that locked questionnaires as disabled fields aren’t submitted, and I think explicitly filtering out locked questionnaires is a good practise in the sense of defensive programming. But, I think a comment here stating that locked questionnaires aren’t submitted in a normal case would help future people.

Also, the expression could be:
```suggestion
            not_locked = list(self.cleaned_data.get('general_questionnaires').filter(is_locked=False))
```

evap/contributor/tests/test_views.py
```diff
@@ -179,6 +228,30 @@ def test_contributor_evaluation_edit(self):
...
+            state='prepared',
+        )
+        evaluation.general_contribution.questionnaires.set([locked_questionnaire])
+
+        page = self.app.get(f'/contributor/evaluation/{evaluation.pk}/edit', user=responsible, status=200)
+        form = page.forms['evaluation-form']
+        response = form.submit(name='operation', value='save')
+        # TODO: Somehow assert that the form was submitted successfully.
+        # follow works in this case, but does not quite hit the point
+        response.follow()
```
pixunil, line 252: I added my regression test which currently has two flaws:

1. The assertion is a bit weird, but maybe you can come up with something nicer.
2. As webtest does wrongly submit disabled fields, I had to add a helper method.

(Also: sorry for breaking your build)

evap/contributor/forms.py
```diff
@@ -53,6 +53,9 @@ def clean(self):
                 self.add_error("vote_start_datetime", "")
                 self.add_error("vote_end_date", _("The first day of evaluation must be before the last one."))
 
+        if not self.cleaned_data.get('general_questionnaires'):
+            self.add_error("general_questionnaires", _("At least one questionnaire must be selected."))
+
```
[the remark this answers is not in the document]
janno42: true. that's based on some thought i had in between and then didn't move back where it belongs. i now moved it into `clean_general_questionnaires`.

evap/evaluation/templates/questionnaires_widget.html
janno42: right, this is needed so that locked questionnaires are checked on form errors, your thoughts are interesting but i'd prefer you investigating them later on :)

evap/contributor/forms.py
janno42: i honestly don't know where to put what comment exactly so that it doesn't feel out of place.

the expression was changed.

evap/contributor/tests/test_views.py
```diff
@@ -122,6 +126,51 @@ def test_wrong_state(self):
         self.app.get(self.url, user="responsible@institution.example.com", status=403)
 
 
+# A submit function which conforms to the specification to not submit disabled fields.
+# See https://github.com/Pylons/webtest/issues/138
+def submit_fields_without_disabled(self, name=None, index=None, submit_value=None):
+    submit = []
+    # Use another name here so we can keep function param the same for BWC.
```
janno42, line 133: bwc?

evap/contributor/tests/test_views.py
```diff
@@ -122,6 +126,51 @@ def test_wrong_state(self):
         self.app.get(self.url, user="responsible@institution.example.com", status=403)
 
 
+# A submit function which conforms to the specification to not submit disabled fields.
+# See https://github.com/Pylons/webtest/issues/138
+def submit_fields_without_disabled(self, name=None, index=None, submit_value=None):
```
janno42, line 131: did you write this yourself?

evap/contributor/tests/test_views.py
janno42: i added an assert for the success message after saving as we usually do

evap/contributor/forms.py
pixunil: Well, then no reason to force it :)

evap/contributor/tests/test_views.py
pixunil: No, I copied it from webtest’s sources `Form.submit_fields` and changed one line:

```patch
- if value is None:
+ if value is None or 'disabled' in field.attrs:
      continue
```

On second thought, this solution doesn’t really make anybody happy (codecov also flags some uncovered lines). During research, I stumpled across https://github.com/Pylons/webtest/issues/138#issuecomment-532760520 and would prefer this approach now – now we’re doing webtest’s job in the test, but this also gives us an opportunity to leave a comment about disabled fields :)

janno42: just removing the `disabled` attribute sounds fine for me

PR #1485, opened 2020-06-22
Show participations and votes on user merge

### LINKED ISSUE #1480: User merge does not show participations and votes
When merging two users, the UI does not show the merged lists for participations and votes. Currently the respective cell is empty.

pixunil:
Resolves #1480 

This was properly introduced in 8a2eaf6dd6961a1f56711f5896b0014da1fa2b88, where the lines were changed to have a consistent ordering.
But, I think it is unnecessary to explicitely sort here, as the `merged_user` is a `dict` which `courses_responsible_for`, `contributions`, `evaluations_participating_in`, `evaluations_voted_for` are sorted according to the semester: https://github.com/e-valuation/EvaP/blob/56c307a7347861ed3bbfd4327be28fca152bc4c8/evap/staff/tools.py#L238-L241

To check that the regression is caught, run:
```
git checkout master
git checkout user-merge-missing-participation evap/staff/tests/test_views.py
./manage.py test
```

But I might overlook something, pinging @karyon for this case :)

### DIFF
--- a/evap/staff/templates/staff_user_merge.html
+++ b/evap/staff/templates/staff_user_merge.html
@@ -141,15 +141,15 @@
                     <tr{% if 'courses_responsible_for' in errors %} class="table-danger"{% endif %}>
                         <td><strong>{% trans 'Courses responsible for' %}</strong></td>
                         {% regroup main_user.get_sorted_courses_responsible_for by semester as course_list %}
-                        <td{% if course_list|length > 0 %} class="info"{% endif %}>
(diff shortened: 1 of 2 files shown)


He3lixxx:
I think we'd want some kind of test on this (we had always-empty and no test detected it?)

janno42:
Yes, a small test (e.g., checking that a certain evaluation is on the page twice (one time for the main user, another time for the merged user) would be nice.

PR #1492, opened 2020-07-06
Fix contributor publish notification missing for responsibles.

### LINKED ISSUE #1483: Send publish notification to responsible
When sending publish notifications in `send_contributor_publish_notifications` they are currently not sent to the person responsible for the course if the average grade was published (only in the `elif` part the responsible is included). The responsible person should get a notification there as well.

He3lixxx:
- Fix contributor publish notification missing for responsibles. Fixes #1483 
- Add unit test for the method - was only indirectly tested through view tests before

I guess it would be best to also split the testing into two parts here: In the view tests, make sure that `send_contributor_publish_notifications` and `send_participant_publish_notifications` are called correctly, and move the assertions for the mail outbox sizes into the `EmailTemplateTest` class. However, currently, the view test nicely specifies our expected behaviour, we wouldn't have that then anymore. @pixunil what do you think?

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -1599,6 +1599,15 @@ class EmailTemplate(models.Model):
 
         evaluations_per_contributor = defaultdict(set)
         for evaluation in evaluations:
+            # an average grade is published or a general text answer exists
+            relevant_information_published_for_responsibles = (
+                evaluation.can_publish_average_grade
+                or evaluation.textanswer_set.filter(contribution=evaluation.general_contribution).exists()
+            )
+            if relevant_information_published_for_responsibles:
+                for responsible in evaluation.course.responsibles.all():
+                    evaluations_per_contributor[responsible].add(evaluation)
+
             # for evaluations with published averaged grade, all contributors get a notification
             # we don't send a notification if the significance threshold isn't met
             if evaluation.can_publish_average_grade:
@@ -1611,9 +1620,6 @@ class EmailTemplate(models.Model):
                 for textanswer in evaluation.textanswer_set:
                     if textanswer.contribution.contributor:
                         evaluations_per_contributor[textanswer.contribution.contributor].add(evaluation)
-
-                for contributor in evaluation.course.responsibles.all():
-                    evaluations_per_contributor[contributor].add(evaluation)
 
         for contributor, evaluation_set in evaluations_per_contributor.items():
             body_params = {'user': contributor, 'evaluations': list(evaluation_set)}
--- a/evap/evaluation/tests/test_models.py
+++ b/evap/evaluation/tests/test_models.py
@@ -1,5 +1,5 @@
 from datetime import datetime, timedelta, date
-from unittest.mock import patch, Mock
+from unittest.mock import patch, Mock, call
 
 from django.contrib.auth.models import Group
 from django.core.exceptions import ValidationError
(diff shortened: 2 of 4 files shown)


evap/staff/tests/test_views.py
```diff
@@ -1101,21 +1101,27 @@ def test_publish_notifications(self):
...
         self.helper_publish_evaluation_with_publish_notifications_for(evaluation, contributors=True, participants=True)
-        self.assertEqual(len(mail.outbox), 3)
-        self.assertIn([participant1.email], [mail.outbox[0].to, mail.outbox[1].to, mail.outbox[2].to])
-        self.assertIn([participant2.email], [mail.outbox[0].to, mail.outbox[1].to, mail.outbox[2].to])
-        self.assertIn([contributor1.email], [mail.outbox[0].to, mail.outbox[1].to, mail.outbox[2].to])
+        self.assertEqual(len(mail.outbox), 4)
+        self.assertEqual(
+            {recipient for msg in mail.outbox for recipient in msg.to},
+            {participant1.email, participant2.email, contributor1.email, self.responsible.email}
+        )
```
He3lixxx, line 1124: On a second thought: The sets will remove duplicates. Thus, the new code will not catch if we are sending emails way too often or add the wrong people as recipients. The old assertions were probably better.

evap/evaluation/models.py
```diff
@@ -1599,6 +1599,10 @@ def send_contributor_publish_notifications(cls, evaluations, template=None):
 
         evaluations_per_contributor = defaultdict(set)
         for evaluation in evaluations:
+            # all responsibles unconditionally get a notification
```
janno42, line 1602: we shouldn't send an email if no results are published, only in the if and elif cases below. i'm sorry that the issue description is not precise enough.

evap/staff/tests/test_views.py
```diff
@@ -1101,8 +1101,9 @@ def test_publish_notifications(self):
         mail.outbox = []
 
         self.helper_publish_evaluation_with_publish_notifications_for(evaluation, contributors=True, participants=False)
-        self.assertEqual(len(mail.outbox), 1)
-        self.assertEqual(mail.outbox[0].to, [contributor1.email])
+        self.assertEqual(len(mail.outbox), 2)
+        self.assertIn([contributor1.email], [mail.outbox[0].to, mail.outbox[1].to])
+        self.assertIn([self.responsible.email], [mail.outbox[0].to, mail.outbox[1].to])
```
pixunil, line 1106: Could these three lines become something like this?
I’d prefer `assertCountEqual` over multiple asserts as it doesn’t stop at the first error and gives a nice summary, while keeping the bag-of-lists-semantic.

```python
self.assertCountEqual([[contributor1.email], [self.responsible.email]], [email.to for email in mail.outbox])

File: evap/staff/tests/test_views.py [hunk 4]
```diff
@@ -1136,11 +1138,20 @@ def helper_semester_state_views(self, evaluation, old_state, new_state):
...
+            voters=[participant1, participant2]
+        )
 
         self.helper_semester_state_views(evaluation, "reviewed", "published")
-        self.assertEqual(len(mail.outbox), 2)
+        self.assertEqual(len(mail.outbox), 3)
+        self.assertEqual(
+            {recipient for msg in mail.outbox for recipient in msg.to},
+            {participant1.email, participant2.email, self.responsible.email}
+        )
```
pixunil, line 1154: This doesn’t keep the bag-of-lists-semantic – we could accidentally send an email twice to an user. Is this intended?

evap/evaluation/tests/test_models.py
```diff
@@ -719,6 +719,71 @@ def test_disable_cc(self):
...
+            evaluation=evaluation2,
+            contributor=contributor_both,
+            role=Contribution.Role.CONTRIBUTOR
+        )
+        contributor2_contribution = baker.make(
+            Contribution,
+            evaluation=evaluation2,
+            contributor=contributor2,
+            role=Contribution.Role.CONTRIBUTOR
+        )
```
pixunil, line 760: These lines are formatted differently than those for `evaluation1` and I can’t spot the difference :)

evap/evaluation/tests/test_models.py
```diff
@@ -719,6 +719,71 @@ def test_disable_cc(self):
...
+            EmailTemplate.send_contributor_publish_notifications([evaluation1, evaluation2])
+            send_to_user_mock.assert_has_calls(expected_calls, any_order=True)
+
+        # if general textanswers for an evaluation exist, all responsibles should also be notified
+        baker.make(TextAnswer, contribution=evaluation2.general_contribution)
+        expected_calls.append(call(responsible2, {}, {'user': responsible2, 'evaluations': [evaluation2]}, use_cc=True))
+
+        with patch('evap.evaluation.models.EmailTemplate.send_to_user') as send_to_user_mock:
+            EmailTemplate.send_contributor_publish_notifications([evaluation1, evaluation2])
+            send_to_user_mock.assert_has_calls(expected_calls, any_order=True)
```
pixunil, line 785: So, I think I _can_ understand this test and it would be fine for me to keep it as-is (there are enough comments to explain what you’re trying to achieve). But, if you want to split this case, I’d suggest to move this part into another case.

evap/evaluation/models.py
```diff
@@ -1599,6 +1599,15 @@ def send_contributor_publish_notifications(cls, evaluations, template=None):
 
         evaluations_per_contributor = defaultdict(set)
         for evaluation in evaluations:
+            # an average grade is published or a general text answer exists
+            relevant_information_published_for_responsibles = (
```
janno42, line 1603: it seems like you're only using this once. why didn't you move it into the if?

He3lixxx: I moved it out to prevent having a multi-line condition in the if (which I always struggle to format nicely) -- with the side benefit, that the condition now has a semantic name.

evap/evaluation/tests/test_models.py
He3lixxx: image
The main reason was that the upper 4 lines stay below 120 chars per line but the bottom three exceed that length significantly due to the variables that are assigned.

If i fully spread them all out, it will be 7 * 6 = 42 lines and I wanted to prevent that.

What do you think? Spread them all out?

evap/evaluation/tests/test_models.py
He3lixxx: I think duplicating the setup code is not a good option, so we'd have to refactor the 36 lines of set up out into a different method, but I struggle naming that as it is a very specific scenario that is being set up. For the second method, we would need exactly the same set up with that one additional TextAnswer.

Did you have a specific refactoring in mind that you thought would work quite well?

evap/staff/tests/test_views.py
He3lixxx: nice, wasn't aware of that one.

evap/staff/tests/test_views.py
He3lixxx: Nope, wasn't intended, good catch

evap/evaluation/tests/test_models.py
He3lixxx: I solved it by adding two helpers, `make_contributor` and `make_editor` that I think are usable in a lot of other places too.

janno42 (approved):

evap/evaluation/tests/test_models.py
He3lixxx: (We wanted to finally get the bug fix done, thus we already merged the current state. I'm still open to refactoring the code, so feel free to reply to this discussion later)

PR #1494, opened 2020-07-19
Introduce staff mode

### LINKED ISSUE #1109: Switchable staff mode

PFischbeck:
This pull request adds a new staff mode. More specifically, all managers see the following navbar.
manager_off
Once they enter staff mode (active for 1 hour), the navbar looks like this:
manager_on

Pretty much all "restrictions" are client-side, since this is not truly security-relevant. However, I tried to fix things like the default index and the full results view if not in staff mode. Please check whether this behaves as expected for different types of users (managers, students, reviewers, grade publishers...).

I had to work on the navbar cache, since the new button uses CSRF, which doesn't play nicely with caching...
This fixes #1109.

### DIFF
--- a/evap/evaluation/auth.py
+++ b/evap/evaluation/auth.py
@@ -66,6 +66,15 @@ def internal_required(view_func):
     return user_passes_test(check_user)(view_func)
 
 
+def staff_permission_required(view_func):
+    """
+    Decorator for views that checks that the user is logged in and staff (regardless of staff mode!)
+    """
+    def check_user(user):
+        return user.has_staff_permission
+    return user_passes_test(check_user)(view_func)
+
+
 def manager_required(view_func):
     """
     Decorator for views that checks that the user is logged in and a manager
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -1360,6 +1360,11 @@ class UserProfile(AbstractBaseUser, PermissionsMixin):
     def is_staff(self):
         return self.is_manager or self.is_reviewer
 
+    # Required for staff mode to work, since several other cached properties (including is_staff) are overwritten
+    @property
+    def has_staff_permission(self):
+        return self.groups.filter(name='Manager').exists() or self.groups.filter(name='Reviewer').exists()
+
     @cached_property
     def is_manager(self):
         return self.groups.filter(name='Manager').exists()
@@ -1394,11 +1399,11 @@ class UserProfile(AbstractBaseUser, PermissionsMixin):
             return False
         return True
 
-    @property
+    @cached_property
     def is_participant(self):
         return self.evaluations_participating_in.exists()
 
-    @property
+    @cached_property
     def is_student(self):
         """
             A UserProfile is not considered to be a student anymore if the
@@ -1415,23 +1420,23 @@ class UserProfile(AbstractBaseUser, PermissionsMixin):
 
         return last_semester_participated.created_at >= last_semester_contributed.created_at
 
-    @property
+    @cached_property
     def is_contributor(self):
         return self.contributions.exists()
 
-    @property
+    @cached_property
     def is_editor(self):
         return self.contributions.filter(role=Contribution.Role.EDITOR).exists() or self.is_responsible
 
-    @property
+    @cached_property
     def is_responsible(self):
         return self.courses_responsible_for.exists()
 
-    @property
+    @cached_property
     def is_delegate(self):
         return self.represented_users.exists()
 
-    @property
+    @cached_property
     def is_editor_or_delegate(self):
         return self.is_editor or self.is_delegate
 
--- a/evap/evaluation/templates/navbar.html
+++ b/evap/evaluation/templates/navbar.html
@@ -21,39 +21,14 @@
                 {% if user|can_reward_points_be_used_by %}
                     <li class="nav-item"><a class="nav-link" href="{% url 'rewards:index' %}">{% trans 'Rewards' %}</a></li>
                 {% endif %}
-                {% if user.is_grade_publisher and not user.is_manager %}
-                    <li class="nav-item dropdown">
-                        <a class="nav-link dropdown-toggle" href="{% url 'grades:index' %}" id="navbarPublishDropdownMenuLink" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">{% trans 'Publish grades' %}</a>
-                        <div class="dropdown-menu" aria-labelledby="navbarPublishDropdownMenuLink">
-                            <a class="dropdown-item" href="{% url 'grades:index' %}">{% trans 'Semesters' %}</a>
-                            {% for semester in grade_document_semesters %}
-                                <a class="dropdown-item dropdown-item-indent" href="{% url 'grades:semester_view' semester.id %}">{{ semester.name }}</a>
-                            {% endfor %}
-                        </div>
-                    </li>
-                {% endif %}
-                {% if not user.is_external %}
-                    <li class="nav-item"><a class="nav-link" href="{% url 'results:index' %}">{% trans 'Results' %}</a></li>
-                {% endif %}
                 {% if user.is_manager %}
+                    <li class="nav-item"><a class="nav-link" href="{% url 'staff:index' %}">{% trans 'Overview' %}</a></li>
                     <li class="nav-item dropdown">
-                        <a class="nav-link dropdown-toggle" href="{% url 'staff:index' %}" id="navbarStaffDropdownMenuLink" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">{% trans 'Manage' %}</a>
-                        <div class="dropdown-menu" aria-labelledby="navbarStaffDropdownMenuLink">
-                            <a class="dropdown-item" href="{% url 'staff:index' %}">{% trans 'Overview' %}</a>
-                            <a class="dropdown-item" href="{% url 'staff:index' %}">{% trans 'Semesters' %}</a>
+                        <a class="nav-link dropdown-toggle" href="#" id="navbarSemestersDropdownMenuLink" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">{% trans 'Semesters' %}</a>
+                        <div class="dropdown-menu" aria-labelledby="navbarSemestersDropdownMenuLink">
                             {% for semester in result_semesters %}
-                                <a class="dropdown-item dropdown-item-indent" href="{% url 'staff:semester_view' semester.id %}">{{ semester.name }}</a>
+                                <a class="dropdown-item" href="{% url 'staff:semester_view' semester.id %}">{{ semester.name }}</a>
                             {% endfor %}
-                            {% if user.is_grade_publisher %}
-                                <a class="dropdown-item" href="{% url 'grades:index' %}">{% trans 'Publish grades' %}</a>
-                            {% endif %}
-                            <a class="dropdown-item" href="{% url 'staff:questionnaire_index' %}">{% trans 'Questionnaires' %}</a>
-                            <a class="dropdown-item" href="{% url 'staff:user_index' %}">{% trans 'Users' %}</a>
-                            <a class="dropdown-item" href="{% url 'staff:degree_index' %}">{% trans 'Degrees' %}</a>
-                            <a class="dropdown-item" href="{% url 'staff:course_type_index' %}">{% trans 'Course types' %}</a>
-                            <a class="dropdown-item" href="{% url 'rewards:reward_point_redemption_events' %}">{% trans 'Reward point redemption events' %}</a>
-                            <a class="dropdown-item" href="{% url 'staff:index' %}">{% trans 'Templates' %}</a>
-                            <a class="dropdown-item" href="{% url 'staff:faq_index' %}">{% trans 'FAQ' %}</a>
                         </div>
                     </li>
                 {% elif user.is_reviewer %}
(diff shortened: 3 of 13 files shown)


evap/evaluation/templates/base.html
```diff
@@ -38,8 +38,8 @@
 
 {% get_current_language as LANGUAGE_CODE %}
 <div class="fixed-top d-print-none">
-    {% cache 3600 navbar request.user.email LANGUAGE_CODE %}
-        {% include_navbar user LANGUAGE_CODE %}
+    {% cache 3600 navbar request.user.email LANGUAGE_CODE in_staff_mode %}
+        {% include_navbar user LANGUAGE_CODE in_staff_mode %}
```
He3lixxx, line 42: Here you let the cache keep two versions, depending on whether the user is in staff mode or not, yet in `enter_staff_mode` and `exit_staff_mode` you delete the cache for both versions. Am I missing something? To me this looks like you mixed up two different approached here.

PFischbeck: You're right, I mixed the two different approaches. I fixed it, and now we only go with the cache deletion approach - the approach of keeping two versions in cache didn't work out due to the navbar now containing a form for the button. The cache would keep old (invalid) CSRF tokens.

evap/staff/staff_mode.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+
+    def middleware(request):
+        if is_in_staff_mode(request):
+            if time.time() <= request.session.get('staff_mode_start_time', 0) + STAFF_MODE_TIMEOUT:
+                # just refresh time
+                update_staff_mode(request)
+            else:
+                exit_staff_mode(request)
+                messages.info(request, _("Your staff mode timed out."))
```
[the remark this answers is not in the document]
PFischbeck: Good idea! The message is now only shown if less than three hours have passed since the timeout.

evap/staff/staff_mode.py
```diff
@@ -0,0 +1,54 @@
...
+
+        response = get_response(request)
+        return response
+
+    return middleware
+
+
+def is_in_staff_mode(request):
+    return request.session.get('in_staff_mode', False)
```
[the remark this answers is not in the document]
PFischbeck: Thanks for the suggestion, I applied it.

evap/results/templates/results_evaluation_detail.html
```diff
@@ -20,7 +20,7 @@
         <div class="alert alert-warning">{% trans 'This is a preview. The results have not been published yet.' %}</div>
     {% endif %}
 
-    {% if is_reviewer or is_responsible_or_contributor_or_delegate %}
+    {% if is_reviewer and in_staff_mode or is_responsible_or_contributor_or_delegate %}
```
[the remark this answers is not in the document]
PFischbeck: You're right. I fixed it so that the `view` is set to `public` if not in staff mode, which also affects `can_export_text_answer` and `view_as_user`.

evap/evaluation/templates/navbar.html
```diff
@@ -12,51 +12,66 @@
...
-                            {% if user.is_grade_publisher %}
-                                <a class="dropdown-item" href="{% url 'grades:index' %}">{% trans 'Publish grades' %}</a>
-                            {% endif %}
-                            <a class="dropdown-item" href="{% url 'staff:questionnaire_index' %}">{% trans 'Questionnaires' %}</a>
-                            <a class="dropdown-item" href="{% url 'staff:user_index' %}">{% trans 'Users' %}</a>
                             <a class="dropdown-item" href="{% url 'staff:degree_index' %}">{% trans 'Degrees' %}</a>
                             <a class="dropdown-item" href="{% url 'staff:course_type_index' %}">{% trans 'Course types' %}</a>
                             <a class="dropdown-item" href="{% url 'rewards:reward_point_redemption_events' %}">{% trans 'Reward point redemption events' %}</a>
                             <a class="dropdown-item" href="{% url 'staff:index' %}">{% trans 'Templates' %}</a>
                             <a class="dropdown-item" href="{% url 'staff:faq_index' %}">{% trans 'FAQ' %}</a>
```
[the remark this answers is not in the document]
PFischbeck: That is intended, as I thought that the staff mode should only be used for staff stuff. The results page only lists public data, and the results details can instead be found through the semesters in staff mode.

Do you think I should re-add it to the navbar?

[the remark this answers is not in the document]
PFischbeck: In the new version, the results page is in the navbar again.

PFischbeck:
I now changed to the cached_property approach you suggested, and I am quite happy about how it turned out! Now there are no weird `in_staff_mode`checks all over the place.
The only 'downsides' were that I had to introduce a non-cached property for checking the actual `is_staff`, and also I had to adjust most view tests.

evap/staff/staff_mode.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def update_staff_mode(request):
+    request.session['staff_mode_start_time'] = time.time()
+    request.session.modified = True
+
+
+def enter_staff_mode(request):
+    request.session['staff_mode_start_time'] = time.time()
+    request.session.modified = True
```
karyon, line 60: `update_staff_mode`?

evap/staff/tests/test_views.py
```diff
@@ -2454,13 +2594,17 @@ def test_review_actions(self):
         let_user_vote_for_evaluation(self.app, self.student2, self.evaluation)
 
         # now reviewing should work
+        helper_enter_staff_mode(self)
```
karyon, line 2597: this seems wrong, there was no exit after setUp

evap/staff/tests/test_views.py
```diff
@@ -18,24 +18,18 @@
...
-        except FileNotFoundError:
-            pass
-
-
 class TestDownloadSampleXlsView(WebTest):
     url = '/staff/download_sample_xls/sample.xls'
     email_placeholder = "institution.com"
 
+    def setUp(self):
+        helper_enter_staff_mode(self)
```
karyon, line 31: I was going to suggest to extend WebTest into a StaffModeWebTest, which contains just this method, but there's also WebTestWith200Check. But we could use a mixin for that, i think (I have never used them). So what about a StaffModeMixin? It doesn't change much but it's a little bit less boilerplate. Those tests that actually implemented setUp before would then need to call super().setUp(), though, and it might be unexpected when creating a new test that overriding setUp breaks staff mode. What do you think?

evap/staff/tests/test_views.py
```diff
@@ -2037,23 +2133,32 @@ def test_textanswers_showing_up(self):
...
+        helper_exit_staff_mode(self)
         let_user_vote_for_evaluation(self.app, self.student2, self.evaluation)
+        helper_enter_staff_mode(self)
         page = self.app.get(self.url, user=self.manager, status=200)
         self.assertContains(page, self.answer)
 
     def test_textanswers_full_view(self):
+        helper_exit_staff_mode(self)
         let_user_vote_for_evaluation(self.app, self.student2, self.evaluation)
+        helper_enter_staff_mode(self)
```
karyon, line 2153: a python context manager might be a good fit for this pattern. would make it a little easier to not get it wrong in the future!

evap/evaluation/templates/navbar.html
```diff
@@ -36,19 +36,29 @@
                     <li class="nav-item"><a class="nav-link" href="{% url 'results:index' %}">{% trans 'Results' %}</a></li>
                 {% endif %}
                 {% if user.is_manager %}
```
janno42, line 38: i like the new menu! it would be even better with some additional rearrangements. could you please order it as follows (renaming in **bold**):
1. Overview
2. Semesters ...
3. Results
4. Users ...
    1. **User list**
    2. Create
    3. Import
    4. Merge
    5. Update 
5. Questionnaires
6. Publish grades ...
7. More ...
    1. Course Types
    2. Degrees
    3.** Email templates**
    4. FAQ
    5. Reward point redemption events

janno42 (changes requested):

evap/context_processors.py
```diff
@@ -12,3 +14,7 @@ def slogan(request):
 
 def debug(request):
     return {'debug': settings.DEBUG}
+
+
+def staff_mode(request):
+    return {'in_staff_mode': is_in_staff_mode(request)}
```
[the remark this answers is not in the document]
PFischbeck: You're right, I replaced it :)

evap/staff/staff_mode.py
PFischbeck: :+1:

evap/staff/tests/test_views.py
PFischbeck: I fixed it (with a context manager)

evap/staff/tests/test_views.py
PFischbeck: Yes, I use a context manager now. This reads much better :)

evap/staff/tests/test_views.py
PFischbeck: I talked about this @janno42, and we decided to go for `WebTestStaffMode` and `WebTestStaffModeWith200Check`, so no mixins. The occasional `super().setUp()` calls are fine.

evap/evaluation/models.py
```diff
@@ -1299,6 +1299,10 @@ def __str__(self):
     def is_staff(self):
         return self.is_manager or self.is_reviewer
 
+    # Required for staff mode to work, since several other cached properties (including is_staff) are overwritten
+    def is_staff_uncached(self):
```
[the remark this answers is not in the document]
PFischbeck: I renamed it to `has_staff_permission` now.

PFischbeck:
This is ready for re-review :)

evap/staff/tests/test_views.py
```diff
@@ -485,8 +478,9 @@ def test_access_to_semester_with_archived_results(self):
         # reviewers shouldn't be allowed to access the semester page
         self.app.get('/staff/semester/2', user=reviewer, status=403)
 
-        # managers can access the page
```
karyon, line 488: (here and above) unnecessary run_in_staff_mode in a WebTestStaffMode?

karyon (approved):
looks great!

evap/staff/tests/test_views.py
```diff
@@ -467,12 +459,14 @@ def test_view_list_sorting(self):
...
         self.app.reset()  # language is only loaded on login, so we're forcing a re-login here
 
-        self.manager.language = 'de'
-        self.manager.save()
-        page = self.app.get(self.url, user=self.manager).body.decode("utf-8")
-        position_evaluation1 = page.find("Evaluation 1")
-        position_evaluation2 = page.find("Evaluation 2")
-        self.assertLess(position_evaluation1, position_evaluation2)
+        # Re-enter staff mode, since the session was just reset
+        with run_in_staff_mode(self):
```
karyon, line 463: here too :)

karyon:
Out of curiosity, why did you decide against the mixin?

evap/staff/tests/test_views.py
PFischbeck: This one was needed because the session behaves weird when switching users. But I mitigated it by changing the order of checks here, so this is now fixed.

evap/staff/tests/test_views.py
PFischbeck: This one is necessary, because the app (and therefore the session) is reset in the middle of the test. I added a comment in the code to explain it.

karyon: oh, right. sorry.

PFischbeck:
We will probably not introduce more types of `WebTest`, so the number of new classes is fine. Also, it feels more like a subclass of WebTest than a mixin that extends functionality. And if we had used mixins, both the class definition and `setUp` super calls would be a bit more cumbersome:
`TestSemesterView(StaffModeMixin, WebTest)` vs `TestSemesterView(WebTestStaffMode)`
`super(StaffModeMixin, self).setUp()` vs `super().setUp()`

evap/staff/tests/test_views.py
```diff
@@ -2617,3 +2620,29 @@ def test_make_other_semester_active(self):
 
         self.assertFalse(semester1.is_active)
         self.assertTrue(semester2.is_active)
+
+
+class TestStaffMode(WebTest):
```
He3lixxx, line 2625: The `enter_staff_mode` view is has the correct access control decorator, and the middleware is written very defensively only ever settings permissions to False, but since this is security relevant, I think I could sleep better if we added a test here that, again, makes sure that:

If you are not a staff user and you access the "enter staff mode" url, you will get permission denied and you will not be in staff mode (aka you will not pass the is_staff check / will not be able to access a staff view / whatever).

Have you considered that?

He3lixxx (commented):
Code is really nice to read and I like the overall structure :+1: 

If @janno42 and @karyon are happy, we can merge, but for me, personally, I'd like one more test to assure that we can not critically mess up permissions.

evap/evaluation/auth.py
```diff
@@ -66,6 +66,15 @@ def check_user(user):
     return user_passes_test(check_user)(view_func)
 
 
+def staff_uncached_required(view_func):
```
karyon, line 69: `[has_]staff_permission_required`?

evap/staff/tests/test_views.py
karyon: While I agree that this is important enough to warrant additional "defense layers", your suggestion would basically test the `staff_uncached_required` decorator. I mean you can do that, but there's really not much in there that can break.

What if we added an `assert user.has_staff_permission` in `update_staff_mode` in the middleware? though upper layers should of course handle that already, this would be an additional defense in the lowest layer and would protect not only against a potentially broken enter_staff_mode view.

or you do both :)

He3lixxx: > your suggestion would basically test the staff_uncached_required decorator

That is true, in my view it was mainly a "the decorator is there" test with the main intention of having it codified in two independent places (test and implementation)

> What if we added an assert user.has_staff_permission in update_staff_mode in the middleware? though upper layers should of course handle that already, this would be an additional defense in the lowest layer and would protect not only against a potentially broken enter_staff_mode view.

I would also be fine with that.

PFischbeck: This was a very good idea! As it turns out, until now the decorator allowed anyone in, because I forgot to mark `has_staff_permission` as a property, and `request.user.has_staff_permission` was truthy...
I fixed it, added the `assert` in the middleware, and added an extra test that a normal user gets 403 when trying to enter or exit staff mode.

karyon: > but there's really not much in there that can break

I think I learned a valuable lesson today :D

He3lixxx: > As it turns out, until now the decorator allowed anyone in, because I forgot to mark has_staff_permission as a property, and request.user.has_staff_permission was truthy

That is such a bad bug and I think your only chance to spot that is a test. I guess in a statically typed language, linters should complain, but in python, pylint won't be able to figure out what we're talking about.

I had something similar recently, also concerning user access, where a code base used `def is_internal()` as a method on the `User` class, but stored `is_internal` as a boolean in a request handler `context`. All code used the context. When I wrote some new code that only had the user accessible, I made the same mistake. I wonder if there are any best practices to prevent this from happening. Do some people advocate using `== True` on truth-checks in security relevant code?

He3lixxx (approved):

janno42 (approved):

janno42:
Thank you so much for this long awaited feature! 🎉

PR #1495, opened 2020-07-27
Redesign the student index page

janno42:
Redesign of the student index page:
- Due evaluations are shown first, followed by upcoming evaluations.
- The complete semester views follow after the "Open and upcoming" card. That leads to todos and results now being more separated from each other.
- Archived semesters are transparent.
- Start and end of the evaluation period are shown as "Begins in ... days, ... hours" instead of the date (which is still available via a tooltip.)
- Responsibles' names have been added.

See commit messages for further details.

Proposed layout:
Screenshot_2020-07-27 Evaluation - EvaP


evap/student/templates/student_vote.html
```diff
@@ -22,8 +22,8 @@
     {% if evaluation.is_midterm_evaluation %}
         <div class="callout callout-warning">{% trans 'The results of this evaluation will be published while the course is still running. This means that the contributors will receive this feedback before the final grades for the course have been published.' %}</div>
     {% endif %}
-    {% if evaluation_ends_soon %}
-        <div class="callout callout-warning">{% blocktrans %}The evaluation period will end in {{ hours_left_for_evaluation }} hours and {{ minutes_left_for_evaluation }} minutes ({{ vote_end_datetime }}). Your evaluation will only be accepted if you send the completed questionnaire before this deadline ends.{% endblocktrans %}</div>
+    {% if evaluation.evaluation_ends_soon %}
```
pixunil, line 25: It feels a bit weird to double the word evaluation here. Maybe: `evaluation.is_in_extra_evalution_period` (I couldn’t think of a smart name for this time period for the time period between displayed deadline and actual deadline. I proposed elsewhere to display “ends soon” for the displayed deadline too, if you don’t want that, `evaluation.does_end_soon` would suffice.)

evap/student/templates/student_index_semester_evaluations_list.html
```diff
@@ -0,0 +1,168 @@
+{% load evaluation_filters %}
+
+{% for semester in semester_list %}
+    {% if semester.evaluations %}
+        <div class="card{% if not forloop.last %} mb-3{% endif %}{% if semester.results_are_archived %} deactivate{% endif %}">
```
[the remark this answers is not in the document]
janno42: the same thing happens for courses with multiple evaluations - i would leave this for another day.

evap/student/templates/student_index_evaluation_period.html
```diff
@@ -0,0 +1,29 @@
...
+
+{% if evaluation.state == 'prepared' or evaluation.state == 'editor_approved' %}
+    <span class="small text-secondary" data-toggle="tooltip" data-placement="left"
+        title="{% trans 'The evaluation is currently prepared by the lecturers and the evaluation team.' %} {% trans 'You will receive an email when the evaluation period begins.' %}">
+        <i>{% trans 'In preparation' %}</i>
+    </span>
+{% elif evaluation.state == 'approved' %}
+    <span class="small{%if evaluation.hours_until_evaluation < 12 %} text-primary{% endif %}" data-toggle="tooltip" data-placement="left"
+        title="{% trans 'Evaluation period' %}:<br />{{ evaluation.vote_start_datetime }} &ndash; {{ evaluation.vote_end_date }}<br /><br />{% trans 'You will receive an email when the evaluation period begins.' %}">
+        <span class="far fa-calendar-alt"></span>
```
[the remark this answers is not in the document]
janno42: that's definitely your screen :)
i don't like the plain calendar icon because i don't think it's clearly identifiable. with normal text size (which i also don't really like here) the squares look irregular on my screen, so i chose the clock icon. it fits to the icon now used for upcoming evaluations in the semester list.

evap/static/scss/_adjustments.scss
```diff
@@ -67,3 +67,9 @@ body, .fixed-top, .fixed-bottom, .is-fixed, .sticky-top {
         flex: 0 0 40%;
     }
 }
+
+
+// make yellow text darker
+.text-warning {
```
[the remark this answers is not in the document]
janno42: now using a badge instead

evap/student/templates/student_index_semester_evaluations_list.html
```diff
@@ -0,0 +1,168 @@
...
+                                            {% if evaluation.voted_for %}
+                                                <span class="text text-success" data-toggle="tooltip" data-placement="top" title="{% trans 'You gave us your feedback in this evaluation. Thank you!' %}"><span class="far fa-fw fa-smile" aria-hidden="true"></span></span>
+                                            {% elif evaluation.participates_in %}
+                                                {% if evaluation.state == 'evaluated' or evaluation.state == 'reviewed' or evaluation.state == 'published' %}
+                                                    <span class="text text-primary" data-toggle="tooltip" data-placement="top" title="{% trans 'You did not take part in this evaluation.' %}"><span class="far fa-fw fa-frown" aria-hidden="true"></span></span>
+                                                {% else %}
+                                                    <span class="text text-secondary" data-toggle="tooltip" data-placement="top" title="{% trans 'You can take part in this evaluation above.' %}"><span class="fas fa-fw fa-arrow-up" aria-hidden="true"></span></span>
```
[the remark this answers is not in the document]
janno42: icons, colors and tooltip changed

evap/student/templates/student_vote.html
janno42: `evaluation.ends_soon`

evap/student/templates/student_index_semester_evaluations_list.html
```diff
@@ -0,0 +1,164 @@
...
+                                        {% if evaluation.is_single_result %}
+                                            <span class="badge badge-success">{% trans 'Single result' %}</span>
+                                        {% endif %}
+                                        {% if course.evaluation_count == 1 %}
+                                            <i class="small">{{ course.responsibles_names }}</i>
+                                        {% endif %}
+                                    </td>
+                                    <td style="width: 20%">
+                                        {% if semester.results_are_archived %}
+                                        {% elif evaluation.is_single_result %}
```
He3lixxx, line 132: huh?

evap/student/templates/student_index_semester_evaluations_list.html
```diff
@@ -0,0 +1,164 @@
...
+                                            <span data-toggle="tooltip" data-placement="top" title="{% trans 'Individual participants are not stored for single results. The number shown here is the number of votes received.' %}">
+                                                {% include 'progress_bar_fill.html' with icon='user' total=evaluation.single_result_rating_result.count_sum fill=100 %}
+                                            </span>
+                                        {% else %}
+                                            {% if evaluation.state == 'evaluated' or evaluation.state == 'reviewed' or evaluation.state == 'published' %}
+                                                {% include 'progress_bar.html' with icon='user' done=evaluation.num_voters total=evaluation.num_participants %}
+                                            {% elif evaluation.participates_in and not evaluation.voted_for %}
+                                            {% else %}
```
He3lixxx, line 140: I don't understand.

evap/student/templates/student_index_semester_evaluations_list.html
janno42: in case the results are archived nothing should be shown there. the cases could be written in a different order with additional checks for not archived instead :)

evap/student/templates/student_index_semester_evaluations_list.html
janno42: similar to the case above. i will rewrite the conditions.

evap/student/templates/student_index_semester_evaluations_list.html
He3lixxx: Ah I see, so its basically a guard clause, but you can't return or do any other kind of early exit, so you just put an empty body. It kind of makes sense. If rewriting makes the code more complex, we can also leave it like this.

evap/student/views.py
```diff
@@ -48,15 +53,34 @@ def index(request):
...
         results_are_archived=semester.results_are_archived,
         grade_documents_are_deleted=semester.grade_documents_are_deleted,
         evaluations=[evaluation for evaluation in evaluations if evaluation.course.semester_id == semester.id]
     ) for semester in semesters]
 
+    unfinished_evaluations = list(
+        Evaluation.objects
+        .filter(participants=request.user, state__in=['prepared', 'editor_approved', 'approved', 'in_evaluation'])
+        .exclude(voters=request.user)
+        .prefetch_related('course__responsibles', 'course__type', 'course__semester', 'participants')
```
karyon, line 65: (both here and above)

Assuming the page just shows the participant count, it might be more appropriate to let the DB do the counting and prefetching that.

What about the voter count? that's also shown on the page, isn't it?

there are two methods both called `get_evaluations_with_prefetched_data` in the codebase, which both prefetch the voter and participant count. in case you're doing it here as well, could you look into factoring that out? (not sure how easy that would be)

evap/evaluation/tools.py
```diff
@@ -11,6 +11,18 @@
 from django.utils.translation import LANGUAGE_SESSION_KEY, get_language
 
 
+def is_prefetched(instance, attribute_name):
+    """
+    Is the given attribute prefetched? Can be used to do ordering or counting
+    in python and avoid additional database queries
+    """
+
+    if hasattr(instance, '_prefetched_objects_cache') and attribute_name in instance._prefetched_objects_cache:
```
karyon, line 20: `return hasattr...`

this method might have been useful elsewhere, why didn't you add it earlier?!

evap/student/views.py
janno42: @He3lixxx if you're interested you're welcome to continue with this. otherwise i'll have a look into it.

He3lixxx: I won't be able to work on this until wednesday.

He3lixxx: > Assuming the page just shows the participant count, it might be more appropriate to let the DB do the counting and prefetching that.

In general, I agree. However, I also think that using `num_participants` is a good idea, to have a global accessor that can be easily changed when needed (For example, we might want to exclude users that dropped out of the course in the future?). So, I'd only do this if it measurably improves performance. What do you think?

On the test dataset, if I put all 950 users in as participants on one of the unfinished_evaluations (Gamification für OpenHPI), the time spent in between here and after the `render()` call varies between 0.18 and 0.23 seconds. If I make the database count the participants (`.annotate(num_participants=Count("participants", distinct=True))
`, this should "override" the cached_property, right?), I get a wrong result (1 instead of 950) and a speedup of maybe 0.01 seconds (0.176 to 0.26 seconds in ~5 tests)

> there are two methods both called get_evaluations_with_prefetched_data in the codebase, which both prefetch the voter and participant count. in case you're doing it here as well, could you look into factoring that out? (not sure how easy that would be)

Both fetch the counts in an extra query and then manually attach the results to the evaluation objects. The stated reason is:
```
        # these annotates and the zip below could be replaced by something like this, but it was 2x slower:
        # annotate(num_participants=Coalesce('_participant_count', Count("participants", distinct=True)))
```
I consider this a rather situational hack, that I'd also want to keep in as little code as possible, so I wouldn't want to use it here if we don't see similar improvements.


Voter count is not shown on the page for unfinished evaluations. For the finished evaluations, we prefetch `voters`, so the same ideas might apply there.
However, we need both prefetches in lines 49 and 50 to determine if our user has to vote or already voted, so here, the count is not sufficient:
```
        evaluation.participates_in = request.user in evaluation.participants.all()
        evaluation.voted_for = request.user in evaluation.voters.all()
```

I tried to make django compute both on the database side. With these changes, I get response times from 390 to 450ms. Without them, I get 450 to 530ms response times, however, still some of the shown results are incorrect. I pushed the code to b3e02b89. Do you see what's wrong there?

evap/student/views.py
```diff
@@ -48,15 +53,34 @@ def index(request):
...
+    )
+
+    # available evaluations come first, ordered by time left for evaluation and the name
+    # evaluations in other (visible) states follow by name
+    def sorter(evaluation):
+        return (
+            evaluation.state != 'in_evaluation',
+            evaluation.state not in ['prepared', 'editor_approved', 'approved'],
+            date_to_datetime(evaluation.vote_end_date) - datetime.now() if evaluation.state == 'in_evaluation' else evaluation.full_name,
+            evaluation.full_name
```
pixunil, line 75: Maybe i’m not quite understanding key functions in Python, but if my considerations are correct, then I have some points:

- `unfinished_evaluations` only have the states `['prepared', 'editor_approved', 'approved', 'in_evaluation']`, so in my head the second line is superflous. If you want to protect yourself for the situation when more states are added, I’d prefer an explicit switch
  ```python
   if state == "in_evaluation": 0
   elif state in ['prepared', 'editor_approved', 'approved']: 1
   else: 2`
  ```
- Wouldn’t just `evaluation.vote_end_date` suffice here instead of subtracting `datetime.now()`? As comparable, it shouldn’t make a difference to ordering if you subtract a constant value
- Also, adding either a `datetime` or a `str` in the second last line, I’d prefer to add a null-value here (maybe `None` works, or a fixed date)

evap/student/templates/student_index_semester_evaluations_list.html
```diff
@@ -0,0 +1,168 @@
...
+            </div>
+            <div class="card-body table-responsive">
+                <table class="table table-seamless-links table-vertically-aligned table-headerless">
+                    <tbody>
+                        {% regroup semester.evaluations by course as course_evaluations %}
+                        {% for course, evaluations in course_evaluations|dictsort:"grouper.name" %}
+                            {% if course.evaluation_count > 1 %}
+                                <tr class="course-row heading-row">
+                                    <td></td>
+                                    <td style="width: 60%">
```
pixunil, line 24: Not necessary, but I would prefer to put the `style="width: …"` into a `<colgroup>` to prevent duplicated percentage values (and may get overlooked later)

evap/student/templates/student_index_evaluation_period.html
```diff
@@ -0,0 +1,29 @@
...
+        </span>
+    {% else %}
+        <span class="small{% if evaluation.display_hours_left_for_evaluation < 48 %} badge badge-warning{% endif %}" data-toggle="tooltip" data-placement="left"
+            title="{% trans 'Evaluation period' %}:<br />{{ evaluation.vote_start_datetime }} &ndash; {{ evaluation.vote_end_date }}">
+            <span class="far fa-clock"></span>
+            {% trans 'Ends in' %} {{ evaluation.display_vote_end_datetime|timeuntil }}
+        </span>
+    {% endif %}
+{% endif %}
+<br />
```
pixunil, line 29: Do you need this?

pixunil (changes requested):
:cat:

evap/student/templates/student_index_unfinished_evaluations_list.html
```diff
@@ -0,0 +1,34 @@
...
+        {% for evaluation in unfinished_evaluations %}
+            <tr{% if evaluation.state == 'in_evaluation' and not evaluation.voted_for %}
+                    class="hover-row hover-row-info" data-url="{% url 'student:vote' evaluation.id %}"
+                {% else %} class="deactivate"{% endif %}>
+                <td style="width: 65%">
+                    <div>
+                        <span class="evaluation-name font-weight-bold {% if evaluation.state == 'in_evaluation' and not evaluation.voted_for %} text-primary{% endif %}">
+                            {{ evaluation.full_name }}
+                        </span>
+                        {% if not evaluation.course.semester.is_active %}({{ evaluation.course.semester.name }}){% endif %}
```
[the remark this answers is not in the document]
janno42: we decided to throw this into garbage 🗑️

evap/student/views.py
karyon: I don't see it, but the docs say that [Combining multiple aggregations with annotate() will yield the wrong results](https://docs.djangoproject.com/en/3.1/topics/db/aggregation/#combining-multiple-aggregations) except for Counts with distinct=True. so you might need subqueries or separate queries after all.

In any case, it looks like the code did the same thing before. So i don't think it should block you from merging. Though i still think it would be better to fix this. It's not even because of the counting itself, I'm more bothered that the code fetches (for an average 6th semester bachelor student) like 2k UserProfiles from the database for no good reason.

pixunil (approved):

evap/student/views.py
He3lixxx: Okay, so I found one source of error. We had a query like this:
```
Evaluation.objects
.filter(participants=request.user, state__in=['prepared', 'editor_approved', 'approved', 'in_evaluation'])
.exclude(voters=request.user)
.annotate(num_participants=Count("participants", distinct=True))
.prefetch_related('course__responsibles', 'course__type', 'course__semester')
```

if you think of this in terms of a database operation, the `participants` m2m field is just another table that is joined in. In the SQL query, the `participants=request.user` will work by simply discarding all the rows where, after the join, the participant is not `request.user`. So, counting the participants afterwards obviously will always return "1", as for each evaluation, all other participants are discarded except for our user.

This can be fixed by simply moving the annotation up above the filter for participants. We will get a correct count then. I think something similar will work for the other broken query.

He3lixxx: @janno42 https://github.com/He3lixxx/EvaP/commit/b741a54ea2a44c962953c07050b1ff72111fcacb solves this discussion and passes all tests. Feel free to cherry-pick it.

He3lixxx: Oh, I just noticed, my commit is not taking care of `_participant_count` after archiving, so it might be broken. I don't think a student should see an evaluation with _participant_count set on their page. However, this might come back to us if for some reason this ever changes.

`get_evaluations_with_prefetched_data` solves this by using `Coalesce` to use either the `Count("participants")` or `_participant_count`. I will think a bit about how we could potentially consolidate the two `get_evaluations_with_prefetched_data` methods and this query at the student index view.

evap/student/views.py
```diff
@@ -24,13 +26,16 @@
...
-    # retrieve all courses which have evaluations that are not in state "new" and in which the user participates
-    courses = Course.objects.filter(
-        evaluations__participants=request.user,
-        evaluations__state__in=['prepared', 'editor_approved', 'approved', 'in_evaluation', 'evaluated', 'reviewed', 'published']
-    ).distinct().prefetch_related('semester', 'grade_documents', 'type', 'evaluations', 'evaluations__participants', 'evaluations__voters')
-    # retrieve all evaluations which the user can see that are not new
-    evaluations = [evaluation for course in courses for evaluation in course.evaluations.all() if evaluation.can_be_seen_by(request.user)]
+    query = (Evaluation.objects
+        .filter(course__evaluations__participants=request.user)
+        .exclude(state="new")
```
karyon, line 31: Here's an edge case: Course C1 has evaluations E1 (new, user does participate) and E2 (not new, user does not participate). Right now the code will include E2 for no reason apparent to the user, because the filter clause looks for any participation, and not only for participations of non-new evaluations.

evap/student/views.py
He3lixxx: Alright, so probably, the reason why the annotations in results/views.py:get_evaluations_with_prefetched_data were slow is explained here: https://stackoverflow.com/a/55482574/12345551
tl;dr: We create a big cartesian product and then count distinct entries. This blows quadratically in size, that's why it was slow. The post suggests a "better" fix with subqueries that allows to circumvent the manual zipping and attribute transfer.

@karyon with that, we should be able to make a `annotate_evaluations_with_participant_and_voter_counts` method and use that in th both `get_evaluations_with_prefetched_data` methods. That would be what you meant when you said "factor that out", right?

I'll try to code that.

He3lixxx: I added another commit (https://github.com/He3lixxx/EvaP/commit/f66cd149734ed75e7a267561913cd434cf403a82) doing what I announced there. Results, Student index and Semester index have similar performance, so it does not seem to impact performance.

Diff of both commits can be found here
https://github.com/janno42/EvaP/compare/evaluationperiod_student...He3lixxx:evaluationperiod_student_test

evap/student/views.py
janno42: added additional filter, thanks!

evap/results/views.py
```diff
@@ -96,24 +96,16 @@ def update_template_cache_of_published_evaluations_in_course(course):
...
                 "course__semester",
                 "course__responsibles",
-            ).order_by('pk')
+            )
         )
-        for evaluation, participant_count, voter_count, course_evaluations_count in zip(evaluations, participant_counts, voter_counts, course_evaluations_counts):
-            if evaluation._participant_count is None:
-                evaluation.num_participants = participant_count
-                evaluation.num_voters = voter_count
-            evaluation.course_evaluations_count = course_evaluations_count
```
karyon, line 116: @He3lixxx the course evaluation count is a separate attribute. could you restore this? (or perhaps first check whether this actually made the query count regress)

He3lixxx: Oh, totally missed that. Grep for `course_evaluations_count` only gives me 3 hits in results/exporters.py, and there, this attribute is not used. Do we even need this attribute attached here?

PR #1496, opened 2020-07-27
Keyboard navigation for voting site

### LINKED ISSUE #1134: Keyboard controls for voting
voting should be easily possible using only the keyboard and without using the tab key too much :)

either use the same approach as in #1094 (show one question at a time, some key advances to the next question) or, alternatively, leave the page design unchanged but have one question focused, use the keys to answer, and then move the focus to the next answer.

niklasmohrin:
Fixes #1134﻿

### DIFF
--- a/evap/static/scss/components/_buttons.scss
+++ b/evap/static/scss/components/_buttons.scss
@@ -191,9 +191,25 @@ a:not([href]):not(.disabled) {
         width: 3.5rem;
     }
 
+    &.focus,
+    &:focus {
+        outline: none;
+        box-shadow: 0 0 0 0.2rem rgba($evap-dark-blue, 0.5);
+    }
+
     .vote-type-bipolar &:not(:last-child) {
         width: 5rem;
         font-size: 0.8rem;
+    }
+}
+
+textarea,
+button,
+a.collapse-toggle {
+    &.tab-selectable.focus,
+    &.tab-selectable:focus {
+        outline: none;
+        box-shadow: 0 0 0 0.2rem rgba($evap-dark-blue, 0.5);
     }
 }
 
--- a/evap/student/templates/student_vote.html
+++ b/evap/student/templates/student_vote.html
@@ -89,7 +89,7 @@
                         <div class="card collapsible{% if not forloop.last %} mb-3{% endif %}">
                             <div class="card-header d-flex tab-row">
                                 <div class="mr-auto">
-                                    <button class="collapse-toggle{% if errors_exist and not contributor_has_errors %} collapsed{% endif %}" data-toggle="collapse"
+                                    <button class="collapse-toggle{% if errors_exist and not contributor_has_errors %} collapsed{% endif %} bg-transparent" data-toggle="collapse"
                                         data-target="#vote-area-{{ contributor.id }}" aria-expanded="false" aria-controls="vote-area-{{ contributor.id }}" tabindex="-1" type="button">
                                         {{ contributor.full_name }}
                                         {% if label %} &ndash; <span class="font-italic">{{ label }}</span>{% endif %}
--- a/evap/student/templates/student_vote.html
+++ b/evap/student/templates/student_vote.html
@@ -208,33 +208,33 @@
             });
 
             function updateLastSavedLabel() {
-                    const timeNow = new Date();
-                    const lastSavedLabel = $('#last-saved');
-                    if (localStorage.getItem(lastSavedStorageKey) !== null) {
-                        const lastSavedDate = new Date(localStorage.getItem(lastSavedStorageKey));
-                        const delta = Math.round((timeNow - lastSavedDate) / 1000)
-                        const languageCode = "{{ LANGUAGE_CODE }}"
-                        const relativeTimeFormat = new Intl.RelativeTimeFormat(languageCode)
-                        let timeStamp = "{% trans 'unknown' %}"
-                        if (delta < 3) {
-                            timeStamp = "{% trans 'just now' %}";
-                        } else if (delta < 60) {
-                            timeStamp = relativeTimeFormat.format(-delta, 'seconds');
-                        } else if (delta < 60 * 30) {
-                            timeStamp = relativeTimeFormat.format(-Math.round(delta / 60), 'minutes');
-                        } else if (delta < 60 * 60 * 12) {
-                            timeStamp = padWithLeadingZeros(lastSavedDate.getHours()) + ":" + padWithLeadingZeros(lastSavedDate.getMinutes());
-                        } else {
-                            timeStamp = lastSavedDate.getFullYear() + "-"
-                            + padWithLeadingZeros(lastSavedDate.getMonth() + 1) + "-"
-                            + padWithLeadingZeros(lastSavedDate.getDate()) + " "
-                            + padWithLeadingZeros(lastSavedDate.getHours()) + ":"
-                            + padWithLeadingZeros(lastSavedDate.getMinutes());
-                        }
-                        lastSavedLabel.text("{% trans 'Last saved locally' %}: " + timeStamp);
+                const timeNow = new Date();
+                const lastSavedLabel = $('#last-saved');
+                if (localStorage.getItem(lastSavedStorageKey) !== null) {
+                    const lastSavedDate = new Date(localStorage.getItem(lastSavedStorageKey));
+                    const delta = Math.round((timeNow - lastSavedDate) / 1000)
+                    const languageCode = "{{ LANGUAGE_CODE }}"
+                    const relativeTimeFormat = new Intl.RelativeTimeFormat(languageCode)
+                    let timeStamp = "{% trans 'unknown' %}"
+                    if (delta < 3) {
+                        timeStamp = "{% trans 'just now' %}";
+                    } else if (delta < 60) {
+                        timeStamp = relativeTimeFormat.format(-delta, 'seconds');
+                    } else if (delta < 60 * 30) {
+                        timeStamp = relativeTimeFormat.format(-Math.round(delta / 60), 'minutes');
+                    } else if (delta < 60 * 60 * 12) {
+                        timeStamp = padWithLeadingZeros(lastSavedDate.getHours()) + ":" + padWithLeadingZeros(lastSavedDate.getMinutes());
                     } else {
-                        lastSavedLabel.text("{% trans 'Could not save your information locally' %}");
+                        timeStamp = lastSavedDate.getFullYear() + "-"
+                        + padWithLeadingZeros(lastSavedDate.getMonth() + 1) + "-"
+                        + padWithLeadingZeros(lastSavedDate.getDate()) + " "
+                        + padWithLeadingZeros(lastSavedDate.getHours()) + ":"
+                        + padWithLeadingZeros(lastSavedDate.getMinutes());
                     }
+                    lastSavedLabel.text("{% trans 'Last saved locally' %}: " + timeStamp);
+                } else {
+                    lastSavedLabel.text("{% trans 'Could not save your information locally' %}");
+                }
             }
 
             function padWithLeadingZeros(number){
--- a/evap/student/templates/student_vote.html
+++ b/evap/student/templates/student_vote.html
@@ -45,14 +45,14 @@
                 <div class="card-header">
                     <span class="fas fa-exclamation-triangle"></span> {% trans 'Small number of participants' %}
                 </div>
-                <div class="card-body">
+                <div class="card-body tab-row">
                     {% blocktrans %}Only a small number of people can take part in this evaluation. You should be aware that contributors might be able to guess who voted for a specific answer and who wrote a text answer. Results will only be published if two or more people participated in the evaluation.{% endblocktrans %}
                     {% if evaluation.num_voters == 0 %}
                         <br />
                         {% blocktrans %}You're the first person taking part in this evaluation. If you want your text answers to be shown to the contributors even if you remain the only person taking part, please check the box below. Otherwise your text answers will be deleted if no other person takes part in the evaluation.{% endblocktrans %}
 
                         <div class="form-check mt-3">
-                            <input class="form-check-input" id="text_results_publish_confirmation_top" type="checkbox"
+                            <input class="form-check-input tab-selectable" id="text_results_publish_confirmation_top" type="checkbox"
                                 name="text_results_publish_confirmation_top" />
                             <label class="form-check-label" for="text_results_publish_confirmation_top">
                                 {% trans 'Show my text answers to the contributors even if I remain the only person who takes part in this evaluation.' %}
@@ -87,16 +87,16 @@
 
                     {% for contributor, label, form_group, contributor_has_errors, textanswers_visible_to in contributor_form_groups %}
                         <div class="card collapsible{% if not forloop.last %} mb-3{% endif %}">
-                            <div class="card-header d-flex">
+                            <div class="card-header d-flex tab-row">
                                 <div class="mr-auto">
-                                    <a class="collapse-toggle{% if errors_exist and not contributor_has_errors %} collapsed{% endif %}" data-toggle="collapse"
-                                        href="#vote-area-{{ contributor.id }}" aria-expanded="false" aria-controls="vote-area-{{ contributor.id }}">
+                                    <button class="collapse-toggle{% if errors_exist and not contributor_has_errors %} collapsed{% endif %}" data-toggle="collapse"
+                                        data-target="#vote-area-{{ contributor.id }}" aria-expanded="false" aria-controls="vote-area-{{ contributor.id }}" tabindex="-1" type="button">
                                         {{ contributor.full_name }}
                                         {% if label %} &ndash; <span class="font-italic">{{ label }}</span>{% endif %}
-                                    </a>
+                                    </button>
                                 </div>
                                 <div>
-                                    <button type="button" class="btn btn-light btn-sm"
+                                    <button type="button" class="btn btn-light btn-sm tab-selectable"
                                         data-mark-no-answers-for="{{ contributor.id }}"
                                         {% if preview %}disabled{% endif %}
                                     >
@@ -129,11 +129,11 @@
                 <div class="card-header">
                     <span class="fas fa-exclamation-triangle"></span> {% trans 'Small number of participants' %}
                 </div>
-                <div class="card-body">
+                <div class="card-body tab-row">
                     {% blocktrans %}You're the first person taking part in this evaluation. If you want your text answers to be shown to the contributors even if you remain the only person taking part, please check the box below. Otherwise your text answers will be deleted if no other person takes part in the evaluation.{% endblocktrans %}
 
                     <div class="form-check mt-3">
-                        <input class="form-check-input" id="text_results_publish_confirmation_bottom" type="checkbox"
+                        <input class="form-check-input tab-selectable" id="text_results_publish_confirmation_bottom" type="checkbox"
                             name="text_results_publish_confirmation_bottom" />
                         <label class="form-check-label" for="text_results_publish_confirmation_bottom">
                             {% trans 'Show my text answers to the contributors even if I remain the only person who takes part in this evaluation.' %}
@@ -144,7 +144,7 @@
         {% endif %}
         {% if not for_rendering_in_modal %}
             <div class="card card-submit-area{% if not preview %} card-submit-area-2{% endif %} text-center">
-                <div class="card-body">
+                <div class="card-body tab-row">
                     {% if preview %}
                         <a class="btn btn-secondary" href="{{ request.META.HTTP_REFERER }}">{% trans 'Back' %}</a>
                     {% else %}
@@ -152,7 +152,7 @@
                             <span id="last-saved"></span>
                             <span class="fas fa-info-circle" data-toggle="tooltip" title="{% trans 'The evaluation can be continued later using the same device and the same browser. But you have to submit it to send it to the server and make it count. After submitting, you can not edit the evaluation anymore.' %}"></span>
                         </div>
-                        <button id="vote-submit-btn" type="submit" class="btn btn-success">{% trans 'Submit questionnaire' %}</button>
+                        <button id="vote-submit-btn" type="submit" class="btn btn-success tab-selectable">{% trans 'Submit questionnaire' %}</button>
                         <p id="submit-error-warning" style="display: none">
                             <span class="fas fa-exclamation-triangle"></span>
                             {% blocktrans %}The server can't be reached. Your answers have been stored in your browser and it's safe to leave the page. Please try again later to submit the questionnaire.{% endblocktrans %}
@@ -302,10 +302,11 @@
             });
 
             document.querySelectorAll("[data-mark-no-answers-for]").forEach(button => {
-                button.addEventListener("click", () => {
-                    const contributorId = button.dataset.markNoAnswersFor;
-                    const voteArea = document.querySelector(`#vote-area-${contributorId}`);
+                const contributorId = button.dataset.markNoAnswersFor;
+                const voteArea = document.querySelector(`#vote-area-${contributorId}`);
+                const collapseToggle = voteArea.closest(".collapsible").querySelector(".collapse-toggle");
 
+                button.addEventListener("click", () => {
                     voteArea.querySelectorAll(".vote-btn.active").forEach(el => el.classList.remove("active"));
                     voteArea.querySelectorAll(".vote-inputs [type=radio][value='6']").forEach(radioInput => {
                         radioInput.checked = true;
@@ -316,7 +317,21 @@
 
                     // hide questionnaire for contributor
                     $(voteArea).collapse("hide");
+                    collapseToggle.classList.add("tab-selectable");
+
+                    // Disable this button, until user changes a value
+                    button.classList.remove("tab-selectable");
+                    button.disabled = true;
                 });
+
+                voteArea.querySelectorAll(".vote-inputs [type=radio]:not([value='6'])")
+                    .forEach(radioInput => {
+                        radioInput.labels[0].addEventListener("click", () => {
+                            collapseToggle.classList.remove("tab-selectable");
+                            button.classList.add("tab-selectable");
+                            button.disabled = false;
+                        });
+                    });
             });
 
             // remove error highlighting when an answer was selected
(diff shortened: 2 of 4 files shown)


niklasmohrin:
Before tackling the design of a "quick-view"-like voting site, I probably want to add keyboard navigation to the existing site, so that you don't need to tab. Maybe this is not needed, but I need to get into this stuff anyway :smile: . I am thinking of a tree based navigation, where you can move in and out of things like "general contribution", a specific contributor, textfields (these are especially hard, since we need controls for entering and exitting, that don't conflict with entering the text itself).

For mobile users, there could be something like the quick view site, that has bigger buttons and stuff.
I am not sure if I would like pagination for the evaluation on the phone, since I wouldn't be able to estimate how long the evaluation is anymore and there would be more taps involved for the same effect. Anyway, we would let the user decide which view to use, so this should fit everyone.

Let me know, if anyone of you has an explicit design in mind

niklasmohrin:
I am not sure where to go with this, I added some controls with `Tab`, arrow keys and the number row.

What do you guys think?

evap/student/templates/student_vote.html
```diff
@@ -308,6 +308,88 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
+                "down": 40,
+            };
+
+            document.getElementById("student-vote-form").addEventListener("keydown", e => {
+                if (things_to_decide[current_row][current_col].tagName == "TEXTAREA"
+                    && e.key != "Tab") {
+                    return;
+                }
+
+                console.log(e);
```
He3lixxx, line 353: It's dangerous to go debugging alone. Take this!

evap/student/templates/student_vote.html
```diff
@@ -308,6 +308,88 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
+                    case "ArrowLeft":
+                        current_col--;
+                        break;
+                    case "ArrowRight":
+                        current_col++;
+                        break;
+                    case "ArrowUp":
+                    case "ArrowDown":
+                        e.preventDefault();
+                        return;
```
He3lixxx, line 378: So you can not use Up and Down to go to the next / previous line?

janno42 (changes requested):
- the page should jump a little further when scrolling down such that the newly selected row is completely visible. it would be even better if the current row is roughly in the middle of the page.
- there might be checkboxes of the "Small number of participants" warning on the top and bottom of the page. those should also be reachable by tabbing.
- the buttons "I can't give feedback about this contributor" should also be handled. do you have an idea how?
- please make the submit button also reachable by tabbing.

evap/student/templates/student_vote.html
```diff
@@ -308,6 +308,88 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
+                    default:
+                        const number_key = e.keyCode - 48;
```
He3lixxx, line 380: I'd prefer
```js
if ((index = parseInt(e.key) - 1) != NaN) {
    if (index < things_to_decide[current_row].length) {
        current_col = index - 1;
    }
}
```
or similar, to get the 48 out of there, and also not use keyCode (https://developer.mozilla.org/de/docs/Web/API/KeyboardEvent/keyCode)

He3lixxx: (if you do the clamping below anyway, it might make sense to remove the inner condition, so that if I press "6", "5" is selected, because that's probably what I wanted)

He3lixxx (changes requested):

niklasmohrin:
I found a perfect method for centering the focussed element as @janno42 suggested, but it says that it is an "experimental API, that should not be used in production" on the [mdn page](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) (at the `scrollIntoViewOptions` parameter description). According to the [entry on caniuse.com](https://caniuse.com/#feat=mdn-api_element_scrollintoview_scrollintoviewoptions), both Firefox and Chrome support the needed feature. Do you think, I can use this here, or should I implement this differently?

evap/student/templates/student_vote.html
niklasmohrin: I am not sure about this yet. Right now, if you tab to the next row, you either land where your current selection is or you start on the neutral option. I feel like the up and down arrow keys should not trigger any horizontal movement, so they would just move up or down in the grid. There are two options for what happens next:
1) The element you land on is selected: This sucks, because you can't go through the page with the arrow keys without overwriting your selections and you would have "made a choice" (selected an option) just by moving on to the next question. Some people might be used to scroll with the arrow keys, so it would be awful for them too.
2) The element you land on is focused, but not selected: This is weird as well, since the other operation with the arrow keys automatically select the element you move on. Now, you have to either press space to select the focussed element, or move away and back on to the focussed element with the left / right key.
I think both options are unintuitive, which is why I disabled the arrow keys for now.

Additionally, in text areas, you _can_ use the left and right key, since these are not needed in this situation for anything else (and are pretty needed for text manipulation). I would want to have the arrow keys in text areas as well, but they would be reserved for navigation.

evap/student/templates/student_vote.html
niklasmohrin: Wow, and I was dead certain that `keyCode` was the sure thing to do here, in order to not fiddle around with weird strings and stuff 🤣. Well, that's that.

I can see how you could be annoyed if nothing happens when you press 6, just because you miscounted, but I am unsure where to stop here. How would I handle 0? In a (theoretical) scenario with 9 options, the 0 is closest to 9 in the number row, but closest to 1 / 2 on a numpad (and numericly). With 5 options, the zero is numericly really close to 1, but on the number row (which I feel like is the more intuitive tool for voting with the keyboard, since it aligns better with how the voting scale thingies look), I would expect that the 5 is selected.

What do you think?

evap/student/templates/student_vote.html
He3lixxx: I'd vote in favor of option 1: You need to select one entry in each row anyway, so selecting the same choice as in the previous answer (that is, "go straight down and select that entry") seems reasonable for me. If some element is already selected, obviously, we shouldn't overwrite that choice, but rather jump to that position. Then, I think this is a good implementation that makes sense for someone going through the questions.

(Also, I think having some consistent behavior on the arrow keys that might not be intuitive to all users is still better than completely disabling the arrow keys at all)

IMHO, inside text areas, all arrow keys should just move around in the text field. TAB should be the only key allowing you to leave it (but I think your current code already has this early check for text fields?)

evap/student/templates/student_vote.html
He3lixxx: That is a good point, I only thought about the linear number line on a small keyboard.
With this in mind, I think ignoring any "invalid" number key is also fine.

He3lixxx:
> I found a perfect method for centering the focussed element as @janno42 suggested, but it says that it is an "experimental API, that should not be used in production" on the [mdn page](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) (at the `scrollIntoViewOptions` parameter description). According to the [entry on caniuse.com](https://caniuse.com/#feat=mdn-api_element_scrollintoview_scrollintoviewoptions), both Firefox and Chrome support the needed feature. Do you think, I can use this here, or should I implement this differently?

So firefox and chrome behave exactly as we want them to, and other browsers support the general API (scroll into view) but will not correctly align the element to the center of the page (so they will probably just ignore the align parameter and scroll the element to the top of the page instead)?
In that case, I'd say use it, we probably won't get any better. We could [reinvent the wheel, compute the position and use window.scrollTo](https://codepen.io/janhoogeveen/pen/GwaeJP), but using a fallback on browsers not supporting the alignment is fine for me.

evap/student/templates/student_vote.html
niklasmohrin: You are right, having at least the option to use up / down without limiting the text area controls shouldn't hurt anyone except people scrolling with the arrow keys. Now that I think about it, this shouldn't be a problem either, considering that the site scrolls automatically if done as proposed in the other comments :)
Yeah, I think all key events that are not the tab key go into the text area already, but I am not standing on the hacked together code as of now. Right now I want to find out what works good for using the site and I can figure out how to do that later
Thanks for your feedback! :smiley:

niklasmohrin:
Okay great, totally agree. It was mainly the "should not be used in production code" that was confusing me here. This should not have any downsides (like security or performance or so, I couldn't think why that would be though) except for maybe breaking at some point if a standard was subject to change?

evap/student/templates/student_vote.html
pixunil: I would prefer if we could get around to imitate tab-behaviour. The default behaviour, like on `master` currently has these issues:

1. <kbd>Tab</kbd> works as expected for changing the current focus to the next item. However, in Firefox, tabbing forwards on a radio group without selected item will focus the next item of the same group. I hope we can solve this without mocking the tabbing to much, maybe only capture <kbd>Tab</kbd> when on a radio item with no item selected in its group.
2. We can’t select a radio button by it’s number. Well, this is something we definitively have to do custom, 
but it _should_ be possible without capturing <kbd>Tab</kbd>. (Nit: I would also put “No Answer” to <kbd>0</kbd>)
3. Tabbing will focus the name of a contributor on a collapsible, which I can’t really see the use-case for it. Maybe adding a `tabindex` or similar.

So, I would only concentrate on the number hot-keys (with a nice keboard legend like for the quick review, so it is discoverable) and as secondary goal to unify the behaviour of <kbd>Tab</kbd> for unselected radio groups.
I hope this will save you from some edge-cases when handling the tabbing and could keep your code small and clean :)

niklasmohrin: I was working on an implementation that gives control back to the browser if the focused item is in a row where something is selected, but I noticed two downsides with this approach:
1. The browser always focuses the first element ("strongly agree"). I think it would be best if, when tabbing onto a row where no option is selected, the "neutral" option would be focused first.
2. I would still have to figure out where the next element that would be focused is, since I want to center it with `scrollIntoView`. I could also hack this with a timeout or maybe another event (keyup or so), but it would be easier to do right away.

The focusing of the contributor collapse was easy to remove with `tabindex` though, thanks for that hint!

niklasmohrin:
@pixunil Alrighty, I am back to catching all Tab-presses so that the neutral element can be selected by default and the row-to-be-tabbed-onto can be centered on the screen. I don't really see any option to communicate to the browser where I want it to tab to (except for maintaining tabnindexes on all selectable elements as we discussed).

---

@ all
The current code correctly selects what has been discussed for these questions:

- Yes/No/No Answer
- Everything symmetrical that has a "No Answer" on the right (like agree/disagree, too much / too little)
- Text
- Grades (haven't tested manually yet but it should tab onto "3")
- (submit button row on the bottom)

Did I forget any question types?

---

@ all
In general, it would be nice if someone could look over the result again and tell me what could be improved. The code is not fully-review ready yet, I left all the previous code in as comments and maybe it needs some refactoring later. Mainly, I am looking for feedback about the user experience. Note: I know that you currently have to click into the form once to get started, I am going to fix it in the next session. That's all :)

niklasmohrin:
What's left:

- [ ] discuss and clear the 2 TODOs left
- [ ] remove code that's commented out

evap/student/templates/student_vote.html
```diff
@@ -312,6 +312,259 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
+                if (num === 0) {
+                    // Select "No answer"
+                    num = row.children.length;
+                }
+                const index = num - 1;
+
+                if (!(0 <= index && index < row.children.length)) return;
+
+                const nextElement = row.children[index];
+                // TODO: should there be check that nextElement is of type "input" here?
```
He3lixxx, line 326: If the javascript accidentally clicked on the submit button, that would be really bad.

From my understanding of the current code, it is rather unlikely that this happens. What do you think? Do you have any idea if we could add another assertion or similiar on this ("Button must not POST / submit / leave page" or whatever?)

evap/student/templates/student_vote.html
```diff
@@ -312,6 +312,259 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
+                            && nextIndex + direction < focussable.length
+                            && focussable[nextIndex].parentElement.parentElement === current.parentElement.parentElement
+                        ) nextIndex += direction;
+
+                        // TODO: possible optimization: if focussable[nextIndex] is a textarea, just focus it
+                        // Is it worth it?
+                        const nextRow = focussable[nextIndex].parentElement.parentElement;
+
+                        if (nextRow.id === "few-participants-warning") {
+                            nextElement = nextRow.children[1].children[0];
```
He3lixxx, line 392: Hmm.... I feel like this will break at some point (especially this line, relying on the specific DOM to be this way, but also the whole script in general).

@janno42 @karyon @pixunil We already have some points where we do stuff on the front end that is currently untested. I feel like this code here is another reason to (finally) add front end tests -- or, to put it differently: I wouldn't feel confident having this without any tests. What do you think?

He3lixxx (commented):
I think the code is okay, I mean, it is what it is™. Javascript manipulating the DOM is always extremely fiddly and about to break. We need to talk about tests.

evap/student/templates/student_vote.html
niklasmohrin: Yeah that would be terrible, I guess the suggested safety guard should be here anyway. To correct myself: It should check for the `type` attribute instead of the tagname, because I believe you can also do `<input type="submit">`.

Currently, this is prevented only in the calling code:

```javascript
if (!insideSubmitRow && current.tagName !== "TEXTAREA") {
  // do the number key thing
}
```

evap/student/templates/student_vote.html
niklasmohrin: I agree, all of this is pretty error prone. Maybe I could make this specifically more robust by introducing more CSS classes and querying here? Or maybe I should use xpath or something more expressive?

pixunil: Yes, having a possibility to unit-test (pure js functions) or integration-test (when you perceive html and js as two different code units) would be very nice. I had the urge to have some in my previous PR with the configurable text answers (where I just missed a `/g` flag in my regex).
I’m a bit overwhelmed with the research here, as the ecosystem is rather huge and I’m unsure whether it is sensible to migrate to e.g. Typescript or not when we are touching our javascript strategy. (but I think this general discussion is better discussed during a call)

----

For this specific line, I would prefer a `nextRow.querySelector(…)` if possible, if it is unique enough, you can match on the tag, otherwise adding a class would be the way to go.

evap/student/templates/student_vote.html
```diff
@@ -312,6 +315,117 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
+                        focussable.length - 1
+                    );
+                    nextElement = focussable[nextIndex];
+                } else {
+                    // The index of the nearest focusable element that is not in the current row
+                    let nextIndex = focussable.indexOf(current);
+                    if (nextIndex === -1) {
+                        nextElement = focussable[0];
+                    } else {
+                        while (0 <= nextIndex + direction
```
pixunil, line 389: I’d like a comment here, please. :bellhop_bell:

evap/student/templates/student_vote.html
```diff
@@ -46,14 +46,14 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
                 <div class="card-header">
                     <span class="fas fa-exclamation-triangle"></span> {% trans 'Small number of participants' %}
                 </div>
-                <div class="card-body">
+                <div class="card-body" id="few-participants-warning">
```
pixunil, line 49: Hm, if we change this id, `#text_results_publish_confirmation_top` and friends should also be renamed. But I’m not quite sure whether you really need this id.

evap/student/templates/student_vote.html
```diff
@@ -312,6 +315,117 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
+
+                // Since the event could be caught on either the outer label or
+                // the nested text / input, the full row could be two steps up
+                let currentRow = current.parentElement;
+                if (currentRow.tagName !== "DIV") currentRow = currentRow.parentElement;
+                if (currentRow.tagName !== "DIV") {
+                    return;
+                }
+
+                const insideSubmitRow = currentRow.parentElement.classList.contains("card-submit-area");
```
pixunil, line 353:
```suggestion
                const insideSubmitRow = currentRow.closest(".card-submit-area") !== null;
```
Just to be a bit more independent of the implemented markup structure.

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -34,7 +34,7 @@
                             <div class="vote-inputs {{ field.field.widget.attrs.choices.cssClass }} btn-group btn-group-toggle" data-toggle="buttons"{% if preview %} disabled{% endif %}>
                             {% for choice, color in field|zip:field.field.widget.attrs.choices.colors %}
                                 <label class="btn btn-sm vote-btn vote-btn-{{ color }} d-flex{% if field.value == choice.data.value %} active{% endif %}{% if field.errors %} choice-error{% endif %}" name="{{ choice.data.name }}"{% if not preview %} onclick="selectedAnswer('{{ choice.data.name }}');"{% endif %}>
-                                    <input id="{{ choice.id_for_label }}" name="{{ choice.data.name }}" type="radio" value="{{ choice.data.value }}" autocomplete="off"{% if field.value == choice.data.value %} checked{% endif %} />
+                                    <input class="tab-selectable num-selectable" id="{{ choice.id_for_label }}" name="{{ choice.data.name }}" type="radio" value="{{ choice.data.value }}" autocomplete="off"{% if field.value == choice.data.value %} checked{% endif %} />
```
pixunil, line 37: Please put `class` behind `id` and `name` (I’m not quite sure about the order of `class` and `name`, but at least `id` must come before `class`)

pixunil (changes requested):
Most of the behaviour is like I like it. The scrolling really adds a nice polished feeling :)


- Pressing <kbd>Backspace</kbd> is still the default action of the browser to go back. I might assume that some people would think that this would undo their choice. While I’m not sure, we want to implement undo, I think we should at least capture this keystroke and prevent the browser action.
- The styles for `.vote-btn:not(.active):focus` could be updated to have a nice border on all sides
- I can’t tab to the submit area

evap/student/templates/student_vote.html
```diff
@@ -91,15 +91,18 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
                                         {{ contributor.full_name }}
                                         {% if label %} &ndash; <span class="font-italic">{{ label }}</span>{% endif %}
                                     </a>
                                 </div>
                                 <div>
-                                    <button type="button"{% if preview %} disabled{% endif %} class="btn btn-light btn-sm" onclick="markNoAnswerFor({{ contributor.id }});">
-                                        {% blocktrans %}I can't give feedback about this contributor{% endblocktrans %}
-                                    </button>
+                                    <span>
+                                        {# Outer span needed for invariance used in <Tab> listener #}
```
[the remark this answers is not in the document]
niklasmohrin: Yeah, the `.closest` method above solves this

evap/student/templates/student_vote.html
```diff
@@ -312,6 +315,117 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
+                                nextElement = nextRow.getElementsByClassName("tab-selectable")[0];
+                            } else {
+                                const alreadySelectedElement = Array.from(nextRow.children)
+                                    .find(el => el.classList.contains("active"));
+
+                                if (alreadySelectedElement) {
+                                    nextElement = alreadySelectedElement.getElementsByClassName("tab-selectable")[0];
+                                } else {
+                                    let index;
+                                    if (nextRow.children.length === 3) { // Yes-No question, should focus "yes"
```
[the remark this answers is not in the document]
niklasmohrin: Good point! I guess it's fine to select the first element in all cases. I'll update the comment

evap/student/templates/student_vote.html
niklasmohrin: I added another class that marks elements that should be selectable by pressing the number keys as a safety guard

evap/student/templates/student_vote.html
```diff
@@ -312,6 +315,117 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
+                const index = num - 1;
+
+                if (!(0 <= index && index < row.children.length)) return;
+
+                const maybeNextElement = row
+                    .children[index]
+                    .getElementsByClassName("num-selectable");
+                if (maybeNextElement.length === 1) {
+                    maybeNextElement[0].click();
+                }
```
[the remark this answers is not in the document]
niklasmohrin: This class is another safety guard to not click on the wrong things (although this method should only be called with safe options). I don't think the querySelector is worth it, because it is slower (only slightly) and with the current code, there is another safety that there really only is one (which should always be the case in the calling context). What do you think?

evap/student/templates/student_vote.html
niklasmohrin: You are right, this was a leftover :)

evap/student/templates/student_vote_questionnaire_group.html
niklasmohrin: I never heard of this, but I can stick to it here where I add the `class` attribute. Note that there is more code around the diff that doesn't comply to this rule, but was in here before this PR, should this be fixed?

niklasmohrin:
Now working on a nicer outline, but logic / behavior can still be reviewed if you feel like it!

niklasmohrin:
Nicer outline is cancelled

janno42 (changes requested):
After the card for a contributor has been collapsed by navigating to "I can't give feedback about this contributor" and hitting Enter, one can no longer tab back over this section. I think the active element should jump back to the last question before the collapsed card.
The same happens when trying to tab forward over a collapsed card.

janno42:
When using keyboard navigation, the currently active button's outline should probably have more contrast so that users don't have to search for the currently focused element. Setting the alpha value to `0.5` looks quite nice, in my opinion.

evap/student/templates/student_vote.html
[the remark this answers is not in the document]
He3lixxx: @pixunil would you be fine with keeping the length check here?

He3lixxx (approved):
Scrolling on tab in firefox is so smooth, I love it! Sorry for the late review.

niklasmohrin:
(only rebased and squashed into two commits)

evap/student/templates/student_vote.html
```diff
@@ -89,16 +89,16 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
 
                     {% for contributor, label, form_group, contributor_has_errors, textanswers_visible_to in contributor_form_groups %}
                         <div class="card collapsible{% if not forloop.last %} mb-3{% endif %}">
-                            <div class="card-header d-flex">
+                            <div class="card-header d-flex tab-row">
                                 <div class="mr-auto">
                                     <a class="collapse-toggle{% if errors_exist and not contributor_has_errors %} collapsed{% endif %}" data-toggle="collapse"
-                                        href="#vote-area-{{ contributor.id }}" aria-expanded="false" aria-controls="vote-area-{{ contributor.id }}">
+                                        href="#vote-area-{{ contributor.id }}" aria-expanded="false" aria-controls="vote-area-{{ contributor.id }}" tabindex="-1">
```
niklasmohrin, line 95: todo: make this not a link and use data-target instead of href

janno42 (changes requested):
Once a contributor was collapsed, focusing the name and pressing space does not re-open the card but makes the page jump.

That's the only thing I found 🎉

evap/student/templates/student_vote.html
```diff
@@ -300,16 +299,33 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
                     voteArea.querySelectorAll(".vote-inputs [type=radio][value='6']")
                         .forEach(radioInput => radioInput.click());
 
                     // hide questionnaire for contributor
                     $(voteArea).collapse("hide");
+                    collapseToggle.classList.add("tab-selectable");
+
+                    // Disable this button, until user changes a value
+                    button.classList.remove("tab-selectable");
+                    button.disabled = true;
```
pixunil, line 317: :bulb: I like the feature that the button gets disabled once you activate it. But it would be even nicer if the vote area gets collapsed and the button disabled if all vote buttons were “no answer” beforehand. If this idea doesn’t fit in here nicely, we can open an issue afterwards (don’t want to do unnecessary scope creeping here)

pixunil (approved):
Mostly cosmetic things (:paintbrush:) from my side, and an idea (:bulb:) which would be nice, but not necessary. Oh, and what happened to the other idea for a keyboard legend, like for the quick review? I think we could promote this feature a bit, especially for the number keys.

evap/static/scss/components/_buttons.scss
```diff
@@ -192,6 +192,21 @@ a:not([href]):not(.disabled) {
...
+        box-shadow: 0 0 0 0.2rem rgba(49, 112, 143, 0.5);
+    }
+}
+
+textarea.tab-selectable:focus,
+textarea.tab-selectable.focus,
+button.tab-selectable:focus,
+button.tab-selectable.focus,
+a.collapse-toggle.tab-selectable:focus,
+a.collapse-toggle.tab-selectable.focus {
```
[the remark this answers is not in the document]
niklasmohrin: This is the kind of magic I was waiting for on this review :smiley:

evap/student/templates/student_vote.html
niklasmohrin: I think it wouldn't be too hard to implement this "hook" that happens on every press of a "no-answer" button and checks if all buttons are "no-answer". The site should also scroll to the contributor name afterwards, so that you wouldn't have to scroll up again.

I don't know how well this would integrate with the upcoming "additional-text-answers-to-all-questions" (naming is hard) feature, which is why I would leave this to wait until after both of the pull requests are merged to come up with something elegant then.

I don't know if this would be widely used though... I mean do you ever go through a contributor and think "hm, no, I really cannot answer this particular question, but maybe something else comes up" at every question? (or is there another use case?)

Maybe a "mark all others as no-answer and collapse" button? Or maybe the collapse button thingy should stick to the top of the main container so that you could collapse the contributor after you are done?

pixunil: Ok, let me rephrase my idea: When clicking »I can’t give feedback for this contributor« and then reloading the page, the card will reappear uncollapsed with all buttons at “no answer”. It is only about the persistence of the ui when the page reloads, sorry for not delivering this keyword to you :)

I think the phrasing of this button is well designed, I think the main use cases include: someone does not know the person’s name, wasn’t in contact with the contributor (like a tutor of another group), or is indifferent about this contributor. But the button could have some unexpected behavior if the voter answers some questions first and then clicks the button and – woopsie – all answers for this contributor were reset. But those advanced thoughts are definitely outside of this scope :)

niklasmohrin: Oh, that makes sense. Honestly, let's just open another issue afterwards like you suggested from the start :sweat_smile:

evap/static/scss/components/_buttons.scss
```diff
@@ -192,6 +192,18 @@ a:not([href]):not(.disabled) {
         width: 5rem;
         font-size: 0.8rem;
     }
+
+    &.focus, &:focus {
+        outline: none;
+        box-shadow: 0 0 0 0.2rem rgba(49, 112, 143, 0.5);
```
niklasmohrin, line 198: Should this be a constant?

niklasmohrin: (also used a couple of lines further down)

niklasmohrin (commented):
Now tabbing over the contributor works as expected (at least that's what I think)

evap/static/scss/components/_buttons.scss
janno42: the shadow doesn't need to be a constant i think, but please use `$evap-dark-blue` for the rgb part.

janno42 (changes requested):
the background color of the contributor card header texts is slightly off :)
image

also, when hitting enter while hovering a vote button, this submits the form.
that should probably not happen. i accidentally submitted a questionnaire because of this 😅

janno42 (approved):
hooray, looks like everything works now :)
if @He3lixxx and @pixunil have no further remarks, this is ready to merge.

evap/student/templates/student_vote.html
```diff
@@ -338,6 +353,123 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
+                }
+            }
+
+            const studentForm = document.getElementById("student-vote-form");
+            const selectable = studentForm.querySelectorAll(".tab-selectable");
+            const rows = Array.from(studentForm.getElementsByClassName("tab-row"));
+            studentForm.addEventListener("keydown", e => {
+                const current = document.activeElement;
+                if (!current.matches("input, label, span, textarea, button")) {
+                    // Don't do anything special
```
He3lixxx, line 390: quite obvious, isn't it? :P

He3lixxx (approved):

He3lixxx (approved):
I'm slightly in favor of allowing `*` and `-` to start typing in the additional text answer field (to start bullet lists). I'd also be okay with allowing more chars than that (e.g. everything that is a letter and not a digit)

PR #1497, opened 2020-08-17
Student index page: Fix O(N) database queries

He3lixxx:
Hopefully, with these changes, the server should be able to deliver the student index page in < 500ms.


janno42 (approved):

He3lixxx:
Production currently delivers the page in 2 seconds. That is still way worse than I would have expected (when my dev server is able to deliver the page in around 500ms). Maybe archived semesters still have some O(N) behaviour or something? In the development data, we don't have an archived semester afaik. I will have to look into that some more.

PR #1498, opened 2020-08-20
Clearable multi select fields

### LINKED ISSUE #1451: "Clear all" option for participant form field
There should be an option to remove all currently selected participants in the Evaluation form.

JannisBerndt:
I havent had a better idea yet. I am open for some suggestions...

Fixes #1451

### DIFF
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -110,7 +110,8 @@
                 language: "{{ LANGUAGE_CODE }}",
                 placeholder: "{% trans 'Please select...' %}",
                 minimumResultsForSearch: 15,
-                minimumInputLength: $(this).children('option').length > 50 ? 3 : 0
+                minimumInputLength: $(this).children('option').length > 50 ? 3 : 0,
+                allowClear: this.multiple
             });
         });
     };


JannisBerndt:
* _I havent had a better idea yet. I am open for some suggestions..._

evap/evaluation/templates/base.html
```diff
@@ -110,7 +110,8 @@
                 language: "{{ LANGUAGE_CODE }}",
                 placeholder: "{% trans 'Please select...' %}",
                 minimumResultsForSearch: 15,
-                minimumInputLength: $(this).children('option').length > 50 ? 3 : 0
+                minimumInputLength: $(this).children('option').length > 50 ? 3 : 0,
+                allowClear: (this.name == "participants") ? true : false
```
He3lixxx, line 114: I'd suggest adding a second argument to this function (`applySelect2`) that allows to provide additional arguments or override some of the default ones. Then, the place where the "exceptional" behaviour is defined is closer to where we actually need it.

(afaik javascript has some object packing and unpacking semantics similar to **kwargs in python)

janno42: i would replace this line with `allowClear: this.multiple` instead. it's actually helpful to add this for all multi select fields and we don't have to handle it individually.

janno42 (changes requested):

janno42:
Very nice and easy solution 👍

janno42 (approved):

He3lixxx (approved):

janno42:
thanks :)

PR #1499, opened 2020-08-24
Fix crash when importing user data with multiple excel sheets

### LINKED ISSUE #1493: Handle multiple sheets in import files correctly
When importing an excel file with multiple sheets (e.g. when using the person importer on the person management page for a certain course) and at least one of the sheets is not correctly formatted, processing the import fails with a `TypeError`:
```
cannot unpack non-iterable NoneType object

File "/opt/evap/evap/staff/importers.py", line 660, in process_file_content
    user_list, importer.success_messages, importer.warnings, importer.errors = UserImporter.process(file_content, test_run)
```

The importer should show an error message instead. The same should be done for other importers as well.

niklasmohrin:
Closes #1493

The logic was correct, there was just a line missing.

### DIFF
--- a/evap/staff/importers.py
+++ b/evap/staff/importers.py
@@ -540,7 +540,7 @@ class UserImporter(ExcelImporter):
                 except Exception as error:
                     self.errors[ImporterError.GENERAL].append(
                         _("A problem occured while writing the entries to the database."
-                          " The error message has been: '{}'").format(error=error))
+                          " The error message has been: '{}'").format(error))
                     raise
 
         msg = format_html(_("Successfully created {} users:"), len(created_users))
--- a/evap/staff/importers.py
+++ b/evap/staff/importers.py
@@ -603,6 +603,7 @@ class UserImporter(ExcelImporter):
             if settings.DEBUG:
                 # re-raise error for further introspection if in debug mode
                 raise
+            return [], importer.success_messages, importer.warnings, importer.errors
 
 
 class PersonImporter:
--- a/evap/staff/tests/test_importers.py
+++ b/evap/staff/tests/test_importers.py
@@ -1,5 +1,7 @@
 import os
 from datetime import date, datetime
+from unittest.mock import patch
+
 from django.test import TestCase, override_settings
 from django.conf import settings
 from model_bakery import baker
@@ -119,6 +121,17 @@ class TestUserImporter(TestCase):
             " None None, lucilia.manilium@institution.example.com"])
 
         self.assertEqual(UserProfile.objects.count(), 2)
+
+    @override_settings(DEBUG=False)
+    @patch("evap.evaluation.models.UserProfile.objects.update_or_create")
+    def test_unhandled_exception(self, mocked_db_access):
+        mocked_db_access.side_effect = Exception("Contact your database admin right now!")
+        result, __, __, errors = UserImporter.process(self.valid_excel_content, test_run=False)
+        self.assertEqual(result, [])
+        self.assertIn(
+            "Import finally aborted after exception: 'Contact your database admin right now!'",
+            errors[ImporterError.GENERAL],
+        )
 
 
 class TestEnrollmentImporter(TestCase):


He3lixxx:
(Reminder: Test)

niklasmohrin:
Right, so I wanted to work on a test case today, but I had to realize, that my broken file (one sheet formatted correctly and another one with mistakes) is actually properly rejected, even without this fix. I threw the file at both the `UserImporter` and the `PersonImporter`, but both generated

> Errors occurred while parsing the input data. No data was imported. 

This error message is appended, if the file is an excel file with four columns, but the user data is invalid. After a first look, it seems as if only the line 585 `importer.for_each_row_in_excel_file_do(importer.read_one_user)` could raise an exception, but that would mean, that `importer.read_one_user` would throw an exception, but it cannot see, where that would happen. I could not find any opportunity for a thrown exception in any of the other methods either, so I am kind of stuck here.

@janno42 Do you still have the broken excel file from when you were submitting the issue? That would be a great help!

janno42:
turns out, the error was something else all along :)
```
AttributeError
'float' object has no attribute 'strip'
/evap/evap/staff/importers.py in __init__, line 40
```

this occurs when the column "title" contains numbers instead of strings (as people often put running numbers in the first column).
so we should add additional data sanity checks for this kind of wrong data.

He3lixxx:
The fix we have here (missing return statement, leading to unpacking of None, raising a TypeError) should still be merged.

As for testing: In the call stack of a normal import execution, we will at some point have
```
PersonImport:process_file_content
UserImporter:process
UserImporter:save_users_to_db
UserData:store_in_database
```
where the case we fix here will happen if UserImporter:process throws an unexpected error. Now, on malformed input, we usually want to give out a precise error message, so I'd suggest mocking a technical failure. `store_in_database` calls `UserProfile.objects.update_or_create`, so we could mock that and make it raise some exception (as would happen if the database was offline at that time, for example), giving us a simple 5 line test.

niklasmohrin:
I added the proposed test, but I stumbled upon something else on the way. See [the second commit](https://github.com/e-valuation/EvaP/pull/1499/commits/ee78c16d265dad3269300363b46498b7c94b3e6a) for that change, this error sneaked in [in this commit](https://github.com/e-valuation/EvaP/commit/e0e379a61ba3e574c0ba5f2f935a53d294417815#diff-da68433c909f6cbfecd51d376e8bf4eaL450) a couple of months ago.

I just saw, that the travis build fails in debug mode, which is to be expected, as the importer re-raises caught exceptions in debug mode instead of just catching them and appending to the `errors` list. Should I write something like
```python
if settings.DEBUG:
    # test for thrown exception
else:
    # test for caught exception
```
in the new test, or should I mock the setting for the test (if that is possible)?

niklasmohrin:
The new data sanity check could just check, that all non-empty cells in the sheet must be of type `str`, so that should be straight forward, right? I could add this here as well.

He3lixxx:
> I just saw, that the travis build fails in debug mode, which is to be expected, as the importer re-raises caught exceptions in debug mode instead of just catching them and appending to the `errors` list. Should I write something like
> 
> ```python
> if settings.DEBUG:
>     # test for thrown exception
> else:
>     # test for caught exception
> ```
> 
> in the new test, or should I mock the setting for the test (if that is possible)?

I think we used the `@override_settings` decorator for that in the past, [for example in this test](https://github.com/e-valuation/EvaP/blob/master/evap/staff/tests/test_views.py#L329)

For your data type correctness, we need our PO (@janno42) to decide. This would mean that users can not upload excel files where any field is of type "number" or "date" (see https://xlrd.readthedocs.io/en/latest/api.html#xlrd.sheet.Cell)

niklasmohrin:
The decorator is just what I am looking for, thanks!

> This would mean that users can not upload excel files where any field is of type "number" or "date"

I thought, that this new check could be provided as a method on the base class `ExcelImporter` and then used in deriving classes only when needed.

Maybe, this could even be more generalized for other use elsewhere, so that you could do something like `self.check_data_type_correctness([int, str, str, float])`, where the given list is just compared to the types of the cells in each row. An implementation like this could be a more specialized version of the `check_column_count` method, that is already in use (i.e. the method also calls `check_column_count(len(argument_list))`).

In case this method is added, we need to come up with a way to deal with `int` and `float`. As you pointed out, both the "number" and the "date" type of excel is a float value in python and a boolean in excel is (only?) an `int` in python (maybe, there could be an enum on the base class, or the constants from xlrd must be used when using `check_data_type_correctness`). I think there sould be a way to restrict the expected cell value to an integer (for example if you want to have student ids or something, where you would have the number type in excel, but would want to check whether it is integer).

What do you think?

janno42:
we just use strings, so every cell could be parsed as string value, regardless of the excel data type.

He3lixxx:
> The decorator is just what I am looking for, thanks! (edit: This now opens the possibility to have two tests, one for each possible environment. Do you think, there should be tests about the debugging instance?)

You can add one if you want, but the re-raise is only for convenience when running a debugging server, so I wouldn't require it.

janno42 (approved):

PR #1502, opened 2020-08-29
Importer string cleaning

### LINKED ISSUE #1489: Importer string cleaning
All importers should replace multiple consecutive spaces in cell values by single spaces.

There should be a test in which an exemplary enrollment file is imported, which includes two courses with a name differing only in the amount of spaces. The test should assert that only one course is created by the import.

JannisBerndt:
Fixes #1489


JannisBerndt:
Just dont ask what I've done there, pls

evap/staff/importers.py
```diff
@@ -244,7 +244,10 @@ def for_each_row_in_excel_file_do(self, row_function):
         for sheet in self.book.sheets():
             try:
                 for row in range(self.skip_first_n_rows, sheet.nrows):
-                    row_function(sheet.row_values(row), sheet, row)
+                    data = []  # container for normalized cell data
+                    for cell in sheet.row_values(row):
+                        data.append(' '.join(cell.split()))  # see https://stackoverflow.com/questions/2077897/substitute-multiple-whitespace-with-single-whitespace-in-python
```
karyon, line 249: a list comprehension is probably more concise here.

evap/staff/tests/test_importers.py
```diff
@@ -125,6 +125,7 @@ class TestEnrollmentImporter(TestCase):
     filename_valid = os.path.join(settings.BASE_DIR, "staff/fixtures/test_enrollment_data.xls")
     filename_valid_degree_merge = os.path.join(settings.BASE_DIR, "staff/fixtures/test_enrollment_data_degree_merge.xls")
     filename_valid_import_names = os.path.join(settings.BASE_DIR, "staff/fixtures/test_enrollment_data_import_names.xls")
+    filename_valid_consecutive_and_trailing_spaces = os.path.join(settings.BASE_DIR, "staff/fixtures/test_enrollment_data_consecutive_and_trailing_spaces.xls")
```
karyon, line 128: not part of this PR, but do you think it might make sense to create some helper methods that create an (in-memory) xls from python data structures? We're adding more and more .xls files to the repo, which are not reviewable and must be manually edited in case the schema changes. we could keep one or two xls files in the tests to make sure we don't break reading "real" xls files. @JannisBerndt @janno42 @He3lixxx

evap/staff/importers.py
JannisBerndt: Yep, way better

evap/staff/tests/test_importers.py
JannisBerndt: I think this makes sense. One could argue that the conversion is an additional point of failure but I think its not that critical.

He3lixxx: We could use csv of xls to make it human readable / reviewable. Schema changes would still be a problem.

I don't know whether using a data structure in code would boost maintainability. Are we talking about using a list of lists here?

Pylint organizes their tests like this: https://github.com/PyCQA/pylint/tree/master/tests/functional/r

janno42 (approved):

evap/staff/tests/test_importers.py
janno42: see #1503

He3lixxx (approved):

He3lixxx:
Thanks for contributing :+1:

PR #1504, opened 2020-08-31
Fix test failures in test_results_cache_refreshed

### LINKED ISSUE #1486: test_results_cache_refreshed is acting weird
`test_results_cache` in `staff/tests/test_forms.py` is failing randomly.

Behaviour observed:
- On master, without `--keepdb`, it seems to pass most of the time (not one crash in about 15 runs in a row)
- It broke three out of eight times on master with `--keepdb`
- On one of my branches (the Exporter refactor), it breaks on every 2nd - 5th test run (with `--keepdb`)
- It broke once for @He3lixxx over the period of the entire evening

My master was at `4595866ed29b3209cfeb66994aa347ec9a66e1e0`

He3lixxx:
fixes  #1486

I initially thought about using `assertEqual(next(filter(...)))` but I felt like this was simpler and open for adding other contributors -- however, if anyone has a strong opinion about the asserts, I can as well change them.

### DIFF
--- a/evap/staff/tests/test_forms.py
+++ b/evap/staff/tests/test_forms.py
@@ -142,14 +142,14 @@ class UserFormTests(TestCase):
 
         results_after = collect_results(evaluation)
 
-        self.assertEqual(
-            results_before.contribution_results[0].contributor.first_name,
-            "Peter"
+        self.assertCountEqual(
+            (result.contributor.first_name for result in results_before.contribution_results if result.contributor),
+            ("Peter", ),
         )
 
-        self.assertEqual(
-            results_after.contribution_results[0].contributor.first_name,
-            "Patrick"
+        self.assertCountEqual(
+            (result.contributor.first_name for result in results_after.contribution_results if result.contributor),
+            ("Patrick", ),
         )
 
 


janno42 (approved):

evap/staff/tests/test_forms.py
```diff
@@ -142,14 +142,14 @@ def test_results_cache_refreshed(self):
...
+            (result.contributor.first_name for result in results_before.contribution_results if result.contributor),
+            ("Peter", ),
         )
 
-        self.assertEqual(
-            results_after.contribution_results[0].contributor.first_name,
-            "Patrick"
+        self.assertCountEqual(
+            (result.contributor.first_name for result in results_after.contribution_results if result.contributor),
+            ("Patrick", ),
```
karyon, line 152: are you sure you want assertCountEqual? the actual string content is not compared anymore.

He3lixxx: `assertCountEqual` is quite a misleading name - see https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual. It does not assert that the lengths are the same, but rather that each element has the same count in the sequence

karyon: right. thanks!

PR #1506, opened 2020-09-07
Proxied evaluation shouldn't show as delegated

### LINKED ISSUE #1419: Proxied evaluations shouldn't show as delegated
Evaluations where a proxy user is responsible shouldn't be shown as "delegated courses" for the delegates of this proxy user. The delegates should see this course in the list of their own courses like those they are directly responsible for and they shouldn't be hidden when the option to hide delegated courses is active.

JannisBerndt:
Fixes #1419

### DIFF
--- a/evap/contributor/views.py
+++ b/evap/contributor/views.py
@@ -28,7 +28,9 @@ def index(request):
     own_courses = Course.objects.filter(
         Q(evaluations__state__in=contributor_visible_states) & (
             Q(responsibles=user) |
-            Q(evaluations__contributions__contributor=user)
+            Q(evaluations__contributions__contributor=user) |
+            Q(evaluations__contributions__contributor__in=user.represented_users.filter(is_proxy_user=True)) |
+            Q(responsibles__in=user.represented_users.filter(is_proxy_user=True))
         )
     )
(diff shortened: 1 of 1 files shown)


evap/contributor/views.py
```diff
@@ -37,7 +39,7 @@ def index(request):
 
     displayed_evaluations = set(own_evaluations)
     if show_delegated:
-        represented_users = user.represented_users.all()
+        represented_users = user.represented_users.filter(is_proxy_user=False)
```
janno42, line 42: please change to `.exclude(is_proxy_user=True)`

evap/contributor/views.py
```diff
@@ -28,7 +28,9 @@ def index(request):
     own_courses = Course.objects.filter(
         Q(evaluations__state__in=contributor_visible_states) & (
             Q(responsibles=user) |
-            Q(evaluations__contributions__contributor=user)
+            Q(evaluations__contributions__contributor=user) |
+            Q(evaluations__contributions__contributor__in=user.represented_users.filter(is_proxy_user=True)) |
```
janno42, line 32: please get the represented proxy users before the `own_courses` query and use the variable in the two occurrences here

janno42 (changes requested):
thanks!

janno42 (approved):

PR #1512, opened 2020-09-15
Automatically remove delegations on user deactivation

### LINKED ISSUE #1305: Automatically remove delegations on user deactivation
This requires #1304 being implemented first.

Being a delegate or CC user should not prevent a UserProfile from being deleted.
When a user is deleted (or set as inactive, manually or via bulk deletion), the user should be removed from all other users' delegates and CC users. A message should inform about all removals that happened.

JannisBerndt:
Fixes #1305

### DIFF
--- a/evap/locale/de/LC_MESSAGES/django.po
+++ b/evap/locale/de/LC_MESSAGES/django.po
@@ -5,7 +5,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: EvaP\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-06-29 10:02+0200\n"
+"POT-Creation-Date: 2020-09-15 10:35+0200\n"
 "PO-Revision-Date: 2020-06-29 10:06+0200\n"
 "Last-Translator: Johannes Wolf <johannes-wolf@posteo.de>\n"
 "Language-Team: Johannes Wolf (janno42)\n"
@@ -34,7 +34,7 @@ msgstr "Name (Deutsch)"
 msgid "Name (English)"
 msgstr "Name (Englisch)"
 
-#: evap/contributor/forms.py:54 evap/staff/forms.py:318
+#: evap/contributor/forms.py:54 evap/staff/forms.py:319
 msgid "The first day of evaluation must be before the last one."
 msgstr "Der erste Evaluierungstag muss vor dem letzten liegen."
 
@@ -81,8 +81,8 @@ msgid "Responsibles"
 msgstr "Verantwortliche"
 
 #: evap/contributor/templates/contributor_evaluation_form.html:47
-#: evap/evaluation/templates/navbar.html:52 evap/results/exporters.py:151
-#: evap/results/templates/results_index.html:26 evap/staff/forms.py:846
+#: evap/evaluation/templates/navbar.html:52 evap/results/exporters.py:152
+#: evap/results/templates/results_index.html:26 evap/staff/forms.py:850
 #: evap/staff/templates/staff_degree_index.html:5
 #: evap/staff/templates/staff_index.html:48
 #: evap/staff/templates/staff_semester_export.html:25 evap/staff/views.py:536
@@ -315,7 +315,7 @@ msgid "Hide"
 msgstr "Verbergen"
 
 #: evap/contributor/templates/contributor_index.html:68
-#: evap/evaluation/models.py:1460
+#: evap/evaluation/models.py:1456
 #: evap/grades/templates/grades_semester_view.html:31
 #: evap/staff/templates/staff_semester_preparation_reminder.html:40
 #: evap/staff/templates/staff_semester_view.html:259
@@ -415,13 +415,13 @@ msgid "Title"
 msgstr "Titel"
 
 #: evap/contributor/templates/contributor_settings.html:20
-#: evap/rewards/exporters.py:27 evap/staff/templates/staff_user_merge.html:33
+#: evap/rewards/exporters.py:12 evap/staff/templates/staff_user_merge.html:33
 #: evap/staff/views.py:1070
 msgid "First name"
 msgstr "Vorname"
 
 #: evap/contributor/templates/contributor_settings.html:24
-#: evap/rewards/exporters.py:26 evap/staff/templates/staff_user_merge.html:39
+#: evap/rewards/exporters.py:11 evap/staff/templates/staff_user_merge.html:39
 #: evap/staff/views.py:1070
 msgid "Last name"
 msgstr "Nachname"
@@ -524,7 +524,7 @@ msgstr ""
 "In Ihrem Web-Browser scheinen Cookies deaktiviert zu sein. Cookies werden "
 "zur Anmeldung benötigt."
 
-#: evap/evaluation/forms.py:55 evap/rewards/exporters.py:28
+#: evap/evaluation/forms.py:55 evap/rewards/exporters.py:13
 msgid "Email address"
 msgstr "E‑Mail-Adresse"
 
@@ -705,7 +705,7 @@ msgid "gets no grade documents"
 msgstr "erhält keine Noten-Dokumente"
 
 #: evap/evaluation/models.py:292 evap/evaluation/models.py:389
-#: evap/grades/forms.py:14 evap/staff/forms.py:340
+#: evap/grades/forms.py:14 evap/staff/forms.py:341
 msgid "Last modified"
 msgstr "Zuletzt verändert"
 
@@ -1417,6 +1417,7 @@ msgid "CC Users"
 msgstr "CC-Benutzer"
 
 #: evap/evaluation/models.py:1245
+#: evap/staff/templates/staff_user_badges.html:17
 msgid "Proxy user"
 msgstr "Proxy-Benutzer"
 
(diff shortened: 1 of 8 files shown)


evap/locale/de/LC_MESSAGES/django.po
```diff
@@ -34,7 +34,7 @@ msgstr "Name (Deutsch)"
 msgid "Name (English)"
 msgstr "Name (Englisch)"
 
-#: evap/contributor/forms.py:54 evap/staff/forms.py:318
+#: evap/contributor/forms.py:54 evap/staff/forms.py:319
```
JannisBerndt, line 37: Regarding the translation: I ran `./manage.py makemessages -l de` which updated very many code line refernces that got changed sometime in the past. Do you prefer manually adding the new translations or is the big "update" done somewhere else once in a while?

janno42: we do the translations on every update of the release branch. so you only have to make sure that all strings are marked for translation in views (`_("translate this")`) and templates (`{% trans 'translate this' %}`) and i'll do the rest. please do not commit any changes on the `django.po` file.

evap/staff/templates/staff_user_index.html
```diff
@@ -64,8 +64,7 @@
                                 ></button>
                             {% else %}
                                 <button type="button" disabled class="btn btn-sm btn-danger" data-toggle="tooltip" data-placement="left"
-                                    title="{% blocktrans %}This user contributes to an evaluation (or is a delegate or CC user of
-                                    such a person), participates in an evaluation whose participations haven't been archived yet or has special
+                                    title="{% blocktrans %}This user contributes to an evaluation, participates in an evaluation whose participations haven't been archived yet or has special
```
janno42, line 67: please have a look at the line breaks, they are a bit mixed up here. if easily possible, lines should not be longer than 120 characters.

evap/staff/templates/staff_user_index.html
```diff
@@ -82,7 +81,7 @@
 {% block modals %}
     {{ block.super }}
     {% trans 'Delete user' as title %}
-    {% trans 'Do you really want to delete the user <strong id="data-label"></strong>?' as question %}
+    {% trans 'Do you really want to delete the user <strong id="data-label"></strong>?<br><strong id="data-label"></strong> will also be removed from every user having <strong id="data-label"></strong> as a delegated or CC-user.' as question %}
```
janno42, line 84: please change to `{% trans 'Do you really want to delete the user <strong id="data-label"></strong>?<br />This person will also be removed from every other user having this person as a delegated or CC-user.' as question %}`

evap/staff/tools.py
```diff
@@ -292,3 +301,21 @@ def find_next_unreviewed_evaluation(semester, excluded):
         .filter(contributions__textanswer_set__state=TextAnswer.State.NOT_REVIEWED) \
         .annotate(num_unreviewed_textanswers=Count("contributions__textanswer_set")) \
         .order_by('vote_end_date', '-num_unreviewed_textanswers').first()
+
+
+def remove_user_from_represented_and_ccing_users(user, ignored_users=None, test_run=False):
+    remove_messages = []
+    ignored_users = list() if ignored_users is None else ignored_users
```
janno42, line 308: you could write `ignored_users = ignored_users or []` instead

evap/staff/tools.py
```diff
@@ -182,6 +188,9 @@ def bulk_update_users(request, user_file_content, test_run):
         messages.info(request, _('No data was changed in this test run.'))
     else:
         with transaction.atomic():
+            for user in deletable_users + users_to_mark_inactive:
+                # this has to be done in an additional for-loop because otherwise it interferes with the user deletion
+                remove_user_from_represented_and_ccing_users(user, deletable_users + users_to_mark_inactive)
```
janno42, line 193: you don't show any messages here, are you?

janno42 (changes requested):
thanks! i still have to do some testing, but here is already some feedback. can you also please add automated tests for this new feature?

evap/staff/tools.py
He3lixxx: or just use `(user, ignored_users=[], test_run=False)` (keep in mind that the actual list instance is then shared across calls, so you can't do that if you modify the list in the method. Shouldn't be a problem here, though)

JannisBerndt: Using '[]' as default parameter throws an error in the codecov test because of the problem @He3lixxx mentioned (which wozld not be an actual problem im this case). I changed it because of that... The Suggestion of @janno42 seems the best solution then...

evap/staff/tools.py
JannisBerndt: No, because they would be doubled with the information messages in the previous for-loops.
Should they be shown twice?
The only real problem is the future tense in the messages, but it is also future in all the other notifications. How should it be done?

JannisBerndt:
> thanks! i still have to do some testing, but here is already some feedback. can you also please add automated tests for this new feature?

Thanks for the feedback, yes I will add some tests.

evap/staff/tools.py
```diff
@@ -292,3 +301,21 @@ def find_next_unreviewed_evaluation(semester, excluded):
...
         .annotate(num_unreviewed_textanswers=Count("contributions__textanswer_set")) \
         .order_by('vote_end_date', '-num_unreviewed_textanswers').first()
+
+
+def remove_user_from_represented_and_ccing_users(user, ignored_users=None, test_run=False):
+    remove_messages = []
+    ignored_users = list() if ignored_users is None else ignored_users
+    for represented_user in user.represented_users.exclude(id__in=[user.id for user in ignored_users]):
+        if test_run:
+            remove_messages.append(_("%s will be removed from the delegates of %s.") % (user.full_name, represented_user.full_name))
```
He3lixxx, line 311: We usually the the more recent .format() syntax of python that gets rid of some of the weirdness of the older %-formatting. I'd appreciate if you changed this here (and in the other few cases)

evap/staff/tools.py
janno42: it's fine how it is. i didn't see that the "test" messages are also shown on the real update run :)

evap/staff/tools.py
```diff
@@ -292,3 +301,21 @@ def find_next_unreviewed_evaluation(semester, excluded):
...
+
+def remove_user_from_represented_and_ccing_users(user, ignored_users=None, test_run=False):
+    remove_messages = []
+    ignored_users = ignored_users or []
+    for represented_user in user.represented_users.exclude(id__in=[user.id for user in ignored_users]):
+        if test_run:
+            remove_messages.append(_("{} will be removed from the delegates of {}.").format(user.full_name, represented_user.full_name))
+        else:
+            represented_user.delegates.remove(user)
+            remove_messages.append(_("Removed {} from the delegates of {}.").format(user.full_name, represented_user.full_name))
```
janno42, line 314: as discussed in a comment above, this message and the one in line 320 are never shown because you're only using them in the bulk update when it's not a test run, correct? then i'd say you can remove the messages altogether.

janno42 (changes requested):
just one more change request from my side :)

evap/staff/tools.py
JannisBerndt: The only place where those messages are used is in line 782 in forms.py.
I am not sure whether we should care about different tenses or use one message tense everywhere.
Furthermore I placed this call into the user change form because the delegates and cc users should get removed on every user change. However, displaying the messages is handled individually outside the form (after changing a user)...

He3lixxx: I feel like this method is fine. Having a distinction between "will be changed" for a test run and "was changed" for a real run is a good thing imo. We should change the three calls in bulk_update_users below.

As far as I can see, in `bulk_update_users`, the simple idiomatic snippet:
```python
for user in deletable_users + users_to_mark_inactive:
    messages = remove_user_from_represented_and_ccing_users(user, deletable_users + users_to_mark_inactive, test_run)
    messages.warning(...)
```

should work -- or am I missing something? We may have to make the database transaction a bit bigger, but that shouldn't be much of a problem.

janno42 (approved):

He3lixxx (approved):

evap/staff/tools.py
```diff
@@ -170,18 +170,20 @@ def bulk_update_users(request, user_file_content, test_run):
                 create_user_list_html_string_for_message(users_to_mark_inactive)
             )
         )
-    if emails_of_users_to_be_created:
```
janno42, line 173: why did you remove this line?

evap/staff/views.py
```diff
@@ -1598,6 +1598,8 @@ def notify_reward_points(grantings, **_kwargs):  # pylint: disable=unused-variab
         form.save()
         delete_navbar_cache_for_users([user])
         messages.success(request, _("Successfully updated user."))
+        for message in form.remove_messages:
+            messages.warning(request, message)
```
janno42, line 1602: codecov reports line 1602 not being covered by tests. did you miss it or was including it in a test too complex?

janno42 (changes requested):

evap/staff/tools.py
JannisBerndt: Was a mistake, oops...

evap/staff/views.py
JannisBerndt: didnt notice because it broke when adding "mocking" to the test. Got fixed in the last commit

janno42 (approved):

janno42:
Thank you so much for your continuous support! 🚀

PR #1513, opened 2020-09-29
Configurable text answer warnings

### LINKED ISSUE #1507: Configurable text answer warnings
There currently exists a text answer warning when students use terms that might reference the answer to a different question (`doesTextareaContainReference` in `student_vote.html`).
Similarly, such warnings could be used for strong language and other use cases.
The current warning should be changed to a customizable system where staff users can define any number of warnings on a new ui page and define a list of triggering strings for each warning.

The warning for meaningless text (`isTextareaInvalid`) can stay hard-coded but should probably be renamed to something like `doesTextareaContainMeaninglessText`.

pixunil:
Closes #1507 

Remember to run:
```
./manage.py migrate student
```

- Creates the `TextAnswerWarning` model (the first one in the student app) which contains so-called trigger strings which will trigger a warning when used in a text answer
- Now, multiple text answers can be displayed at the same time, so the styling is a bit adjusted
- Staff not only gets a configuration form, but a little preview textarea as well

textarea-with-warnings

staff-configuration

### DIFF
--- a/evap/staff/templates/staff_text_answer_warnings.html
+++ b/evap/staff/templates/staff_text_answer_warnings.html
@@ -1,5 +1,8 @@
 {% extends 'staff_base.html' %}
 
+{% load static %}
+{% load student_filters %}
+
 {% block breadcrumb %}
     {{ block.super }}
     <li class="breadcrumb-item">{% trans 'Text answer warnings' %}</li>
@@ -63,13 +66,44 @@
                 <button type="submit" class="btn btn-primary">{% trans 'Save text answer warnings' %}</button>
             </div>
         </div>
+        <div class="card">
+            <div class="card-header">
+                {% trans 'Preview' %}
+            </div>
+            <div class="card-body">
+                <p>
+                    {% blocktrans %}
+                        Changes of the form above are only reflected after the form has been saved.
+                    {% endblocktrans %}
+                </p>
+                <div class="row">
+                    <div class="col-question col-lg-4 col-xl-5 d-flex flex-column">
+                        <label for="preview-textarea">
+                            {% trans 'Test textarea' %}
+                        </label>
+                        {% include 'student_text_answer_warnings.html' with text_answer_warnings=text_answer_warnings %}
+                    </div>
+                    <div class="col-answer col-lg-8 col-xl-7 d-flex">
+                        <div class="vote-inputs">
+                            <textarea id="preview-textarea"></textarea>
+                        </div>
+                    </div>
+                </div>
+            </div>
+        </div>
     </form>
 {% endblock %}
 
 {% block additional_javascript %}
     {% include 'sortable_form_js.html' %}
 
+    {{ text_answer_warnings|text_answer_warning_trigger_strings|json_script:'text-answer-warnings' }}
+
     <script type="module">
+        import {initTextAnswerWarnings} from "{% static 'js/text-answer-warnings.js' %}";
+
+        initTextAnswerWarnings($("#preview-textarea"), JSON.parse($("#text-answer-warnings").text()));
+
         const rowChanged = function(row) {
             const triggerStrings = row.find("select[id$=-trigger_strings]").val();
             const warningTextDe = row.find("textarea[id$=-warning_text_de]").val();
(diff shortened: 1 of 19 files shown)


karyon:
Maybe add an explanation to the form along the lines of "The meaningless text detection is always active and cannot be configured. All characters are normalized to lowercase and any whitespace is normalized to a single space before comparison with the trigger strings."

Maybe allow regex? Has little cost in the code and allows for greater flexibility for the user.

I still think it would be cool to integrate the meaningless text warning into this, so it's configurable via the same mechanism. When allowing regexes, the current rules would translate to something like `'^\W*$', '^\W*k\W*a\W*$', '^\W*n\W*a\W*$'`. It's actually a little horrible. Maybe there's a better way? But I'd still prefer that to "just read the code", and the code is actually not that straightforward either.

evap/staff/templates/staff_text_answer_warnings.html
```diff
@@ -0,0 +1,141 @@
...
+                            <span class="fas fa-exclamation-triangle"></span>
+                            {% trans "This answer will probably be filtered out. You can simply leave the textbox empty." %}
+                        </span>
+                        {% for warning in text_answer_warnings %}
+                            <span class="mt-2 warning-label d-none" data-warning="trigger-string-{{ forloop.counter0 }}">
+                                <span class="fas fa-exclamation-triangle"></span>
+                                {{ warning.warning_text }}
+                            </span>
+                        {% endfor %}
+                    </div>
```
He3lixxx, line 94: This part (obviously) is extemely similar to the element on the vote page. Have you considered factoring it out? It would remove duplication and make sure that the preview always matches the actual form element.

He3lixxx (commented):
Code looks good to me, very pleasant to read.

He3lixxx:
I'm unsure about using regexes as @karyon proposed.
Con: I like clear strings in the UI better than regexes. Humans are bad at parsing regexes.
Pro: I don't like the hardcoded, unconfigurable warning.

@janno42 what's your opinion on that?

evap/staff/templates/staff_text_answer_warnings.html
pixunil: I have considered to factor out the entire text answer row, but not the (smaller) warning text elements. Will do it.

janno42:
i have pretty much the same opinion. i still think that the empty/meaningless textanswer warning can be handled separately and would vote for that option and not using regexes.

evap/student/models.py
```diff
@@ -0,0 +1,19 @@
...
+
+class TextAnswerWarning(models.Model):
+    warning_text_de = models.CharField(max_length=1024, verbose_name=_("Warning text (German)"))
+    warning_text_en = models.CharField(max_length=1024, verbose_name=_("Warning text (English)"))
+    warning_text = translate(en='warning_text_en', de='warning_text_de')
+
+    trigger_strings = ArrayField(models.CharField(max_length=1024), default=list,
+        verbose_name=_("Trigger strings (case-insensitive)"), blank=True)
+
+    order = models.IntegerField(verbose_name=_("Warning ordering"), default=-1)
```
janno42, line 16: this should probably be `Warning order`

janno42 (approved):
The warning textareas could be smaller, I'd say 5 rows would suffice.

evap/static/js/text-answer-warnings.js
```diff
@@ -0,0 +1,45 @@
+function isTextMeaningless(text) {
+    return text.length > 0 && ["", "ka", "na"].includes(text.replace(/\W/g,''));
+}
+
+function doesTextContainTriggerString(text, triggerStrings) {
+    return triggerStrings.some(triggerString => text.includes(triggerString));
+}
+
+function updateTextareaWarning(textarea, textAnswerWarnings) {
+    const text = textarea.val().toLowerCase().replace(/\s{2,}/, " ");
```
janno42, line 10: handling spaces seems not to be consistent. using the trigger string `test_spaces` (`_` standing for a space character) the warning is shown for these texts:
- `this__includes__multiple__test_spaces`
- `this_includes_multiple_test__spaces`

but not for this one:
- `this_includes_multiple__test__spaces`

janno42 (changes requested):

janno42 (approved):
as soon as travis deigns to run the tests and everything is fine, i am happy with this pr :)

janno42 (approved):

PR #1514, opened 2020-10-05
More results caching for staff users

### LINKED ISSUE #1510: Calculate grades only after evaluation period passed
Evaluation results should only be calculated after the evaluation period passed. Staff users should only see grades for evaluations in states `evaluated` and later.

Staff users can access the results page during the evaluation period and see the current votes. Currently they can also see the grade that would be given if the evaluation would end in this moment, but the grade should not be shown neither on results index nor on results detail pages before the evaluation period passed.

karyon:
This should speed up the results index for staff users by a lot. (1300 -> 350ms with the test data, the effect is larger the more evaluations have begun but are not published yet).

First, this caches results also for the states `evaluated` and `reviewed`, since the results in those states doesn't change frequently anymore. The exception is reviewing textanswers, which does change the results. In this PR, the results are refreshed on review_finished. That means that if some textanswers are reviewed already, but others aren't, they still wont show up on the results page if staff users looked at it. 

Sadly, for the assertion that those two states always have the results cached, I had to add a `cache_results` to lots of tests. An alternative would be perhaps to send an email to the admins instead, but then calculate the results on the fly. If we didn't do any of that, I wouldn't be surprised if we'd introduce a silent performance degradation in the future.

The second change is not displaying results on the results index for evaluations that are in_evaluation, which fixes #1510 (and also speeds up the results index).

Best reviewed per commit ;)

### DIFF
--- a/evap/results/tests/test_tools.py
+++ b/evap/results/tests/test_tools.py
@@ -25,14 +25,17 @@ class TestCalculateResults(TestCase):
 
         self.assertIsNotNone(caches['results'].get(get_results_cache_key(evaluation)))
 
-    def test_cache_unpublished_evaluation(self):
-        evaluation = baker.make(Evaluation, state='reviewed')
-        evaluation.publish()
+    def test_caching_lifecycle(self):
+        evaluation = baker.make(Evaluation, state='in_evaluation')
+
+        self.assertIsNone(caches['results'].get(get_results_cache_key(evaluation)))
+
+        evaluation.evaluation_end()
         evaluation.save()
 
         self.assertIsNotNone(caches['results'].get(get_results_cache_key(evaluation)))
 
-        evaluation.unpublish()
+        evaluation.reopen_evaluation()
         evaluation.save()
 
         self.assertIsNone(caches['results'].get(get_results_cache_key(evaluation)))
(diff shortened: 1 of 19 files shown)


He3lixxx (approved):
Very nice.

evap/results/tests/test_tools.py
```diff
@@ -7,30 +7,35 @@
 
 from evap.evaluation.models import (Contribution, Course, Evaluation, Question, Questionnaire, RatingAnswerCounter,
                                     TextAnswer, UserProfile)
-from evap.results.tools import (calculate_average_course_distribution, calculate_average_distribution, collect_results,
-                                distribution_to_grade, get_collect_results_cache_key, get_single_result_rating_result,
+from evap.results.tools import (calculate_average_course_distribution, calculate_average_distribution,
+                                cache_results, get_results,
+                                distribution_to_grade, get_results_cache_key, get_single_result_rating_result,
```
janno42, line 12: please order alphabetically

janno42 (commented):
very nice!
i don't get what you mean by "An alternative would be perhaps to send an email to the admins instead, but then calculate the results on the fly." can you please tell me what email could be send when?

also, there are some codecov and some codacy warnings in case you want to look into those.

karyon:
i ordered the imports, removed some unused ones and fixed one of the codacy issues (though it still shows up, no idea why). I don't think fixing the other two issues improves anything. the codecov comments look ok to me.

janno42 (approved):

PR #1519, opened 2020-10-12
Fix autofocus on email/OpenID button on login page

karyon:
The focus via `$("#id_email")` didn't work when developing locally because there were two email fields. I fixed that.

When OpenID is available, I made it autofocus the OpenID button. Though it looks a little weird and I don't know whether anyone would ever think of using the enter key to press that button... I can also remove it again.

image

### DIFF
--- a/evap/evaluation/templates/index.html
+++ b/evap/evaluation/templates/index.html
@@ -17,7 +17,7 @@
                             {% trans 'Log in using HPI OpenID Connect.' %}
                         </p>
                         <div class="text-center mt-5">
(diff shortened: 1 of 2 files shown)


evap/evaluation/templates/index.html
```diff
@@ -81,10 +81,4 @@ <h5 class="modal-title" id="loginProblemsModalLabel">{% trans 'Login problems' %
...
     </div>
 </div>
-{% endblock %}
-
-{% block additional_javascript %}
-    <script type="text/javascript">
-        $("#id_email").focus();
-    </script>
-{% endblock %}
+{% endblock %}
```
janno42, line 84: please add a newline :)

janno42 (approved):
i'd say having the autofocus on the button is okay

PR #1521, opened 2020-10-19
Excel test data generator

### LINKED ISSUE #1503: Excel test data generator
it might make sense to create some helper methods that create an (in-memory) xls from python data structures. We're adding more and more .xls files to the repo, which are not reviewable and must be manually edited in case the schema changes. we could keep one or two xls files in the tests to make sure we don't break reading "real" xls files.

_Originally posted by @karyon in https://github.com/e-valuation/EvaP/pull/1502#discussion_r479787832_

jb3rndt:
Fixes #1503


janno42 (approved):

jb3rndt:
I used python dictionaries for the different sheets and a 2D-array for each sheet to represent the table data of that sheet.
By using python data structures we do not have to worry about reading files, as we would have to if we used csv files to represent the data for the excel files. Here we can use them directly. I used a dictionary to also save the sheet names because they are relevant in some tests.
I did not delete the excel files yet, so we can still compare the data.
I kept using the original excel files `valid_user_import.xls` and `invalid_user_import.xls` for the tests.

evap/staff/tests/test_importers.py
```diff
@@ -359,3 +353,14 @@ def test_import_new_participant(self):
 
         self.assertEqual(self.evaluation1.participants.count(), 2)
         self.assertEqual(set(self.evaluation1.participants.all()), set([self.participant1, self.participant2]))
+
+def create_memory_excel_file(data):
+    memory_excel_file = io.BytesIO()
+    wb = xlwt.Workbook()
+    for sheet_name, sheet_data in data.items():
+        sheet = wb.add_sheet(sheet_name)
+        for (row_data, row_num) in zip(sheet_data, range(len(sheet_data))):
```
pixunil, line 362: You can replace `zip(x, range(len(x)))` with [`enumerate(x)`](https://docs.python.org/3/library/functions.html#enumerate):
```suggestion
        for (row_num, row_data) in enumerate(sheet_data):
```

evap/staff/fixtures/excel_files_test_data.py
```diff
@@ -0,0 +1,132 @@
+duplicate_user_import_filedata = {
+    "Users": [
```
pixunil, line 2: Just because I’m curious: Do you have a pattern when to use `'` and when `"`? I would prefer to keep them consistently `'`, but that’s not a strong opinion here

evap/staff/tests/test_importers.py
```diff
@@ -25,8 +26,7 @@ def setUpTestData(cls):
             cls.invalid_excel_content = excel_file.read()
         with open(cls.filename_random, "rb") as excel_file:
             cls.random_excel_content = excel_file.read()
-        with open(cls.filename_duplicate, "rb") as excel_file:
```
pixunil, line 28: Just a reminder for your later cleanup that `filename_duplicate` is unused (and I think that no linter will catch this)

pixunil (changes requested):
Nice one! I did not look into the details of the test data itself, but the overall framework is neat.

evap/staff/tests/test_importers.py
jb3rndt: Nice, Thanks 👍

evap/staff/fixtures/excel_files_test_data.py
jb3rndt: Nope, I got no pattern 😅

He3lixxx (approved):
Thanks!

pixunil (approved):

janno42:
in the meantime we have a new test file where we can't just use the new method because it must have specific data types in certain cells. we should probably just keep the excel file in this special case.

karyon:
can we make the generator put those specific datatypes into the in-memory excel file?

evap/staff/tests/test_importers.py
```diff
@@ -1,17 +1,19 @@
...
 from evap.evaluation.models import Course, Degree, UserProfile, Semester, Evaluation, Contribution, CourseType
 from evap.staff.importers import UserImporter, EnrollmentImporter, PersonImporter, ImporterError, ImporterWarning
 from evap.staff.tools import ImportType
+import evap.staff.fixtures.excel_files_test_data as excel_data
 
 
 class TestUserImporter(TestCase):
     filename_valid = os.path.join(settings.BASE_DIR, "staff/fixtures/valid_user_import.xls")
-    filename_duplicate = os.path.join(settings.BASE_DIR, "staff/fixtures/duplicate_user_import.xls")
     filename_invalid = os.path.join(settings.BASE_DIR, "staff/fixtures/invalid_user_import.xls")
```
karyon, line 17: why not replace filename_valid and filename_invalid as well?

jb3rndt:
It works with no programmatical overhead... So we could do it, but I understand if you would want to keep the original file to make sure it gets read properly.
(I also updated the branch to have the new file and tests... thats the cause of that merge conflict commit)

janno42 (approved):

PR #1525, opened 2020-10-19
Importer type checking

### LINKED ISSUE #1522: Excel importer: Data types of cells. Float cells might cause import errors
Successor of #1493.

With the current code, we had occurences of
```
AttributeError
'float' object has no attribute 'strip'
/evap/evap/staff/importers.py in __init__, line 40
```
which we think is caused by the excel file cell being set to type "float", so xlrd gives us a float value instead of a string value, but we expect it to be a string.

@niklasmohrin had some proposals on what to do in #1499. I think our last consent is given in https://github.com/e-valuation/EvaP/pull/1499#issuecomment-688454822 by @janno42:
> we just use strings, so every cell could be parsed as string value, regardless of the excel data type.

niklasmohrin:
Closes #1522

This is what was proposed in #1499 plus a test.


He3lixxx (approved):

janno42 (approved):

janno42:
Thanks :)

Issue #1524, opened 2020-10-19
Add evaluation period check when rendering "Evaluate" button


He3lixxx:
**Edit: See latest comment for up-to-date description of what this issue is about**

---

With the current testdata, you can do the following: reload testdata, login as evap, go to student index page, click on "Evaluate" for "Game Programming".

You will get access denied, triggered by these two lines in evap/student/views.py:138:
```python
    if not evaluation.can_be_voted_for_by(request.user):
        raise PermissionDenied
```
I think we can agree that in general, the evaluation should not be shown on the student index page, or the student should be able to access it.

What happened here is that the cron job that was supposed to move the evaluation from state `in_evaluation` to `evaluated` didn't run (in the dev vm). We rely on it to run on more places than just here, and there might be a lot of other stuff that can break if the cron jobs stop running. However, I think there are not that many that are in direct contact with students, which is the majority of our users.

Still, I think we should make sure that one of the following holds:
- If the cronjob should have run but didn't run, notify someone (`assert evaluation.can_be_voted_for_by(request.user)` for each evaluation in the student index page would do that)
- Filter out evaluations whose evaluation period has ended, even if their state is still `in_evaluation`, on the student index view. This would allow the platform to continue running even if cron fails for some reason.

@janno42, @karyon, @pixunil, @felixrindt what do you think?

janno42:
an assertion is perfectly fine for me.
if we would do state checks on the student index page i would say we should do them everywhere. but i think the assert is sufficient and avoids too much code.

He3lixxx:
Hmm, we should keep in mind that adding an insertion will actually make the error worse, when it occurs. Right now, users will the evaluation that has already ended, but as long as they don't click on it, everything is fine. If we add the assertion, the student index page will 500 instead of giving the semi-correct version (which in this case would still be better).

He3lixxx:
Me and @janno42 agreed that the student index page with the "Evaluate" button is probably the only place where we have two different sources of state (system time and cron job execution state) that could collide (if cron job execution is delayed). Thus, adding another check on system time, just as the vote view does, on the student index page, and not showing that "Evaluate"- button if the time is passed should be an okay solution.

PR #1526, opened 2020-10-19
Change test data timestamps for the next two years

He3lixxx:
(Do we want to set this for a longer future period?)

temporary fix for what makes us observe #1524 with the current test data


janno42:
please make it 2099 like other evaluations already use

janno42 (approved):

PR #1527, opened 2020-10-20
Remove last_modified fields for Course and Evaluation

janno42:
after #1424 now logs changes, we don't need the `last_modified` fields anymore


karyon (commented):
Looks good to me, but the removed tests made me concerned.

I'm a bit undecided whether testing the logging code on the form level would still be beneficial: on the one hand this would properly verify a specific user action is resulting in the expected log entries, on the other hand the logging works on such a low level that unit-tests would work as well. What do you think?

Revisiting the unit tests of the logging code, I think they are rather spotty, and with the complexity of the code and low knowledge distribution among us, proper tests would be even more important. Here's what I think is missing (at least): saving without changes, saving with changes in ignored properties, saving with changes, saving a change in a related object, delete related object. At least least verify that something was logged (or not), ideally also verify that the right thing was logged. Maybe we should go through all the probably actions in the logging code and collect more cases that need to be tested.

janno42:
True, i put that into a new issue (#1528)

karyon:
What do you think about the form-level tests?

janno42:
the unit tests should be fine, i wouldn't focus on the forms here. as you said logging is so deeply integrated that we probably don't need to test the forms independently.

karyon (approved):

Issue #1515, opened 2020-10-05
Users might receive a stale datagrid.js / Rethink JS asset handling


karyon:
datagrid.js (and also text-answer-warning.js from #1513) is currently served from the static directory without being processed by django-compressor. apache serves those files with a cache expiry date of `access plus 1 month`. I suspect that when we modify datagrid.js, clients might use an old cached version for up to a month.

django-compressor doesn't play well with ES6 modules I'm afraid. There's a PR open so you can output a <script type="module ...>, but this would even help with the `import from ...` style we're currently using.

What if we removed django-compressor? Here's what we use it for and what we could do instead:

1. concatenation
    * instead: enable HTTP/2. we should do that anyways, and concatenating assets is not a best practice with that anymore.
2. minification
    * instead: do nothing. our dependencies are minified either way, and the couple KB of JS that we hand-write might as well go without minification.
    * while we are at it, we could enable brotli. ideally, we would pre-compress our static assets and configure apache to use those.
3. cache-busting (appending hashes to filenames of the assets)
    * Instead: use [ManifestStaticFilesStorage](https://docs.djangoproject.com/en/3.1/ref/contrib/staticfiles/#manifeststaticfilesstorage), which does the same
4. scss processing
    * Instead: (not so sure about this one) we add a `./manage.py run scss [--watch]`, which compiles the scss into static_collected. it must be called whenever collectstatic is called, and when editing scss files. I'm not 100% sure that would even work, does django's development server actually looks into static_collected...?
    * Or we keep django-compressor specifically for scss files, but that feels kind of wrong.

Any other ideas/input?

pixunil:
First, my input to your points:

1. concatenation
    - I agree with you that concatenating will be rather obsolete through HTTP/2. I haven’t researched a lot on this topic, but afaik this can be achieved by a better reuse of connections. In a second stage, we could make use of Server Push, which may need additional configuration.
2. minification
    - I would also add that there exists minifiers / bundlers for JS, which could also make some polyfills automatically (if necessary)
3. cache-busting
    - For appending hashes, this looks like a sensible alternatve, but I think we may want an entirely different approach as changing names doesn’t work nicely with `import from`
4. scss processing
[...]

PR #1529, opened 2020-10-21
Static assets rework

pixunil:
Addresses #1515. This implements all base requirements except HTTP/2 (which I’ll delegate to @karyon) and is working for `localhost:8001` (Note: if you use <kbd>F5</kbd>, you may force the browser to fetch the resources, normal navigation through the page shows the expected behaviour). All code can be reviewed and I’ll be happy if you have ideas for my open questions.

Remember to provision your virtual machine to use the new configurations:
```bash
vagrant provision
```

This adds a new command `management.py scss`, if you don’t work actively with style sheets, you are all set if you run `management.py run` after a `git pull`, when actively developing, you should use `management.py scss --watch`.

- [x] Remove some unneeded files
- [x] Replace `sassc` with `dart-sass` (as it has better long-term support and has a `--watch` option)
- [x] Look into better ways to acquire `dart-sass`, the `wget | tar` command feels a bit hacky (resolution: seems fine for development purposes)
- [x] Error handling when `dart-sass` is not available on production (resolution: will print an error during deployment, but won’t break it)
- [x] Look into why we need to use `sass --watch --poll` for effectively watching for changes, I suspect that we need polling as watching via operating system events doesn’t work in a virtual machine. **Result**: Polling is necessary in virtual machines, leaving it as-is.
- [x] Add a new `management.py scss` utility to compile stylesheets, supporting `--watch` and `--release` flags
- [x] Consistently use `{% static %}` instead of `{% get_static_prefix %}`
- [x] Use `ManifestStaticFilesStorage` to append hashes to files
- [x] Look into how we could get rid of the additional webfont formats (they are referenced by fontawesome, but according to our compatibility rule of modern browsers, [we should get away with only `woff2`](https://css-tricks.com/snippets/css/using-font-face/#practical-level-of-browser-support))
- [x] Remove compressor with its configuration and calls (I have not a good overview about this, so I haven’t touched this yet)

- [x] Change server configuration to set a rather huge `Expires` header for static files with hash and a rather small one for those without hash
- [x] Check how we can satisfy Codacy


karyon:
nice!

regarding removing compressor, a ctrl-f `compress` should point you to all the locations. should be a couple templates (the tag is called `compress`), the settings file (the settings are all prefixed with `COMPRESS_`), and some calls to `manage.py compress`.

evap/evaluation/templates/sortable_form_js.html
```diff
@@ -51,7 +51,7 @@
             formTemplate: (usesTemplate ? ".form-template" : null)
         });
 
-        $('#' + tableId + " tbody").sortable({
```
karyon, line 54: 100% worth it!

karyon:
why exactly did those font formats become necessary now?

karyon:
to expand on your suspicion, the --poll might be necessary because watching files in the mirrored vagrant directory might not work.

karyon:
no idea why the caching might not work for you. if you want, we can have a look together.

He3lixxx:
Cool stuff!

>  Look into better ways to acquire dart-sass, the wget | tar command feels a bit hacky

It is hacky, but then again, the whole VM is set up in a hacky way. I still think it's fine.

> Error handling when dart-sass is not available on production (should the deployment fail then?)

We at least need it on every production update to compile the new css files, right? In that case, I think it's fine to fail if it's missing.

> Look into why we need to use sass --watch --poll for effectively watching for changes, I suspect that we need polling as watching via operating system events doesn’t work in a virtual machine.

@karyon is probably right here. Virtualbox will not forward any filesystem change events (https://www.virtualbox.org/ticket/10660), so we will have to rely on polling.

> Check how we can satisfy Codacy

The current failure is due to two messages from Bandit about using subprocess. I think using subprocess is fine there, so I would simply add a `# nosec` to these lines -- see https://github.com/PyCQA/bandit#exclusions.

I'm currently looking into the pylint behaviour. I didn't find a bug report for the specific error we are receiving, however, I can see how the line that it lints confuses it. I think the main problem with pylint currently is that it does not run in a deterministic way if running with multiple processes, meaning that some messages might not occur in some execution orders. We might have to fall back to set it to always just run with one runner. In the meantime, if the false-positive here keeps being an issue, feel free to just globally exclude that error in .pylintrc.

He3lixxx (approved):
Very cool 👍

evap/evaluation/tests/test_commands.py
```diff
@@ -137,52 +137,58 @@ def test_single_result_anonymization(self):
...
+        self.assertEqual(mock.call_count, Evaluation.objects.count())
 
-class TestReloadTestdataCommand(TestCase):
-    @patch('builtins.input')
-    @patch('evap.evaluation.management.commands.reload_testdata.call_command')
-    def test_aborts(self, mock_call_command, mock_input):
-        mock_input.return_value = 'not yes'
 
-        management.call_command('reload_testdata', stdout=StringIO())
+class TestScssCommend(TestCase):
```
He3lixxx, line 150: Commend

deployment/update_production.sh
```diff
@@ -31,6 +31,9 @@ set -x # print executed commands. enable this here to not print the if above.
 
 sudo -H -u $USERNAME git fetch
 
+# install sass
+./deployment/install_dart_sass.sh
+
```
karyon, line 36: I think this is not the right place, at least we don't have e.g. a pip install in here either. I suppose a "setup production" script simply doesn't exist at the moment.

evap/evaluation/management/commands/scss.py
```diff
@@ -0,0 +1,39 @@
...
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--watch', action='store_true',
+            help='Watch stylesheets and recompile when they change.',
+        )
+        parser.add_argument(
+            '--release', action='store_true',
+            help='Compress output stylesheet and do not generate source maps.'
+                 ' Indented to use in production deployment.',
```
karyon, line 17: Indented ;)

evap/development/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.urls import path
+
+from evap.development import views
+
+urlpatterns = [
+    path("components", views.development_components, name="development_components"),
```
karyon, line 6: I would suggest `dev_helpers` as a more obvious name, but maybe that's just personal preference.

evap/development/management/commands/run.py
```diff
@@ -9,6 +9,8 @@ class Command(BaseCommand):
     help = 'Execute "runserver 0.0.0.0:8000"'
 
     def handle(self, *args, **options):
+        self.stdout.write('Executing "manage.py scss"')
+        execute_from_command_line(['manage.py', 'scss'])
```
karyon, line 13: how fast is that?

deployment/load_production_backup.sh
```diff
@@ -47,6 +47,7 @@ sudo -H -u $USERNAME $ENVDIR/bin/pip install -r requirements.txt
 
 # compilemessages and compress regularly fail without any real issue.
 sudo -H -u $USERNAME $ENVDIR/bin/python manage.py compilemessages || true
+sudo -H -u $USERNAME $ENVDIR/bin/python manage.py scss --release
```
karyon, line 50: that doesn't really fit into loading backups. as you already added it to update_production.sh, I'd remove it here.

evap/evaluation/management/commands/scss.py
```diff
@@ -23,6 +28,9 @@ def handle(self, *args, **options):
         if options['watch']:
             command += ['--watch', '--poll']
 
+        if options['release']:
```
karyon, line 31: maybe call this "production"? i think that's how it's usually called in the javascript world (e.g. in webpack or rollup), "release" sounds more like compiled languages

deployment/provision_vagrant_vm.sh
```diff
@@ -46,6 +46,7 @@ sudo -H -u $USER $ENV_FOLDER/bin/pip install mod_wsgi
 
 # setup apache
 a2enmod expires
+a2enmod headers
```
karyon, line 49: https://stackoverflow.com/questions/5799906/what-s-the-difference-between-expires-and-cache-control-headers sounds like Cache-Control more or less superseded Expires. maybe we can remove the latter?

deployment/update_production.sh
pixunil: Hm, so should this only be documented somewhere else (e.g. in the wiki) or should there be a new `setup_production.sh`?

evap/evaluation/management/commands/scss.py
pixunil: That’s a tricky flip-flop of characters ^^

deployment/load_production_backup.sh
pixunil: Afaik, `collectstatic --noinput` had the same intent – to update all static assets, including sass files. And I can’t quite see where we are calling `update_production.sh` after `load_production_backup.sh`, so I’m not sure where I would put this call if not here (without changing the script structure).

evap/development/management/commands/run.py
pixunil: On my machine™ around 1s normally, but I experienced a runtime of 2s. I think this is fine for `scss`, but I need to investigate something for the follow-up PR as typescript takes around 7s which is more noticable. This is just convenience so that the workflow does not change “just run `management.py run` and you are all set”.
I’m tending to run those in parallel (rather unlikely that the first browser request is made before sass has finished), but I don’t know how I would produce some sensible output.

deployment/load_production_backup.sh
karyon: yeah, on a second look it seems like load_production_backup is meant to do a full rollback and not just load the backup.

deployment/update_production.sh
karyon: the number of times we had to spin up a new production instance in the last 8 years has been zero i think :D if we ever need to do that, the vagrant provision script should contain mostly complete instructions.

speaking of which, could you add this command there? :)

pixunil: I now inlined the content of `install_dart_sass.sh` into `provision_vagrant.sh`. Just to be sure – the command for installing dart has always have been in `provision_vagrant.sh`, so I don’t know whether I understood your last question correctly?

deployment/provision_vagrant_vm.sh
pixunil: Only disadvantage I could imagine would be that the numbers are not that readable as they were before. Sample configuration (not tested nor commited yet):
```
    <Directory ${REPO_FOLDER}/evap/static_collected>
        Header set Cache-Control "max-age: 10800" // 3 hours

        <FilesMatch "\.[0-9a-f]{12}\.\w+$">
            Header set Cache-Control "max-age: 31536000, immutable" // 1 year
        </FilesMatch>
    </Directory>
```

evap/development/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.urls import path
```
He3lixxx, line 1: If codecov complains that the lines here and in views.py are uncovered: I'd vote in favor of ignoring these files, I consider them to be expected to be untested. https://docs.codecov.io/docs/ignoring-paths

deployment/provision_vagrant_vm.sh
karyon: looks good to me!

deployment/update_production.sh
karyon: sorry, i thought you added it only to update_production.sh. I think inlining into provision_vagrant.sh is fine until we need it in another place.

karyon: seems like you need it for the new travis.yml in the typescript PR...?

pixunil: Good catch, then I’ll revert to the version with an extra shell script

deployment/provision_vagrant_vm.sh
pixunil: Despides some modifications to the syntax (comments need an own line and `max-age` wants a `=` not a `:` after it), I tested this locally and seams to work

evap/development/urls.py
pixunil: @janno42 your opinion here? :cat:

janno42: helpers sounds good to me, but components is also fine :)

karyon:
hey @pixunil do you need anything here? anything I can help you with?

PR #1531, opened 2020-10-27
Fix access denied on manager login, fixes #1523

### LINKED ISSUE #1523: Access denied on manager login
Currently, after logging in, a manager is redirected to /staff/, but staff mode will not be active, so they will get a 403 access denied.

@janno42 what behavior do we want here? Redirect as if they weren't a manager or enable staff mode?

karyon:
The evaluation:index view was logging the user in and would redirect them immediately afterwards. Since no new page load happened, the staff mode middleware wouldn't run before determining the redirect target. By reloading the index page, this is fixed at the expense of an additional HTTP request.

On production, this doesn't change anything, as the only login methods there (OpenID and login url) don't have this issue.

fixes #1523

### DIFF
--- a/evap/evaluation/views.py
+++ b/evap/evaluation/views.py
@@ -74,6 +74,7 @@ def index(request):
             # clean up our test cookie
             if request.session.test_cookie_worked():
                 request.session.delete_test_cookie()
+            return redirect('evaluation:index')
 
     # if not logged in by now, render form
     if not request.user.is_authenticated:


janno42 (approved):

He3lixxx:
Do we want to add a regression test? I feel like this should have been caught by a test.

karyon:
I created https://github.com/e-valuation/EvaP/issues/1536.

PR #1532, opened 2020-11-02
Update pylint

He3lixxx:
I kind of hope that the unexplainable "unused local variable: <lambda>" errors go away with updating pylint, we might have to add two ignores on that line: one to suppress the spontaneus false positive, and one to suppress the error that happens if the disable command is there, but the error is not detected in that run.

### DIFF
--- a/evap/evaluation/models_logging.py
+++ b/evap/evaluation/models_logging.py
@@ -41,10 +41,11 @@ class LogJSONEncoder(JSONEncoder):
     As JSON can't store datetime objects, we localize them to strings.
     """
 
-    def default(self, obj):
-        if isinstance(obj, (date, time, datetime)):
-            return localize(obj)
-        return super().default(obj)
+    def default(self, o):
+        # o is the object to serialize -- we can't rename the argument in JSONEncoder
+        if isinstance(o, (date, time, datetime)):
+            return localize(o)
+        return super().default(o)
 
 
 def _choice_to_display(field, choice):  # does not support nested choices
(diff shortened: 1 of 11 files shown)


evap/evaluation/models_logging.py
```diff
@@ -41,10 +41,11 @@ class LogJSONEncoder(JSONEncoder):
     As JSON can't store datetime objects, we localize them to strings.
     """
 
-    def default(self, obj):
-        if isinstance(obj, (date, time, datetime)):
-            return localize(obj)
-        return super().default(obj)
+    def default(self, o):
+        """ o is an object instance """
```
karyon, line 45: why?

karyon (approved):

evap/evaluation/models_logging.py
He3lixxx: `signature-differs` triggered. In the parent class, the argument is named `o`, not `obj`. I forgot to undo this when I disabled `signature-differs`. (Kind of makes sense though, substitution principle and keyword arguments - a call with `instance.default(o=False)` wouldn't work with this child class)

janno42 (approved):

He3lixxx:
The weird false-positive by pylint in `translate` still occurred on travis with pylint 2.6.0. As it only happens on travis and I can't reproduce it locally, I don't feel like filing an issue with pylint, so I just put an ignore statement in that method.

@karyon: My statement about the `o` vs `obj`was wrong, it wasn't `signature-differs` that triggered, but `arguments-differ`. Thus, I added the change you annotated again (I changed `obj` to `o` and added a comment on why we use that (bad) name).

Issue #1426, opened 2020-02-10
Auto-save information


janno42:
Answers in the student vote form are auto-saved locally. There is a button that can be clicked for manually saving locally but it is more or less useless, because the save is already done automatically on every form change.
Users wanted an "easier auto-save" by using hotkeys - we should instead show the information about the automatic saves in a more detailed way and remove the button for saving manually.

A small info text like `Last auto-save in browser: <Timestamp>` should be displayed instead, giving more details about the continuous saving. Info that the evaluation can be continued another time when using the same browser on the same device should be placed in a tooltip.

felixrindt:
I'd like something human for the timestamp like "3 seconds ago", similar to django's [humanize](https://docs.djangoproject.com/en/3.0/ref/contrib/humanize/#naturaltime).

janno42:
@ScrappyDieRaupe do you still want to work on this?

ScrappyDieRaupe:
Feel free to unassign me if it needs to be done quickly. 
I might work on this some time in the future.

PR #1533, opened 2020-11-02
Replace save button on student vote with infotext

Sohn123:


karyon (approved):
code looks good, @janno42 re strings?

janno42:
Looks good in general, but I'm not yet very happy with the user experience. Currently you are showing the time when the form was last saved as HH:MM. I would argue that this precision is too low, because users can enter a lot of information in multiple fields within a minute. So to make sure that the form was really saved some people might wait on the page for about a minute, change a thing and check whether it was saved again.

Just adding the seconds doesn't feel right either though. As [commented on the issue](https://github.com/e-valuation/EvaP/issues/1426#issuecomment-584607224) a continuously updated humanized time stamp might help. Unfortunately, this would mean a bit more work. We would need some JavaScript code updating the information and can't directly access Django's `naturaltime` from there. @pixunil do you have additional thoughts about that?

karyon:
It's probably not thaaat much work. To save on typing, you could start with copypasting something like this [SO answer](https://stackoverflow.com/a/7641812/4312944), but it's just a couple lines. And then adjust the strings to what we want.

janno42:
I think the relative time should be displayed as follows:
- < 3 seconds: `just now`
- < 60 seconds: `N seconds ago`
- < 30 minutes: `N minutes ago`
- < 12 hours: `at HH:MM`
- \>= 12 hours: `at yyyy-mm-dd HH:MM`

We need to pay attention to correct pluralization ("1 minute ago") and moving parts in translations ("vor 1 Minute"). You could for example use the [Intl.RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat).

janno42:
i was playing with strings and positioning on the page and noticed there were some problems with the time stamp formatting because javascript is weird. i pushed some fixes for that and hope you are fine with the changes, @Sohn123.

evap/student/templates/student_vote.html
```diff
@@ -194,12 +197,52 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
+                    const delta = Math.round((timeNow - lastSavedDate) / 1000)
+                    const languageCode = "{{ LANGUAGE_CODE }}"
+                    const relativeTimeFormat = new Intl.RelativeTimeFormat(languageCode)
+                    let timeStamp = "{% trans 'unknown' %}"
+                    if (delta < 3) {
+                        timeStamp = "{% trans 'just now' %}";
+                    } else if (delta < 60) {
+                        timeStamp = relativeTimeFormat.format(-delta, 'seconds');
+                    } else if (delta < 60 * 30) {
+                        timeStamp = relativeTimeFormat.format(-Math.round(delta / 60), 'minutes');
```
karyon, line 221: add "ago" here and above?

evap/student/templates/student_vote.html
```diff
@@ -173,7 +176,7 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
 
         <script type="module">
             import {initTextAnswerWarnings} from "{% static 'js/text-answer-warnings.js' %}";
-
+            const lastSavedStorageKey = "student-vote-{{evaluation.id}}-{{request.user.id}}"
```
karyon, line 179: since you added the evaluation id... should that also be added to sysiphus' customKeySuffix?

evap/student/templates/student_vote.html
janno42: that is done by the relative time format :)

evap/student/templates/student_vote.html
janno42: for the vote form fields the evaluation id is already part of the key because sisyphus put the complete url into the key. or do you mean a different place?

karyon: no, that sounds good.

karyon (approved):

evap/student/templates/student_vote.html
```diff
@@ -194,12 +197,52 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
+                    const lastSaved = $('#last-saved');
+                    const timeNow = new Date();
+                    localStorage.setItem(lastSavedStorageKey, timeNow.toString());
+                },
             });
+
+            function updateLastSavedLabel() {
+                    const timeNow = new Date();
+                    const lastSavedLabel = $('#last-saved');
+                    const lastSavedDate = new Date(localStorage.getItem(lastSavedStorageKey))
```
pixunil, line 211:
```suggestion
                    const lastSavedDate = new Date(localStorage.getItem(lastSavedStorageKey));
```
:paintbrush: Playing manual linter here: There are some semicolon missing (multiple occurences)

*****

Also, you should check for `localStorage.getItem(lastSavedStorageKey) !== null` here (see below)

evap/student/templates/student_vote.html
```diff
@@ -194,12 +197,52 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
                     if (textResultsPublishConfirmation.bottomCard) {
                         updateTextResultsPublishConfirmation();
                     }
-                }
+                },
+                onSave: function () {
+                    const lastSaved = $('#last-saved');
```
pixunil, line 202:
```suggestion
```
:paintbrush: It’s unused, right?

evap/student/templates/student_vote.html
```diff
@@ -194,12 +197,52 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
+                    let timeStamp = "{% trans 'unknown' %}"
+                    if (delta < 3) {
+                        timeStamp = "{% trans 'just now' %}";
+                    } else if (delta < 60) {
+                        timeStamp = relativeTimeFormat.format(-delta, 'seconds');
+                    } else if (delta < 60 * 30) {
+                        timeStamp = relativeTimeFormat.format(-Math.round(delta / 60), 'minutes');
+                    } else if (delta < 60 * 60 * 12) {
+                        timeStamp = padWithLeadingZeros(lastSavedDate.getHours()) + ":" + padWithLeadingZeros(lastSavedDate.getMinutes());
+                    } else {
```
pixunil, line 224: I’d check here that there is a sensible `delta` here. If – for any reason – the local storage item is empty, this code will print out 1970-01-01 01:00 as saved date. (Or for the value `""`: NaN-NaN-NaN NaN:NaN).

evap/student/templates/student_vote.html
pixunil, line 179:
```suggestion
            const lastSavedStorageKey = "student-vote-last-saved-at-{{evaluation.id}}-{{request.user.id}}";
```
:paintbrush: Just to avoid possible collisions if we want to add more local storage items for this page

pixunil (changes requested):
One small request for a missing storage item (which might happen when local storage is deactivated), and some code style notices (:paintbrush:). Otherwise, looks great!

evap/student/templates/student_vote.html
He3lixxx: NaN-NaN-NaN NaN:NaN Batman!

https://www.destroyallsoftware.com/talks/wat

Ok I'll show myself out.

Sohn123: Should be resolved with newest commit

pixunil (approved):

evap/student/templates/student_vote.html
```diff
@@ -173,7 +177,7 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
 
         <script type="module">
             import {initTextAnswerWarnings} from "{% static 'js/text-answer-warnings.js' %}";
-
+            const lastSavedStorageKey = "student-vote-last-saved-at-{{evaluation.id}}-{{request.user.id}}";
```
janno42, line 180:
```suggestion
            const lastSavedStorageKey = "student-vote-last-saved-at-{{ evaluation.id }}-{{ request.user.id }}";
```

evap/student/templates/student_vote.html
```diff
@@ -150,12 +150,16 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
+                            <span id="last-saved"></span>
+                            <span class="fa fa-info-circle" data-toggle="tooltip" title="{% trans 'The evaluation can be continued later using the same device and the same browser. But you have to submit it to send it to the server and make it count. After submitting, you can not edit the evaluation anymore.' %}"></span>
+                        </div>
                         <button id="vote-submit-btn" type="submit" class="btn btn-success">{% trans 'Submit questionnaire' %}</button>
                         <p id="submit-error-warning" style="display: none">
                             <span class="fas fa-exclamation-triangle"></span>
                             {% blocktrans %}The server can't be reached. Your answers have been stored in your browser and it's safe to leave the page. Please try again later to submit the questionnaire.{% endblocktrans %}
                         </p>
+                        <div id="last-saved" class="text-secondary pt-2" data-toggle="tooltip" title="{% trans 'This evaluation can be continued another time using the same device and the same browser'%}"></div>
```
janno42, line 162: this line should be removed, doesn't it?

janno42 (changes requested):

evap/student/templates/student_vote.html
Sohn123: Jap

janno42 (approved):
Thank you! 👍

PR #1535, opened 2020-11-06
Fix Javascript on student vote

### LINKED ISSUE #1534: JavaScript on student vote page does not work properly
When clicking on the "I can't give feedback about this contributor" button, nothing happens.
All voting answers should receive the selection "no answer" and the contributor should be collapsed.

The bug occurs in the current Firefox version (82.0.2) on Windows as well as in Safari on MacOS.

pixunil:
Fixes #1534 :see_no_evil: 

The issue was that functions are not visible anymore when using `type=module` which broke the `onclick` and `onchange` attributes.
Thus, I replaced them with `addEventListener` (and migrated some code away from jQuery, except the save button which gets reworked in #1533). If you don’t like touching the code in the way I did it, I could also move all called functions into a non-module `<script type="text/javascript">`.

Yet another reason to look into frontend tests (which I am on it)…

Places I tested:

- http://localhost:8000/student/vote/1634 (has a “small number of participants warning”)
- http://localhost:8000/staff/semester/21/evaluation/1525/preview

### DIFF
--- a/evap/student/templates/student_vote.html
+++ b/evap/student/templates/student_vote.html
@@ -55,7 +55,7 @@
 
                         <div class="form-check mt-3">
                             <input class="form-check-input" id="text_results_publish_confirmation_top" type="checkbox"
-                                name="text_results_publish_confirmation_top" onchange="textResultsPublishConfirmationTopChanged();" />
+                                name="text_results_publish_confirmation_top" />
                             <label class="form-check-label" for="text_results_publish_confirmation_top">
                                 {% trans 'Show my text answers to the contributors even if I remain the only person who takes part in this evaluation.' %}
                             </label>
@@ -98,7 +98,10 @@
                                     </a>
                                 </div>
                                 <div>
-                                    <button type="button"{% if preview %} disabled{% endif %} class="btn btn-light btn-sm" onclick="markNoAnswerFor({{ contributor.id }});">
+                                    <button type="button" class="btn btn-light btn-sm"
+                                        data-mark-no-answers-for="{{ contributor.id }}"
+                                        {% if preview %}disabled{% endif %}
+                                    >
                                         {% blocktrans %}I can't give feedback about this contributor{% endblocktrans %}
                                     </button>
                                 </div>
@@ -133,7 +136,7 @@
 
                     <div class="form-check mt-3">
                         <input class="form-check-input" id="text_results_publish_confirmation_bottom" type="checkbox"
-                            name="text_results_publish_confirmation_bottom" onchange="textResultsPublishConfirmationBottomChanged();" />
+                            name="text_results_publish_confirmation_bottom" />
                         <label class="form-check-label" for="text_results_publish_confirmation_bottom">
                             {% trans 'Show my text answers to the contributors even if I remain the only person who takes part in this evaluation.' %}
                         </label>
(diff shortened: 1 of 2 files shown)


He3lixxx (approved):

evap/student/templates/student_vote.html
```diff
@@ -98,7 +98,10 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
                                     </a>
                                 </div>
                                 <div>
-                                    <button type="button"{% if preview %} disabled{% endif %} class="btn btn-light btn-sm" onclick="markNoAnswerFor({{ contributor.id }});">
+                                    <button type="button" class="btn btn-light btn-sm"
+                                        data-mark-no-answers-for="{{ contributor.id }}"
+                                        {% if preview %} disabled{% endif %}
```
janno42, line 103: you can remove the space here

evap/student/templates/student_vote.html
```diff
@@ -240,29 +249,34 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
-                // check "no answer" and activate the button
-                $("#vote-area-" + contributorId + " div.vote-inputs [type=radio][value=6]").prop('checked', true);
-                $("#vote-area-" + contributorId + " div.vote-inputs [type=radio][value=6]").parent().addClass("active");
+                    // hide questionnaire for contributor
+                    $(voteArea).collapse("hide");
 
-                // hide questionnaire for contributor
-                $("#vote-area-" + contributorId).collapse("hide");
+                    // setting the 'checked' prop does not trigger a change event, so manually save all data here
+                    sisyphus.saveAllData();
```
janno42, line 264: you are not setting the `checked` property manually anymore. is the manual save still required?

janno42 (changes requested):
When clicking the bottom checkbox in the "Small number of participants" warning, the card should not disappear, only when the top checkbox is clicked.

evap/student/templates/student_vote.html
pixunil: Nice catch, sisyphus saves implicitly :)

evap/student/templates/student_vote.html
```diff
@@ -273,16 +287,19 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
-                $("#text_results_publish_confirmation_top").prop('checked', $("#text_results_publish_confirmation_bottom").prop('checked'));
+            if (textResultsPublishConfirmation.bottomCard) {
+                textResultsPublishConfirmation.top.addEventListener("change", () => {
+                    const isChecked = textResultsPublishConfirmation.top.checked;
+                    textResultsPublishConfirmation.bottom.checked = isChecked;
+                    textResultsPublishConfirmation.bottomCard.classList.toggle("d-none", isChecked);
+                });
+                textResultsPublishConfirmation.bottom.addEventListener("change", () => {
+                    textResultsPublishConfirmation.top.checked = textResultsPublishConfirmation.bottom.checked;
+                    sisyphus.saveAllData();
```
janno42, line 301: please add a comment why this is needed

pixunil: I came up with this:
```
// The top checkbox should only be visually checked without triggering the change event,
// which would hide the bottom card.
// To keep the top checkbox checked (after a reload or submit), save the form manually.
```
Too long? Too verbose? Then I’d like another proposal (`my` :gb: `is` :exclamation: :yellow_square: `from` :egg: )

janno42: it's fine 👍

evap/student/templates/student_vote.html
```diff
@@ -182,8 +191,8 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
                     $("div.vote-inputs input[type=radio]:checked").parent().addClass("active");
 
                     // restore publish confirmation state
-                    if($("#text_results_publish_confirmation_top").prop('checked')) {
-                        $("#bottom_text_results_publish_confirmation_card").hide();
+                    if (textResultsPublishConfirmation.bottomCard) {
+                        updateTextResultsPublishConfirmation();
                     }
```
pixunil, line 196: Just a note why I extracted `updateTextResultsPublishConfirmation` here: Only asking for `textResultsPublishConfirmation.top.checked` will raise an error when the checkbox doesn’t exist and I thought that calling a shared function makes it clearer what state we are restoring

janno42 (approved):
thanks :)

Issue #1508, opened 2020-09-10
Skip evaluation in quick review


janno42:
When reviewing text answers in the quick view there should be an option to skip the current evaluation in case reviewers are involved themselves or they don't want to review this particular evaluation for any other reason. The skipped evaluation should then not be shown to this staff user for review again at least for some time.

He3lixxx:
fixed in #1537

PR #1537, opened 2020-11-16
Add 'Skip evaluation' button to quick review mode

rgwohlbold:
This pull request adds a "Skip evaluation button" to the quick review mode, see #1508 .

### DIFF
--- a/evap/staff/templates/quick-review.js
+++ b/evap/staff/templates/quick-review.js
@@ -22,20 +22,22 @@ $(document).ready(() => {
     });
 
     $(document).keydown(event => {
-        let actions = {
-            37: "[data-slide=left]",          // arrow left
-            39: "[data-slide=right]",         // arrow right
-            74: "[data-action=publish]",      // j
-            75: "[data-action=make_private]", // k
-            76: "[data-action=hide]",         // l
-             8: "[data-action=unreview]",     // backspace
-            13: "[data-url=next-evaluation]", // enter
-            77: "[data-startover=unreviewed]",// m
-            78: "[data-startover=all]"        // n
+        const actions = {
+            "arrowleft":    "[data-slide=left]",
+            "arrowright":   "[data-slide=right]",
+            "j":            "[data-action=publish]",
+            "k":            "[data-action=make_private]",
+            "l":            "[data-action=hide]",
+            "backspace":    "[data-action=unreview]",
+            "enter":        `[data-url=next-evaluation][data-next-evaluation-index=${nextEvaluationIndex}]`,
+            "m":            "[data-startover=unreviewed]",
+            "n":            "[data-startover=all]",
+            "s":            "[data-skip-evaluation]",
         };
+        const action = actions[event.key.toLowerCase()];
 
-        if(actions[event.which] && !event.originalEvent.repeat) {
-            let element = slider.find(actions[event.which]);
+        if(action && !event.originalEvent.repeat) {
+            let element = slider.find(action);
             if(element.length) {
                 element[0].click();
             }
@@ -108,6 +110,37 @@ $(document).ready(() => {
             element.addClass("to-" + direction);
             element.position();
             element.addClass("active");
+        }
+    }
+
+    let nextEvaluationIndex = 0;
+    updateNextEvaluation(0);
+
+    $("[data-skip-evaluation]").on("click", event => {
+        const skippedEvaluationId = $("[data-evaluation]").not(":hidden").data("evaluation");
+        if (skippedEvaluationId !== undefined) {
+            $.ajax({
+                type: "POST",
+                data: {evaluation_id: skippedEvaluationId},
+                url: "{% url 'staff:evaluation_textanswers_skip' %}",
+                error: function(){ window.alert("{% trans 'The server is not responding.' %}"); }
+            });
+            updateNextEvaluation(++nextEvaluationIndex);
+        }
+    });
+
+    function updateNextEvaluation(index) {
+        let nextEvaluation = false;
+        $("[data-next-evaluation-index]").each(function() {
+            if ($(this).data("next-evaluation-index") == index) {
+                $(this).show();
+                nextEvaluation = true;
+            } else {
+                $(this).hide();
+            }
+        });
+        if (!nextEvaluation) {
+            $("[data-skip-evaluation]").hide();
         }
     }
 
--- a/evap/staff/templates/staff_evaluation_textanswers_quick.html
+++ b/evap/staff/templates/staff_evaluation_textanswers_quick.html
@@ -19,6 +19,7 @@
                     <tr><td><kbd>L</kbd></td><td>{% trans 'Remove answer' %}</td></tr>
                     <tr><td><kbd>&#x232b;</kbd></td><td>{% trans 'Unreview answer' %}</td></tr>
                     <tr><td><kbd>&#x21b2;</kbd></td><td>{% trans 'Review next evaluation' %}</td></tr>
+                    <tr><td><kbd>S</kbd></td><td>{% trans 'Skip evaluation' %}</td></tr>
                     <tr><td><kbd>N</kbd></td><td>{% trans 'Show all again' %}</td></tr>
                     <tr><td><kbd>M</kbd></td><td>{% trans 'Show unreviewed' %}</td></tr>
                 </table>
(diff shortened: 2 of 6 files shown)


evap/staff/templates/quick-review.js
```diff
@@ -111,6 +113,35 @@ $(document).ready(() => {
         }
     }
 
+    var nextEvaluationIndex = 0;
+    setNextEvaluation(0)
```
janno42, line 117: there is a semicolon missing here :)

evap/staff/templates/quick-review.js
```diff
@@ -111,6 +113,35 @@ $(document).ready(() => {
...
+    $("#skip-next-evaluation").on("click", event => {
+        const skippedEvaluationId = $("[data-evaluation").not(":hidden").data("evaluation")
+        $.ajax({
+            type: "POST",
+            data: {evaluation_id: skippedEvaluationId},
+            url: "{% url 'staff:evaluation_textanswers_skip' %}",
+            error: function(){ window.alert("{% trans 'The server is not responding.' %}"); }
+        });
+        setNextEvaluation(++nextEvaluationIndex);
+    })
```
janno42, line 128: here as well

janno42 (commented):
looks good!
you don't need to do translations, all strings seem to be correctly marked for translation already and i'll do the rest with the next production update.

and you are right: a test not returning a skipped evaluation on requesting `find_unreviewed_evaluations` would be nice.

evap/staff/templates/quick-review.js
```diff
@@ -111,6 +113,35 @@ $(document).ready(() => {
         }
     }
 
+    var nextEvaluationIndex = 0;
+    setNextEvaluation(0);
+
+    $("#skip-next-evaluation").on("click", event => {
+        const skippedEvaluationId = $("[data-evaluation").not(":hidden").data("evaluation");
```
pixunil, line 120:
```suggestion
        const skippedEvaluationId = $("[data-evaluation]").not(":hidden").data("evaluation");
```
Unless the selector without `]` has a special meaning unknown to me

evap/staff/templates/staff_evaluation_textanswers_quick.html
```diff
@@ -76,10 +77,12 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
                             {% trans 'There are no text answers for this evaluation.' %}
                         </span>
                     {% endif %}
-                    {% if next_evaluation %}
-                        <span>
-                            {% blocktrans with name=next_evaluation.full_name count answers=next_evaluation.num_unreviewed_textanswers %}The next evaluation "{{ name }}" has got {{ answers }} unreviewed text answer.{% plural %}The next evaluation "{{ name }}" has got {{ answers }} unreviewed text answers.{% endblocktrans %}
-                        </span>
+                    {% if next_evaluations %}
+                        {% for next_evaluation in next_evaluations %}
+                            <span data-index="{{ forloop.counter0 }}" data-evaluation="{{ next_evaluation.pk }}">
```
pixunil, line 82: `[data-index]` is a bit too generic for my taste. Would it suffice to include `[data-evaluation]` and depending on the currently visible next evaluation to change the `[href]` of `[data-url=next-evaluation]`? Or include a `[data-next-evaluation-url]` here?

pixunil (changes requested):
Nice solution :cat:

evap/staff/templates/staff_evaluation_textanswers_quick.html
rgwohlbold: I don't think this would be enough. We don't only need the url, but also the title. I will rename it to data-next-evaluation-index

rgwohlbold:
I added a test, but since there is no way to check if a review will occur again without parsing HTML, the test only checks that the HTTP status is 200

evap/staff/templates/staff_evaluation_textanswers_quick.html
```diff
@@ -21,6 +21,7 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
                     <tr><td><kbd>&#x21b2;</kbd></td><td>{% trans 'Review next evaluation' %}</td></tr>
                     <tr><td><kbd>N</kbd></td><td>{% trans 'Show all again' %}</td></tr>
                     <tr><td><kbd>M</kbd></td><td>{% trans 'Show unreviewed' %}</td></tr>
+                    <tr><td><kbd>S</kbd></td><td>{% trans 'Skip evaluation' %}</td></tr>
```
janno42, line 24: please move the new entry between "Review next evaluation" and "Show all again", then the order of the hotkeys and buttons is consistent.

janno42 (changes requested):
when skipping an evaluation and then pressing enter (hotkey for next evaluation), the skipped evaluation will still show up for review. it works correctly when using the button instead.

rgwohlbold:
Implemented changes, fixed tests :+1:

janno42 (approved):

evap/staff/tests/test_views.py
```diff
@@ -2383,6 +2383,30 @@ def test_finishing_review_updates_results(self):
...
+        )
+        top_general_questionnaire = baker.make(Questionnaire, type=Questionnaire.Type.TOP)
+        baker.make(Question, questionnaire=top_general_questionnaire, type=Question.TEXT)
+        evaluation.general_contribution.questionnaires.set([top_general_questionnaire])
+        baker.make(TextAnswer, contribution=evaluation.general_contribution)
+
+        skip_url = "/staff/textanswers/skip"
+        self.app.post(skip_url, user=manager, status=200, params={
+            'evaluation_id': evaluation.id
+        })
```
pixunil, line 2407: You should test some behaviour in your test, here it would be to check that the skipped evaluation is saved in the session. This works on my machine: (why `self.client.session` is not working, but `repsonse.client.session` does – I don’t know)
```suggestion
        response = self.app.post(skip_url, user=manager, status=200, params={
            'evaluation_id': evaluation.id,
        })
        self.assertEqual(response.client.session['review-skipped'], {evaluation.id})
```

pixunil (changes requested):

evap/staff/templates/quick-review.js
```diff
@@ -22,20 +22,22 @@ $(document).ready(() => {
...
+            "j":            "[data-action=publish]",
+            "k":            "[data-action=make_private]",
+            "l":            "[data-action=hide]",
+            "backspace":    "[data-action=unreview]",
+            "enter":        `[data-url=next-evaluation][data-next-evaluation-index=${nextEvaluationIndex}]`,
+            "m":            "[data-startover=unreviewed]",
+            "n":            "[data-startover=all]",
+            "s":            "[data-skip-evaluation]",
         };
+        const action = actions[event.key.toLowerCase()]
```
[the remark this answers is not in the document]
rgwohlbold: I think the readability of the code should be preferred over not using the object as a map. I will add the semicolon.

evap/staff/templates/quick-review.js
```diff
@@ -111,6 +113,37 @@ $(document).ready(() => {
         }
     }
 
+    let nextEvaluationIndex = 0;
+    updateNextEvaluation(0);
+
+    $("[data-skip-evaluation]").on("click", event => {
+        const skippedEvaluationId = $("[data-evaluation]").not(":hidden").data("evaluation");
+        if (skippedEvaluationId !== undefined) {
```
He3lixxx, line 121: In what case is this equal to undefined?

evap/staff/views.py
```diff
@@ -1106,20 +1106,35 @@ def evaluation_textanswers(request, semester_id, evaluation_id):
...
 
     template_data.update(dict(evaluation_sections=evaluation_sections, contributor_sections=contributor_sections))
     return render(request, "staff_evaluation_textanswers_full.html", template_data)
 
+@reviewer_required
+def evaluation_textanswers_skip(request):
+    evaluation_id = request.POST["evaluation_id"]
+    evaluation = get_object_or_404(Evaluation, id=evaluation_id)
+    if evaluation.is_fully_reviewed:
+        raise PermissionDenied
```
He3lixxx, line 1131: Why is this required?

evap/staff/tests/test_views.py
```diff
@@ -2383,6 +2383,31 @@ def test_finishing_review_updates_results(self):
...
+        student = baker.make(UserProfile, email="student@institution.example.com")
+        evaluation = baker.make(
+            Evaluation,
+            participants=[student],
+            voters=[student],
+            state="in_evaluation",
+            can_publish_text_results=True,
+        )
+        top_general_questionnaire = baker.make(Questionnaire, type=Questionnaire.Type.TOP)
+        baker.make(Question, questionnaire=top_general_questionnaire, type=Question.TEXT)
```
He3lixxx, line 2400: Could you see if the test works without some of these models? (I would especially expect that these two are not necessary for the test to pass)

pixunil (approved):

He3lixxx (commented):
Very cool, I'm impressed.

With this code, I have to say I've never seen the UI nor the whole js code here so far, so I don't feel able to properly review this.

evap/staff/views.py
rgwohlbold: I will remove this. Can I remove the two lines above as well?

evap/staff/tests/test_views.py
rgwohlbold: When I remove the is_fully_reviewed check, this can be removed as well

evap/staff/templates/quick-review.js
rgwohlbold: When a user presses the "skip" key, the button is pressed (virtually). This check assures that no AJAX request is sent if the "skip evaluation" button is hidden, in which case skippedEvaluationId is undefined

He3lixxx (approved):

PR #1539, opened 2020-11-16
Regression test for "Access denied on manager login"

### LINKED ISSUE #1536: Write test for the login workflow
https://github.com/e-valuation/EvaP/pull/1531 should've been caught by a test. While at it, verify there's a test for logging in via a login URL.

karyon:
Add regression test for #1523, fixes #1536

First commit contains some minor cleanups.

### DIFF
--- a/evap/evaluation/tests/test_views.py
+++ b/evap/evaluation/tests/test_views.py
@@ -1,5 +1,6 @@
 from django.core import mail
 from django.contrib.auth.hashers import make_password
+from django.contrib.auth.models import Group
 
 from django_webtest import WebTest
 from model_bakery import baker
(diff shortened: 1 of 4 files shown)


evap/evaluation/tests/test_views.py
```diff
@@ -22,6 +23,19 @@ def test_passworduser_login(self):
         password_form['password'] = 'evap'
         self.assertEqual(password_form.submit().status_code, 302)
 
+    def test_login_for_staff_users_correctly_redirects(self):
+        """ Regression test for #1523: Access denied on manager login """
+        internal_email = 'manager@institution.example.com' # external users don't necessarily have a proper redirect page
```
janno42, line 28: manual linting (don't we have a tool for that?): there a space missing before the `#` :)

janno42 (approved):

evap/evaluation/tests/test_views.py
He3lixxx: Welcome to the magic world of linters, where what you want seems so simple yet it is impossible to reach.

In today's episode: I want meaningful linting of my code, including pep-8 style guides- Turns out pylint, the tool we chose, does not include linting rules for all pep-8 conventions. Especially, the two-spaces-before-comment rule was explicitly marked as #wontfix here: https://github.com/PyCQA/pylint/issues/1124

Now, we could, for this one error, additionally run flake8 or something. However, I already see you frowning while writing this comment, and I myself am not a fan of this solution. So I guess, as long as we want to rely on pylint (please don't make me swap linters), we will have to keep manually linting this.

He3lixxx (approved):

evap/evaluation/tests/test_views.py
janno42: sometimes i ~hate~ dislike technology :)
for me it's fine to keep our linting as it is, at least for the moment.

PR #1540, opened 2020-11-30
Fixing navbar position to cover no content anymore

### LINKED ISSUE #1442: Wrapping the breadcrumbs partially covers the content
When the screen is small enough for the breadcrumbs to get wrapped, the content of the page gets partially covered.

Here's an example on mobile where the breadcrumbs cover the title (scrolled all the way to the top):

image

This issue is probably a part of https://github.com/e-valuation/EvaP/issues/883, but not necessarily specific to mobile.

jb3rndt:
Fixes #1442

### DIFF
--- a/evap/static/scss/components/_grid.scss
+++ b/evap/static/scss/components/_grid.scss
@@ -1,8 +1,3 @@
-.container {
-    padding-top: 15px; // to have some space between content and navbar/breadcrumbs
-    flex: 1 0 auto;
-}
-
 .container-striped .row:nth-child(odd) {
     background-color: $table-accent-bg;
 }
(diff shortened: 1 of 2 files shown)


janno42 (approved):
beautiful :)

evap/static/scss/components/_grid.scss
```diff
@@ -1,5 +1,5 @@
 .container {
-    padding-top: 80px; // for navbar (and breadcrumbs)
+    padding-top: 15px; // to have some space between content and navbar/breadcrumbs
     flex: 1 0 auto;
 }
```
pixunil, line 4: Cool that `sticky` is working out of the box :)
- I’d replace `padding-top` with `.pt-3` directly on the `.container` (slightly different in height, but I guess you aren’t attached to the exact amount of `15px`)
- According to my firefox, `flex` has no effect on this element as it is not flexible, so the wohle stylesheet rule can be removed (unless I have overseen a different usage of `.container`, but my search only deteced one in `base.html`)

pixunil (changes requested):

He3lixxx (approved):

pixunil (approved):

janno42:
Thank you! :)

PR #1544, opened 2020-12-14
Student index page: Fix O(n) database queries - second attempt

He3lixxx:
follow up for #1497. There, we didn't add a test, and turns out the "fix" was not good enough. This time, I added a test - so lets hope we have better luck with that.


janno42 (approved):
⏱️

Issue #1303, opened 2019-02-11
Replace option for participant adding


janno42:
When copying or importing participants, there should be an option to replace the current participants instead of just adding the new ones.

PR #1546, opened 2021-01-04
Replace all participants during import

dieknolle3333:
Work in progress on #1303


He3lixxx:
I think we can close this now with #1579 open.

Issue #1509, opened 2020-09-14
Refactor contributor index queries


janno42:
On merging #1506 I noticed that the queries for the contributor index page should be refactored similar to those for the student index as done in #1495.

PR #1548, opened 2021-01-06
Some query optimizations for contributor index view

karyon:
I looked at https://github.com/e-valuation/EvaP/issues/1509, but didn't find good opportunities for refactoring. I did some optimizations instead.

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -705,14 +705,14 @@ class Evaluation(LoggedModel):
 
     def is_user_editor_or_delegate(self, user):
         represented_users = user.represented_users.all() | UserProfile.objects.filter(pk=user.pk)
-        return self.contributions.filter(contributor__pk__in=represented_users, role=Contribution.Role.EDITOR).exists() or self.course.responsibles.filter(pk__in=represented_users).exists()
+        return self.contributions.filter(contributor__in=represented_users, role=Contribution.Role.EDITOR).exists() or self.course.responsibles.filter(pk__in=represented_users).exists()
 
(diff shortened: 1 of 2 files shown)


d-croft (approved):
Looks good to me.

evap/evaluation/models.py
```diff
@@ -704,17 +704,15 @@ def hours_until_evaluation(self):
         return (self.vote_start_datetime - datetime.now()) / timedelta(hours=1)
 
     def is_user_editor_or_delegate(self, user):
-        represented_user_pks = [represented_user.pk for represented_user in user.represented_users.all()]
-        represented_user_pks.append(user.pk)
-        return self.contributions.filter(contributor__pk__in=represented_user_pks, role=Contribution.Role.EDITOR).exists() or self.course.responsibles.filter(pk__in=represented_user_pks).exists()
+        represented_users = user.represented_users.all() | UserProfile.objects.filter(pk=user.pk)
+        return self.contributions.filter(contributor__pk__in=represented_users, role=Contribution.Role.EDITOR).exists() or self.course.responsibles.filter(pk__in=represented_users).exists()
```
janno42, line 708: why is this `contributor__pk__in` and not `contributor__in`? you are now using the user profile objects, right?

janno42 (approved):
thank you! 👍

evap/evaluation/models.py
karyon: Good question. Yes, I'm using the objects, and simply forgot to remove the __pk here. But it seems to work either way, the generated sql is identical. I assume this works to enable second case in this line, where the pk__ must be preserved...

Issue #1384, opened 2019-11-29
It is possible to reconstruct the author of all text answers and some rating answers


karyon:
Follow-Up to https://github.com/fsr-de/EvaP/issues/1335.

Postgres saves the transaction ID of every tuple inserted into the database (xmin of the system columns). Since the vote operation by design is one transaction, and that transaction saves both all the answers and the addition to the `voters` table, there's a direct link between authors and all their answers.

For rating answers, this is not entirely solved by the use of aggregating counters: the tables in postgres are insert-only, that means old versions of those counters stay in the table at least until the next VACUUM. So you just need to diff the version in question with the previous version. Though i couldn't find how to access such old tuples, it would be nice to not have them. But in any case, the comments are the bigger problem.

Possible solutions from less-bad to bad:
* ~~CLUSTER + VACUUM might just work. We should verify what happens when inserting a tuple with a UUID PK, then clustering the table, and then vacuuming. In case the CLUSTER has it's own transaction (it might not, since it has ACCESS EXCLUSIVE on the entire table, i dunno?), it might create new versions of the reordered tuples just like other updates do. In my naive understanding, since our answers have random PKs they will usually not go into the last page of the table and will have to be moved by the next CLUSTER (as long as the [fillfactor](https://www.postgresql.org/docs/12/sql-cluster.html) stays at 100%). Combined with a VACUUM, that might be it already. So maybe we don't even need to change anything! :D but we should verify.~~ CLUSTER does not change the xmin even if the tuples move.
* When inserting votes/comments, touch existing votes/comments of the same evaluation (e.g. flip a dummy boolean, or maybe an UPDATE with no changes is enough). VACUUM periodically.
* Periodically touch all votes/comments of the active semester. VACUUM afterwards.
* We could aggregate all comments for a question to one large text in the database with some separators, but since old versions of this large text would be kept by postgres at least until the next vacuum, that doesn't seem like a full solution. Also, we'd still have the order of when the user added their comment, which would be nice to not save as well. Also, we'd probably need to shuffle the order of the comments within that large text, and that won't help with questions that receive only one comment...
* [On stackoverflow](https://stackoverflow.com/questions/58758713) people suggested to put our voters-table and the answers on different databases. I think that does not solve anything, since the consecutive transaction ID sequences on both databases still allow for a full correlation of the two tables.
* Putting the inserts of the vote operation in different transactions doesn't really solve anything since transaction IDs are consecutive. Also we'd lose the transaction guarantees.

[...]

PR #1549, opened 2021-01-18
Do bogus update on Answer update to hide insertion transaction number

He3lixxx:
Fixes parts of #1384.

This will update all existing Answer rows when adding a new answer, inside the same transaction. Thus, from the new tuples, it is not clear which Answer was added / changed.

However, When adding a text answer or a RatingAnswerCounter, obviously, the old tuple is still in the database (and now considered "dead").
You can check out how many dead tuples are currently in the database using
```
SELECT relname, n_dead_tup FROM pg_stat_user_tables;
```

We would want to keep this as close to 0 as possible. We could either add a cron job that runs [`vacuum`](https://www.postgresql.org/docs/9.5/sql-vacuum.html) just as we did for the physical tuple-reordering, or we could try to rely on the [autovacuum](https://www.postgresql.org/docs/9.5/routine-vacuuming.html#AUTOVACUUM), which will be enabled by default and has to be enabled for the database to run smoothly. [This page](https://www.shubhamdipt.com/blog/how-to-clean-dead-tuples-and-monitor-postgresql-using-vacuum/) lists a few commands to inspect the current configuration.

The biggest concern with vacuum is that you will require a `FULL` vacuum for the database to actually give back the memory to the OS, and then we're still waiting for someone to overwrite it. This requires exclusive access to the database, so we will have to stop serving for the time this runs. (EDIT: Currently, the `CLUSTER` we run also requires exclusive access during that period. It might be feasible to actually do a full vacuum then, too).

A non-full vacuum will keep the memory reserved in postgres and use it when it requires new space (which in our should be when someone else votes). Until then, it will not be overwritten.

I'd say we should monitor our dead tuple count on production during a voting phase. If it stays low enough (default autovacuum starts with 50 dead tuples IMO), we might not have to do any additional steps. Otherwise, we might want to add the cronjob. For the actual overwrite, I think hoping that the memory will be overwritten at some point is the best we can do.

(If someone gets a memory dump, actually finding the piece of memory with the dead tuples is really easy, as we use UUIDS, so you can search for their values in the memory dump)

### DIFF
--- a/evap/student/tests/test_views.py
+++ b/evap/student/tests/test_views.py
@@ -6,7 +6,7 @@ from django_webtest import WebTest
 from model_bakery import baker
 
 from evap.evaluation.models import (UserProfile, Evaluation, Questionnaire, Question, Contribution,
-                                    TextAnswer, RatingAnswerCounter, Semester)
+                                    TextAnswer, RatingAnswerCounter, Semester, Answer)
 from evap.evaluation.tests.tools import WebTestWith200Check
 from evap.student.tools import question_id
 from evap.student.views import SUCCESS_MAGIC_STRING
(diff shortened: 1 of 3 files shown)


He3lixxx:
(note for myself, in case I want to inspect some more internal stuff later on: Instructions on how to access dead tuples can be found here: https://www.percona.com/blog/2018/08/06/basic-understanding-bloat-vacuum-postgresql-mvcc/)

karyon (approved):

evap/student/tests/test_views.py
```diff
@@ -292,3 +292,25 @@ def test_textanswer_visibility_is_shown(self):
             page.body.decode(),
             r"can be seen by:<br />\s*{}".format(self.contributor1.full_name.replace('(', '\\(').replace(')', '\\)'))
         )
+
+    def test_xmin_of_all_answes_is_updated(self):
```
He3lixxx, line 296: answerrrrrrs

karyon:
I just said that in person but I'll do it again for the record: I just *LOVE* how we discussed this thoroughly over the course of a year to arrive at a solution that with surgical precision solves the problem in two lines of code. That unit test of yours is the icing on the cake. Great work! :D

karyon:
Regarding vacuums and memorydumps: I wouldn't worry too much about it. Doing a VACUUM FULL might be ok, depending on how long it takes, we're already doing a periodic CLUSTER which also requires exclusive access. But even a non-full vacuum might already be enough, as reading postgres' memory requires root access and, well, I'm not sure whether an attack with root privileges that aims to deanonymize the last couple votes is a scenario we want to defend against.

janno42 (approved):
⭐

PR #1550, opened 2021-02-01
Log improvements

janno42:
small log improvements, see commit messages


evap/evaluation/migrations/0119_auto_20210201_1523.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 3.0.10 on 2021-02-01 15:23
```
He3lixxx, line 1: maybe rename the file to 0119_contribution_verbose_names.py or something?

evap/evaluation/models_logging.py
```diff
@@ -113,7 +115,7 @@ def message(self):
             message = _("A {cls} was deleted.")
 
         return message.format(
-            cls=self.content_type.model_class()._meta.verbose_name_raw,
+            cls=capitalize_first(self.content_type.model_class()._meta.verbose_name),
```
He3lixxx, line 118: why did we use _raw here before (_raw should be the untranslated version - is it possible that translation is in some cases not available, which will break this code?)

janno42: @felixrindt did you have any specific reasons for that?

felixrindt: No, I suppose it works without raw as well 😅

janno42: i think so too. we shouldn't have any user-facing strings that aren't translated so everything should be fine.

He3lixxx (approved):

PR #1551, opened 2021-02-01
Take user's names from OIDC if empty

janno42:
Partly revert 94fc6dc8b87
First name and last name of a user should be updated via OIDC if they are currently empty


He3lixxx (approved):
Testing would be cool, but seems hard to properly realize without https://github.com/mozilla/mozilla-django-oidc/issues/113.

PR #1552, opened 2021-02-01
Update dependency versions

janno42:
I checked the changelogs and didn't find any changes we have to do on our side.
@He3lixxx, are there any linting things we waited for?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -60,6 +60,8 @@ jobs:
         uses: actions/checkout@v2
       - name: Install dependencies
         run: pip install -r requirements-test.txt
+      - name: Add localsettings
+        run: cp evap/settings_test.py evap/localsettings.py
       - name: Run linter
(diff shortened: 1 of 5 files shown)


He3lixxx:
For pylint-django to work properly, I (locally) need something like this in my pylintrc:
```
DJANGO_SETTINGS_MODULE=your.app.settings
```

In the .pylintrc, we still have one check that we would like to enable:
```
disable =
    ....
    signature-differs,           # can be used if https://github.com/PyCQA/pylint/issues/3737 is fixed
    ....
```
the PR that closed this issue was merged on Dec 31, the release we are upgrading to here is from August 21, so the fix is not (yet) in pylint, thus the line has to stay
-> Nothing to change for the linting things we're waiting on.

karyon (approved):
I also went through some of the changelogs and didn't find anything that needs changes.

He3lixxx (approved):

PR #1553, opened 2021-02-01
Make Evaluation.state an FSMIntegerField

### LINKED ISSUE #1191: Make Course.state an FsmIntegerField
django-fsm can do integers: https://github.com/kmmbvnr/django-fsm#integer-field-support

We should use that for Course.state, which has the following benefits:
- we'd finally be forced to use some constants instead of the magic strings which are the different course states
- comparisons like `if state >= Course.IN_EVALUATION` just work

Cons:
- Course.IN_EVALUATION is a little more verbose than 'in_evaluation'
- the database becomes less human readable
- how to access constants in templates (non-ugly)?

We decided the cons are not really significant, in addition it's kind of the same thing for Questionnaire.type, where we made the very same decision to use integers.

niklasmohrin:
Closes #1191

Don't need a review yet :)

### DIFF
--- a/evap/contributor/views.py
+++ b/evap/contributor/views.py
@@ -23,7 +23,7 @@ def index(request):
     show_delegated = get_parameter_from_url_or_session(request, "show_delegated", True)
 
     represented_proxy_users = user.represented_users.filter(is_proxy_user=True)
-    contributor_visible_states = ['prepared', 'editor_approved', 'approved', 'in_evaluation', 'evaluated', 'reviewed', 'published']
+    contributor_visible_states = [Evaluation.State.PREPARED, Evaluation.State.EDITOR_APPROVED, Evaluation.State.APPROVED, Evaluation.State.IN_EVALUATION, Evaluation.State.EVALUATED, Evaluation.State.REVIEWED, Evaluation.State.PUBLISHED]
     own_courses = Course.objects.filter(
         Q(evaluations__state__in=contributor_visible_states) & (
             Q(responsibles=user) |
@@ -63,8 +63,7 @@ def index(request):
 
     displayed_evaluations.sort(key=lambda evaluation: (evaluation.course.name, evaluation.name))  # evaluations must be sorted for regrouping them in the template
 
-    annotate_distributions_and_grades(e for e in displayed_evaluations if e.state == "published")
-
+    annotate_distributions_and_grades(e for e in displayed_evaluations if e.state == Evaluation.State.PUBLISHED)
     displayed_evaluations = get_evaluations_with_course_result_attributes(displayed_evaluations)
 
     semesters = Semester.objects.all()
@@ -108,7 +107,7 @@ def evaluation_view(request, evaluation_id):
     evaluation = get_object_or_404(Evaluation, id=evaluation_id)
 
     # check rights
-    if not evaluation.is_user_editor_or_delegate(user) or evaluation.state not in ['prepared', 'editor_approved', 'approved', 'in_evaluation', 'evaluated', 'reviewed']:
+    if not evaluation.is_user_editor_or_delegate(user) or evaluation.state not in [Evaluation.State.PREPARED, Evaluation.State.EDITOR_APPROVED, Evaluation.State.APPROVED, Evaluation.State.IN_EVALUATION, Evaluation.State.EVALUATED, Evaluation.State.REVIEWED]:
         raise PermissionDenied
 
     InlineContributionFormset = inlineformset_factory(Evaluation, Contribution, formset=ContributionFormSet, form=EditorContributionForm, extra=0)
@@ -146,7 +145,7 @@ def evaluation_edit(request, evaluation_id):
     evaluation = get_object_or_404(Evaluation, id=evaluation_id)
 
     # check rights
-    if not (evaluation.is_user_editor_or_delegate(request.user) and evaluation.state == 'prepared'):
+    if not (evaluation.is_user_editor_or_delegate(request.user) and evaluation.state == Evaluation.State.PREPARED):
         raise PermissionDenied
 
     post_operation = request.POST.get('operation') if request.POST else None
@@ -199,7 +198,7 @@ def evaluation_preview(request, evaluation_id):
     evaluation = get_object_or_404(Evaluation, id=evaluation_id)
 
     # check rights
-    if not (evaluation.is_user_responsible_or_contributor_or_delegate(user) and evaluation.state in ['prepared', 'editor_approved', 'approved', 'in_evaluation', 'evaluated', 'reviewed']):
+    if not (evaluation.is_user_responsible_or_contributor_or_delegate(user) and evaluation.state in [Evaluation.State.PREPARED, Evaluation.State.EDITOR_APPROVED, Evaluation.State.APPROVED, Evaluation.State.IN_EVALUATION, Evaluation.State.EVALUATED, Evaluation.State.REVIEWED]):
         raise PermissionDenied
 
     return get_valid_form_groups_or_render_vote_page(request, evaluation, preview=True)[1]
--- a/evap/contributor/views.py
+++ b/evap/contributor/views.py
@@ -107,7 +107,7 @@ def evaluation_view(request, evaluation_id):
     evaluation = get_object_or_404(Evaluation, id=evaluation_id)
 
     # check rights
-    if not evaluation.is_user_editor_or_delegate(user) or not Evaluation.State.NEW < evaluation.state < Evaluation.State.PUBLISHED:
+    if not evaluation.is_user_editor_or_delegate(user) or not Evaluation.State.PREPARED <= evaluation.state <= Evaluation.State.REVIEWED:
         raise PermissionDenied
 
     InlineContributionFormset = inlineformset_factory(Evaluation, Contribution, formset=ContributionFormSet, form=EditorContributionForm, extra=0)
@@ -198,7 +198,7 @@ def evaluation_preview(request, evaluation_id):
     evaluation = get_object_or_404(Evaluation, id=evaluation_id)
 
     # check rights
-    if not (evaluation.is_user_responsible_or_contributor_or_delegate(user) and Evaluation.State.NEW < evaluation.state < Evaluation.State.PUBLISHED):
+    if not (evaluation.is_user_responsible_or_contributor_or_delegate(user) and Evaluation.State.PREPARED <= evaluation.state <= Evaluation.State.REVIEWED):
         raise PermissionDenied
 
     return get_valid_form_groups_or_render_vote_page(request, evaluation, preview=True)[1]
--- a/evap/contributor/views.py
+++ b/evap/contributor/views.py
@@ -107,7 +107,7 @@ def evaluation_view(request, evaluation_id):
     evaluation = get_object_or_404(Evaluation, id=evaluation_id)
 
     # check rights
-    if not evaluation.is_user_editor_or_delegate(user) or evaluation.state not in [Evaluation.State.PREPARED, Evaluation.State.EDITOR_APPROVED, Evaluation.State.APPROVED, Evaluation.State.IN_EVALUATION, Evaluation.State.EVALUATED, Evaluation.State.REVIEWED]:
+    if not evaluation.is_user_editor_or_delegate(user) or Evaluation.State.NEW < evaluation.state < Evaluation.State.PUBLISHED:
         raise PermissionDenied
 
     InlineContributionFormset = inlineformset_factory(Evaluation, Contribution, formset=ContributionFormSet, form=EditorContributionForm, extra=0)
@@ -198,7 +198,7 @@ def evaluation_preview(request, evaluation_id):
     evaluation = get_object_or_404(Evaluation, id=evaluation_id)
 
     # check rights
-    if not (evaluation.is_user_responsible_or_contributor_or_delegate(user) and evaluation.state in [Evaluation.State.PREPARED, Evaluation.State.EDITOR_APPROVED, Evaluation.State.APPROVED, Evaluation.State.IN_EVALUATION, Evaluation.State.EVALUATED, Evaluation.State.REVIEWED]):
+    if not (evaluation.is_user_responsible_or_contributor_or_delegate(user) and Evaluation.State.NEW < evaluation.state < Evaluation.State.PUBLISHED):
         raise PermissionDenied
 
     return get_valid_form_groups_or_render_vote_page(request, evaluation, preview=True)[1]
--- a/evap/contributor/views.py
+++ b/evap/contributor/views.py
@@ -107,7 +107,7 @@ def evaluation_view(request, evaluation_id):
     evaluation = get_object_or_404(Evaluation, id=evaluation_id)
 
     # check rights
-    if not evaluation.is_user_editor_or_delegate(user) or Evaluation.State.NEW < evaluation.state < Evaluation.State.PUBLISHED:
+    if not evaluation.is_user_editor_or_delegate(user) or not Evaluation.State.NEW < evaluation.state < Evaluation.State.PUBLISHED:
         raise PermissionDenied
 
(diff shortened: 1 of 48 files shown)


niklasmohrin:
When running `./manage.py makemigrations`, I was given a migration that did a `migrations.AlterField' operation, but as far as I can tell, this can only be used to alter the type and does not allow me to specify how to map the strings to the numeric constants. 

Should I add a temporary field `_state`, map the old values to that, remove the old field and then rename the `_state` field to `state`?

Does anybody know what is best to do here? 

https://docs.djangoproject.com/en/3.1/ref/migration-operations/#alterfield

karyon:
Seems like you've figured it out already :) Check out [0088_prepare_use_integers_and_groups_for_question_type.py](https://github.com/e-valuation/EvaP/blob/master/evap/evaluation/migrations/0088_prepare_use_integers_and_groups_for_question_type.py) and [0089_use_integers_for_question_type.py](https://github.com/e-valuation/EvaP/blob/master/evap/evaluation/migrations/0089_use_integers_for_question_type.py)) which do pretty much what you want to do. I'm not sure why they're split into two migrations, but sometimes the django migrations have limitations that require that.

evap/evaluation/templatetags/evaluation_filters.py
```diff
@@ -112,27 +116,29 @@ def weight_info(evaluation):
...
     return None
 
 
 @register.filter
 def approval_state_icon(state):
     if state in APPROVAL_STATES:
         return APPROVAL_STATES[state].icon
-    if state in ['in_evaluation', 'evaluated', 'reviewed', 'published']:
-        return APPROVAL_STATES['approved'].icon
+    if state > Evaluation.State.APPROVED:
```
niklasmohrin, line 140: Should this go back to checking in the list, or is it fair to assume that `state' is actually a valid state number?

niklasmohrin (commented):
I think this is all the spots where the magic strings come up. will rebase another day

evap/evaluation/templatetags/evaluation_filters.py
karyon: i think that's fair to assume after you've verified every caller ;)

evap/staff/templates/staff_semester_view.html
```diff
@@ -287,15 +287,15 @@ <h3>
                         <div>
                             <button name="target_state" value="prepared" type="submit" class="btn btn-sm btn-light"
                                 data-toggle="tooltip" data-placement="top" title="{% trans 'Evaluations in preparation or approved by editors' %}">
-                                <span class="{{ 'new'|approval_state_icon }}"></span>
-                                <span class="{{ 'editor_approved'|approval_state_icon }}"></span>
+                                <span class="{{ 'new'|state_from_english|approval_state_icon }}"></span>
```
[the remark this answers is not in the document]
niklasmohrin: I guess `approval_state_icon` could work with the string constants, but I thought that it would be more consistent if everything was based on `Evaluation.State`. Then, one would convert from a string to an int as early as possible to use all of the "library" with the int.

I could perform the conversion in the icon function too though if you think that's a better fit

evap/staff/tests/test_views.py
```diff
@@ -1114,7 +1114,7 @@ def helper_publish_evaluation_with_publish_notifications_for(self, evaluation, c
         page = self.app.get("/staff/semester/1", user=self.manager)
         form = page.forms["evaluation_operation_form"]
         form['evaluation'] = evaluation.pk
-        response = form.submit('target_state', value="published")
+        response = form.submit('target_state', value=Evaluation.state_to_str(Evaluation.State.PUBLISHED))
```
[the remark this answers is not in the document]
niklasmohrin: Your next comment is actually where this string is handled. The strings are converted to / used to look up these operation classes which call the transition methods on an evaluation and these transitions have the actual `Evaluation.State` values already replaced. I had left these as strings, because I thought that they would be "user-facing" in a URL, but they could of course also be integers. What do you think?

evap/staff/views.py
```diff
@@ -313,11 +313,11 @@ def apply(request, evaluations, email_template=None, email_template_contributor=
...
-        'new': RevertToNewOperation,
-        'prepared': ReadyForEditorsOperation,
-        'in_evaluation': BeginEvaluationOperation,
-        'reviewed': UnpublishOperation,
-        'published': PublishOperation,
+    'new': RevertToNewOperation,
+    'prepared': ReadyForEditorsOperation,
+    'in_evaluation': BeginEvaluationOperation,
+    'reviewed': UnpublishOperation,
+    'published': PublishOperation,
```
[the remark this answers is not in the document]
niklasmohrin: see above

evap/staff/templates/staff_semester_view.html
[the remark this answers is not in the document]
karyon: Nono, not the string constants, the integer constants! :) Maybe you didn't do that because you don't have an `evaluation` object available? Is there any way to make the enum available in the templates regardless?

evap/staff/tests/test_views.py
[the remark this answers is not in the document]
karyon: I wouldn't mind if the operations were integers. I mean, our courses and semesters and everything is integers too in urls. But that's something for @janno42 to decide i would say.

evap/staff/templates/staff_semester_view.html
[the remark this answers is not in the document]
niklasmohrin: Oh, now I get it :laughing: yeah, I just passed `Evaluation` now :+1:

niklasmohrin:
I think that's all, at least I didn't find anything with

```bash
rg "(\"|')(new|prepared|editor_approved|approved|in_evaluation|evaluated|reviewed|published)(\"|')" -g "!*.js"
```

anymore.

Question remains whether the constants should also be used in query parameters (so they contain numbers instead of the strings).

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -194,15 +194,15 @@
                 <span class="fas fa-envelope" aria-hidden="true"></span>
             </a>
         {% endif %}
-        {% if state == 'new' or state == 'prepared' or state == 'editor_approved' or state == 'approved' %}
+        {% if state == evaluation.State.NEW or state == evaluation.State.PREPARED or state == evaluation.State.EDITOR_APPROVED or state == evaluation.State.APPROVED %}
             <a href="{% url 'staff:evaluation_preview' semester.id evaluation.id %}" class="btn btn-sm btn-light" data-toggle="tooltip" data-placement="top" title="{% trans 'Preview' %}">
                 <span class="fas fa-eye" aria-hidden="true"></span>
             </a>
-        {% elif state == 'in_evaluation' or state == 'evaluated' or state == 'reviewed' and evaluation|can_results_page_be_seen_by:request.user %}
+        {% elif state == evaluation.State.IN_EVALUATION or state == evaluation.State.EVALUATED or state == evaluation.State.REVIEWED and evaluation|can_results_page_be_seen_by:request.user %}
```
He3lixxx, line 201: `elif < PUBLISHED` should match semantics here

evap/student/templates/student_index_semester_evaluations_list.html
```diff
@@ -138,13 +138,13 @@
...
                                             {% if evaluation.is_single_result %}
-                                                {% if evaluation.state == 'published' %}
+                                                {% if evaluation.state == evaluation.State.PUBLISHED %}
                                                     <span data-toggle="tooltip" data-placement="top" title="{% trans 'Individual participants are not stored for single results. The number shown here is the number of votes received.' %}">
                                                         {% include 'progress_bar_fill.html' with icon='user' total=evaluation.single_result_rating_result.count_sum fill=100 %}
                                                     </span>
                                                 {% endif %}
                                             {% else %}
-                                                {% if evaluation.state == 'evaluated' or evaluation.state == 'reviewed' or evaluation.state == 'published' %}
+                                                {% if evaluation.state > evaluation.State.IN_EVALUTION %}
```
He3lixxx, line 147: It's dangerous to go alone. Here, take this: "A"

Did no test catch this? Feels risky...

evap/student/templates/student_index_semester_evaluations_list.html
```diff
@@ -154,9 +154,9 @@
                                     </td>
                                     <td class="text-right">
                                         {% if not semester.results_are_archived %}
-                                            {% if evaluation.state == 'published' %}
+                                            {% if evaluation.state == evaluation.State.PUBLISHED %}
                                                 {% include 'evaluation_result_widget.html' with course_or_evaluation=evaluation %}
-                                            {% elif not evaluation.participates_in or evaluation.voted_for and evaluation.state == 'in_evaluation' or evaluation.state == 'evaluated' or evaluation.state == 'reviewed' %}
+                                            {% elif not evaluation.participates_in or evaluation.voted_for and evaluation.state == evaluation.State.IN_EVALUTION or evaluation.state == evaluation.State.EVALUATED or evaluation.state == evaluation.State.REVIEWED %}
```
He3lixxx, line 159: EVALUTION here too

He3lixxx (commented):
Looks good, thanks for the work.

Mainly suggestions on simplifying some of the long `and state == state` chains.

evap/staff/tests/test_views.py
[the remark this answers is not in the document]
janno42: yes, please use the integers. people are not meant to fiddle with urls manually anyway :)

evap/student/templates/student_index_semester_evaluations_list.html
janno42: `evaluation.State.IN_EVALUATION <= evaluation.state <= evaluation.State.REVIEWED`?

janno42:
something seems to be not working:
image

janno42:
and please have a look at the logs, they now need special handling of the statenames:
image

evap/student/templates/student_index_semester_evaluations_list.html
niklasmohrin: Actually, this is how I avoided passing more template data into `render`, because whenever there is an evaluation, I thought "why wouldn't I just access the class member through the instance?" (or maybe it was someone else's idea, I don't remember). Does it not work? I thought I tested this view ... :thinking:

He3lixxx: I'm fine with using `evaluation.State.WHATEVER`. I just wanted to point out that you wrote `EVALUTION` instead of `EVALUATION` -- this definitely should fail. Keep in mind: The template renderer will simply ignore all errors, it would just act as if the condition was false.

evap/evaluation/models.py
```diff
@@ -818,6 +848,18 @@ def annotate_with_participant_and_voter_counts(cls, evaluation_query):
     def unlogged_fields(self):
         return super().unlogged_fields + ["voters", "is_single_result", "can_publish_text_results", "_voter_count", "_participant_count"]
 
+    @classmethod
+    def transform_log_action(cls, field_action):
+        import pdb
+        pdb.set_trace()
```
karyon, line 854: leftover debug code

evap/evaluation/models_logging.py
```diff
@@ -271,6 +275,10 @@ def unlogged_fields(self):
         """Specify a list of field names so that these fields don't get logged."""
         return ['id', 'order']
 
+    @classmethod
```
karyon, line 278: why is this a classmethod and not a regularly overridable method?

evap/evaluation/models_logging.py
```diff
@@ -68,6 +68,7 @@ def _field_actions_for_field(field, actions):
         elif hasattr(field, "choices") and field.choices:
             # convert values from choice-based fields to their display equivalent
             items = [_choice_to_display(field, item) for item in items]
+        # TODO: check for Evaluation.State
```
karyon, line 71: leftover TODO

karyon (commented):
I don't know whether you wanted to have a review at this point, but anyway, here it is, I hope you don't mind :)

evap/evaluation/models_logging.py
niklasmohrin: I changed it to a `staticmethod` for now. I think it's a bit strange to override a static method with a class method (like `Evaluation` does), but in a way it makes sense. I don't know if this behavior is backed by the language standard though :man_shrugging:

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -5,7 +5,7 @@
 
 <td>
     {% if request.user.is_manager and not info_only and not semester.participations_are_archived %}
-        {% if state == 'new' or state == 'prepared' or state == 'editor_approved' or state == 'approved' or state == 'reviewed' or state == 'published' %}
+        {% if not evaluation.State.APPROVED < state < evaluation.State.REVIEWED %}
```
janno42, line 8: this doesn't seem correct

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -20,19 +20,19 @@
     </td>
 {% endwith %}
 
-{% if state == 'new' or state == 'prepared' or state == 'editor_approved' %}
+{% if state < evaluation.State.IN_EVALUATION %}
```
janno42, line 23: `< evaluation.State.APPROVED`?

janno42 (changes requested):

evap/staff/templates/staff_semester_view_evaluation.html
niklasmohrin: I don't see it :thinking:

evap/evaluation/templatetags/evaluation_filters.py
```diff
@@ -112,29 +116,20 @@ def weight_info(evaluation):
...
 
-
 @register.filter
 def approval_state_values(state):
     if state in APPROVAL_STATES:
         return APPROVAL_STATES[state]
-    if state in ['in_evaluation', 'evaluated', 'reviewed', 'published']:
-        return APPROVAL_STATES['approved']
-    return None
+    return APPROVAL_STATES[Evaluation.State.APPROVED]
```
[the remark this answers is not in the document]
niklasmohrin: I verified all the callers. Should we still add a check, in case a new user of this filter ever messes it up?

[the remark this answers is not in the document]
He3lixxx: Hmm so this could break if someone added a new state that should not be approved. However, I think when adding a new state, one would look at all occurrences of APPROVAL_STATES anyway, so I'm fine with how it is right now.

He3lixxx (approved):

evap/contributor/views.py
```diff
@@ -108,7 +107,7 @@ def evaluation_view(request, evaluation_id):
     evaluation = get_object_or_404(Evaluation, id=evaluation_id)
 
     # check rights
-    if not evaluation.is_user_editor_or_delegate(user) or evaluation.state not in ['prepared', 'editor_approved', 'approved', 'in_evaluation', 'evaluated', 'reviewed']:
+    if not evaluation.is_user_editor_or_delegate(user) or not Evaluation.State.NEW < evaluation.state < Evaluation.State.PUBLISHED:
```
janno42, line 110: inclusive bounds would be better here imho

evap/contributor/views.py
```diff
@@ -199,7 +198,7 @@ def evaluation_preview(request, evaluation_id):
     evaluation = get_object_or_404(Evaluation, id=evaluation_id)
 
     # check rights
-    if not (evaluation.is_user_responsible_or_contributor_or_delegate(user) and evaluation.state in ['prepared', 'editor_approved', 'approved', 'in_evaluation', 'evaluated', 'reviewed']):
+    if not (evaluation.is_user_responsible_or_contributor_or_delegate(user) and Evaluation.State.NEW < evaluation.state < Evaluation.State.PUBLISHED):
```
janno42, line 201: here as well

janno42 (approved):

PR #1556, opened 2021-02-13
Update Django to 3.1

janno42:
- The providing_args argument is deprecated. As it is purely documentational, it has no replacement.
- django.conf.urls.url() is deprecated in favor of django.urls.re_path()
- django.contrib.postgres.fields.JSONField is deprecated. Use django.db.models.JSONField instead.

Also updating django-extensions to the most current version


evap/evaluation/models.py
```diff
@@ -382,7 +382,7 @@ class Evaluation(LoggedModel):
     # Disable to prevent editors from changing evaluation data
     allow_editors_to_edit = models.BooleanField(verbose_name=_("allow editors to edit"), default=True)
 
-    evaluation_evaluated = Signal(providing_args=['request', 'semester'])
+    evaluation_evaluated = Signal()
```
He3lixxx, line 385: Would it make sense to put these in a comment so people potentially receiving the signal can quickly figure out what arguments to expect?

He3lixxx (approved):

PR #1560, opened 2021-02-15
'I can't give feedback about this contributor' button: performance, don't scroll away

He3lixxx:
The old code will save `questions_of_the_contributor` * `total_evaluation_questions` many elements, which is slow (we've had reports of up to 5 seconds). 

Sisyphus does not support a nice way to say "Hey, I'm going to bulk update, please only save once", so we have to monkey-patch it. It supports a `onBeforeSave` method, which can abort saving if it returns `false`. However, up to the point where it evaluates this function, it has already spent half the time on just finding all the elements, so we have to stop it earlier.
Fun Fact: Sisyphus also always saves all elements, even if just one element was changed #scalesnicely.

Bootstrap's `button.js` click event listener causes a scroll to any button that is clicked. So, if the contributor's card is bigger than the screen, we get weird scrolling. I didn't find any way to tell bootstrap to stop this, but maybe, I just didn't look in the right place -- ideas are welcome.

With both these changes, it feels like we don't want to actually use `.click()` anymore, but I don't see any way to nicely set the bootstrap buttons (which from my experience are a pain to work with), so I guess we're stuck with that.

### DIFF
--- a/evap/student/templates/student_vote.html
+++ b/evap/student/templates/student_vote.html
@@ -228,7 +228,6 @@
                     } else {
                         lastSavedLabel.text("{% trans 'Could not save your information locally' %}");
                     }
-                    
             }
 
             function padWithLeadingZeros(number){
@@ -299,9 +298,14 @@
                 button.addEventListener("click", () => {
                     const contributorId = button.dataset.markNoAnswersFor;
                     const voteArea = document.querySelector(`#vote-area-${contributorId}`);
-                    // check "no answer" and uncheck all other options
-                    voteArea.querySelectorAll(".vote-inputs [type=radio][value='6']")
-                        .forEach(radioInput => radioInput.click());
+
+                    voteArea.querySelectorAll(".vote-btn.active").forEach(el => el.classList.remove("active"));
+                    voteArea.querySelectorAll(".vote-inputs [type=radio][value='6']").forEach(radioInput => {
+                        radioInput.checked = true;
+                        radioInput.closest(".vote-btn").classList.add("active");
+                    });
+
+                    sisyphus.saveAllData();
 
                     // hide questionnaire for contributor
                     $(voteArea).collapse("hide");


karyon:
https://stackoverflow.com/questions/17034804/how-to-prevent-page-from-scrolling-back-up-when-a-button-is-clicked suggests using preventDefault to prevent the scrolling, did you try that?

https://getbootstrap.com/docs/4.4/components/buttons/#checkbox-and-radio-buttons says "The checked state for these buttons is only updated via click event on the button. If you use [...] manually apply the input’s checked property you’ll need to toggle .active on the <label> manually", but I agree that's not a very nice way.

He3lixxx:
I thought about `preventDefault`, but I didn't quite see how it could be applicable here. From what I understand, if you have an event handler, calling `event.preventDefault()` will stop event propagation after your handler has finished (aka: The "more default" handlers that would come after your specialized handler will not be called). However, in this case, we _do_ want most parts of the default behaviour of bootstrap and the browser (radio input should be selected, button should get `.active`), and I dont see how that would work out.

The documentation you linked for bootstrap is nice, it seems to be easier than I thought. So, theoretically, this should also work:
```javascript
const contributorId = button.dataset.markNoAnswersFor;
const voteArea = document.querySelector(`#vote-area-${contributorId}`);

voteArea.querySelectorAll(".vote-btn").forEach(el => el.classList.remove("active"));
voteArea.querySelectorAll(".vote-inputs [type=radio][value='6']").forEach(radioInput =>{
    radioInput.checked = true;
    radioInput.closest(".vote-btn").classList.add("active");
});

sisyphus.saveAllData();

// hide questionnaire for contributor
$(voteArea).collapse("hide");
```

We have the advantage here that if we don't trigger a click, we don't need to monkey-patch sisyphus.
This could be worse if:
- We add some kind of click event listeners ourselves and introduce a bug when the "I can't evaluate this contributor" button is clicked (Seems somewhat reasonable, but I don't really see how this would happen. We shouldn't rely on `on-click` to be triggerend on every change)
- Bootstrap changes their "API" and we don't notice it (I don't think this will happen)

But in general, I think it's actually nicer - or is there some behaviour that `.click()` initially caused that I missed here @janno42 @pixunil?
Also, if we have #1545 merged, we could(?) and should add a frontend test for this, then I'd feel reasonably safe about it.

janno42:
The second solution seems a lot better to me. The click itself shouldn't be needed.
But don't we have to manually set `checked` to false for all other buttons as well?

He3lixxx:
> The second solution seems a lot better to me. The click itself shouldn't be needed.

Nice, will change the PR to use that.

> But don't we have to manually set `checked` to false for all other buttons as well?

The underlying inputs are radio inputs and [HTML guarantees](https://html.spec.whatwg.org/#radio-button-state-(type=radio)) that when we set one radio input of the group as checked, all others are unchecked. (Again: the html standard is incredibly readable - how did they do it? 😄)

Of course, it could break if either the radio buttons are not correctly grouped (which I think can't realistically happen) or we somehow switch away from radio buttons (which I see no reason for). I think the code is safe to go with for now, and we should add frontend tests for the vote page as soon as we have the infrastructure.

evap/student/templates/student_vote.html
```diff
@@ -299,9 +298,14 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
                 button.addEventListener("click", () => {
                     const contributorId = button.dataset.markNoAnswersFor;
                     const voteArea = document.querySelector(`#vote-area-${contributorId}`);
-                    // check "no answer" and uncheck all other options
-                    voteArea.querySelectorAll(".vote-inputs [type=radio][value='6']")
-                        .forEach(radioInput => radioInput.click());
+
+                    voteArea.querySelectorAll(".vote-btn").forEach(el => el.classList.remove("active"));
+                    voteArea.querySelectorAll(".vote-inputs [type=radio][value='6']").forEach(radioInput =>{
```
pixunil, line 303: :paintbrush:
```suggestion
                    voteArea.querySelectorAll(".vote-inputs [type=radio][value='6']").forEach(radioInput => {

File: evap/student/templates/student_vote.html [hunk 2]
```diff
@@ -299,9 +298,14 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
                 button.addEventListener("click", () => {
                     const contributorId = button.dataset.markNoAnswersFor;
                     const voteArea = document.querySelector(`#vote-area-${contributorId}`);
-                    // check "no answer" and uncheck all other options
-                    voteArea.querySelectorAll(".vote-inputs [type=radio][value='6']")
-                        .forEach(radioInput => radioInput.click());
+
+                    voteArea.querySelectorAll(".vote-btn").forEach(el => el.classList.remove("active"));
```
pixunil, line 302: I don’t know if it makes a difference, but wouldn’t this be even faster when pre-selecting with `voteArea.querySelectorAll(".vote-btn.active")`?

pixunil (approved):
Wow, I actually didn’t expect to notice a change, but I didn’t perceive how slow it was beforehand :)

I think your approach is sufficiently future-proof. As you state, `on-click` event handler shouldn’t be a correct use of this situation and it is unlikely that Bootstrap changes something depending on when you change the `checked` property or add a class.

janno42 (approved):

PR #1563, opened 2021-02-17
Grant eligible reward points on semester activation

janno42:
This will retroactively grant all eligible reward points for a semester once reward points are activated.

### DIFF
--- a/evap/rewards/views.py
+++ b/evap/rewards/views.py
@@ -16,7 +16,7 @@ from evap.staff.views import semester_view
 
 from evap.rewards.models import RewardPointGranting, RewardPointRedemption, RewardPointRedemptionEvent, \
                                 SemesterActivation, NoPointsSelected, NotEnoughPoints, RedemptionEventExpired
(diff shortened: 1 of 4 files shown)


evap/rewards/views.py
```diff
@@ -122,5 +122,7 @@ def semester_activation(request, semester_id, active):
     active = active == 'on'
 
     SemesterActivation.objects.update_or_create(semester=semester, defaults={'is_active': active})
+    if active:
+        grant_eligible_reward_points_for_semester(request, semester)
```
He3lixxx, line 126: We agreed on adding a confirmation modal (because this is irreversible from the UI). Modal should also contain a note that the request will take a while to process.

He3lixxx (commented):
Code and test looks good. We talked about the modal dialog.

evap/rewards/views.py
janno42: done

He3lixxx (approved):
nice, thanks

PR #1564, opened 2021-02-19
Add additional text answers for rating questions

### LINKED ISSUE #1542: Attached text answer fields for rating questions
Rating questions should get an attached text answer field that allows an additional text answer to be submitted for that rating question. The text answer field should be hidden and only be shown after clicking on a button/dropdown arrow/etc.
By default, every rating question should have this additional text answer field, but this must be configurable for each question in a questionnaire.

This is meant to be used in situations like a rating question stating a contributor "showed respect for the participants" and a participant wanting to add a specific comment to this situation. Having the text answer fields hidden does not increase the number of questions a participant feels pushed to answer while still giving the opportunity for additional specific feedback if required.

janno42:
work in progress to fix #1542 

some open todos:

- [x] uncheck and disable "allow textanswers" field when text question is selected in question form
- [x] hide additional textfield and add button to show it in vote form
- [x] show textfield on page load if not empty (locally saved answers)
- [x] don't show question text twice in vote form and on results page, use rows correctly
- [x] improve results view
- [x] refactor everything
- [x] test all the things


janno42:
preview of the vote page (top: initial state, bottom: after clicking on button)

image

---

image

janno42:
Tests are still missing, but I'm looking forward to your feedback on the current feature-complete implementation.

evap/staff/forms.py
```diff
@@ -650,13 +650,18 @@ def save(self, commit=True):
...
         widgets = {
             'text_de': forms.Textarea(attrs={'rows': 2}),
             'text_en': forms.Textarea(attrs={'rows': 2}),
             'order': forms.HiddenInput(),
         }
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self.instance.pk and self.instance.type in [Question.TEXT, Question.HEADING]:
+            self.fields['allow_textanswers'].disabled = True
```
He3lixxx, line 663: No custom clean method: If users change the type _and_ the allow_textanswers on the form, they can submit successfully (however, the save method will set allow_textanswers back to False). I think they should get an error. (JS should disable the checkbox for them, but django will not verify this.)

evap/staff/templates/staff_questionnaire_form.html
```diff
@@ -49,9 +49,18 @@ <h5 class="card-title">{% trans 'Questions' %}</h5>
...
+                                    <td class="question-type">
+                                        {% include 'bootstrap_form_field_widget.html' with field=form_element.type %}
+                                        <div class="form-check mt-1">
+                                            <input class="form-check-input allow-textanswer-checkbox" id="{{ form_element.allow_textanswers.id_for_label }}" name="{{ form_element.allow_textanswers.html_name }}"
+                                                type="checkbox"{% if form_element.allow_textanswers.value %} checked{% endif %}
+                                                {% if form_element.allow_textanswers.field.disabled %} disabled{% endif %} />
+                                            <label class="form-check-label" for="{{ form_element.allow_textanswers.id_for_label }}">{{ form_element.allow_textanswers.label }}</label>
+                                        </div>
+                                    </td>
+                                    <td>{% include 'bootstrap_form_field_widget.html' with field=form_element.DELETE %}</td>
```
He3lixxx, line 63: Why is it necessary to unroll the loop? I see that we somehow need to reference the input, but I don't see why the full unroll is necessary for that.

evap/staff/views.py
```diff
@@ -1068,7 +1068,7 @@ def get_evaluation_and_contributor_textanswer_sections(evaluation, filter_textan
         for questionnaire in contribution.questionnaires.all():
             text_results = []
 
-            for question in questionnaire.text_questions:
+            for question in questionnaire.questions.all():
```
He3lixxx, line 1071: (Not your change, but I think we should prevent O(N) queries here - probably something for another issue).

evap/student/forms.py
```diff
@@ -12,8 +12,45 @@ def __init__(self, label):
         super().__init__(label=label, required=False)
 
 
+class TextAnswerField(forms.CharField):
+
+    def __init__(self, related_question_id="", *args, **kwargs):
+        # pylint: disable=keyword-arg-before-vararg
```
He3lixxx, line 18: could we change it to
```
    def __init__(self, *args, related_question_id="", **kwargs):
```
? Same below.

evap/student/forms.py
```diff
@@ -12,8 +12,45 @@ def __init__(self, label):
...
+        self.related_question_id = related_question_id
+        kwargs['required'] = False
+        kwargs['widget'] = forms.Textarea(attrs={'related_question_id': self.related_question_id})
+        super().__init__(*args, **kwargs)
+
+    def deconstruct(self):
+        name, path, args, kwargs = super().deconstruct()
+        del kwargs['required']
```
He3lixxx, line 26: I'm not sure if I understand correctly what's happening here, but from what I understand, this is supposed to return information on how to construct the current instance. We `del kwargs['required']` here, because in the constructor, we will manually add it. Why isn't the same true for `kwargs['widget']`?

evap/student/views.py
```diff
@@ -149,6 +149,7 @@ def get_valid_form_groups_or_render_vote_page(request, evaluation, preview, for_
 
 @participant_required
 def vote(request, evaluation_id):
+    # pylint: disable=too-many-locals,too-many-nested-blocks,too-many-branches
```
He3lixxx, line 152: Maybe we should consider opening a refactor issue?

He3lixxx (changes requested):
Most parts of the code look good to me, I have some questions though.

evap/results/views.py
```diff
@@ -225,10 +226,14 @@ def remove_textanswers_that_the_user_must_not_see(evaluation_result, user, repre
...
-        questionnaire_result.question_results = [
-            result for result in questionnaire_result.question_results
-            if not isinstance(result, TextResult) or len(result.answers) > 0
-        ]
+        cleaned_results = []
+        for result in questionnaire_result.question_results:
+            if not isinstance(result, TextResult) or len(result.answers) > 0:
+                cleaned_results.append(result)
+            # if previous result was a RatingResult, update info about additional text result
+            elif cleaned_results and isinstance(cleaned_results[-1], RatingResult):
```
karyon, line 234: it took me a while to understand why this is correct :) how about `elif is_additional_text_result`?

evap/staff/templates/staff_questionnaire_form.html
```diff
@@ -82,6 +91,17 @@ <h5 class="card-title">{% trans 'Questions' %}</h5>
                 applySelect2(row.find("select"));
             };
             makeFormSortable("question_table", "questions", rowChanged, rowAdded, "", true, true);
+
+            // enable or disable checkbox for additional text answers
+            $('.question-type select').on('select2:select', function (e) {
+                const checkbox = $($(this).closest("td.question-type")).find("input");
+                if (e.params.data.id == 0 || e.params.data.id == 5) {  // 0: Text question; 5: Heading
+                    checkbox.prop("checked", false);
+                    checkbox.prop("disabled", true);
```
karyon, line 100: shouldn't it be possible to disable the form fields in the form class?

evap/student/forms.py
```diff
@@ -12,8 +12,45 @@ def __init__(self, label):
...
+        if self.related_question_id != "":
+            kwargs['related_question_id'] = self.related_question_id
+        return name, path, args, kwargs
+
+
+class RatingAnswerField(forms.TypedChoiceField):
+
+    def __init__(self, widget_choices, has_additional_textanswer=False, *args, **kwargs):
```
karyon, line 34: i would suggest `allows_textanswer`, `has_additional_textanswer_field`, or `allows_additional_textanswer` as parameter name.

karyon (commented):
Nothing groundbreaking except for the proposal to move the additional TextResult into the RatingResult, but I'm not sure whether it's worth it.

evap/student/forms.py
janno42: i copied it from the django docs but sure, they should better be swapped

evap/results/templates/results_evaluation_detail_questionnaires.html
```diff
@@ -14,17 +16,25 @@ <h4>{{ questionnaire_result.questionnaire.public_name }}</h4>
...
-            <div class="text-answer row">
-                <div class="col-question col-lg-broad col-lg-4">{{ question_result.question.text }}</div>
-                <div class="col-answer col-lg-broad col-lg-8 d-flex">
+            {% if not question_result.has_additional_text_result %}
+                </div>
+            {% endif %}
+        {% elif question_result|is_of_type:'TextResult' %}
+            {% if not question_result.is_additional_text_result %}
+                <div class="row">
+            {% endif %}
```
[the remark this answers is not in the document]
pixunil: I think the first step could be to replace `is_of_type` filters with static methods if possible, imo `is_text_result` and `is_rating_result` would be sensible here.

I couldn’t try it locally, but do I interpret it correctly that the question text is doubled if a rating question has additional text answers? If yes, why?

A more higher view proposol would be to _somehow_ implement a more tree-oriented structure. I mean, currently we have headers which divide the linear-looking question results into multiple sections, which feels more like two layers instead of one. Similary, we now put results for the same question in the same row. So maybe a reshuffling of the template data into: sections (with header) → question → results. Or the merging of additional text answers into the `RatingResult` helps, idk

pixunil (commented):
Didn’t finish yet, but want to give you at least some comments :)

evap/results/templates/results_evaluation_detail_questionnaires.html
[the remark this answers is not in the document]
janno42: i now used a different approach that should make this template much easier to read

evap/results/tools.py
```diff
@@ -156,17 +158,24 @@ def _get_results_impl(evaluation):
...
+                    continue
                 if question.is_rating_question:
                     if evaluation.can_publish_rating_results:
                         answer_counters = RatingAnswerCounter.objects.filter(contribution=contribution, question=question)
                     else:
                         answer_counters = None
-                    results.append(RatingResult(question, answer_counters))
-                elif question.is_text_question and evaluation.can_publish_text_results:
+                    has_additional_text_result = question.allow_textanswers and evaluation.can_publish_text_results
+                    results.append(RatingResult(question, answer_counters, has_additional_text_result=has_additional_text_result))
```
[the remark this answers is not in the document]
janno42: moving the additional text result into the rating result is probably better. i now implemented such a version and i think it helps.

evap/results/tools.py
```diff
@@ -156,17 +158,24 @@ def _get_results_impl(evaluation):
...
+                    results.append(HeadingResult(question=question))
+                    continue
                 if question.is_rating_question:
                     if evaluation.can_publish_rating_results:
                         answer_counters = RatingAnswerCounter.objects.filter(contribution=contribution, question=question)
                     else:
                         answer_counters = None
-                    results.append(RatingResult(question, answer_counters))
-                elif question.is_text_question and evaluation.can_publish_text_results:
+                    has_additional_text_result = question.allow_textanswers and evaluation.can_publish_text_results
```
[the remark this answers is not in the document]
janno42: empty text results are always removed by `remove_textanswers_that_the_user_must_not_see`

evap/staff/forms.py
janno42: my thoughts were that this part here disables the checkboxes on page load if the initial question type is `TEXT` or `HEADING`, so for questions that don't allow additional text answers.
the same happens via js when these types are selected when editing the form. in combination, this disallows selecting one of these types together with additional text answers in the ui. so there should be no case where any errors are shown. i think forcing correct input data in the ui is better than showing an error afterward.

we could then clean the data after submit to prevent any form data fiddling or - as i implemented it - always set `allow_textanswers` to `False` on saving the model if one of the disallowed types was chosen. you don't seem to like the current implementation, what would you prefer instead?
it is a staff form, so we don't have to fear people manually editing post data. by extending the save method, a consistent state is enforced on django side. we could also implement an additional database constraint, but that would still require the other things if we don't want to run into a database error on saving.
i think a falsely set `True` value for other question types in the end also would not break anything because `allow_textanswers` is only evaluated for rating questions.

evap/evaluation/models.py
```diff
@@ -934,6 +935,11 @@ class Meta:
         verbose_name = _("question")
         verbose_name_plural = _("questions")
 
+    def save(self, *args, **kwargs):
+        if self.type in [Question.TEXT, Question.HEADING]:
+            self.allow_textanswers = False
```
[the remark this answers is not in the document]
janno42: see below (https://github.com/e-valuation/EvaP/pull/1564#discussion_r614129803)

evap/staff/templates/staff_questionnaire_form.html
janno42: i used this to put the checkbox under the dropdown:
image

otherwise, it would be placed into its own `td` and would not have a label:
image

evap/student/forms.py
janno42: you are right, the widget is as well always set in the init method. so we can remove it as well. i changed that now.

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -14,41 +14,62 @@
...
+                                {% for choice, color in field|zip:field.field.widget.attrs.choices.colors %}
+                                    <label class="btn btn-sm vote-btn vote-btn-{{ color }} d-flex{% if field.value == choice.data.value %} active{% endif %}{% if field.errors %} choice-error{% endif %}">
+                                        <input id="{{ choice.id_for_label }}" name="{{ choice.data.name }}" type="radio" value="{{ choice.data.value }}" autocomplete="off"{% if field.value == choice.data.value %} checked{% endif %} />
+                                        <span class="m-auto vote-btn-text">{{ choice.choice_label|linebreaksbr }} {{ choice.id }}</span>
+                                    </label>
+                                {% endfor %}
+                                </div>
                             </div>
+                        {% else %}
+                            <div class="{% if related_question_id %}collapse collapse-{{ related_question_id }} {% endif %}col-lg-8 col-xl-7 p-0">
```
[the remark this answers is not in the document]
janno42: the toggle button has a `data-target` having this class as its value. this is used for the collapse effect for the additional text answer field (https://getbootstrap.com/docs/4.6/components/collapse/)

evap/student/tools.py
```diff
@@ -1,8 +1,8 @@
-def question_id(contribution, questionnaire, question):
+def question_id(contribution, questionnaire, question, additional_textanswer=False):
```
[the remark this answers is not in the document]
janno42: `answer_field_id` is probably the best choice, changing it.

evap/student/forms.py
janno42: went for `allows_textanswer`

evap/staff/templates/staff_questionnaire_form.html
janno42: yes, the fields are disabled on loading the form. this here is to disable them when editing the form and choosing one of the options that don't allow additional text answers.

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -14,41 +14,62 @@
...
-                                    <span class="m-auto vote-btn-text">{{ choice.choice_label|linebreaksbr }} {{ choice.id }}</span>
-                                </label>
-                            {% endfor %}
+                        {% if field|is_choice_field %}
+                            <div class="col-question col-lg-4 col-xl-5 d-flex my-auto{% if field.errors %} choice-error{% endif %}">
+                                <span class="mr-1">
+                                    {{ field.label }}
+                                </span>
+                                {% if has_additional_textanswer %}
+                                    <span class="ml-auto my-auto" data-toggle="tooltip" title="{% trans 'Add a text answer to this question' %}">
```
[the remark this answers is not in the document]
janno42: `align-self-end` is not even necessary when using `justify-content-between`

evap/results/views.py
```diff
@@ -224,16 +225,35 @@ def remove_textanswers_that_the_user_must_not_see(evaluation_result, user, repre
...
 
 
 def filter_text_answers(evaluation_result):
     for questionnaire_result in evaluation_result.questionnaire_results:
-        questionnaire_result.question_results = [result for result in questionnaire_result.question_results if isinstance(result, TextResult)]
+        question_results = []
+        for result in questionnaire_result.question_results:
+            if isinstance(result, TextResult):
+                question_results.append(result)
+            elif result.additional_text_result:
```
karyon, line 254: why don't you need to check for `isinstance(RatingResult)` before accessing `additional_text_result`?

evap/evaluation/models.py
```diff
@@ -934,6 +935,11 @@ class Meta:
         verbose_name = _("question")
         verbose_name_plural = _("questions")
 
+    def save(self, *args, **kwargs):
+        if self.type in [Question.TEXT, Question.HEADING]:
+            self.allows_textanswers = False
```
karyon, line 940: but... isn't that weird? a text question doesn't allow textanswers?

It's a bit verbose but I'd suggest naming this `allows_additional_textanswers`.

evap/staff/views.py
karyon: That should be simple to fix with something like `contribution.questionnaires.prefetch_related('questions')`. The `TextAnswer.objects.filter` below is still O(N) of course.

To not increase the number of queries (right now the PR adds queries for heading results and rating results that don't allow textanswers), you could add a `if not question.allows_textanswers): continue`.

evap/staff/views.py
```diff
@@ -1269,13 +1269,14 @@ def make_questionnaire_edit_forms(request, questionnaire, editable):
...
         editable_fields = ['visibility', 'is_locked', 'name_de', 'name_en', 'description_de', 'description_en', 'type']
+        editable_question_fields = ['allows_textanswers']
 
         for name, field in form.fields.items():
             if name not in editable_fields:
                 field.disabled = True
         for question_form in formset.forms:
             for name, field in question_form.fields.items():
-                if name != 'id':
+                if name not in editable_question_fields + ['id']:
```
karyon, line 1279: It's already disabled in the QuestionForm, why is that necessary here as well?

evap/staff/views.py
```diff
@@ -1296,8 +1297,7 @@ def questionnaire_edit(request, questionnaire_id):
 
     if form.is_valid() and formset.is_valid():
         form.save()
-        if editable:
```
karyon, line 1299: why could you remove this line?

evap/student/forms.py
```diff
@@ -22,23 +59,22 @@ def __init__(self, *args, contribution, questionnaire, **kwargs):
...
-                choices = CHOICES[question.type]
-                field = forms.TypedChoiceField(widget=forms.RadioSelect(attrs={'choices': choices}),
-                                               choices=zip(choices.values, choices.names),
-                                               coerce=int,
-                                               **field_args)
+                field = RatingAnswerField(
+                    widget_choices=CHOICES[question.type],
+                    choices=zip(CHOICES[question.type].values, CHOICES[question.type].names),
+                    label=question.text,
+                    allows_textanswer=question.allows_textanswers,
```
karyon, line 69: if you made RatingAnswerField (and perhaps TextAnswerField) only take a question as a parameter, the code would probably get simpler. They would also be less generic, but then again they are not meant to be generic.

evap/student/views.py
karyon: I looked through it and I think the code pretty much states what needs to be done, not less, not more. If anyone finds a way to make this simpler, sure go ahead! But splitting it up into multiple methods would, in my opinion, not make it simpler. It would just introduce a bunch of parameters that one also needs to read and understand, and I'd do it only if it helped with something else, e.g. if we could reasonably unit-test the individual methods.

karyon:
I think moving the additional text results into the rating results was really good, a bunch of questions marks that arose when I first read the code simply weren't there when I read this now :)

evap/staff/templates/staff_questionnaire_form.html
```diff
@@ -82,6 +91,17 @@ <h5 class="card-title">{% trans 'Questions' %}</h5>
                 applySelect2(row.find("select"));
             };
             makeFormSortable("question_table", "questions", rowChanged, rowAdded, "", true, true);
+
+            // enable or disable checkbox for additional text answers
+            $('.question-type select').on('select2:select', function (e) {
+                const checkbox = $($(this).closest("td.question-type")).find("input");
+                if (e.params.data.id == 0 || e.params.data.id == 5) {  // 0: Text question; 5: Heading
+                    checkbox.prop("checked", false);
```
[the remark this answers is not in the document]
janno42: that reminded me to fix the migration. the default value is true, but for text and heading questions its now always false (made sure by `save`) so the checkbox is unchecked and disabled.

evap/staff/views.py
janno42: the formset needs to be saved because the `allows_textanswers` field is an editable field also for forms that are otherwise not editable (see above in `make_questionnaire_edit_forms`)

evap/staff/views.py
janno42: the `allows_textanswers` does not get disabled here

evap/evaluation/models.py
janno42: done

evap/student/templates/student_vote_questionnaire_group.html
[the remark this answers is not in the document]
pixunil: Well, turns out I gave you the wrong instructions and you did the right thing either way :)
(`.my-auto` could be replaced with `.align-self-center`, but there is no big difference in intuition here)

evap/staff/templates/staff_questionnaire_form.html
```diff
@@ -49,9 +49,18 @@ <h5 class="card-title">{% trans 'Questions' %}</h5>
...
-                                        <td>{% include 'bootstrap_form_field_widget.html' with field=field %}</td>
-                                    {% endfor %}
+                                    <td>{% include 'bootstrap_form_field_widget.html' with field=form_element.text_de %}</td>
+                                    <td>{% include 'bootstrap_form_field_widget.html' with field=form_element.text_en %}</td>
+                                    <td class="question-type">
+                                        {% include 'bootstrap_form_field_widget.html' with field=form_element.type %}
+                                        <div class="form-check mt-1">
+                                            <input class="form-check-input allow-textanswer-checkbox" id="{{ form_element.allows_additional_textanswers.id_for_label }}" name="{{ form_element.allows_additional_textanswers.html_name }}"
+                                                type="checkbox"{% if form_element.allows_additional_textanswers.value %} checked{% endif %}
+                                                {% if form_element.allows_additional_textanswers.field.disabled %} disabled{% endif %} />
```
pixunil, line 59: It looks like you’re mirroring some basic form logic. In a little test, it worked fine for me to include in the `forms.py` the additional attribute as `widgets = {…, 'allows_additional_textanswers': forms.CheckboxInput(attrs={'data-keep': 'checked'})}` and replacing the `<input>…</input>` with just `{{ form_element.allows_additional_textanswers }}`.
Or was there a different reason I don’t see?

pixunil (commented):
Some random frontend stuff

evap/evaluation/templates/sortable_form_js.html
```diff
@@ -23,8 +23,8 @@
                 row.find('input[id$=-order]').val(row.parent().children().length);
 
                 // We have to empty the formset, otherwise sometimes old contents from
-                // invalid forms are copied
-                row.find("input:checkbox,:radio").removeAttr("checked");
+                // invalid forms are copied. The textanswer checkbox needs to stay checked.
+                row.find("input:checkbox:not(.allow-textanswer-checkbox),:radio").removeAttr("checked");
```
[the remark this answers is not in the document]
janno42: if it's just about mentioning the issue, i can add that reference now.

evap/staff/forms.py
He3lixxx: Hmm, I think enforcing on the frontend and in safe() is okay -- however, I still think that having only the check in `save` is a bit too weak and would vote in favor of also adding a [database constraint](https://docs.djangoproject.com/en/3.0/ref/models/constraints/#module-django.db.models.constraints).

evap/staff/templates/staff_questionnaire_form.html
janno42: yes, it works in part. but unfortunately, the label then isn't caught by the `:disabled` anymore, and the styling is broken.
so this needs additional changes.

janno42: it turns out this is more complicated than i thought and not worth it for me right now. will leave it as is.

evap/staff/forms.py
```diff
@@ -650,13 +650,18 @@ def save(self, commit=True):
...
+        fields = ('order', 'questionnaire', 'text_de', 'text_en', 'type', 'allows_additional_textanswers')
         widgets = {
             'text_de': forms.Textarea(attrs={'rows': 2}),
             'text_en': forms.Textarea(attrs={'rows': 2}),
             'order': forms.HiddenInput(),
         }
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self.instance.pk and self.instance.type in [Question.TEXT, Question.HEADING]:
```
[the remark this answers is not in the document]
janno42: would you prefer having it in the form's `clean` or instead in the model's `save`?

janno42:
squashed everything and added tests. looking for full reviews because i hope there aren't too many things left open.

evap/results/tools.py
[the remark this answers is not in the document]
He3lixxx: (I will assume this discussion is resolved. If not, please reopen)

evap/staff/forms.py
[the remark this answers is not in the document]
He3lixxx: "clean" so we can show useful errors on invalid input -- should be what is implemented right now

He3lixxx (approved):

PR #1567, opened 2021-03-01
Order evaluations on result index page

### LINKED ISSUE #1557: Order evaluations on result index page
Evaluations of a course are ordered alphabetically in the overview on the results detail pages of this course. On the results index page, however, they are not.
They should be ordered alphabetically there as well.

rgwohlbold:
Fixes #1557 
This commit introduces the same sorting behavior as the results detail page. The empty evaluation (same name as the course) is shown at the top.

### DIFF
--- a/evap/results/tests/test_views.py
+++ b/evap/results/tests/test_views.py
@@ -62,11 +62,8 @@ class TestResultsView(WebTest):
         evaluation2 = baker.make(Evaluation, name_de='random_evaluation_b', name_en='random_evaluation_b', course=course, state='published')
         evaluation1 = baker.make(Evaluation, name_de='random_evaluation_a', name_en='random_evaluation_a', course=course, state='published')
 
-        page = self.app.get(self.url, user=student)
-        elem1 = page.html.find('span', {'class':'evaluation-name'}, text=re.compile('.*{}.*'.format(evaluation1.name_en)))
-        elem2 = page.html.find('span', {'class':'evaluation-name'}, text=re.compile('.*{}.*'.format(evaluation2.name_en)))
-        evaluation_names = page.html.find_all('span', {'class':'evaluation-name'})
-        self.assertTrue(evaluation_names.index(elem1) < evaluation_names.index(elem2))
+        page = str(self.app.get(self.url, user=student))
+        self.assertTrue(page.index(evaluation1.name_en) < page.index(evaluation2.name_en))
 
 
     # using LocMemCache so the cache queries don't show up in the query count that's measured here
--- a/evap/results/tests/test_views.py
+++ b/evap/results/tests/test_views.py
@@ -1,6 +1,7 @@
 from unittest.mock import patch
 from io import StringIO
 import random
+import re
 
 from django.contrib.auth.models import Group
 from django.core.cache import caches
@@ -51,6 +52,22 @@ class TestResultsView(WebTest):
         self.assertNotContains(page, evaluation.full_name)
         self.assertNotContains(page, evaluation2.full_name)
         caches['results'].clear()
+
+
+    @patch('evap.evaluation.models.Evaluation.can_be_seen_by', new=(lambda self, user: True))
+    def test_order(self):
+        student = baker.make(UserProfile, email="student@institution.example.com")
+
+        course = baker.make(Course)
+        evaluation2 = baker.make(Evaluation, name_de='random_evaluation_b', name_en='random_evaluation_b', course=course, state='published')
+        evaluation1 = baker.make(Evaluation, name_de='random_evaluation_a', name_en='random_evaluation_a', course=course, state='published')
+
+        page = self.app.get(self.url, user=student)
+        elem1 = page.html.find('span', {'class':'evaluation-name'}, text=re.compile('.*{}.*'.format(evaluation1.name_en)))
+        elem2 = page.html.find('span', {'class':'evaluation-name'}, text=re.compile('.*{}.*'.format(evaluation2.name_en)))
+        evaluation_names = page.html.find_all('span', {'class':'evaluation-name'})
+        self.assertTrue(evaluation_names.index(elem1) < evaluation_names.index(elem2))
+
 
     # using LocMemCache so the cache queries don't show up in the query count that's measured here
     @override_settings(CACHES={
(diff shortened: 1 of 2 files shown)


rgwohlbold:
I looked into testing this. As far as I can tell, there is no integrated way to test the order of two elements in `django_webtest`. It would be possible to do by searching the entire HTML for occurrences of the evaluation titles and comparing their indices. My feeling is that this would not be a great way of going about it. What are your thoughts?

He3lixxx:
You are right: It's not a great way, but imho it's a very cheap way and gets the job done quite reliably without any additional hassle. We're usually just searching for the first occurence of the strings that should be ordered and comparing their indices

Some reasoning:
It is true that order of appearance in a web browser after styling is applied is not necessarily the same as the order of the DOM nodes. However, for accessibility reasons, you usually want to keep them as close together as possible, and we usually do that, so it usually works for us to simply assert node order in the DOM. I don't think we've ever had a case of bad ordering where such a test has or would have failed us.

Now, this would be a problem if we were to majorly redesign the way we display content (make evap a single page application), but this is very unlikely to happen, would take a huge effort, and adapting some small tests would not be the biggest concern then.

Some random example from the current code base (you can find quite some tests with pattern `test.*ordering`):
https://github.com/e-valuation/EvaP/blob/949189f9fec924d84fea132e245ca4af003fb26c/evap/contributor/tests/test_forms.py#L185

evap/results/tests/test_views.py
```diff
@@ -52,6 +52,19 @@ def test_multiple_evaluations_per_course(self):
...
         caches['results'].clear()
 
+
+    @patch('evap.evaluation.models.Evaluation.can_be_seen_by', new=(lambda self, user: True))
+    def test_order(self):
+        student = baker.make(UserProfile, email="student@institution.example.com")
+
+        course = baker.make(Course)
+        evaluation2 = baker.make(Evaluation, name_de='random_evaluation_b', name_en='random_evaluation_b', course=course, state='published')
+        evaluation1 = baker.make(Evaluation, name_de='random_evaluation_a', name_en='random_evaluation_a', course=course, state='published')
```
He3lixxx, line 62: You currently set `name_de`, which is never used as far as I see. I guess this was initially to switch the language and then check that their order changes? However, below, you only assert the order for the English page, not for the German one.

I guess it would probably be cool if you added the second check (the names would need to be so different that the ordering changes). Right now, this has a 50% chance of succeeding if the output order was random.

evap/results/tests/test_views.py
```diff
@@ -52,6 +52,19 @@ def test_multiple_evaluations_per_course(self):
         self.assertNotContains(page, evaluation2.full_name)
         caches['results'].clear()
 
+
+    @patch('evap.evaluation.models.Evaluation.can_be_seen_by', new=(lambda self, user: True))
+    def test_order(self):
+        student = baker.make(UserProfile, email="student@institution.example.com")
```
He3lixxx, line 58: The student doesn't need this email, right?

evap/results/tests/test_views.py
```diff
@@ -52,6 +52,19 @@ def test_multiple_evaluations_per_course(self):
...
+
+        course = baker.make(Course)
+        evaluation2 = baker.make(Evaluation, name_de='random_evaluation_b', name_en='random_evaluation_b', course=course, state='published')
+        evaluation1 = baker.make(Evaluation, name_de='random_evaluation_a', name_en='random_evaluation_a', course=course, state='published')
+
+        page = str(self.app.get(self.url, user=student))
```
He3lixxx, line 64: I'd slightly be in favor of using
```python
self.app.get(self.url, user=student).body.decode()
```
here -- sorry, I just saw the test I linked does it using `str()`. However, I feel that the string representation is more of a debug / user-facing thing, and we want just the html content, so this feels a bit cleaner to me.

(However, not a strong opinion, the current code also works)

He3lixxx (commented):
Quite nice so far, but I'd still change some things

evap/results/tests/test_views.py
rgwohlbold: Sounds reasonable, I changed it in the latest commit

evap/results/tests/test_views.py
rgwohlbold: If I don't set this email, the call to `self.app.get` fails since email is nullable and not set automatically. Then, django_webtest does not serialize the user correctly

evap/results/tests/test_views.py
rgwohlbold: This is more elegant than using `str()`, changed it in the latest commit

evap/results/tests/test_views.py
```diff
@@ -52,6 +52,22 @@ def test_multiple_evaluations_per_course(self):
         self.assertNotContains(page, evaluation2.full_name)
         caches['results'].clear()
 
+
```
He3lixxx, line 55: We try to style our code according to [pep8](https://www.python.org/dev/peps/pep-0008/#blank-lines), which says that
> Method definitions inside a class are surrounded by a single blank line.

So it would be cool if you could remove one blank line before and after this test

evap/results/tests/test_views.py
```diff
@@ -52,6 +52,22 @@ def test_multiple_evaluations_per_course(self):
...
+    @patch('evap.evaluation.models.Evaluation.can_be_seen_by', new=(lambda self, user: True))
+    def test_order(self):
+        student = baker.make(UserProfile, email="student@institution.example.com")
+
+        course = baker.make(Course)
+        evaluation1 = baker.make(Evaluation, name_de='random_evaluation_d', name_en='random_evaluation_a', course=course, state='published')
+        evaluation2 = baker.make(Evaluation, name_de='random_evaluation_c', name_en='random_evaluation_b', course=course, state='published')
+
+        page = self.app.get(self.url, user=student).body.decode()
+        self.assertTrue(page.index(evaluation1.name_en) < page.index(evaluation2.name_en))
```
He3lixxx, line 65: We should also be able to use `self.assertLess` and `self.assertGreater` here (see [docs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertLess)), which would be a little more precise I think.

He3lixxx (approved):
Code looks good and should work fine. I have two more annotations for code style, if you want you can adopt these, but otherwise I'm also fine with merging this as is.

janno42 (approved):
thank you! 👍

He3lixxx:
Also thanks from my side! :)

PR #1576, opened 2021-03-28
Allow sending HTML formatted emails

### LINKED ISSUE #1044: Allow sending HTML formatted emails
It would be cool if we could send emails as formatted HTML as well as plain text. This should not break compatibility for users using email clients that only allow plain text emails.

jb3rndt:
Fixes #1044
Every email is now sent with a plain text part and a html part. If no specific html body is set, the plain text gets copied into the html template just before sending the email.
For the email templates the html version now uses html lists for listing evaluations and some other styling for other things.
  
### And Mobile (Apple Mail):

  
One can notice some differences in the representation. Here are the major ones I'm trying to explain and for which I found no better solution:
1. The EvaP logo and header bar appear in different heights: I had many problems with email clients not displaying images in the sizes I wanted. In many cases they ignored any sizes of divs or table data tags (`<td>`) around them and inside the `` itself that were supposed to shrink the image to a specific size. Instead the image always appeared in original size. The only way that works an all email clients I tested is setting the `width` of the image to 100% in the `` tag (not even the inline css style attribute worked) and then use surrounding tags (here `<td>`) for the actual size. The `height` attribute didnt work as well. This is why the image is always 25% of the total device width. That causes it to be very thin on narrow screens and very thick on wide screens as it keeps the aspect ratio.
2. The size of the text boxes: In gmail there is much padding to the left and right of the text box which is unproportionally smaller on outlook and mobile. The thought behind it: On wide screens it is probably nice to shrink the text box in its width to have the message more centered. On narrow screens this would take up way to much space and the text would stretch in length. That is why i wanted to integrate media queries that apply different behavior depending on the screen width. Media queries only seem to have an effect on gmail which is why it looks weird in outlook.
3. I have no clue why outlook puts a white underline below the subject.
4. A bit conceptual: For now, the subject is part of the email template. I thought it would be nice to see the html email as a whole and thus have the subject repeated there. In addition to that it gives a bit more structure and variability to the template.

For the email templates in the staff view and the email operations for sending emails, the email body forms support changing the plain text and html body independent of each other. This is done with bootstrap pills which allow switching between the textareas.
This independent editing is only allowed for the above cases (unless I forgot any) because for emails that dont use any template we decided to not provide the ability of changing the contents independently for simplicity and integrity reasons.

Furthermore I'm not sure how to preferably code the UI for the forms. One can notice that I reused the `bootstrap_form.html` for the staff email templates by adding a case for email forms which uses the new `bootstrap_form_pill_group.html`. For the email operations however I adjusted the UI directly inside the `staff_email_preview_form.html` because there are the different id suffix(es?) and thus the `bootstrap_form.html` isnt used here. This could maybe be a place to refactor e.g. by adding the ability to add suffix(es?) to the `bootstrap_form.html` or something like that. Or maybe I dont understand why this cant be used in this case.
  
I hope I could explain the main things :)


evap/evaluation/templates/email_base.html
```diff
@@ -0,0 +1,88 @@
...
+    }
+  </style>
+</head>
+
+<body style="margin: 0px; font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, 'Helvetica Neue', Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol'; font-size: 1rem; font-weight: 400; line-height: 1.5; color: #212529; background-color: #ffffff;">
+  <table style="background-color: #495057; width: 100%; border-collapse: collapse;" cellspacing="0" cellpadding="0">
+    <tr>
+      <td style="width: 25%;">
+        <a href="https://evap.hpi.de">
+          <img src="https://evap.hpi.de/static/images/evap.png" width="100%" style="height: auto; display: block;"/>
```
He3lixxx, line 36: I read your paragraph about the image never scaling the way you wanted it to. I'm fine with any solution that (a) works with modern & widespread email clients and (b) is valid HTML / CSS. It would be very interesting to know how other big companies circumvent the problems that some email clients seem to have here.

The `width="100%"` tag is not valid html, as [per standard](https://html.spec.whatwg.org/#attr-dim-width), you can only give a number ("non-negative integer") there, which will then be interpreted as CSS pixels. Have you tested out the HTML5 `sizes` attribute? Maybe that is used in most clients and gives you the flexibility that you need.

evap/evaluation/templates/email_base.html
```diff
@@ -0,0 +1,88 @@
...
+      </td>
+    </tr>
+    <tr>
+      <td style="background-color: #495057; width: 100%; height: 25px;"></td>
+    </tr>
+  </table>
+
+</body>
+
+</html>
```
He3lixxx, line 88: Missing newline -- maybe you want to set up your editor to add one?

evap/staff/templates/staff_email_preview_form.html
```diff
@@ -12,8 +12,24 @@
                 <div class="form-field"><input class="form-control" name="email_subject{{ id_suffix }}" type="text" value="{{ email_template.subject }}" /></div>
             </div>
             <div class="form-group d-flex">
-                <div class="form-label">Email body</div>
-                <div class="form-field"><textarea class="form-control" cols="40" rows="10" name="email_body{{ id_suffix }}">{{ email_template.body }}</textarea></div>
+                <div class="form-label">
+                    <div class="nav nav-pills btn-switch btn-group-vertical float-right" role="tablist">
+                        <a class="p-1 space-normal font-weight-bold btn btn-light nav-link show active" href="#tab-plain{{ id_suffix }}" data-toggle="pill" role="tab" aria-controls="pills-plain{{ id_suffix }}" aria-selected="true">
```
He3lixxx, line 17: You're using an `id_suffix` for all ids here -- do we need them? For me, it looks like we could leave them out.

He3lixxx (changes requested):
Looks pretty nice overall. I had some places that confused me, see the comments.

evap/evaluation/models.py
```diff
@@ -31,6 +32,10 @@
 logger = logging.getLogger(__name__)
 
 
+from email import charset
+charset.add_charset('utf-8',charset.SHORTEST,None,'utf-8')
```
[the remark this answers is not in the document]
jb3rndt: Yeah sorry that is outdated and my fault. In the past I tried to change the email encoding to something different but didnt help anything. I'll remove it.

evap/evaluation/migrations/0121_add_html_body.py
```diff
@@ -0,0 +1,25 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='emailtemplate',
+            name='html_body',
+            field=models.TextField(default='', validators=[evap.evaluation.models.validate_template], verbose_name='HTML'),
+            preserve_default=False,
+        ),
+        migrations.AlterField(
+            model_name='emailtemplate',
+            name='body',
```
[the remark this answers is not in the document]
jb3rndt: We only briefly discussed the verbose names. So I am happy to change the field to something like `plain_body` and `html_body` or `plain_text` and `html_body` maybe?

evap/evaluation/models.py
```diff
@@ -1627,14 +1633,18 @@ def send_to_user(self, user, subject_params, body_params, use_cc, additional_cc_
 
         subject = self.render_string(self.subject, subject_params)
         body = self.render_string(self.body, body_params)
+        self.html_body = body if self.html_body == '' else self.render_string(self.html_body, body_params)
+        html_params = dict({'content': self.html_body, 'subject': subject}, **body_params)
+        html_body = render_to_string('email_base.html', html_params)
```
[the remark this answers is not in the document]
jb3rndt: 1. Paragraph: This step takes place immediately before sending the email. So for my understanding there is just an attribute of a class that gets changed and there is no further user interaction with the EmailTemplate object after this point. But maybe I'm missing something here.
2. Paragraph: I dont exactly understand what you mean as I cannot uniquely identify the code your `render_to_string(self.html_body)` reference references, but let me explain what has to be done before sending the html email, maybe that clarifies some things. First, the text of the input field has to be rendered so that the django template tags are resolved. That happens in line 1636. It either renders the `self.html_body` or just takes the `body` which is already rendered. In the next step the html template has to be filled with the subject and the content that just got rendered. This html template gets rendered in line 1638. The preparation for the rendering context parameters happens in 1637. One could argue that we should not override the `self.html_body` attribute but rather create a temporary variable.
3. Paragraph: I'm not sure if I discussed that with someone or if that was my subconsciuos understanding of the whole issue. Independent of that I think it would be cool to send every single email as a html email. For now, the only arguments I see are the following:
pro: Receivers of emails could wonder why some emails are just plain and some are html emails.
con: Users of the website are sending individual emails via the website and are wondering why their simple emails are wrapped into a big html email.

evap/evaluation/templates/email_base.html
```diff
@@ -0,0 +1,88 @@
...
+  <style>
+    @media (max-width: 768px) {
+      .spacer {
+        width: 0px;
+      }
+
+      .content {
+        width: 100%;
+      }
+    }
```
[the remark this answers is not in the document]
jb3rndt: It is valid afaik. I think its only allowed for media queries. It actually works on gmail but doesnt on outlook.

evap/evaluation/templates/email_base.html
jb3rndt: While trying to understand how all of this works I looked into some html emails of some companies. But other than hard-coded image sizes I was not able to find anything special in the auto-generated templates with a ton of lines.
Sadly, the sizes attribute didnt help but I probably got it working with the width and a hard-coded pixel value. But I still have to test it.

evap/evaluation/templates/email_base.html
jb3rndt: Yup, did it, thx

evap/evaluation/migrations/0121_add_html_body.py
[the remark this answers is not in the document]
He3lixxx: I think `html_body` is fine, I'm unsure about the other name. I think it should end in `_body` as well, so I'd consider `plain_text_body` or `plain_body`. What do think?

evap/evaluation/models.py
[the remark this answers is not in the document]
He3lixxx: 1 and 2: I mixed two different thoughts here, let me try to make them clear(er).

I think my key point is: Methods should not have unexpected side effects.

The method name `send_to_user` does not imply any change of state on the object it is called on, so it should not change the state. This might work fine for the current code we have, but keep in mind any future code that might be written. The following code seems perfectly fine:
```python
# Send out emails to all participants
template = EmailTemplate.objects.get(type=PARTICIPANT_NOTIFICATION, ...)
for user in course.users:
    template.send_to_user(user, ...)
```

However, `template.html_body` is "corrupted" after the first loop iteration. So, as you said, we shouldn't overwrite the attribute but use a temporary variable.

---

Now, I first thought you wanted to achieve this behavior:
```
If we have no html_body yet:
    Set html_body to plain_text_body
    save the template
```

This would act as a kind of "migration", i.e. for every email template that we use, we lazily give it some html_body in case it doesn't have one. This brings us to point 3 here.

I dislike this idea. If we send html, we should utilize proper html in these emails, that is, we should use html lists and paragraphs and so on. Keep in mind that e.g. line breaks, that might be carefully crafted for the plain text mails, will be ignored for the rendering of html.

I agree that it would be cool to uniformly send out all emails as html emails, but imo this should be done by manually "translating" all templates.

---

In the end, I'd expect this method to behave like this:
```python
subject = self.render_string(self.subject, subject_params)
body = self.render_string(self.body, body_params)
mail = EmailMultiAlternatives(...)

if self.html_body:
    # notice mismatch here: we use the words "body" and "content". Maybe we can find a unified naming scheme?
    # Keep in mind: In the context of html, "body" is pretty strictly defined term that we're misusing here.
    rendered_content = self.render_string(self.html_body, body_params)
    wrapper_template_params = dict({'content': rendered_content, 'subject': subject}, **body_params)
    wrapped_content = render_to_string('email_base.html', wrapper_template_params)
    mail.attach_alternative(wrapped_content, "text/html")
```

evap/evaluation/templates/email_base.html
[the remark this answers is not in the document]
He3lixxx: Ah, indeed, I didn't know that. 👍

evap/evaluation/models.py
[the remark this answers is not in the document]
jb3rndt: > 
> 
> 1 and 2: I mixed two different thoughts here, let me try to make them clear(er).
> 
> I think my key point is: Methods should not have unexpected side effects.
> 
> The method name `send_to_user` does not imply any change of state on the object it is called on, so it should not change the state. This might work fine for the current code we have, but keep in mind any future code that might be written. The following code seems perfectly fine:
> 
> ```python
> # Send out emails to all participants
> template = EmailTemplate.objects.get(type=PARTICIPANT_NOTIFICATION, ...)
> for user in course.users:
>     template.send_to_user(user, ...)
> ```
> 
> However, `template.html_body` is "corrupted" after the first loop iteration. So, as you said, we shouldn't overwrite the attribute but use a temporary variable.
> 
Okay, I see your point.
> Now, I first thought you wanted to achieve this behavior:
> 
> ```
> If we have no html_body yet:
>     Set html_body to plain_text_body
>     save the template
> ```
> 
> This would act as a kind of "migration", i.e. for every email template that we use, we lazily give it some html_body in case it doesn't have one. This brings us to point 3 here.
> 
> I dislike this idea. If we send html, we should utilize proper html in these emails, that is, we should use html lists and paragraphs and so on. Keep in mind that e.g. line breaks, that might be carefully crafted for the plain text mails, will be ignored for the rendering of html.
> 
Linebreaks are taken care of by the `|linebreaksbr` template filter, but that does not greatly weaken your argument 😅. So yes, I see your point.
> I agree that it would be cool to uniformly send out all emails as html emails, but imo this should be done by manually "translating" all templates.
> 
If you mean the staff email templates like participant notification, they all have a separate html body that uses html lists for example (see `test_data.json` lines 128103 ff.). So I guess we are only talking about individual emails that a user writes by himself on the website and then sends out to others.
To sum it up: If we would want to send every email as an html email, we should give the user a separate html email text field for editing the html body of individual emails. And as we already decided against that we shouldn't send every email as html email. Unless we find another solution...
I dont know how often such individual email get sent and whether they often have lists that should be html but that is probably no argument. So I guess I can live with only sending html emails if they have a "staff template" ^^.
> In the end, I'd expect this method to behave like this:
> 
> ```python
> subject = self.render_string(self.subject, subject_params)
> body = self.render_string(self.body, body_params)
> mail = EmailMultiAlternatives(...)
> 
> if self.html_body:
>     # notice mismatch here: we use the words "body" and "content". Maybe we can find a unified naming scheme?
>     # Keep in mind: In the context of html, "body" is pretty strictly defined term that we're misusing here.
>     rendered_content = self.render_string(self.html_body, body_params)
>     wrapper_template_params = dict({'content': rendered_content, 'subject': subject}, **body_params)
>     wrapped_content = render_to_string('email_base.html', wrapper_template_params)
>     mail.attach_alternative(wrapped_content, "text/html")
> ```
Looks good :)

evap/evaluation/migrations/0121_add_html_body.py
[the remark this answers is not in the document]
jb3rndt: I think I'd prefer `html_body` and `plain_body`.

evap/evaluation/models.py
```diff
@@ -1642,7 +1639,12 @@ def send_to_user(self, user, subject_params, body_params, use_cc, additional_cc_
             cc=cc_addresses,
             bcc=[a[1] for a in settings.MANAGERS],
             headers={'Reply-To': settings.REPLY_TO_EMAIL})
-        mail.attach_alternative(html_body, "text/html")
+
+        if self.html_body:
+            rendered_html_email_body = self.render_string(self.html_body, body_params)
+            html_template_params = dict({'email_body': rendered_html_email_body, 'email_subject': subject}, **body_params)
+            wrapped_html_email_body = render_to_string('email_base.html', html_template_params)
+            mail.attach_alternative(wrapped_html_email_body, "text/html")
```
jb3rndt, line 1647: > ```python3
> if self.html_body:
>     # notice mismatch here: we use the words "body" and "content". Maybe we can find a unified naming scheme?
>     # Keep in mind: In the context of html, "body" is pretty strictly defined term that we're misusing here.
>     rendered_content = self.render_string(self.html_body, body_params)
>     wrapper_template_params = dict({'content': rendered_content, 'subject': subject}, **body_params)
>     wrapped_content = render_to_string('email_base.html', wrapper_template_params)
>     mail.attach_alternative(wrapped_content, "text/html")
> ```

As @He3lixxx already mentioned, the naming here is pretty difficult/confusing. I tried it this way, but it still feels weird. So maybe you have some ideas?

evap/staff/templates/staff_email_preview_form.html
jb3rndt: Do you mean this `aria-controls="pills-plain{{ id_suffix }}"` for example? I think for accessibility the id_suffix would be useful here aswell but I dont know how much we focus on accessibility? I could also leave these aria things completely out if you want´.

He3lixxx: I'm talking about all locations where `{{ id_suffix }}` is used. Having aria-attributes is fine.

My point is that we could have `href="#tab-plain"` and `aria-controls="pills-plain"` -- we don't have multiple "items" here, so adding an id suffix doesn't help with anything, or am I missing something?

jb3rndt: In the case of an evalutation operation in which both contributors and participants can get reminded, this form is imported twice. And then we would have colliding `href` attributes which breaks the pill switching mechanism by bootstrap.
And for all aria things they are not needed, but I guess it would be consistent in the described case.

He3lixxx: That makes sense, thanks.

evap/evaluation/migrations/0121_add_html_body.py
```diff
@@ -0,0 +1,30 @@
...
+
+    dependencies = [
+        ('evaluation', '0120_use_django_db_models_jsonfield'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='emailtemplate',
+            name='body',
+        ),
```
He3lixxx, line 17: We should use RenameField and AlterField here. This RemoveField followed by AddField would delete all email templates when migrating.

evap/evaluation/models.py
```diff
@@ -31,6 +32,8 @@
 logger = logging.getLogger(__name__)
 
 
+
+
```
He3lixxx, line 36: I think we want 2 empty lines here, so I'd remove these again (very interesting that codacy doesn't complain here, I don't think that this is pep8 compliant)

evap/staff/tests/test_views.py
```diff
@@ -2432,19 +2436,27 @@ def setUpTestData(cls):
 
     def test_emailtemplate(self):
         """
-            Tests the emailtemplate view with one valid and one invalid input datasets.
+            Tests the emailtemplate view with one valid and two invalid input datasets.
```
He3lixxx, line 2439: I think its fair to just remove the comment, it doesn't add much value (however, now that you've changed it, I don't mind, we can also keep it)

evap/staff/tests/test_views.py
```diff
@@ -2432,19 +2436,27 @@ def setUpTestData(cls):
...
+        form["plain_body"] = " invalid tag: {{}}"
         form.submit()
-        self.assertEqual(EmailTemplate.objects.get(pk=1).body, "body: mflkd862xmnbo5")
+        self.assertEqual(EmailTemplate.objects.get(pk=1).plain_body, "plain_body: mflkd862xmnbo5")
+        self.assertEqual(EmailTemplate.objects.get(pk=1).html_body, "html_body: <p>mflkd862xmnbo5</p>")
+
+        form["html_body"] = ""
+        form.submit()
+        self.assertEqual(EmailTemplate.objects.get(pk=1).plain_body, "plain_body: mflkd862xmnbo5")
+        self.assertEqual(EmailTemplate.objects.get(pk=1).html_body, "html_body: <p>mflkd862xmnbo5</p>")
```
He3lixxx, line 2459: Sooo we can't delete the html_body? Is this what we want? I expected an invalid tag test here, just as it was with the plain body.

evap/evaluation/models.py
```diff
@@ -1626,16 +1630,22 @@ def send_to_user(self, user, subject_params, body_params, use_cc, additional_cc_
...
             to=[user.email],
             cc=cc_addresses,
             bcc=[a[1] for a in settings.MANAGERS],
             headers={'Reply-To': settings.REPLY_TO_EMAIL})
 
+        if self.html_body:
+            rendered_html_email_body = self.render_string(self.html_body, body_params)
+            html_template_params = dict({'email_body': rendered_html_email_body, 'email_subject': subject}, **body_params)
+            wrapped_html_email_body = render_to_string('email_base.html', html_template_params)
+            mail.attach_alternative(wrapped_html_email_body, "text/html")
```
He3lixxx, line 1647: The template might contain user-specific data that the user might also be able to set themselves. so, using `self.render_string`, which sets `autoescape=False` isn't correct here (it might work for the current test data, but I'd rather have correct escaping in place. I'd even add a test for that.).

To test this, insert a `<` into a user's name. This must be escaped to `&lt;` in `rendered_html_email_body` (which with the current code shouldn't happen, if I'm not mistaken). It may be a good idea to also rename render_string to indicate that it disables autoescaping.

Then, the `render_to_string` call must not escape html, as we want to use the tags already in the content. This seems to already be the case -- but why? I quickly tried to find why no escaping happens here, but I didn't find anything.

He3lixxx (changes requested):

evap/evaluation/migrations/0121_add_html_body.py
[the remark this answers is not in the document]
janno42: i think we should use `html_content` and `plain_content`

evap/evaluation/models.py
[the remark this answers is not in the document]
He3lixxx: In a discussion with @janno42 we agreed that in most cases, we will only send the templates in the database. These will have `html_content` set. For the rare occasion that we send a custom email, where staff users only provide a `text_content`, we prefer a uniform look with the other emails, so we will take the plaintext, pass it through the `linebreaksbr` filter and then put it into the HTML email -- even though it may not be utilizing HTML properly in this case.

evap/evaluation/migrations/0121_add_html_body.py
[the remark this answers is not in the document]
jb3rndt: Ok, so while discussing https://github.com/e-valuation/EvaP/pull/1576#discussion_r611814588 we figured out that `plain_content` and `html_content` would be the best choice here.

evap/evaluation/models.py
jb3rndt: > ```python3
> if self.html_body:
>     # notice mismatch here: we use the words "body" and "content". Maybe we can find a unified naming scheme?
>     # Keep in mind: In the context of html, "body" is pretty strictly defined term that we're misusing here.
>     rendered_content = self.render_string(self.html_body, body_params)
>     wrapper_template_params = dict({'content': rendered_content, 'subject': subject}, **body_params)
>     wrapped_content = render_to_string('email_base.html', wrapper_template_params)
>     mail.attach_alternative(wrapped_content, "text/html")
> ```
We agreed on this suggestion.

evap/evaluation/models.py
jb3rndt: Yep that is a good point. I change it so the html content parameters and the subject string that is inserted into the template get escaped. For now I just changed the `self.render_string` method to optionally take an `autoescape` boolean that defaults to `False`. I dont know if that is the solution you looked forward to, but I think its not necessary to rename the method if we leave it this way. Especially because no autoescape it still mostly used.
For the `render_to_string` method: I could not find anything yet either and it is pretty strange...

evap/evaluation/templates/email_base.html
```diff
@@ -0,0 +1,79 @@
...
+  <meta name="viewport" content="width=device-width; initial-scale=1; shrink-to-fit=no"/>
+  <title>EvaP</title>
+  <style>
+    @media (min-width: 768px) {
+      .spacer {
+        width: 15% !important;
+      }
+
+      .email-content {
+        width: 70% !important;
```
janno42, line 16: just a question because i'm not sure: do you think it would be a good idea to set a max width here? for the website, the main content part has a max width so it doesn't span the whole screen in large browser windows.

evap/evaluation/templates/bootstrap_form.html
```diff
@@ -1,4 +1,13 @@
 {% include 'bootstrap_form_errors.html' with errors=form.non_field_errors %}
-{% for field in form %}
-    {% include 'bootstrap_form_field.html' with field=field %}
-{% endfor %}
+{% if email_form and form.html_content in form.visible_fields %}
```
janno42, line 2: would it be possible (without a lot of additional work) to generalize this a bit more? in case we want to use such pill group forms in other places in the future it would be nice if this part isn't specialized on the `email_form`.

evap/evaluation/templates/email_base.html
```diff
@@ -0,0 +1,79 @@
...
+      </td>
+      <td class="email-content" style="padding: 4%; width: 100%">
+        <table style="background-color: white; width: 100%; border: 2px solid #F0F0F0; border-collapse: collapse;">
+          <tr>
+            <td style="text-align: center; font-size: 18px; padding: 1em; background-color: #F8F8F8; border-bottom: 2px solid #F0F0F0">
+              {{ email_subject|linebreaksbr }}
+            </td>
+          </tr>
+          <tr>
+            <td style="padding: 2%">
```
janno42, line 47: regarding the non-fixed width mentioned above, percentage values for padding here and in other places feel a bit strange. how does it look when stretching the window?

evap/evaluation/models.py
```diff
@@ -1625,25 +1628,35 @@ def send_to_user(self, user, subject_params, body_params, use_cc, additional_cc_
...
             if send_separate_login_url:
                 self.send_login_url_to_user(user)
         except Exception:  # pylint: disable=broad-except
             logger.exception('An exception occurred when sending the following email to user "{}":\n{}\n'.format(user.full_name_with_additional_info, mail.message()))
 
+    def construct_mail(self, to_email, cc_addresses, subject_params, body_params):
+        subject = self.render_string(self.subject, subject_params)
+        plain_content = self.render_string(self.plain_content, body_params)
+
+        rendered_content = self.render_string(self.html_content if self.html_content else self.plain_content, body_params, True)
```
janno42, line 1645: can you please make this `..., autoescape=True)`?

evap/static/scss/evap.scss
```diff
@@ -138,6 +138,10 @@ td.icon {
     padding-left: 4px;
 }
 
+.space-normal {
+    white-space: normal;
```
janno42, line 142: why do we need this?

evap/evaluation/templates/email_base.html
jb3rndt: Yes thats much better. I used `max-width: 1000px` because `720px` seemed a bit less and `1000px` was the next threshold used on the website as well. Do you think that is ok?

janno42: yes, seems reasonable

evap/static/scss/evap.scss
janno42: we discussed that the line breaks in the email form buttons would be strange otherwise.

evap/evaluation/templates/bootstrap_form.html
jb3rndt: Is b0ddb08 a fitting solution for that?

janno42: it's beautiful! 👍

evap/staff/templates/staff_email_preview_form.html
```diff
@@ -12,8 +12,24 @@
                 <div class="form-field"><input class="form-control" name="email_subject{{ id_suffix }}" type="text" value="{{ email_template.subject }}" /></div>
             </div>
             <div class="form-group d-flex">
-                <div class="form-label">Email body</div>
-                <div class="form-field"><textarea class="form-control" cols="40" rows="10" name="email_body{{ id_suffix }}">{{ email_template.body }}</textarea></div>
+                <div class="form-label">
+                    <div class="nav nav-pills btn-switch btn-group-vertical float-right" role="tablist">
+                        <a class="p-1 space-normal font-weight-bold btn btn-light text-center nav-link show active" href="#tab-plain{{ id_suffix }}" data-toggle="pill" role="tab" aria-controls="pills-plain{{ id_suffix }}" aria-selected="true">
+                            Plain Text
```
janno42, line 18: translation missing here and below :)

evap/staff/templates/staff_evaluation_email.html
```diff
@@ -12,27 +12,32 @@
...
                                         </div>
-                                        <button name="export" type="submit" formnovalidate class="btn btn-sm btn-secondary">{% trans 'Show Recipients' %}</button>
-                                    </div>
-                                </div>
-                            {% else %}
-                                {% include 'bootstrap_form_field.html' with field=field %}
-                            {% endif %}
-                        {% endfor %}
+                                    {% else %}
+                                        {% include 'bootstrap_form_field.html' with field=field email_form=True %}
```
janno42, line 35: this wasn't updated yet

janno42 (changes requested):

evap/staff/templates/staff_evaluation_email.html
janno42: oh sorry, `email_form` was already used before here.

jb3rndt:
For sending emails in debug mode to your real email to view html emails:
```
localsettings.py:

EMAIL_BACKEND = 'django.core.mail.backends.smtp.EmailBackend'
EMAIL_HOST = 'smtp.gmail.com'
EMAIL_PORT = 587
EMAIL_USE_TLS = True
EMAIL_HOST_USER = <your_email_address>
EMAIL_HOST_PASSWORD = <your_email_password>
```
Must be google mail for this config.

You now need another email to send the emails to. Can be any email I think.
```
models.py: L1649
mail = EmailMultiAlternatives(
            subject=subject,
            body=plain_content,
            to=[<your_other_email_address>],#[to_email],
            cc=[],#cc_addresses,
            bcc=[],#[a[1] for a in settings.MANAGERS],
            headers={'Reply-To': settings.REPLY_TO_EMAIL},
            alternatives=[(wrapped_content, "text/html")])
```

evap/staff/templates/staff_email_preview_form.html
```diff
@@ -1,24 +1,24 @@
...
+         {% if heading %}<h4 class="card-title">{{ heading }}</h4>{% endif %}
         <div class="form-check mb-2">
             <input class="form-check-input" id="send_email{{ id_suffix }}" type="checkbox" name="send_email{{ id_suffix }}"
                 checked onclick="$('#email_form{{ id_suffix }}').toggle();" />
             <label class="form-check-label" for="send_email{{ id_suffix }}">{% trans 'Send email notifications' %}</label>
         </div>
         <div class="email-form" id="email_form{{ id_suffix}}">
             <div class="form-group d-flex">
-                <div class="form-label">Email subject</div>
+                <div class="form-label">{% trans 'Email subject' %}<div>
```
janno42, line 11: `</div>` :)

jb3rndt: Aaah I was wondering why the site is so weird an didnt find the issue, thanks :D

evap/evaluation/tests/test_models.py
```diff
@@ -657,6 +657,28 @@ def test_missing_email_address():
...
+        template = EmailTemplate(subject='Example', plain_content='Example plain content', html_content='<p>Example html content</p>')
+        template.send_to_user(self.user, {}, {}, False)
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertTrue(isinstance(mail.outbox[0], mail.message.EmailMultiAlternatives))
+        self.assertEqual(mail.outbox[0].body, 'Example plain content')
+        self.assertEqual(len(mail.outbox[0].alternatives), 1)
+        self.assertEqual(mail.outbox[0].alternatives[0][1], 'text/html')
+        self.assertIn('<p>Example html content</p>', mail.outbox[0].alternatives[0][0])
+
+    def test_copy_plain_content_on_empty_html_content(self):
```
He3lixxx, line 670: Maybe add a newline to check that it also is converted to a HTML linebreak?

He3lixxx (commented):
Code looks good. I will still have to validate generated HTML, didn't do that yet.

Have you considered adding your instructions on how to get the emails sent to you to our documentation? I think this might be useful in the future.

evap/evaluation/tests/test_models.py
jb3rndt: Yep, fixed in `27d8262`. (I added this case to `test_copy_plain_content_on_empty_html_content` because it fits well in there)

evap/development/fixtures/test_data.json
```diff
@@ -128101,7 +128101,8 @@
...
... one-time URL you can login to the evaluation platform: <a href=\"{{ login_url }}\">{{ login_url }}</a>{% elif user.needs_login_key %}We will send you a one-time login URL in a separate email.{% endif %}\r\n\r\nTo prepare your evaluations we would like to ask you for the following within a week:\r\n<ul><li>Is the evaluation period appropriate? Please let the evaluation end before the final exam (written or oral examination, final assignment, etc.) of your course.</li><li>Are the selected questionnaires adequate for the evaluation? Please adapt the selection if necessary.</li><li>Are all contributors (lecturers, teaching assistants, etc.) included in the evaluation? Please add all additional persons with their appropriate questionnaires.</li></ul>\r\nThese evaluations need your approval:\r\n<ul>{% for evaluation in evaluations|order_by:\"full_name_en\" %}<li>{{ evaluation.full_name_en }}</li>{% endfor %}</ul>\r\nThank you very much in advance for your efforts!\r\nIf you have any questions or feedback, please contact the evaluation team (<a href=\"mailto:{{ contact_email }}\">{{ contact_email }}</a>).\r\n\r\nKind regards,\r\nthe Evaluation Team\r\n\r\n(This is an automated message.)"
```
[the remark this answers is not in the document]
jb3rndt: We agreed on not converting newlines to `<br />` by using the `linebreaksbr` filter when rendering the html content for the email. The content now has tu use `<br />` explicitly to have linebreaks in the html email. With that we can better control the linebreaks and have an independent layout of the html templated when editing them.
However, when using the plain content for the html email, the newlines have to be replaced with `<br />`.
(See `cb1cd47` for the changes.)

evap/evaluation/templates/email_base.html
```diff
@@ -0,0 +1,64 @@
...
+</head>
+
+<body style="margin: 0px; padding: 0px; font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, 'Helvetica Neue', Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol'; font-size: 1rem; font-weight: 400; line-height: 1.5; color: #212529; background-color: #ffffff;">
+  <table style="background-color: #495057; width: 100%; border-collapse: collapse;" cellspacing="0" cellpadding="0">
+    <tr>
+      <td>
+        <a href="{{ page_url }}">
+          <img src="https://evap.hpi.de/static/images/evap.png" height="50" style="display: block;"/>
+        </a>
+      </td>
```
He3lixxx, line 18: Do we need this empty td? I don't see how it could change anything? Maybe we could remove the whole table if just one row and one column is left?

He3lixxx (approved):
Code looks good. I haven't tested anything by hand.

janno42 (approved):

evap/evaluation/templates/email_base.html
jb3rndt: You are right, it is not needed. But replacing the table with e.g. a `div` sadly breaks the template on my mobile email client, so I'm going to leave the table in.

janno42:
Nice! :shipit:

PR #1577, opened 2021-03-29
xmin update test: Link relevant doku page to also ensure vacuum + cluster

### LINKED ISSUE #1384: It is possible to reconstruct the author of all text answers and some rating answers

He3lixxx:
We need to update this page and apply the changes on production when adding a new question type.

Fixes #1384

### DIFF
--- a/evap/student/tests/test_views.py
+++ b/evap/student/tests/test_views.py
@@ -305,7 +305,13 @@ class TestVoteView(WebTest):
         form[question_id(self.evaluation.general_contribution, self.top_general_questionnaire, self.top_grade_question)] = 2
         form.submit()
 
-        self.assertEqual(set(Answer.__subclasses__()), {RatingAnswerCounter, TextAnswer}, "This requires an update if a new answer type is added")
+        self.assertEqual(
+            set(Answer.__subclasses__()),
+            {RatingAnswerCounter, TextAnswer},
+            "This test requires an update if a new answer type is added. Also, when adding a new answer type, "
+            "the new table should probably also be vacuumed and clustered -- see and update "
+            "https://github.com/e-valuation/EvaP/wiki/Installation#database-vacuuming-and-clustering"
+        )
 
         query = RatingAnswerCounter.objects.raw("SELECT id, xmin FROM evaluation_ratinganswercounter")
         rating_answer_xmins = [row.xmin for row in query]


evap/student/tests/test_views.py
```diff
@@ -305,7 +305,13 @@ def test_xmin_of_all_answers_is_updated(self):
         form[question_id(self.evaluation.general_contribution, self.top_general_questionnaire, self.top_grade_question)] = 2
         form.submit()
 
-        self.assertEqual(set(Answer.__subclasses__()), {RatingAnswerCounter, TextAnswer}, "This requires an update if a new answer type is added")
+        self.assertEqual(
+            set(Answer.__subclasses__()),
+            {RatingAnswerCounter, TextAnswer},
+            "This test requires an update if a new answer type is added. Also, when adding a new question, "
```
janno42, line 311: shouldn't this be "answer" as well?
```suggestion
            "This test requires an update if a new answer type is added. Also, when adding a new answer type, "
```

He3lixxx: hah, that's true.

karyon (approved):

PR #1578, opened 2021-03-29
Add pip self-update to provision script. Fixes provisioning error.

He3lixxx:
Fixes current error when setting up the VM.

### DIFF
--- a/deployment/provision_vagrant_vm.sh
+++ b/deployment/provision_vagrant_vm.sh
@@ -11,7 +11,7 @@ export DEBIAN_FRONTEND=noninteractive
 
 # install python stuff
 apt-get -q update
-apt-get -q install -y python3.7 python3.7-dev python3.7-venv python3-pip gettext
+apt-get -q install -y python3.7 python3.7-dev python3-venv python3.7-venv gettext
 
 # install sass
 $REPO_FOLDER/deployment/install_dart_sass.sh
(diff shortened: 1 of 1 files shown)


janno42 (approved):

janno42 (approved):

He3lixxx:
This all is way to complicated...

So: We don't need pip installed on the system, which allows us to also not think about having version mismatches between our package manager and pip's own version.
This is because the venv module will use `ensurepip` to install a new pip-installation into any newly created venv. For python3.7, this is pip9. So, it doesn't make any difference if we update pip on the system or not, venvs are not affected. This means that we do not have to install the distribution package python3-pip.

Alright, now one might think it's enough to install python3.7-venv, and call `python3.7 -m venv some/path`. However, this doesn't work, as `python3.7-venv` (for whatever reason?) doesn't ship ensurepip. So, you have to install `python3-venv` additionally, which will also install python3.6 and also ship ensurepip. Alternatively, we could call `python3 -m venv some/path` to create the venv and only install python3-venv (without installing python3.7-venv).

I don't know what the compatibility is, there, but using 3.7 to create the venv if we plan to use 3,7 within the venv seems safer for me, so I opted for that right now.

The current script works. I removed some comments that explained what the line below did, where the line itself was verbose enough.

PR #1579, opened 2021-03-29
Fixes Issue #1303. Allows to replace participants. (#1303)

### LINKED ISSUE #1303: Replace option for participant adding

FSadrieh:
Fixes #1303

### DIFF
--- a/evap/staff/templates/staff_evaluation_person_management.html
+++ b/evap/staff/templates/staff_evaluation_person_management.html
@@ -20,17 +20,17 @@
                             <button type="button" class="btn-link" onclick="copyHeaders(['Title', 'First name', 'Last name', 'Email'])">{% trans 'Copy headers to clipboard' %}</button>).
                             {% trans 'This will create all containing users.' %}
                         </p>
-            	        {% include 'bootstrap_form.html' with form=participant_excel_form wide=True %}                    
+            	        {% include 'bootstrap_form.html' with form=participant_excel_form wide=True %}
                     </div>
-                    <div class="card-footer text-center">                      
+                    <div class="card-footer text-center">
                         {% if not participant_test_passed %}
                             <button name="operation" value="test-participants" type="submit" class="btn btn-sm btn-light form-submit-btn">{% trans 'Upload and test' %}</button>
                         {% else %}
                             <button name="operation" value="test-participants" type="submit" class="btn btn-sm btn-light form-submit-btn">{% trans 'Upload and test' %}</button>
-                        <div class="mt-2">
-                            <button name="operation" value="import-participants" type="button" onclick="importParticipantsModalShow('import-participants');" class="btn btn-sm btn-primary mr-1 form-submit-btn";>{% trans 'Import previously uploaded file' %}</button>
-                            <button name="operation" value="import-replace-participants" type="button" onclick="replaceParticipantsModalShow('import-replace-participants');" class="btn btn-sm btn-danger form-submit-btn">{% trans 'Replace all' %}</button>
-                        </div>
+                            <div class="mt-2">
+                                <button name="operation" value="import-participants" type="button" onclick="importParticipantsModalShow('import-participants');" class="btn btn-sm btn-primary mr-1 form-submit-btn";>{% trans 'Import previously uploaded file' %}</button>
+                                <button name="operation" value="import-replace-participants" type="button" onclick="replaceParticipantsModalShow('import-replace-participants');" class="btn btn-sm btn-danger form-submit-btn">{% trans 'Replace all' %}</button>
+                            </div>
                         {% endif %}
                     </div>
                 </div>
@@ -76,10 +76,10 @@
                             <button name="operation" value="test-contributors" type="submit" class="btn btn-sm btn-light form-submit-btn">{% trans 'Upload and test' %}</button>
                         {% else %}
                             <button name="operation" value="test-contributors" type="submit" class="btn btn-sm btn-light form-submit-btn">{% trans 'Upload and test' %}</button>
-                        <div class="mt-2">
-                            <button name="operation" value="import-contributors" type="button" onclick="importContributorsModalShow('import-contributors');" class="btn btn-sm btn-primary form-submit-btn">{% trans 'Import previously uploaded file' %}</button>
-                            <button name="operation" value="import-replace-contributors" type="button" onclick="replaceContributorsModalShow('import-replace-contributors');" class="btn btn-sm btn-danger ml-1 form-submit-btn">{% trans 'Replace all' %}</button>
-                        </div>
+                            <div class="mt-2">
+                                <button name="operation" value="import-contributors" type="button" onclick="importContributorsModalShow('import-contributors');" class="btn btn-sm btn-primary form-submit-btn">{% trans 'Import previously uploaded file' %}</button>
+                                <button name="operation" value="import-replace-contributors" type="button" onclick="replaceContributorsModalShow('import-replace-contributors');" class="btn btn-sm btn-danger ml-1 form-submit-btn">{% trans 'Replace all' %}</button>
+                            </div>
                         {% endif %}
                     </div>
                 </div>
@@ -142,19 +142,19 @@
         };
     </script>
 
-{% trans 'Replace participants' as title %}
-{% blocktrans asvar question %}Do you really want to delete all existing participants and replace them with participant data from the Excel file?{% endblocktrans %}
-{% trans 'Replace participants' as action_text %}
-{% include 'confirmation_modal.html' with modal_id='replaceParticipantsModal' title=title question=question action_text=action_text btn_type='danger' %}
-<script type="text/javascript">
-    function replaceParticipantsModalAction(dataId) {
-        var input = $("<input>")
-           .attr("type", "hidden")
-           .attr("name", "operation").val(dataId);
-        $('#participant-import-form').append(input);
-        $('#participant-import-form').submit();
-    };
-</script>
+    {% trans 'Replace participants' as title %}
+    {% blocktrans asvar question %}Do you really want to delete all existing participants and replace them with participant data from the Excel file?{% endblocktrans %}
+    {% trans 'Replace participants' as action_text %}
+    {% include 'confirmation_modal.html' with modal_id='replaceParticipantsModal' title=title question=question action_text=action_text btn_type='danger' %}
+    <script type="text/javascript">
+        function replaceParticipantsModalAction(dataId) {
+            var input = $("<input>")
+            .attr("type", "hidden")
+            .attr("name", "operation").val(dataId);
+            $('#participant-import-form').append(input);
+            $('#participant-import-form').submit();
+        };
+    </script>
 
     {% trans 'Copy participants' as title %}
     {% blocktrans asvar question %}Do you really want to copy the participants?{% endblocktrans %}
@@ -166,7 +166,7 @@
             .attr("type", "hidden")
             .attr("name", "operation").val(dataId);
             $('#participant-copy-form').append(input);
-            $('#participant-copy-form').submit();  
+            $('#participant-copy-form').submit();
         };
     </script>
 
@@ -194,7 +194,7 @@
             .attr("type", "hidden")
             .attr("name", "operation").val(dataId);
             $('#contributor-import-form').append(input);
-            $('#contributor-import-form').submit(); 
+            $('#contributor-import-form').submit();
         };
     </script>
 
@@ -207,8 +207,8 @@
             var input = $("<input>")
             .attr("type", "hidden")
             .attr("name", "operation").val(dataId);
-            $('#participant-import-form').append(input);
-            $('#participant-import-form').submit();
+            $('#contributor-import-form').append(input);
+            $('#contributor-import-form').submit();
         };
     </script>
 
@@ -222,7 +222,7 @@
             .attr("type", "hidden")
             .attr("name", "operation").val(dataId);
             $('#contributor-copy-form').append(input);
-            $('#contributor-copy-form').submit(); 
+            $('#contributor-copy-form').submit();
         };
     </script>
 
--- a/evap/staff/templates/staff_evaluation_person_management.html
+++ b/evap/staff/templates/staff_evaluation_person_management.html
@@ -20,7 +20,7 @@
                             <button type="button" class="btn-link" onclick="copyHeaders(['Title', 'First name', 'Last name', 'Email'])">{% trans 'Copy headers to clipboard' %}</button>).
                             {% trans 'This will create all containing users.' %}
                         </p>
-            	        {% include 'bootstrap_form.html' with form=participant_excel_form wide=True %}
+                    {% include 'bootstrap_form.html' with form=participant_excel_form wide=True %}
                     </div>
                     <div class="card-footer text-center">
                         {% if not participant_test_passed %}
@@ -48,7 +48,7 @@
                     </div>
                     <div class="card-footer text-center">
                         <button name="operation" value="copy-participants" type="button" onclick="copyParticipantsModalShow('copy-participants');" class="btn btn-sm btn-primary form-submit-btn">{% trans 'Copy participants' %}</button>
-                        <button name="operation" value="copy-replace-participants" type="button" onclick="copyreplaceParticipantsModalShow('copy-replace-participants');" class="btn btn-sm btn-danger ml-1 form-submit-btn">{% trans 'Replace all' %}</button>
+                        <button name="operation" value="copy-replace-participants" type="button" onclick="copyReplaceParticipantsModalShow('copy-replace-participants');" class="btn btn-sm btn-danger ml-1 form-submit-btn">{% trans 'Replace all' %}</button>
                     </div>
                 </div>
             </form>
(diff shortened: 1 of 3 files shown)


evap/staff/templates/staff_evaluation_person_management.html
```diff
@@ -20,15 +20,17 @@ <h6 class="card-subtitle mb-2 text-muted">{% trans 'From Excel file' %}</h6>
                             <button type="button" class="btn-link" onclick="copyHeaders(['Title', 'First name', 'Last name', 'Email'])">{% trans 'Copy headers to clipboard' %}</button>).
                             {% trans 'This will create all containing users.' %}
                         </p>
-                        {% include 'bootstrap_form.html' with form=participant_excel_form wide=True %}
+            	        {% include 'bootstrap_form.html' with form=participant_excel_form wide=True %}                    
                     </div>
-                    <div class="card-footer text-center">
+                    <div class="card-footer text-center">
```
janno42, line 25: please pay attention to the leading and trailing whitespace changes here and above

evap/staff/templates/staff_evaluation_person_management.html
```diff
@@ -20,15 +20,17 @@ <h6 class="card-subtitle mb-2 text-muted">{% trans 'From Excel file' %}</h6>
...
                         {% else %}
                             <button name="operation" value="test-participants" type="submit" class="btn btn-sm btn-light form-submit-btn">{% trans 'Upload and test' %}</button>
-                            <div class="form-submit-btn-divider"></div>
-                            <button name="operation" value="import-participants" type="button" onclick="importParticipantsModalShow('import-participants');" class="btn btn-sm btn-primary form-submit-btn">{% trans 'Import previously uploaded file' %}</button>
+                        <div class="mt-2">
+                            <button name="operation" value="import-participants" type="button" onclick="importParticipantsModalShow('import-participants');" class="btn btn-sm btn-primary mr-1 form-submit-btn";>{% trans 'Import previously uploaded file' %}</button>
+                            <button name="operation" value="import-replace-participants" type="button" onclick="replaceParticipantsModalShow('import-replace-participants');" class="btn btn-sm btn-danger form-submit-btn">{% trans 'Replace all' %}</button>
+                        </div>
```
janno42, line 33: the lines of this `div` should be indented one step further

evap/staff/templates/staff_evaluation_person_management.html
```diff
@@ -73,8 +76,10 @@ <h6 class="card-subtitle mb-2 text-muted">{% trans 'From Excel file' %}</h6>
                             <button name="operation" value="test-contributors" type="submit" class="btn btn-sm btn-light form-submit-btn">{% trans 'Upload and test' %}</button>
                         {% else %}
                             <button name="operation" value="test-contributors" type="submit" class="btn btn-sm btn-light form-submit-btn">{% trans 'Upload and test' %}</button>
-                            <div class="form-submit-btn-divider"></div>
+                        <div class="mt-2">
                             <button name="operation" value="import-contributors" type="button" onclick="importContributorsModalShow('import-contributors');" class="btn btn-sm btn-primary form-submit-btn">{% trans 'Import previously uploaded file' %}</button>
+                            <button name="operation" value="import-replace-contributors" type="button" onclick="replaceContributorsModalShow('import-replace-contributors');" class="btn btn-sm btn-danger ml-1 form-submit-btn">{% trans 'Replace all' %}</button>
+                        </div>
```
janno42, line 82: please check the indentation :)

evap/staff/templates/staff_evaluation_person_management.html
```diff
@@ -136,6 +142,20 @@ <h6 class="card-subtitle mb-2 text-muted">{% trans 'To CSV file' %}</h6>
...
+<script type="text/javascript">
+    function replaceParticipantsModalAction(dataId) {
+        var input = $("<input>")
+           .attr("type", "hidden")
+           .attr("name", "operation").val(dataId);
+        $('#participant-import-form').append(input);
+        $('#participant-import-form').submit();
+    };
+</script>
+
```
janno42, line 158: and here again, please

evap/staff/tests/test_views.py
```diff
@@ -1733,6 +1752,17 @@ def test_copy_participants(self):
 
         self.assertEqual(self.evaluation.participants.count(), original_participant_count + self.evaluation2.participants.count())
 
+    def test_copy_participants(self):
+        page = self.app.get(self.url, user=self.manager)
+
+        original_participant_count = self.evaluation.participants.count()
```
janno42, line 1758: you are not using this value again. maybe make sure that is is 0?
that would help to understand the `assertEqual` below. right now it is weird because one could think that copying should not make the number equal in contrast to the replacement operation. it is only true here, because `self.evaluation` didn't have any participants before.

evap/staff/tests/test_views.py
```diff
@@ -1764,6 +1813,18 @@ def test_copy_contributors(self):
         new_contributor_count = UserProfile.objects.filter(contributions__evaluation=self.evaluation).count()
         self.assertEqual(new_contributor_count, original_contributor_count + UserProfile.objects.filter(contributions__evaluation=self.evaluation2).count())
 
+    def test_copy_replace_contributors(self):
```
janno42, line 1816: shouldn't this be `test_copy_contributors`?
then, the same situation as in `test_copy_participants` applies here as well.

evap/staff/views.py
```diff
@@ -996,14 +1010,15 @@ def evaluation_person_management(request, semester_id, evaluation_id):
...
-                             'test-contributors', 'import-contributors', 'copy-contributors'):
+        if operation not in ('test-participants', 'import-participants', 'copy-participants', 'import-replace-participants', 'copy-replace-participants',
+                             'test-contributors', 'import-contributors', 'copy-contributors', 'import-replace-contributors', 'copy-replace-contributors'):
             raise SuspiciousOperation("Invalid POST operation")
 
         import_type = ImportType.Participant if 'participants' in operation else ImportType.Contributor
         excel_form = participant_excel_form if 'participants' in operation else contributor_excel_form
         copy_form = participant_copy_form if 'participants' in operation else contributor_copy_form
 
+
```
janno42, line 1021: please remove this line

janno42 (changes requested):
works nicely! there are some formatting and test things left to do.

evap/staff/templates/staff_evaluation_person_management.html
FSadrieh: I think this is now indented correctly

evap/staff/tests/test_views.py
FSadrieh: I dont thinl so, the first test is for just copying the contributors into the evaluation. The second test is for replacing the current contributors.

evap/staff/templates/staff_evaluation_person_management.html
```diff
@@ -20,15 +20,17 @@ <h6 class="card-subtitle mb-2 text-muted">{% trans 'From Excel file' %}</h6>
                             <button type="button" class="btn-link" onclick="copyHeaders(['Title', 'First name', 'Last name', 'Email'])">{% trans 'Copy headers to clipboard' %}</button>).
                             {% trans 'This will create all containing users.' %}
                         </p>
-                        {% include 'bootstrap_form.html' with form=participant_excel_form wide=True %}
+            	        {% include 'bootstrap_form.html' with form=participant_excel_form wide=True %}
```
He3lixxx, line 23: accidentally changed indentation(?)

evap/staff/templates/staff_evaluation_person_management.html
```diff
@@ -46,6 +48,7 @@ <h6 class="card-subtitle mb-2 text-muted">{% trans 'From other evaluation' %}</h
                     </div>
                     <div class="card-footer text-center">
                         <button name="operation" value="copy-participants" type="button" onclick="copyParticipantsModalShow('copy-participants');" class="btn btn-sm btn-primary form-submit-btn">{% trans 'Copy participants' %}</button>
+                        <button name="operation" value="copy-replace-participants" type="button" onclick="copyreplaceParticipantsModalShow('copy-replace-participants');" class="btn btn-sm btn-danger ml-1 form-submit-btn">{% trans 'Replace all' %}</button>
```
He3lixxx, line 51: Consistent naming would be `copyReplaceParticipantsModalShow`, right?

evap/staff/templates/staff_evaluation_person_management.html
```diff
@@ -136,6 +142,20 @@ <h6 class="card-subtitle mb-2 text-muted">{% trans 'To CSV file' %}</h6>
...
 
+    {% trans 'Replace participants' as title %}
+    {% blocktrans asvar question %}Do you really want to delete all existing participants and replace them with participant data from the Excel file?{% endblocktrans %}
+    {% trans 'Replace participants' as action_text %}
+    {% include 'confirmation_modal.html' with modal_id='replaceParticipantsModal' title=title question=question action_text=action_text btn_type='danger' %}
+    <script type="text/javascript">
+        function replaceParticipantsModalAction(dataId) {
+            var input = $("<input>")
+            .attr("type", "hidden")
+            .attr("name", "operation").val(dataId);
```
He3lixxx, line 153: personally, I'd indent these two lines, but I can also live with how it is now.

evap/staff/tests/test_views.py
```diff
@@ -1733,6 +1754,15 @@ def test_copy_participants(self):
 
         self.assertEqual(self.evaluation.participants.count(), original_participant_count + self.evaluation2.participants.count())
 
+    def test_replace_copy_participants(self):
+        page = self.app.get(self.url, user=self.manager)
+
+        form = page.forms["participant-copy-form"]
+        form["evaluation"] = str(self.evaluation2.pk)
+        page = form.submit(name="operation", value="copy-participants")
```
He3lixxx, line 1762: test is named `replace_copy`, but here you only `copy`. Was this intended?

evap/staff/tests/test_views.py
```diff
@@ -1722,6 +1724,25 @@ def test_import_valid_participants_file(self):
...
+        page = self.app.get(self.url2, user=self.manager)
+
+        original_participant_count = self.evaluation2.participants.count()
+
+        form = page.forms["participant-import-form"]
+        form["excel_file"] = (self.filename_valid,)
+        page = form.submit(name="operation", value="test-participants")
+
+        self.assertContains(page, 'Import previously uploaded file')
+        self.assertEqual(self.evaluation2.participants.count(), original_participant_count)
```
He3lixxx, line 1737: In theory, we should split up testing the test run, which is the same for import and replace, and then testing the actual import and replace runs (so these assertions here are not duplicated) (same for the contributors tests)

evap/staff/tests/test_views.py
```diff
@@ -1722,6 +1724,25 @@ def test_import_valid_participants_file(self):
...
+
+        form = page.forms["participant-import-form"]
+        form.submit(name="operation", value="import-replace-participants")
+        self.assertEqual(self.evaluation2.participants.count(), 2)
```
He3lixxx, line 1741: This assertion only works because the test setup currently ensures that they the participant count was unequal to two before. However, in the future, anyone might change the test setup. To prevent this from making the test meaningless, I like to add an assertion that the old state was differend. In this case, that would be:
```
self.assertNotEqual(self.evaluation2.participants.count(), 2)
```
before the submission of the form.

(same for the other tests)

evap/staff/tests/test_views.py
```diff
@@ -1733,6 +1754,15 @@ def test_copy_participants(self):
...
         self.assertEqual(self.evaluation.participants.count(), original_participant_count + self.evaluation2.participants.count())
 
+    def test_replace_copy_participants(self):
+        page = self.app.get(self.url, user=self.manager)
+
+        form = page.forms["participant-copy-form"]
+        form["evaluation"] = str(self.evaluation2.pk)
+        page = form.submit(name="operation", value="copy-participants")
+
+        self.assertEqual(self.evaluation.participants.count(), self.evaluation2.participants.count())
```
He3lixxx, line 1764: If this test did the wrong thing and still passed, we should think about how to prevent this in the future.

evap/staff/views.py
```diff
@@ -1014,21 +1027,28 @@ def evaluation_person_management(request, semester_id, evaluation_id):
...
                     save_import_file(excel_file, request.user.id, import_type)
 
-        elif 'import' in operation:
-            file_content = get_import_file_content_or_raise(request.user.id, import_type)
-            success_messages, warnings, __ = PersonImporter.process_file_content(import_type, evaluation, test_run=False, file_content=file_content)
-            delete_import_file(request.user.id, import_type)
+        else:
+            if 'import' in operation:
+                if 'replace' in operation:
+                    deleted_person_count, deletion_message = helper_delete_users_from_evaluation(evaluation, operation)
```
He3lixxx, line 1033: This is duplicated. You could move it out one scope, so that the logic is like:
```python3
if "test:
    pass # do test stuff
else:
    additional_messages = []
    if "replace":
        ... = helper_delete_users_from_evaluation(...)  # before the actual operation: clear everything
        # change additional messages
      
    if "copy":
        pass # do copy stuff
    elif "import":
        pass # do import stuff
       
    # merge messages and additional messages, unconditionally
```
Then, we don't have to check for "replace" three times.

Also, I'd suggest moving everything inside a [database transaction](https://docs.djangoproject.com/en/3.2/topics/db/transactions/), to make sure that either
- existing users get removed and the new users get added automatically, or,
- nothing is changed

It would be weird if, for whatever reason, you click on "replace" and end up with an empty data set because something went wrong.

He3lixxx (changes requested):
Cool stuff!

evap/staff/templates/staff_evaluation_person_management.html
FSadrieh: You just picked them as an example, but would intend them at every button?

He3lixxx: Uh, good point, I'd make an extra commit that "fixes" the indentation for this and the other 6 spots where it is not indented.

evap/staff/views.py
```diff
@@ -1014,21 +1028,28 @@ def evaluation_person_management(request, semester_id, evaluation_id):
...
 
-        elif 'import' in operation:
-            file_content = get_import_file_content_or_raise(request.user.id, import_type)
-            success_messages, warnings, __ = PersonImporter.process_file_content(import_type, evaluation, test_run=False, file_content=file_content)
-            delete_import_file(request.user.id, import_type)
+        else:
+            additional_messages = []
+            if 'replace' in operation:
+                deleted_person_count, deletion_message = helper_delete_users_from_evaluation(evaluation, operation)
+                additional_messages = format_html(deletion_message, deleted_person_count, evaluation.name)
```
janno42, line 1035: please use `evaluation.full_name` instead of `evaluation.name`. the `name` can be empty, `full_name` would then use the course's name instead.

evap/staff/templates/staff_evaluation_person_management.html
```diff
@@ -136,17 +142,45 @@ <h6 class="card-subtitle mb-2 text-muted">{% trans 'To CSV file' %}</h6>
...
+                .attr("name", "operation").val(dataId);
+            $('#participant-copy-form').append(input);
+            $('#participant-copy-form').submit();
+        };
+    </script>
+
+    {% trans 'Replace participants' as title %}
+    {% blocktrans asvar question %}Do you really want to delete all existing participants and copy the participants into the evaluation?{% endblocktrans %}
+    {% trans 'Replace participants' as action_text %}
+    {% include 'confirmation_modal.html' with modal_id='copyreplaceParticipantsModal' title=title question=question action_text=action_text btn_type='danger' %}
```
janno42, line 176: you missed the uppercase letter here

janno42 (changes requested):
almost there! 🙂

evap/staff/templates/staff_evaluation_person_management.html
```diff
@@ -78,7 +78,7 @@ <h6 class="card-subtitle mb-2 text-muted">{% trans 'From Excel file' %}</h6>
                             <button name="operation" value="test-contributors" type="submit" class="btn btn-sm btn-light form-submit-btn">{% trans 'Upload and test' %}</button>
                             <div class="mt-2">
                                 <button name="operation" value="import-contributors" type="button" onclick="importContributorsModalShow('import-contributors');" class="btn btn-sm btn-primary form-submit-btn">{% trans 'Import previously uploaded file' %}</button>
-                                <button name="operation" value="import-replace-contributors" type="button" onclick="replaceContributorsModalShow('import-replace-contributors');" class="btn btn-sm btn-danger ml-1 form-submit-btn">{% trans 'Replace all' %}</button>
+                                <button name="operation" value="import-replace-contributors" type="button" onclick="replaceContributorsModalShow('import-replace-contributors');" class="btn btn-sm btn-danger ml-1 form-submit-btn">{% trans 'Replace contribors' %}</button>
```
janno42, line 81: `contributors` :)

evap/staff/templates/staff_evaluation_person_management.html
```diff
@@ -231,7 +231,7 @@ <h6 class="card-subtitle mb-2 text-muted">{% trans 'To CSV file' %}</h6>
     {% trans 'Replace contributors' as action_text %}
     {% include 'confirmation_modal.html' with modal_id='copyreplaceContributorsModal' title=title question=question action_text=action_text btn_type='danger' %}
```
janno42, line 232: another `copyReplace` here please

janno42 (changes requested):

He3lixxx (approved):

janno42 (approved):

PR #1580, opened 2021-04-12
vagrant thingies

niklasmohrin:
- Move execution of install_dart_sass.sh below the mounting of the folder containing the script
- Remove unneeded uses of sudo in provision_vagrant_vm.sh
- Only use de_DE and en_US locales with `manage.py compilemessages`


He3lixxx (approved):
Thanks :)

PR #1581, opened 2021-04-12
Pylint all the things. Even more than before!

He3lixxx:
Edit: Probably best reviewed per commit.

Will we ever get done with #1092? No one knows!

However, I have to say, most code that pylint complained about here in the update is actually worth changing.

We have 3 or 4 places where pylint detects boilerplate code as duplicate, however, currently, we can not exclude that part from the detection, so we still need to keep the duplicate code detection disabled on a global level. I fixed the other places (where I thought it was appropriate) -- I'm open for discussing these changes, we can also revert / not merge them.

### DIFF
--- a/evap/student/views.py
+++ b/evap/student/views.py
@@ -17,9 +17,8 @@ from evap.student.models import TextAnswerWarning
 from evap.student.forms import QuestionnaireVotingForm
 from evap.student.tools import question_id
 
-from evap.results.tools import (calculate_average_distribution, distribution_to_grade,
-                                get_evaluations_with_course_result_attributes, get_single_result_rating_result,
-                                textanswers_visible_to, normalized_distribution)
+from evap.results.tools import (annotate_distributions_and_grades, get_evaluations_with_course_result_attributes,
+                                textanswers_visible_to)
 
 SUCCESS_MAGIC_STRING = 'vote submitted successfully'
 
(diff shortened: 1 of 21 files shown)


evap/student/views.py
```diff
@@ -54,12 +53,7 @@ def index(request):
 
     for evaluation in evaluations:
         if evaluation.state == "published":
-            if not evaluation.is_single_result:
-                evaluation.distribution = calculate_average_distribution(evaluation)
-            else:
-                evaluation.single_result_rating_result = get_single_result_rating_result(evaluation)
-                evaluation.distribution = normalized_distribution(evaluation.single_result_rating_result.counts)
-            evaluation.avg_grade = distribution_to_grade(evaluation.distribution)
+            annotate_distribution_and_grade(evaluation)
```
karyon, line 56: you could make the annotate function take a list. you'd have to filter here though for published evaluations...

karyon (approved):

karyon:
the duplicate detection is pretty neat.

evap/student/views.py
He3lixxx: Hmm. I considered it, but we then have the following code:
```python
evaluations = [some list of evaluations]  # computed somehow

published_evaluations = [e for e in evaluations if e.state == published]
annotate_distribution_and_grade(published_evaluations)

# code goes on to use evaluations instead of published_evaluations
```

This is weird because when I read this, I'd first think that this is an unused computation or unused variable. The real intention is hidden behind a somewhat obscure side effect, which I don't like. What do you think?

karyon: hm, true. you could do `annotate_distribution_and_grade(e for e in evaluations if e.state == published)`, maybe that is less weird. but maybe it's not :D you can of course leave it as it is if you want :)

He3lixxx: Implemented. I was a bit unsure about the name, now we have `annotate_evaluation_*` directly next to `get_evaluations_with_*` where both do more or less the same. We have quite a few other occurences of `annotate_evaluation_*` but also other occurences of `get_evaluations_with_*`

It would be cool to have this refactored, and I started doing that, but it turns out that there is a bit more work to do and some discussions to have, so I'd just put it in a new issue.

PR #1582, opened 2021-04-12
Detect installed CPU cores and system memory and initialize VM accordingly

niklasmohrin:
I think I will remove the part where I open files and make assumptions with another awk oneliner from the Internet. Originally, I wanted to not rely on having awk installed, but these are just so much simpler and since this branch doesn't run on Windows hosts, installing awk should be painful when in doubt.


He3lixxx:
I think it is somewhat unmaintainable (mainly because its not our main language), but I mean ... if it works ... guess we can throw it away any time if it stops working.

PR #1583, opened 2021-04-12
Fix #1571: Copying Courses

### LINKED ISSUE #1571: Allow copying evaluation to other semester
When copying an evaluation, the copy will currently be created in the same semester as the original.
This should be changed so that the semester can be selected in a field at the top of the form, defaulting to the current semester.

felixrindt:
Fixes #1571, although that requested Evaluation copy to other semesters. Problem though, we are missing the Course, which is why @janno42 suggested to create Course Copy instead, on which I had a go here.

Somewhat complicated is copying the evaluations.

### DIFF
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -216,12 +216,11 @@ class CourseTypeMergeSelectionForm(forms.Form):
 
 
 class CourseFormMixin:
-    def _set_responsibles_queryset(self):
-        self.fields["responsibles"].queryset = UserProfile.objects.exclude(is_active=False)
-        if self.instance.pk:
-            self.fields["responsibles"].queryset |= UserProfile.objects.filter(
-                pk__in=[user.pk for user in self.instance.responsibles.all()]
-            )
+    def _set_responsibles_queryset(self, existing_course=None):
+        queryset = UserProfile.objects.exclude(is_active=False)
+        if existing_course:
+            queryset = (queryset | existing_course.responsibles.all()).distinct()
+        self.fields["responsibles"].queryset = queryset
 
     def validate_unique(self):
         super().validate_unique()
@@ -256,7 +255,7 @@ class CourseForm(CourseFormMixin, forms.ModelForm):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self._set_responsibles_queryset()
+        self._set_responsibles_queryset(self.instance if self.instance.pk else None)
         if not self.instance.can_be_edited_by_manager:
             disable_all_fields(self)
 
@@ -276,7 +275,7 @@ class CourseCopyForm(CourseFormMixin, forms.ModelForm):
         opts = self._meta
         initial = forms.models.model_to_dict(instance, opts.fields, opts.exclude)
         super().__init__(data=data, initial=initial)
-        self._set_responsibles_queryset()
+        self._set_responsibles_queryset(instance)
 
     # To ensure we don't forget about copying a relevant field, we explicitly list copied and ignored fields and test against that
     EVALUATION_COPIED_FIELDS = {
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -342,7 +342,6 @@ class CourseCopyForm(CourseFormMixin, forms.ModelForm):
                 new_contribution.save()
                 new_contribution.questionnaires.set(old_contribution.questionnaires.all())
 
-        new_course.responsibles.filter(is_active=False).update(is_active=True)
         return new_course
 
 
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -4,6 +4,7 @@ from collections.abc import Iterable
 from django import forms
 from django.contrib.auth.models import Group
 from django.core.exceptions import SuspiciousOperation, ValidationError
+from django.db import transaction
 from django.db.models import Max, Q
 from django.forms.models import BaseInlineFormSet
 from django.forms.widgets import CheckboxSelectMultiple
@@ -147,8 +148,8 @@ class SemesterForm(forms.ModelForm):
         model = Semester
         fields = ("name_de", "name_en", "short_name_de", "short_name_en")
 
-    def save(self, *args, **kwargs):
-        semester = super().save(*args, **kwargs)
+    def save(self, commit=True):
+        semester = super().save(commit)
         if "short_name_en" in self.changed_data or "short_name_de" in self.changed_data:
             update_template_cache(semester.evaluations.filter(state__in=STATES_WITH_RESULT_TEMPLATE_CACHING))
         return semester
@@ -214,7 +215,28 @@ class CourseTypeMergeSelectionForm(forms.Form):
             raise ValidationError(_("You must select two different course types."))
 
 
-class CourseForm(forms.ModelForm):
+class CourseFormMixin:
+    def _set_responsibles_queryset(self):
+        self.fields["responsibles"].queryset = UserProfile.objects.exclude(is_active=False)
+        if self.instance.pk:
+            self.fields["responsibles"].queryset |= UserProfile.objects.filter(
+                pk__in=[user.pk for user in self.instance.responsibles.all()]
+            )
+
+    def validate_unique(self):
+        super().validate_unique()
+        # name_xy and semester are unique together. This will be treated as a non-field-error since two
+        # fields are involved. Since we only show the name_xy field to the user, assign that error to this
+        # field. This hack is not documented, so it might be broken when you are reading this.
+        for e in self.non_field_errors().as_data():
+            if e.code == "unique_together" and "unique_check" in e.params:
+                if "semester" in e.params["unique_check"]:
+                    # The order of the fields is probably determined by the unique_together constraints in the Course class.
+                    name_field = e.params["unique_check"][1]
+                    self.add_error(name_field, e)
+
+
+class CourseForm(CourseFormMixin, forms.ModelForm):
     semester = forms.ModelChoiceField(Semester.objects.all(), disabled=True, required=False, widget=forms.HiddenInput())
 
     class Meta:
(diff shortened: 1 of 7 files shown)


felixrindt:
I can't get one of my tests to pass with

```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "evaluation_course_temp_pkey"
DETAIL:  Key (id)=(1) already exists.
```

and I can't reproduce the problem in a running dev instance.

janno42:
all tests succeed on my machine. can you tell me which test fails on which line and which django version you are using? (have you installed the latest requirements?)

felixrindt:
Weird, I did not get this error today when running the `./manage.py test`.

evap/staff/forms.py
```diff
@@ -233,27 +255,70 @@ class Meta:
...
+        opts = self._meta
+        initial = forms.models.model_to_dict(instance, opts.fields, opts.exclude)
+        super().__init__(data=data, initial=initial)
+        self._set_responsibles_queryset()
+
+    @transaction.atomic()
+    def save(self, commit=True):
+        new_course: Course = super().save()
+        # we need to create copies of evaluations and their participation as well
+        for old_evaluation in self.old_course.evaluations.all():
```
janno42, line 282: i think we can skip all single results. copying those doesn't really make sense.

evap/staff/forms.py
```diff
@@ -233,27 +255,70 @@ class Meta:
...
-        # field. This hack is not documented, so it might be broken when you are reading this.
-        for e in self.non_field_errors().as_data():
-            if e.code == "unique_together" and "unique_check" in e.params:
-                if "semester" in e.params["unique_check"]:
-                    # The order of the fields is probably determined by the unique_together constraints in the Course class.
-                    name_field = e.params["unique_check"][1]
-                    self.add_error(name_field, e)
+
+class CourseCopyForm(CourseFormMixin, forms.ModelForm):
+    semester = forms.ModelChoiceField(Semester.objects.all())
```
janno42, line 264: can the semester field easily be moved to the top of the form?

janno42 (changes requested):

evap/staff/tests/test_views.py
```diff
@@ -1642,6 +1642,57 @@ def test_evaluation_copy(self):
...
+    def test_course_copy(self):
+        response = self.app.get(self.url, user=self.manager, status=200)
+        form = response.forms["course-form"]
+        form["semester"] = self.other_semester.pk
+        form["vote_start_datetime"] = datetime.datetime(2099, 1, 1, 0, 0)
+        form["vote_end_date"] = datetime.date(2099, 12, 31)
+        response = form.submit()
+
+        self.assertEqual(Course.objects.count(), 2)
+        copied_course = Course.objects.exclude(pk=self.course.pk).get()
```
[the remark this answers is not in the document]
felixrindt: that's not the same. I want to get *the other* evaluation. Though I can do `get(pk__ne=...)`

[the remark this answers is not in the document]
felixrindt: That's not the same :)

evap/staff/templates/staff_course_copyform.html
```diff
@@ -0,0 +1,44 @@
...
+        </div>
+
+        <div class="card mb-3">
+            <div class="card-body">
+                <h5 class="card-title">{% trans 'Evaluations that will be copied' %}</h5>
+                {% if course.evaluations.count > 0 %}
+                    <ul>
+                        {% for evaluation in course.evaluations.all %}
+                            <li>
+                                {{ evaluation.full_name }}
```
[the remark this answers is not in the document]
felixrindt: I don't think it matters. Do you know how to do that in a template?

evap/staff/forms.py
felixrindt: Aye

evap/staff/forms.py
felixrindt: Yes! simply putting `field_order = ['semester']` make django put it at the top. All other fields follow with their default ordering.

evap/staff/templates/staff_course_copyform.html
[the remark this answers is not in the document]
janno42: it would be nice to bring them into an order because we also do it in many other places :)
have a look at, e. g., `contributor/templates/contributor_index.html` and the `dictsort` filter.

evap/staff/templates/staff_course_copyform.html
```diff
@@ -0,0 +1,44 @@
...
+                {% include 'bootstrap_form.html' with form=course_form %}
+            </div>
+        </div>
+
+        <div class="card mb-3">
+            <div class="card-body">
+                <h5 class="card-title">{% trans 'Evaluations that will be copied' %}</h5>
+                {% if course.evaluations.count > 0 %}
+                    <ul>
+                        {% for evaluation in course.evaluations.all %}
```
janno42, line 22: as single results are not copied, the list generated here is not correct

evap/staff/forms.py
```diff
@@ -233,27 +255,94 @@ class Meta:
...
+        "is_rewarded",
+        "is_midterm_evaluation",
+        "allow_editors_to_edit",
+        "wait_for_grade_upload_before_publishing",
+    }
+
+    EVALUATION_EXCLUDED_FIELDS = {
+        "id",
+        "course",
+        "can_publish_text_reuslts",
```
janno42, line 295:
```suggestion

```
i first wondered how the typo in here (`reuslts`) did not make the test fail but then saw the attribute is listed below again

evap/staff/forms.py
```diff
@@ -233,27 +255,94 @@ class Meta:
...
-            if e.code == "unique_together" and "unique_check" in e.params:
-                if "semester" in e.params["unique_check"]:
-                    # The order of the fields is probably determined by the unique_together constraints in the Course class.
-                    name_field = e.params["unique_check"][1]
-                    self.add_error(name_field, e)
+
+class CourseCopyForm(CourseFormMixin, forms.ModelForm):
+    semester = forms.ModelChoiceField(Semester.objects.all())
+    vote_start_datetime = forms.DateTimeField(label=_("Start of associated evaluations"), localize=True)
+    vote_end_date = forms.DateField(label=_("End of associated evaluations"), localize=True)
```
janno42, line 266:
```suggestion
    vote_start_datetime = forms.DateTimeField(label=_("Start of evaluations"), localize=True)
    vote_end_date = forms.DateField(label=_("Last day of evaluations"), localize=True)
```
Ahem, you marked this as resolved but it was the same as before 😱

evap/staff/views.py
```diff
@@ -946,6 +947,30 @@ def course_create(request, semester_id):
...
 
+@manager_required
+def course_copy(request, semester_id, course_id):
+    semester = get_object_or_404(Semester, id=semester_id)
+    course = get_object_or_404(Course, id=course_id, semester=semester)
+    course_form = CourseCopyForm(request.POST or None, instance=course)
+
+    if course_form.is_valid():
+        copied_course = course_form.save()
+        messages.success(request, _("Successfully created course."))
```
janno42, line 958:
```suggestion
        messages.success(request, _("Successfully copied course."))
```

janno42:
oh, and please remove the merge commit and rebase instead :)

evap/staff/templates/staff_course_copyform.html
```diff
@@ -0,0 +1,44 @@
...
+                {% if course.evaluations.count > 0 %}
+                    <ul>
+                        {% for evaluation in evaluations %}
+                            <li>
+                                {{ evaluation.full_name }}
+                            </li>
+                        {% endfor %}
+                    </ul>
+                {% else %}
+                    <span class="font-italic">{% trans 'There are no evaluations for this course.' %}</span>
```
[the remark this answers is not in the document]
felixrindt: That doesn't seem to work. I couldn't find other usages of `fst-italic`, only `font-italic`.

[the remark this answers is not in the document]
janno42: are you sure you updated the bootstrap submodule?
https://getbootstrap.com/docs/5.1/utilities/text/#font-weight-and-italics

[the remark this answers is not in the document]
janno42: but good to know we missed something in the bootstrap update 😅

evap/staff/forms.py
```diff
@@ -214,7 +215,28 @@ def clean(self):
             raise ValidationError(_("You must select two different course types."))
 
 
-class CourseForm(forms.ModelForm):
+class CourseFormMixin:
+    def _set_responsibles_queryset(self):
+        self.fields["responsibles"].queryset = UserProfile.objects.exclude(is_active=False)
+        if self.instance.pk:
+            self.fields["responsibles"].queryset |= UserProfile.objects.filter(
+                pk__in=[user.pk for user in self.instance.responsibles.all()]
```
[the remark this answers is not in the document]
janno42: focus, @felixrindt. focus!
"... and a respective info message should be shown" :)

evap/staff/views.py
```diff
@@ -950,6 +951,43 @@ def course_create(request, semester_id):
...
+def course_copy(request, semester_id, course_id):
+    semester = get_object_or_404(Semester, id=semester_id)
+    course = get_object_or_404(Course, id=course_id, semester=semester)
+    course_form = CourseCopyForm(request.POST or None, instance=course)
+
+    if course_form.is_valid():
+        copied_course = course_form.save()
+        messages.success(request, _("Successfully copied course."))
+
+        inactive_users = copied_course.responsibles.filter(is_active=False)
```
janno42, line 964: please also activate all other contributors

evap/staff/tests/test_views.py
```diff
@@ -1653,6 +1653,73 @@ def test_evaluation_copy(self):
...
 
 
+class TestCourseCopyView(WebTestStaffMode):
+    url = "/staff/semester/1/course/1/copy"
+
+    @classmethod
+    def setUpTestData(cls):
+        cls.manager = make_manager()
+        cls.semester = baker.make(Semester, pk=1)
+        cls.other_semester = baker.make(Semester, pk=2)
```
janno42, line 1663: please don't use explicit pks to avoid database integrity errors and create the url like in, e.g., `TestEvaluationCopyView`

janno42 (changes requested):

evap/staff/tests/test_views.py
He3lixxx: Interesting issue here
Run `./manage.py test evap.staff.tests.test_views.TestCourseCopyView`. You will get an IntegrityError on
```
form.submit().follow()
```
in line 1706. This is triggered by the CourseCopyForm doing
```
new_course: Course = super.save()
```
in staff/forms.py:327

In super.save(), it will do `self.instance.save()`, which fails because it violates the unique constraint on the course primary key. However, `self.instance` correctly does not have a pk set, so we're just trying to use the key the database gives us here.

Seems to me like the auto-counter of the db is at 1 and it doesn't check whether that key can actually be used.

He3lixxx: minimally reproducible in `./manage.py shell_plus`:
```python
>>> from model_bakery import baker
>>> s = baker.make(Semester)
>>> ct = baker.make(CourseType)
>>>
>>> c1 = baker.make(Course, semester=s, type=ct)
>>> c1.pk
144
>>> # So next ID would be 145. Mess it up with
>>> c2 = baker.make(Course, semester=s, type=ct, pk=145)
>>> c2.pk
145
>>> c3 = baker.prepare(Course, semester=s, type=ct)
>>> print(c3.pk)
None
>>> 
>>> c3.save()  # boom
```

karyon: sorry, i missed this. If a test uses only hard-coded non-conflicting IDs, all should be fine. A test never specifying pks should also be fine. hard-coded pks at all in a test should also be fine. Different tests shouldn't affect one another. I imagine that only mixing both methods within a test, like in your repro case, would break.

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

evap/staff/tests/test_forms.py
```diff
@@ -678,6 +679,21 @@ def test_handle_multivaluedicts(self):
         self.assertEqual(Questionnaire.objects.filter(contributions=self.contribution2).count(), 2)
 
 
+class CourseCopyFormTests(TestCase):
+    # pylint: disable=no-self-use
```
He3lixxx, line 683: We've used `@staticmethod` in other places -- any specific reason to not use it here?

evap/staff/tests/test_views.py
```diff
@@ -1653,6 +1653,70 @@ def test_evaluation_copy(self):
...
+        response = self.app.get(self.url, user=self.manager, status=200)
+        form = response.forms["course-form"]
+        form["semester"] = self.other_semester.pk
+        form["vote_start_datetime"] = datetime.datetime(2099, 1, 1, 0, 0)
+        form["vote_end_date"] = datetime.date(2099, 12, 31)
+
+        # check that the user activation is mentioned
+        assert not self.responsibles[1].is_active
+        response = form.submit().follow()
+        assert self.responsibles[1].full_name in str(response)
```
He3lixxx, line 1704: Was there a reason to not use `self.assertFalse` and `self.assertContains`?

He3lixxx: again in 1717

evap/staff/views.py
```diff
@@ -950,6 +951,45 @@ def course_create(request, semester_id):
...
+    course = get_object_or_404(Course, id=course_id, semester=semester)
+    course_form = CourseCopyForm(request.POST or None, instance=course)
+
+    if course_form.is_valid():
+        copied_course = course_form.save()
+        messages.success(request, _("Successfully copied course."))
+
+        inactive_users = UserProfile.objects.filter(
+            Q(contributions__evaluation__course__in=[copied_course], is_active=False)
+            | Q(courses_responsible_for__in=[copied_course], is_active=False)
```
He3lixxx, line 966: think we don't need to use `__in` if we use it with a list with a single element.

evap/staff/forms.py
```diff
@@ -234,27 +255,94 @@ class Meta:
...
-        if self.instance.pk:
-            self.fields["responsibles"].queryset |= UserProfile.objects.filter(
-                pk__in=[user.pk for user in self.instance.responsibles.all()]
+
+class CourseCopyForm(CourseFormMixin, forms.ModelForm):
+    semester = forms.ModelChoiceField(Semester.objects.all())
+    vote_start_datetime = forms.DateTimeField(label=_("Start of evaluations"), localize=True)
+    vote_end_date = forms.DateField(label=_("Last day of evaluations"), localize=True)
+
+    class Meta(CourseForm.Meta):
```
He3lixxx, line 268: If the same meta class is shared between `CourseForm` and `CourseCopyForm`, is there a way to put it in `CourseFormMixin` and inherit it from there?

He3lixxx (changes requested):

evap/staff/forms.py
janno42: i don't know :)
do you think that's important? (and if yes: do you want to have a look at it?)

He3lixxx (approved):

PR #1586, opened 2021-04-19
excludes evaluations that haven't terminated from fast review view

### LINKED ISSUE #1572: Filter evaluations on fast review
When finishing the review of an evaluation's text answers in fast mode, only evaluations where the evaluation period already ended should be selected for continuing the review. Currently, also evaluations that are still in evaluation are shown.

dieknolle3333:
fixes #1572

### DIFF
--- a/evap/staff/tests/test_views.py
+++ b/evap/staff/tests/test_views.py
@@ -1896,7 +1896,6 @@ class TestEvaluationEmailView(WebTestStaffMode):
         self.assertEqual(len(mail.outbox), 2)
 
 
-@override_settings(EVALUATION_END_OFFSET_HOURS=26)
 class TestEvaluationTextAnswerView(WebTest):
     url = '/staff/semester/1/evaluation/1/textanswers'
 
@@ -1910,7 +1909,7 @@ class TestEvaluationTextAnswerView(WebTest):
         cls.evaluation = baker.make(
             Evaluation,
             pk=1,
-            course=baker.make(Course, semester=semester),
+            course__semester=semester,
             participants=[student1, cls.student2],
             voters=[student1],
             state="in_evaluation"
@@ -1932,16 +1931,13 @@ class TestEvaluationTextAnswerView(WebTest):
 
         cls.evaluation2 = baker.make(
             Evaluation,
-            pk=2,
-            course=baker.make(Course, semester=semester),
+            course__semester=semester,
             participants=[student1],
             voters=[student1, cls.student2],
             vote_start_datetime=datetime.datetime.now() - datetime.timedelta(days=5),
             vote_end_date=datetime.date.today() - datetime.timedelta(days=4),
             can_publish_text_results=True
         )
-
-        cls.evaluation2.general_contribution.questionnaires.set([top_general_questionnaire])
 
         contribution2 = baker.make(
             Contribution,
@@ -1980,18 +1976,19 @@ class TestEvaluationTextAnswerView(WebTest):
             page = self.app.get(self.url + '?view=full', user=self.manager, status=200)
             self.assertContains(page, self.answer)
 
-    def test_exclude_unreviewed_evaluations(self):
-        # evaluation that ended 2 days ago is definitely finished and should be proposed
+    @override_settings(EVALUATION_END_OFFSET_HOURS=26)
+    def test_exclude_unfinished_evaluations(self):
         let_user_vote_for_evaluation(self.app, self.student2, self.evaluation)
         with run_in_staff_mode(self):
             page = self.app.get(self.url, user=self.manager, status=200)
+            # evaluation2 is finished and should show up
             self.assertContains(page, self.evaluation2.full_name)
 
-        # with an offset of 26 hours, evaluations with vote_end_date yesterday haven't finished yet
         self.evaluation2.vote_end_date = datetime.date.today() - datetime.timedelta(days=1)
         self.evaluation2.save()
         with run_in_staff_mode(self):
             page = self.app.get(self.url, user=self.manager, status=200)
+            # unfinished because still in EVALUATION_END_OFFSET_HOURS
             self.assertNotContains(page, self.evaluation2.full_name)
 
 
--- a/evap/staff/tests/test_views.py
+++ b/evap/staff/tests/test_views.py
@@ -1976,6 +1976,7 @@ class TestEvaluationTextAnswerView(WebTest):
             page = self.app.get(self.url + '?view=full', user=self.manager, status=200)
             self.assertContains(page, self.answer)
 
+    # keeps test safe from potential time zone changes and leap seconds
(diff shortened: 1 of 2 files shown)


evap/staff/tests/test_views.py
```diff
@@ -1896,6 +1896,7 @@ def test_emails_are_sent(self):
         self.assertEqual(len(mail.outbox), 2)
 
 
+@override_settings(EVALUATION_END_OFFSET_HOURS=26)
```
He3lixxx, line 1899: We only need this setting changed for the one test method, right? In this case, could you move the decorator there so it doesn't affect the whole class but only the one test method?

evap/staff/tests/test_views.py
```diff
@@ -1929,6 +1930,32 @@ def setUpTestData(cls):
...
+        cls.evaluation2 = baker.make(
+            Evaluation,
+            pk=2,
+            course=baker.make(Course, semester=semester),
+            participants=[student1],
+            voters=[student1, cls.student2],
+            vote_start_datetime=datetime.datetime.now() - datetime.timedelta(days=5),
+            vote_end_date=datetime.date.today() - datetime.timedelta(days=4),
+            can_publish_text_results=True
+        )
```
He3lixxx, line 1942: We don't specifically need the pk to be 2 here, right? If so, I'd remove it.

`course=...` could be simplified to `course__semester=semester` (here and above where this was taken from)

evap/staff/tests/test_views.py
```diff
@@ -1929,6 +1930,32 @@ def setUpTestData(cls):
...
+
+        cls.evaluation2.general_contribution.questionnaires.set([top_general_questionnaire])
```
He3lixxx, line 1944: Do we need this here? Do we need this above (with the LIKERT question)? I don't see why it would be necessary, but I'm probably just missing something.

evap/staff/tests/test_views.py
```diff
@@ -1953,6 +1980,20 @@ def test_textanswers_full_view(self):
             page = self.app.get(self.url + '?view=full', user=self.manager, status=200)
             self.assertContains(page, self.answer)
 
+    def test_exclude_unreviewed_evaluations(self):
```
He3lixxx, line 1983: exclude unreviewed evaluations? Shouldn't it be something like "exclude unfinished evaluations"?

evap/staff/tests/test_views.py
```diff
@@ -1953,6 +1980,20 @@ def test_textanswers_full_view(self):
...
+        # evaluation that ended 2 days ago is definitely finished and should be proposed
+        let_user_vote_for_evaluation(self.app, self.student2, self.evaluation)
+        with run_in_staff_mode(self):
+            page = self.app.get(self.url, user=self.manager, status=200)
+            self.assertContains(page, self.evaluation2.full_name)
```
He3lixxx, line 1988: I would shorten the comment a bit to not replicate content of the code. If someone changes the end date in `setUpTestData`, the "2" will become wrong (which I think it already is? Shouldn't it be 3?). So, I'd propose to make it something like

> evaluation2 is finished and should show up

or whatever. Also, I'd move the comment directly to the assertion doing the test

evap/staff/tests/test_views.py
```diff
@@ -1953,6 +1980,20 @@ def test_textanswers_full_view(self):
...
+
+        # with an offset of 26 hours, evaluations with vote_end_date yesterday haven't finished yet
```
He3lixxx, line 1990: something similar here. I'd suggest
> unfinished because still in EVALUATION_END_OFFSET_HOURS

evap/staff/tools.py
```diff
@@ -294,8 +295,14 @@ def merge_users(main_user, other_user, preview=False):
 
 
 def find_unreviewed_evaluations(semester, excluded):
+    # as evaluations are open for an offset of hours after vote_end_datetime, the evaluations ending yesterday are also xcluded during offset
```
He3lixxx, line 298: typo: xcluded

evap/staff/tools.py
```diff
@@ -294,8 +295,14 @@ def merge_users(main_user, other_user, preview=False):
...
+    exclude_date = date.today()
+    if datetime.now().hour < settings.EVALUATION_END_OFFSET_HOURS:
+        exclude_date -= timedelta(days=1)
```
He3lixxx, line 301: @janno42 just making sure: There is no case we ever expect this setting to be more than 24 hours, right?

He3lixxx (changes requested):
Cool! I have some minor refactoring annotations.

evap/staff/tests/test_views.py
dieknolle3333: no, we don't need it here.

evap/staff/tests/test_views.py
dieknolle3333: oh you're right, I was off by one there.
it's obvoiously a good idea to avoid the calculus (at least for me) ^^

evap/staff/tools.py
janno42: right, that woudn't really make sense

He3lixxx:
Uhh I think something went wrong here when you did the merge? It gives me changes that have already been merged into master now when I try to review this.

He3lixxx (approved):
Thanks :)

evap/staff/tests/test_views.py
```diff
@@ -1953,6 +1976,21 @@ def test_textanswers_full_view(self):
             page = self.app.get(self.url + '?view=full', user=self.manager, status=200)
             self.assertContains(page, self.answer)
 
+    @override_settings(EVALUATION_END_OFFSET_HOURS=26)
```
janno42, line 1979: as discussed in this pr, we don't expect this value to be more than 24 hours. so maybe it's a bit strange to have this test value in here?

He3lixxx: (I think the idea here was to make sure the test even runs in the weirdest of time-zone-change and leap-second combinations. If it makes sense to keep 26 here, we'd probably want to add an explaining comment)

janno42: okay. i don't think its very important to make the test time change proof (we didn't specify what should happen when reviewing text answers on a time zone change on production 😉), but if you add a comment it should be fine.

dieknolle3333: ok, I hope this is better.

janno42: probably something like this would explain it even better:

```
# use offset of more than 25 hours to make sure the test doesn't fail even on combined time zone change and leap second
```

dieknolle3333: got it, thanks for the suggestion!

janno42: thanks!

janno42 (approved):

Issue #1561, opened 2021-02-17
Investigate navbar cache removal


janno42:
See also discussion in #1530.

We should decide whether the navbar cache should be removed because it caused several troubles in the past. A performance check is called for to base the decision on.

He3lixxx:
Closing in favor of #1645, we will remove it there.

PR #1587, opened 2021-04-26
Fix staff mode problems

janno42:
The `OIDC_AFTER_USERLOGIN_HOOK` introduced in #1562 is actually not working. It looks like I found this part in the docs of a library that we're not using. 🤦 

This PR removes the `after_login_ function` added in the previous PR and instead deletes the navbar cache when accessing the `evaluation:index` view as a logged-in user. This is the redirect location after login, which is always passed.

Also, because entering the staff mode takes a few seconds on production (loading the active semester's index), a spinner is added to the button for immediate feedback.

### DIFF
--- a/evap/evaluation/views.py
+++ b/evap/evaluation/views.py
@@ -13,8 +13,10 @@ from django.views.decorators.debug import sensitive_post_parameters
 from django.views.i18n import set_language
 
 from evap.evaluation.forms import NewKeyForm, LoginEmailForm
-from evap.middleware import no_login_required
 from evap.evaluation.models import FaqSection, EmailTemplate, Semester
+from evap.middleware import no_login_required
+from evap.staff.tools import delete_navbar_cache_for_users
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


karyon:
I would suggest treating the loading time of the staff index page as a bug and the spinner as a workaround.

janno42:
Good point. Multiple pages could load a bit faster :)
I'll open an issue with a list of particular slow pages.

evap/staff/staff_mode.py
```diff
@@ -59,11 +58,9 @@ def update_staff_mode(request):
 
 def enter_staff_mode(request):
     update_staff_mode(request)
-    delete_navbar_cache_for_users([request.user])
```
He3lixxx, line 62: This works because our enter_staff_mode view will, afterwards, redirect to evaluation:index, which will do the deletion.

If, at some point, we decide to change this redirect, we will likely break it again, but the test should catch that, because re-entering after leaving should then fail, right?

evap/evaluation/views.py
```diff
@@ -88,6 +90,8 @@ def index(request):
         )
         return render(request, "index.html", template_data)
 
+    delete_navbar_cache_for_users([request.user])
```
He3lixxx, line 93: I had to dive into the old PR again to find out why exactly this was necessary. Do we maybe want to add a simple comment like:
```python
# The navbar might contain CSRF tokens that will be invalidated by django on login
```
?

He3lixxx (approved):

evap/staff/staff_mode.py
janno42: yes, i think so, too. and hopefully #1561 will make all of this obsolete :)

PR #1589, opened 2021-05-21
Update testdata to new default format from django 3.1 on

He3lixxx:
Quote from [the django 3.1 release notes](https://docs.djangoproject.com/en/3.2/releases/3.1/):
> JSON and YAML serializers, used by dumpdata, now dump all data with Unicode by default. If you need the previous behavior, pass ensure_ascii=True to JSON serializer, or allow_unicode=False to YAML serializer.

I'm fine with this, but it's polluting PRs (#1576, #1564). I think it's best to move this change into a single, own commit -- so here it is.

(I simply executed `./manage.py reload_testdata` followed by `./manage.py dump_testdata`)

### DIFF
--- a/evap/development/fixtures/test_data.json
+++ b/evap/development/fixtures/test_data.json
@@ -105,7 +105,7 @@
     "type": 10,
     "name_de": "Vorlesung 1 - Inhalt",
     "name_en": "Lecture 1 - Content",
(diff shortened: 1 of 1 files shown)


pixunil (approved):
Nice :sparkles:

PR #1597, opened 2021-07-05
Importer messages: More intuitive counts / numbers

### LINKED ISSUE #1590: Importer message refactoring
This should be implemented after the changes in #1579 have been merged.

The importers (`EnrollmentImporter`, `UserImporter`) and person management operations (importing and copying contributors/participants) show messages like
- "N users will be created: [list of users]"
- "Successfully created N users: [list of users]"
- "N participants added to the evaluation X: [list of users]"

In case N=0, the messages should instead read something like "No users will be/were created." or "No participants were added to the evaluation X."
Also, the case N=1 should be handled as well (see https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#pluralization).

FSadrieh:
Fixes #1590


evap/staff/importers.py
```diff
@@ -653,21 +680,29 @@ def process_contributors(self, evaluation, test_run, user_list):
         already_related = [contribution.contributor for contribution in already_related_contributions]
         if already_related:
             msg = format_html(_("The following {} users are already contributing to evaluation {}:"), len(already_related), evaluation.full_name)
+            msg = format_html(ngettext('The following user is already contributing to evaluation %(name)s:', 'The following %(users)d users are already contributing to evaluation %(name)s:', len(already_related),) % {'users': len(already_related),'name': evaluation.full_name})
```
niklasmohrin, line 683: this should be `+=` or the line above should be removed

niklasmohrin: We don't usually do percent-formatted strings, please use [f-strings](https://www.geeksforgeeks.org/formatted-string-literals-f-strings-python/) instead

evap/staff/importers.py
```diff
@@ -653,21 +680,29 @@ def process_contributors(self, evaluation, test_run, user_list):
...
         users_to_add = [user for user in user_list if user not in already_related]
 
-        if not test_run:
-            for user in users_to_add:
-                order = Contribution.objects.filter(evaluation=evaluation).count()
-                Contribution.objects.create(evaluation=evaluation, contributor=user, order=order)
-            msg = format_html(_("{} contributors added to the evaluation {}:"), len(users_to_add), evaluation.full_name)
+        if len(users_to_add) == 0:
+            if not test_run:
+                evaluation.participants.add(*users_to_add)
```
niklasmohrin, line 693: why is this needed? this expands to 0 users

evap/staff/importers.py
```diff
@@ -653,21 +680,29 @@ def process_contributors(self, evaluation, test_run, user_list):
...
-            msg = format_html(_("{} contributors would be added to the evaluation {}:"), len(users_to_add), evaluation.full_name)
-        msg += create_user_list_html_string_for_message(users_to_add)
+            if not test_run:
+                for user in users_to_add:
+                    order = Contribution.objects.filter(evaluation=evaluation).count()
+                    Contribution.objects.create(evaluation=evaluation, contributor=user, order=order)
+                msg = format_html(ngettext('One contributor added to the evaluation %(name)s:', '%(user)s contributors added to the evaluation %(name)s:', len(users_to_add),) % {'user': len(users_to_add), 'name': evaluation.full_name})
+            else:
+                msg = format_html(ngettext('One contributor would be added to the evaluation %(name)s:', '%(user)s contributors would be added to the evaluation %(name)s:', len(users_to_add),) % {'user': len(users_to_add), 'name': evaluation.full_name})
+            msg += create_user_list_html_string_for_message(users_to_add)
```
niklasmohrin, line 705: I feel like you can reduce the intersection between the branches a bit, maybe start with `msg = ""` before this hunk, then conditionally add the count and the words `"contributor"` or `"contributors"`, then check for `test_run` (and append `"would be"`), and so on. I don't know if it would play out as nicely as I am imagining it right now

evap/staff/tests/test_importers.py
```diff
@@ -149,9 +153,28 @@ def test_disallow_non_string_types(self):
             set(errors[ImporterError.SCHEMA])
         )
 
+    def test_user_success_messages(self):
+        user_list, success_messages, warnings, errors = UserImporter.process(self.valid_excel_content, test_run=True)
+        self.assertIn('The import run will create 2 users:<br />Lucilia Manilium (lucilia.manilium@institution.example.com)<br />Bastius Quid (bastius.quid@external.example.com)', success_messages)
```
niklasmohrin, line 158: It is a bit odd that the full html is here, although that is not really what you want to test right now, maybe you could check for only the message and then check the two names in separate asserts. I am not even sure whether the check for the names is appropriate here

evap/staff/tests/test_importers.py
```diff
@@ -297,8 +320,29 @@ def test_replace_consecutive_and_trailing_spaces(self):
...
... />Bastius Quid (bastius.quid@external.example.com)<br />Diam Sed (345@external.example.com)<br />Eirmod Nonumy (456@external.example.com)<br />Electram Romano (111@external.example.com)<br />Invidunt Tempor (789@external.example.com)<br />Diam Synephebos (diam.synephebos@institution.example.com)<br />Elitr Sadipscing (234@external.example.com)<br />Graecis Sic (890@external.example.com)<br />Labore Takimata (678@external.example.com)<br />Sanctus Aliquyam (567@external.example.com)<br />Sit Dolor (asd@external.example.com)<br />Ipsum Lorem (ipsum.lorem@institution.example.com)<br />Consetetur Amet (123@external.example.com)<br />Latinas Menandri (latinas.menandri@institution.example.com)<br />Lucilia Manilium (lucilia.manilium@institution.example.com)<br />Christoph Prorsus (123@institution.example.com)<br />Eadamque Praeterea (345@institution.example.com)<br />Ferdi Itaque (789@institution.example.com)<br />Probabo Itch (234@institution.example.com)<br />Tam Sed (456@institution.example.com)<br />Torquate Metrodorus (torquate.metrodorus@institution.example.com)<br />Augendas Multi (567@institution.example.com)<br />Invitat Honoris (111@institution.example.com)', success_messages)
```
niklasmohrin, line 328: same here, we only really care about the first sentence here, right?

niklasmohrin (changes requested):
I noticed that in the issue, the proposed message is "No contributors ....", here it is "Zero contributors ..."

evap/staff/importers.py
janno42: keeping translations in mind, it is better to create complete sentences and not to add words to the middle of a sentence

evap/staff/importers.py
```diff
@@ -688,51 +701,53 @@ def process_participants(self, evaluation, test_run, user_list):
...
-        if not test_run:
-            for user in users_to_add:
-                order = Contribution.objects.filter(evaluation=evaluation).count()
-                Contribution.objects.create(evaluation=evaluation, contributor=user, order=order)
-            msg = format_html(_("{} contributors added to the evaluation {}:"), len(users_to_add), evaluation.full_name)
+        if len(users_to_add) == 0:
+            if not test_run:
+                msg = format_html(_("Zero contributors added to the evaluation {}"), evaluation.full_name)
+            else:
+                msg = format_html(_("Zero contributors would be added to the evaluation {}"), evaluation.full_name)
```
He3lixxx, line 741: Shouldn't this be "No" to be consistent?

evap/staff/importers.py
```diff
@@ -498,25 +498,33 @@ def write_enrollments_to_db(self, semester, vote_start_datetime, vote_end_date):
...
+                if len(students_created) == 0:
+                    msg += format_html(_(" no students"))
+                else:
+                    msg += format_html(ngettext(' one student', ' {users} students', len(students_created),).format(users = len(students_created)))
+                    if len(responsibles_created) == 0:
+                        msg += format_html(_(" and no contributors:"))
+                    else:
+                        msg += format_html(ngettext(' and one contributor:', ' and {users} contributors:', len(responsibles_created),).format(users = len(responsibles_created)))
+                msg += create_user_list_html_string_for_message(students_created + responsibles_created)
+
```
He3lixxx, line 514: Maybe we should try to prevent these 3 levels of nested if-else blocks?
We could try something like:
```python3
student_count = len(students_created)
student_word = _("no") if student_count == 0 else str(student_count)  # TODO: Should use pgettext
responsible_count = len(responsibles_created)
responsible_word = _("no") if responsible_count == 0 else str(responsible_count)  # TODO: Should use pgettext

student_phrase = ngettext("{count} student", "{count} students", student_count).format(count=student_word)
responsible_phrase = ngettext("{count} responsible", "{count} responsibles", responsible_count).format(count=responsible_word)

message = ngettext(
    'Successfully created one course/evaluation, {student_phrase} and {reponsible_phrase}',
    'Successfully created {evaluation} courses/evaluations, {student_phrase} and {reponsible_phrase}''
).format(responsible_phrase=responsible_phrase, student_phrase=student_phrase)
```

what do you think?

evap/staff/importers.py
```diff
@@ -498,25 +498,33 @@ def write_enrollments_to_db(self, semester, vote_start_datetime, vote_end_date):
...
+            msg = format_html(ngettext('Successfully created one course/evaluation,', 'Successfully created {evaluation} courses/evaluations,', len(self.evaluations),).format(evaluation = len(self.evaluations)))
+            if len(students_created) == 0 and len(responsibles_created) == 0:
+                msg += format_html(_(" no students and no contributors."))
+            else:
+                if len(students_created) == 0:
+                    msg += format_html(_(" no students"))
+                else:
+                    msg += format_html(ngettext(' one student', ' {users} students', len(students_created),).format(users = len(students_created)))
+                    if len(responsibles_created) == 0:
+                        msg += format_html(_(" and no contributors:"))
```
He3lixxx, line 510: mismatch between code name (responsibles) and text name (contributors). One should be fixed (but if you don't want to, you don't have to).

evap/staff/importers.py
```diff
@@ -497,26 +497,51 @@ def write_enrollments_to_db(self, semester, vote_start_datetime, vote_end_date):
...
+
+        message = ngettext(
+            "Successfully created one course/evaluation, {student_phrase} and {responsible_phrase}",
+            "Successfully created {evaluation} courses/evaluations, {student_phrase} and {responsible_phrase}",
+            len(self.evaluations),
+        ).format(responsible_phrase=responsible_phrase, student_phrase=student_phrase, evaluation=len(self.evaluations))
+        if student_word != "no" or responsible_word != "no":
+            message += ":"
+        message += create_user_list_html_string_for_message(students_created)
+        self.success_messages.append(message)
```
janno42, line 521: shouldn't this be within the if?

evap/staff/importers.py
```diff
@@ -610,9 +635,17 @@ def save_users_to_db(self):
                         ).format(error)
                     )
                     raise
-
-        msg = format_html(_("Successfully created {} users:"), len(created_users))
-        msg += create_user_list_html_string_for_message(created_users)
+        if len(created_users) == 0:
+            msg = format_html(_("No users were created"))
```
janno42, line 639:
```suggestion
            msg = format_html(_("No users were created."))
```

evap/staff/importers.py
```diff
@@ -630,8 +663,17 @@ def create_test_success_messages(self):
         filtered_users = [user_data for user_data in self.users.values() if not user_data.user_already_exists()]
 
         self.success_messages.append(_("The test run showed no errors. No data was imported yet."))
-        msg = format_html(_("The import run will create {} users:"), len(filtered_users))
-        msg += create_user_list_html_string_for_message(filtered_users)
+        if len(filtered_users) == 0:
+            msg = format_html(_("The import run will create no users"))
```
janno42, line 667:
```suggestion
            msg = format_html(_("The import run will create no users."))
```

evap/staff/importers.py
```diff
@@ -723,16 +786,32 @@ def process_contributors(self, evaluation, test_run, user_list):
...
         users_to_add = [user for user in user_list if user not in already_related]
 
-        if not test_run:
-            for user in users_to_add:
-                order = Contribution.objects.filter(evaluation=evaluation).count()
-                Contribution.objects.create(evaluation=evaluation, contributor=user, order=order)
-            msg = format_html(_("{} contributors added to the evaluation {}:"), len(users_to_add), evaluation.full_name)
+        if len(users_to_add) == 0:
+            if not test_run:
+                msg = format_html(_("No contributors added to the evaluation {}"), evaluation.full_name)
```
janno42, line 791:
```suggestion
                msg = format_html(_("No contributors added to the evaluation {}."), evaluation.full_name)
```

evap/staff/importers.py
```diff
@@ -723,16 +786,32 @@ def process_contributors(self, evaluation, test_run, user_list):
...
+            else:
+                msg = format_html(_("No contributors would be added to the evaluation {}"), evaluation.full_name)
```
janno42, line 793:
```suggestion
                msg = format_html(_("No contributors would be added to the evaluation {}."), evaluation.full_name)
```

janno42 (changes requested):

evap/staff/importers.py
```diff
@@ -497,26 +497,51 @@ def write_enrollments_to_db(self, semester, vote_start_datetime, vote_end_date):
...
+            count=responsible_word
         )
-        msg += create_user_list_html_string_for_message(students_created + responsibles_created)
-        self.success_messages.append(msg)
+
+        message = ngettext(
+            "Successfully created 1 course/evaluation, {student_phrase} and {responsible_phrase}",
+            "Successfully created {evaluation} courses/evaluations, {student_phrase} and {responsible_phrase}",
+            len(self.evaluations),
+        ).format(responsible_phrase=responsible_phrase, student_phrase=student_phrase, evaluation=len(self.evaluations))
```
niklasmohrin, line 517: Why do the evaluations get special treatment here? Why isn't it analogous to `responsible_phrase` and `student_phrase`?

evap/staff/importers.py
```diff
@@ -497,26 +497,51 @@ def write_enrollments_to_db(self, semester, vote_start_datetime, vote_end_date):
...
+        if student_word != "no" or responsible_word != "no":
```
niklasmohrin, line 518: I think a `student_count > 0` would fit better

evap/staff/importers.py
```diff
@@ -497,26 +497,51 @@ def write_enrollments_to_db(self, semester, vote_start_datetime, vote_end_date):
...
             for evaluation, participants in participants_per_evaluation.items():
                 evaluation.participants.add(*participants)
+            self.create_success_messages(students_created, responsibles_created)
 
-        msg = format_html(
-            _("Successfully created {} courses/evaluations, {} participants and {} contributors:"),
-            len(self.evaluations),
-            len(students_created),
-            len(responsibles_created),
+    def create_success_messages(self, students_created, responsibles_created):
```
niklasmohrin, line 502: The name seems a bit off to me: I would assume that a function `create*` returns whatever was created. Maybe something like "add" or "append" would fit better? What do you think?

This only creates one message, so it shouldn't be "messages" either

evap/staff/importers.py
```diff
@@ -497,26 +497,51 @@ def write_enrollments_to_db(self, semester, vote_start_datetime, vote_end_date):
...
-            len(filtered_users),
+            ngettext(
+                "The import run will create 1 course/evaluation",
+                "The import run will create {evaluation} courses/evaluations",
+                len(self.evaluations),
+            ).format(evaluation=len(self.evaluations))
         )
-        msg += create_user_list_html_string_for_message(filtered_users)
+        if len(filtered_users) == 0:
+            msg += format_html(_(" and no users."))
```
niklasmohrin, line 535: I think I like the `*_phrase` approach from the other method better than adding these strings together. Ignoring translation for a second, you could have something like `msg = f"The import run will create {evaluations_phrase} and {users_phrase}"` here. What do you think?

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked
🔽 
🎉 Frederic is hopefully happy

evap/staff/importers.py
```diff
@@ -689,21 +731,40 @@ def process_participants(self, evaluation, test_run, user_list):
...
+                    ).format(name=evaluation.full_name, users=len(users_to_add))
+                )
+            else:
+                msg = format_html(
+                    ngettext(
+                        "1 participant would be added to the evaluation {name}:",
+                        "{users} participants would be added to the evaluation {name}:",
+                        len(users_to_add),
+                    ).format(name=evaluation.full_name, users=len(users_to_add))
+                )
```
niklasmohrin, line 766: There are multiple places with this much duplication, just for the word "would". Is there any chance that one could define
```python
if test_run:
    would_word = "would be "
else:
    would_word = ""
# ...
x = f"1 participant {would_word}added to the evaluation"
```
or similar? Or is that too much trouble for translation @janno42 ?

evap/staff/importers.py
```diff
@@ -723,16 +786,32 @@ def process_contributors(self, evaluation, test_run, user_list):
...
+                msg = format_html(_("No contributors would be added to the evaluation {}."), evaluation.full_name)
         else:
-            msg = format_html(
-                _("{} contributors would be added to the evaluation {}:"), len(users_to_add), evaluation.full_name
-            )
-        msg += create_user_list_html_string_for_message(users_to_add)
+            if not test_run:
+                for user in users_to_add:
+                    order = Contribution.objects.filter(evaluation=evaluation).count()
+                    Contribution.objects.create(evaluation=evaluation, contributor=user, order=order)
```
niklasmohrin, line 798: I think this should be outside of the `len(users_to_add) == 0` block, the loop doesn't run if the length is zero anyways and then one would be able to spot / skip the pluralization code better

niklasmohrin (commented):
This seems rather involved, but if that's what's necessary I am fine with it. I would prefer if we could deduplicate some strings though

evap/staff/importers.py
niklasmohrin: fyi he said its complicated

janno42: right, we can't assume there is just one word because the german translation can have a splitted verb ("wird hinzugefügt"/"würde hinzugefügt werden").
therefore translations always need to be complete phrases.

evap/staff/tests/test_importers.py
```diff
@@ -372,7 +402,37 @@ def test_replace_consecutive_and_trailing_spaces(self):
...
         success_messages, __, __ = EnrollmentImporter.process(excel_content, self.semester, None, None, test_run=True)
-        self.assertIn("The import run will create 1 courses/evaluations and 3 users", "".join(success_messages))
+        self.assertIn("The import run will create 1 course/evaluation and 3 users:", "".join(success_messages))
+
+    def test_enrollment_success_messages(self):
+        excel_content = excel_data.create_memory_excel_file(excel_data.test_enrollment_data_filedata)
+        success_messages, __, __ = EnrollmentImporter.process(excel_content, self.semester, None, None, test_run=True)
+        self.assertIn("The import run will create 23 courses/evaluations and 23 users:", success_messages[4])
+        success_messages, __, __ = EnrollmentImporter.process(
+            excel_content, self.semester, self.vote_start_datetime, self.vote_end_date, test_run=False
```
He3lixxx, line 412: It might make sense to create a `importer_args` variable and do the calls like `.process(*importer_args, test_run=True)`.

evap/staff/tests/test_importers.py
```diff
@@ -372,7 +402,37 @@ def test_replace_consecutive_and_trailing_spaces(self):
...
+        self.assertIn("The import run will create 1 course/evaluation and 1 user:", success_messages[3])
+        success_messages, __, __ = EnrollmentImporter.process(
+            excel_content, self.semester, self.vote_start_datetime, self.vote_end_date, test_run=False
+        )
+        self.assertIn("Successfully created 1 course/evaluation, no students", success_messages[2])
+
+        old_evaluation_count = Evaluation.objects.all().count()
+        Evaluation.objects.get(course__name_de="newBauen").delete()
+        Course.objects.get(name_de="newBauen").delete()
+        self.assertEqual(Evaluation.objects.all().count(), old_evaluation_count - 1)
```
He3lixxx, line 429: same as above (I'd remove the old_count variable and the assertion).

evap/staff/tests/test_importers.py
```diff
@@ -459,14 +516,65 @@ def test_import_new_participant(self):
...
-        self.assertIn("1 participants added to the evaluation", "".join(success_messages))
+        self.assertIn("1 participant added to the evaluation", "".join(success_messages))
         self.assertIn("{}".format(self.participant2.email), "".join(success_messages))
 
         self.assertEqual(self.evaluation1.participants.count(), 2)
         self.assertEqual(set(self.evaluation1.participants.all()), set([self.participant1, self.participant2]))
+
+    def test_person_success_messages(self):
+        success_messages, __, __ = PersonImporter.process_source_evaluation(
+            ImportType.PARTICIPANT, self.evaluation1, test_run=True, source_evaluation=self.evaluation3
```
He3lixxx, line 533: same with a local variable containing the arguments we don't really care about in this test.

evap/staff/tests/test_importers.py
```diff
@@ -200,10 +230,10 @@ def test_valid_file_import(self):
             excel_content, self.semester, self.vote_start_datetime, self.vote_end_date, test_run=False
         )
         self.assertIn(
-            "Successfully created 23 courses/evaluations, 6 participants and 17 contributors:",
+            "Successfully created 23 courses/evaluations, 6 students and 17 responsibles",
             "".join(success_messages),
         )
-        self.assertIn("Ferdi Itaque (789@institution.example.com)", "".join(success_messages))
+        self.assertIn("Torquate Metrodorus (torquate.metrodorus@institution.example.com)", "".join(success_messages))
```
He3lixxx, line 236: What happened here? The source file was not changed, was it?

He3lixxx (changes requested):
git commit -am "Oh no, jetzt will der wieder alles ändern und danach geht wieder nichts mehr" :P

janno42:
just a quick reminder that we currently have other ongoing changes at the importer as well and keeping this laying around won't make the merge easier :)

evap/staff/importers.py
```diff
@@ -572,26 +572,57 @@ def write_enrollments_to_db(self, semester, vote_start_datetime, vote_end_date):
...
+
+        msg = _("The import run will create {evaluation_phrase} and {user_phrase}").format(
+            evaluation_phrase=evaluation_phrase, user_phrase=user_phrase
         )
-        msg += create_user_list_html_string_for_message(filtered_users)
+        if user_count == 0:
+            msg += format_html(_("."))
+        else:
+            msg += format_html(_(":"))
+            msg += create_user_list_html_string_for_message(filtered_users)
```
niklasmohrin, line 625: Neither `.` nor `:` are translated or escaped, so both `_` and `format_html` are unneeded. 

Additionally, the `create_user_list_html_string_for_message` returns an already escaped string (it is an instance of `django.utils.safestring.SafeData`), but when you "add" it onto another string, this knowledge is lost and the created string will be escaped again - we don't want that. This is where `format_html` comes in, it does a `conditional_escape` on the inserted parameters (except the first one), so the safe data wont be escaped again. Since your phrases from above need to be escaped too, you probably want something like `format_html("{}: {}", msg, create_user_list(...))`

Interestingly, your tests don't fail although this gets escaped twice :thinking:

evap/staff/importers.py
```diff
@@ -705,8 +744,17 @@ def create_test_success_messages(self):
...
+        if len(filtered_users) == 0:
+            msg = format_html(_("The import run will create no users."))
+        else:
+            msg = format_html(
+                ngettext(
+                    "The import run will create {user_count} user:",
+                    "The import run will create {user_count} users:",
+                    len(filtered_users),
+                ).format(user_count=len(filtered_users))
+            )
```
niklasmohrin, line 756: This `.format` for the first argument of `format_html` is an anti-pattern here and a vulnerability in other places. As a rule of thumb, you should _never_ use `format_html` with only one argument. In your case here, you could write
```python
msg = format_html(
    _("Successfully created {count} {object}:{list}"),
    count=len(created_users),
    object=ngettext("user", "users", len(created_users)),
    list=create_user_list_html_string_for_message(created_users),
)
```

evap/staff/importers.py
```diff
@@ -764,21 +812,42 @@ def process_participants(self, evaluation, test_run, user_list):
...
         if already_related:
             msg = format_html(
-                _("The following {} users are already participants in evaluation {}:"),
-                len(already_related),
-                evaluation.full_name,
+                ngettext(
+                    "The following user is already participant in evaluation {name}:",
+                    "The following {user_count} are already participants in evaluation {name}:",
+                    len(already_related),
+                ).format(user_count=len(already_related), name=evaluation.full_name)
```
niklasmohrin, line 819: ... and here, it is a vulnerability, because `evaluation.full_name` is not escaped anymore and could contain malicious contents. Pls fix :)

niklasmohrin (changes requested):
Somehow you have fallen right into the same hole as we all did with django string escaping - please take a look at https://docs.djangoproject.com/en/4.0/ref/utils/#module-django.utils.html . Note that the linter is complaining too :^)

evap/staff/importers.py
```diff
@@ -574,55 +574,45 @@ def write_enrollments_to_db(self, semester, vote_start_datetime, vote_end_date):
...
-        evaluation_word = pgettext("count", "no") if evaluation_count == 0 else str(evaluation_count)
-
-        student_phrase = ngettext("{count} student", "{count} students", student_count).format(count=student_word)
-        responsible_phrase = ngettext("{count} responsible", "{count} responsibles", responsible_count).format(
-            count=responsible_word
-        )
-        evaluation_phrase = ngettext(
-            "{count} course/evaluation", "{count} courses/evaluations", evaluation_count
-        ).format(count=evaluation_word)
+        if not students_created and not responsibles_created and not self.evaluations:
```
niklasmohrin, line 583: codecov says this case is never true in tests, a test for this should be easy

evap/staff/importers.py
```diff
@@ -574,55 +574,45 @@ def write_enrollments_to_db(self, semester, vote_start_datetime, vote_end_date):
...
             responsible_phrase=responsible_phrase, student_phrase=student_phrase, evaluation_phrase=evaluation_phrase
         )
-        if student_count == 0 and responsible_count == 0 and evaluation_count == 0:
-            message = "Nothing changend"
-        elif student_count > 0 or responsible_count > 0:
+
+        if students_created or responsibles_created:
             message += ":"
             message += create_user_list_html_string_for_message(students_created)
             message += create_user_list_html_string_for_message(responsibles_created)
```
niklasmohrin, line 597: This discards the knowledge that the value from `create_user_list_html_string_for_message` is already escaped. The part inside the `if` should probably be one `message = format_html(...)`

evap/staff/importers.py
```diff
@@ -574,55 +574,45 @@ def write_enrollments_to_db(self, semester, vote_start_datetime, vote_end_date):
...
         self.success_messages.append(message)
+        return message
```
niklasmohrin, line 599: No one uses the returned value here

evap/staff/importers.py
```diff
@@ -574,55 +574,45 @@ def write_enrollments_to_db(self, semester, vote_start_datetime, vote_end_date):
...
             evaluation_phrase=evaluation_phrase, user_phrase=user_phrase
         )
-        if user_count == 0:
-            msg += format_html(_("."))
+        if not filtered_users:
+            msg += "."
         else:
-            msg += format_html(_(":"))
-            msg += create_user_list_html_string_for_message(filtered_users)
+            msg += format_html("{}: {}", msg, create_user_list_html_string_for_message(filtered_users))
```
niklasmohrin, line 615: Same thing here: This does `str + SafeString`, you probably meant to do `=` and not `+=`

evap/staff/importers.py
```diff
@@ -717,16 +707,15 @@ def save_users_to_db(self):
...
-            msg = format_html(_("No users were created."))
+            msg = _("No users were created.")
         else:
             msg = format_html(
-                ngettext(
-                    "Successfully created {user_count} user:",
-                    "Successfully created {user_count} users:",
-                    len(created_users),
-                ).format(user_count=len(created_users))
+                _("Successfully created {object}: {list}"),
```
niklasmohrin, line 713: Ah, sorry, this doesn't work as I thought: In German, the "created" also needs to know about pluralization ("wurde erfolgreich erstellt" vs. "wurden ..."), so the whole sentence before the ":" should use `ngettext`

evap/staff/importers.py
```diff
@@ -744,17 +733,15 @@ def create_test_success_messages(self):
...
+        if not filtered_users:
+            msg = _("The import run will create no users.")
         else:
             msg = format_html(
-                ngettext(
-                    "The import run will create {user_count} user:",
-                    "The import run will create {user_count} users:",
-                    len(filtered_users),
-                ).format(user_count=len(filtered_users))
+                _("The import run will create {count} {object}: {list}"),
```
niklasmohrin, line 740: Same probably applies here

niklasmohrin (commented):
The pluralization happens in more places, it is probably most efficient to take the old version and go from there

evap/staff/fixtures/excel_files_test_data.py
```diff
@@ -162,6 +162,19 @@
...
+        ['Master', 'Quid', 'Bastius', 'bastius.quid@external.example.com', 'Seminar', 'no', 'newBauen', 'newBuild', '', 'newSed', 'newDiam', 'new345@external.institution.com']
+    ]
+}
+
+one_new_course_and_user_filedata = {
+    'MA Belegungen': [
+        ['Degree', 'Student last name', 'Student first name', 'Student email address', 'Course kind', 'Course is graded', 'Course name (de)', 'Course name (en)', 'Responsible title', 'Responsible last name', 'Responsible first name', 'Responsible email address'],
+        ['Master', 'Quid', 'Bastius', 'bastius.quid@external.example.com', 'Seminar', 'no', 'newBauen', 'newBuild', '', 'newSed', 'newDiam', 'new345@external.institution.com']
+    ]
+}
```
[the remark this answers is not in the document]
He3lixxx: Unresolving, as this is still duplicated.

evap/staff/importers.py
He3lixxx: (Codecov says the `if` is not covered, but the if will always be covered, only the return might not be covered. So, maybe the if and the return are actually both covered. See #1723 for details.)

evap/staff/tests/test_importers.py
```diff
@@ -206,9 +236,11 @@ def create_existing_course(self):
         return existing_course, existing_course_evaluation
 
     def test_valid_file_import(self):
-        success_messages, warnings, errors = EnrollmentImporter.process(
-            self.default_excel_content, self.semester, None, None, test_run=True
-        )
+        excel_content = excel_data.create_memory_excel_file(excel_data.test_enrollment_data_filedata)
+        import_args_withoutDates = [excel_content, self.semester, None, None]
+        import_args_withDates = [excel_content, self.semester, self.vote_start_datetime, self.vote_end_date]
```
He3lixxx, line 241: This must be the famous `snake_has_eatenPascalCase` convention.

Please go with consistent snake case: `import_args_without_dates` and `import_args_with_dates`

evap/staff/tests/test_importers.py
```diff
@@ -234,11 +264,11 @@ def test_valid_file_import(self):
 
     def test_degrees_are_merged(self):
         excel_content = excel_data.create_memory_excel_file(excel_data.test_enrollment_data_degree_merge_filedata)
+        import_args_withoutDates = [excel_content, self.semester, None, None]
+        import_args_withDates = [excel_content, self.semester, self.vote_start_datetime, self.vote_end_date]
```
He3lixxx, line 268: naming style

evap/staff/tests/test_importers.py
```diff
@@ -297,13 +324,12 @@ def test_course_type_and_degrees_are_retrieved_with_import_names(self):
 
     @override_settings(IMPORTER_MAX_ENROLLMENTS=1)
     def test_enrollment_importer_high_enrollment_warning(self):
+        excel_content = excel_data.create_memory_excel_file(excel_data.test_enrollment_data_filedata)
+        import_args_withoutDates = [excel_content, self.semester, None, None]
+        import_args_withDates = [excel_content, self.semester, self.vote_start_datetime, self.vote_end_date]
```
He3lixxx, line 329: naming style

evap/staff/tests/test_importers.py
```diff
@@ -401,16 +430,30 @@ def test_replace_consecutive_and_trailing_spaces(self):
...
-        self.assertIn("The import run will create 1 courses/evaluations and 3 users", "".join(success_messages))
+        success_messages, __, __ = EnrollmentImporter.process(*import_args, test_run=True)
+        self.assertIn("The import run will create 1 course/evaluation and 3 users:", "".join(success_messages))
 
-    def test_existing_course_is_not_recreated(self):
         existing_course, existing_course_evaluation = self.create_existing_course()
 
+        excel_content = excel_data.create_memory_excel_file(excel_data.one_new_course_and_user_filedata)
+        import_args_withoutDates = [excel_content, self.semester, None, None]
+        import_args_withDates = [excel_content, self.semester, self.vote_start_datetime, self.vote_end_date]
```
He3lixxx, line 442: naming style

He3lixxx (changes requested):

evap/staff/importers.py
FSadrieh: Fixed the +=. I will leave this open as a reminder for you two to figure out your differences ;)

evap/staff/importers.py
FSadrieh: Depending on how the importers work, this is never really reached. If the evaluations always get made

He3lixxx: So if we upload an empty file, it will result in an error earlier?

@janno42 On merge situations, this is basically the same as https://github.com/e-valuation/EvaP/pull/1683#issuecomment-1115302197 -- `self.evaluations` might contain evaluations where nothing will be changed, due to the merge.

evap/staff/importers.py
```diff
@@ -710,11 +713,13 @@ def save_users_to_db(self):
             msg = _("No users were created.")
         else:
             msg = format_html(
-                _("Successfully created {object}: {list}"),
-                object=ngettext("{user_count} user", "{user_count} users", len(created_users)).format(
-                    user_count=len(created_users)
-                ),
-                list=create_user_list_html_string_for_message(created_users),
+                _("{}: {}"),
```
He3lixxx, line 716: If we were really serious about internationalization with the goal of supporting all languages, we would have to do something to allow translating this here. With german and english, however, I think we don't have to translate this one.

evap/staff/tests/test_importers.py
```diff
@@ -237,10 +237,10 @@ def create_existing_course(self):
 
     def test_valid_file_import(self):
         excel_content = excel_data.create_memory_excel_file(excel_data.test_enrollment_data_filedata)
-        import_args_withoutDates = [excel_content, self.semester, None, None]
-        import_args_withDates = [excel_content, self.semester, self.vote_start_datetime, self.vote_end_date]
+        import_args_without_Dates = [excel_content, self.semester, None, None]
+        import_args_with_Dates = [excel_content, self.semester, self.vote_start_datetime, self.vote_end_date]
```
He3lixxx, line 241: Closer, but not yet fully there. You can do it!

evap/staff/importers.py
```diff
@@ -592,11 +593,13 @@ def create_success_messages(self, students_created, responsibles_created):
         )
 
         if students_created or responsibles_created:
-            message += ":"
-            message += create_user_list_html_string_for_message(students_created)
-            message += create_user_list_html_string_for_message(responsibles_created)
+            message = format_html(
+                "{}: {} {}",
```
He3lixxx, line 597: I don't think `create_user_list_html_string_for_message` ends or begins with a line break, does it?

If so, this would need a line break in between the last two pairs of braces.

He3lixxx (changes requested):

He3lixxx:
The linting error regarding `no-init` should go away after merging / rebasing on main. The other two linter messages should be fixed.

evap/staff/tests/test_importers.py
```diff
@@ -232,13 +261,26 @@ def test_valid_file_import(self):
         expected_user_count = old_user_count + 23
         self.assertEqual(UserProfile.objects.all().count(), expected_user_count)
 
+        success_messages, warnings, errors = EnrollmentImporter.process(*import_args, test_run=True)
+        self.assertIn("The import run will create no courses/evaluations and no users.", "".join(success_messages))
+        self.assertEqual(errors, {})
+
+        success_messages, warnings, errors = EnrollmentImporter.process(*import_args, test_run=False)
```
He3lixxx, line 268: warnings unused

evap/staff/tests/test_importers.py
```diff
@@ -541,24 +596,25 @@ def setUpTestData(cls):
...
 
         success_messages, warnings, __ = PersonImporter.process_source_evaluation(
             ImportType.CONTRIBUTOR, self.evaluation1, test_run=True, source_evaluation=self.evaluation1
         )
-        self.assertIn("0 contributors would be added to the evaluation", "".join(success_messages))
-        self.assertIn(
-            "The following 1 users are already contributing to evaluation", warnings[ImporterWarning.GENERAL][0]
-        )
+        self.assertIn("No contributors would be added to the evaluation", "".join(success_messages))
+        self.assertIn("The following user is already contributing to evaluation", warnings[ImporterWarning.GENERAL][0])
```
He3lixxx, line 611: hardcoded array indices here and below again.

evap/staff/tests/test_importers.py
```diff
@@ -571,16 +627,16 @@ def test_import_new_contributor(self):
         success_messages, __, __ = PersonImporter.process_source_evaluation(
             ImportType.CONTRIBUTOR, self.evaluation1, test_run=True, source_evaluation=self.evaluation2
         )
-        self.assertIn("1 contributors would be added to the evaluation", "".join(success_messages))
-        self.assertIn(f"{self.contributor2.email}", "".join(success_messages))
+        self.assertIn("1 contributor would be added to the evaluation", "".join(success_messages))
+        self.assertIn(str(self.contributor2.email), "".join(success_messages))
```
He3lixxx, line 631: what is the purpose of this change? Is this an accidental merge-undo?

He3lixxx (changes requested):

evap/staff/importers.py
```diff
@@ -777,21 +835,47 @@ def process_participants(self, evaluation, test_run, user_list):
...
+                msg = format_html(_("No participants added to the evaluation {}"), evaluation.full_name)
+
         else:
-            msg = format_html(
-                _("{} participants would be added to the evaluation {}:"), len(users_to_add), evaluation.full_name
-            )
-        msg += create_user_list_html_string_for_message(users_to_add)
+            if test_run:
+                msg = format_html(
+                    _("{object} would be added to the evaluation {name}: {list}"),
```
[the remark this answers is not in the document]
FSadrieh: The correct translation would be, e.g. "3 Teilnehmer würden zu der Evaluierung Backen hinzugefügt werden: ..." right?
Then this should not be a correct fix, right?:
msg = format_html(
                    _("{sentence} {name}: {list}"),
                    sentence=ngettext("{user_count} participant would be added to the evaluation", "{user_count} participants would be added to the evaluation", len(users_to_add))....

evap/staff/tests/test_importers.py
FSadrieh: I changed it back. I think the str was a mistake

evap/staff/importers.py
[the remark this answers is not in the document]
He3lixxx: Yes, and yes, you'd probably have to add a `{evaluation_name}` to the message for that to work, if I understand your question correctly.

evap/staff/importers.py
```diff
@@ -777,21 +830,46 @@ def process_participants(self, evaluation, test_run, user_list):
 
         if already_related:
             msg = format_html(
-                _("The following {} users are already participants in evaluation {}:"),
-                len(already_related),
-                evaluation.full_name,
+                "{sentence}: {list}",
+                sentence=ngettext(
+                    "The following user is already participating in evaluation {name}",
+                    "The following {user_count} are already participating in evaluation {name}",
```
He3lixxx, line 836: missing "users": "The following 3 are already participating [...]"

evap/staff/importers.py
```diff
@@ -815,12 +896,34 @@ def process_contributors(self, evaluation, test_run, user_list):
...
+            else:
+                msg = format_html(_("No contributors added to the evaluation {}."), evaluation.full_name)
         else:
-            msg = format_html(
-                _("{} contributors would be added to the evaluation {}:"), len(users_to_add), evaluation.full_name
-            )
-        msg += create_user_list_html_string_for_message(users_to_add)
+            if test_run:
+                msg = format_html(
+                    _("{sentence}: {list}"),
```
He3lixxx, line 908: Some other place above, this is not translated. We should probably unify this.

evap/staff/importers.py
```diff
@@ -815,12 +896,34 @@ def process_contributors(self, evaluation, test_run, user_list):
...
+                    sentence=ngettext(
+                        "{user_count} contributor would be added to the evaluation {name}",
+                        "{user_count} contributors would be added to the evaluation {name}",
+                        len(users_to_add),
+                    ).format(user_count=len(users_to_add), name=evaluation.full_name),
+                    list=create_user_list_html_string_for_message(users_to_add),
+                )
+            else:
+                msg = format_html(
+                    _("{sentence} added to the evaluation {name}: {list}"),
```
He3lixxx, line 918: Uniformly with the other places, this could be a `ngettext` on the "outer" translation, i.e. translate between:
* {count} contributor added to the evaluation {name}: {list}
* {count} contributors added to the evaluation {name}: list{}

(if we don't, `sentence` is a weird formatter string, we should use `contributor_object`)

evap/staff/tests/test_importers.py
```diff
@@ -169,9 +174,35 @@ def test_disallow_non_string_types(self):
...
+
+        __, success_messages, __, __ = UserImporter.process(self.valid_excel_content, test_run=True)
+        self.assertIn("The import run will create no users", "".join(success_messages))
+        __, success_messages, __, __ = UserImporter.process(self.valid_excel_content, test_run=False)
+        self.assertIn("No users were created.", "".join(success_messages))
+
 
 class TestEnrollmentImporter(TestCase):
     filename_random = os.path.join(settings.BASE_DIR, "staff/fixtures/random.random")
+    filename_one_course = os.path.join(settings.BASE_DIR, "staff/fixtures/random.random")
```
He3lixxx, line 205: something's wrong here.

evap/staff/tests/test_importers.py
```diff
@@ -612,14 +667,65 @@ def test_import_new_participant(self):
...
+        )
+        self.assertIn("2 participants would be added to the evaluation", "".join(success_messages))
+        success_messages, __, __ = PersonImporter.process_source_evaluation(
+            ImportType.PARTICIPANT, self.evaluation1, test_run=False, source_evaluation=self.evaluation3
+        )
+        self.assertIn("2 participants added to the evaluation", "".join(success_messages))
+        success_messages, __, __ = PersonImporter.process_source_evaluation(
+            ImportType.PARTICIPANT, self.evaluation2, test_run=True, source_evaluation=self.evaluation3
+        )
+        self.assertIn("1 participant would be added to the evaluation", "".join(success_messages))
```
He3lixxx, line 694: Why? We just imported participants from evaluation3 to evaluation2, why does it say "1 would be added" now?

evap/staff/importers.py
He3lixxx: (All tests passing indicates there is a test missing here)

He3lixxx (changes requested):

niklasmohrin:
PR #1634 has been blocked on this for a while and we would like to merge it next week. Do you think you can finish it by then? Otherwise, we have to see how to port your changes to the new architecture

PR #1598, opened 2021-07-12
Fix illegal / broken HTML

He3lixxx:
I ran all the pages through https://validator.w3.org and fixed the errors. I guess best reviewed by commit?


janno42 (approved):
very html. much beauty. wow.

PR #1585, opened 2021-04-18
Fix formatting discussions (introduce Black)

karyon:
I always wanted to try [Black](https://github.com/psf/black), an opinionated code formatter. It provides very few configuration options, which is nice: There's not much to discuss about, either we take it or we don't.

One of the few options that could be discussed about is the line length. I arbitrarily set it to 140, which is kind of a lot, but it's already adding quite a lot of lines. 150 added 300 lines less, 160 another 150. A line length of 130 added 500 more lines, 120 added another 800.

And then there's --skip-string-normalization. Without it, black would change all strings to use double quotes. That's another 2000 lines changed, and I thought that's not worth it.

I excluded a few files where I felt black didn't help. For migrations I felt that formatting existing ones is just unnecessary code churn. It would be nice to apply it to new ones though, but that's not so easy. Anyway, I just excluded them all.

What do you think?

Todos left:
* [x] Discuss whether we actually want this
* [x] Update to 21.4b0
* [x] ~Exclude settings.py?~
* [ ] check the 140 commit for formatting fails
* [ ] check the 120 commit for formatting fails
* [ ] Fail CI if not formatted properly
* [ ] Give management command
* [ ] Check for conflicts with pylint
* [ ] create .git-blame-ignore-revs
* [ ] enable string normalization

Nice to haves:
* [ ] Pyproject.toml: Move pylint settings there? Reference: https://github.com/ephios-dev/ephios/blob/main/pyproject.toml
* [ ] Add isort
* [ ] git precommit hook: Can we give something to users that automatically runs? Might produce "weird" formatting of some lines. See https://github.com/psf/black#version-control-integration / https://github.com/ephios-dev/ephios/blob/main/.pre-commit-config.yaml

PR #1600, opened 2021-07-19
Code formatting: Introduce black

He3lixxx:
Continuation of #1585 

Todos left:
* [x] Discuss whether we actually want this
* [x] Update to current version
* [x] ~Exclude settings.py?~
* [x] check for formatting fails (I will just re-check the ones given in #1585)
* [x] Fail CI if not formatted properly
* [x] Check for conflicts with pylint
* [x] enable string normalization
* [x] Management command (for black and pylint)
* [x] create .git-blame-ignore-revs (will do last, when commit hash is fixed)

Nice to haves:
* [ ] Pyproject.toml: Move pylint settings there? Reference: https://github.com/ephios-dev/ephios/blob/main/pyproject.toml
* [ ] Add isort


evap/evaluation/models.py
```diff
@@ -848,64 +957,87 @@ def annotate_with_participant_and_voter_counts(cls, evaluation_query):
...
 
     class Meta:
-        unique_together = (
-            ('evaluation', 'contributor'),
-        )
-        ordering = ['order', ]
+        unique_together = (("evaluation", "contributor"),)
+        ordering = [
+            "order",
+        ]
```
janno42: this feels inconsistent, but i don't think it's important to change it

janno42 (approved):
i looked trough the tiny diff and think its ready to :shipit:

He3lixxx:
I've added a .git-ignore-blame-rev and updated https://github.com/e-valuation/EvaP/wiki/Development-Tips with instructions on how to use it.

For new contributors, it might make sense to put a note in the README.md that they probably want their IDE to lint + format files on save or something? Since there might be discussion about this, I'll make a new PR for that. Pyproject.toml merge will also go in there.

PR #1602, opened 2021-07-26
Fix degree check when importing enrollments

### LINKED ISSUE #1595: Missing degree check in enrollment import
When importing enrollment data, the importer must check whether the degrees used in the uploaded file already exist (and show an error message if a degree doesn't yet exist and must be manually created).

However, the error message for non-existent degrees currently only works correctly if this degree is the one in the first occurrence of an evaluation. If the first occurrence has a valid degree and any of the following degrees does not exist, no error will be shown and the import will fail silently.

Instead, an error message should be shown if any entry in the imported file uses a non-existent degree.

niklasmohrin:
Closes #1595﻿

### DIFF
--- a/evap/staff/fixtures/excel_files_test_data.py
+++ b/evap/staff/fixtures/excel_files_test_data.py
@@ -132,6 +132,14 @@ test_enrollment_data_import_names_filedata = {
     ]
 }
 
+test_unknown_degree_error_filedata = {
+    'Sheet 1': [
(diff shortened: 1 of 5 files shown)


He3lixxx (approved):
Noice

janno42 (changes requested):
looks nice, just one little change request

janno42 (approved):
👍

PR #1605, opened 2021-07-26
Codacy: Various stuff. I want to get the A grade back!

He3lixxx:
I've disabled a lot of stuff on codacy that we do not care about, but some of the suggestions actually made sense, so I changed them here.

(Note: The disabled stuff alone lifted the grade back to A)

### DIFF
--- a/.prospector.yaml
+++ b/.prospector.yaml
@@ -12,10 +12,10 @@ doc-warnings: no
 strictness: veryhigh
 pep8:
   disable:
-    - E501  # line too long - we don't enforce some fixed line length but rather complain about long lines in reviews.
-    - W503  # line break before operator - looks better and PEP8 allows it
-    - E128  # hanging indents not allowed - we indent differently
-    - E126  # overindentation - we indent differently
+    - E501  # line too long - code formatting is done by black
+    - W503  # line break before operator
+    - E128  # hanging indents not allowed
+    - E126  # overindentation
     - N802  # function name should be lowercase - lots of false positives in tests that pylint doesn't show
     - N803  # argument name should be lowercase - pylint has this as well (and there we tweaked it)
     - N806  # variable in function should be lowercase - lots of false positives, mostly Formset classes made from factories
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -5,14 +5,12 @@
 [![Codacy Badge](https://app.codacy.com/project/badge/Grade/2cf538781fdc4680a7103bcf96417a9a)](https://www.codacy.com/gh/e-valuation/EvaP/dashboard)
 [![codecov](https://codecov.io/gh/e-valuation/EvaP/branch/master/graph/badge.svg)](https://codecov.io/gh/e-valuation/EvaP)
 
-
-## What is EvaP?
+## What is EvaP
```
janno42, line 8: it doesn't allow question marks in headings?

He3lixxx: [Yup](https://app.codacy.com/gh/e-valuation/EvaP/issues?bid=23052067&filters=W3siaWQiOiJMYW5ndWFnZSIsInZhbHVlcyI6WyJNYXJrZG93biJdfSx7ImlkIjoiQ2F0ZWdvcnkiLCJ2YWx1ZXMiOltdfSx7ImlkIjoiTGV2ZWwiLCJ2YWx1ZXMiOltdfSx7ImlkIjoiUGF0dGVybiIsInZhbHVlcyI6WyI2NjI2Il19LHsiaWQiOiJBdXRob3IiLCJ2YWx1ZXMiOltdfV0=), the [no-heading-punctuation](https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-no-heading-punctuation) rule is active by default. I don't actually care about the question mark here, I can see how someone could argue it to be a smell, but I'm also fine with just disabling the rule if you prefer that.

He3lixxx: Btw, same for the "no-multiple-empty-lines" rule. Personally, I don't really care if we put two empty lines before a new heading. With syntax highlighting, it shouldn't be necessary, but I don't think it does any damage.

karyon (approved):

README.md
janno42: as discussed, two empty lines before the heading are nice but i really don't understand the problem with question marks in headings from a linguistic perspective. in this case, both solutions (removing the question mark, disabling the rule) are fine for me :)

janno42 (approved):

README.md
He3lixxx: Removed the changes from README.md. Deactivated the two checks on codacy.

PR #1608, opened 2021-08-01
Remove manual reloading of DB models in tests

karyon:
This has been made obsolete by django 3.2, which now properly isolates
test data created in setUpTestData between tests.

(see https://docs.djangoproject.com/en/3.2/releases/3.2/#tests)


He3lixxx (approved):
Nice

PR #1614, opened 2021-08-09
Fix M2M logging when adding and removing participants in one request

### LINKED ISSUE #1594: Log M2M Clear
With #1579 participants and contributors of an evaluation can be replaced. In the code, the participants/contributors are cleared first and then the new ones are added.
The log only lists the newly added participants/contributors and does not show an entry for the deleted ones.
The deleted participants/contributors should also be logged.

PFischbeck:
This was due to the second change overwriting the first change within the same LogEntry. We fix this by collecting all changes so far, and then updating it collectively.

This fixes #1594.

### DIFF
--- a/evap/evaluation/models_logging.py
+++ b/evap/evaluation/models_logging.py
@@ -131,6 +131,7 @@ class LoggedModel(models.Model):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._logentry = None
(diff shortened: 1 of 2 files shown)


janno42 (approved):
besides the comment text in `log_m2m_change` that we talked about, this is very nice! thank you :)

janno42 (commented):
🎆

He3lixxx (approved):

PR #1616, opened 2021-08-09
Submit form and show spinner icon when changing language

### LINKED ISSUE #1609: Language change page reload
When changing the language by clicking on the button in the navbar, an AJAX request is sent to change the language, and then the page reloads in the new language. In some situations (especially on the semester page in staff mode), this leads to several seconds of inactivity after clicking the button before the page visibly reloads in the browser.
This should be changed to receive an immediate response, e.g., by using form submission like the one used for the staff mode buttons and adding the animated spinner.

PFischbeck:
This fixes #1609.

And this forced me to slightly improve a few tests, since they now referred to the wrong form 😄

### DIFF
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -152,18 +152,8 @@
             .mouseout(() => row.addClass("hover-row"));
     });
 
-    function setLanguage(language) {
-        $.ajax({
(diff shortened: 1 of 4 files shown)


janno42 (approved):
b-e-a-utiful!

PR #1618, opened 2021-08-16
http2

niklasmohrin:
Related to #1515

I don't know if there is prior work on this, if so let me know

This enables h2c, which is http2 without tls, in vagrant. I don't know how tls is set up for http/1.1, so I didn't start working on this for now :) The [documentation](https://httpd.apache.org/docs/2.4/howto/http2.html) has a hint "Choose a strong SSLCipherSuite", I don't know if we have anything set there either. I will have to see how h2 works once I find out

This would also be a chance to document the tls setup somewhere

### DIFF
--- a/deployment/apache.template.conf
+++ b/deployment/apache.template.conf
@@ -1,3 +1,8 @@
+LoadModule http2_module modules/mod_http2.so
+# NOTE: h2c means http, h2 meaning https
+# I have no idea how tls is set up in production :(
+Protocols h2c http/1.1
+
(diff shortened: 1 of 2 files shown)


karyon:
there was prior work, see https://github.com/karyon/EvaP/commit/bb45eadf9449edf0c08b24ae8f4924810f729e7b

production uses letsencrypt, and that can't be used for development environments, that's all I remember :)

niklasmohrin:
This isn't going anywhere really, closing

Issue #1118, opened 2018-03-01
Text answer flagging


janno42:
It would be nice to have an option for flagging text answers during the review process. Staff users and reviewers should then have the possibility to display all flagged text answers of a semester in a new view. Flags won't influence anything else, especially approving text answers is independent.

pixunil:
Some ~80% code here: https://github.com/pixunil/EvaP/tree/textanswer-flagging

janno42:
An additional requirement came up a few days ago: In quick text answer review, evaluations, where all remaining unreviewed text answers are flagged, should not be suggested as the following evaluation for review.

BenBals:
We just discussed our review workflow on at a FSR meeting. We currently use the skip feature for exactly that reason. But flagging, especially with a unified list/view across evaluations, would be a much better solution.

PR #1619, opened 2021-08-16
textanswer flagging

LucPrestin:
**Summary**:
Adds flags to textanswer reviews to indicate the need for an discussion about a text answer with other staff members

**Resolves**: #1118 

**Changes**:
- Add TextAnswer.is_flagged boolean field
- Add buttons to toggle text answer flags
- Add view to list only flagged text answers
- enable flagging in the texanswers quick view
- enable flagging in the textanswers normal view (full, unreviewed and flagged)
- make all flag buttons yellow
- fix flag status display
- update tests to new url
- add safeguard to textanswer toggling


janno42 (commented):
i didn't test the functionality yet, but it looks promising so far.

please note that the evaluations suggested in quick review need to be filtered (https://github.com/e-valuation/EvaP/issues/1118#issuecomment-891029989) and that the issue calls for an overview page for all flagged text answers of a semester.

He3lixxx:
Did we forget properly reviewing this? If so, I'm sorry. There are quite some merge conflicts now, but they should be solvable.

On a quick glance, most of the code looks good. I would, however, try to abstain from "toggle" semantics with values where your local representation might be outdated. Consider this scenario:

Staff user 1 opens the review page. They are interrupted, so they suspend their computer. 90 Minutes later, they open it again, to continue reviewing.
In the mean time, staff user 2 has already flagged the answer. But, since staff user 1 didn't reload the page, they are still shown "not flagged" as state. Staff user 1 now also wants to flag, but the "toggle" at the server actually unflags.

(What makes it even worse here is that user 2 will have no indication, their screen will show "flagged" now -- but even without that: Just always send a "target" state to make your operations idempotent, and we don't have to worry about this)

janno42:
Will be reimplemented by @niklasmohrin after #1776 is merged

Issue #1592, opened 2021-06-21
Follow up to #1564 "Add additional text answers for rating questions"


He3lixxx:
Before merging #1564, there were still some open discussions regarding refactoring or performance considerations (query count). This is the resulting follow-up issue:

- https://github.com/e-valuation/EvaP/pull/1564#discussion_r580428637 indicated we might have O(N) queries in `get_evaluation_and_contributor_textanswer_sections`
- https://github.com/e-valuation/EvaP/pull/1564#discussion_r615369227 had another O(N) query issue there
- https://github.com/e-valuation/EvaP/pull/1564#discussion_r633689897 indicated a refactoring of `make_questionnaire_edit_forms` would be appropriate
- https://github.com/e-valuation/EvaP/pull/1564#discussion_r615370553 indicated it might make sense to let a `RatingAnswerField` take a `question`
- https://github.com/e-valuation/EvaP/pull/1564#discussion_r633643347 indicated the line `                            $($(this).closest(".row")).find(".btn-textanswer").click()` in student_vote.html might be slow
- There was a discussion that we should maybe add a database constraint: Only rating answers can have additional text answers. This was, for some time, asserted in the save method, but this will not help for .update() queries.

janno42:
all done (or decided not to change) in #1620

PR #1620, opened 2021-08-16
Text follow ups

niklasmohrin:
Some things mentioned in #1592 

okay, I cant remember what is already done all the time anymore
- [x] (comment) indicated we might have O(N) queries in `get_evaluation_and_contributor_textanswer_sections`
- [x] (comment) had another O(N) query issue there
- [x] (comment) indicated a refactoring of `make_questionnaire_edit_forms` would be appropriate
- [x] (comment) indicated it might make sense to let a `RatingAnswerField` take a `question`
- [x] (comment) indicated the line `$($(this).closest(".row")).find(".btn-textanswer").click()` in student_vote.html might be slow; edit: we concluded that we really want to click those buttons, only replaced jquery with vanilla js on that one
- [x] There was a discussion that we should maybe add a database constraint: Only rating answers can have additional text answers. This was, for some time, asserted in the save method, but this will not help for .update() queries.; edit: not possible as per my comment below
- [x] Related to "Keyboard navigation for voting site" We would want that when a user clicks on the "add additional text answer", the cursor focus goes to the input field automatically.
- [x] When the additional textanswer button should look "activated" / "pressed" when when the additional text answer field is visible.

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -1337,9 +1337,11 @@ CHOICES = {
 
 
 class Answer(models.Model):
-    """An abstract answer to a question. For anonymity purposes, the answering
-    user ist not stored in the object. Concrete subclasses are `RatingAnswerCounter`,
-    and `TextAnswer`."""
+    """
+    An abstract answer to a question. For anonymity purposes, the answering
+    user ist not stored in the object. Concrete subclasses are
+    `RatingAnswerCounter`, and `TextAnswer`.
+    """
 
     # we use UUIDs to hide insertion order. See https://github.com/e-valuation/EvaP/wiki/Data-Economy
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
@@ -1354,11 +1356,13 @@ class Answer(models.Model):
 
 
 class RatingAnswerCounter(Answer):
-    """A rating answer counter to a question.
+    """
+    A rating answer counter to a question.
     The interpretation depends on the type of question:
     unipolar: 1, 2, 3, 4, 5; where lower value means more agreement
     bipolar: -3, -2, -1, 0, 1, 2, 3; where a lower absolute means more agreement and the sign shows the pole
-    yes / no: 1, 5; for 1 being the good answer"""
+    yes / no: 1, 5; for 1 being the good answer
+    """
 
     answer = models.IntegerField(verbose_name=_("answer"))
     count = models.IntegerField(verbose_name=_("count"), default=0)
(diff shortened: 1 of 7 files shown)


evap/staff/views.py
```diff
@@ -1581,50 +1583,50 @@ def questionnaire_create(request):
     return render(request, "staff_questionnaire_form.html", dict(form=form, formset=formset, editable=True))
 
 
+def disable_all_except_named(fields: Dict[str, Any], names_of_editable: List[str]):
```
He3lixxx, line 1586: Did we discuss type annotations before? I feel like we did but I can't remember

He3lixxx: `names_of_editable` just has to be some kind of collection that can be tested with `in` / `__contains__()`, so it should probably be [`Container[str]`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Container). The call could be made with a tuple or a set instead of a list.

He3lixxx (commented):
Code looks good. I haven't tested. Comment because I don't really know about type annotations.

niklasmohrin:
According to https://forum.djangoproject.com/t/checkconstraint-involving-related-model/5351

> This is actually a PostgreSQL limitation. From their docs, the first Note in section 5.4.1 5:
> 
> >    Note
> > PostgreSQL does not support CHECK constraints that reference table data other than the new or updated row being checked.

Thus, having the additional textanswer check like proposed is not possible

evap/static/scss/components/_buttons.scss
```diff
@@ -262,4 +262,17 @@ a.collapse-toggle {
     height: 2.5rem;
     background-color: $lighter-gray;
     border-color: $medium-gray;
+
+    &.has-contents {
+        box-shadow: inset 0 0 10px $evap-dark-blue;
+        outline: 3px solid $evap-dark-blue;
+        // border-color: $evap-dark-blue;
+        // border-width: 0.2rem;
+    }
```
janno42, line 271: even fancier:

```suggestion
    &.has-contents > .fa-comment {
        font-weight: 900;
    }
```

evap/staff/views.py
```diff
@@ -1374,31 +1376,42 @@ def get_evaluation_and_contributor_textanswer_sections(evaluation, filter_textan
 
     evaluation_sections = []
     contributor_sections = []
+    evaluation_contributors = list(evaluation.course.responsibles.all())
```
He3lixxx, line 1379: `evaluation_responsibles` or `course_responsibles`?

He3lixxx (approved):
One typo, one variable renaming suggestion. Other code looks good to me :)

evap/evaluation/models.py
```diff
@@ -1372,6 +1376,7 @@ class Meta:
 class TextAnswer(Answer):
     """A free-form text answer to a question."""
 
+    # The original answer from the user may be altered by staff users before publishing.
```
janno42, line 1379:
```suggestion
    # If the text answer was changed during review, original_answer holds the original text
```

janno42 (approved):

PR #1626, opened 2021-09-06
Add a test utility function that creates RatingAnswerCounters

karyon:
Admittedly, the syntax is not 100% intuitive. An alternative for the dict would be a list, instead of 
```
make_rating_answer_counters(rating_question, contribution, {1:15, 2:12, 4:100, 5:2})
make_rating_answer_counters(rating_question, contribution, {2:6})
make_rating_answer_counters(yesno_question, contribution, {1:15, 5:2})
make_rating_answer_counters(bipolar_question, contribution, {-3:20, -2:15, 0:1, 1:15, 3:2})
```
we could have 
```
make_rating_answer_counters(rating_question, contribution, [15, 12, 0, 100, 2])
make_rating_answer_counters(rating_question, contribution, [0, 6, 0, 0, 0])
make_rating_answer_counters(yesno_question, contribution, [15, 2])
make_rating_answer_counters(bipolar_question, contribution, [20, 15, 0, 1, 15, 0, 2])
```

Syntactically I like the lists a bit more, but one would always need to specify all answers (...but maybe that's an advantage actually?) and the helper method would require more code for knowing which question types need which answers, a different default argument per type, and for filtering out counters with zero answers. What do you think?

### DIFF
--- a/evap/evaluation/tests/test_models.py
+++ b/evap/evaluation/tests/test_models.py
@@ -25,7 +25,12 @@ from evap.evaluation.models import (
     UserProfile,
 )
 from evap.grades.models import GradeDocument
-from evap.evaluation.tests.tools import let_user_vote_for_evaluation, make_contributor, make_editor
+from evap.evaluation.tests.tools import (
+    let_user_vote_for_evaluation,
+    make_contributor,
+    make_editor,
+    make_rating_answer_counters,
+)
 from evap.results.tools import calculate_average_distribution, cache_results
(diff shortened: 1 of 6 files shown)


janno42:
I would prefer the lists because they are similar to what we use for results calculation, but the dicts are also fine.

janno42 (approved):

karyon:
good point! no need to introduce another format. lists it is!

janno42 (approved):
nice work :)

PR #1627, opened 2021-09-19
Bootstrap 5

### LINKED ISSUE #1617: Bootstrap update
Update Bootstrap to version 5

janno42:
fix #1617. work in progress. please have a look at all things you can think of and report all remaining styling problems :)

open issues:

- [ ] tooltip for additional text answer button on student vote page is moving weirdly


evap/student/templates/student_vote.html
```diff
@@ -368,22 +362,18 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
             }
 
             function selectByNumberKey(row, num) {
+                var index = 2 * num - 1;
```
niklasmohrin, line 365:
```suggestion
                let index = 2 * num - 1;
```

niklasmohrin (commented):
After tabbing onto "I cant give feedback ...", the button (that we are still on) does not use the custom keyhandler anymore and the next tab event will be browser behavior. Still need to check whether that was broken before or not :man_shrugging:

niklasmohrin:
When a contributor section is closed with the normal collapse button, the "I cant give feedback" button opens the section again

niklasmohrin:
Some ideas for the FAQ site:
- more space after the last question
- an outline around the selected question + answer so that one can better tell where the answer ends / the next question starts
- the top heading looks very similar to the sub headings / it doesn't look very top-ish to me; maybe it's because the "are" and "the" are not capitalized? (just remove the "These are the"? center the top heading?)

janno42:
@niklasmohrin addressed your comments (responses in your posts)

PR #1629, opened 2021-09-20
isort this would be a great addition

niklasmohrin:


janno42 (approved):
1327 had this for ages and finally, here we are and now have it in EvaP as well! 🎉

He3lixxx (approved):
thanks :)

PR #1633, opened 2021-10-11
Use middleware for setting language

### LINKED ISSUE #1631: Lost in translation: Language preference dropped when browser is closed
The language preference is lost after I restart my browser. This happens on both Firefox and Chromium.

1. log in
2. set language to German (my system language is English)
3. close and reopen browser
4. go to evap, now the site is English again

@He3lixxx reproduced and said that the django cookie for language is discarded after the session ends. It gets restored if you have the "restore session when opening firefox" setting enabled

niklasmohrin:
Fixes #1631

Instead of setting session and user settings for language when they log in, it is now updated on every request. Now, when a logged in user visits the site with a fresh session, the preference from the database is used.

I couldn't figure out another way to do this easily. [An answer to this question](https://stackoverflow.com/questions/42953521/set-or-change-the-default-language-dynamically-according-to-the-user-django) also suggests using a third party middleware, https://pypi.python.org/pypi/django-user-language-middleware/. It does the same as this.

One test broke - it used the `HTTP_ACCEPT_LANGUAGE` header. I believe that it was handled by the django locale middleware.

Is this a problem? Other thoughts?

### DIFF
--- a/evap/results/tests/test_views.py
+++ b/evap/results/tests/test_views.py
@@ -93,7 +93,9 @@ class TestResultsView(WebTest):
         page = self.app.get(self.url, user=student).body.decode()
         self.assertLess(page.index(evaluation1.name_en), page.index(evaluation2.name_en))
 
-        page = self.app.get(self.url, user=student, extra_environ={"HTTP_ACCEPT_LANGUAGE": "de"}).body.decode()
+        student.language = "de"
+        student.save()
+        page = self.app.get(self.url, user=student).body.decode()
         self.assertGreater(page.index(evaluation1.name_de), page.index(evaluation2.name_de))
 
     # using LocMemCache so the cache queries don't show up in the query count that's measured here


evap/middleware.py
```diff
@@ -26,3 +28,21 @@ def process_view(request, view_func, _view_args, _view_kwargs):
...
+
+        if request.user.language:
+            translation.activate(request.user.language)
+        else:
+            request.user.language = translation.get_language()
+            request.user.save()
+        lang = request.user.language
+        response = get_response(request)
+        response.set_cookie(settings.LANGUAGE_COOKIE_NAME, lang)
+        return response
```
He3lixxx, line 46: - The average user will be logged in and authenticated, right? With this code, they will always have a `user.language` set except for their very first request after account creation, right? If that is the case, I would add a comment to that branch explaining that
- For the average request, we would call `translation.activate` and `set_cookie` here, right? Feels a bit heavy to me, and I think it's weird that we would need a custom middleware. But if that is what it takes to make it work, then that's just the way it is ¯\_(ツ)_/¯
- Setting the cookie is only necessary so the language is kept after the user logs out, right? In that case, I would also add a comment.

This will sync the languages between all devices, so a language change on one device will affect all other devices, right? @janno42 is that what we want?

evap/middleware.py
```diff
@@ -26,3 +28,23 @@ def process_view(request, view_func, _view_args, _view_kwargs):
...
+            return get_response(request)
+        if request.user.language == translation.get_language():
+            return get_response(request)
+
+        if request.user.language:
+            translation.activate(request.user.language)
+        else:
+            request.user.language = translation.get_language()
+            request.user.save()
+        lang = request.user.language
```
He3lixxx, line 45: You could inline this now, but I think this is also ok.

Issue #1574, opened 2021-03-14
Import data test improvements


janno42:
Importing enrollment data often is a tedious process and should be assisted as far as possible. Some potential problems in the data are currently not checked by the importer, and others generate a large number of warnings for the same problem.

These are changes that should be made to the testing phase of importing enrollment data:
- If the study degrees of a course do not match in all lines of the import file, only one warning should be shown, listing all study degrees found for the course.
- If a user has different study degrees in one import file, this should be shown as an error.
- A new warning section, "Similar course names", should be added. This should be filled with warnings for all pairs of course names with a small string distance. Examples are typos, where cases change, or a small number of letters are missing, changed, or added.

He3lixxx:
Regarding degrees:
* We change it so that only a single degree can be provided for one enrollment (i.e. the "degree of the user"). This removes the need for point 1 from the issue description.
* We add a check that the degree is identical in all enrollment rows of a single user. This solves point 2.


Point 3 remains as described above.

niklasmohrin:
Dependency seems appropriate; the triangle inequality should hold for edit distance, so maybe we can use something fancy to avoid n^2 runtime? I was first thinking about that TI exercise about closest-points-problem, but that uses metric space stuff; maybe k-neighbours kind of stuff is better? or we just take the n^2 :D

PR #1634, opened 2021-10-15
Rewrite staff importers (UserImporter, EnrollmentImporter)

He3lixxx:
Preparation to solving #1574.

Minimal implementation to pass all existing tests. Putting this here already to get some feedback on the overall architecture.

The idea is that importers start with the binary excel content and then put their current state through a series of steps that each gets them a bit closer to their result. On errors, it simply raises an error aborting the import. After each step, checkers can be run to add errors or warnings that are more user friendly. This hopefully allows to understand the existing checks and add new checks very easily.
I tried to separate and encapsulate the different concerns as much as possible: Logic required for an import, errors/warnings, message handling

Its probably best check out how user import and enrollment import files look like before looking at the code.

related code changes that need to be integrated:
- [x] xls was changed to xlsx in #1703.
- [x] #1714 added a test that didn't really match code anymore. I adapted it to the new infrastructure, but we might consider removing it.
- [x] #1683 allowed importing into existing courses.
- [ ] #1597 changes importer success messages to include user lists.
- [x] #1730 Adds links to user profiles in importer messages.


niklasmohrin (commented):
didn't read till the end, but here are some thoughts along the way

evap/staff/importers.py
```diff
@@ -1,691 +1,1281 @@
...
+    @staticmethod
+    def bulk_update_fields():
+        """Fields passed to bulk_update when updating existing users with new UserData"""
+        return ["first_name", "last_name", "title", "email", "is_active"]
+
+    def __init__(self, first_name: str, last_name: str, title: str, email: str):
+        object.__setattr__(self, "first_name", first_name.strip())
+        object.__setattr__(self, "last_name", last_name.strip())
+        object.__setattr__(self, "title", title.strip())
+        object.__setattr__(self, "email", clean_email(email))
```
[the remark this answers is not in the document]
He3lixxx: The intention was to make this usable as a key in a dictionary. For this, it needs `__eq__()` and `__hash__()`.
Maybe I should remove `eq=True` because that is true by default.

What comment would you like? Frozen tries its best to make the instances immutable by hooking setattr (I think). Immutable instances are necessary to have a safe `__hash__()` implementation, which `@dataclass` will then provide.

evap/staff/importers.py
```diff
@@ -1,691 +1,1281 @@
...
-    NAME = ("name", gettext_lazy("Name mismatches"), 1)
-    INACTIVE = ("inactive", gettext_lazy("Inactive users"), 2)
-    DUPL = ("duplicate", gettext_lazy("Possible duplicates"), 3)
-    IGNORED = ("ignored", gettext_lazy("Ignored duplicates"), 4)
+        return rows
 
-    DEGREE = ("degree", gettext_lazy("Degree mismatches"), 5)
-    MANY = ("too_many_enrollments", gettext_lazy("Unusually high number of enrollments"), 6)
 
+class DegreeImportMapper:
```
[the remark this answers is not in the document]
He3lixxx: How to organize all these classes if open for discussion, I'm fine with having them all in one file, but I'd also be fine with multiple files.

You would propose having three files, `importers.py`, `importer_checkers.py` and `importer_mappers.py`? Where would the `ExcelFileRowMapper` be?

evap/staff/importers.py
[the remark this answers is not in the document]
niklasmohrin: I think a short "`object.__setattr__` is needed to initialize instances of frozen dataclasses" and / or a link to https://stackoverflow.com/a/58336722/13679671 or a similar reference should suffice. It's just to immediately answer the question that first readers will ask themselves; I know that I "wasted" time on finding out what this did and why because it totally caught me off guard :laughing:

karyon:
Random thought, if you want to scope-creep your refactor, you could improve testability of the importer. A big step towards better testability might be having the test data in code instead of less-nice-to-edit xls files. Either the test data from the code could first be converted to an xls file, or the importers would need to enable tests to insert non-xls input somewhere.

evap/staff/importers.py
```diff
@@ -1,691 +1,1281 @@
...
+        return self._messages_with_level_by_category(ImporterMessage.Level.WARNING)
+
+    def errors_by_category(self):
+        return self._messages_with_level_by_category(ImporterMessage.Level.ERROR)
+
+    def forward_messages_to_django(self, request):
+        method_by_level = {
+            ImporterMessage.Level.SUCCESS: messages.success,
+            ImporterMessage.Level.WARNING: messages.warning,
+            ImporterMessage.Level.ERROR: messages.error,
```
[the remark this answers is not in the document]
He3lixxx: Forward to django is necessary when we redirect to a different page after the import is successful. Then, we can't use the custom message rendering that we have in the importer template, but still want to have all messages shown.

I think the custom rendering was required to have correct grouping and sorting of the messages.

evap/staff/importers.py
```diff
@@ -1,691 +1,1281 @@
...
+        )
+        self.existing_names_en: set[str] = set(
+            Course.objects.filter(semester=semester).values_list("name_en", flat=True).distinct()
         )
-        msg += create_user_list_html_string_for_message(students_created + responsibles_created)
-        self.success_messages.append(msg)
+        self.existing_names_tracker: FirstLocationAndCountTracker = FirstLocationAndCountTracker()
+
+        self.names_en_by_name_de: Dict[str, str] = defaultdict(set)
+        self.name_de_mismatch_tracker: FirstLocationAndCountTracker = FirstLocationAndCountTracker()
```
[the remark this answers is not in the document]
He3lixxx: CourseDataMismatchChecker checks for mismatches between the course data given for same course. The course is identified by its name_en ("Cooking|Kochen" and "Cooking|Backen")

This checks for the same name_de being used for two different courses ("Cooking|Backen" and "Baking|Backen").

evap/staff/importers.py
```diff
@@ -1,691 +1,1281 @@
...
-        new_participants = []
-        for user_data in self.users.values():
+        def finalize(self):
+            self.user_data_checker.finalize()
+
+    class CourseDataAdapter(RowCheckerMixin):
+        """Adapter to use CourseDataChecker with EnrollmentInputRow"""
+
+        def __init__(self, course_data_checker):
+            self.course_data_checker: Checker = course_data_checker
```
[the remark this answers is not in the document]
He3lixxx: We are going to add more Checkers that work on CourseData. The intention here was to keep this independent on whatever checker for course data is used. But I guess we're going to remove this adapter anyway.

evap/staff/importers.py
```diff
@@ -1,691 +1,1281 @@
...
-            is_graded=data[5],
-            responsible_email=responsible_data.email,
+class UserDataValidationChecker(Checker):
+    """Run django validation against UserData instances"""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Only give one error per unique user_data.
+        self.already_checked: Set[UserData] = set()
```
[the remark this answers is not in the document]
He3lixxx: I don't know how much latency a `.validate()` call on a model instance adds (I guess it causes DB queries?), so I intended to not use a tracker here with a "and X more places" semantics, I really only wanted to have one validate per user profile.

evap/staff/importers.py
```diff
@@ -1,691 +1,1281 @@
...
+                category=ImporterMessage.Category.USER,
+            )
+
+
+class DuplicateUserDataChecker(Checker):
+    """Check for duplicate users"""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.first_location_by_user_data: Dict[UserData, Location] = {}
```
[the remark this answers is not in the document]
He3lixxx: This uses a tracker for the error locations -- however, we still need to have the first location for each user available, independently on any errors. I think it's fine the way it is?

He3lixxx:
@karyon @niklasmohrin all the stuff suggested that I marked as resolved is now incorporated.

The structuring of components across files is still open, there's one TODO in the code for that.

IDK. If you want to review again, you can do that. I can also try to move stuff around files first if you prefer that.

It's not much less code than before, mainly because the distinction between "PartialCourseData", which might be incomplete, and "CourseData" which is guaranteed to be complete, forces a translation step between them. I couldn't bring myself to store invalid values without any reminders/guarantees in CourseData though. I think the current approach allows humans to reason about correctness that wouldn't be given otherwise.

evap/staff/importers/base.py
```diff
@@ -0,0 +1,283 @@
...
+    """Just a fancy wrapper around a collection of messages with some utility functions"""
+
+    def __init__(self):
+        self.messages: List[ImporterLogEntry] = []
+
+    def __repr__(self):
+        return f"{type(self).__name__}({self.messages})"
+
+    def __eq__(self, other):
+        return self.messages == other.messages
```
He3lixxx, line 62: remove, never used (was intended for tests)

evap/staff/importers/base.py
```diff
@@ -0,0 +1,283 @@
...
+                category=ImporterLogEntry.Category.SCHEMA,
+            ) from e
+
+        rows = []
+        for sheet in book.sheets():
+            if sheet.nrows <= self.skip_first_n_rows:
+                continue
+
+            if sheet.ncols != self.row_cls.column_count:
+                raise ImporterException(
```
He3lixxx, line 202: add a test for these

evap/staff/importers/base.py
```diff
@@ -0,0 +1,283 @@
...
+
+    def finalize(self):
+        # can be overriden by implementations to generate aggregated messages
+        pass
+
+
+class RowCheckerMixin(ABC):
+    @abstractmethod
+    def check_row(self, row):
+        pass
```
He3lixxx, line 278: add `# pragma: no cover`

evap/staff/importers/base.py
```diff
@@ -0,0 +1,283 @@
...
+
+class InputRow(ABC):
+    @property
+    @abstractmethod
+    def column_count(self):
+        pass
+
+    @abstractmethod
+    def __init__(self, location: ExcelFileLocation, cells: Iterable[str]):
+        pass
```
He3lixxx, line 174: add `# pragma: no cover`

evap/staff/importers/base.py
```diff
@@ -0,0 +1,283 @@
...
+
+    def __str__(self):
+        return _('Sheet "{}", row {}').format(self.sheet_name, self.row_number + 1)
+
+
+class InputRow(ABC):
+    @property
+    @abstractmethod
+    def column_count(self):
+        pass
```
He3lixxx, line 170: add `# pragma: no cover`

evap/staff/importers/user.py
```diff
@@ -0,0 +1,384 @@
...
+
+    def check_userdata(self, user_data: UserData, location: ExcelFileLocation):
+        if user_data.email == "":
+            self.importer_log.add_error(
+                _("{location}: Email address is missing.").format(location=location),
+                category=ImporterLogEntry.Category.USER,
+            )
+
+        if user_data.first_name == "":
+            self.importer_log.add_error(
```
He3lixxx, line 128: Add this case to a test.

evap/evaluation/models_logging.py
```diff
@@ -202,32 +202,40 @@ def log_instance_delete(self):
         changes = self._get_change_data(InstanceActionType.DELETE)
         self._update_log(changes, InstanceActionType.DELETE)
 
-    def _update_log(self, changes, action_type: InstanceActionType):
+    def _create_log_entry(self, action_type=InstanceActionType.CREATE):
```
niklasmohrin, line 205: Since this doesn't actually write to `self`, maybe this should not be a method and instead a classmethod that also takes an instance ~_or_ a constructor on `LogEntry`~ (edit: need `thread`)?

niklasmohrin (commented):
(review in progress)

evap/evaluation/models_logging.py
He3lixxx: Didn't really change anything about the logic here.

Thinking about it, I think the current code is wrong, because the initialization of `thread` is done in the class declaration, which afaik will be done exactly once, shared for all threads. Then, all threads would share the same "local" context.

evap/staff/importers/user.py
```diff
@@ -0,0 +1,384 @@
...
+                    first_location=first_location,
+                ),
+                category=ImporterLogEntry.Category.IGNORED,
+            )
+
+
+# TODO: Not a real class that we want to instantiate --> make a set of functions? At least make functions private.
+class UserImporter:
+    class UserDataAdapter(RowCheckerMixin):
+        """Adapter to use UserDataChecker with UserParsedRow"""
```
niklasmohrin, line 307: There is no one `UserDataChecker` (anymore?)

evap/staff/importers/enrollment.py
```diff
@@ -0,0 +1,666 @@
...
+    responsible_email: str
+
+    def equals_when_ignoring_degrees(self, other):
+        return (
+            self.name_de == other.name_de
+            and self.name_en == other.name_en
+            and self.course_type == other.course_type
+            and self.is_graded == other.is_graded
+            and self.responsible_email == other.responsible_email
+        )
```
niklasmohrin, line 45:
```suggestion
        def key(data):
            return (data.name_de, data.name_en, data.course_type, data.is_graded, data.responsible_email)
        return key(self) == key(other)
```

evap/staff/importers/enrollment.py
```diff
@@ -0,0 +1,666 @@
...
+
+@dataclass
+class PartialCourseData:
+    """Same as CourseData, but some fields might be unpopulated because parsing them failed"""
+
+    name_de: str
+    name_en: str
+    degrees_or_none: Set[Degree]
+    course_type_or_none: CourseType
+    is_graded_or_none: bool
```
niklasmohrin, line 56: Hm, I suppose it would be clearer to make the types `Optional` instead of naming them `*_or_none`? (Or maybe do both); for me, I will see / find type hints relatively quickly so the `or_none` is more noise

edit: ah, I see, this is to distinguish it from the fully parsed data; hm I guess we can stick to this name

evap/staff/importers/enrollment.py
```diff
@@ -0,0 +1,666 @@
...
+    due to parsing failure.
+    """
+
+    location: ExcelFileLocation
+
+    student_data: UserData
+    responsible_data: UserData
+    partial_course_data: PartialCourseData
+
+    def get_enrollment_parsed_row(self):
```
niklasmohrin, line 219: I personally don't like the word `get` and would prefer `as`, `into` or so here too :)

niklasmohrin (commented):
I like what I see :+1: (not completely through yet)

evap/staff/importers/user.py
```diff
@@ -0,0 +1,384 @@
...
+    column_count = 4
+
+    location: ExcelFileLocation
+
+    title: str
+    first_name: str
+    last_name: str
+    email: str
+
+    def __init__(self, location: ExcelFileLocation, cells: Sequence[str]):
```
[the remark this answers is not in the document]
He3lixxx: I considered the fact that the attributes and cell values line up to be accidental here, so I decided I did not want to rely on that. I guess this is up to personal taste? Considering that the current method is 4 lines only, and just the declaration of the abstract method would also be 4 lines, I don't see how it could be any shorter.
Would you also be ok with the current code?

evap/staff/importers/user.py
He3lixxx: changed wording to "Checker for UserData" -- or would you rather have the ABC `UserDataChecker` back? I removed it to reduce class count.

niklasmohrin: Nope, now it's good :+1:

evap/staff/importers/user.py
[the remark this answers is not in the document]
niklasmohrin: What really led me to this suggestion was `EnrollmentInputRow.__init__` which assigns 12 cells out of order, which is harder to grasp.

You are right, whether we want to have the dataclass declaration be exactly equivalent to the order of cells is a elegant-or-awful debate. I think that it would be a compact and declarative way of communicating the order, and since the `InputRow` classes don't do much more anyway, it would not be too complicated either.

Of course, one can list the same arguments for the other side and the current version isn't doing any actual harm, so I am not strongly for or against either side, the current code would also be fine for me

niklasmohrin:
I really like the changes, I feel like it really helps when trying to find / understand where something particular is happening without having to backward-slice (swt namedrop, I know) for some importer-wide state modifications :+1:

I haven't proof-read whether the logic is actually correctly mapped to the new code, but I would defer that until final review (that is, until @karyon is through again); and the tests should have already done most of this work for me :^)

evap/evaluation/models_logging.py
He3lixxx: Usage of `thread` is necessary and seems to be correct (contrary to what I first thought, see #1701 for details). Resolving this for now.

evap/staff/importers/enrollment.py
He3lixxx: Made them `Optional[X]`. I'm open for removing the `or_none` part of the name -- I think for developing, it might help people to distinguish where they are (and if this is optional or not), depending on what tools they have available.

evap/staff/importers/user.py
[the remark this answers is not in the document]
He3lixxx: Thought about it again, agreed, applied your version.

He3lixxx:
I've removed the importer classes, they are now methods (`import_users`, `import_enrollments`, `import_persons_from_file`). This resolves the last todo that I still had on my list.

Some discussions in this PR are still unresolved, waiting for input there. Apart from that, waiting for review.

He3lixxx:
@janno42 in https://github.com/e-valuation/EvaP/pull/1703 we replaced xls import with xlsx import. With the architecture here, we might be able to support both, xls and xlsx. Is that desirable, or do we prefer only supporting xlsx?

evap/staff/importers/enrollment.py
```diff
@@ -0,0 +1,652 @@
...
+        assert course_data.equals_when_ignoring_degrees(row.course_data)
+
+        # Not a checker to keep merging and notifying about the merge together.
+        if not row.course_data.degrees.issubset(course_data.degrees):
+            course_data.degrees.update(row.course_data.degrees)
+
+            importer_log.add_warning(
+                _(
+                    "{location}: The course's \"{course_name}\" degree differs from it's degrees in previous rows."
+                    " All degrees have been set for the course."
```
niklasmohrin, line 544: I would merge for this line on its own, on `main` it says "both" which is just factually incorrect

janno42:
I'd say only xlsx is fine. No need to add more complexity to the importer :)

evap/staff/importers/enrollment.py
He3lixxx: (note to self: should be "its". Also the sentence structure is weird: The course's "introduction to X" degree differs [...].)

evap/staff/importers/user.py
```diff
@@ -0,0 +1,374 @@
...
+
+
+class UserDataMismatchChecker(Checker):
+    """Assert UserData matches previous occurrences in the import as well as the database"""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.users: Dict[
+            str, UserData
+        ] = {}  # maps user's mail to UserData instance where it was first seen to have O(1) lookup
```
niklasmohrin, line 139:
```suggestion
        # maps user's mail to UserData instance where it was first seen to have O(1) lookup
        self.users: Dict[str, UserData] = {}
```

evap/staff/importers/user.py
```diff
@@ -0,0 +1,374 @@
...
+                ).format(location=location, user_email=user_email),
+                category=ImporterLogEntry.Category.USER,
+            )
+
+        # Mismatches to database entries
+        user_query = UserProfile.objects.filter(email__in=self.users.keys())
+        db_users_by_email = {db_user.email: db_user for db_user in user_query}
+
+        for user_data in self.users.values():
+            db_user = db_users_by_email.get(user_data.email, None)
```
niklasmohrin, line 167:
```suggestion
            db_user = db_users_by_email.get(user_data.email)
```

niklasmohrin (commented):
haven't read everything (again), good stuff

evap/staff/importers/person.py
```diff
@@ -0,0 +1,100 @@
+from typing import Iterable
+
+from django.utils.html import format_html
+from django.utils.translation import gettext as _
+
+from evap.evaluation.models import Contribution, Evaluation, UserProfile
+from evap.staff.tools import ImportType, create_user_list_html_string_for_message
+
+from .base import ImporterLog
+from .user import import_users
```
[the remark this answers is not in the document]
He3lixxx: I used relative imports in `staff/importers/` so the whole directly can be moved without having to adjust anything, because I considered the file structure an implementation detail not relevant to the rest of the code. Absolute imports should also work fine.

Do you have a good reason for using absolute imports? The absolute path is longer, so I don't really see a benefit.

evap/staff/importers/person.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def add_participants_to(
+    evaluation: Evaluation, users: Iterable[UserProfile], test_run: bool, importer_log: ImporterLog
+):
+    evaluation_participants = evaluation.participants.all()
+    already_related = [user for user in users if user in evaluation_participants]
+    users_to_add = [user for user in users if user not in evaluation_participants]
```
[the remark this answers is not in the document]
He3lixxx: Searching for other places where we could use that method using grep with
```regexp
for.*in.*if.*not.*in
```
gives me no other results. Since we're talking about two lines, I'd argue we should leave it inlined until we'd have at least one more usage.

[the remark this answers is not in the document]
niklasmohrin: Sure

evap/staff/importers/person.py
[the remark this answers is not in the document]
niklasmohrin: One reason for me would be that when I have a `UserImporter` and a method that returns `ImporterLog`, I could not see what to import after going to the `user.py` file. But your point makes sense too, so you can decide

evap/staff/importers/enrollment.py
niklasmohrin: Yes, let's keep them

evap/staff/importers/user.py
```diff
@@ -0,0 +1,384 @@
...
+    def check_userdata(self, user_data: UserData, location: ExcelFileLocation):
+        if user_data.email == "":
+            # UserDataEmptyFieldsChecker will give an error for these, no need to spam additional errors
+            return
+
+        stored_user_data = self.users.setdefault(user_data.email, user_data)
+        if user_data != stored_user_data:
+            self.in_file_mismatch_tracker.add_location_for_key(location, user_data.email)
+
+    def finalize(self):
```
niklasmohrin, line 152: As this method is not trivial, I suggest already today:
```suggestion
    def finalize(self) -> None:
```

evap/staff/importers/user.py
```diff
@@ -0,0 +1,384 @@
...
+                    first_location=first_location,
+                ),
+                category=ImporterLogEntry.Category.IGNORED,
+            )
+
+
+class UserDataAdapter(RowCheckerMixin):
+    """Adapter to use Checkers for UserData with UserParsedRow"""
+
+    def __init__(self, user_data_checker):
```
niklasmohrin, line 309:
```suggestion
    def __init__(self, user_data_checker: Checker):
```

niklasmohrin (commented):
In person we found that functions / methods that dont have any annotations _in the signature_ are not checked; in a follow up, we / I should add `-> None` as a default

evap/staff/importers/base.py
```diff
@@ -0,0 +1,285 @@
...
+    def __init__(self, test_run: bool, importer_log: ImporterLog):
+        self.test_run: bool = test_run
+        self.importer_log: ImporterLog = importer_log
+
+    def finalize(self):
+        # can be overriden by implementations to generate aggregated messages
+        pass
+
+
+class RowCheckerMixin(ABC):
```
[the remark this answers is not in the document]
He3lixxx: Added the missing finalize, added `-> None`, but haven't added annotations to `row` and `rows`. I think that's fine for now(?)

evap/staff/importers/user.py
He3lixxx: Doesn't work with the current code base: `Checker` only defined `__init__` and `finalize`, so we can have checkers for `UserData`, `CourseData`, different kinds of rows, ...

We could add a protocol for `UserDataChecker`, but then we'd probably have to add protocols for all other cases, too?

niklasmohrin: Hm, both protocols and generics _could_ make this work, but we need to draw the line somewhere and I am fine with doing that here - otherwise I have to open PRs at mypy

evap/staff/importers/base.py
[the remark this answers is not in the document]
niklasmohrin: Yeah, its fine

evap/staff/importers/user.py
```diff
@@ -0,0 +1,383 @@
...
+            checker.check_rows(rows)
+        importer_log.raise_if_has_errors()
+
+        users = [row.user_data for row in rows]
+
+        # Both will contain the updated data from the UserData instances, but the existing users are not yet saved.
+        existing_user_profiles, new_user_profiles = get_user_profile_objects(users)
+        resulting_user_profiles = existing_user_profiles + new_user_profiles
+
+        importer_log.raise_if_has_errors()
```
niklasmohrin, line 344: I think the log is not touched since the last `raise_if_has_errors`

evap/staff/importers/user.py
```diff
@@ -0,0 +1,383 @@
...
+        if test_run:
+            importer_log.add_success(_("The test run showed no errors. No data was imported yet."))
+            msg = format_html(_("The import run will create {} users:"), len(new_user_profiles))
+            msg += create_user_list_html_string_for_message(new_user_profiles)
```
niklasmohrin, line 348: How do you feel about:
```suggestion
            msg = format_html(
                _("The import run will create {} users:{}"),
                len(new_user_profiles),
                create_user_list_html_string_for_message(new_user_profiles),
            )
```
Its slightly longer, but doesn't involve mutation of `msg` (same for the `else` case)

evap/staff/importers/user.py
```diff
@@ -0,0 +1,383 @@
...
+
+    def check_row(self, row: UserParsedRow):
+        self.user_data_checker.check_userdata(row.user_data, row.location)
+
+    def finalize(self) -> None:
+        self.user_data_checker.finalize()
+
+
+@transaction.atomic
+def import_users(excel_content, test_run: bool) -> Tuple[Iterable[UserProfile], ImporterLog]:
```
niklasmohrin, line 319: Secondly: `excel_content` is implicitly `typing.Any`
```suggestion
def import_users(excel_content: bytes, test_run: bool) -> Tuple[Iterable[UserProfile], ImporterLog]:
```

evap/staff/importers/user.py
```diff
@@ -0,0 +1,383 @@
...
+            operator.or_,
+            (
+                Q(first_name=user.first_name) & Q(last_name=user.last_name) & ~Q(email=user.email)
+                for user in self.users.values()
+            ),
+            Q(pk=None),  # always false Q as initializer, required for empty self.users
+        )
+        db_users_by_name = defaultdict(list)
+        for db_user in UserProfile.objects.filter(same_name_filter):
+            db_users_by_name[(db_user.first_name, db_user.last_name)].append(db_user)
```
niklasmohrin, line 191: Isn't this `unordered_groupby` again?

niklasmohrin (commented):
Looked over `user.py`, seems good, some notes:

evap/staff/importers/enrollment.py
```diff
@@ -0,0 +1,764 @@
...
+
+        try:
+            degrees = self.degree_mapper.degree_set_from_import_string(row.evaluation_degree_names)
+        except DegreeImportMapper.InvalidDegreeNamesException as e:
+            degrees = invalid_value
+            for invalid_degree in e.invalid_degree_names:
+                self.invalid_degrees_tracker.add_location_for_key(row.location, invalid_degree)
+
+        try:
+            course_type: Union[InvalidValue, CourseType] = self.course_type_mapper.course_type_from_import_string(
```
niklasmohrin, line 234: This is only needed because of the assignment below, how do you feel about declaring the three variables that can be invalid just before the first `try` (this is also shorter):

```python
degrees: ...
course_type: Union[CourseType, InvalidValue]
is_graded: Union[bool, InvalidValue]
```

evap/staff/importers/enrollment.py
```diff
@@ -0,0 +1,764 @@
...
+
+@dataclass
+class CourseData:
+    """Holds information about a course, retrieved from an import file."""
+
+    name_de: str
+    name_en: str
+    degrees: Union[Set[Degree], InvalidValue]
+    course_type: Union[CourseType, InvalidValue]
+    is_graded: Union[bool, InvalidValue]
```
niklasmohrin, line 51: This can be a bit nicer, using
```python
T = TypeVar("T")
MaybeInvalid = Union[T, InvalidValue]
```

and then `MaybeInvalid[CourseType]` and such

evap/staff/importers/enrollment.py
```diff
@@ -0,0 +1,764 @@
...
+            is_graded = invalid_value
+            self.invalid_is_graded_tracker.add_location_for_key(row.location, e.invalid_is_graded)
+
+        course_data = CourseData(
+            name_de=row.evaluation_name_de.strip(),
+            name_en=row.evaluation_name_en.strip(),
+            degrees=degrees,
+            course_type=course_type,
+            is_graded=is_graded,
+            responsible_email=row.responsible_email,
```
niklasmohrin, line 255: why are the names stripped, but this is not?

evap/staff/importers/enrollment.py
```diff
@@ -0,0 +1,764 @@
...
+        self.courses_by_name_en = {course.name_en: course for course in courses}
+
+    @staticmethod
+    def get_merge_hindrances(course_data: CourseData, merge_candidate: Course):
+        hindrances = []
+        if merge_candidate.type != course_data.course_type:
+            hindrances.append(_("the course type does not match"))
+
+        responsibles = merge_candidate.responsibles.all()
+        if len(responsibles) != 1 or responsibles.get().email != course_data.responsible_email:
```
niklasmohrin, line 332: Does `.get` here make a new query? If so, let's make it `[0]`

evap/staff/importers/enrollment.py
```diff
@@ -0,0 +1,764 @@
...
+        course_with_same_name_de = self.courses_by_name_de.get(course_data.name_de, None)
+
+        if course_with_same_name_en is None and course_with_same_name_de is None:
+            # No name matches --> no merging, all is fine
+            course_data.merge_into_course = None
+            return
+
+        if course_with_same_name_en != course_with_same_name_de:
+            if course_with_same_name_en is not None:
+                # Course with same name_en, but different name_de exists
```
niklasmohrin, line 353: Should these comments be moved to be doc comments in the exception classes?

niklasmohrin (commented):
Now read until the `CourseNameChecker` in `enrollment.py`

evap/staff/importers/enrollment.py
```diff
@@ -0,0 +1,764 @@
...
+    def check_row(self, row: EnrollmentParsedRow) -> None:
+        self.course_data_checker.check_course_data(row.course_data, row.location)
+
+    def finalize(self) -> None:
+        self.course_data_checker.finalize()
+
+
+@transaction.atomic
+def import_enrollments(
+    excel_content, semester: Semester, vote_start_datetime, vote_end_date, test_run: bool
```
niklasmohrin, line 556: what is holding us back from the following?
```python
def import_enrollments(
    excel_content: bytes,
    semester: Semester,
    vote_start_datetime: Optional[datetime],
    vote_end_date: Optional[date],
    test_run: bool,
) -> ImporterLog:
```

accordingly, `update_existing_and_create_new_courses` should be annotated with non-optional versions

evap/staff/importers/enrollment.py
```diff
@@ -0,0 +1,764 @@
...
+) -> ImporterLog:
+    # pylint: disable=too-many-locals
+    importer_log = ImporterLog()
+
+    with ConvertExceptionsToMessages(importer_log):
+        input_rows = ExcelFileRowMapper(skip_first_n_rows=1, row_cls=EnrollmentInputRow, importer_log=importer_log).map(
```
niklasmohrin, line 562: adding a comma after `importer_log` formats this longer, but less awkward in my opinion - your call :)

evap/staff/importers/enrollment.py
```diff
@@ -0,0 +1,764 @@
...
+                new_participants_count,
+                new_responsibles_count,
+            )
+            msg += create_user_list_html_string_for_message(new_user_profiles)
+            importer_log.add_success(msg)
+
+    return importer_log
+
+
+def normalize_rows(enrollment_rows: Iterable[EnrollmentParsedRow], importer_log: ImporterLog):
```
niklasmohrin, line 615: this is missing a return type

evap/staff/importers/enrollment.py
```diff
@@ -0,0 +1,764 @@
...
+
+        importer_log.raise_if_has_errors()
+        if test_run:
+            importer_log.add_success(_("The test run showed no errors. No data was imported yet."))
+            msg = format_html(
+                _("The import run will create {evaluation_count} courses/evaluations and {user_count} users:"),
+                evaluation_count=new_course_count,
+                user_count=len(new_user_profiles),
+            )
+            msg += create_user_list_html_string_for_message(new_user_profiles)
```
niklasmohrin, line 596: I am slightly in favor of using the result of `create_user_list_html_string_for_message` in the `format_html` itself instead of appending, but you can decide

niklasmohrin (commented):
done with enrollment

niklasmohrin:
reading `import_enrollments`, I felt like these abstractions we now have really want to be composed in a more abstract fashion and I thought about something like sklearn's pipeline:
```python
with ConvertExceptionsToMessages(importer_log):
    pipeline = Pipeline(
        ExcelFileRowMapper(skip_first_n_rows=1, row_cls=EnrollmentInputRow),
        RaiseIfErrors(),
        EnrollmentInputRowMapper(),
        TooManyEnrollmentsChecker(test_run),
        CourseDataAdapter(CourseNameChecker(test_run, semester=semester)),
        CourseDataAdapter(CourseDataMismatchChecker(test_run)),
        UserDataAdapter(UserDataEmptyFieldsChecker(test_run)),
        UserDataAdapter(UserDataMismatchChecker(test_run)),
        UserDataAdapter(UserDataValidationChecker(test_run)),
        RaiseIfErrors(),
        importer_log=importer_log,
    )
    parsed_rows = pipeline.run(excel_content)
```

This will however require some more (typing?) effort without much "business value" - maybe it will be interesting to look into at some point though :^)

evap/staff/importers/enrollment.py
```diff
@@ -0,0 +1,764 @@
...
+
+    def equals_when_ignoring_degrees(self, other) -> bool:
+        def key(data):
+            return (data.name_de, data.name_en, data.course_type, data.is_graded, data.responsible_email)
+
+        return key(self) == key(other)
+
+
+class ValidCourseData(CourseData):
+    """Typing: CourseData instance where no element is invalid_value"""
```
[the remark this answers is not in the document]
He3lixxx: Yes, but a missing field is by definition `invalid_value` (we don't allow `None`, by choice)

evap/staff/importers/enrollment.py
He3lixxx: ¯\_(ツ)_/¯

He3lixxx: Added a `.strip()` call here, too. @janno42 any reasons why we wouldn't want to do that?

evap/staff/importers/enrollment.py
```diff
@@ -0,0 +1,764 @@
...
+                _(
+                    '{location}: Course "{course_name}" (EN) already exists in this semester with a different german name.'
+                ).format(
+                    location=location_string,
+                    course_name=name_en,
+                ),
+                category=ImporterLogEntry.Category.COURSE,
+            )
+
+        for key, location_string in self.name_de_mismatch_tracker.aggregated_keys_and_location_strings():
```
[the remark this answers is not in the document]
He3lixxx: Changed the loop header to
```python3
        for (name_en, name_de), location_string in self.name_de_mismatch_tracker.aggregated_keys_and_location_strings():
```

Does that work for you? Any more chars require a linebreak.

evap/staff/importers/enrollment.py
```diff
@@ -0,0 +1,764 @@
...
+        if not all_fields_valid(course_data):
+            return
+
+        stored = self.course_data_by_name_en.setdefault(course_data.name_en, course_data)
+
+        # degrees will be merged together
+        if course_data.equals_when_ignoring_degrees(stored):
+            return
+
+        self.tracker.add_location_for_key(location, course_data.name_en)
```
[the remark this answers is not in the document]
He3lixxx: I slightly favor the other way, because we have 2 conditions here that allow a course to "pass".

With that schema (multiple passes, a single failure), I like it more when we use early returns. See https://github.com/e-valuation/EvaP/blob/main/evap/evaluation/models.py#L1611-L1620

evap/staff/importers/enrollment.py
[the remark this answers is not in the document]
niklasmohrin: Sounds good, we could even go for `(name_en, __)`, but both are fine :)

He3lixxx:
> reading `import_enrollments`, I felt like these abstractions we now have really want to be composed in a more abstract fashion and I thought about something like sklearn's pipeline:
> 
> ```python
> with ConvertExceptionsToMessages(importer_log):
>     pipeline = Pipeline(
>         ExcelFileRowMapper(skip_first_n_rows=1, row_cls=EnrollmentInputRow),
>         RaiseIfErrors(),
>         EnrollmentInputRowMapper(),
>         TooManyEnrollmentsChecker(test_run),
>         CourseDataAdapter(CourseNameChecker(test_run, semester=semester)),
>         CourseDataAdapter(CourseDataMismatchChecker(test_run)),
>         UserDataAdapter(UserDataEmptyFieldsChecker(test_run)),
>         UserDataAdapter(UserDataMismatchChecker(test_run)),
>         UserDataAdapter(UserDataValidationChecker(test_run)),
>         RaiseIfErrors(),
>         importer_log=importer_log,
>     )
>     parsed_rows = pipeline.run(excel_content)
> ```
> 
> This will however require some more (typing?) effort without much "business value" - maybe it will be interesting to look into at some point though :^)

Created #1795 for this.

evap/staff/importers/enrollment.py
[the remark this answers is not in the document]
niklasmohrin: ah, I get it now; I was wondering why `name_en` etc. is missing, but those are inherited from `CourseData`, didn't see that

niklasmohrin (approved):
:tada:

PR #1637, opened 2021-10-18
Prevent text answer updates after publishing evaluation

### LINKED ISSUE #1623: Prevent text answer updates after publishing evaluation
After an evaluation was published, changes to text answers should not be possible anymore.
The staff text answer page of an evaluation (`staff.views.evaluation_textanswers`) should then not be accessible (the UI doesn't link to it but manually entering the URL is currently possible) and `evaluation_textanswers_update_publish` as well as `evaluation_textanswer_edit` should also raise `PermissionDenied`.

rgwohlbold:
Fixes #1623


He3lixxx (commented):
Could you add three short tests?

evap/staff/tests/test_views.py
```diff
@@ -2255,6 +2255,18 @@ def test_num_queries_is_constant(self):
             with self.assertNumQueries(FuzzyInt(0, self.num_questions)):
                 self.app.get(self.url, user=self.manager)
 
+    def test_published(self):
+        let_user_vote_for_evaluation(self.app, self.student2, self.evaluation)
+        Evaluation.objects.filter(id=self.evaluation.id).update(state=Evaluation.State.PUBLISHED)
+        with run_in_staff_mode(self):
+            self.app.get(self.url, user=self.manager, status=403)
```
He3lixxx, line 2262: If possible, I'd add the assertion that before the `update()`, we got a 200 response, so that we can be sure the 403 is caused by what we changed here (same for the other tests).

rgwohlbold: That makes sense. I added in the extra assertions :+1:

PR #1639, opened 2021-10-18
Participants editable by editors

### LINKED ISSUE #1569: Make participants editable by editors
Editors are currently not able to change participants of an evaluation and have to send a message asking for changes instead. This results in a lot of work for the evaluation team. Since recently, all changes are logged, so editors can be allowed to change the participants themselves, and staff users can see what changes have been made.

This should be done as follows;
- The list of participants should become an editable form field just like in the staff form and should be placed at the same position as in the staff form instead of at the end of the page. 
- The field (both for editors and staff) should be changed to contain exactly one person per line (by changing the select2 elements' CSS).
- If the log contains participant changes done by users who are not staff and the evaluation was not yet approved by a staff user, the participant field in the staff form should show a warning (e.g., by having a yellow border).

dieknolle3333:
fixes #1569 (hopefully)

### DIFF
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -5,7 +5,7 @@ from django.db.models import Q
 from django.forms.widgets import CheckboxSelectMultiple
 from django.utils.translation import gettext_lazy as _
 
-from evap.evaluation.forms import UserModelChoiceField
+from evap.evaluation.forms import UserModelChoiceField, UserModelMultipleChoiceField
 from evap.evaluation.models import Course, Evaluation, Questionnaire, UserProfile
 from evap.evaluation.tools import vote_end_datetime
 from evap.staff.forms import ContributionForm
@@ -26,9 +26,13 @@ class EvaluationForm(forms.ModelForm):
             "name_en_field",
             "vote_start_datetime",
             "vote_end_date",
+            "participants",
             "general_questionnaires",
             "course",
         )
+        field_classes = {
+            "participants": UserModelMultipleChoiceField,
+        }
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -44,6 +48,8 @@ class EvaluationForm(forms.ModelForm):
 
         self.fields["vote_start_datetime"].localize = True
         self.fields["vote_end_date"].localize = True
+
+        self.fields["participants"].queryset = UserProfile.objects.exclude(is_active=False)
 
         if self.instance.general_contribution:
             self.fields["general_questionnaires"].initial = [
@@ -72,6 +78,9 @@ class EvaluationForm(forms.ModelForm):
             raise forms.ValidationError(_("The last day of evaluation must be in the future."))
 
         return vote_end_date
+
+    def clean_participants(self):
+        return self.cleaned_data.get("participants")
 
     def clean_general_questionnaires(self):
         # Ensure all locked questionnaires still have the same status (included or not)
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -33,6 +33,9 @@ class EvaluationForm(forms.ModelForm):
         field_classes = {
             "participants": UserModelMultipleChoiceField,
         }
+        widgets = {
+            "participants": forms.SelectMultiple(attrs={"data-selection-css-class": "participants_multi_select"}),
+        }
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -42,8 +45,8 @@ class EvaluationForm(forms.ModelForm):
 
         self.fields["general_questionnaires"].queryset = (
             Questionnaire.objects.general_questionnaires()
-            .filter(Q(visibility=Questionnaire.Visibility.EDITORS) | Q(contributions__evaluation=self.instance))
-            .distinct()
+                .filter(Q(visibility=Questionnaire.Visibility.EDITORS) | Q(contributions__evaluation=self.instance))
+                .distinct()
         )
 
         self.fields["vote_start_datetime"].localize = True
@@ -78,9 +81,6 @@ class EvaluationForm(forms.ModelForm):
             raise forms.ValidationError(_("The last day of evaluation must be in the future."))
 
         return vote_end_date
-
-    def clean_participants(self):
-        return self.cleaned_data.get("participants")
 
     def clean_general_questionnaires(self):
         # Ensure all locked questionnaires still have the same status (included or not)
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -52,7 +52,8 @@ class EvaluationForm(forms.ModelForm):
         self.fields["vote_start_datetime"].localize = True
         self.fields["vote_end_date"].localize = True
 
-        self.fields["participants"].queryset = UserProfile.objects.exclude(is_active=False)
+        self.fields["participants"].queryset = (
+                UserProfile.objects.exclude(is_active=False) | self.instance.participants.all()).distinct()
 
         if self.instance.general_contribution:
             self.fields["general_questionnaires"].initial = [
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -57,7 +57,8 @@ class EvaluationForm(forms.ModelForm):
         else:
             prior_participants = self.instance.participants.all()
         self.fields["participants"].queryset = (
-                UserProfile.objects.exclude(is_active=False) | prior_participants).distinct()
+            UserProfile.objects.exclude(is_active=False) | prior_participants
+        ).distinct()
 
         if self.instance.general_contribution:
             self.fields["general_questionnaires"].initial = [
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -52,13 +52,10 @@ class EvaluationForm(forms.ModelForm):
         self.fields["vote_start_datetime"].localize = True
         self.fields["vote_end_date"].localize = True
 
-        if self.instance.pk is None:
-            prior_participants = UserProfile.objects.filter(pk=None)
-        else:
-            prior_participants = self.instance.participants.all()
-        self.fields["participants"].queryset = (
-            UserProfile.objects.exclude(is_active=False) | prior_participants
-        ).distinct()
+        queryset = UserProfile.objects.exclude(is_active=False)
+        if self.instance.pk is not None:
+            queryset = (queryset | self.instance.participants.all()).distinct()
+        self.fields["participants"].queryset = queryset
 
         if self.instance.general_contribution:
             self.fields["general_questionnaires"].initial = [
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -34,7 +34,7 @@ class EvaluationForm(forms.ModelForm):
             "participants": UserModelMultipleChoiceField,
         }
         widgets = {
-            "participants": forms.SelectMultiple(attrs={"data-selection-css-class": "participants_multi_select"}),
+            "participants": forms.SelectMultiple(attrs={"data-selection-css-class": "user-multi-select"}),
         }
 
     def __init__(self, *args, **kwargs):
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -33,9 +33,6 @@ class EvaluationForm(forms.ModelForm):
         field_classes = {
             "participants": UserModelMultipleChoiceField,
         }
-        # widgets = {
-        #   "participants": forms.SelectMultiple(attrs={"data-selection-css-class": "user-multi-select"}),
-        # }
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -45,8 +45,8 @@ class EvaluationForm(forms.ModelForm):
 
         self.fields["general_questionnaires"].queryset = (
             Questionnaire.objects.general_questionnaires()
-                .filter(Q(visibility=Questionnaire.Visibility.EDITORS) | Q(contributions__evaluation=self.instance))
-                .distinct()
+            .filter(Q(visibility=Questionnaire.Visibility.EDITORS) | Q(contributions__evaluation=self.instance))
+            .distinct()
         )
 
         self.fields["vote_start_datetime"].localize = True
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -33,9 +33,9 @@ class EvaluationForm(forms.ModelForm):
         field_classes = {
             "participants": UserModelMultipleChoiceField,
         }
-        widgets = {
-            "participants": forms.SelectMultiple(attrs={"data-selection-css-class": "user-multi-select"}),
-        }
+        #widgets = {
+         #   "participants": forms.SelectMultiple(attrs={"data-selection-css-class": "user-multi-select"}),
+        #}
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -33,9 +33,9 @@ class EvaluationForm(forms.ModelForm):
         field_classes = {
             "participants": UserModelMultipleChoiceField,
         }
-        #widgets = {
-         #   "participants": forms.SelectMultiple(attrs={"data-selection-css-class": "user-multi-select"}),
-        #}
+        # widgets = {
+        #   "participants": forms.SelectMultiple(attrs={"data-selection-css-class": "user-multi-select"}),
+        # }
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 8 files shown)


evap/static/scss/_adjustments.scss
```diff
@@ -27,6 +27,7 @@ body, .fixed-top, .fixed-bottom, .is-fixed, .sticky-top {
 // select2 adjustments
 .select2-container--default .select2-selection--single {
     height: 38px;
+    width: 94%;
```
He3lixxx, line 30: Why did you add this? This is for a single selection select2 field, right?

evap/static/scss/_adjustments.scss
```diff
@@ -46,6 +47,12 @@ body, .fixed-top, .fixed-bottom, .is-fixed, .sticky-top {
     white-space: normal;
 }
 
+.select2-container--default .select2-selection--multiple .select2-selection__choice{
+    float: unset;
+    max-width: 90%;
+    width: fit-content;
+}
```
He3lixxx, line 54: I think this affect all select2 multi-selects, right? @janno42 is this what we want?

janno42: we should only change the user select fields in this way

evap/contributor/forms.py
```diff
@@ -26,9 +26,16 @@ class Meta:
...
             "vote_end_date",
+            "participants",
             "general_questionnaires",
             "course",
         )
+        field_classes = {
+            "participants": UserModelMultipleChoiceField,
+        }
+        widgets = {
+            "participants": forms.SelectMultiple(attrs={"data-selection-css-class": "participants_multi_select"}),
```
janno42, line 37: please rename the css class to `user-multi-select`

janno42 (changes requested):
the user select field for me looks like this:
image

i assume it looks better for you - have you tested it on the current main branch and did you update the bootstrap submodule and the npm sass installation?

evap/static/scss/_adjustments.scss
```diff
@@ -140,6 +140,14 @@ $alert-colors: (
...
 }
 
+.user-multi-select {
+    padding-right: 36px !important;
+}
+
+.user-multi-select li {
+    width: 100%;
+}
+
```
janno42, line 150: since you're at it: please set a lighter background color like this:
```
.user-multi-select.select2-selection--multiple .select2-selection__choice {
    background-color: $lighter-gray;
}
```

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

evap/static/scss/_adjustments.scss
```diff
@@ -140,6 +140,18 @@ $alert-colors: (
     white-space: normal;
 }
 
+.user-multi-select {
+    padding-right: 36px !important;
```
niklasmohrin, line 144: @janno42 Should this be converted to another unit or is it fine like this? Note that there is one rule with `38px` a couple of lines above

niklasmohrin (approved):
I haven't really checked whether everything that needs to be deleted actually is deleted, but the diff looks good at least :)

evap/contributor/forms.py
```diff
@@ -45,6 +52,8 @@ def __init__(self, *args, **kwargs):
         self.fields["vote_start_datetime"].localize = True
         self.fields["vote_end_date"].localize = True
 
+        self.fields["participants"].queryset = UserProfile.objects.exclude(is_active=False)
```
He3lixxx, line 55: @janno42 This will give the usual issues with inactive participants not being listed and then removed on save. I guess that is not expected?

evap/contributor/tests/test_forms.py
```diff
@@ -21,6 +21,24 @@ def test_fields_disabled_when_editors_disallowed_to_edit(self):
...
+        form = EvaluationForm(form_data, instance=evaluation)
+        self.assertEqual(len(form["participants"].initial), 1)
+
+        form_data["participants"].remove(student.pk)
+        EvaluationForm(form_data, instance=evaluation).save()
+        self.assertEqual(evaluation.num_participants, 0)
+
+        form_data["participants"].append(student.pk)
+        EvaluationForm(form_data, instance=evaluation).save()
+        del evaluation.num_participants  # discard cached property
```
He3lixxx, line 39:
```suggestion
        del evaluation.num_participants
```

evap/contributor/tests/test_views.py
```diff
@@ -245,8 +245,6 @@ def test_contact_modal_escape(self):
         self.evaluation.save()
         page = self.app.get(self.url, user=self.responsible, status=200)
 
-        self.assertIn("changeParticipantRequestModalLabel", page)
-
```
He3lixxx, line 249: The changeEvaluationRequestModal still contains the evaluation title, so I think this should be replaced.

He3lixxx (changes requested):

evap/static/scss/_adjustments.scss
janno42: does the delete button have a fixed size or does it change when changing text sizes? if the button size is fixed, we can keep the pixel value.

evap/contributor/forms.py
janno42: right, the queryset should also include all already selected users

evap/static/scss/_adjustments.scss
dieknolle3333: From what I found out zooming in and out in the browser, the width of the button doesn't really change (it grows and shrinks within an interval of less than 0.05px but not proportional to the browser's zoom). It's probably because the box containing all the user multi selects doesn't change its size with zooming in or out (at least in Chromium and Firefox).
Both the height and the padding-right are hard-coded in select2.min.css. 
So even if the outer box did change its size, I would expect the delete button to still be in the middle of the border of the outer box and the user multi selects.
Is that good enough? :sweat_smile:

janno42: when changing the `font-size` of the document (changing the actual value, not zooming in the browser), the fixed width isn't correct anymore (see below). but as other things would break as well (e.g., the individual removal buttons) i don't mind keeping it like it is (and i hope we'll get rid of select2 at some point anyway).

image

janno42 (changes requested):
I just remembered one thing that needs to be done in this PR as well (sorry).
Similar to the button requesting a new account in the contributor section, we need the same button (opening the same modal) also for participants, in case editors want to add someone who doesn't have an account yet.

dieknolle3333:
> I just remembered one thing that needs to be done in this PR as well (sorry). Similar to the button requesting a new account in the contributor section, we need the same button (opening the same modal) also for participants, in case editors want to add someone who doesn't have an account yet.

Is the button only needed when editors are allowed to edit?

janno42:
The button should only be shown when editors are allowed to edit. It should optimally be placed next to the participant field but I assume this isn't easy. So it is also okay to place it where the request changes button would be if editors aren't allowed to edit.

dieknolle3333:
> The button should only be shown when editors are allowed to edit. It should optimally be placed next to the participant field but I assume this isn't easy. So it is also okay to place it where the request changes button would be if editors aren't allowed to edit.

Screenshot from 2022-01-31 22-08-52
Should this be styled differently, e.g. with a bigger margin?

janno42:
yes, please :)
the following form field should have at least the same distance to the button as the button has to the top of the card.

dieknolle3333:
> yes, please :) the following form field should have at least the same distance to the button as the button has to the top of the card.

(Sorry for deleting my previous question. After I found the solution it just looked so much better that I didn't think I needed to ask it anymore, and I deleted it before seeing your answer, however I managed to do that.) Anyway, I hope it looks fine now!
Screenshot from 2022-01-31 22-57-13

evap/contributor/templates/contributor_evaluation_form.html
```diff
@@ -63,7 +63,15 @@ <h5 class="card-title">{% trans 'Course data' %}</h5>
             <div class="card-body">
                 <div class="d-flex">
                     <h5 class="card-title me-auto">{% trans 'Evaluation data' %}</h5>
-                    {% if not evaluation.allow_editors_to_edit %}
+                    {% if evaluation.allow_editors_to_edit %}
+                        {% if not manager %}
```
janno42, line 67: you can skip this check. it is the contributor form and not the staff form, so managers won't use it :)

evap/contributor/forms.py
```diff
@@ -45,6 +52,14 @@ def __init__(self, *args, **kwargs):
         self.fields["vote_start_datetime"].localize = True
         self.fields["vote_end_date"].localize = True
 
+        if self.instance.pk is None:
+            prior_participants = UserProfile.objects.filter(pk=None)
+        else:
+            prior_participants = self.instance.participants.all()
+        self.fields["participants"].queryset = (
+            UserProfile.objects.exclude(is_active=False) | prior_participants
+        ).distinct()
```
janno42, line 61: You could do something like this instead:
```
queryset = UserProfile.objects.exclude(is_active=False)
if self.instance.pk is not None:
    queryset = (queryset | self.instance.participants.all()).distinct()
self.fields["participants"].queryset = queryset
```

evap/contributor/templates/contributor_evaluation_form.html
```diff
@@ -147,10 +135,6 @@ <h5 class="modal-title" id="previewModalLabel">{% trans 'Preview' %}</h5>
         };
     </script>
 
-    {% blocktrans asvar title with evaluation_name=evaluation.full_name|safe %}Request participant changes for {{ evaluation_name }}{% endblocktrans %}
-    {% trans 'Please tell us what changes to the participants list we should make.' as teaser %}
-    {% include 'contact_modal.html' with modal_id='changeParticipantRequestModal' user=request.user title=title teaser=teaser %}
-
     {% trans 'Request account creation' as title %}
```
janno42, line 138: in case managers are asked to add the newly created accounts to the evaluation, this gives them the required info where they belong:
```suggestion
    {% blocktrans asvar title with evaluation_name=evaluation.full_name|safe %}Request account creation for {{ evaluation_name }}{% endblocktrans %}
```

evap/staff/forms.py
```diff
@@ -382,7 +385,13 @@ def __init__(self, *args, **kwargs):
...
         )
 
-        self.fields["participants"].queryset = UserProfile.objects.exclude(is_active=False)
+        if self.instance.pk is None:
+            prior_participants = UserProfile.objects.filter(pk=None)
+        else:
+            prior_participants = self.instance.participants.all()
+        self.fields["participants"].queryset = (
+            UserProfile.objects.exclude(is_active=False) | prior_participants
+        ).distinct()
```
janno42, line 394: see contributor form

janno42 (commented):
nice, almost done =)

evap/contributor/tests/test_views.py
```diff
@@ -259,3 +259,11 @@ def test_information_message(self):
...
             "Once everything is okay, please approve the evaluation on the bottom of the page.",
         )
+
+    def test_display_request_buttons(self):
+        self.evaluation.name_en = "Adam & Eve"
+        self.evaluation.allow_editors_to_edit = False
+        self.evaluation.save()
+        page = self.app.get(self.url, user=self.responsible)
+        self.assertIn("Request changes", page)
+        self.assertNotIn("Request creation of new account", page)
```
dieknolle3333, line 269: With this (unfinished) test I wanted to make sure that only one button is shown, in this case, that the "Request changes" button is visible while the "Request creation of new account" button is not. The problem is that in this view, the "Request creation of new account" button exists for contributors as well, so the string is always present in the page.
That's why I think I need to parse HTML in order to test whether the button is in the right location. 
Is there a way to do that already? 
If no, can we maybe add a dependency that allows me to parse HTML in tests? @niklasmohrin expressed support for adding beautiful soup, arguing that it would also be useful in other cases.

He3lixxx: You could assert that "Request creation of new account" is on the page twice when `allow_editors_to_edit` is True and only once when `allow_editors_to_edit` is False. We do something similar in [`test_textanswer_review_state_tags`](https://github.com/e-valuation/EvaP/blob/main/evap/staff/tests/test_views.py#L589)

evap/contributor/tests/test_forms.py
```diff
@@ -21,6 +21,24 @@ def test_fields_disabled_when_editors_disallowed_to_edit(self):
         form = EvaluationForm(instance=evaluation)
         self.assertTrue(all(form.fields[field].disabled for field in form.fields))
 
+    def test_edit_participants_in_form(self):
```
He3lixxx, line 24:
```suggestion
    def test_edit_participants(self):
```

evap/contributor/tests/test_views.py
```diff
@@ -259,3 +259,11 @@ def test_information_message(self):
             "Please review the evaluation's details below, add all contributors and select suitable questionnaires. "
             "Once everything is okay, please approve the evaluation on the bottom of the page.",
         )
+
+    def test_display_request_buttons(self):
+        self.evaluation.name_en = "Adam & Eve"
```
He3lixxx, line 264: I think the name is irrelevant for this test?
```suggestion
```

evap/staff/tests/test_forms.py
```diff
@@ -954,6 +954,29 @@ def test_unused_questionnaire_visibility(self):
...
+        form_data = get_form_data_from_instance(EvaluationForm, evaluation)
+        form = EvaluationForm(form_data, instance=evaluation)
+        self.assertEqual(len(form["participants"]), 1)
+
+    def test_inactive_participants_not_in_queryset(self):
+        evaluation = baker.make(Evaluation, course__degrees=[baker.make(Degree)])
+
+        form_data = get_form_data_from_instance(EvaluationForm, evaluation)
+        form = EvaluationForm(form_data, instance=evaluation)
+        self.assertEqual(len(set(form.fields["participants"].queryset)), 0)
```
He3lixxx, line 970: The querysets should be distinct, so the intermediate sets shouldn't matter. `.count()` should work, so you could use `queryset.count()` instead of `len(set(queryset))`.

Same for the assertions below.

evap/static/scss/_adjustments.scss
```diff
@@ -140,6 +140,18 @@ $alert-colors: (
     white-space: normal;
 }
 
+.user-multi-select {
```
He3lixxx, line 143: Technically, this class just makes the multi select have one entry per line. We have some other user multi selects that will not use this class, so it might make sense to have the name reflecting that.

He3lixxx (changes requested):
select2 was updated to 4.1.0-rc1 on upstream:main in between (see #1702) -- so after merging or rebasing, the select2 stuff should disappear from the diff.

evap/contributor/tests/test_views.py
dieknolle3333: Thank you for the hint!

evap/contributor/templates/contributor_evaluation_form.html
```diff
@@ -63,7 +63,13 @@ <h5 class="card-title">{% trans 'Course data' %}</h5>
             <div class="card-body">
                 <div class="d-flex">
                     <h5 class="card-title me-auto">{% trans 'Evaluation data' %}</h5>
-                    {% if not evaluation.allow_editors_to_edit %}
+                    {% if evaluation.allow_editors_to_edit %}
+                            <div>
+                                <button type="button" class="btn btn-sm btn-light mb-3" onclick="createAccountRequestModalShow();" id="createAccountRequestModalButton">
+                                    {% trans 'Request creation of new account' %}
+                                </button>
+                            </div>
```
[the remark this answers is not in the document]
He3lixxx: (In any way: The same `id` is used in contribution_formset.html, which is included in line 79, but the html standard [requires `id`s to be unique](https://www.w3.org/TR/2014/REC-html5-20141028/dom.html#the-id-attribute))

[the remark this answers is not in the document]
He3lixxx: Okay, then I'd say we should create a new issue for that.

For this PR, I'd still ask to fix the non-unique id.

He3lixxx (approved):
Code is fine with me, haven't tested manually.

evap/contributor/tests/test_views.py
```diff
@@ -259,3 +259,18 @@ def test_information_message(self):
...
+        self.evaluation.allow_editors_to_edit = False
+        self.evaluation.save()
+        page = self.app.get(self.url, user=self.responsible)
+        expected_request_changes_count = page.body.decode().count("Request changes")
+        expected_request_account_count = page.body.decode().count("Request creation of new account")
+
+        self.evaluation.allow_editors_to_edit = True
+        self.evaluation.save()
+        page = self.app.get(self.url, user=self.responsible)
+        self.assertEqual(page.body.decode().count("Request changes"), expected_request_changes_count - 1)
```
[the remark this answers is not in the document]
dieknolle3333: I think the idea was to make the count independent of the rest of the view. Is that unnecessary?

niklasmohrin (approved):
Code looks good, UI does too as far as I can tell - @janno42 I am relying on your thorough review though :)

janno42 (approved):

janno42:
Thank you so much 👍

dieknolle3333:
yeeeey, this PR finally made it! :D

PR #1072, opened 2018-01-02
Ui updates

janno42:
Many small and a few larger updates to the UI. Many things are getting even better in 2018!
See commit messages for individual changes.

PR #1643, opened 2021-10-18
sort due evaluations by due date, then by name (closes #1612)

### LINKED ISSUE #1612: Due evaluations ordering
The reminder message about due evaluations (management command `send_reminders`) sorts them by the number of remaining days. A second ordering level should be added, ordering evaluations with the same number of remaining days by their full name.

Skn0tt:


evap/evaluation/models.py
```diff
@@ -1719,7 +1718,7 @@ def get_sorted_due_evaluations(self):
 
         # Sort evaluations by number of days left for evaluation and bring them to following format:
         # [(evaluation, due_in_days), ...]
-        return sorted(due_evaluations.items(), key=operator.itemgetter(1))
+        return sorted(due_evaluations.items(), key=lambda tuple: (tuple[1], tuple[0].full_name))
```
He3lixxx, line 1721: We don't use `tuple` as a variable name since it is a python builtin. In this case, it wouldn't cause any problems except messing up syntax highlighting. One could argue whether it's a bug that pylint doesn't issue a [`redefined-builtin` error](https://vald-phoenix.github.io/pylint-errors/plerr/errors/variables/W0622.html) for this.

@janno42 This method was originally introduced in https://github.com/e-valuation/EvaP/pull/1072/commits/11f5b846d571a7288e2cf8128fa038b1b7249665#diff-ae63028ddccb2a801f831dd096db77b51c87859ac2d9ad8cfd2ee972bde59d78R149. I don't think we need the dictionary, the QuerySet does not contain duplicates. Thus, a much more straightforward
```python3
def get_sorted_due_evaluations(self):
    evaluations_and_days_left = (
        (evaluation, evaluation.days_left_for_evaluation())
        for evaluation in Evaluation.objects.filter(
            participants=self, state=Evaluation.State.IN_EVALUATION
        ).exclude(voters=self)
    )

    return sorted(evaluations_and_days_left, key=lambda tup: (tup[1], tup[0].full_name))
```
should get the job done and be easier to understand. What do you think?

janno42: sure, fine with that 👍

janno42:
replaced by #1674

PR #1646, opened 2021-10-18
Introduce centralised sorting_key property 

### LINKED ISSUE #1624: Handle user profile sorting with incomplete names
Sorting user profiles by first and last name currently fails when not both are specified. This happens for special users such as, e.g., "Student Representatives" who do not have a first and last name specified.

For example, when opening a semester's grade reminder page, loading the page currently fails when at least one of the responsible users of a course listed on the page does not have a first and last name specified (`AttributeError at /staff/semester/<id>/grade_reminder 'NoneType' object has no attribute 'lower'`).

As a solution, if only a last name is specified, the first name can be ignored. If both values are None, the email must be used for sorting.

These places need to be checked and updated:
- `evaluation.models.responsibles_names`
- `staff.views.semester_preparation_reminder`
- `staff.views.semester_grade_reminder`

Skn0tt:
(Closes #1624)

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -1513,7 +1513,7 @@ class UserProfile(AbstractBaseUser, PermissionsMixin):
     is_active = models.BooleanField(default=True, verbose_name=_("active"))
 
     class Meta:
-        ordering = ["last_name", "first_name", "email"]
+        ordering = ["last_name", "first_name", "email"] # keep in sync with #sorting_key
         verbose_name = _("user")
         verbose_name_plural = _("users")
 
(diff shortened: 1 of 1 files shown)


evap/evaluation/models.py
```diff
@@ -1552,6 +1552,14 @@ def full_name_with_additional_info(self):
...
         return name + " (" + self.email + ")"
 
+    @property
+    def sorting_key(self): # keep in sync with Meta#ordering
+        return (
+            (self.last_name or "").lower(),
+            (self.first_name or "").lower(),
+            (self.email or "").lower()
+        )
+
```
He3lixxx, line 1562: I'd assume that when ordering by string, we usually want case insensitivity (@janno42?). In this case, [django supports case insensitive model ordering](https://code.djangoproject.com/ticket/6498#comment:24) using the [`Lower` function](https://docs.djangoproject.com/en/3.2/ref/models/database-functions/#lower).

For the three changed sort calls, we could then just use database sorting, which we get anyway. In staff/views.py, this would even gives us a chance for optimization, it currently looks like we have O(N) queries there.

Any design change we do here should probably be done uniformly across the code base (probably another issue)

PR #1647, opened 2021-10-18
Use language-level ordering sugar

Skn0tt:
Depends on #1646 to be merged first

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -5,6 +5,7 @@ import uuid
 from collections import defaultdict, namedtuple
 from datetime import date, datetime, timedelta
 from enum import Enum, auto
+from functools import total_ordering
 
 from django.conf import settings
 from django.contrib import messages
@@ -354,9 +355,7 @@ class Course(LoggedModel):
 
     @cached_property
     def responsibles_names(self):
-        ordered_responsibles = sorted(
-            self.responsibles.all(), key=lambda responsible: responsible.sorting_key
-        )
+        ordered_responsibles = sorted(self.responsibles.all())
         return ", ".join([responsible.full_name for responsible in ordered_responsibles])
 
     @property
@@ -1483,6 +1482,7 @@ class UserProfileManager(BaseUserManager):
         return user
 
 
+@total_ordering
 class UserProfile(AbstractBaseUser, PermissionsMixin):
     # null=True because certain external users don't have an address
     email = models.EmailField(max_length=255, unique=True, blank=True, null=True, verbose_name=_("email address"))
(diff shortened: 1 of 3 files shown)


evap/evaluation/models.py
```diff
@@ -1483,6 +1482,7 @@ def create_superuser(self, email, password, first_name=None, last_name=None):
         return user
 
 
+@total_ordering
```
He3lixxx, line 1485: The ordering relation we use here is not a [total order](https://en.wikipedia.org/wiki/Total_order). We would have
```python3
>>> user1, user2 = UserProfile(), UserProfile()
>>> user1 < user2 
False
>>> user2 < user1
False
>>> user2 == user1
False
```
because django's default implementation of `__eq__()` uses the primary key to detect equality.

I think this inconsistency will cause bugs in the future and thus would vote strongly against using [`@total_ordering`](https://docs.python.org/3/library/functools.html#functools.total_ordering) here.

If we want to use comparison for sorting, IMO, we should _only_ implement `__lt__()`. However, I'd also vote against doing that, see discussion below.

evap/evaluation/models.py
```diff
@@ -1560,6 +1560,9 @@ def sorting_key(self): # keep in sync with Meta#ordering
             (self.email or "").lower()
         )
 
+    def __lt__(self, other):
+        return self.sorting_key < other.sorting_key
```
He3lixxx, line 1564: Django has a standardized way of specifying the default ordering, though [`Meta.ordering`](https://docs.djangoproject.com/en/3.2/ref/models/options/#django.db.models.Options.ordering). Django has explicitly decided _not_ to implement `__lt__()` for the models when it was requested ([ticket 26141](https://code.djangoproject.com/ticket/26141), [ticket 22843](https://code.djangoproject.com/ticket/22843)).

When defining one comparison operator, users of the class would usually expect that you have some kind of total or partial order that is being expressed, and that the remaining comparison operators are defined and behave as [would commonly be expected](https://stackoverflow.com/a/1061350/12345551).
However, we can not achieve a meaningful implementation for all the operators here (see discussion above). Django ships a default implementation for `__eq__()` that uses an object's primary key. Our comparison operators would use some other attributes. I don't see how this could work together semantically.
We would be left with implementing only `__lt__()` for the purpose of sorting objects, but not comparing them otherwise.

Considering that we often have new contributors in this project, I think this would be a source of confusion. In most cases, django's ordering feature for database queries should be usable (also in this case, see https://github.com/e-valuation/EvaP/pull/1646#discussion_r732000220), and if we don't have a real need for this, I would avoid it.

PR #1651, opened 2021-10-25
Prevent deleting contributors voted for in the form

### LINKED ISSUE #1558: Prevent deleting contributors voted for in the form
Staff users can delete contributors in the evaluation form even if answers to questions of their contribution have already been given. However, this correctly results in a `ProtectedError` when saving the form:
```
Cannot delete some instances of model 'Contribution' because they are referenced through a protected foreign key
```
Deleting these contributors should not be possible in the form.

jb3rndt:
Fixes #1558

The delete button will not be displayed for contributions when editing an evaluation if they got voted for.

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -1046,9 +1046,10 @@ class Contribution(LoggedModel):
 
     @property
     def can_be_deleted(self):
-        return (
+        return (self.pk is None and self.is_general) or (
             not RatingAnswerCounter.objects.filter(contribution=self).exists()
             and not TextAnswer.objects.filter(contribution=self).exists()
+            and not self.is_general
         )
 
     @property
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -1045,16 +1045,6 @@ class Contribution(LoggedModel):
         return self.contributor_id is None
 
     @property
-    def can_be_deleted(self):
-        if self.pk is None:
-            return True  # not stored in the DB. Required so temporary instances in the formset can be deleted.
-
-        if not self.ratinganswercounter_set.exists() and not self.textanswer_set.exists():
-            return True
-
-        return False
-
-    @property
     def object_to_attach_logentries_to(self):
         return Evaluation, self.evaluation_id
 
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -1046,14 +1046,13 @@ class Contribution(LoggedModel):
 
     @property
     def can_be_deleted(self):
-        if self.is_general:
-            # The pk is used to distinct the general contribution of an evaluation from a new contribution in a ContributionFormset.
-            return self.pk is None
-        else:
-            return (
-                not RatingAnswerCounter.objects.filter(contribution=self).exists()
-                and not TextAnswer.objects.filter(contribution=self).exists()
-            )
+        if self.pk is None:
+            return True  # not stored in the DB. Required so temporary instances in the formset can be deleted.
+
+        if not self.ratinganswercounter_set.exists() and not self.textanswer_set.exists():
+            return True
+
+        return False
 
     @property
     def object_to_attach_logentries_to(self):
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -1045,6 +1045,13 @@ class Contribution(LoggedModel):
         return self.contributor_id is None
 
     @property
+    def can_be_deleted(self):
+        return (
+            not RatingAnswerCounter.objects.filter(contribution=self).exists()
+            and not TextAnswer.objects.filter(contribution=self).exists()
+        )
+
+    @property
     def object_to_attach_logentries_to(self):
         return Evaluation, self.evaluation_id
 
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -1046,11 +1046,14 @@ class Contribution(LoggedModel):
 
     @property
     def can_be_deleted(self):
-        return (self.pk is None and self.is_general) or (
-            not RatingAnswerCounter.objects.filter(contribution=self).exists()
-            and not TextAnswer.objects.filter(contribution=self).exists()
-            and not self.is_general
-        )
+        if self.is_general:
+            # The pk is used to distinct the general contribution of an evaluation from a new contribution in a ContributionFormset.
+            return self.pk is None
+        else:
+            return (
+                not RatingAnswerCounter.objects.filter(contribution=self).exists()
+                and not TextAnswer.objects.filter(contribution=self).exists()
+            )
 
     @property
     def object_to_attach_logentries_to(self):
--- a/evap/evaluation/templates/contribution_formset.html
+++ b/evap/evaluation/templates/contribution_formset.html
@@ -66,7 +66,7 @@
                         {% include 'bootstrap_form_errors.html' with errors=form.label.errors %}
                     </td>
                     <td>
-                        {% if editable and form.instance.can_be_deleted %}
+                        {% if editable and form.show_delete_button %}
                             {% include 'bootstrap_form_field_widget.html' with field=form.DELETE class="d-none" %}
                         {% endif %}
                     </td>
--- a/evap/evaluation/templates/contribution_formset.html
+++ b/evap/evaluation/templates/contribution_formset.html
@@ -66,7 +66,7 @@
                         {% include 'bootstrap_form_errors.html' with errors=form.label.errors %}
                     </td>
                     <td>
-                        {% if editable and form.is_deletable.value %}
+                        {% if editable and form.can_be_deleted %}
                             {% include 'bootstrap_form_field_widget.html' with field=form.DELETE class="d-none" %}
                         {% endif %}
                     </td>
--- a/evap/evaluation/templates/contribution_formset.html
+++ b/evap/evaluation/templates/contribution_formset.html
@@ -66,7 +66,7 @@
                         {% include 'bootstrap_form_errors.html' with errors=form.label.errors %}
                     </td>
                     <td>
-                        {% if editable and form.can_be_deleted %}
+                        {% if editable and form.instance.can_be_deleted %}
                             {% include 'bootstrap_form_field_widget.html' with field=form.DELETE class="d-none" %}
                         {% endif %}
                     </td>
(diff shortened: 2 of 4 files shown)


evap/staff/forms.py
```diff
@@ -495,6 +495,7 @@ def __init__(self, *args, evaluation=None, **kwargs):
 
         if self.instance.contributor:
             self.fields["contributor"].queryset |= UserProfile.objects.filter(pk=self.instance.contributor.pk)
+            self.fields["contributor"].initial = self.instance.contributor
```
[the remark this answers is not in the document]
jb3rndt: Ah right, its not needed

evap/evaluation/templates/contribution_formset.html
```diff
@@ -66,9 +66,7 @@ <h5 class="card-title mr-auto">{% trans 'Contributors' %}</h5>
                         {% include 'bootstrap_form_errors.html' with errors=form.label.errors %}
                     </td>
                     <td>
-                    {% if editable %}
-                        {{ form.DELETE }}
-                    {% endif %}
+                        {% if editable and form.can_be_deleted %}{{ form.DELETE }}{% endif %}
```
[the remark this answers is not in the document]
jb3rndt: Yes that works as well 👍

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

evap/staff/forms.py
```diff
@@ -506,6 +507,8 @@ def __init__(self, *args, evaluation=None, **kwargs):
             .distinct()
         )
 
+        self.can_be_deleted = self.instance.can_be_deleted and not self.instance.is_general
```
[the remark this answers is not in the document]
He3lixxx: Hmm I think it would be cool to have `can_be_deleted` be False on every general contribution. Could you maybe circumvent the issue with the new contributor by checking whether the instance has a primary key set? The new instance should not exits in the database, so it should not have a primary key set. For instances without a PK, I'd say it's fair that they can be deleted.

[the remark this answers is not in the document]
jb3rndt: Ok I see. I hope I unterstood it correctly. I made it so a general contribution can only be deleted if it has no pk of if a contribution is not general and has no answers which should be what we want (I hope) :)

evap/evaluation/models.py
```diff
@@ -1044,6 +1044,14 @@ def unlogged_fields(self):
     def is_general(self):
         return self.contributor_id is None
 
+    @property
+    def can_be_deleted(self):
+        return (self.pk is None and self.is_general) or (
+            not RatingAnswerCounter.objects.filter(contribution=self).exists()
+            and not TextAnswer.objects.filter(contribution=self).exists()
+            and not self.is_general
+        )
```
niklasmohrin, line 1053: So, this is equivalent to the following, right? Why do we care about the primary key when the contribution is general but don't when it is not? I would best prefer the version below with a comment about the pk :)
```suggestion
        if self.is_general:
            return self.pk is None
        else:
            return (
                not RatingAnswerCounter.objects.filter(contribution=self).exists()
                and not TextAnswer.objects.filter(contribution=self).exists()
            )
```

jb3rndt: Yes it is equivalent, and much more readable 👍
If the question was not rhetorical: If a contribution is the general contribution for an evaluation, it should not be deletable at all. But the extra contribution in the Contribution formset is considered general as well, based on the property. So to distinct both of them, the pk is used because the real general contribution for an evaluation does already exist in the database.

He3lixxx: Radical opinion: Make the flow even more human, which the following structure:
- In general, deletion is not possible
- Exception: Object is not yet stored in database (technically no deletion, but required for the button)
- Exception: No Answers exist
which yields
```python
@property
def can_be_deleted(self):
    if self.pk is None:
        return True  # not stored in the DB. Required so temporary instances in the formset can be deleted.
    
    if not self.ratinganswercounters_set.exists() and not self.textanswers_set.exists():
        return True
    
    return False
```

using `self.ratinganswercounters_set` and the equivalent for the textanswers allows prefetching for the property, so we don't always need 2 queries to compute this property.

evap/evaluation/models.py
```diff
@@ -1044,6 +1044,16 @@ def unlogged_fields(self):
     def is_general(self):
         return self.contributor_id is None
 
+    @property
+    def can_be_deleted(self):
+        if self.pk is None:
+            return True  # not stored in the DB. Required so temporary instances in the formset can be deleted.
+
+        if not self.ratinganswercounter_set.exists() and not self.textanswer_set.exists():
```
niklasmohrin, line 1052: It seems odd that the knowledge about the possible kinds of answers is also kept here, would there be any way to avoid that (without loosing the advantages @He3lixxx named)? Or should we add an assertion about the available subclasses like on line 462 in `student/*/test_views.py` to the new test?

niklasmohrin (approved):

evap/evaluation/models.py
He3lixxx: Adding an assertion either to the test or to the method is probably a good idea.

I'd be slightly in favor of adding it to the method because that's where the assertion is really needed, and the test would fail in both ways, but considering that we added it to the test in #1549, that's probably what we should do for uniformity.

evap/staff/tests/test_forms.py
```diff
@@ -554,6 +555,30 @@ def test_staff_can_select_proxy_user(self):
...
+            questionnaires=[questionnaire],
+        )
+        baker.make(RatingAnswerCounter, contribution=contribution1)
+
+        contribution_formset = inlineformset_factory(
+            Evaluation, Contribution, formset=ContributionFormSet, form=ContributionForm, extra=1
+        )
+        formset = contribution_formset(instance=evaluation, form_kwargs={"evaluation": evaluation})
+        self.assertFalse(formset.forms[0].instance.can_be_deleted)
+        self.assertTrue(formset.forms[1].instance.can_be_deleted)
```
He3lixxx, line 580: `forms[0]` is `contribution1` and `forms[1]` is the extra instance, right? So we're not really testing that `contribution1` was ever deletable.

In order to make sure that the test really tests the right thing, I usually advocate doing assertions before and after. In this case, that would be:
- Assert that the contribution is deletable
- Add the RatingAnswerCounter
- Assert that the contribution is not deletable anymore.

This is a Forms test, but you're not actually testing any real forms logic. It might make sense to just make it a models test.

evap/evaluation/models.py
```diff
@@ -1044,6 +1044,16 @@ def unlogged_fields(self):
     def is_general(self):
         return self.contributor_id is None
 
+    @property
+    def can_be_deleted(self):
```
He3lixxx, line 1048: I'm a bit uneasy about the following scenario:
- Issue for a new requirement: Contributions should not be deletable anymore if X.
- User stumbles across this method, and just adds `if X: return False´, thinking that it solves the issue
- The delete button will not be shown anymore -- but malicious staff users will be able to delete the instances. Also: Do we have other locations where Contributions can be deleted?

Considering this, I'd say for maintainability, it would be best if we either added this as a proper permissions check to wherever a contribution is deleted, or renamed it to "show_delete_button" or something. The latter could contain an explaining comment.

He3lixxx (changes requested):
Sorry I should have brought up the `can_be_deleted` thought earlier. I didn't think about it properly in past reviews.

evap/evaluation/models.py
jb3rndt: Makes sense. As far as I know this is the only place where contributions are explicitly deleted.
Currently, deleting contributions (with answers) is not possible for any user (including malicious staff users) becaase of the database protection. So when we care about deletion protection, this would only be useful if there are other conditions (like the case you described) in which the database would actually allow the deletion. That leads me to the point that this would be caring for future conditions, which is not necessarily bad, but I am not sure if it should be handled right now and here?
If we want to do it I would probably override the `delete` method so we dont have to search for all places where contributions are deleted.

Otherwise I would probably just rename it to "show_delete_button", as you mentioned. What bothers me about this is that the model implements view logic, which does not really fit, right?

What do you think?

evap/evaluation/models.py
jb3rndt: Seems reasonable, I just copied the assertion from #1549 :P

evap/evaluation/models.py
He3lixxx: Totally agree. I'd suggest renaming the method and moving it into the form.

jb3rndt: Good Idea 👍

evap/staff/tests/test_forms.py
```diff
@@ -554,6 +557,39 @@ def test_staff_can_select_proxy_user(self):
...
+        """
+        self.assertEqual(
+            set(Answer.__subclasses__()),
+            {RatingAnswerCounter, TextAnswer},
+            "This requires an update if a new answer type is added",
+        )
+        evaluation = baker.make(Evaluation)
+        user1 = baker.make(UserProfile)
+        questionnaire = baker.make(Questionnaire, type=Questionnaire.Type.CONTRIBUTOR)
+        contribution1 = baker.make(
```
niklasmohrin, line 572: You could inline the creation of `contributor` and omit the mention of the questionnaire altogether. Secondly, there isn't a `contribution2`, so why is this one called `contribution1`?

niklasmohrin (approved):
Haven't rechecked the site, but nothing changed semantically, soooo ...

I have one nitpick for the test left though :)

janno42 (approved):
functionality still given. i like the new implementation 👍

evap/staff/tests/test_forms.py
He3lixxx: (for the contributor, ` _fill_optional=['contributor']` would completely remove `user1`)

He3lixxx (commented):
(Just to clear my negative review -- hoping that "Comment" does that)

Code is nice. The model creation could be slimmed a bit as @niklasmohrin mentioned, but I'd also accept if the current version was merged.

evap/staff/tests/test_forms.py
jb3rndt: Yep makes sense :)

He3lixxx (approved):
Thanks! :)

Issue #1648, opened 2021-10-21
VM provisioning fails on windows hosts because symlinks can not be created


He3lixxx:
Currently, when not running vagrant/virtualbox as admin on a windows host, the provisioning fails at `npm ci` because npm can not create the symlinks from `./node_modules/.bin/` to the binaries in package-local bin directories (e.g. `node_modules/typescript/bin/`).

We had a similar problem with the venv directory in #1520. There, our fix was moving the venv out of the mounted directory, with the main drawback that the venv can not be used on the host (It may cause trouble using it on the host anyway, since the binary links in there pointing to e.g. python3 can not be made available on the host).

I guess it would make sense to have the same "solution" to this problem applied for both the `venv` and the `node_modules` directory.

Possible solutions with npm:
- It is possible to have symlinks in the mounted directory [_if_ virtualbox can map them to symlinks in windows](https://stackoverflow.com/questions/23936458/correct-way-to-setup-virtualbox-4-3-to-use-symlinks-on-guest-for-meteor). However, to do that, it needs to [either be run as admin](https://security.stackexchange.com/questions/10194/why-do-you-have-to-be-an-admin-to-create-a-symlink-in-windows) of have [a security policy flag changed for the user running it](https://superuser.com/a/105381).
    - We could just ask windows users to run vagrant/virtualbox as admin when they plan to provision the vm or use npm in any way, maybe add a note in the README, and be done with it
    - We could also try to make some script that runs when users start vagrant that detects whether creating a symlink is currently possible. If not, it could [check whether the current user has the necessary flag set, and if it hasn't, it could ask for admin permission (UAC) and set the flag](https://dbondarchuk.com/2016/09/23/adding-permission-for-creating-symlink-using-powershell/). This would be around 70 lines of powershell code we would need to provide. It would mean we permanently change some system configuration on the host machine.

- Have it install somewhere else (same as we did with the venv) -- I don't even think that's possible for a "local" installation ([npm doc page](https://docs.npmjs.com/cli/v7/configuring-npm/folders) doesn't list any such possibility), but it seems like [using a "global" installation could do the trick](https://stackoverflow.com/questions/14742553/how-to-set-custom-location-for-local-installation-of-npm-package)

[...]

PR #1657, opened 2021-10-31
Update README.md

He3lixxx:
Related to #1648, but probably not a final solution(?)

It's a struggle to be as open/informing about what's happening on one side, but also keeping the text short and accessible to people that might be new to all this stuff, and might not know line endings or symlinks.

- Linux Subsystem on Windows: I don't know what steps are necessary there, but I'd guess, neither the git configuration nor the symlink privileges are required. We could test this if someone uses the subsystem by running `git config --global core.autocrlf` and `ln -s some_target some_source`.
- Technically, the symlink privileges will be required every time users use NPM inside the VM. That includes the first time setup or any later NPM usage. However, I think it's fair to only go into detail with this when users face this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,17 +21,23 @@ The easiest setup using [Vagrant](https://www.vagrantup.com) and [VirtualBox](ht
 
 1. Fork the Evap repository (using the Fork-button in the upper right corner on GitHub)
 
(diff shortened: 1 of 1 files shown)


janno42 (approved):

niklasmohrin (approved):
Seems appropriate :+1: Codacy is still crying, but honestly these markdown lints are a bit ridiculous :D

PR #1660, opened 2021-11-01
Dark Mode

### LINKED ISSUE #1039: Dark Theme
We had a user request for a dark mode. Hopyfully, this only involves making a new set of css file.

rgwohlbold:
Fixes #1039


karyon:
Fwiw, sass has an inline if: `$variable: if(condition, result-when-true, result-when-false);` ([via stackoverflow](https://stackoverflow.com/questions/25300453/does-sass-support-inline-if-statements))

I imagine this would make the code more concise and less duplicated.

Also I was about to suggest to use methods for common color pairs, but it seems like you introduced some already :)

rgwohlbold:
Thanks @karyon , that was exactly what I needed :)
I will look into refactoring when the styling is done :+1:

niklasmohrin:
@rgwohlbold Do you want to continue working on this? Otherwise, we would like to unassign you from the issue, so that others can pick it up :)

rgwohlbold:
I'll probably not work in this in the near future, you can unassign me

01.08.2022 17:25:13 Niklas Mohrin ***@***.***>:

Issue #1640, opened 2021-10-18
Wrong grade document edit form title


janno42:
When editing a grade document that holds final grades, the title of the form wrongly shows "Upload midterm grades" (should be "Upload final grades" instead) because the parameter `final_grades` is not correctly set for the template.
This can for example be seen at the course "Operating Systems I (Summer term 2014)" in the test data.

PR #1661, opened 2021-11-04
#1640 grade document

fidoriel:
fix for #1640 grade document

using the same 'final' flag as the upload_grades func

### DIFF
--- a/evap/grades/templates/grades_course_view.html
+++ b/evap/grades/templates/grades_course_view.html
@@ -29,7 +29,7 @@
                                 <td>
                                     <a href="{% url 'grades:download_grades' grade_document.id %}" class="btn btn-sm btn-light" data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans 'Download' %}"><span class="fas fa-download"></span></a>
                                     {% if user.is_grade_publisher %}
-                                        <a href="{% url 'grades:edit_grades' semester.id course.id grade_document.id %}?final={% if final_grades %}true{% else %}false{% end %}" class="btn btn-sm btn-light" data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans 'Edit' %}"><span class="fas fa-pencil-alt"></span></a>
+                                        <a href="{% url 'grades:edit_grades' semester.id course.id grade_document.id %}" class="btn btn-sm btn-light" data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans 'Edit' %}"><span class="fas fa-pencil-alt"></span></a>
                                         <button type="button" {{ disable_if_archived }} onclick="deleteGradedocumentModalShow({{ grade_document.id }}, '{{ grade_document.description|escapejs }}');" class="btn btn-sm btn-danger" data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans 'Delete' %}">
                                             <span class="fas fa-trash"></span>
                                         </button>
--- a/evap/grades/templates/grades_course_view.html
+++ b/evap/grades/templates/grades_course_view.html
@@ -29,11 +29,7 @@
                                 <td>
                                     <a href="{% url 'grades:download_grades' grade_document.id %}" class="btn btn-sm btn-light" data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans 'Download' %}"><span class="fas fa-download"></span></a>
                                     {% if user.is_grade_publisher %}
-                                        {% if final_grades %}
-                                            <a href="{% url 'grades:edit_grades' semester.id course.id grade_document.id %}?final=true" class="btn btn-sm btn-light" data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans 'Edit' %}"><span class="fas fa-pencil-alt"></span></a>
-                                        {% else %}
-                                            <a href="{% url 'grades:edit_grades' semester.id course.id grade_document.id %}" class="btn btn-sm btn-light" data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans 'Edit' %}"><span class="fas fa-pencil-alt"></span></a>
-                                        {% endif %}
+                                        <a href="{% url 'grades:edit_grades' semester.id course.id grade_document.id %}?final={% if final_grades %}true{% else %}false{% end %}" class="btn btn-sm btn-light" data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans 'Edit' %}"><span class="fas fa-pencil-alt"></span></a>
                                         <button type="button" {{ disable_if_archived }} onclick="deleteGradedocumentModalShow({{ grade_document.id }}, '{{ grade_document.description|escapejs }}');" class="btn btn-sm btn-danger" data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans 'Delete' %}">
                                             <span class="fas fa-trash"></span>
                                         </button>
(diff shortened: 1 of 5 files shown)


evap/grades/templates/grades_course_view.html
```diff
@@ -29,7 +29,11 @@ <h3 class="mb-3">{{ course.name }} ({{ semester.name }})</h3>
                                 <td>
                                     <a href="{% url 'grades:download_grades' grade_document.id %}" class="btn btn-sm btn-light" data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans 'Download' %}"><span class="fas fa-download"></span></a>
                                     {% if user.is_grade_publisher %}
-                                        <a href="{% url 'grades:edit_grades' semester.id course.id grade_document.id %}" class="btn btn-sm btn-light" data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans 'Edit' %}"><span class="fas fa-pencil-alt"></span></a>
+                                        {% if "final" in grade_document.get_type_display %}
+                                            <a href="{% url 'grades:edit_grades' semester.id course.id grade_document.id %}?final=true" class="btn btn-sm btn-light" data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans 'Edit' %}"><span class="fas fa-pencil-alt"></span></a>
```
He3lixxx, line 33: The view that we are linking to here will get the `grade_document.id` passed in and load the `GradeDocument` from the database anyway. Thus, we could just do all the logic in the view, and would not require the additional parameter -- or am I missing something?

evap/grades/views.py
```diff
@@ -165,6 +165,8 @@ def edit_grades(request, semester_id, course_id, grade_document_id):
 
     form = GradeDocumentForm(request.POST or None, request.FILES or None, instance=grade_document)
 
+    final_grades = request.GET.get("final") == "true"  # if parameter is not given, assume midterm grades
```
He3lixxx, line 168: This could then become
```
final_grades = grade_document.type == GradeDocument.Type.FINAL_GRADES
```
which would properly utilize the constant enum-like names we have for the two types without relying on whatever string is used in the database.

He3lixxx (changes requested):
Thanks again!

We probably also want to have a small test for this.

fidoriel:
I added the requested changes, now it works only with the bool flag final_grades.

evap/grades/templates/grades_course_view.html
He3lixxx: This change here is still part of the PR -- I guess you didn't add the file before committing?

evap/grades/templates/grades_course_view.html
```diff
@@ -29,7 +29,11 @@ <h3 class="mb-3">{{ course.name }} ({{ semester.name }})</h3>
...
                                     {% if user.is_grade_publisher %}
-                                        <a href="{% url 'grades:edit_grades' semester.id course.id grade_document.id %}" class="btn btn-sm btn-light" data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans 'Edit' %}"><span class="fas fa-pencil-alt"></span></a>
+                                        {% if final_grades %}
+                                            <a href="{% url 'grades:edit_grades' semester.id course.id grade_document.id %}?final=true" class="btn btn-sm btn-light" data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans 'Edit' %}"><span class="fas fa-pencil-alt"></span></a>
+                                        {% else %}
+                                            <a href="{% url 'grades:edit_grades' semester.id course.id grade_document.id %}" class="btn btn-sm btn-light" data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans 'Edit' %}"><span class="fas fa-pencil-alt"></span></a>
```
niklasmohrin, line 35: These two lines are almost the same, you could also inline the `if` into the url:
```django
href="{% url 'grades:edit_grades' semester.id course.id grade_document.id %}?final={% if final_grades %}true{% else %}false{% end %}"
```

Maybe you could even just render variable into a string using `{{ final_grades }}`. On that note: The variable should probably be called `is_final_grades` or similar to suggest it is a boolean and not a collection of the actual grades.

evap/grades/tests.py
```diff
@@ -132,6 +132,23 @@ def test_upload_final_grades(self):
         evaluation.publish()
         evaluation.save()
         self.helper_check_final_grade_upload(course, 0)
+    
+    def test_grades_headlines(self):
+        # check midterm headline
+        self.assertEqual(self.course.midterm_grade_documents.count(), 0)
+        semester = baker.make(Semester, pk=1, grade_documents_are_deleted=False)
+        baker.make(Evaluation, course=baker.make(Course, pk=1, semester=semester), state=Evaluation.State.PREPARED)
```
niklasmohrin, line 140: This can be shortened to
```python
baker.make(Evaluation, course__pk=1, course__semester__pk=1, state=...)
```
Additionally, the `grade_documents_are_deleted` is `False` by default

evap/grades/tests.py
```diff
@@ -132,6 +132,23 @@ def test_upload_final_grades(self):
...
+        response = self.app.get("/grades/semester/1/course/1/upload", user=self.grade_publisher)
+        self.assertContains(response, "Upload midterm grades")
+        self.assertNotContains(response, "Upload final grades")
+
+        # check final headline
+        self.assertEqual(self.course.final_grade_documents.count(), 0)
+        semester = baker.make(Semester, pk=2, grade_documents_are_deleted=False)
+        baker.make(Evaluation, course=baker.make(Course, pk=2, semester=semester), state=Evaluation.State.PREPARED)
+        response = self.app.get("/grades/semester/1/course/1/upload?final=true", user=self.grade_publisher)
```
niklasmohrin, line 149: You use the id `1` again here, I suppose this is not what you intended?

niklasmohrin (changes requested):

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

evap/grades/tests.py
```diff
@@ -238,3 +239,30 @@ def test_403_on_archived_semester(self):
...
+    def test_grades_headlines(self):
+        # check midterm headline
+        grade_publisher = baker.make(
+            UserProfile,
+            email="grade_publisher@institution.example.com",
+            groups=[Group.objects.get(name="Grade publisher")],
+        )
+        grade_document = baker.make(GradeDocument)
+        response = self.app.get(
+            f"/grades/semester/{grade_document.course.semester.pk}/course/{grade_document.course.pk}/edit/{grade_document.pk}",
```
niklasmohrin, line 254: You could save this into a variable `url` once and make the other two uses shorter, your call :)

niklasmohrin (commented):
Not much to criticize about the code :)

After rebasing on current `main`, CI should pass

evap/grades/tests.py
fidoriel: I think this way ist shows what it does because it changes settings in between.

He3lixxx (approved):
Personally, I'd remove both comments from the test and store the url in a local variable as @niklasmohrin suggested:
```python3
url = f"/grades/semester/{grade_document.course.semester.pk}/course/{grade_document.course.pk}/edit/{grade_document.pk}"
# ...
response = self.app.get(url, user=grade_publisher)
```

But I think the current version does the job as well.

niklasmohrin (approved):
what richard said

He3lixxx:
Alrighty so @fidoriel currently the formatting test fails, but I think that's due to a mistake we had in main. If you merge from or rebase onto upstream/main this should resolve.

If you want, you can apply what I suggested in https://github.com/e-valuation/EvaP/pull/1661#pullrequestreview-820163332.

Otherwise, we can merge this once all checks pass.

Issue #1658, opened 2021-11-01
Redirect after login not working


janno42:
When opening a link such as `http://localhost:8000/rewards/` while not being logged in, the login page is shown and a parameter `next` with the originally requested page is set. The parameter specifies, which page should be loaded after a successful login.
Currently however, after being logged in, the index page is shown instead of the requested page.

PR #1662, opened 2021-11-05
fix for #1658

fidoriel:
fixing no redirection if user not auth after login with email and pw

### DIFF
--- a/evap/evaluation/tests/test_views.py
+++ b/evap/evaluation/tests/test_views.py
@@ -42,7 +42,7 @@ class TestIndexView(WebTest):
         self.assertRedirects(response, self.url, fetch_redirect_response=False)
         self.assertRedirects(response.follow(), "/results/")
 
-    def test_password_respects_redirect_parameter(self):
+    def test_login_view_respects_redirect_parameter(self):
         """Regression test for #1658: redirect after login"""
         # external users don't necessarily have a proper redirect page
         internal_email = "manager@institution.example.com"
--- a/evap/evaluation/tests/test_views.py
+++ b/evap/evaluation/tests/test_views.py
@@ -41,6 +41,25 @@ class TestIndexView(WebTest):
         response = password_form.submit()
         self.assertRedirects(response, self.url, fetch_redirect_response=False)
         self.assertRedirects(response.follow(), "/results/")
+    
+    def test_password_respects_redirect_parameter(self):
+        """Regression test for #1658: redirect after login"""
+        internal_email = (
+            "manager@institution.example.com"  # external users don't necessarily have a proper redirect page
+        )
+        baker.make(
+            UserProfile,
+            email=internal_email,
+            password=make_password("evap"),
+            groups=[Group.objects.get(name="Manager")],
+        )
+
+        response = self.app.get(self.url + "?next=/test42/")
+        password_form = response.forms["email-login-form"]
+        password_form["email"] = internal_email
+        password_form["password"] = "evap"
+        response = password_form.submit()
+        self.assertRedirects(response.follow(), "/test42/", fetch_redirect_response=False)
 
     def test_send_new_login_key(self):
         """Tests whether requesting a new login key is only possible for existing users,
--- a/evap/evaluation/tests/test_views.py
+++ b/evap/evaluation/tests/test_views.py
@@ -44,7 +44,6 @@ class TestIndexView(WebTest):
 
     def test_login_view_respects_redirect_parameter(self):
         """Regression test for #1658: redirect after login"""
-        # external users don't necessarily have a proper redirect page
         internal_email = "manager@institution.example.com"
         baker.make(
             UserProfile,
--- a/evap/evaluation/tests/test_views.py
+++ b/evap/evaluation/tests/test_views.py
@@ -50,7 +50,6 @@ class TestIndexView(WebTest):
             UserProfile,
             email=internal_email,
             password=make_password("evap"),
-            groups=[Group.objects.get(name="Manager")],
         )
 
         response = self.app.get(self.url + "?next=/test42/")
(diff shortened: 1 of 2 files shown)


evap/evaluation/views.py
```diff
@@ -75,6 +75,12 @@ def index(request):
             # clean up our test cookie
             if request.session.test_cookie_worked():
                 request.session.delete_test_cookie()
+
+            # check for redirect variable
+            redirect_to = request.GET.get("next", None)
+            if redirect_to is not None:
+                return redirect(redirect_to)
+
```
He3lixxx, line 83: Why did this not work before? Redirecting to "evaluation:index" means the browser would send a GET request for this view, again. Then, the next cookie should still be set, and the "correct" redirect should be handled in lines 100 to 106.

There is already a test for redirecting to the correct start page -- see evaluation/tests/test_views.py line 25 (`test_login_for_staff_users_correctly_redirects`). If I read it correctly, this asserts the redirection to `evaluation:index` followed by the redirection to the correct page. This makes me think there is something wrong with this snippet of code.

In general, I think we just want this branch to fall through to the end of the view, where the case that a user logged in is handled, skipping the whole double-redirect-thing.

He3lixxx (changes requested):
Thanks for contributing!

I think we'd want a proper case for this.

evap/evaluation/views.py
He3lixxx: Ah ok, so I missed that we're not talking about a cookie here, but just a GET parameter. Then, I see that it is lost on the redirect.

I still think that we should simply do nothing here in this if-block and just let the program flow continue until it reaches the handling for the next parameter below in lines 100 to 106.

evap/evaluation/tests/test_views.py
```diff
@@ -42,6 +42,25 @@ def test_login_for_staff_users_correctly_redirects(self):
         self.assertRedirects(response, self.url, fetch_redirect_response=False)
         self.assertRedirects(response.follow(), "/results/")
 
+    def test_password_respects_redirect_parameter(self):
+        """Regression test for #1658: redirect after login"""
+        internal_email = (
+            "manager@institution.example.com"  # external users don't necessarily have a proper redirect page
+        )
```
janno42, line 49:
```suggestion
        # external users don't necessarily have a proper redirect page
        internal_email = "manager@institution.example.com"
```

evap/evaluation/views.py
```diff
@@ -75,6 +76,12 @@ def index(request):
             # clean up our test cookie
             if request.session.test_cookie_worked():
                 request.session.delete_test_cookie()
+
+            # redirect to self needed to call middleware again to set corrtect permission
```
janno42, line 80:
```suggestion
            # redirect to self needed to call middleware again to set correct permission
```

evap/evaluation/views.py
```diff
@@ -75,6 +76,12 @@ def index(request):
...
+            redirect_to = request.GET.get("next", None)
+            if redirect_to is not None:
```
janno42, line 82:
```suggestion
            if redirect_to:
```

janno42 (changes requested):

evap/evaluation/views.py
niklasmohrin: I think the original version is better :D

evap/evaluation/views.py
```diff
@@ -75,6 +76,12 @@ def index(request):
             # clean up our test cookie
             if request.session.test_cookie_worked():
                 request.session.delete_test_cookie()
+
+            # redirect to self needed to call middleware again to set correct permission
```
niklasmohrin, line 80: I think the word "staff mode" earns a mention in the comment :)

niklasmohrin (commented):
Note: Test correctly breaks before this PR :)

I only have two nitpicks that are - of course - open for discussion

evap/evaluation/views.py
He3lixxx: I'd also say the explicit None check is a bit more canonical here. 

However, @janno42's comment gave me the idea of using another value for "no redirect" than None to simplify the rest of the logic.
@fidoriel This is mainly code golf, I think we can also merge the current version.

---

The [specification for `application/x-www-form-urlencoded`](https://www.w3.org/TR/REC-html40/interact/forms.html#form-content-type) does not require a GET parameter value to be non-empty. The query string `?var=&next=&test=true` is perfectly valid.

This means that we could simplify the 4 lines here to a single line:
```python3
return redirect(reverse("evaluation:index") + "?next=" + request.GET.get("next", ""))
```

Which might append a parameter with an empty value, which we could treat just as "no value given" below in 104 with
```python3
redirect_to = request.GET.get("next", "")
if redirect_to:
    return redirect(redirect_to)
```

---

However, I think it would match our intention if we just redirect to the current url again, and not explicitly handle any GET parameters. Django has [`HttpRequest.get_full_path_info()`](https://docs.djangoproject.com/en/3.2/ref/request-response/#django.http.HttpRequest.get_full_path_info), which we could use for that:
```python3
return redirect(request.get_full_path_info())
```
You could construct cases where unconditionally keeping all GET-parameters for a second request causes bugs, but it really shouldn't happen for us as [`GET` should be a safe HTTP-method](https://www.rfc-editor.org/rfc/rfc7231#section-4.2.1).

evap/evaluation/tests/test_views.py
```diff
@@ -42,6 +42,24 @@ def test_login_for_staff_users_correctly_redirects(self):
         self.assertRedirects(response, self.url, fetch_redirect_response=False)
         self.assertRedirects(response.follow(), "/results/")
 
+    def test_password_respects_redirect_parameter(self):
```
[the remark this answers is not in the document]
He3lixxx: I'd also be fine `test_password_login_respects_redirect_parameter`.

He3lixxx (changes requested):
Three small changes, if they are resolved we can merge this. Sorry for the back and forth.

He3lixxx (approved):

niklasmohrin (approved):

janno42 (approved):

janno42:
Thanks! :)

PR #1664, opened 2021-11-08
add alias command for makemessages and used parameters

janno42:


He3lixxx (approved):
You could add a test, but, like, it really doesn't make a difference I think.

karyon:
If django changed e.g. the command's parameters and this custom one would start failing, a test that just checks for the return value might help us notice it sooner than on the next release. Then again, django's command didn't change for years now.

On November 8, 2021 10:27:03 PM GMT+01:00, Richard Ebeling ***@***.***> wrote:
***@***.*** approved this pull request.

niklasmohrin (approved):

janno42:
the test doesn't seem to be the best idea in the current implementation

janno42:
i removed the test again, will be merged without it

PR #1666, opened 2021-11-15
Make TypeScript Prettier

### LINKED ISSUE #1652: Make Typescript code Prettier
We should add automated formatting for our typescript files. I think https://prettier.io/ is pretty good, but the choice is open for discussion. The formatting should be done in `manage.py format` and be checked in CI.

niklasmohrin:
Closes #1652

- `npm install --save-dev prettier`
- `npx prettier --write src`
- Add Prettier to management command
- Add Prettier to CI

### DIFF
--- a/evap/static/ts/src/csrf-utils.ts
+++ b/evap/static/ts/src/csrf-utils.ts
@@ -1,7 +1,8 @@
 // based on: https://docs.djangoproject.com/en/3.1/ref/csrf/#ajax
 function getCookie(name: string): string | null {
     if (document.cookie !== "") {
-        const cookie = document.cookie.split(";")
+        const cookie = document.cookie
+            .split(";")
             .map(cookie => cookie.trim())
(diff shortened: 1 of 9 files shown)


evap/static/ts/src/csrf-utils.ts
```diff
@@ -19,8 +22,9 @@ function isMethodCsrfSafe(method: string): boolean {
 
 // setup ajax sending csrf token
 $.ajaxSetup({
-    beforeSend: function(xhr: JQuery.jqXHR, settings: JQuery.AjaxSettings) {
-        const isMethodSafe = settings.method && isMethodCsrfSafe(settings.method);
```
He3lixxx, line 23: Does it have a ~80 chars line length limit? If so, do we want to allow lines to be a bit longer? In python, we allow 120.

He3lixxx (approved):

PR #1667, opened 2021-11-15
fix #1333

### LINKED ISSUE #1333: Redesign user index
The user index page loads quite slowly. This could probably be improved be removing badges that are not really needed ("Contributor", "Responsible", "External", "No email") and postponing the `can_be_deleted_by_manager` check to be run only after clicking the button and returning an error message if not possible
It should be investigated if further improvements to the loading time are possible.

lill28:


evap/staff/templates/staff_user_index.html
```diff
@@ -9,37 +9,39 @@
...
-                <input type="search" name="search" class="form-control" placeholder="{% trans 'Search...' %}" />
-                <button class="btn btn-light text-secondary" type="button" data-reset="search" data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans 'Clear search filter' %}">
-                    <span class="fas fa-backspace"></span>
-                </button>
+        <div class="col">
+            <div class="card h-100">
+                <div class="card-body">
+                    <form method="POST" class="form-horizontal">
+                        {% csrf_token %}
+                            <h4 class="card-title text-center">{% trans 'Search user' %}</h4>
```
He3lixxx, line 33: This one has the `text-center` class, the heading on the other card doesn't. It would be consistent with the staff index page to not use `text-center` (@janno42?)

evap/staff/templates/staff_user_index.html
```diff
@@ -9,37 +9,39 @@
...
+                                    <button type="submit" class="btn btn-md btn-primary">{% trans 'Edit user' %}</button>
+                                </div>
+                    </form>    
+                </div>
             </div>
         </div>
     </div>
 
-    <div class="card mb-3">
+    <!-- <div class="card mb-3">
```
He3lixxx, line 44: Please remove all the code that was commented out.

evap/staff/templates/staff_user_list.html
```diff
@@ -0,0 +1,105 @@
...
+                    <a href="{% url 'staff:user_list' %}?filter_users=false" role="button" class="btn btn-sm btn-light{% if not filter_users %} active{% endif %}">
+                        {% trans 'Show' %}
+                    </a>
+                    <a href="{% url 'staff:user_list' %}?filter_users=true" role="button" class="btn btn-sm btn-light{% if filter_users %} active{% endif %}">
+                        {% trans 'Hide' %}
+                    </a>
+                </div>
+            </div>
+        </div>
+        <div class="input-group col-3">
```
He3lixxx, line 27: The two divs here are stacked on top of each other without any padding/margin for me (firefox 94.0). With the old page, they were next to each other.

The issue seems to be that the `input-group` class sets `width=100%`, making this not fit in the row anymore. The "old" page had this nested structure to circumvent that:
```html
<div class="col-3">
    <div class="input-group">
        {# CONTENT #}
    </div>
</div>

File: evap/staff/templates/staff_user_list.html [hunk 4]
```diff
@@ -0,0 +1,105 @@
...
+                {% endfor %}
+                </tbody>
+            </table>
+        </div>
+    </div>
+{% endblock %}
+
+{% block modals %}
+    {{ block.super }}
+    {% trans 'Delete user' as title %}
```
He3lixxx, line 78: @janno42 do we want to keep the delete button here?

evap/staff/templates/staff_user_list.html
```diff
@@ -0,0 +1,105 @@
...
+                <thead>
+                    <tr>
+                        <th style="width: 25%">{% trans 'Name' %}</th>
+                        <th style="width: 35%">{% trans 'Email' %}</th>
+                        <th style="width: 30%">{% trans 'Information' %}</th>
+                        <th style="width: 10%">{% trans 'Actions' %}</th>
+                    </tr>
+                </thead>
+                <tbody>
+                {% for user in users %}
```
He3lixxx, line 49: Usually, we'd indent this (since its inside the `<tbody>` we just opened.

evap/staff/templates/staff_user_list.html
```diff
@@ -0,0 +1,105 @@
...
+                            <a href="{% url 'staff:user_edit' user.id request.path %}" class="btn btn-secondary btn-sm"><span class="fas fa-pencil-alt"></span></a>
+                            {% if user.can_be_deleted_by_manager %}
+                                <button type="button" class="btn btn-danger btn-sm" onclick="deleteModalShow({{ user.id }}, '{{ user.full_name|escapejs }}');">
+                                    <span class="fas fa-trash"></span
+                                ></button>
+                            {% else %}
+                                <button type="button" disabled class="btn btn-sm btn-danger" data-toggle="tooltip" data-placement="left"
+                                    title="{% blocktrans %}This user contributes to an evaluation, participates in an evaluation whose participations 
+                                    haven't been archived yet or has special rights and as such cannot be deleted.{% endblocktrans %}">
+                                    <span class="fas fa-trash"></span>
```
He3lixxx, line 64: multiline-tag with child situation again.

evap/staff/templates/staff_user_list.html
```diff
@@ -0,0 +1,105 @@
...
+    <script type="module">
+        import {TableGrid} from "{% static 'js/datagrid.js' %}";
+        new TableGrid().init({
+            storageKey: "user-index-data-grid",
+            table: $(".user-table"),
+            searchInput: $("input[name=search]"),
+            resetSearch: $("[data-reset=search]"),
+        });
+    </script>
+{% endblock %}
```
He3lixxx, line 105: The file is missing a newline character at the end of the last line. There is a preference for this in VS:Code somewhere.

evap/staff/views.py
```diff
@@ -1893,6 +1898,16 @@ def text_answer_warnings_index(request):
 
 @manager_required
 def user_index(request):
+    form = UserEditSelectionForm(request.POST or None)
+
+    if form.is_valid():
+        user = form.cleaned_data['user']
+        return redirect('staff:user_edit', user.id, request.path)
```
He3lixxx, line 1905: Usually, when we want to store a "next"-location, we pass it as a parameter using the [query string](https://en.wikipedia.org/wiki/Query_string) (e.g. when redirecting to the login page). This prevents problems with slashes being part of the value you want to pass.

@janno42 could we live with a static redirect when saving / deleting a user from staff:user_edit?

evap/staff/views.py
```diff
@@ -1893,6 +1898,16 @@ def text_answer_warnings_index(request):
 
 @manager_required
 def user_index(request):
```
He3lixxx, line 1900: It would be nice if you could add a small test for this view that loads the page and submits the user selection form, asserting that the redirect happens as expected.

evap/staff/views.py
```diff
@@ -1925,7 +1940,7 @@ def user_index(request):
         .order_by("last_name", "first_name", "email")
     )
 
-    return render(request, "staff_user_index.html", dict(users=users, filter_users=filter_users))
+    return render(request, "staff_user_list.html", dict(users=users, filter_users=filter_users))
```
He3lixxx, line 1943: codecov says this was not covered by tests -- it seems like there was no test before for the user index page, so it would be nice if you could add a simple one that just loads the page and asserts that there is no server error.

evap/staff/views.py
He3lixxx: Currently, https://localhost:8000/staff/user/949/edit// throws an error for me because a `reverse` call fails in there because no "next" argument is given. I can imagine that there are a few additional places that link to the user edit page. These would need to be changed with the current solution.

If we really need the "next"-page passed here, we should probably just give it a sane default value, probably the user list or the user index. Then, only the other of the two would need to use that parameter.

He3lixxx (changes requested):
Cool, this has very nice solutions to a lot of problems you seem to already have encountered and solved. Good job!

evap/staff/templates/staff_user_list.html
janno42: no, please remove the delete button. as specified, the delete button should only be available in the user form to further increase the page loading time for the user list.

evap/staff/views.py
janno42: a static redirect to `staff/user` would be fine

evap/staff/templates/staff_user_form.html
```diff
@@ -75,7 +75,37 @@ <h5 class="card-title me-auto">{% trans 'Export evaluation results' %}</h5>
...
+    {% trans 'Do you really want to delete the user <strong id="data-label"></strong>?<br/>This person will also be removed from every other user having this person as a delegated or CC-user.' as question %}
+    {% trans 'Delete user' as action_text %}
+    {% include 'confirmation_modal.html' with modal_id='deleteModal' title=title question=question action_text=action_text btn_type='danger' %}
+    <script type="text/javascript">
+        function deleteModalAction(dataId) {
+            $.ajax({
+                type: "POST",
+                url: "{% url 'staff:user_delete' %}",
+                data: {"user_id": dataId},
+                success: function(){window.location="{% url 'staff:user_index' %}"},
```
janno42, line 106: this is different than other form operations - usually, when submitting the form (or performing a different option that afterwards reloads the page) we want to show a message about the successful operation on the page. so in this case we need a message, that the user was successfully deleted.
so maybe we don't want to delete the user via an ajax request here? though this would require changes in the view as well. @He3lixxx what do you think?

evap/staff/templates/staff_user_list.html
```diff
@@ -0,0 +1,105 @@
...
+                        {% trans 'Show' %}
+                    </a>
+                    <a href="{% url 'staff:user_list' %}?filter_users=true" role="button" class="btn btn-sm btn-light{% if filter_users %} active{% endif %}">
+                        {% trans 'Hide' %}
+                    </a>
+                </div>
+            </div>
+        </div>
+        <div class="input-group col-3">
+            <input type="search" name="search" class="form-control" placeholder="{% trans 'Search...' %}" />
```
janno42, line 28: the search field is currently not working. probably the TableGrid initialization below broke during moving the list to this new page.

janno42 (changes requested):

niklasmohrin (commented):
Works very nicely!

janno42:
just reformating left (you can simply use `python manage.py format`) and i think then we're done here :)

niklasmohrin (approved):

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

He3lixxx (commented):
https://github.com/e-valuation/EvaP/pull/1667/commits/cdd0c810e8de7180f63ef28af96b57a7089aa3f6 is named " Change TestUserIndexView to TestUserListView and write new TestUserIndexView" but only does the renaming, and doesn't add the new test. Is it possible that you forgot to `git add` your changes or something?

He3lixxx (approved):
Thanks! :)

He3lixxx:
There are merge conflicts in staff_user_index -- would you like to resolve them yourself, or should we do that?

PR #1669, opened 2021-11-17
Minify triangles_*.svg

karyon:
I know, I know, when thinking of important changes that EvaP needs,
minifying svgs is not what comes to mind first. It was fun ok? :) 
If accepted I can apply this to the dark mode PR as well.

This reduces the gzipped size from 4 to 2.7 kb, mostly by removing
the fractional part of all the coordinates. The visual result is
virtually the same, some edges shift by less than a pixel.

I used https://jakearchibald.github.io/svgomg/ with defaults plus
"Precision" set to zero and "prettify markup" enabled.


He3lixxx (commented):
Fine with me. We've done some design work (name tags, presentations) that used the old file, but I don't see how some subpixel-changes would cause real problems there.

On the other hand, the file will be cached anyways, and just the fontawesome woff2 file is ~74KB.

It looks to me as if the file is not shipped gzipped currently, if this is the case, we might want to change that.

janno42 (approved):

niklasmohrin (approved):

PR #1670, opened 2021-11-22
Fix codecov reporting status too early

He3lixxx:
Currently, the coverage reports on github will report failure for the first few minutes after running the checks, because the coverage for render_pages was measured and reported.

This fixes it.

### DIFF
--- a/.codecov.yaml
+++ b/.codecov.yaml
@@ -8,6 +8,9 @@ coverage:
         target: auto
         threshold: 0.05%
         base: auto
+codecov:
+  notify:
+    after_n_builds: 2
 comment: false
 ignore:
   - "evap/development/views.py"


janno42 (dismissed):

He3lixxx:
Ok so turns out codecov does not overwrite the results with whatever was sent to them most recently, but instead, merges the results (see https://docs.codecov.com/docs/merging-reports). However, on their documentation page, it also says that notifications and comments should be delayed automatically until all CI runs are finished. We did not observe that.

I wonder whether the CI run status is a "notification" in their eyes. I sure feels like one.

To test whether maybe their detection is broken, I manually set the required results to 2. Let's see if this helps.

He3lixxx:
image
Nope, it doesn't. Well.

I'll try to contact their support and ask for help.

janno42:
to be honest: i wouldn't bother too much with codecov. why can't we just remove the coverage part from the ts rendering?

He3lixxx:
@janno42 because we also have quite some python code for rendering the HTML that would then be reported as "not covered". That dropped the project coverage by 0.3%. In the long term, I'd like to get back to 95%.

I had a silly mistake in the yaml config file, I set `coverage.notify.after_n_builds` instead of `codecov.notify.after_n_builds`.
Now it works fine.

janno42 (approved):
i'll just believe you this will work :)

PR #1671, opened 2021-11-22
More aggressive linebreaking

### LINKED ISSUE #1654: Result bars overflow in narrow windows
In narrow windows on the results page, the result bars can overflow the cards.
This should be changed so that the different viewport widths are correctly handled.

Kakadus:
Fixes #1654 

Adds more media-breakpoints for grid and raises the existing ones by one level.

### DIFF
--- a/evap/static/scss/components/_grid.scss
+++ b/evap/static/scss/components/_grid.scss
@@ -77,19 +77,19 @@
         "name semester responsible participants result"
         / auto    6rem       12rem         7rem  11rem;
 
-    @include media-breakpoint-down(md) {
+    @include media-breakpoint-down(lg) {
         grid:
             "name semester  responsible responsible"
             "name semester participants      result"
             / auto    3rem         7rem       10rem;
     }
 
-    @include media-breakpoint-down(sm) {
+    @include media-breakpoint-down(md) {
         grid:
             "        name        name    semester"
             " responsible responsible responsible"
             "participants      result      result"
-            /         1fr         1fr        3rem;
+            /         1fr         1fr        6rem;
     }
 
     &:not(.grid-header) [data-col=responsible] {


evap/static/scss/components/_grid.scss
```diff
@@ -77,19 +77,17 @@
...
         / auto    6rem       12rem         7rem  11rem;
 
-    @include media-breakpoint-down(md) {
+    @include media-breakpoint-down(lg) {
         grid:
-            "name semester  responsible responsible"
-            "name semester participants      result"
-            / auto    3rem         7rem       10rem;
+            "name semester responsible participants result"
+            / auto    3rem       7rem         10rem;
```
janno42, line 83:
```suggestion
            / auto    3rem        7rem        6rem    9rem;
```
there seems to be a value missing. what do you think of this configuration?

evap/static/scss/components/_grid.scss
```diff
@@ -77,19 +77,17 @@
...
-    @include media-breakpoint-down(sm) {
+    @include media-breakpoint-down(md) {
         grid:
-            "        name        name    semester"
-            " responsible responsible responsible"
-            "participants      result      result"
-            /         1fr         1fr        3rem;
+            "name semester  responsible responsible"
+            "name semester participants      result"
+            / auto    1fr         1fr       3rem;
```
janno42, line 90:
```suggestion
            / auto     1fr          1fr        6rem;
```
3rem results in very small bars that are hardly useful (not your fault). what do you think about using this larger value?

janno42 (changes requested):

evap/static/scss/components/_grid.scss
Kakadus: Yes, I must have forgotten to copy some lines there... Now it is the same configuration as before, except the boundaries

evap/static/scss/components/_grid.scss
Kakadus: This is true. I updated it (see last commit)

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

He3lixxx (approved):
Thanks for contributing!

PR #1677, opened 2021-12-06
Introduce maintenance mode

janno42:
This introduces a maintenance mode that is active during a production update.
When in maintenance mode, Apache always returns a maintenance page informing users that EvaP is currently not available.

### DIFF
--- a/deployment/apache.maintenance-template.conf
+++ b/deployment/apache.maintenance-template.conf
@@ -1,0 +1,24 @@
+<VirtualHost *:80>
+    ServerName evap
+
+    DocumentRoot /var/www/html
+
+    RewriteEngine on
+    RewriteCond %{DOCUMENT_ROOT}/maintenance.html -f
+    RewriteCond %{SCRIPT_FILENAME} !maintenance.html
+    RewriteRule !\.(png|css|svg)$ /maintenance.html [R=503,L]
+    RewriteRule maintenance\.css$ /maintenance.css
+    RewriteRule triangles_gray\.svg$ /background_gray.svg
+    RewriteRule triangles_color\.svg$ /background_color.svg
+    RewriteRule favicon\.png$ /favicon.png
+    ErrorDocument 503 /maintenance.html
+    Header Set Cache-Control "max-age=0, no-store"
+
+    ErrorLog /var/log/apache2/error.log
+
+    # Possible values include: debug, info, notice, warn, error, crit,
+    # alert, emerg.
+    LogLevel info
+
+    CustomLog /var/log/apache2/access.log combined
+</VirtualHost>
--- a/deployment/disable_maintenance_mode.sh
+++ b/deployment/disable_maintenance_mode.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+set -e # abort on error
+
+rm -f /var/www/html/maintenance.css
+rm -f /var/www/html/background_gray.svg
+rm -f /var/www/html/background_color.svg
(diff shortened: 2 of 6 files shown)


evap/static/maintenance/maintenance.html
```diff
@@ -0,0 +1,36 @@
+<html>
+	<head>
+		<meta charset="utf-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
+        <title>EvaP</title>
```
He3lixxx, line 5: mixed indentation, these lines use spaces, above and below uses tabs. Indentation is also weird in 14, 25 and 26.

He3lixxx (approved):

He3lixxx (approved):

deployment/apache.maintenance-template.conf
```diff
@@ -0,0 +1,24 @@
+<VirtualHost *:80>
+    ServerName evap
+
+    DocumentRoot /var/www/html
+
+    RewriteEngine on
```
niklasmohrin, line 6: In the vagrant vm, I had to do `sudo a2enmod rewrite` once to get it working, maybe we want to do this in the provision script? (... or my vm was misconfigured :man_shrugging: )

deployment/enable_maintenance_mode.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
```
niklasmohrin, line 2: It's a good practice to have `set -e` here which makes the script execution behave as if it was started with the `-e` flag to bash, so that the script stops if any commands of the script fail. Then it would for example end in `cp: cannot create regular file '/var/www/html/maintenance.css': Permission denied` if one forgets to use `sudo` instead of proceeding and producing more output. Any reasons to not include that here?

deployment/update_production.sh
```diff
@@ -34,7 +34,7 @@ sudo -H -u $USERNAME git fetch
 # Note that apache should not be running during most of the upgrade,
 # since then e.g. the backup might be incomplete or the code does not
 # match the database layout, or https://github.com/e-valuation/EvaP/issues/1237.
-[[ -z "$GITHUB_WORKFLOW" ]] && sudo service apache2 stop
+[[ -z "$GITHUB_WORKFLOW" ]] && sudo ./deployment/enable_maintenance_mode.sh
```
niklasmohrin, line 37: When I checked out the branch, the scripts did not have the executable bit set, so this line would break

deployment/disable_maintenance_mode.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+rm /var/www/html/maintenance.css
```
niklasmohrin, line 3: _If_ we decide to add `set -e`, these should maybe be `rm -f` so that the script still works after a change of this list / if this script is used for a first setup. But that depends on how you want this to work in your workflow and how it is used :)

niklasmohrin (commented):
Nice! I have some thoughts that may turn out to not be important :man_shrugging:

deployment/apache.maintenance-template.conf
janno42: you are absolutely right, i forgot to add the `a2enmod`

deployment/enable_maintenance_mode.sh
janno42: i'm fine with adding it :)

niklasmohrin (approved):

Issue #1596, opened 2021-07-05
Import enrollment data with existing courses


janno42:
When importing enrollments, it currently is not possible to import data with courses that already exist in the semester.
Instead,
- if all attributes of the course (names, type, responsible person, graded) match with the existing one and
- just one evaluation exists for this course,

the importer should show a warning and
- add all new participants from the enrollment data to the existing evaluation and
- add degrees to the course if necessary.

janno42:
@jb3rndt do you want to continue working on this?

jb3rndt:
Yeah I would like to :)

niklasmohrin:
Fixed in #1683

PR #1683, opened 2021-12-13
Import enrollment data with existing courses

jb3rndt:
Solves #1596
Todo:

- [x] Write Tests
- [x] Rework architecture for checking for existing courses


evap/staff/importers.py
```diff
@@ -97,28 +99,87 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
+            )
+            evaluation.save()
+            evaluation.contributions.create(
+                evaluation=evaluation,
+                contributor=responsible_dbobj,
+                role=Contribution.Role.EDITOR,
+                textanswer_visibility=Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS,
+            )
+
+    def check_existing_course(self, semester, importer):
```
jb3rndt, line 128: I am not sure how we should handle the generation of error messages. Putting those into `check_evaluation_data_correctness` would make that method extremely long. Generating those error messages outside of it needs the importer object and hides the places where the error messages get generated.
Another solution could be writing the error messages into `EvaluationData.errors` and convert them to importer errors later on like here: (`importers.py` Line 501-509)
```python
if "is_graded" in evaluation_data.errors:
      self.errors[ImporterError.IS_GRADED].append(
          _('"is_graded" of course {} is {}, but must be {} or {}').format(
              evaluation_data.name_en,
              evaluation_data.errors["is_graded"],
              settings.IMPORTER_GRADED_YES,
              settings.IMPORTER_GRADED_NO,
          )
      )
```

evap/staff/importers.py
```diff
@@ -70,7 +70,7 @@ def validate(self):
 
 
 @dataclass
-class EvaluationData:
+class EvaluationData:  # pylint: disable=too-many-instance-attributes
```
jb3rndt, line 73: Do you think this is valid? I dont know how to reduce the number of attributes and one could argue that this is a data class

evap/staff/importers.py
```diff
@@ -97,28 +98,85 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
+            if name_collision == "both":
+                importer.errors[ImporterError.COURSE].append(
+                    _("Course {}({}) does already exist in this semester.").format(self.name_en, self.name_de)
+                )
+            else:
+                importer.errors[ImporterError.COURSE].append(
+                    _("Course {} does already exist in this semester.").format(
+                        self.name_en if name_collision == "name_en" else self.name_de
+                    )
+                )
```
jb3rndt, line 142: Previously, the error messages for a name collision were generated depending on whether the german name already existed, or the english name. Additionally, both of them got generated when both names already existed. So I tried to keep this logic but merge the error messages if both names already existed.

evap/staff/importers.py
```diff
@@ -97,28 +98,85 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
+                    _("Course {}({}) does already exist in this semester but the only evaluation is {}graded.").format(
+                        self.name_en, self.name_de, "not " if self.is_graded else ""
+                    )
+                )
+            else:
+                importer.errors[ImporterError.COURSE].append(
+                    _(
+                        "Course {}({}) does already exist in this semester and is identical but has {} evaluations instead of 1."
+                    ).format(self.name_en, self.name_de, len(evaluations))
+                )
```
jb3rndt, line 178: Do you think these error messages are too detailed? Alternatively there could be a general error that just states that the course is similar to one that exists but differs in some attributes that must be equal. With this, these two error messages could also be merged into one

evap/staff/importers.py
niklasmohrin: I think this is fine

evap/staff/importers.py
```diff
@@ -82,6 +82,7 @@ class EvaluationData:
     is_graded: bool
     responsible_email: str
     errors: Dict
+    existing_course: Course
```
niklasmohrin, line 85: Below, you set this to `None`, so this should probably be `Optional[Course]`

evap/staff/importers.py
niklasmohrin: Sounds good! In the rare case that the German and English names collide with two different courses, this error could give the impression that it is only one. This change is for the better in most cases though, I like :)

evap/staff/importers.py
niklasmohrin: Hm, I feel like we shouldn't make this super complicated because error generation will move around in / after #1634 anyways. I think all of your solutions are fine and should work, I think that its okay to pass the importer into these methods here

... but, @He3lixxx probably has the most up-to-date opinions on the matter for reasons not disclosed here

evap/staff/importers.py
janno42: The error messages for differing attributes should list which attributes are different but don't need to include details about the values.
An explicit message for courses that already have more than one evaluation is also needed to explain this different kind of obstacle.

evap/staff/importers.py
He3lixxx: In case we merge #1634, structuring of code in this PR wouldn't matter, so then you wouldn't have to worry too much. If you want to think about useful structuring of importer code, maybe take a look at that PR, suggestions are welcome there.

jb3rndt: Ok, so I will leave it as is :)

evap/staff/importers.py
jb3rndt: Ok, sounds good. I dont know if its too much but now I included another error message if the course type or responsible person dont match

evap/staff/importers.py
```diff
@@ -97,28 +98,98 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
+                importer.errors[ImporterError.COURSE].append(
+                    _("Course {} ({}) does already exist in this semester.").format(self.name_en, self.name_de)
+                )
+            else:
+                importer.errors[ImporterError.COURSE].append(
+                    _("Course {} does already exist in this semester.").format(
+                        self.name_en if name_collision == "name_en" else self.name_de
+                    )
+                )
+        return False
```
niklasmohrin, line 143: I can't really follow what the returned value means from just reading this method, I would personally write it more like this, but this really is more of a personal preference. (this doesn't factor in any other comments I wrote)
```python
name_collision = self.check_name_collision(semester)
if not name_collision:
    return False

self.find_and_set_existing_course(semester, importer)  # (returns nothing)
if self.existing_course is None:
    if name_collision == "both":
        importer.errors[ImporterError.COURSE].append(
            _("Course {} ({}) does already exist in this semester.").format(self.name_en, self.name_de)
        )
    else:
        importer.errors[ImporterError.COURSE].append(
            _("Course {} does already exist in this semester.").format(getattr(self, name_collision))
        )

return self.existing_course is not None
```

evap/staff/importers.py
```diff
@@ -97,28 +98,98 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
+            else:
+                importer.errors[ImporterError.COURSE].append(
+                    _("Course {} ({}) does already exist in this semester but the {} {} not match.").format(
+                        self.name_en,
+                        self.name_de,
+                        " and ".join(differing_attributes),
+                        "does" if len(differing_attributes) == 1 else "do",
+                    )
+                )
+        return None
```
niklasmohrin, line 192: This happens implicitly :)

evap/staff/importers.py
```diff
@@ -97,28 +98,98 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
-        )
-        evaluation.save()
-        evaluation.contributions.create(
-            evaluation=evaluation,
-            contributor=responsible_dbobj,
-            role=Contribution.Role.EDITOR,
-            textanswer_visibility=Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS,
-        )
+        if courses.exists():
+            course = courses.first()
```
niklasmohrin, line 160: I think this does two queries when only one is needed, just check whether `course is None` after assigning `course = courses.first()` and return early :)

evap/staff/importers.py
```diff
@@ -97,28 +98,98 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
+            differing_attributes = []
+            if course.type != self.course_type:
+                differing_attributes.append("course type")
+            if len(course.responsibles.all()) != 1 or course.responsibles.first().email != self.responsible_email:
```
niklasmohrin, line 164:
```suggestion
            if course.responsibles.count() != 1 or course.responsibles.first().email != self.responsible_email:
```

evap/staff/importers.py
```diff
@@ -97,28 +98,98 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
+                            "Course {} ({}) does already exist in this semester and is identical but must have only one evaluation."
+                        ).format(self.name_en, self.name_de)
+                    )
+            else:
+                importer.errors[ImporterError.COURSE].append(
+                    _("Course {} ({}) does already exist in this semester but the {} {} not match.").format(
+                        self.name_en,
+                        self.name_de,
+                        " and ".join(differing_attributes),
+                        "does" if len(differing_attributes) == 1 else "do",
```
niklasmohrin, line 189: This could probably use `ngettext`, see https://docs.djangoproject.com/en/4.0/topics/i18n/translation/#pluralization and the examples already in the codebase

evap/staff/importers.py
```diff
@@ -97,28 +98,98 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
+            if course.type != self.course_type:
+                differing_attributes.append("course type")
+            if len(course.responsibles.all()) != 1 or course.responsibles.first().email != self.responsible_email:
+                differing_attributes.append("responsible person")
+            if not differing_attributes:
+                evaluations = Evaluation.objects.filter(course=course)
+                if len(evaluations) == 1:
+                    if evaluations.first().wait_for_grade_upload_before_publishing == self.is_graded:
+                        self.existing_course = course
+                        return course
```
niklasmohrin, line 171: I think this should probably only do one of those two things, I would only set the instance var which would go along nicely with the suggested change to `check_existing_course`

niklasmohrin (changes requested):
Quite a wall of text that I wrote together here, mostly reordering things or shortcuts :)

evap/staff/tests/test_importers.py
```diff
@@ -374,6 +387,117 @@ def test_replace_consecutive_and_trailing_spaces(self):
...
+                )
+            ),
+            1,
+        )
+
+    def test_existing_course_degree_is_added(self):
+        self.create_existing_course()
+        excel_content = excel_data.create_memory_excel_file(excel_data.test_enrollment_data_existing_course)
+
+        self.existing_course.degrees.set([Degree.objects.get(name_de="Master")])
```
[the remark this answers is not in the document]
He3lixxx: Agree -- this currently tests whether "Bachelor" will be added to the already existing "Master", we could just do it the other way around. (Or did you maybe want to test importing with both, "Bachelor" and "Master" already set?)

evap/staff/importers.py
He3lixxx: (Although, I'd say its fine to explicitly write when we want to show intent: "We thought about this fall-through, and we definitely want to return None in this case.")

evap/staff/tests/test_importers.py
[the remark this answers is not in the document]
jb3rndt: What I did there was just me being lazy to work around creating another memory_excel_file data that would have the Master degree in it so it can be added to the existing course. Would you like to see another excel file data or rather a second variant of the `existing_course` instance?

evap/staff/tests/test_importers.py
```diff
@@ -374,6 +387,117 @@ def test_replace_consecutive_and_trailing_spaces(self):
...
+                Course.objects.filter(
+                    semester=self.semester, name_de=self.existing_course.name_de, name_en=self.existing_course.name_en
+                )
+            ),
+            1,
+        )
+
+    def test_existing_course_degree_is_added(self):
+        self.create_existing_course()
+        excel_content = excel_data.create_memory_excel_file(excel_data.test_enrollment_data_existing_course)
```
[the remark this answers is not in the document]
jb3rndt: So, creating the correct `memory_excel_file` may be different for each test, and yes, in this case it is the same for each of the new tests. Imo creating the correct `memory_excel_file` directly inside the test makes it better understandable. Moving all different excel data out of the methods into class variables sounds like bloating up the test class pretty much. Just my initial throughts. I dont know of any superior reasons for this. :D (I hope this was even part of your regards)
Regarding the `self.create_existing_course()`: yes seems valid to more it into `setUpTestData`.

evap/staff/tests/test_importers.py
```diff
@@ -374,6 +387,117 @@ def test_replace_consecutive_and_trailing_spaces(self):
...
+                )
+            ),
+            1,
+        )
+
+    def test_existing_course_different_grading(self):
+        self.create_existing_course()
+        excel_content = excel_data.create_memory_excel_file(excel_data.test_enrollment_data_existing_course)
+
+        evaluation = Evaluation.objects.get(course=self.existing_course)
```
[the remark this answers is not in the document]
jb3rndt: You mean putting the Evaluation into a class variable? Yeah, can do that ^^

evap/staff/tests/test_importers.py
```diff
@@ -182,6 +182,19 @@ def setUpTestData(cls):
         Degree.objects.filter(name_de="Bachelor").update(import_names=["Bachelor", "B. Sc."])
         Degree.objects.filter(name_de="Master").update(import_names=["Master", "M. Sc."])
 
+    @classmethod
```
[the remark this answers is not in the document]
jb3rndt: 🤷‍♂️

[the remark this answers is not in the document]
jb3rndt: Removed it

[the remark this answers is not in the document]
niklasmohrin: yes, but currently we are not doing so, right?

evap/staff/importers.py
jb3rndt: Yeah its a more clearer structure, I like it

evap/staff/importers.py
jb3rndt: I think Im going to remove the returns, as the returned value is not used by `check_existing_course` anymore

evap/staff/tests/test_importers.py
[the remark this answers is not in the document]
jb3rndt: Currently not, but you suggested doing so in https://github.com/e-valuation/EvaP/pull/1683#discussion_r790988950, so I think is still needs to be a classmethod?

evap/staff/importers.py
jb3rndt: Yeah I should read all comments before replying to them one by one :D
I totally agree ^^

evap/staff/tests/test_importers.py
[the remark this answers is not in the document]
jb3rndt: Hm I am not sure about moving the `self.create_existing_course()` into `setUpTestData` because that breaks the evaluation counts for other tests. We could make those counts independent of the evaluations in the database or not move `self.create_existing_course()`. What so you think?

evap/staff/importers.py
```diff
@@ -427,13 +500,11 @@ def check_evaluation_data_correctness(self, semester):
...
         for evaluation_data in self.evaluations.values():
-            if Course.objects.filter(semester=semester, name_en=evaluation_data.name_en).exists():
-                self.errors[ImporterError.COURSE].append(
-                    _("Course {} does already exist in this semester.").format(evaluation_data.name_en)
-                )
-            if Course.objects.filter(semester=semester, name_de=evaluation_data.name_de).exists():
-                self.errors[ImporterError.COURSE].append(
-                    _("Course {} does already exist in this semester.").format(evaluation_data.name_de)
+            if evaluation_data.check_existing_course(semester, self):
+                self.warnings[ImporterWarning.DUPL].append(
```
He3lixxx, line 504: I'd say we should use a new warning category for this -- it's not really a duplicate.

evap/staff/importers.py
```diff
@@ -97,28 +98,99 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
+                importer.errors[ImporterError.COURSE].append(
+                    _("Course {} does already exist in this semester.").format(getattr(self, name_collision))
+                )
+
+        return self.existing_course is not None
+
+    def check_name_collision(self, semester):
+        name_collision = None
+        if Course.objects.filter(semester=semester, name_en=self.name_en).exists():
+            name_collision = "name_en"
```
He3lixxx, line 148: This should probably be a [Flag](https://docs.python.org/3/library/enum.html#flag). I'd say this whole method can then be inlined.

evap/staff/importers.py
```diff
@@ -97,28 +98,99 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
-            wait_for_grade_upload_before_publishing=self.is_graded,
-        )
-        evaluation.save()
-        evaluation.contributions.create(
-            evaluation=evaluation,
-            contributor=responsible_dbobj,
-            role=Contribution.Role.EDITOR,
-            textanswer_visibility=Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS,
-        )
+        course = courses.first()
```
He3lixxx, line 159: you never reuse `courses`, so you could assign `course` directly.

evap/staff/importers.py
```diff
@@ -97,28 +98,99 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
+        if course is None:
+            return
+
+        differing_attributes = []
+        if course.type != self.course_type:
+            differing_attributes.append("course type")
```
He3lixxx, line 165: Flag again.

evap/staff/importers.py
```diff
@@ -97,28 +98,99 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
+                            "Course {} ({}) does already exist in this semester but the grading of the evaluation does not match."
+                        ).format(self.name_en, self.name_de)
+                    )
+            else:
+                importer.errors[ImporterError.COURSE].append(
+                    _(
+                        "Course {} ({}) does already exist in this semester and is identical but must have exactly one evaluation."
+                    ).format(self.name_en, self.name_de)
+                )
+        else:
```
He3lixxx, line 185: invert if and else branches here to make the control flow easier to follow ([example](https://www.refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html))

evap/staff/importers.py
```diff
@@ -97,28 +98,99 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
+        self.find_and_set_existing_course(semester, importer)
+        if self.existing_course is None:
+            if name_collision == "both":
+                importer.errors[ImporterError.COURSE].append(
+                    _("Course {} ({}) does already exist in this semester.").format(self.name_en, self.name_de)
+                )
+            else:
+                importer.errors[ImporterError.COURSE].append(
+                    _("Course {} does already exist in this semester.").format(getattr(self, name_collision))
+                )
```
He3lixxx, line 141: For the sanity of the reader, I think it would make sense to create these errors where the other errors are created. Currently, following the logic of this is quite the hopping between this, `check_name_collision` and `find_and_set_existing_course`, and you have to keep 6 combinations of `name_collision` and `existing_course` in mind while doing so.

Maybe one more inversion of the flow could help with that, too:
- If there exists a course with the same name_de and name_en, handle it (add mismatch error or set existing_course) and return
- (Implicit else) If there exists a course with either the same name_de or the same name_en, handle it (add errors I think)

He3lixxx (changes requested):

evap/staff/tests/test_importers.py
```diff
@@ -374,6 +387,117 @@ def test_replace_consecutive_and_trailing_spaces(self):
...
+        EnrollmentImporter.process(excel_content, self.semester, None, None, test_run=False)
+        self.assertSetEqual(
+            set(self.existing_course.degrees.all()), set(Degree.objects.filter(name_de__in=["Master", "Bachelor"]))
+        )
+
+    def test_existing_course_users_added_to_evaluation(self):
+        self.create_existing_course()
+        excel_content = excel_data.create_memory_excel_file(excel_data.test_enrollment_data_existing_course)
+
+        self.assertEqual(0, len(self.existing_course.evaluations.all()[0].participants.all()))
```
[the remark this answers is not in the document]
He3lixxx: @jb3rndt You made this into `.all()[0]` instead of `.get()`, which would not throw if we had more than one evaluation. Was there a particular reason for that?

evap/staff/importers.py
```diff
@@ -97,28 +98,99 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
+                        "Course {} ({}) does already exist in this semester and is identical but must have exactly one evaluation."
+                    ).format(self.name_en, self.name_de)
+                )
+        else:
+            importer.errors[ImporterError.COURSE].append(
+                _("Course {} ({}) does already exist in this semester but the {} {} not match.").format(
+                    self.name_en,
+                    self.name_de,
+                    " and ".join(differing_attributes),
+                    ngettext("does", "do", len(differing_attributes)),
```
He3lixxx, line 191: This does not work for translations. In german, the words "does" and "do" would disappear -- we wouldn't have: "aber die Attribute X und Y tun nicht übereinstimmen", instead the "match" would be the word we conjugate. You will have to put a bit more context into the translated phrase to make it work -- probably two copies of the phrase, one for singular and one for plural, is what we need.

evap/staff/importers.py
He3lixxx: @janno42 any cases where users need the explicit information what language they are looking at? I could see someone complaining that the course doesn't exist when they just don't find it because they are looking at the English names instead of the German ones.

evap/staff/importers.py
He3lixxx: Just realized: If you use it only to build the message, the list (or possibly a set) is probably the more suitable option.

evap/staff/tests/test_importers.py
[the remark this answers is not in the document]
jb3rndt: I'm not sure if I miss something...
If we remove the last line here (420) (which was the suggested change, from what I understood), the `memory_excel_file` would only add a Bachelor degree to `self.existing_course`, which already has a Bachelor degree. So we could not really assert that some degree got added.
Should I change the degree of self.existing_course to Master inside the `create_existing_course` function?
I think I dont know where the Master degree should come from, if line 420 is removed.

evap/staff/tests/test_importers.py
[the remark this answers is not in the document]
jb3rndt: Nope, there was no reason. It is now
`self.assertFalse(self.existing_course_evaluation.participants.exists())`

evap/staff/tests/test_importers.py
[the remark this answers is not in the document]
jb3rndt: Adapting the existing course to be able to use the `test_enrollment_data_filedata` is a good idea.
Regarding the excel content: So I am going to put
`excel_content = excel_data.create_memory_excel_file(excel_data.test_enrollment_data_fildata)`
into `setUpTestData` also for the already existing tests.

evap/staff/tests/test_importers.py
```diff
@@ -374,6 +386,106 @@ def test_replace_consecutive_and_trailing_spaces(self):
...
+            set(self.existing_course.degrees.all()), set(Degree.objects.filter(name_de__in=["Master", "Bachelor"]))
+        )
+
+    def test_existing_course_users_added_to_evaluation(self):
+        self.create_existing_course()
+        excel_content = excel_data.create_memory_excel_file(excel_data.test_enrollment_data_existing_course)
+
+        self.assertFalse(self.existing_course_evaluation.participants.exists())
+        EnrollmentImporter.process(excel_content, self.semester, None, None, test_run=False)
+        UserProfile.objects.get(email="lucilia.manilium@institution.example.com")
```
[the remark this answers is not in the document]
jb3rndt: Seems so 😅

evap/staff/tests/test_importers.py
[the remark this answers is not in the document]
He3lixxx: Sorry, somehow, I was under the impression that the memory_excel_file contained both degrees for the course. I probably looked at the wrong line. Ignore my comments.

evap/staff/importers.py
jb3rndt: What do you think of

- `MODIF = ("modification", gettext_lazy("Possible modifications"), x)`
- `EXISTS= ("existing", gettext_lazy("Existing courses", x)`
- `ADJ= ("adjustments", gettext_lazy("Automatic adjustments", x)`

Or any combination of these?

evap/staff/importers.py
jb3rndt: I just had another thought. Why do we display different error messages if there is already a name mismatch (either name_de or name_en does not match). We could also integrate those attributes in this error message:
```
importer.errors[ImporterError.COURSE].append(
    _("Course {} ({}) does already exist in this semester but the {} {} not match.").format(
        self.name_en,
        self.name_de,
        " and ".join(differing_attributes),
        ngettext("does", "do", len(differing_attributes)),
    )
)
```
So the result would be: There is one error message that lists all differing attributes of a course, if one of name_de or name_en matches but the other does not, or if both are equal and any other attribute is different.
Would that be a viable solution?

evap/staff/importers.py
He3lixxx: I was thinking something like:
```python3
Course.objects.filter(semester=semester, name_en=self.name_en).exists() and NameCollision.NAME_EN
| Course.objects.filter(semester=semester, name_en=self.name_de).exists() and NameCollision.NAME_DE
```

@niklasmohrin what do you think about abusing the "X and Y returns Y if X is truthy" thing here?

evap/staff/importers.py
He3lixxx: In general: I think handling all the attributes equally makes sense, especially if it simplifies stuff.

The error message you listed still has the translation problem (https://github.com/e-valuation/EvaP/pull/1683#discussion_r796899279).

evap/staff/importers.py
jb3rndt: Yep, is https://github.com/e-valuation/EvaP/pull/1683/commits/8c68ddf8638d0da82430f630f19336bf0cd19be6 doing that as expected?

evap/staff/importers.py
jb3rndt: One could see the merging of degrees and adding the user to the evaluation as modification/adjustment of the course/evaluation. At least that is where the thought came from. But maybe `EXISTS` is the best out of these...

evap/staff/importers.py
jb3rndt: I hope https://github.com/e-valuation/EvaP/pull/1683/commits/54a459ef4111a5462e00b490f950cc3c448e2c30 does what was expected

evap/staff/importers.py
jb3rndt: https://github.com/e-valuation/EvaP/pull/1683/commits/8b7fbc5ba9de0aa6e0d5efcd3f0e704acf339e0b should fix this problem, right?

evap/staff/importers.py
jb3rndt: Ok, I tried to move the generation of error message into one method and also have that early check for the name_collision you suggested in your first comment (I did it the other way around, but should be the same). See https://github.com/e-valuation/EvaP/pull/1683/commits/5897433102b11c9757214fdc507fb132794e35a0.
So now there is a method that checks a given course for its attributes (`check_existing_course_attributes`) whether they match with the current evaluationData instance. Then it returns all mismatching attributes and whether the number of evaluations is correct. These information then get transformed into an error message in the calling method. (Little side note: in `check_existing_course_attributes` there is an early return for a mismatching evaluation count because if there was no early return, the check for the grading property would maybe give false information or would crash)
Does the new structure improve understandability?

Would you still like to see what I suggested here: https://github.com/e-valuation/EvaP/pull/1683#discussion_r800857244? Because I am not sure if it makes the code even more complicated.

evap/staff/tests/test_importers.py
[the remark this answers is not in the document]
jb3rndt: I'm not sure what to do about this 😅...
Is it really a problem to rely on `degrees.set(...)`? I mean yeah, it could break... But every other component in django could break as well, or is there something special or unstable about `degrees.set(...)`?

evap/staff/tests/test_importers.py
```diff
@@ -181,34 +182,54 @@ def setUpTestData(cls):
         baker.make(CourseType, name_de="Vorlesung", import_names=["Vorlesung", "V"])
         Degree.objects.filter(name_de="Bachelor").update(import_names=["Bachelor", "B. Sc."])
         Degree.objects.filter(name_de="Master").update(import_names=["Master", "M. Sc."])
+        cls.default_excel_content = excel_data.create_memory_excel_file(excel_data.test_enrollment_data_filedata)
+        cls.create_existing_course(cls)
```
niklasmohrin, line 186: You can just make `create_existing_course` a `@classmethod` and use `cls.create_existing_course()`

evap/staff/tests/test_importers.py
```diff
@@ -181,34 +182,54 @@ def setUpTestData(cls):
...
-            excel_content, self.semester, None, None, test_run=True
+            self.default_excel_content, self.semester, None, None, test_run=True
         )
-        self.assertIn("The import run will create 23 courses/evaluations and 23 users:", "".join(success_messages))
+        self.assertIn("The import run will create 23 courses/evaluations and 22 users:", "".join(success_messages))
         # check for one random user instead of for all 23
         self.assertIn("Ferdi Itaque (789@institution.example.com)", "".join(success_messages))
         self.assertEqual(errors, {})
-        self.assertEqual(warnings, {})
+        self.assertNotEqual(warnings, {})
```
niklasmohrin, line 216: Wait, why isn't `warnings` empty anymore? These tests should definitely not change, right? (I mean, this test is called `valid_file_import`)

edit: or is the excel file just wrong, according to the new logic?

evap/staff/importers.py
janno42: Yes 👍

evap/staff/importers.py
janno42: i'd say `EXISTS` suffices and you can put all of these into there

evap/staff/importers.py
```diff
@@ -97,28 +105,111 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
+        else:
+            self.existing_course = course
+
+        return self.existing_course is not None
+
+    def check_name_collision(self, semester):
+        flag = NameCollision.NONE
+        flag |= NameCollision(
+            Course.objects.filter(semester=semester, name_en=self.name_en).exists() and NameCollision.NAME_EN
         )
```
niklasmohrin, line 193: Ah, this didn't work out like we first thought; I guess we should just have 
```suggestion
        flag = NameCollision.NONE
        if Course.objects.filter(semester=semester, name_en=self.name_en).exists():
            flag |= NameCollision.NAME_EN
```

evap/staff/importers.py
```diff
@@ -97,28 +105,111 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
-            role=Contribution.Role.EDITOR,
-            textanswer_visibility=Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS,
+        flag |= NameCollision(
+            Course.objects.filter(semester=semester, name_de=self.name_de).exists() and NameCollision.NAME_DE
         )
+        return flag
+
+    def check_existing_course_attributes(self, course):
+        mismatches = []
+        has_wrong_evaluation_count = False
```
niklasmohrin, line 201: The only branch that sets this to `True` returns immediately, so it can just be replaced by `True` and `False` in the two return statements directly

niklasmohrin (changes requested):
Sorry that I have to come back with more comments, I hope we can close this PR soon ! :)

evap/staff/importers.py
niklasmohrin: If I understand correctly, the "the" and " and " are not translated at all right now

evap/staff/importers.py
```diff
@@ -97,28 +105,111 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
         # This is safe because the user's email address is checked before in the importer (see #953)
         responsible_dbobj = UserProfile.objects.get(email=self.responsible_email)
-        course = Course(
-            name_de=self.name_de,
-            name_en=self.name_en,
-            type=self.course_type,
+        if self.existing_course:
+            self.existing_course.degrees.add(*self.degrees)
+        else:
+            course = Course(
```
niklasmohrin, line 111: If you would change it to this:
```python
        if self.existing_course:
            self.existing_course.degrees.add(*self.degrees)
            return
        course = Course(
        # ...
```

you wouldn't have the large diff for the else-branch

evap/staff/importers.py
niklasmohrin: Sorry, this didn't work out in the one-line fashion we were aiming at due to the type-error reasons you listed; see the new comment I made below :sweat_smile:

evap/staff/tests/test_importers.py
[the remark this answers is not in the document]
niklasmohrin: Hm, this really shouldn't happen! Optimally, we would find out why this happens and do something in `setupTestData` to restore the old behavior, otherwise, let's just leave it with the call in every test and open an investigative-follow-up-issue. Sorry, you are really stumbling upon weird stuff in this PR.

evap/staff/tests/test_importers.py
He3lixxx: (also, I'd argue that the merge for existing courses is a special case, not a base case. Thus I'd say `cls.create_existing_course()` should not be called in `setUpTestData`, but just in tests that need it)

He3lixxx:
I'm still finding it difficult to follow the logic in the importer. We currently have 98 additional lines for that and I feel like I have to keep too many things in my head when trying to verify its correctness.

<details>
<summary>Essentially, isn't what we're trying to do basically this? (Mockup with 50 additional lines)</summary>

```python3
# both in EnrollmentImporter
def check_evaluation_data_correctness(self, semester):
    # [...]
    for evaluation_data in self.evaluations.values():
        merge_candidate = Course.objects.filter(
            semester=semester, name_de=evaluation_data.name_de, name_en=evaluation_data.name_en
        ).first()
        if merge_candidate:
            merge_hindrances = self.get_course_merge_hindrances(evaluation_data, merge_candidate)
            if merge_hindrances:
                self.errors[ImporterError.COURSE].append(
                    _(
                        "Course {name_en} ({name_de}) already exists in this semester, but the courses can not be merged for the following reasons: {merge_hindrances}"
                    ).format(
                        name_en=evaluation_data.name_en,
                        name_de=evaluation_data.name_de,
                        merge_hindrances=", ".join(merge_hindrances),
                    )
                )
            else:
                evaluation_data.existing_course = merge_candidate
                self.warnings[ImporterWarning.COURSE_MERGE].append(
                    _(
                        "Course {} ({}) already exists with identical attributes. Course is not created and users are put into the evaluation of that course."
                    ).format(evaluation_data.name_en, evaluation_data.name_de)
                )
            continue

        if Course.objects.filter(semester=semester, name_en=evaluation_data.name_en).exists():
            self.errors[ImporterError.COURSE].append(
                _("Course {} (EN) already exists in this semester with differing name_de.").format(
                    evaluation_data.name_en
                )
            )
        if Course.objects.filter(semester=semester, name_de=evaluation_data.name_de).exists():
            self.errors[ImporterError.COURSE].append(
                _("Course {} (DE) already exists in this semester with differing name_en.").format(
                    evaluation_data.name_de
                )
            )
    # old `for evaluation_data in self.evaluations.values()` loop here, without the name_de and name_en checks


def get_course_merge_hindrances(evaluation_data: EvaluationData, course: Course) -> Set[Str]:
    hindrances = set()
    if course.type != evaluation_data.course_type:
        hindrances.append(_("the course type does not match"))
    if course.responsibles.count() != 1 or course.responsibles.first().email != evaluation_data.responsible_email:
        hindrances.append(_("the responsibles of the course do not match"))

    evaluations = Evaluation.objects.filter(course=course)
    if len(evaluations) != 1:
        hindrances.append(_("the existing course does not have exactly one evaluation"))
    elif evaluations.first().wait_for_grade_upload_before_publishing != evaluation_data.is_graded:
        mismatches.append(_("the evaluation of the existing course has a mismatching grading specification"))

    return hindrances
```
</details>

evap/staff/tests/test_importers.py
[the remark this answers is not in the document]
jb3rndt: I think restoring the old behavior would either require removing the course again after creating it (which obviously does not help) or using an existing_course that is not in `default_excel_content` which in turn requires a different `excel_content` for all the new tests.
Cosidering that and what @He3lixxx said in https://github.com/e-valuation/EvaP/pull/1683#discussion_r806138012 I'll revert https://github.com/e-valuation/EvaP/commit/e789a8370541c215785c9183b41adfa55721df01. Shall I open an investigation issue after this pr is merged or is it resolved with the argument in https://github.com/e-valuation/EvaP/commit/e789a8370541c215785c9183b41adfa55721df01?

evap/staff/tests/test_importers.py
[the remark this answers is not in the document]
jb3rndt: Yep seems valid. In `test_existing_course_different_attributes`, `test_existing_course_equal_except_evaluations` and `test_existing_course_different_grading` such adjustments are done as well. Do you think one one comment in `test_existing_course_degree_is_added` is enough or should we at least add references to it in the other tests?

jb3rndt:
Yeah I agree that it is somehow more complicated than it should be... I tried to use your suggestions and the simplification from https://github.com/e-valuation/EvaP/pull/1683#discussion_r800857244. https://github.com/e-valuation/EvaP/pull/1683/commits/8d1a1c82e6cc5c69d364344c705d7f186e8fab27 is the result (at least from a structure perspective).

evap/staff/importers.py
```diff
@@ -120,6 +132,54 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
 
+    def check_for_existing_course(self, semester):
+        course = Course.objects.filter(
+            semester=semester,
+            name_en=self.name_en,
+            name_de=self.name_de,
+        ).first()
+
+        if course:
+            merge_hindraces = self.course_merge_hindraces(course)
```
niklasmohrin, line 143: This should be hindra**n**ces, in a couple of places

evap/staff/importers.py
```diff
@@ -182,6 +244,7 @@ def create(self, name_de, name_en, degree_names, course_type_name, is_graded, re
             is_graded=is_graded,
             responsible_email=responsible_email,
             errors=errors,
+            existing_course=None,
```
niklasmohrin, line 247: Note: We could make this a default on the dataclass itself by specifying `= None` on the field declaration, but it would also be a bit confusing, so let's keep it like it is right now :)

evap/staff/importers.py
```diff
@@ -427,13 +490,24 @@ def check_evaluation_data_correctness(self, semester):
...
                 self.errors[ImporterError.COURSE].append(
-                    _("Course {} does already exist in this semester.").format(evaluation_data.name_en)
+                    format_html(
+                        _(
+                            "Course {} ({}) does already exist in this semester, but the courses can not be merged for the following reasons:<br /> - {}."
+                        ).format(
+                            evaluation_data.name_en,
+                            evaluation_data.name_de,
+                            "<br /> - ".join(merge_hindraces),
+                        )
```
niklasmohrin, line 503: To get rid of the `<br />` duplication we could:
```suggestion
                        _(
                            "Course {} ({}) does already exist in this semester, but the courses can not be merged for the following reasons:{}."
                        ).format(
                            evaluation_data.name_en,
                            evaluation_data.name_de,
                            "".join(f"<br /> - {msg}" for msg in merge_hindraces),
                        )
```

evap/staff/tests/test_importers.py
```diff
@@ -374,6 +393,113 @@ def test_replace_consecutive_and_trailing_spaces(self):
...
+
+        old_course_count = Course.objects.count()
+        old_dict = model_to_dict(self.existing_course)
+
+        __, __, errors = EnrollmentImporter.process(
+            self.default_excel_content, self.semester, self.vote_start_datetime, self.vote_end_date, test_run=False
+        )
+
+        self.assertIn("the course type does not match", errors[ImporterError.COURSE][0])
+        self.assertIn("the responsibles of the course do not match", errors[ImporterError.COURSE][0])
```
niklasmohrin, line 459: What is this `[0]`? To check that this string is among any of the errors, we usually assert that it is in `"".join(errors[ImporterError.COURSE])`

niklasmohrin (commented):
Very nice, I have reached the nitpicking phase, other than that this looks good to me :+1:

evap/staff/tests/test_importers.py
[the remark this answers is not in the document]
niklasmohrin: it looks good as it is now :)

Feel free to open an issue, even if we figure it out instantly, it's nice to have it documented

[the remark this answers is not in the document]
He3lixxx: > Hm, this really shouldn't happen! Optimally, we would find out why this happens and do something in `setupTestData` to restore the old behavior, otherwise, let's just leave it with the call in every test and open an investigative-follow-up-issue. Sorry, you are really stumbling upon weird stuff in this PR.

> Feel free to open an issue, even if we figure it out instantly, it's nice to have it documented

Looking into it again, really nothing unexpected happened, it was just the combination of the two things that confused us:
- `test_enrollment_data_filedata`, the "import sheet" used for all other tests, was used for the new merge tests, thus the setup for the new tests had to create database entries for the courses that would be imported for that sheet.
- `create_exisiting_course` was moved to `setUpTestData`. For the reasons above, this method sets up the "course already exists" situation with the usual import data that normally shouldn't cause any errors -- which obviously then affects all tests, as we now basically start every test with a "course already exists" situation.

In detail:
- Warnings were unequal to {} because obviously, with one course from the import already existing, there would be a warning that the course already exists.
- Users were reduced from 23 to 22 because a user's email (123@institution.example.com) was re-used for a responsible (probably on purpose to trigger a warning / error)
- All other increments / decrements are explainable similarly.

Resolving this now: The current code is what we want (reuse error-free import data, create error/warning situation in tests that want it).

evap/staff/importers.py
He3lixxx: > Would you still like to see what I suggested here: [#1683 (comment)](https://github.com/e-valuation/EvaP/pull/1683#discussion_r800857244)? Because I am not sure if it makes the code even more complicated.

I think the current code in the PR is fine. imho the aggregation you proposed above does not work well because there could be multiple merge candidates then, and then the messages would become misleading: The candidates would likely be two different courses, and obviously the user would not want to merge with both of them.

Resolving for now.

evap/staff/tests/test_importers.py
He3lixxx: Resolving, explanation in https://github.com/e-valuation/EvaP/pull/1683#discussion_r816165654.

evap/staff/tests/test_importers.py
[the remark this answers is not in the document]
He3lixxx: (for me, all methods, including this one, would be fine without the comment)

evap/staff/importers.py
```diff
@@ -120,6 +132,55 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
+            self.existing_course = course
+            return set()
+
+        name_collision = self.check_name_collision(semester)
+        if name_collision == NameCollision.NONE:
+            return set()
+        if name_collision == NameCollision.NAME_DE:
+            return set([_("the german course name does not match")])
+        if name_collision == NameCollision.NAME_EN:
+            return set([_("the english course name does not match")])
```
He3lixxx, line 155: With this, the final message to the user would be:
> Course Shake (Schütteln) does already exist in this semester, but the courses can not be merged for the following reasons:
> - the german course name does not match

and, similarly,
> Course Shake (Schütteln) does already exist in this semester, but the courses can not be merged for the following reasons:
> - the english course name does not match

For the user, this requires quite some thinking to figure out what is actually going on, compared to
> Course Shake (english name) already exists in this semester with differing german name

and
>Course Schütteln (german name) already exists in this semester with differing english name

I'd argue: Move this into the importer, so we have access to self.errors here and are not forced to have one unified message constructed around it.

Then, I'd think we can more easily find a name, such as `set_existing_course_to_merge_or_add_errors`, or simply `handle_course_name_collisions`, or whatever floats your boat.

evap/staff/importers.py
```diff
@@ -120,6 +132,55 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
             textanswer_visibility=Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS,
         )
 
+    def check_for_existing_course(self, semester):
```
He3lixxx, line 135: I'd say the name doesn't really describe what the method does, or returns, anymore. Change suggestion a few comments below

He3lixxx (changes requested):

evap/staff/importers.py
He3lixxx: (I'd let you and @niklasmohrin overrule me here if you disagree, because I'm afraid I might be unfair here. Background: I've played around with this for a while for https://github.com/e-valuation/EvaP/pull/1683#issuecomment-1039503929 and think the general control flow structure there is close to the best we could achieve for simplest code with best user messages, apart from whether it is in an extra method or directly in the importer's loop. But, since it's my code, I'm probably biased, thus the overruling.)

evap/staff/importers.py
```diff
@@ -151,10 +212,12 @@ def __new__(cls, value, label, order):
     NAME = ("name", gettext_lazy("Name mismatches"), 1)
     INACTIVE = ("inactive", gettext_lazy("Inactive users"), 2)
     DUPL = ("duplicate", gettext_lazy("Possible duplicates"), 3)
-    IGNORED = ("ignored", gettext_lazy("Ignored duplicates"), 4)
+    MODIF = ("modification", gettext_lazy("Possible modifications"), 4)
```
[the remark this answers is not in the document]
jb3rndt: Oops, probably accidently left it in

evap/staff/importers.py
niklasmohrin: "below" - doesn't write another comment :microphone: :arrow_down:

He3lixxx: (below in code. Turns out github keeps the order comments were added at. I wanted to reference https://github.com/e-valuation/EvaP/pull/1683#discussion_r816193814)

jb3rndt: method is gone anyway :D

evap/staff/importers.py
jb3rndt: Yep, understand your point and agree, see https://github.com/e-valuation/EvaP/pull/1683/commits/79d7b1f845d6f8beab446440f2cbaaf0b8ac9ae6

evap/staff/importers.py
```diff
@@ -120,6 +132,55 @@ def store_in_database(self, vote_start_datetime, vote_end_date, semester):
...
+        ).first()
+
+        if course:
+            merge_hindrances = self.course_merge_hindrances(course)
+            if merge_hindrances:
+                return merge_hindrances
+            self.existing_course = course
+            return set()
+
+        name_collision = self.check_name_collision(semester)
```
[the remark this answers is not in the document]
jb3rndt: Yep, gone since https://github.com/e-valuation/EvaP/pull/1683/commits/79d7b1f845d6f8beab446440f2cbaaf0b8ac9ae6

evap/staff/tests/test_importers.py
```diff
@@ -385,6 +405,111 @@ def test_replace_consecutive_and_trailing_spaces(self):
...
+        self.assertFalse(existing_course_evaluation.participants.exists())
+
+        EnrollmentImporter.process(
+            self.default_excel_content, self.semester, self.vote_start_datetime, self.vote_end_date, test_run=False
+        )
+
+        self.assertIn(
+            UserProfile.objects.get(email="lucilia.manilium@institution.example.com"),
+            existing_course_evaluation.participants.all(),
+        )
```
He3lixxx, line 453: If you want to compact this a bit more: This could be part of `test_existing_course_is_not_recreated` (where we test that the corresponding warning is given) (you might want to change the test name then, maybe `test_existing_course_is_merged` or whatever) (feel free to just resolve if you want to keep it)

evap/staff/importers.py
```diff
@@ -476,6 +496,47 @@ def check_evaluation_data_correctness(self, semester):
...
+                            evaluation_data.name_de,
+                            "".join(f"<br /> - {msg}" for msg in merge_hindrances),
+                        )
+                    )
+                )
+            else:
+                evaluation_data.existing_course = merge_candidate
+                self.warnings[ImporterWarning.EXISTS].append(
+                    _(
+                        "Course {} ({}) already exists with identical attributes. Course is not created and users are put into the evaluation of that course."
```
He3lixxx, line 521: Two small things:
- "identical attributes" might be misleading, as degrees could differ.
- The message does not say that degrees will be merged.

@janno42 is this ok, or do we want a different message?

He3lixxx (commented):
Code looks good to me 👍

evap/staff/importers.py
janno42: Both things should optimally be included in the warning message :)

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

evap/staff/tests/test_importers.py
jb3rndt: Yeah I think thats ok. If the merging of the participants breaks it should hopefully be obvious by the `assertIn`, right? (Because one could argue that two things are tested at once and thus its harder to see what actually broke).
One thing I am not sure about: Is it weird that the participants merge is combined with `test_existing_course_is_not_recreated`, while the degree merge has a separate test? I would definitely not combine all three though.

niklasmohrin (approved):
After moving a couple of lines around as discussed in https://github.com/e-valuation/EvaP/pull/1683#discussion_r831454747, I don't see anything holding this back any longer :+1:

He3lixxx:
@janno42 
The counts in the success messages were not updated, we still issue these messages:
https://github.com/e-valuation/EvaP/blob/196fe16373850b2890326eca4b2bf05da1e14b76/evap/staff/importers.py#L585-L592
and
https://github.com/e-valuation/EvaP/blob/196fe16373850b2890326eca4b2bf05da1e14b76/evap/staff/importers.py#L598-L604

where self.evaluations is all evaluation in the import file, ignoring any merging that will happen. This means that the importer will say
> The import run will create N courses/evaluations

and
> Successfully created N courses/evaluations

when in fact less than N courses or evaluations were added. Is this expected what we want to happen? I think it's a bit misleading.

jb3rndt:
Quickly created a [PR](https://github.com/e-valuation/EvaP/pull/1753) for it :)

PR #1686, opened 2021-12-13
Update `Evaluation.state` values in log entries too

### LINKED ISSUE #1678: Migrate Evaluation.state in LogEntries
In 37508eb2c426, the `Evaluation.state` was turned into an `FSMIntegerField`. However, the migration didn't update the evaluations' log entries. This leads to errors in `evaluation.models.transform_log_action` when rendering log messages because old entries hold strings (e.g. `"in_evaluation"`) which results in a `KeyError` for `STATE_STR_CONVERSION` that expects a number (e.g. `State.IN_EVALUATION`).

niklasmohrin:
Closes #1678

### DIFF
--- a/evap/evaluation/migrations/0127_fix_logentry_types.py
+++ b/evap/evaluation/migrations/0127_fix_logentry_types.py
@@ -1,0 +1,19 @@
+import importlib
+from django.db import migrations
+
+original_migration_module = importlib.import_module("evap.evaluation.migrations.0123_evaluation_state_fsm_int")
+
+
+class Migration(migrations.Migration):
+    """
+    Initially, we forgot to migrate logentries in 0123, so this migration cleans this up.
+    Note that 0123 is now modified to also take care of the logentries in the first place.
+    """
+
+    dependencies = [
+        ("evaluation", "0126_add_textanswer_review_email_template"),
+    ]
+
+    operations = [
+        migrations.RunPython(original_migration_module.logentries_str_to_int, migrations.RunPython.noop),
+    ]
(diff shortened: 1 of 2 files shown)


evap/evaluation/migrations/0127_fix_logentry_types.py
```diff
@@ -0,0 +1,23 @@
+import importlib
+from django.db import migrations
+
+original_migration_module = importlib.import_module("evap.evaluation.migrations.0123_evaluation_state_fsm_int")
+
+
+def noop(_apps, _schema_editor):
+    pass
```
He3lixxx, line 8: Django already has this: https://docs.djangoproject.com/en/4.0/ref/migration-operations/#django.db.migrations.operations.RunPython.noop

evap/evaluation/migrations/0127_fix_logentry_types.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class Migration(migrations.Migration):
+    """
+    Initially, we forgot to migrate logentries in 0123, so this migration cleans this up.
+    Note that 0123 is now modified to also take care of the logentries in thee first place.
```
He3lixxx, line 14:
```suggestion
    Note that 0123 is now modified to also take care of the logentries in the first place.
```

He3lixxx (changes requested):

janno42 (approved):
✔️ Meets requirements

He3lixxx (approved):

niklasmohrin:
I noticed something else / for another PR: In the displayed log, the state names are always in English (also happens on master). This somewhat makes sense because we are displaying database value changes here and these are just these direct values, but since we are mapping them from integers to strings anyways, we _could_ also translate them while we are at it :D What do you think?

janno42:
I would agree, that we *could* translate them, but it is not really necessary. Go ahead if you want to open an issue for that.

niklasmohrin:
See #1693

PR #1688, opened 2021-12-20
Introduce MyPy

### LINKED ISSUE #1653: Static type checking for Python
There are some options, the two I know are
- mypy
- pyright

Of course, other options are open for discussion. (more options [in this article](https://www.infoworld.com/article/3575079/4-python-type-checkers-to-keep-your-code-clean.html))

It's still not clear how / whether we can add type checking that doesn't complain about the parts of the code that just don't have type annotations.

niklasmohrin:
Closes #1653

As you can see, there are some places that need a `type: ignore`, but overall I was surprised how little I had to change to make mypy pass. It seems to handle the untyped code fairly well.

How do you all feel about static type checking?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -47,6 +47,24 @@ jobs:
         run: ${{ matrix.command }}
 
 
+  mypy:
+    runs-on: ubuntu-18.04
+
+    container:
+      image: python:3.7
+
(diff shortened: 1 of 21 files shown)


He3lixxx (commented):
Good stuff, looks nice!

He3lixxx (approved):

PR #1689, opened 2021-12-20
Link to `textanswer_edit` with a button instead of in the answer text

fidoriel:


evap/staff/templates/quick-review.js
```diff
@@ -29,6 +32,7 @@ $(document).ready(() => {
             "k":            "[data-action=make_private]",
             "l":            "[data-action=hide]",
             "backspace":    "[data-action=unreview]",
+            "e":            "[data-action=textanswer_link]",
```
janno42, line 35: this action should probably better be called `textanswer_edit` (here and everywhere else)

evap/staff/templates/staff_evaluation_textanswers_quick.html
```diff
@@ -16,6 +16,7 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
                     <tr><td><kbd>K</kbd></td><td>{% trans 'Make answer private to the contributor' %}</td></tr>
                     <tr><td><kbd>L</kbd></td><td>{% trans 'Remove answer' %}</td></tr>
                     <tr><td><kbd>&#x232b;</kbd></td><td>{% trans 'Unreview answer' %}</td></tr>
+                    <tr><td><kbd>F</kbd></td><td>{% trans 'Edit answer' %}</td></tr>
```
janno42, line 19:
```suggestion
                    <tr><td><kbd>E</kbd></td><td>{% trans 'Edit answer' %}</td></tr>
```

evap/staff/templates/staff_evaluation_textanswers_quick.html
```diff
@@ -52,9 +53,8 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
                         </div>
                         {% for answer in result.answers %}
                             <div class="slider-item card-body" data-layer="2" data-id="{{ answer.id }}"{% if contributor %} data-contribution="yes"{% endif %}
-                                {% if answer.is_reviewed %}data-review="{% if answer.is_published %}publish{% elif answer.is_private %}make_private{% else %}hide{% endif %}"{% endif %}
-                            >
-                            <a class="textanswer-link" href="{% url 'staff:evaluation_textanswer_edit' semester.id evaluation.id answer.id %}">{{ answer.answer|linebreaksbr }}</a>
+                                {% if answer.is_reviewed %}data-review="{% if answer.is_published %}publish{% elif answer.is_private %}make_private{% else %}hide{% endif %}"{% endif %}>
+                            <p>{{ answer.answer|linebreaksbr }}</p>
```
janno42, line 57:
```suggestion
                            {{ answer.answer|linebreaksbr }}
```

janno42: we don't need to add a tag here

janno42: please also check the indentation inside the `div`, it currently doesn't seem right

evap/staff/templates/staff_evaluation_textanswers_section.html
```diff
@@ -11,18 +11,24 @@
...
                             <tr id="{{ answer.id }}">
                                 <td class="text-answer">
-                                    <a class="textanswer-link" href="{% url 'staff:evaluation_textanswer_edit' semester.id evaluation.id answer.id %}">{{ answer.answer|linebreaksbr }}</a>
                                     {% if answer.original_answer %}
-                                        <br />
-                                        <span class="textanswer-original">({{ answer.original_answer|linebreaksbr }})</span>
+                                        ({{ answer.original_answer }}
+                                    {% else %}
+                                        {{ answer.answer }}
+                                    {% endif %}
```
janno42, line 27: the `answer` should always be shown - the `original_answer` is then shown additionally if it exists. so the previous behavior (and styling) was correct, please only remove the `a` tag.

evap/staff/templates/staff_evaluation_textanswers_section.html
```diff
@@ -11,18 +11,24 @@
                 <table class="table table-striped{% if not forloop.last %} mb-4{% endif %}">
                     <thead>
                         <tr>
-                            <th style="width: 85%">{% trans 'Text answer' %}</th>
-                            <th style="width: 15%">{% trans 'Publish' %}</th>
+                            <th style="width: 75%">{% trans 'Text answer' %}</th>
+                            <th style="width: 15%"></th>
+                            <th style="width: 10%">{% trans 'Publish' %}</th>
```
janno42, line 16:
```suggestion
                            <th style="width: 80%">{% trans 'Text answer' %}</th>
                            <th style="width: 5%"></th>
                            <th style="width: 15%">{% trans 'Publish' %}</th>
```

janno42: the one-button column can be smaller and the "Publish" column needs a bit more space

janno42 (changes requested):

evap/staff/templates/quick-review.js
```diff
@@ -164,6 +168,7 @@ $(document).ready(() => {
             type: "POST",
             url: "{% url 'staff:evaluation_textanswers_update_publish' %}",
             data: parameters,
+            success: function(data) { if(action == "textanswer_link" && data) window.location = data; },
```
niklasmohrin, line 171: This seems a bit hacky, I don't know what we concluded the last time we talked about it. ~~We now thought that it could possibly be clearer if the button would be in a form by default (which would submit and redirect by default) and we only cancel the submission when we want to do ajax instead.~~

Even better: Always submit forms and use standard conventions to ignore the response.  According to https://stackoverflow.com/a/3283126/13679671 (I also tested with the evaluation_edit view) the browser will not reload the page if the status code is 204 - that means you could just return that for all cases except for the one where you really want to redirect. This means a small refactor is needed about the way these buttons work - I don't know how keen you are on HTML forms, but it probably shouldn't be too bad. 

If you don't want to spend more time on this, you can also leave it like this for now and we will fix it later :)

niklasmohrin (changes requested):

niklasmohrin:
Two things functionality-wise:

1. When editing an answer and saving, I start over at the first textanswer instead of where I left off
2. When clicking the edit button, the quick view still shifts to the next question just before the site reloads:
slidingbug

Not sure how much weight should be put on those things. I would prefer if both were fixed, but I leave the final call to @janno42 and others :)

evap/staff/templates/quick-review.js
```diff
@@ -154,7 +158,7 @@ $(document).ready(() => {
     }
 
     function reviewAction(action) {
-        if(index === items.length || action === "make_private" && !items.eq(index).data("contribution")) {
+        if(!action || index === items.length || action === "make_private" && !items.eq(index).data("contribution")) {
```
niklasmohrin, line 161: I don't think this is needed (anymore)?

niklasmohrin (changes requested):

fidoriel:
There is a third thing too. If you open the edit window from full review and save you will be redirected to the quick review. All those three things have been there before my PR. If you like I would fix them. But this is a very rarely used feature, if I understood correctly 10x a year and the bugs are not critical. I think it is not worth the time. You could open a issue with very low priority?

niklasmohrin:
The two things I wrote are old? How do they occur? Aren't these problems that are specific to the new feature?

It would of course be nice to have all of it fixed, but we can probably first merge this PR for the functionality and make a second issue that makes-it-right :tm:, this one shouldn't be put back for too long though

fidoriel:
I did not change the links, just where they are located.

evap/staff/templates/staff_evaluation_textanswers_section.html
```diff
@@ -11,18 +11,25 @@
...
-                                        <br />
                                         <span class="textanswer-original">({{ answer.original_answer|linebreaksbr }})</span>
+                                        {{ answer.original_answer }}
+                                    {% else %}
+                                        {{ answer.answer }}
+                                    {% endif %}
+                                </td>
+                                <td>
+                                    {% if user.is_manager %}
+                                        <a class="btn btn-sm btn-outline-secondary" href="{% url 'staff:evaluation_textanswer_edit' semester.id evaluation.id answer.id %}"><span class="fas fa-pencil-alt"></a>
```
niklasmohrin, line 32: This is not sufficient for the requirement "Editing should only be possible for managers, not for reviewers.". This change hides the functionality of editing from reviewers, they are still permitted to do so if they find the right url though (for example, as the `reviewer` user, I can still visit `staff/semester/21/evaluation/1600/textanswers#176c70ed-0a94-4d34-a51f-c50f307fbdd6` and make the change). The views responsible need to be protected with `@manager_required`.

(a test that tries to access such a view as a reviewer and asserts a failure would be nice)

evap/staff/templates/staff_evaluation_textanswers_quick.html
```diff
@@ -121,11 +127,13 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
                 </button>
             </div>
             <div class="lcr-right">
+                {% if user.is_manager %}
                 <div data-action-set="reviewing">
-                    <button data-action="unreview" class="btn btn-sm btn-outline-secondary">
-                        {% trans 'unreview' %}
+                    <button data-action="textanswer_edit" class="btn btn-sm btn-outline-secondary">
+                        <span class="fas fa-pencil-alt">
                     </button>
```
niklasmohrin, line 134: Same here

niklasmohrin (changes requested):

evap/staff/templates/staff_evaluation_textanswers_section.html
```diff
@@ -11,18 +11,25 @@
...
                             <tr id="{{ answer.id }}">
                                 <td class="text-answer">
-                                    <a class="textanswer-link" href="{% url 'staff:evaluation_textanswer_edit' semester.id evaluation.id answer.id %}">{{ answer.answer|linebreaksbr }}</a>
                                     {% if answer.original_answer %}
-                                        <br />
                                         <span class="textanswer-original">({{ answer.original_answer|linebreaksbr }})</span>
+                                        {{ answer.original_answer }}
+                                    {% else %}
+                                        {{ answer.answer }}
+                                    {% endif %}
```
janno42, line 28: if an `original_answer` exists, it is now displayed twice :)
please also pay attention to keeping a linebreak between the `answer` and the `original_answer` and use the `linebreaksbr` filter in all places where an answer is shown.

janno42 (changes requested):

evap/staff/templates/staff_evaluation_textanswers_section.html
```diff
@@ -11,18 +11,26 @@
...
                             <tr id="{{ answer.id }}">
                                 <td class="text-answer">
-                                    <a class="textanswer-link" href="{% url 'staff:evaluation_textanswer_edit' semester.id evaluation.id answer.id %}">{{ answer.answer|linebreaksbr }}</a>
                                     {% if answer.original_answer %}
+                                        {{ answer.answer }}
                                         <br />
                                         <span class="textanswer-original">({{ answer.original_answer|linebreaksbr }})</span>
+                                    {% else %}
+                                        {{ answer.answer }}
+                                    {% endif %}
```
janno42, line 29:
```suggestion
                                    {{ answer.answer|linebreaksbr }}
                                    {% if answer.original_answer %}
                                        <br />
                                        <span class="textanswer-original">({{ answer.original_answer|linebreaksbr }})</span>
                                    {% endif %}
```

evap/staff/templates/staff_evaluation_textanswers_quick.html
```diff
@@ -52,12 +53,11 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
...
-                                {% if answer.is_reviewed %}data-review="{% if answer.is_published %}publish{% elif answer.is_private %}make_private{% else %}hide{% endif %}"{% endif %}
-                            >
-                            <a class="textanswer-link" href="{% url 'staff:evaluation_textanswer_edit' semester.id evaluation.id answer.id %}">{{ answer.answer|linebreaksbr }}</a>
-                            {% if answer.original_answer %}
-                                <span class="textanswer-original">({{ answer.original_answer|linebreaksbr }})</span>
-                            {% endif %}
+                                {% if answer.is_reviewed %}data-review="{% if answer.is_published %}publish{% elif answer.is_private %}make_private{% else %}hide{% endif %}"{% endif %}>
+                                {{ answer.answer|linebreaksbr }}
+                                {% if answer.original_answer %}
+                                    <span class="textanswer-original">({{ answer.original_answer }})</span>
```
janno42, line 59:
```suggestion
                                    <span class="textanswer-original">({{ answer.original_answer|linebreaksbr }})</span>
```

janno42 (changes requested):

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

evap/staff/views.py
```diff
@@ -1535,6 +1535,11 @@ def evaluation_textanswers_update_publish(request):
         answer.hide()
     elif action == "unreview":
         answer.unreview()
+    elif action == "textanswer_edit":
+        url = reverse(
+            "staff:evaluation_textanswer_edit", args=[evaluation.course.semester.id, evaluation_id, textanswer_id]
+        )
+        return HttpResponse(url)
```
niklasmohrin, line 1542: I wonder if we should `raise PermissionDenied` (or `SuspiciousOperation`) if the requests user is not a manger to not leak ids. It is suspicious after all :thinking: @He3lixxx what do you think?

niklasmohrin (commented):
One more thought from me, other than that this looks good :+1:

evap/staff/views.py
He3lixxx: This view is accessible to reviewers only. Reviewers can see all textanswers, so they would have all textanswer IDs anyway(?).

Thinking about this -- this view will give a 500 (in 1520 and 1528) if the provided textanswer id or evaluation id does not exist, instead of giving 404. We might want to fix that.

niklasmohrin: You are right, leaking ids is not a problem :+1:. Requesting this view with `action == "textanswer_edit"` as a reviewer is still, excuse my language, sus

He3lixxx: yup, definitely sus, but since there's nothing to be gained here, I wouldn't put any additional access control.

niklasmohrin (approved):

He3lixxx (approved):
Current code is okay for now, @niklasmohrin will create a follow up issue where we'll try to refactor this to use less javascript and instead utilize html forms and behavior of browsers on different http status codes.

PR #1690, opened 2022-01-01
Fix single results having their results hidden in evaluation_detail

### LINKED ISSUE #1621: Single results' results are hidden in course detail page if not published
If a single result is shown on a course detail page of another evaluation of the same course, its results are hidden if it's not yet published. Non-published regular evaluations have their results shown, just as all published evaluations do.

The fix is simple, but I wanted to make sure this behavior is not expected...? It seemed inconsistent to me. 

See screenshot. Note how Bar's results are not shown, but "Software Architecture" and "Bar2" have their results shown.

image

karyon:
They were hidden for staff users if the results weren't published yet, which was not intended.

Fixes #1621

### DIFF
--- a/evap/results/tests/test_views.py
+++ b/evap/results/tests/test_views.py
@@ -475,15 +475,17 @@ class TestResultsSemesterEvaluationDetailView(WebTestStaffMode):
     def test_unpublished_single_results_show_results(self):
         """Regression test for #1621"""
         # make regular evaluation with some answers
+        participants = baker.make(UserProfile, _bulk_create=True, _quantity=20)
         evaluation = baker.make(
-            Evaluation, state=Evaluation.State.REVIEWED, course=baker.make(Course, semester=self.semester)
+            Evaluation,
+            state=Evaluation.State.REVIEWED,
+            course=baker.make(Course, semester=self.semester),
+            participants=participants,
+            voters=participants,
         )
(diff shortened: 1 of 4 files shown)


janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

evap/results/tests/test_views.py
```diff
@@ -473,6 +472,46 @@ def test_preview_with_rating_answers(self):
...
+        # make regular evaluation with some answers
+        evaluation = baker.make(
+            Evaluation, state=Evaluation.State.REVIEWED, course=baker.make(Course, semester=self.semester)
+        )
+        questionnaire = baker.make(Questionnaire, type=Questionnaire.Type.TOP)
+        likert_question = baker.make(Question, type=Question.LIKERT, questionnaire=questionnaire, order=1)
+        evaluation.general_contribution.questionnaires.set([questionnaire])
+        participants = baker.make(UserProfile, _bulk_create=True, _quantity=20)
+        evaluation.participants.set(participants)
+        evaluation.voters.set(participants)
```
He3lixxx, line 486: could in theory directly be passed to the `baker.make` call for `evaluation`, or is there some related-objects-not-saved-blah issue I don't see? The objects should be stored and have their pk set here. Same for below (499-500)

He3lixxx (approved):
Noice

niklasmohrin (approved):

PR #1692, opened 2022-01-02
Fix csrf protection

### LINKED ISSUE #1645: The navbar cache is at it again: CSRF super rude failure
TLDR: After #1616 we have a form in the navbar again, invalidating the evaluation at https://github.com/e-valuation/EvaP/issues/1208#issuecomment-398153367. Thus, we made a full cycle back to #862. It is exactly the same problem.

---

From the makers of _[CSRF problems on language change](https://github.com/e-valuation/EvaP/issues/862)_ and _[CSRF 2: Navbar cache refresh is broken](https://github.com/e-valuation/EvaP/issues/1208)_ and the creator of [the infamous language change spinner](https://github.com/e-valuation/EvaP/pull/1616): The all new _CSRF 3: CSRF super rude failure_.

The navbar cache is at it again. In this iteration of the renown franchise, doubt reigns upon the once-beloved django feature. How will the EvaP developer team respond this time? Will they fix the problem in exactly the same way they did 5 years ago? Or will they eliminate the navbar cache once and for all?

karyon:
This adds the CSRF token to all form via javascript instead of the templates, similar to how it's done already for all AJAX requests.

Fixes #1645. It became a bit hackier than I hoped and I don't mind if you reject this. Also you have to believe me I'm really not trying to save the navbar caching here :D 

The upside is that this removes a bunch of boilerplate code (in separate commits for easier review) (also it's caching-friendly, which might not count). The downside is that this prevents us from testing CSRF verification in our python tests, I had to remove a test that does this. As far as I understand, the TS tests we have cannot actually visit a page, interact with it, submit a form, and visit the redirect page. 

On the other hand, with this proposal, all previous breakages wouldn't have happened in the first place.

I believe we're not losing any security here as the validation is done on the server and that part is unchanged.

### DIFF
--- a/evap/settings.py
+++ b/evap/settings.py
@@ -249,6 +249,7 @@ TEMPLATES = [
                 "django.template.context_processors.static",
                 "django.template.context_processors.request",
                 "django.contrib.messages.context_processors.messages",
+                "evap.context_processors.set_csrf_cookie",
                 "evap.context_processors.slogan",
                 "evap.context_processors.debug",
             ],
@@ -392,7 +393,7 @@ try:
 except ImportError:
     pass
 
-TESTING = "test" in sys.argv
+TESTING = "test" in sys.argv or "render_pages" in sys.argv
 
 # speed up tests
 if TESTING:
--- a/evap/static/ts/src/csrf-utils.ts
+++ b/evap/static/ts/src/csrf-utils.ts
@@ -35,25 +35,18 @@ $.ajaxSetup({
 // Add CSRF token to regular forms on submit.
 // This reduces boilerplate code in the templates and makes the form's HTML cacheable.
 for (const form of document.forms) {
+    let submitter: HTMLElement | null;
     form.addEventListener("submit", (ev: Event) => {
-        // in case a form gets submitted twice
-        if (form.querySelectorAll("input[name=csrfmiddlewaretoken]").length > 0) {
-            return;
-        }
-
-        const { submitter } = ev as SubmitEvent;
-        const action = submitter?.getAttribute("formaction") ?? form.getAttribute("action");
-        const method = submitter?.getAttribute("formmethod") ?? form.getAttribute("method");
+        submitter = (ev as SubmitEvent).submitter;
+    });
+    form.addEventListener("formdata", (ev: Event) => {
+        const action = submitter?.getAttribute("formaction") ?? form.action;
+        const method = submitter?.getAttribute("formmethod") ?? form.method;
 
         if (method && doesMethodRequireCsrfToken(method) && action && !isUrlCrossDomain(action)) {
-            const inputElement = document.createElement("input");
-            inputElement.setAttribute("type", "hidden");
-            inputElement.setAttribute("name", "csrfmiddlewaretoken");
-            inputElement.setAttribute("value", csrftoken);
-
-            form.insertAdjacentElement("afterbegin", inputElement);
+            (ev as FormDataEvent).formData.set("csrfmiddlewaretoken", csrftoken);
         }
-    });
+    })
 }
 
 export const testable = {
--- a/evap/static/ts/src/csrf-utils.ts
+++ b/evap/static/ts/src/csrf-utils.ts
@@ -13,9 +13,9 @@ function getCookie(name: string): string | null {
 }
 const csrftoken = getCookie("csrftoken")!;
 
-function isMethodCsrfSafe(method: string): boolean {
+function doesMethodRequireCsrfToken(method: string): boolean {
     // these HTTP methods do not require CSRF protection
-    return ["GET", "HEAD", "OPTIONS", "TRACE"].includes(method.toUpperCase());
+    return ["POST", "PUT", "DELETE", "PATCH"].includes(method.toUpperCase());
 }
 
 function isUrlCrossDomain(url: string): boolean {
@@ -25,35 +25,38 @@ function isUrlCrossDomain(url: string): boolean {
 // setup ajax sending csrf token
 $.ajaxSetup({
     beforeSend: function (xhr: JQuery.jqXHR, settings: JQuery.AjaxSettings) {
-        const isMethodSafe = settings.method && isMethodCsrfSafe(settings.method);
-        if (!isMethodSafe && !this.crossDomain) {
+        const methodRequiresCsrfToken = settings.type && doesMethodRequireCsrfToken(settings.type);
+        if (methodRequiresCsrfToken && !this.crossDomain) {
             xhr.setRequestHeader("X-CSRFToken", csrftoken);
         }
     },
 });
 
-function addCsrfToken(this: HTMLFormElement, ev: Event) {
-    const inputElement = document.createElement("input");
-    inputElement.setAttribute("type", "hidden");
-    inputElement.setAttribute("name", "csrfmiddlewaretoken");
-    inputElement.setAttribute("value", csrftoken);
-
-    const { submitter } = ev as SubmitEvent;
-    const action = (submitter && submitter.getAttribute("formaction")) || this.getAttribute("action");
-    const method = (submitter && submitter.getAttribute("formmethod")) || this.getAttribute("method");
-
-    if (method && !isMethodCsrfSafe(method) && action && !isUrlCrossDomain(action)) {
-        this.insertAdjacentElement("afterbegin", inputElement.cloneNode() as Element);
-    }
-}
-
 // Add CSRF token to regular forms on submit.
 // This reduces boilerplate code in the templates and makes the form's HTML cacheable.
 for (const form of document.forms) {
-    form.addEventListener("submit", addCsrfToken);
+    form.addEventListener("submit", (ev: Event) => {
+        // in case a form gets submitted twice
+        if (form.querySelectorAll("input[name=csrfmiddlewaretoken]").length > 0) {
+            return;
+        }
+
+        const { submitter } = ev as SubmitEvent;
+        const action = submitter?.getAttribute("formaction") ?? form.getAttribute("action");
+        const method = submitter?.getAttribute("formmethod") ?? form.getAttribute("method");
+
+        if (method && doesMethodRequireCsrfToken(method) && action && !isUrlCrossDomain(action)) {
+            const inputElement = document.createElement("input");
+            inputElement.setAttribute("type", "hidden");
+            inputElement.setAttribute("name", "csrfmiddlewaretoken");
+            inputElement.setAttribute("value", csrftoken);
+
+            form.insertAdjacentElement("afterbegin", inputElement);
+        }
+    });
 }
 
 export const testable = {
     getCookie,
-    isMethodCsrfSafe,
+    doesMethodRequireCsrfToken,
 };
--- a/evap/static/ts/src/csrf-utils.ts
+++ b/evap/static/ts/src/csrf-utils.ts
@@ -28,6 +28,15 @@ $.ajaxSetup({
     },
 });
 
+const inputElement = document.createElement("input");
+inputElement.setAttribute("type", "hidden");
+inputElement.setAttribute("name", "csrfmiddlewaretoken");
+inputElement.setAttribute("value", csrftoken);
+
+for (const form of document.forms) {
+    form.insertAdjacentElement("afterbegin", inputElement.cloneNode() as Element);
+}
+
 export const testable = {
     getCookie,
     isMethodCsrfSafe,
--- a/evap/static/ts/src/csrf-utils.ts
+++ b/evap/static/ts/src/csrf-utils.ts
@@ -15,7 +15,11 @@ const csrftoken = getCookie("csrftoken")!;
 
 function isMethodCsrfSafe(method: string): boolean {
     // these HTTP methods do not require CSRF protection
-    return ["GET", "HEAD", "OPTIONS", "TRACE"].includes(method);
+    return ["GET", "HEAD", "OPTIONS", "TRACE"].includes(method.toUpperCase());
+}
+
+function isUrlCrossDomain(url: string): boolean {
+    return new URL(document.baseURI).origin !== new URL(url, document.baseURI).origin;
 }
 
 // setup ajax sending csrf token
@@ -28,13 +32,25 @@ $.ajaxSetup({
     },
 });
 
-const inputElement = document.createElement("input");
-inputElement.setAttribute("type", "hidden");
-inputElement.setAttribute("name", "csrfmiddlewaretoken");
-inputElement.setAttribute("value", csrftoken);
+function addCsrfToken(this: HTMLFormElement, ev: Event) {
+    const inputElement = document.createElement("input");
+    inputElement.setAttribute("type", "hidden");
+    inputElement.setAttribute("name", "csrfmiddlewaretoken");
+    inputElement.setAttribute("value", csrftoken);
 
+    const { submitter } = ev as SubmitEvent;
+    const action = (submitter && submitter.getAttribute("formaction")) || this.getAttribute("action");
+    const method = (submitter && submitter.getAttribute("formmethod")) || this.getAttribute("method");
+
+    if (method && !isMethodCsrfSafe(method) && action && !isUrlCrossDomain(action)) {
+        this.insertAdjacentElement("afterbegin", inputElement.cloneNode() as Element);
+    }
+}
+
+// Add CSRF token to regular forms on submit.
+// This reduces boilerplate code in the templates and makes the form's HTML cacheable.
 for (const form of document.forms) {
-    form.insertAdjacentElement("afterbegin", inputElement.cloneNode() as Element);
+    form.addEventListener("submit", addCsrfToken);
 }
 
 export const testable = {
--- a/evap/context_processors.py
+++ b/evap/context_processors.py
@@ -1,6 +1,7 @@
 import random
 
 from django.conf import settings
+from django.middleware.csrf import get_token
 from django.utils.translation import get_language
 
 
@@ -12,3 +13,11 @@ def slogan(request):
 
 def debug(request):
     return {"debug": settings.DEBUG}
+
+
+def set_csrf_cookie(request):
+    # this does not add anything to the context, but ensures that the
+    # csrf cookie is set in every view. It is only set if
+    # get_token is called, which we don't do in our templates anymore.
+    get_token(request)
+    return {}
--- a/evap/contributor/templates/contributor_evaluation_form.html
+++ b/evap/contributor/templates/contributor_evaluation_form.html
@@ -27,7 +27,6 @@
     </h3>
 
     <form method="POST" class="form-horizontal multiselect-form select2form" id="evaluation-form">
-        {% csrf_token %}
         <div class="card mb-3">
             <div class="card-body">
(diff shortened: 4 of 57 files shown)


He3lixxx (commented):
We don't have any tests for CSRF checks now, do we?

I feel a bit uncomfortable with that. A simple whitespace change (unindenting the line removing the middleware) could disable the checks, also on production. I could see that slipping through for a maliciously crafted PR that makes us use "hide whitespace changes" in review.

I feel like I've been tempted to use a custom base class for our tests at some occasions now, here, it could be useful as well: We could always enable the middleware, but have `csrf_checks = False` (or similar) in the baseclass. Then, we could also still have a test for CSRF checks with the "production" middleware configuration. What do you think?

(Edit: In general, I think doing the csrf stuff globally in JS is even somewhat more elegant, because it's less repetitive)

evap/settings.py
```diff
@@ -418,6 +419,10 @@
     # give random char field values a reasonable length
     BAKER_CUSTOM_FIELDS_GEN = {"django.db.models.CharField": lambda: random_gen.gen_string(20)}
 
+    # csrf tokens are added in JS and do not work in the tests at all. The django test runner
+    # does disable CSRF protection by default, but the WebTest app does not. So we disable it here.
+    MIDDLEWARE = [m for m in MIDDLEWARE if m != "django.middleware.csrf.CsrfViewMiddleware"]
```
niklasmohrin, line 424: Why not `MIDDLEWARE.remove("django.middleware.csrf.CsrfViewMiddleware")`? This would give us an error if this was somehow not in there already

evap/static/ts/src/csrf-utils.ts
```diff
@@ -28,6 +28,15 @@ $.ajaxSetup({
     },
 });
 
+var inputElement = document.createElement("input");
```
niklasmohrin, line 31:
```suggestion
let inputElement = document.createElement("input");
```

evap/static/ts/src/csrf-utils.ts
```diff
@@ -28,6 +28,15 @@ $.ajaxSetup({
...
+inputElement.setAttribute("type", "hidden");
+inputElement.setAttribute("name", "csrfmiddlewaretoken");
+inputElement.setAttribute("value", csrftoken);
+
+for (const key of Array.from(document.getElementsByTagName("form"))) {
+    key.insertAdjacentHTML("afterbegin", inputElement.outerHTML);
+}
```
niklasmohrin, line 38: I think the following should be equivalent:

```suggestion
for (const form of document.forms) {
    form.insertAdjacentElement("afterbegin", inputElement);
}
```

I am unsure if it is bad to insert the same element multiple times, I think not, but then there is `inputElement.cloneNode() as Element`. Don't know, maybe serializing and deserializing is best after all :thinking:

evap/settings.py
```diff
@@ -418,6 +419,10 @@
...
+
```
niklasmohrin, line 425: A small defense against the unindent-attack that @He3lixxx mentioned could be to insert a `assert TESTING` here, so that indentation is broken if only the middleware line is unindented but the assertion fails if both are

Another thought: Will malicious whitespace-only diffs still tell us if a line was joined with the (commented) line above?

niklasmohrin (commented):
I like the approach, I don't know what the django built in does right now (supposedly something more static?), and I wouldn't know whether anything could be (security-)wrong about inserting the token dynamically. Maybe this is obviously correct, but depending on how (un)sure you all are, we could maybe reach out to the cybersec club and ask them?

If @He3lixxx 's idea with the base class works out as simple as it sounds, I would be in favor of that too :)

About navigating the site in tests: It sounds like someone should sit down and setup all of the selenium and maybe `LiveServerTestCase` thingies (which is the best option according to [three people on the internet](https://stackoverflow.com/a/30607252/13679671) - they were the first three that I found but what they say looks very fitting). I think we can do that later though

Finally, I already have some comments on the code, nothing grave, probably unnecessary until general discussion is done :sweat_smile:

evap/static/ts/src/csrf-utils.ts
He3lixxx:
```suggestion
const inputElement = document.createElement("input");
```

evap/settings.py
karyon: good idea. looks a bit odd but I mean why not.

I'm pretty sure joining lines does not count as whitespace-change.

evap/static/ts/src/csrf-utils.ts
karyon: insert seems to move, not copy, the element. so the clone is actually necessary.

evap/settings.py
karyon: because I didn't know that exists :)

karyon:
@He3lixxx if the only concern is a malicious/accidental removal of the middleware, I think the additional assert should be sufficient. 

If we had a test with the setting enabled, it can't fully verify whether CSRF protection is working, because it's not running JS code. The best it can do is see that the `Set-Cookie: csrftoken=foobar` header is set.

Do you remember where else you wanted to have the base class? I also have a similar fuzzy memory but don't remember the details :( In any case I can totally do that if you want. FWIW, for this PR here we would only need to subclass WebTest. 

@niklasmohrin `{% csrf_token %}` expands to `<input type="hidden" name="csrfmiddlewaretoken" value="kDUGYI5bsUnXBL7IBC4KF1RseQQvYwvf8NtS6fsixVgsoAs7hmFrhmNaiEFkpOWN">`. The token is similar to the one in the cookie (not identical though, but they are derived from the same secret). The JS code I added pretty much does what we're already doing for AJAX requests, which follows what Django's documentation recommends for that case: https://docs.djangoproject.com/en/4.0/ref/csrf/#ajax

niklasmohrin:
Ah, right, dynamically inserting the token into requests is old news. Thanks!

niklasmohrin:
Some things are not working when testing:

1. (of course compile new typescript stuff and clear caches)
2. log in as evap
3. visit /staff by clicking on overview or visit a semester
4. Permission denied

additionally, after step 2, I cannot change languages, exit staff mode etc. with csrf errors

karyon:
Are you *sure* you cleared the default cache? Because that must be done because of, well, the navbar cache. With clearing the cache I can't repro. It would be great if you could check the repro steps!

niklasmohrin:
ah, you are right, we only tried clearing browser caches, not the django cache; well that resolves it :+1:

He3lixxx:
> @He3lixxx if the only concern is a malicious/accidental removal of the middleware, I think the additional assert should be sufficient, what do you think?

It would work, but I still think its a bit hacky.  test that the csrf middleware works as expected are more beautiful™

> If we had a test with the setting enabled, it can't fully verify whether CSRF protection is working, because it's not running JS code. The best it can do is see that the `Set-Cookie: csrftoken=foobar` header is set.

I didn't think about a "it works test" (requests with the correct csrf token will be accepted), but rather a simple "security is given" test that asserts that with a wrong or missing csrf token, the requests will be denied.

> Do you remember where else you wanted to have the base class? I also have a similar fuzzy memory but don't remember the details :( In any case I can totally do that if you want. FWIW, for this PR here we would only need to subclass WebTest.

No, sadly I don't remember the details. Yes, subclassing WebTest and simply setting `csrf_checks = False` was exactly what I was thinking. The "security test" could then simply overwrite that to true again try to post something, and assert that it fails. This has the advantage that we test the csrf middleware, exactly how it is configured in settings.py, and there is no fishy removal code that we'd need to worry about.

karyon:
Done! Commits 3 and 4 are new, replacing the previous commits that dealt with the tests.

evap/static/ts/src/csrf-utils.ts
```diff
@@ -28,6 +28,15 @@ $.ajaxSetup({
...
 });
 
+const inputElement = document.createElement("input");
+inputElement.setAttribute("type", "hidden");
+inputElement.setAttribute("name", "csrfmiddlewaretoken");
+inputElement.setAttribute("value", csrftoken);
+
+for (const form of document.forms) {
+    form.insertAdjacentElement("afterbegin", inputElement.cloneNode() as Element);
+}
```
He3lixxx, line 38: After reading through django's documentation about csrf protection, two more considerations:
- We do not need this if the form method is safe. In staff_semester_view.html, we currently have a form with method="GET". I think this would add the csrftoken as a GET parameter, so the token would be visible on the users screen (in the url bar), making screen sharing unsafe. I'd say we want to prevent that.
- [django documentation on CSRF token](https://docs.djangoproject.com/en/4.0/ref/csrf/#how-to-use-it):
  > This should not be done for POST forms that target external URLs, since that would cause the CSRF token to be leaked, leading to a vulnerability.
  
  If we ever have any form with `action="https://hpi.de/grade-download"` or whatever, we would expose that token. I quickly grepped the source code and currently, I couldn't find any forms with external targets.
  
So, I'd add some kind of check that we're sending a POST request, and it's going to our server. In the existing ajax code, [we check that we have an unsafe method and its not a crossdomain request](https://github.com/e-valuation/EvaP/blob/main/evap/static/ts/src/csrf-utils.ts#L25) before attaching the CSRFToken Header.

Checking the method should be a simple string comparison. Checking for cross domain requests is trickier, and the [number one entry on google, linking to stackoverflow](https://stackoverflow.com/questions/10687099/how-to-test-if-a-url-string-is-absolute-or-relative) makes it look like a futile undertaking, but I feel like abusing the [URL api](https://developer.mozilla.org/en-US/docs/Web/API/URL) should work. This is also what the [5th answer](https://stackoverflow.com/a/57047786/12345551) suggests, after scrolling 7 browser pages.

karyon: Very good points, thank you for going in-depth on this. I've fixed this in the last commit.

karyon: ...I just learned you can override the form action and method using `formaction` and `formmethod` on the submit buttons. Even though we're not using that, I added support for that. While doing that, I also discovered all this can be done lazily  which is nicer in my opinion.

karyon: If all this becomes too much complexity, I'm fine with not merging it!

evap/static/ts/src/csrf-utils.ts
```diff
@@ -28,6 +32,27 @@ $.ajaxSetup({
...
+
+    if (method && !isMethodCsrfSafe(method) && action && !isUrlCrossDomain(action)) {
+        this.insertAdjacentElement("afterbegin", inputElement.cloneNode() as Element);
+    }
+}
+
+// Add CSRF token to regular forms on submit.
+// This reduces boilerplate code in the templates and makes the form's HTML cacheable.
+for (const form of document.forms) {
+    form.addEventListener("submit", addCsrfToken);
```
niklasmohrin, line 53: I think it would be easier to inline the function definition here and use `form` instead of `this` (it took me a minute to find out what the syntax for `this` type hints means :D)

evap/static/ts/src/csrf-utils.ts
```diff
@@ -28,6 +32,27 @@ $.ajaxSetup({
...
+    inputElement.setAttribute("type", "hidden");
+    inputElement.setAttribute("name", "csrfmiddlewaretoken");
+    inputElement.setAttribute("value", csrftoken);
+
+    const { submitter } = ev as SubmitEvent;
+    const action = (submitter && submitter.getAttribute("formaction")) || this.getAttribute("action");
+    const method = (submitter && submitter.getAttribute("formmethod")) || this.getAttribute("method");
+
+    if (method && !isMethodCsrfSafe(method) && action && !isUrlCrossDomain(action)) {
+        this.insertAdjacentElement("afterbegin", inputElement.cloneNode() as Element);
```
niklasmohrin, line 46: Since we now create the `inputElement` on each submit, we don't need to clone anymore; the creation of the element can also be moved into this branch :)

There is one problem though: if we have forms that get submitted multiple times without the site reloading, we insert multiple inputs for the csrf token. One such example view would be the refactor planned in #1698.

I think using the `formdata` event would not have this problem, but I don't think we can check for submitter and target origin in the `formdata` handler sooo :man_shrugging:

evap/static/ts/src/csrf-utils.ts
```diff
@@ -15,7 +15,11 @@ const csrftoken = getCookie("csrftoken")!;
 
 function isMethodCsrfSafe(method: string): boolean {
     // these HTTP methods do not require CSRF protection
-    return ["GET", "HEAD", "OPTIONS", "TRACE"].includes(method);
+    return ["GET", "HEAD", "OPTIONS", "TRACE"].includes(method.toUpperCase());
```
niklasmohrin, line 18: @He3lixxx said it earlier today, and I agree: We should allow-list the methods that do get the token and not the other way around; if we ever fail to send a token, django will, I hope, happily tell us about it :^)

evap/static/ts/src/csrf-utils.ts
```diff
@@ -28,6 +32,27 @@ $.ajaxSetup({
...
 });
 
+function addCsrfToken(this: HTMLFormElement, ev: Event) {
+    const inputElement = document.createElement("input");
+    inputElement.setAttribute("type", "hidden");
+    inputElement.setAttribute("name", "csrfmiddlewaretoken");
+    inputElement.setAttribute("value", csrftoken);
+
+    const { submitter } = ev as SubmitEvent;
+    const action = (submitter && submitter.getAttribute("formaction")) || this.getAttribute("action");
```
niklasmohrin, line 42: This can use the really fancy not-so-new (I was really surprised that not only ts, but also browsers have them already) operators:
```javascript
const action = submitter?.getAttribute("formaction") ?? form.getAttribute("action");
```

evap/static/ts/src/csrf-utils.ts
karyon: I added an explicit check.

karyon:
All done, please review again!

evap/static/ts/src/csrf-utils.ts
```diff
@@ -25,35 +25,38 @@ function isUrlCrossDomain(url: string): boolean {
 // setup ajax sending csrf token
 $.ajaxSetup({
     beforeSend: function (xhr: JQuery.jqXHR, settings: JQuery.AjaxSettings) {
-        const isMethodSafe = settings.method && isMethodCsrfSafe(settings.method);
-        if (!isMethodSafe && !this.crossDomain) {
+        const methodRequiresCsrfToken = settings.type && doesMethodRequireCsrfToken(settings.type);
```
karyon, line 28: weirdly, settings.method was undefined in the js tests, only settings.type was set, which is supposed to be an alias to method. I did not dig into this further.

He3lixxx: @pixunil any idea why this would be the case?

He3lixxx (approved):
I like it now. Although, I'd be interested in why the `type`/`method` thing happens.

evap/static/ts/src/csrf-utils.ts
```diff
@@ -13,22 +13,50 @@ function getCookie(name: string): string | null {
...
+for (const form of document.forms) {
+    form.addEventListener("submit", (ev: Event) => {
+        // in case a form gets submitted twice
+        if (form.querySelectorAll("input[name=csrfmiddlewaretoken]").length > 0) {
+            return;
+        }
+
+        const { submitter } = ev as SubmitEvent;
+        const action = submitter?.getAttribute("formaction") ?? form.getAttribute("action");
+        const method = submitter?.getAttribute("formmethod") ?? form.getAttribute("method");
```
niklasmohrin, line 46: Consider the following scenario:
1. There is a form with multiple submit-inputs, one of which has `formmethod=post`, and another with `formmethod=get`
2. First, we submit with the post method which inserts the token
3. the server responds with http 204 and the site doesn't reload (or it doesn't reload for any other reason)
4. now, the token is in the form and subsequent submits with the get-input will contain the token :(

The same problem occurs with `formaction` which eliminates the `isUrlCrossDomain` check on subsequent calls.

@He3lixxx and I weren't able to come up with a satisfying solution that fits into the current model. Richard identified that this currently does two things when we really only cared about one:

1. Insert the _value_ of the token dynamically
2. Insert the _field_ for the token dynamically

If we went back to having a `{{ evap_dynamic_csrf }}` or so, we could still insert the value dynamically, but would still have the responsibility for deciding whether or not to include csrf tokens in the form code itself. This is analogous to the responsibilities when using the django csrf tag.

niklasmohrin (changes requested):

evap/static/ts/src/csrf-utils.ts
karyon: Doesn't the same problem occur with the previously used {{csrf_token}} as well?

I don't think we need to support this. I can try and add an assert to catch this if you want.

I don't think the cacheability alone is worth the added complexity of this PR. So the point of this PR *is* to insert everything dynamically for easier development and to avoid the boilerplate code.

He3lixxx: Hmm, tough situation.

> Doesn't the same problem occur with the previously used {{csrf_token}} as well?

My thoughts were:
Any form that mixes POST and GET is weird. Any form that has external or mixed endpoints/actions is weird. I don't see how we'd ever need something like that. However, if someone opened a PR with a good reason, we might accept it without thinking about security implications. Now, if the PR contains the `{{csrf_token}}` in this form, we'd be more likely to think about it compared to if the token was added silently.
(One could argue that a piece of javascript that changes a form's action or method has the same issue, though.)

> I don't think the cacheability alone is worth the added complexity of this PR. So the point of this PR is to insert everything dynamically for easier development and to avoid the boilerplate code.

Easier Development is usually better, but in this case, I'd say it gets us closer to doing security stuff wrong, so personally, I wouldn't do it -- I feel like this PR shows how much we're playing with fire here.

(For #1645, a clear_navbar_cache call at the right location should also work to fix the issue for now)

karyon: I tried the formdata event. The cool thing is that the FormData is constructed from scratch for each request, so submitting a form multiple times is not an issue anymore, and no html element construction is necessary anymore. I still had to use the submit event to get the submitter, though.

To put the "playing with fire" into perspective: The actual CSRF verification is done on the server and unchanged, so this PR cannot make us vulnerable to missing tokens. Regarding leaking the token somewhere where it shouldn't be leaked: Consider that it is already stored in the cookie, which is not an uber secure place to begin with. I doubt it would make anything considerably more insecure if we e.g. accidentally put the CSRF token into a GET request (though I do believe I have made that pretty much impossible with my recent change).

@He3lixxx and/or @niklasmohrin can you please make a call? If you give me a thumbs up, I'd summarize all the edge cases we found into a comment before merging. Otherwise I'd close this, which is also fine.

He3lixxx:
(from https://github.com/e-valuation/EvaP/pull/1692#discussion_r802119846)
> @He3lixxx and/or @niklasmohrin can you please make a call? If you give me a thumbs up, I'd summarize all the edge cases we found into a comment before merging. Otherwise I'd close this, which is also fine.

I discussed with @niklasmohrin and we think that overall, we will probably be happier with not merging, so I'm closing this for now. Still, thank you @karyon for going through our concerns and creating the different versions, I learned a thing or two in the process.

karyon:
all good, I also learned a lot!

PR #1695, opened 2022-01-10
Fix #1691: Make evaluation weight add up to 100%

### LINKED ISSUE #1691: Evaluation weight doesn't add up to 100%
Split out from #1621. See screenshot:
image

Kakadus:
Fixes #1691 by sending an additional subquery as described in @karyon's comment


janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

janno42:
why does swapping the filter and the annotate work now although @karyon wrote it returns the sum multiplied by the number of evaluations per course for him?

Kakadus:
I do not know. However, when only swapping the annotate, it is still sometimes wrong

He3lixxx:
> why does swapping the filter and the annotate work now although @karyon wrote it returns the sum multiplied by the number of evaluations per course for him?

The order of these two doesn't matter anymore, as the filter does not work by joining evaluations anymore. The old query was something like:

```SQL
select course.id, sum(evaluation.weight)
from course inner join evaluation
where evaluation.id in [...]
group by course.id
```

so here, the grouping and summing interfered with the filtering.
The new query results in something like

```SQL
select course.id, sum(evaluation.weight)
from course inner join evaluation
where course.id in (
    select course.id 
    from course inner join evaluation
    where evaluation.id in [...]
)
group by course.id
```

where the filtering does not have any impact on the joined evaluations in the context where the grouping and summing happens

niklasmohrin (approved):
Everything you guys write sounds right, there isn't much to be wrong about the code :+1:

He3lixxx (commented):
The code looks good to me.

I'd like to have a regression test (create a szenario where the displayed weights were wrong, just as in the screenshot, and assert that they are now correct) -- seems like the displayed weights are untested right now, otherwise we should have noticed.

evap/results/tests/test_views.py
```diff
@@ -264,6 +265,38 @@ def make_course_with_evaluations(unique_suffix):
         caches["sessions"].clear()
         caches["results"].clear()
 
+    def test_evaluation_weight_sums(self):
+        """
+        ensures that the sum of evaluation weights add up to 100%,
+        even if some evaluations are not shown to the user
```
He3lixxx, line 271: In this test, the `student` user can see all the evaluations, right? That is a mismatch to this comment. If I remember correctly, the wrong display occurred when users could not see all evaluations(?), so it might make sense to make some of these evaluations not have results for the user. I'd suggest having a few in the "new" state.

evap/results/tests/test_views.py
```diff
@@ -264,6 +265,38 @@ def make_course_with_evaluations(unique_suffix):
...
+                    name_de=f"evaluation{i}",
+                    state=Evaluation.State.PUBLISHED,
+                    weight=5 - i,  # weights 1 to 5
+                    is_single_result=True,
+                )
+            )
+            for i in range(0, 5)
+        ]
+        for evaluation in evaluations:
+            baker.make(RatingAnswerCounter, contribution=evaluation.general_contribution, answer=2, count=2)
```
He3lixxx, line 290: The model bakery library has a feature to create multiple instances at once, so you don't have to write the loops yourself:
```suggestion
        names = [f"evaluation{i}" for i in range(5)]
        evaluations = baker.make(
            Evaluation,
            course=course,
            name_en=iter(names),
            name_de=iter(names),
            state=Evaluation.State.PUBLISHED,
            weight=iter(range(5, 0, -1)),
            is_single_result=True,
            _quantity=5,
        )
        contributions = [e.general_contribution for e in evaluations]
        baker.make(RatingAnswerCounter, contribution=iter(contributions), answer=2, count=2, _quantity=len(evaluations))

```

evap/results/tests/test_views.py
```diff
@@ -264,6 +265,38 @@ def make_course_with_evaluations(unique_suffix):
...
+        warm_up_template_cache(evaluations)
+        page = self.app.get(self.url, user=student).body.decode()
+        percentages = [6, 13, 20, 26, 33]
+        # percentage are floored, see evaluation/templatetags/evaluation_filters.py:100
+        prev_index = 0
+        for i in range(0, 5):
+            next_index = page.index(str(percentages[i]) + "%", prev_index)
```
He3lixxx, line 297: You pass prev_index here as a search start, so next_index would always have to be bigger than that. I'd just remove that argument. While at it, you could also use f-strings here.
```suggestion
            next_index = page.index(f"{percentages[i]}%")
```

evap/results/tests/test_views.py
```diff
@@ -264,6 +265,38 @@ def make_course_with_evaluations(unique_suffix):
...
+            self.assertTrue(next_index > prev_index)
```
He3lixxx, line 298: It would be nice if you could also check that it lines up with the evaluation names, so that
index("evaluation0") < index("6%") < index("evaluation1") < index("12%") ...

He3lixxx (changes requested):
Thanks for writing the test! I think we have to make some small changes to it, though.

He3lixxx (approved):
Thanks :)

PR #1700, opened 2022-01-18
Update `COMMANDS` in `manage_autocompletion.sh`

niklasmohrin:
We forgot to do so in #1688


He3lixxx (approved):
What an absolute failure ;)

PR #1706, opened 2022-02-14
Fix #1675: Relocate button "I can't give feedback"

### LINKED ISSUE #1675: Relocate button "I can't give feedback"
The buttons "I can't give feedback about this contributor" in the student voting form should be placed inside the respective cards' bodies. They should stay right-aligned at the top, alongside the heading text.

The buttons must be `tab-selectable` when the respective cards are open and the buttons are not disabled (the buttons are disabled when they are pressed and get enabled again when any voting option is selected).

Kakadus:
Fixes #1675


janno42 (commented):
button location and functionality are perfect. unfortunately, the keyboard navigation doesn't work anymore, tabbing over the repositioned button is currently not possible. @niklasmohrin can probably help you with that :)

evap/static/ts/src/student-vote.ts
```diff
@@ -130,7 +130,7 @@ studentForm.addEventListener("keydown", (e: KeyboardEvent) => {
             selectables[selectables.length - 1].focus({ preventScroll: true });
             return;
         }
-    } while (isInvisible(rows[nextRowIndex]));
+    } while (isInvisible(rows[nextRowIndex]) || rows[nextRowIndex].querySelector(".tab-selectable") == null);
```
niklasmohrin, line 133:
```suggestion
    } while (isInvisible(rows[nextRowIndex]) || rows[nextRowIndex].querySelector(".tab-selectable") === null);
```

He3lixxx (approved):
Code looks good, behavior seems fine.

@janno42 when using the keyboard to select "I can't give feedback [...]", the keyboard focus stays on this button while the card collapses. So, to advance, users have to press "tab" afterwards. Until they do, they will not see which item has keyboard focus. My first intuition was that it should automatically forward to the next element, but the more I think about it, the less I think I'd like it. What do you think?

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -1,7 +1,19 @@
 {% load evaluation_filters %}
-
+<div class="d-flex">
+    <h4 class="me-auto">{{ questionnaire_group.0.questionnaire.public_name }}</h4>
```
niklasmohrin, line 3: This now does the first questionnaire outside of the loop, I wonder if we could restructure it so that we have the "I can't give feedback" button in the loop, but only included in the first iteration, so instead of
```
<h4>...</h4>
{% if contributor %} ... {% endif %}

{% for form %}
  {% if not forloop.first %} <h4> ... </h4> {% endif %}
  ...
{% endfor %}
```

we would have

```
{% for form %}
  <h4>...</h4>
  {% if forloop.first and contributor %}
    ...
  {% endif ⅜}
{% endfor %}
```

For me, this better captures the intent here and it removes the duplicated "knowledge" about how one questionnaire is layed out in the html (though the last one is rather minor)

He3lixxx: Yeah, sounds good

janno42:
> Code looks good, behavior seems fine. Thanks!
> 
> @janno42 when using the keyboard to select "I can't give feedback [...]", the keyboard focus stays on this button while the card collapses. So, to advance, users have to press "tab" afterwards. Until they do, they will not see which item has keyboard focus. My first intuition was that it should automatically forward to the next element, but the more I think about it, the less I think I'd like it. What do you think?

For me, both solutions are fine. Because keeping it as it is now produces less overhead, I'd say it can stay as it is :)

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

niklasmohrin (changes requested):
(blocking for [my comment earlier](https://github.com/e-valuation/EvaP/pull/1706#discussion_r825477945))

Kakadus:
I am refactoring it now...

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -1,7 +1,18 @@
 {% load evaluation_filters %}
-
 {% for form in questionnaire_group %}
-    <h4{% if not forloop.first %} class="mt-3"{% endif %}>{{ form.questionnaire.public_name }}</h4>
```
niklasmohrin, line 4: Before, the first `h4` didnt have the margin at the top, now there is some new whitespace - please remove again :)

niklasmohrin (changes requested):

niklasmohrin (approved):

PR #1714, opened 2022-02-21
Fix importer error-merge error

### LINKED ISSUE #1711: Dictionary merge error in course importer
the following code wants to merge error dicts. however, this fails with a `TypeError` (`unsupported operand type(s) for |: 'str' and 'str'`) in `merge_dictionaries_of_sets`.
https://github.com/e-valuation/EvaP/blob/32caa9d799a938f88fa49940b16f4f44bd89d431/evap/staff/importers.py#L413-L415

reproduce by importing a file with two entries: two participants having the same course that differs in its degree (e.g., `BA` and `MA`) and having another error (e.g. grade `ungraded` instead of `no`).

niklasmohrin:
Closes #1711, mostly reverts #1602

### DIFF
--- a/evap/staff/fixtures/excel_files_test_data.py
+++ b/evap/staff/fixtures/excel_files_test_data.py
@@ -122,6 +122,14 @@ test_enrollment_data_degree_merge_filedata = {
     ]
 }
 
+test_enrollment_data_error_merge_filedata = {
+    'MA Belegungen': [
+        ['Degree', 'Student last name', 'Student first name', 'Student email address', 'Course kind', 'Course is graded', 'Course name (de)', 'Course name (en)', 'Responsible title', 'Responsible last name', 'Responsible first name', 'Responsible email address'],
+        ['Grandmaster', 'Quid', 'Bastius', 'bastius.quid@external.example.com', 'jaminar', 'probably not', 'Bauen', 'Build', '', 'Sed', 'Diam', '345@external.example.com'],
+        ['Beginner,Bachelor', 'Lorem', 'Ipsum', 'ipsum.lorem@institution.example.com', 'jaminar', 'probably not', 'Bauen', 'Build', '', 'Sed', 'Diam', '345@external.example.com'],
+    ],
+}
+
 test_enrollment_data_import_names_filedata = {
     'MA Belegungen': [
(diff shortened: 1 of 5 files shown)


janno42 (approved):
✔️ Meets requirements
✔️ Manual tests passed

evap/staff/importers.py
```diff
@@ -402,18 +402,25 @@ def process_evaluation(self, evaluation_data, sheetname, row):
...
                 )
-                self.evaluations[evaluation_id].degrees |= evaluation_data.degrees
-                self.evaluations[evaluation_id].errors = merge_dictionaries_of_sets(
-                    self.evaluations[evaluation_id].errors, evaluation_data.errors
-                )
-            elif evaluation_data != self.evaluations[evaluation_id]:
+
+                known_data.degrees |= evaluation_data.degrees
+
+                assert all(key in ["degrees", "course_type", "is_graded"] for key in evaluation_data.errors)
```
He3lixxx, line 416: you could use set semantics, if you wanted
```python3
assert evaluation_data.errors.keys() <= {"degrees", "course_type", "is_graded"}
```

evap/staff/fixtures/excel_files_test_data.py
```diff
@@ -122,6 +122,14 @@
     ]
 }
 
+test_enrollment_data_error_merge_filedata = {
+    'MA Belegungen': [
+        ['Degree',            'Student last name', 'Student first name', 'Student email address',               'Course kind', 'Course is graded', 'Course name (de)', 'Course name (en)', 'Responsible title', 'Responsible last name', 'Responsible first name', 'Responsible email address'],
```
niklasmohrin, line 127: @He3lixxx I comma-aligned these cells, should I align the other literals here too, or just format them like we had it before?

He3lixxx: I don't know, I think I'm fine with both.

Pro: Alignment is nice, makes it easier to read.
Con: If line wrapping is on (in PR reviews), it doesn't help. Aligning by hand is tedious. People might forget. It would probably be best if we had it auto formatted?

He3lixxx (approved):
code looks good to me 👍

evap/staff/fixtures/excel_files_test_data.py
niklasmohrin: Hm, auto-format seems like a good call, but every solution that is not a straight up black-plugin for exactly this will probably do more harm than good; I will revert for now and we can think about it another time

evap/staff/importers.py
niklasmohrin: nice

PR #1718, opened 2022-02-25
Delete answers for removed questionnaires

### LINKED ISSUE #1716: Delete answers for questionnaires removed from evaluation
Staff users can remove questionnaires from running or finished evaluations so that the contained questions will not be shown to voters participating in the evaluation from that point on. However, answers that have already been given for these questions before the questionnaire was removed, remain in the database. This leads to inaccessible text answers, blocking the review process.

It is necessary for staff users to be able to remove questionnaires in case they have been selected by mistake and are not applicable for the evaluation. So all answers for these questions need to be deleted when removing the questionnaires.
In addition, questionnaires for which answers have already been recorded should be highlighted in the evaluation form with an accompanying warning that answers for these will be deleted if the questionnaires are removed.
This applies to general questions as well as contributor questions.

janno42:
fix #1716
adds danger warnings in staff evaluation form for those questionnaires that already have answers for questions in these questionnaires and deletes these answers if the questionnaires are removed.

### DIFF
--- a/evap/evaluation/templatetags/evaluation_filters.py
+++ b/evap/evaluation/templatetags/evaluation_filters.py
@@ -219,7 +219,5 @@ def order_by(iterable, attribute):
 
 
 @register.filter
-def value_at(dictionary, key):
-    if key in dictionary:
-        return dictionary[key]
-    return None
+def get(dictionary, key):
+    return dictionary.get(key)
--- a/evap/evaluation/templatetags/evaluation_filters.py
+++ b/evap/evaluation/templatetags/evaluation_filters.py
@@ -216,3 +216,10 @@ def has_nonresponsible_editor(evaluation):
 @register.filter
 def order_by(iterable, attribute):
     return sorted(iterable, key=lambda item: getattr(item, attribute))
+
+
+@register.filter
+def value_at(dictionary, key):
+    if dictionary and key in dictionary:
+        return dictionary[key]
+    return None
--- a/evap/evaluation/templatetags/evaluation_filters.py
+++ b/evap/evaluation/templatetags/evaluation_filters.py
@@ -220,6 +220,6 @@ def order_by(iterable, attribute):
 
 @register.filter
 def value_at(dictionary, key):
-    if dictionary and key in dictionary:
+    if key in dictionary:
         return dictionary[key]
     return None
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -453,12 +453,7 @@ class EvaluationForm(forms.ModelForm):
         removed_questionnaires = set(self.instance.general_contribution.questionnaires.all()) - set(
             selected_questionnaires
         )
-        TextAnswer.objects.filter(
-            contribution=evaluation.general_contribution, question__questionnaire__in=removed_questionnaires
-        ).delete()
-        RatingAnswerCounter.objects.filter(
-            contribution=evaluation.general_contribution, question__questionnaire__in=removed_questionnaires
-        ).delete()
+        evaluation.general_contribution.remove_answers_to_questionnaires(removed_questionnaires)
         evaluation.general_contribution.questionnaires.set(selected_questionnaires)
         if hasattr(self.instance, "old_course"):
             if self.instance.old_course != evaluation.course:
@@ -633,13 +628,7 @@ class ContributionForm(forms.ModelForm):
         if self.instance.pk:
             selected_questionnaires = self.cleaned_data.get("questionnaires")
             removed_questionnaires = set(self.instance.questionnaires.all()) - set(selected_questionnaires)
-            TextAnswer.objects.filter(
-                contribution=self.instance, question__questionnaire__in=removed_questionnaires
-            ).delete()
-            RatingAnswerCounter.objects.filter(
-                contribution=self.instance, question__questionnaire__in=removed_questionnaires
-            ).delete()
-
+            self.instance.remove_answers_to_questionnaires(removed_questionnaires)
         return super().save(*args, **kwargs)
 
 
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -449,7 +449,17 @@ class EvaluationForm(forms.ModelForm):
 
     def save(self, *args, **kw):
         evaluation = super().save(*args, **kw)
-        evaluation.general_contribution.questionnaires.set(self.cleaned_data.get("general_questionnaires"))
+        selected_questionnaires = self.cleaned_data.get("general_questionnaires")
+        removed_questionnaires = set(self.instance.general_contribution.questionnaires.all()) - set(
+            selected_questionnaires
+        )
+        TextAnswer.objects.filter(
+            contribution=evaluation.general_contribution, question__questionnaire__in=removed_questionnaires
+        ).delete()
+        RatingAnswerCounter.objects.filter(
+            contribution=evaluation.general_contribution, question__questionnaire__in=removed_questionnaires
+        ).delete()
+        evaluation.general_contribution.questionnaires.set(selected_questionnaires)
         if hasattr(self.instance, "old_course"):
             if self.instance.old_course != evaluation.course:
                 update_template_cache_of_published_evaluations_in_course(self.instance.old_course)
@@ -618,6 +628,19 @@ class ContributionForm(forms.ModelForm):
             return True
 
         return False
+
+    def save(self, *args, **kwargs):
+        if self.instance.pk:
+            selected_questionnaires = self.cleaned_data.get("questionnaires")
+            removed_questionnaires = set(self.instance.questionnaires.all()) - set(selected_questionnaires)
+            TextAnswer.objects.filter(
+                contribution=self.instance, question__questionnaire__in=removed_questionnaires
+            ).delete()
+            RatingAnswerCounter.objects.filter(
+                contribution=self.instance, question__questionnaire__in=removed_questionnaires
+            ).delete()
+
+        return super().save(*args, **kwargs)
 
 
 class ContributionCopyForm(ContributionForm):
--- a/evap/staff/tests/test_forms.py
+++ b/evap/staff/tests/test_forms.py
@@ -590,12 +590,7 @@ class ContributionFormsetTests(TestCase):
 
     def test_answers_for_removed_questionnaires_deleted(self):
         # pylint: disable=too-many-locals
-        semester = baker.make(Semester)
-        evaluation = baker.make(
-            Evaluation,
-            course=baker.make(Course, semester=semester),
-        )
-        baker.make(Questionnaire, questions=[baker.make(Question)])
+        evaluation = baker.make(Evaluation)
         general_question_1 = baker.make(Question, type=Question.LIKERT)
         general_question_2 = baker.make(Question, type=Question.LIKERT)
         general_questionnaire_1 = baker.make(Questionnaire, questions=[general_question_1])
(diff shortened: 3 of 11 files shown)


evap/staff/tests/test_forms.py
```diff
@@ -588,6 +588,79 @@ def test_prevent_contribution_deletion_with_answers(self):
         self.assertFalse(formset.forms[0].show_delete_button)
         self.assertTrue(formset.forms[1].show_delete_button)
 
+    def test_answers_for_removed_questionnaires_deleted(self):
+        # pylint: disable=too-many-locals
+        semester = baker.make(Semester)
+        evaluation = baker.make(
+            Evaluation,
+            course=baker.make(Course, semester=semester),
+        )
```
He3lixxx, line 597:
```suggestion
        evaluation = baker.make(Evaluation)
```
or am I missing something?

evap/staff/tests/test_forms.py
```diff
@@ -588,6 +588,79 @@ def test_prevent_contribution_deletion_with_answers(self):
...
+        baker.make(Questionnaire, questions=[baker.make(Question)])
```
He3lixxx, line 598: This looks unused, too?

evap/staff/tests/test_views.py
```diff
@@ -1896,6 +1906,46 @@ def test_remove_participants_proportional_reward_points(self):
         )
         self.assertNotIn("The removal as participant has granted the user &quot;d@institution.example.com&quot;", page)
 
+    def test_questionnaire_with_answers_warning(self):
+        page = self.app.get(self.url, user=self.manager)
+        self.assertIn(
```
He3lixxx, line 1911: Probably a good case for [`self.assertInHtml`](https://docs.djangoproject.com/en/4.0/topics/testing/tools/#django.test.SimpleTestCase.assertInHTML)

evap/staff/forms.py
```diff
@@ -616,6 +626,19 @@ def show_delete_button(self):
...
+    def save(self, *args, **kwargs):
+        if self.instance.pk:
+            selected_questionnaires = self.cleaned_data.get("questionnaires")
+            removed_questionnaires = set(self.instance.questionnaires.all()) - set(selected_questionnaires)
+            TextAnswer.objects.filter(
+                contribution=self.instance, question__questionnaire__in=removed_questionnaires
+            ).delete()
+            RatingAnswerCounter.objects.filter(
+                contribution=self.instance, question__questionnaire__in=removed_questionnaires
+            ).delete()
```
He3lixxx, line 638: Should be maybe add some kind of database constraint to model the constraint that we violated here before?

> answers for (contribution, question) can only exist if question.questionnaire is included in contribution.questionnaires

I think it would be good to have it, but I don't know if it's feasable with the tools we have. There is quite some joining going on.

He3lixxx (changes requested):
Some serious testing going on there 👍.

niklasmohrin (commented):
(whoops, apparently I didn't hit submit yesterday :thinking: )

evap/staff/views.py
```diff
@@ -1207,6 +1207,15 @@ def notify_reward_points(grantings, **_kwargs):
...
         return redirect("staff:semester_view", semester.id)
 
+    questionnaires_with_answers_per_contributor = {}
+    for contribution in evaluation.contributions.all():
+        questionnaires = Questionnaire.objects.filter(
+            Q(questions__textanswer__contribution=contribution)
+            | Q(questions__ratinganswercounter__contribution=contribution)
+        ).distinct()
+        if questionnaires.count() > 0:
+            questionnaires_with_answers_per_contributor[contribution.contributor] = questionnaires
```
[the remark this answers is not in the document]
He3lixxx: Yes. In general, neither `qs.count()` nor `qs.exists()` fill the query set cache because they are implemented at the database level without actually fetching records. `len(qs)` and `bool(qs)` do fetch the records and fill the cache.

Here, as a micro-optimization, we could do
```suggestion
        if len(questionnaires) > 0:
            questionnaires_with_answers_per_contributor[contribution.contributor] = questionnaires
```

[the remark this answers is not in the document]
janno42: i'd prefer having logic in the views and not in the template filters but in this case it would be okay for me - your call.

evap/staff/tests/test_forms.py
janno42: wth did i do here?

evap/staff/forms.py
janno42: seems quite complicated for a not really that important thing. but if you'd feel better with it, let's discuss that :)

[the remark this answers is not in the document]
janno42: would you define a method `objects` that manually returns a QuerySet with both answer types?

evap/staff/views.py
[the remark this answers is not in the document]
He3lixxx: Hmm, yeah, but a more explicit logic in the view would require a custom data structure / object or an additional template argument. We have tests, and the worst thing that would happen is that the explanation is shown although it's not necessary, so resolving for now.

evap/evaluation/templatetags/evaluation_filters.py
```diff
@@ -216,3 +216,10 @@ def has_nonresponsible_editor(evaluation):
 @register.filter
 def order_by(iterable, attribute):
     return sorted(iterable, key=lambda item: getattr(item, attribute))
+
+
+@register.filter
+def value_at(dictionary, key):
+    if dictionary and key in dictionary:
+        return dictionary[key]
+    return None
```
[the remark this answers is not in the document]
He3lixxx: Hmm. One issue seems to be that staff_evaluation_form.html and contributor_evaluation_form.html are very similar and share their includes for evaluation_form_general_questionnaires.html and contribution_formset.html. However, in the contributor views (`evaluation_view` and `evaluation_edit`), no `questionnaires_with_answers_per_contributor` argument is passed, so the template engine puts an empty string there as argument for the template tag, resulting in a
```python3
TypeError: 'in <string>' requires string as left operand, not NoneType
```

The other issue is that staff_evaluation_form.html was used also for `evaluation_copy` and `evaluation_create`, also without the argument, so these tests also caused errors.

This works for me: https://github.com/He3lixxx/EvaP/commit/14fbd8a3a5cf96f1650562d46da0a20d2d6d0aed

(Btw: In staff_evaluation_form.html, currently, `questionnaires_with_answers_per_contributor` is not passed explicitly, as we usually do, but captured implicitly. I think we want to be explicit about it?)

I wonder what's the best way to handle this. Maybe make the argument explicit in all include-tags, add a constant in `{}` in contributor_evaluation_form.html and add the template parameter to evaluation_create and evaluation_copy? Semantically, it somewhat collides with the editable flag. I don't find it too trivial to reason about the correctness of just passing an empty dict.

(Unrelated: There was a lot of noise in the testing output, with failed tests during teardown that seemed to be caused by logging. I'll see why that happened and if we can do something about it.)

evap/staff/forms.py
janno42: see #1728

niklasmohrin:
I found out why the `assertInHTML` breaks: According to https://docs.djangoproject.com/en/4.0/topics/testing/tools/#django.test.SimpleTestCase.assertHTMLEqual this would be interpreted to match the html `<label ...></label>` without any content. Now that @He3lixxx found out that we get a BeautifulSoup object from the response itself, you could use a selector like: `self.assertIsNotNone(page.html.select_one('label.form-check-label.badge.bg-danger[for="id_contributions-1-questionnaires_0"]'))`, **but**, this is significantly slower than the `assertIn`. @He3lixxx and I agreed that `assertIn` is superior and we should use it as long as it doesn't break.

```python
def x1():
    self.assertIsNotNone(
        page.html.select_one(
            'label.form-check-label.badge.bg-danger[for="id_contributions-1-questionnaires_0"]'
        )
    )

def x2():
    self.assertIn(
        '<label class="form-check-label badge bg-danger" for="id_contributions-1-questionnaires_0">',
        page,
    )

In [3]: timeit(x1, number=100)
Out[3]: 1.841594752000674

In [4]: timeit(x2, number=100)
Out[4]: 0.003163628000038443
```

evap/staff/views.py
```diff
@@ -1207,6 +1207,15 @@ def notify_reward_points(grantings, **_kwargs):
 
         return redirect("staff:semester_view", semester.id)
 
+    questionnaires_with_answers_per_contributor = {}
+    for contribution in evaluation.contributions.all():
+        questionnaires = Questionnaire.objects.filter(
```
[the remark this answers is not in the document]
He3lixxx: Hmm, this looks easy at first glance, but after trying for quite a while, I feel like this is another one of those queries where django just doesn't quite provide the flexibility -- all my attempts to direct select questionnaires or contributors failed.

Essentially, we want (which isn't really that complicated)
```sql
SELECT UserProfile, Questionnaire
FROM (pairwise inner joins) UserProfile, Contribution, Questionnaire, Question, Answer
WHERE Answer.contribution == Contribution.id AND Contribution.evaluation = INPUT;
```
The best code I could come up with is starting at answer, since we can follow foreign-keys from here to questionnaire and to contribution. I noticed this code already passes the test:
```python3
    answers = [
        *TextAnswer.objects.select_related("question__questionnaire", "contribution__contributor"),
        *RatingAnswerCounter.objects.select_related("question__questionnaire", "contribution__contributor"),
    ]

    questionnaires_with_answers_per_contributor = defaultdict(list)
    for answer in answers:
        questionnaires_with_answers_per_contributor[answer.contribution.contributor].append(
            answer.question.questionnaire
        )
```

Even though it doesn't filter for the correct evaluation, because in the test, there are no answers for other evaluations. Adding this to the view test's setUpTestData solves that:
```python3
        baker.make(
            TextAnswer, question__questionnaire=cls.contributor_questionnaire, contribution__contributor=cls.editor
        )
```

fixing the filtering and adding some syntactic sugar leaves us at:
```python3
    contributor_questionnaire_pairs = [
        (answer.contribution.contributor, answer.question.questionnaire)
        for answer_cls in [TextAnswer, RatingAnswerCounter]
        for answer in answer_cls.objects.filter(contribution__evaluation=evaluation).select_related(
            "question__questionnaire", "contribution__contributor"
        )
    ]

    questionnaires_with_answers_per_contributor = defaultdict(list)
    for (contributor, questionnaire) in contributor_questionnaire_pairs:
        questionnaires_with_answers_per_contributor[contributor].append(questionnaire)
```

it's 10 lines instead of 8, and a bit harder to read. What do you think, is it worth it?

@niklasmohrin is there a one-liner to aggreate a list of (key, value) pairs into a dict[str, list[value-type]]?

evap/evaluation/templatetags/evaluation_filters.py
```diff
@@ -216,3 +216,10 @@ def has_nonresponsible_editor(evaluation):
 @register.filter
 def order_by(iterable, attribute):
     return sorted(iterable, key=lambda item: getattr(item, attribute))
+
+
+@register.filter
+def value_at(dictionary, key):
+    if key in dictionary:
+        return dictionary[key]
+    return None
```
niklasmohrin, line 225: We still wanted to use `get` for name and implementation (https://github.com/e-valuation/EvaP/pull/1718#discussion_r817201881) :)

janno42:
Linter: "evap/evaluation/models.py:1065:0: W0232: Class has no __init__ method (no-init)"
true, but why does it bother now?

He3lixxx:
> Linter: `evap/evaluation/models.py:1065:0: W0232: Class has no __init__ method (no-init)` true, but why does it bother now?

At first I thought it might be related to some smart decision making on pylint's side, because we added the first real public method to the class. But, after checking their source code: [Nope, it should be a simple "Does `__init__` exist" check for all classes](https://github.com/PyCQA/pylint/blob/main/pylint/checkers/classes/class_checker.py#L805).

I can't even reproduce locally, neither with `pylint evap -j 0` (CI command) nor with `pylint evap.evaluation.models`, neither on your branch nor on the hypothetical merge result that the CI ran on. However, the CI error seems to be consistent (restarted the linter 2 times, failed every time.)

evap/staff/views.py
```diff
@@ -1205,6 +1212,18 @@ def notify_reward_points(grantings, **_kwargs):
 
         return redirect("staff:semester_view", semester.id)
 
+    contributor_questionnaire_pairs = [
+        (answer.contribution.contributor, answer.question.questionnaire)
+        for answer_cls in [TextAnswer, RatingAnswerCounter]
```
He3lixxx, line 1217: Might be worth adding the `set(Answer.__subclasses__()) == {TextAnswer, RatingAnswerCounter}` assertion, too.

He3lixxx (approved):

niklasmohrin (approved):

Issue #1719, opened 2022-03-01
[waiting / blocked until 2023] Remove Debug-Mode Testrun Main-Push


He3lixxx:
In https://github.com/e-valuation/EvaP/pull/1257 we introduced that CI [also runs the tests in debug mode](https://github.com/e-valuation/EvaP/pull/1257/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485R15).

During debug execution, it seems that django does lots of additional work, e.g. it [keeps a log of all database queries around](https://docs.djangoproject.com/en/3.2/faq/models/#how-can-i-see-the-raw-sql-queries-django-is-running), which occupies loads of ram. For running the tests with debug=true in the VM, you need around 4,5GB of memory available. With the default VM configuration, it will simply fail. Test execution is around 3 times slower, on CI, a non-debug run takes 1:45 where a debug run takes 5:49.

[The django feature ticket for running the tests in debug mode](https://code.djangoproject.com/ticket/27008) says that this flag is intended to troubleshoot test failures.

So far, I know of one instance where reportedly non-debug tests passed and debug tests would have given an error, because we messed up some variables in a template: https://github.com/e-valuation/EvaP/issues/868. However, this was in 2016, many django versions ago. From the current django documentation, I'd say it shouldn't make a difference for template errors whether debug is enabled. The only situation where a test failure depends on DEBUG that I can think of is in our own code, if we depend on DEBUG.

Pro:
- Reduce energy consumed on every single CI run, and CI runs will finish faster (2 minutes rather than 6 minutes)

Con:
- We will not test code paths of ours that explicitly branch on `settings.DEBUG` anymore. We currently have this 3 times, and its always a re-raise for debugging. We will not lose any coverage on codecov, because the debug runs are not measured for coverage currently.
    - Theoretically, this might create situations where a test will succeed but a `DEBUG=True` instance will show an error. I'd say that this is unlikely enough that we should do it and benefit from the shorter CI run time. If it happens, we will notice.
    - We can still annotate tests with a settings override to have them run with `DEBUG=True` if we want to test code that relies on this flag. (We currently have 2 tests that force `DEBUG=False` to prevent the re-raise mentioned above)

[...]

PR #1720, opened 2022-03-07
Don't run debug mode tests for PRs

niklasmohrin:
See #1719

For now, we only run tests in debug mode when code is pushed to main. I also tried some other stuff, maybe it worked, maybe these changes are gone when you are reading this 🤷


He3lixxx:
Thanks for tackling this.

While you're at it: Currently, some of the CI actions are marked as "required" in a PR (linter, backup, coverage, debug mode, reverse order) while the others are not. I don't think this is intentional, I think we regard all of them as required. Maybe you could fix that.

janno42:
I added the other checks as being required

niklasmohrin:
I tried things :tm: to deduplicate some jobs, but failed to do so because of container/no-container shenanigans :shrug: it just becomes more duplicated now

there isn't really a way to test that the action is working without pushing to main, maybe I can try it out on the main branch of my fork :thinking:

janno42:
yes, please test it on your fork and not in the main repo :)

janno42 (approved):

Issue #1630, opened 2021-09-27
Missing test coverage in `staff/views.py`


niklasmohrin:
The coverage in `staff/views.py` is rather low, here are some spots that seem important-ish

- [ ] `semester_grade_reminder`
- [ ] `semester_archive_{participants, results}, semester_delete_grade_documents, {evaluation, course}_delete`: these just forward to the methods on the database objects / instances; I suppose these are tested then, right?
- [ ] `course_create` with `semester.participations_are_archived`
- [ ] `evaluation_edit` with `evaluation.is_single_result`and `helper_single_result_edit` itself
- [ ] we never `evaluation.reopen_evaluation()` or `evaluation.manager_approve()` within `evaluation_edit`
- [ ] `get_evaluation_and_contributor_textanswer_sections` with `filter_textanswers` or with general contributions or the cases where no answers exist (not sure if these are even possible)
- [ ] `questionnaire_index` with `filter_questionnaires`
- [ ] `make_questionnaire_edit_forms` with `editable`
- [ ] `questionnaire_edit` where the form and formset are valid
- [ ] `user_index` with `filter_users`
- [ ] `export_contributor_results_view`

PR #1721, opened 2022-03-14
Fix internal server errors on requests with unexpected parameters

He3lixxx:
We had some views where users could trigger server errors if they wanted. Fix them so they result in bad requests, as they should.

For the manager-only views, it makes sense to keep the logging, as we deal with trusted users, so we would assume that the request was triggered by our code and we might want to investigate.

The naming of `unexpected_bad_request` is worth discussion. I could also imagine something like `bad_request_by_trusted_user` or whatever.


evap/staff/views.py
```diff
@@ -1500,7 +1500,11 @@ def evaluation_textanswers(request, semester_id, evaluation_id):
 
 @reviewer_required
 def evaluation_textanswers_skip(request):
-    evaluation_id = request.POST["evaluation_id"]
+    try:
+        evaluation_id = request.POST["evaluation_id"]
+    except KeyError:
+        return unexpected_bad_request()
+
     evaluation = get_object_or_404(Evaluation, id=evaluation_id)
```
niklasmohrin, line 1508: Can't we just do `request.POST.get("evaluation_id")` and have the 404 returned by `get_object_or_404`? Edit: Ah, do we want to distinguish between 400 and 404? If so, forget what I said :D

evap/staff/views.py
```diff
@@ -1512,9 +1516,12 @@ def evaluation_textanswers_skip(request):
 @require_POST
 @reviewer_required
 def evaluation_textanswers_update_publish(request):
-    textanswer_id = request.POST["id"]
-    action = request.POST["action"]
-    evaluation_id = request.POST["evaluation_id"]
+    try:
+        textanswer_id = request.POST["id"]
+        action = request.POST["action"]
+        evaluation_id = request.POST["evaluation_id"]
```
niklasmohrin, line 1522: Same here, depending on whether we want the logging or not, by replacing all of these with `dict.get` and using `get_object_or_404` below, we will produce 400s if any of these are empty. Don't know if we want to distinguish between 400 and 404 here either

evap/staff/views.py
He3lixxx: Yes, I explicitly wanted to keep the logging of any request that doesn't have a `evaluation_id` in its data, because I'd consider it likely to indicate a bug at another place.

At the same time, the same would be true for the 404. Haven't thought about that.

He3lixxx:
I added tests in a new commit, so you may want to review by commit. It's a rather big pile of code for the tests, partly because most of the views were completely untested before. On the bright side, these tests cover 8 views from #1630.

evap/staff/tests/test_views.py
```diff
@@ -215,6 +215,26 @@ def test_reward_points_granting_message(self):
         )
 
 
+class TestUserDeleteView(WebTestStaffMode):
```
He3lixxx, line 218: TestUserDeleteView, TestCourseDeleteView, TestEvaluationDeleteView share the exact same structure. It might make sense to and deduplicate their logic (into a base class?)

That could even be extended to TestSemesterArchiveParticipationsView or GradeDeleteTest, although they are slightly more different.

I'll play around and see if I can find something promising.

janno42 (approved):

niklasmohrin (commented):
Haven't looked at tests yet, rest looks good

evap/staff/tests/test_views.py
He3lixxx: @niklasmohrin (and anyone else who cares): I put a mockup here: https://github.com/He3lixxx/EvaP/commit/4413db8b09fbe0e2c0029c08b22f2d54e0bc4f94. I think it works quite well for TestUserDeleteView, TestSemesterDeleteGradeDocumentsView, TestCourseDeleteView, and TestQuestionnaireDeleteView, and it works somewhat okay for TestSemesterDeleteView (where we have one additional test I wouldn't want to leave out). What do you think?

It would be nice if we could do the same thing with these tests, but that would be more complicated:
* EvaluationDelete - has significant other assertions that I'd like to keep
* ArchiveParticipations -- doesn't delete a db object, but asserts relevant model method is called
* GradeDelete - no "bad request" case, always succeeds

The test runner will find all classes that inherit from unittest.TestCase, so we can't make the Mixin derive from WebTest -- that's why I chose the current mixin design.

evap/evaluation/tools.py
```diff
@@ -1,14 +1,25 @@
...
 import xlwt
 from django.conf import settings
+from django.core.exceptions import SuspiciousOperation, ValidationError
+from django.db.models import Model
 from django.http import HttpResponse
+from django.shortcuts import get_object_or_404
 from django.utils.translation import get_language
 
 
+def get_object_from_dict_pk_entry_or_logged_40x(model_cls: Type[Model], dict_obj: Dict, key: str):
```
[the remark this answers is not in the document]
He3lixxx:
```
(venv) evap@ubuntu-bionic:/opt/evap$ ./manage.py lint
************* Module evap.evaluation.tools
evap/evaluation/tools.py:15:0: C0103: Class name "M" doesn't conform to PascalCase naming style (invalid-name)
```

So `ModelType` or is there some other canonical name?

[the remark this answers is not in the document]
niklasmohrin: Fixed in https://github.com/PyCQA/pylint/pull/5894 :)

evap/staff/tests/test_views.py
```diff
@@ -687,56 +709,41 @@ class TestSemesterDeleteView(WebTestStaffMode):
...
-        self.assertFalse(semester.can_be_deleted_by_manager)
+    @patch.object(Semester, "can_be_deleted_by_manager", False)
+    def test_invalid_deletion(self):
+        self.app.post(self.url, user=self.manager, params=self.post_params, status=400)
+        self.assertTrue(Semester.objects.filter(pk=self.semester.pk).exists())
 
-        self.app.post(self.url, params={"semester_id": semester.pk}, user=self.manager, status=400)
-        self.assertTrue(Semester.objects.filter(pk=semester.pk).exists())
+    @patch.object(Semester, "can_be_deleted_by_manager", True)
+    def test_valid_deletion(self):
```
niklasmohrin, line 722: Here, the `valid` case comes after the `invalid` one, but in `TestUserDeleteView`, it is the other way around

evap/staff/tests/test_views.py
```diff
@@ -687,56 +709,41 @@ class TestSemesterDeleteView(WebTestStaffMode):
...
+        self.app.post(self.url, user=self.manager, params=self.post_params)
+        self.assertFalse(Semester.objects.filter(pk=self.semester.pk).exists())
```
niklasmohrin, line 724: Foreshadowing my opinion on your base-class suggestion: Once we have that, we could also make `self.assertDeleted` and `self.assertNotDeleted` helpers to de-noise assertions like this one

evap/staff/tests/test_views.py
```diff
@@ -1462,6 +1513,15 @@ def test_operation_prepare_does_not_send_email_to_contributors(self):
...
         self.assertEqual(actual_emails[0]["additional_cc_users"], set())
 
+    def test_invalid_target_states(self):
+        evaluation = baker.make(Evaluation, state=Evaluation.State.APPROVED, course=self.course)
+        base_url = self.url + f"?evaluation={evaluation.pk}"
+
+        self.app.get(base_url + "&target_state=133742", user=self.manager, status=400)
+        self.app.get(base_url + "&target_state=asdf", user=self.manager, status=400)
+        self.app.get(base_url + "&target_state=", user=self.manager, status=400)
+        self.app.get(base_url + f"&target_state={Evaluation.State.IN_EVALUATION}", user=self.manager, status=200)
```
niklasmohrin, line 1523:
```suggestion
        base_url = f"{self.url}?evaluation={evaluation.pk}"

        self.app.get(f"{base_url}&target_state=133742", user=self.manager, status=400)
        self.app.get(f"{base_url}&target_state=asdf", user=self.manager, status=400)
        self.app.get(f"{base_url}&target_state=", user=self.manager, status=400)
        self.app.get(f"{base_url}&target_state={Evaluation.State.IN_EVALUATION}", user=self.manager, status=200)
```

evap/staff/tests/test_views.py
niklasmohrin: I like :+1:

He3lixxx: Added the commit.

evap/staff/tests/test_views.py
He3lixxx: Should be resolved with the DeleteViewTestMixin.

niklasmohrin (approved):

PR #1726, opened 2022-03-21
Provision evap via vbox and docker

fidoriel:
`vagrant` up --provider=docker` can now be used to provision docker on e.g. macOS Monterey (Intel and M1) where vbox can currently not run.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -17,10 +17,18 @@ Vagrant.configure("2") do |config|
     end
   end
 
-  config.vm.provider :docker do |d|
+  config.vm.provider :docker do |d, override|
     d.image = "ubuntu:bionic"
     # Docker container really are supposed to be used differently. Hacky way to make it into a "VM".
     d.cmd = ["tail", "-f", "/dev/null"]
+
+    # Workaround for no SSH server as long as https://github.com/hashicorp/vagrant/issues/8145 is still open
+    override.trigger.before :provision do |trigger|
+      trigger.ruby do |env, machine| system("vagrant docker-exec -it -- /evap/deployment/provision_vagrant_vm.sh") end
+    end
+    override.trigger.before :ssh do |trigger|
+      trigger.ruby do |env, machine| system("vagrant docker-exec -it -- sudo -H -u evap bash") end
+    end
   end
 
   # port forwarding
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -4,25 +4,29 @@
 Vagrant.require_version ">= 1.8.1"
 
 Vagrant.configure("2") do |config|
-  config.vm.box = "ubuntu/bionic64"
-  config.vm.box_version = "= 20210928.0.0 "
+  config.vm.provider :virtualbox do |v, override|
+    override.vm.box = "ubuntu/bionic64"
+    override.vm.box_version = "= 20210928.0.0 "
+    override.vm.provision "shell", path: "deployment/provision_vagrant_vm.sh"
 
-  # port forwarding
-  config.vm.network :forwarded_port, guest: 8000, host: 8000 # django server
-  config.vm.network :forwarded_port, guest: 80, host: 8001 # apache
-  config.vm.network :forwarded_port, guest: 6379, host: 6379 # redis. helpful when developing on windows, for which redis is not available
-
-  config.vm.provider :virtualbox do |v, _override|
     # disable logfile
     if Vagrant::Util::Platform.windows?
       v.customize [ "modifyvm", :id, "--uartmode1", "file", "nul" ]
     else
       v.customize [ "modifyvm", :id, "--uartmode1", "file", "/dev/null" ]
     end
-
-    # show virtualbox gui, uncomment this to debug startup problems
-    #v.gui = true
   end
+
+  config.vm.provider :docker do |d|
+    d.image = "ubuntu:bionic"
+    # Docker container really are supposed to be used differently. Hacky way to make it into a "VM".
+    d.cmd = ["tail", "-f", "/dev/null"]
+  end
+
+  # port forwarding
+  config.vm.network :forwarded_port, guest: 8000, host: 8000 # django server
+  config.vm.network :forwarded_port, guest: 80, host: 8001 # apache
+  config.vm.network :forwarded_port, guest: 6379, host: 6379 # redis. helpful when developing on windows, for which redis is not available
 
   # override username to be evap instead of vagrant, just as it is on production.
   # This is necessary so management script can assume evap is the correct user to
@@ -34,6 +38,4 @@ Vagrant.configure("2") do |config|
   if ARGV[0] == "ssh" or ARGV[0] == "ssh-config"
     config.ssh.username = 'evap'
   end
-
-  config.vm.provision "shell", path: "deployment/provision_vagrant_vm.sh"
 end
--- a/deployment/provision_vagrant_vm.sh
+++ b/deployment/provision_vagrant_vm.sh
@@ -71,12 +71,15 @@ a2dissite 000-default.conf
 # see https://github.com/e-valuation/EvaP/issues/626
 # and https://docs.djangoproject.com/en/dev/howto/deployment/wsgi/modwsgi/#if-you-get-a-unicodeencodeerror
 sed -i s,\#.\ /etc/default/locale,.\ /etc/default/locale,g /etc/apache2/envvars
-systemctl reload apache2
+service apache2 reload
 
 cp /etc/skel/.bashrc /home/$USER/
 # auto cd into /$USER on login and activate venv
 echo "cd $REPO_FOLDER" >> /home/$USER/.bashrc
 echo "source $ENV_FOLDER/bin/activate" >> /home/$USER/.bashrc
+
+# required for docker (no-op if already started)
+echo "sudo service postgresql start && sudo service redis-server start" >> /home/$USER/.bashrc
 
 # install requirements
 sudo -H -u $USER $ENV_FOLDER/bin/pip install -r $REPO_FOLDER/requirements-dev.txt
@@ -96,7 +99,7 @@ wget https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh --no-verbos
 
 # setup evap
 cd "$MOUNTPOINT"
-git submodule update --init
+sudo -H -u $USER git submodule update --init
 sudo -H -u $USER bash -c "source /home/$USER/.nvm/nvm.sh; nvm install --no-progress node; npm ci"
 echo "nvm use node" >> /home/$USER/.bashrc
 sudo -H -u $USER $ENV_FOLDER/bin/python manage.py migrate --noinput
--- a/deployment/provision_vagrant_vm.sh
+++ b/deployment/provision_vagrant_vm.sh
@@ -2,6 +2,7 @@
 
 set -x # print executed commands
 
+MOUNTPOINT="/evap"
 USER="evap"
 REPO_FOLDER="/opt/evap"
 ENV_FOLDER="/home/$USER/venv"
@@ -9,8 +10,14 @@ ENV_FOLDER="/home/$USER/venv"
 # force apt to not ask, just do defaults.
 export DEBIAN_FRONTEND=noninteractive
 
-# install python stuff
 apt-get -q update
+
+# system utilities that docker containers don't have
+apt-get -q install -y sudo wget git bash-completion
+# docker weirdly needs this -- see https://stackoverflow.com/questions/46247032/how-to-solve-invoke-rc-d-policy-rc-d-denied-execution-of-start-when-building
+printf '#!/bin/sh\nexit 0' > /usr/sbin/policy-rc.d
+
+# install python stuff
 apt-get -q install -y python3.7 python3.7-dev python3-venv python3.7-venv gettext
 
 # setup postgres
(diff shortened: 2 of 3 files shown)


deployment/provision_vagrant_vm.sh
```diff
@@ -1,8 +1,19 @@
...
+docker=false
+
 set -x # print executed commands
 
-USER="evap"
+if [ "$1" = "docker" ]
+then
+    docker=true
+    echo "provisioning docker"
+    USER="vagrant"
```
niklasmohrin, line 11: ? In docker, the user is called vagrant? How did that happen?

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+# django
+RUN apt-get install -y python3 python3-pip
+
+RUN useradd --create-home -s /bin/bash vagrant
+RUN echo -n 'vagrant:vagrant' | chpasswd
+RUN echo 'vagrant ALL = NOPASSWD: ALL' > /etc/sudoers.d/vagrant
+RUN chmod 440 /etc/sudoers.d/vagrant
+RUN mkdir -p /home/vagrant/.ssh
+RUN chmod 700 /home/vagrant/.ssh
+RUN echo "ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEA6NF8iallvQVp22WDkTkyrtvp9eWW6A8YVr+kz4TjGYe7gHzIw+niNltGEFHzD8+v1I2YJ6oXevct1YeS0o9HZyN1Q9qgCgzUFtdOKLv6IedplqoPkcmF0aYet2PkEDo3MlTBckFXPITAMzF8dJSIFo9D8HfdOV0IAdx4O7PtixWKn5y2hMNG0zQPyUecp4pzC6kivAIhyfHilFR61RGL+GPXQ2MWZWFYbAGjyiYJnAmCP3NOTd0jMZEnDkbUvxhMmBYSdETk1rRgm+R4LOzFUGaHqHDLKLX+FIPKcF96hrucXzcWyLbIbEgE98OHlnVYCzRdK8jlqm8tehUc9c9WhQ==" > /home/vagrant/.ssh/authorized_keys
```
niklasmohrin, line 17: Where is this coming from? A comment is needed, but it would be best if this would be a command that explained itself :)

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+# django
+RUN apt-get install -y python3 python3-pip
+
+RUN useradd --create-home -s /bin/bash vagrant
+RUN echo -n 'vagrant:vagrant' | chpasswd
+RUN echo 'vagrant ALL = NOPASSWD: ALL' > /etc/sudoers.d/vagrant
+RUN chmod 440 /etc/sudoers.d/vagrant
+RUN mkdir -p /home/vagrant/.ssh
+RUN chmod 700 /home/vagrant/.ssh
```
niklasmohrin, line 15:
```suggestion
RUN mkdir --mode 700 -p /home/vagrant/.ssh
```

Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM ubuntu:bionic
+
+RUN apt-get update -y
+
+RUN apt-get install -y --no-install-recommends ssh sudo iproute2
+
+# django
+RUN apt-get install -y python3 python3-pip
```
niklasmohrin, line 8:
```suggestion
RUN apt-get install -y --no-install-recommends ssh sudo iproute2 python3 python3-pip
```

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+RUN chmod 700 /home/vagrant/.ssh
+RUN echo "ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEA6NF8iallvQVp22WDkTkyrtvp9eWW6A8YVr+kz4TjGYe7gHzIw+niNltGEFHzD8+v1I2YJ6oXevct1YeS0o9HZyN1Q9qgCgzUFtdOKLv6IedplqoPkcmF0aYet2PkEDo3MlTBckFXPITAMzF8dJSIFo9D8HfdOV0IAdx4O7PtixWKn5y2hMNG0zQPyUecp4pzC6kivAIhyfHilFR61RGL+GPXQ2MWZWFYbAGjyiYJnAmCP3NOTd0jMZEnDkbUvxhMmBYSdETk1rRgm+R4LOzFUGaHqHDLKLX+FIPKcF96hrucXzcWyLbIbEgE98OHlnVYCzRdK8jlqm8tehUc9c9WhQ==" > /home/vagrant/.ssh/authorized_keys
+RUN chmod 600 /home/vagrant/.ssh/authorized_keys
+RUN chown -R vagrant:vagrant /home/vagrant/.ssh
+RUN sed -i -e 's/Defaults.*requiretty/#&/' /etc/sudoers
+RUN sed -i -e 's/\(UsePAM \)yes/\1 no/' /etc/ssh/sshd_config
+
+RUN mkdir /var/run/sshd
+
+RUN apt-get -y install openssh-client
```
niklasmohrin, line 25: Please move this into the installation command above :)

deployment/provision_vagrant_vm.sh
```diff
@@ -11,9 +22,13 @@ export DEBIAN_FRONTEND=noninteractive
 
 # install python stuff
 apt-get -q update
-apt-get -q install -y python3.7 python3.7-dev python3-venv python3.7-venv gettext
+apt-get -q install -y sudo python3.7 python3.7-dev python3-venv python3.7-venv gettext git wget npm
```
niklasmohrin, line 25: Why did you add `npm`? We install node and npm using `nvm` at the end of this file

Vagrantfile
```diff
@@ -30,10 +39,20 @@ Vagrant.configure("2") do |config|
...
   # Also, provision needs to use vagrant as ssh user (since evap does not exist yet)
-  config.vm.synced_folder ".", "/evap", mount_options: ["uid=1042", "gid=1042"]
-  if ARGV[0] == "ssh" or ARGV[0] == "ssh-config"
-    config.ssh.username = 'evap'
+
+  config.vm.provider :docker do |d, override|
+    override.vm.synced_folder ".", "/evap", docker_consistency: "cached"
+
+    override.vm.provision "shell", path: "deployment/provision_vagrant_vm.sh", args: "docker"
+    override.vm.provision "shell", inline: "service redis-server start && service postgresql start", run: "always", privileged: true
```
niklasmohrin, line 47: This runs when _provisioning_, but shouldn't it run every time the container is booted? I suppose there should be an equivalent to `systemctl enable`, probably `service enable` or so. You can also run this in the script, this is what we expect in both setups anyway

deployment/provision_vagrant_vm.sh
```diff
@@ -11,9 +22,13 @@ export DEBIAN_FRONTEND=noninteractive
...
 
 # setup postgres
+if [ "$docker" = true ]
```
niklasmohrin, line 28: If [this article](https://www.cyberciti.biz/faq/how-to-declare-boolean-variables-in-bash-and-use-them-in-a-shell-script/) didn't lie to me, this should be
```suggestion
if "$docker"
```

deployment/provision_vagrant_vm.sh
```diff
@@ -11,9 +22,13 @@ export DEBIAN_FRONTEND=noninteractive
...
+then
+    printf '#!/bin/sh\nexit 0' > /usr/sbin/policy-rc.d
```
niklasmohrin, line 30: Why is this needed? Also: there is no trailing newline here

deployment/provision_vagrant_vm.sh
```diff
@@ -58,7 +84,13 @@ a2dissite 000-default.conf
...
 # and https://docs.djangoproject.com/en/dev/howto/deployment/wsgi/modwsgi/#if-you-get-a-unicodeencodeerror
 sed -i s,\#.\ /etc/default/locale,.\ /etc/default/locale,g /etc/apache2/envvars
-systemctl reload apache2
+
+if [ "$docker" = true ]
+then
+    service apache2 reload
+else
+    systemctl reload apache2
+fi
```
niklasmohrin, line 93: The `service` command should work in both cases

deployment/provision_vagrant_vm.sh
```diff
@@ -73,7 +105,10 @@ cp $REPO_FOLDER/deployment/localsettings.template.py $REPO_FOLDER/evap/localsett
 sed -i -e "s/\${SECRET_KEY}/$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 32)/" $REPO_FOLDER/evap/localsettings.py
 
 # setup vm auto-completion
-cp $REPO_FOLDER/deployment/manage_autocompletion.sh /etc/bash_completion.d/
+if [ "$docker" = true ]
+then
+    cp $REPO_FOLDER/deployment/manage_autocompletion.sh /etc/bash_completion.d/
+fi
```
niklasmohrin, line 111: We want to also have the completions please :D

deployment/provision_vagrant_vm.sh
```diff
@@ -82,7 +117,13 @@ apt-get -q install -y libasound2 libgconf-2-4 libgbm1 libgtk-3-0 libnss3 libx11-
 wget https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh --no-verbose --output-document - | sudo -H -u $USER bash
 
 # setup evap
-cd /$USER
+if [ "$docker" = true ]
+then
+    cd /evap
+else
+    cd /$USER
+fi
```
niklasmohrin, line 125: Isn't this `/evap` in both cases?

niklasmohrin (changes requested):

He3lixxx:
A few thoughts:
- I like our current provisioning script. It's simple and easy to understand. It only makes a minimal set of assumptions about the system it's running on. Basically, it could be used to set up a development environment in any linux environment. I'd like to keep it unaware of where it's running.
- The canonical way to start an interactive shell in a docker container is `docker exec [...] -it bash`, not SSH into it. Setting up a SSH server seems overkill to me, even if this means that `vagrant ssh` will not work. Note that if we set up SSH servers for users, we are somewhat responsible for their security.
- We have vagrant as a tool to provide us with some kind of virtualized linux environment, and the shell script to set up this environment. This should be complete, so I'd like to not add a third file (Dockerfile).

I've tried to come up with the minimal modifications needed for this, resulting in a total of 7 additional relevant lines, not counting comments and readme entries: https://github.com/He3lixxx/EvaP/commit/7c08e9328981e2717cb38ff74bf2021a925384a8

No SSH server means users will not be able to `vagrant ssh` and vagrant will not be able to automatically provision the container on the first `vagrant up`. The workflow for users would thus be (as written in the readme):
```bash
vagrant up --provider docker  # to start the container
vagrant docker-exec -it -- /evap/deployment/provision_vagrant_vm.sh   # to set up the environment (once)
vagrant docker-exec -it -- sudo -H -u evap bash  # to run an interactive shell
```

now I'd say this would be perfect if we somehow could tell vagrant how to do provisioning with this setup, and if we could provide some kind of `vagrant shell` command that either executes `vagrant ssh` or starts a shell in the docker container. Since the Vagrantfile is written in ruby, it would make perfect sense to just script this there in a few lines, but it seems that [custom provisioners](https://www.vagrantup.com/docs/plugins/provisioners) and [custom commands](https://www.vagrantup.com/docs/plugins/commands) can only be done with plugins, which I'd consider overkill again.

Overall, I'd vote in favor of going this approach: Get some "VM"-like container with as little configuration as possible, run the provisioning script in there, profit. What do you think?

deployment/provision_vagrant_vm.sh
fidoriel: it should be except if the username is changed

Vagrantfile
fidoriel: it does run always, i think. this is only needed if docker is used. I had some issues with it, this is the best and simplest I was able to find.

deployment/provision_vagrant_vm.sh
fidoriel: for other usernames provisioning fails

He3lixxx:
Updated code with your latest changes at [He3lixxx:docker](https://github.com/He3lixxx/EvaP/tree/docker) / c762ccfa267570e841abdebc4ffee7b3027e5a99

I did some more refactoring. You should be able to test locally with:
```bash
vagrant destroy
git remote add He3lixxx git@github.com:He3lixxx/EvaP.git
git fetch He3lixxx
git switch -d He3lixxx/docker
vagrant up --provider docker
```

He3lixxx (approved):

He3lixxx:
(note: chome headless won't start inside docker container, because it can not sandbox itself. Reproduce with `./manage.py ts test`)

He3lixxx:
Another thing is that the `postgresql` and `redis-server` services will not auto-start on `vagrant up`, so if you do `vagrant halt && vagrant up && vagrant ssh`, you will have to manually start them again (`service postgresql start && service redis-server start`).

I couldn't find a "clean" way to auto-start them. It's a bit complicated, since starting them will only work after provisioning was done. We could add the commands that start them to bash.rc or similar.

niklasmohrin (approved):

niklasmohrin:
About chromium, docker and macos: @FranzSw noticed that on his m1 mac, the npm installation fails because of puppeteer something and then chromium something. This, interestingly, results in `npx tsc` running a different (outdated?) package, because the right one didn't get installed, very awkward. Adding a `sudo apt-get install -y chromium-browser` fixed this and we only have the normal sandbox errors again.

We could mark puppeteer as an optional dependency, so that the current script doesn't fail - that would mean that one has to run `npm ci` before `ts test` and `precommit` work. We could also install chromium in the provision script. @He3lixxx What do you think?

He3lixxx:
I think that an actual error message and/or full provisioning log would help :P If there are bugs with puppeteer on M1, I'm pretty sure they would have an open issue, possible with a workaround or solution -- I'd say we should do whatever that is.

PR #1727, opened 2022-03-28
Fix LoggedModel.thread.request not being cleared in failing tests

He3lixxx:
The deletion was not working for failing tests. This caused the setup logic for following tests to assume that the "old" `request.user` is responsible for the current change (usually a `baker.make()` call), while the user object didn't even exist in the database anymore. In consequence, it caused foreign key integrity errors for tests that are totally unrelated to the actual failure.

This is interesting because it contradicts (my interpretation) of the [django documentation](https://docs.djangoproject.com/en/4.0/topics/http/middleware/#writing-your-own-middleware), which reads as if just running teardown code after calling `self.get_response(request)` should be fine, because there should be some wrapper that catches exceptions and turns them into 400 or 500 responses. Maybe this does not apply when running tests with our setup so we can get meaningful stack traces?

Anyway, it doesn't look to me as if this would be critical for any other middleware. The language middleware might not correctly set the language cookie, but that shouldn't be an issue.

### DIFF
--- a/evap/evaluation/middleware.py
+++ b/evap/evaluation/middleware.py
@@ -19,9 +19,14 @@ class LoggingRequestMiddleware:
         LoggedModel.thread.request = request
         LoggedModel.thread.request_id = str(uuid.uuid4())
 
-        response = self.get_response(request)
-
-        del LoggedModel.thread.request
-        del LoggedModel.thread.request_id
+        try:
+            # django documentation says to just do this without the try, and that exceptions will be handled:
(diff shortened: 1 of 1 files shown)


niklasmohrin:
In the exception handling for responses, I found that there is a setting `DEBUG_PROPAGATE_EXCEPTIONS` which is read in `handle_uncaught_exception` in `core.handlers.exception` and re-raises. Webtest activates this setting `WebtestMixin._patch_settings`

niklasmohrin (approved):

PR #1731, opened 2022-03-28
Update all the python dependencies (except django)

He3lixxx:
Leaving the joy of updating django for someone else.

Probably best reviewed by commit.

### DIFF
--- a/evap/results/tools.py
+++ b/evap/results/tools.py
@@ -142,7 +142,7 @@ def get_single_result_rating_result(evaluation):
 
 
 def get_results_cache_key(evaluation):
-    return "evap.staff.results.tools.get_results-{:d}".format(evaluation.id)
+    return f"evap.staff.results.tools.get_results-{evaluation.id:d}"
 
 
 def cache_results(evaluation):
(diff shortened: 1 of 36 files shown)


evap/results/tools.py
```diff
@@ -142,7 +142,7 @@ def get_single_result_rating_result(evaluation):
 
 
 def get_results_cache_key(evaluation):
-    return "evap.staff.results.tools.get_results-{:d}".format(evaluation.id)
```
niklasmohrin, line 145: Any idea why we have the explicit base-10 here?

requirements.txt
```diff
@@ -1,14 +1,11 @@
...
-model-bakery == 1.4.0
-django-extensions == 3.1.5
-django-sendfile2 == 0.6.0
-mozilla-django-oidc == 1.2.4
+django-extensions==3.1.5
+django-fsm==2.8.0
+django-redis==5.2.0
+django-sendfile2==0.6.1
+django>=3.2,<3.3
+mozilla-django-oidc==2.0.0
```
niklasmohrin, line 6: From what I can tell, the only breaking changes really only are that they dropped support for python 2 and old 3 versions, right?

niklasmohrin (approved):

evap/results/tools.py
He3lixxx: We brainstormed and got two points
* uuids would break, which is bad
* this breaks on evaluation.id == None, which is good, cause otherweise, we could have key collisions

PR #1668, opened 2021-11-15
 Prioritize evaluations with finished grading for review

PonderKoKo:
Closes #1663

PR #1732, opened 2022-03-30
Prioritize evaluations with finished grading for review

PonderKoKo:
This seems to work - the two things I am still unsure about are the difference between evaluation.id and evaluation.pk (which one should I use) and whether there is a more sensible way to assert the correct evaluation order than checking for a substring in the first span.

### DIFF
--- a/evap/staff/tests/test_views.py
+++ b/evap/staff/tests/test_views.py
@@ -2187,16 +2187,16 @@ class TestEvaluationTextAnswerView(WebTest):
     @classmethod
     def setUpTestData(cls):
         cls.manager = make_manager()
-        semester = baker.make(Semester, pk=1)
-        student1 = baker.make(UserProfile, email="student@institution.example.com")
+        cls.semester = baker.make(Semester, pk=1)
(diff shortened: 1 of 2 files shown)


He3lixxx:
This is a continuation of #1668 (Note that you committed to your `main` branch here -- the PR before was based on your `reviewpriority` branch. That's also probably the reason why github guided you towards opening a new PR).

For our bookkeeping, it would be nice if you could close one of the PRs. Afaik GitHub doesn't allow changing source branches for an existing PR, so if you want to work off your `reviewpriority` branch, you could close this PR here and update #1668. If you want to work off your main branch, you could just close #1668. We usually encourage people to use feature branches. (Feel free to PM us if you have any questions on this)

Regarding your points: In new code, usually use `.pk` because it encapsulates what the actual field is named where we don't care.
About the assertion: The substring approach is perfectly fine. The HTML parsing might even be a bit slow. The actual render location of an element can always differ from what is in the html, so apart from opening the page in a webbrowser and taking a screenshot, there's not really a way to assert the location of certain elements on the page. So, we usually assume that our rendered page represents the flow in the html document, and with that assertion, the substring approach does the job.

PR #1734, opened 2022-04-04
Add SECURITY.md

He3lixxx:
This is a minimum version right now.

We could add more (for example any of the following) if we wanted, but would it help?
* ("Disclosure") We will track security-relevant changes in the changelog (would wait with that until we have a changelog set up).
    * Might be worth adding: Vulnerability Information will be made public after patches have been applied to production -- but it's kind of pointless without specifying, where exactly it will be made public?
* ("Supported version") - We do not officially give any warranty or support (license already says that -- but most open source projects have similar licenses, yet try to offer supported versions of some kind)

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,7 +2,7 @@
 
 Please do not open GitHub issues for anything you think might have a security implication.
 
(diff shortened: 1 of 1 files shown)


janno42 (approved):
fine with me 👍

niklasmohrin (approved):

PR #1737, opened 2022-04-11
Fix broken html escaping in various places

He3lixxx:
We might want to discuss how much testing we want to have going on for these.

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -19,11 +19,12 @@ from django.db.models import Count, Manager, OuterRef, Q, Subquery
 from django.db.models.functions import Coalesce
 from django.dispatch import Signal, receiver
 from django.template import Context, Template
+from django.template.defaultfilters import linebreaksbr
 from django.template.exceptions import TemplateSyntaxError
 from django.template.loader import render_to_string
 from django.urls import reverse
 from django.utils.functional import cached_property
-from django.utils.html import escape
+from django.utils.safestring import SafeData
 from django.utils.translation import gettext_lazy as _
 from django_fsm import FSMIntegerField, transition
 from django_fsm.signals import post_transition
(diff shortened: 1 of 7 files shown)


evap/evaluation/models.py
```diff
@@ -1894,12 +1894,9 @@ def construct_mail(self, to_email, cc_addresses, subject_params, body_params):
         subject = self.render_string(self.subject, subject_params, autoescape=False)
         plain_content = self.render_string(self.plain_content, body_params, autoescape=False)
 
-        rendered_content = self.render_string(
-            self.html_content if self.html_content else self.plain_content.replace("\n", "<br />"), body_params
-        )
-        wrapper_template_params = dict(
-            {"email_content": rendered_content, "email_subject": escape(subject)}, **body_params
-        )
+        html_content = self.html_content if self.html_content else linebreaksbr(self.plain_content)
```
He3lixxx, line 1897: @niklasmohrin: [`linebreaksbr`](https://github.com/django/django/blob/main/django/template/defaultfilters.py#L469) happens to do exactly what we need here (escape, then replace \n with <br/>). However, this is not guaranteed by the [documentation for this filter specifically](https://docs.djangoproject.com/en/4.0/ref/templates/builtins/#linebreaksbr). What do you think? We probably want a test for escaping here, anyway, and if we have that, we could keep `linebreaksbr`?

evap/staff/importers.py
```diff
@@ -505,13 +505,10 @@ def set_existing_course_to_merge_or_add_errors(self, semester, evaluation_data):
...
                 self.errors[ImporterError.COURSE].append(
                     format_html(
-                        _(
-                            "Course {} ({}) does already exist in this semester, but the courses can not be merged for the following reasons:{}."
-                        ).format(
-                            evaluation_data.name_en,
-                            evaluation_data.name_de,
-                            "".join(f"<br /> - {msg}" for msg in merge_hindrances),
-                        )
+                        "Course {} ({}) does already exist in this semester, but the courses can not be merged for the following reasons:{}.",
```
niklasmohrin, line 508: The translation here was lost in translation

Additionally, it would be nice if the German name would be first in the German translation, named format arguments should work for that

evap/evaluation/views.py
```diff
@@ -101,8 +101,8 @@ def index(request):
 
     # check for redirect variable
     redirect_to = request.GET.get("next", None)
-    if redirect_to is not None:
-        return redirect(redirect_to)
+    if redirect_to is not None and url_has_allowed_host_and_scheme(redirect_to, None):
+        return redirect(iri_to_uri(redirect_to))
```
niklasmohrin, line 105: There is some documentation out there that wants us to do this, let's link it

evap/evaluation/models.py
niklasmohrin: If we have a test as our safety net, I don't have anything against `linebreaksbr` :+1:

niklasmohrin (approved):

Issue #1280, opened 2018-11-01
Look at result caching again


karyon:
An alternative approach to results caching could look like this:

- We still have one row in one template, `results_index_course.html`, but that one does not do the caching anymore.
- instead, the warmup method renders the template and writes the result into the cache manually.
- the results index view would, for each course, fetch the rendered template from the cache and put it into a list of rendered templates. `result_index.html` would then do nothing more than printing out all these rendered templates in a loop.

disadvantages:
- the template does not use the usual including and caching mechanisms, we do that manually instead

advantages:
- caching is handled completely on the python side and not split between python and the template. we have to deal with cache invalidation on the python side anyways, so moving cache-warmup there as well makes sense.
- we can use `get_many` of django's caching API, which should improve page performance.

we should wait for #1278 and decide on #1229 (at least fix the two tests in there) before doing this.

He3lixxx:
What do we want to do when the view detects that there is no cache entry? Calculate results and populate cache? We didn't want that so far because we were afraid that we might do something wrong so the page is rerendered at every call, right? If we do this, we could just encorporate it into the current system, afaik. If we don't want to do that, what should happen in this case?

I think I'm happy with our current approach, the only thing that is breaking so far is that the `cache` template tag will store the (bad) content when you render the template with a missing cache entry. This will impact all following renderings. Assuming that we will run into race conditions anyway, I think just populating the cache when we encounter a missing entry is the most reasonable approach.

Improved performance sounds good. I don't see putting this code on python side as a disadvantage, but I also don't think that it would much of an advantage. In the end, we _are_ caching templates or part of it, right? So imo it's always some mixup between views and templates, no matter where we do stuff.

In the end, I don't have a strict opinion on this, so putting it all in python would be fine for me, too. I think the question from paragraph one is the main question, no matter where we do all the caching stuff.

karyon:
what about this: with the approach i detailed above, we would use get_many to get all the cache entries. we can simply compare the length of the result with the number of courses, and boom we know whether there's something missing! at a minimum, we can retry a few times, that would be a fix for the race conditions. we could also trigger a recalculation like you suggested. that would be more robust in case there's an entry missing permanently, but that would be a bunch of code and should never happen. but, since that would be a plain addition to my approach, we could implement my approach, see how good that looks, and then see whether we want even more robustness.

so in addition to the advantages above, doing the caching in python
* will never poison the cache like the template fragment caching does
[...]

PR #1738, opened 2022-04-11
Make refresh_results_cache 6x faster

He3lixxx:
Initially, `./manage.py refresh_results_cache` took approximately 42 seconds for our test data on my machine.

* Prefetching on the `refresh_results_cache` management command brought it down to around 14 seconds (we had around 20'000 db queries) in the process
* Caching the template fragments (the same few fragments were loaded from disk once per render call, totalling in ~9000 file system accesses) brought it down to around 6 seconds (only noticable with DEBUG=True. With debug enabled, this was already cached).

Related to a discussion in #1280:
> you asked me to remind you to keep refresh_results_cache in mind. it currently needs a lot of time on production - so during the update the server is down for several minutes.

6x doesn't sound that impressive, but in comparison, this just flies away:

Edit: Current state is even more faster (x8, 4.7s for `refresh_results_cache`)

### DIFF
--- a/evap/evaluation/management/commands/refresh_results_cache.py
+++ b/evap/evaluation/management/commands/refresh_results_cache.py
@@ -31,7 +31,7 @@ class Command(BaseCommand):
 
         for counter, evaluation in enumerate(evaluations):
             progress_bar.update(counter + 1)
-            cache_results(evaluation)
+            cache_results(evaluation, refetch_related_objects=False)
 
         self.stdout.write("Prerendering result index page...\n")
 
--- a/evap/evaluation/management/commands/refresh_results_cache.py
+++ b/evap/evaluation/management/commands/refresh_results_cache.py
@@ -3,7 +3,12 @@ from django.core.management.base import BaseCommand
 from django.core.serializers.base import ProgressBar
 
 from evap.evaluation.models import Evaluation
-from evap.results.tools import STATES_WITH_RESULT_TEMPLATE_CACHING, STATES_WITH_RESULTS_CACHING, cache_results
+from evap.results.tools import (
+    GET_RESULTS_PREFETCH_LOOKUPS,
+    STATES_WITH_RESULT_TEMPLATE_CACHING,
+    STATES_WITH_RESULTS_CACHING,
+    cache_results,
+)
 from evap.results.views import warm_up_template_cache
 
 
@@ -19,8 +24,11 @@ class Command(BaseCommand):
         self.stdout.write("Calculating results for all evaluations...")
 
         self.stdout.ending = None
-        evaluations = Evaluation.objects.filter(state__in=STATES_WITH_RESULTS_CACHING)
+        evaluations = Evaluation.objects.filter(state__in=STATES_WITH_RESULTS_CACHING).prefetch_related(
+            *GET_RESULTS_PREFETCH_LOOKUPS,
+        )
         progress_bar = ProgressBar(self.stdout, evaluations.count())
+
         for counter, evaluation in enumerate(evaluations):
             progress_bar.update(counter + 1)
             cache_results(evaluation)
--- a/evap/results/tools.py
+++ b/evap/results/tools.py
@@ -1,10 +1,10 @@
 from collections import OrderedDict, defaultdict, namedtuple
 from math import ceil, modf
-from typing import Tuple, cast
+from typing import Dict, List, Tuple, cast
 
 from django.conf import settings
 from django.core.cache import caches
-from django.db.models import Exists, OuterRef, Q, Sum
+from django.db.models import Exists, OuterRef, Sum, prefetch_related_objects
 
 from evap.evaluation.models import (
     CHOICES,
@@ -16,7 +16,6 @@ from evap.evaluation.models import (
     Questionnaire,
     RatingAnswerCounter,
     TextAnswer,
-    UserProfile,
 )
 
 STATES_WITH_RESULTS_CACHING = {Evaluation.State.EVALUATED, Evaluation.State.REVIEWED, Evaluation.State.PUBLISHED}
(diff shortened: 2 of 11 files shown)


evap/results/tools.py
```diff
@@ -393,23 +402,24 @@ def get_grade_color(grade):
...
     non_proxy_contributors = [contributor for contributor in contributors if not contributor.is_proxy_user]
-    num_delegates = len(
-        set(UserProfile.objects.filter(represented_users__in=non_proxy_contributors).distinct()) - set(contributors)
+
+    num_delegates = sum(
+        1
+        for contributor in non_proxy_contributors
+        for delegate in contributor.delegates.all()
+        if delegate not in contributors
     )
```
He3lixxx, line 421: This is actually wrong, two contributors sharing a delegate would make that count go up by 2. I should add a test for that.

He3lixxx:
I don't want this to end up in more code. At the same time, there were a few things that I think we want to handle a bit differently:

* I'd argue it makes sense to put those prefetches into `_get_result_impl`, just before using the values. However, when doing that, `test_finishing_review_updates_results` starts failing -- because with the prefetching, django keeps all the old answers in the "prefetching-cache" (stored in the model's `_prefetched_objects_cache`), even when changing stuff. I think this can be an issue in real code, because we usually call `cache_results` and `get_results` after changing related objects of an evaluation, so the evaluation might contain outdated cached instance. For `create_cache`, our (current) uses seem to be safe in that regard. I tried to go through callers of `get_results` to see if this could actually produce wrong results somewhere, but I wasn't able to come to a conclusion there. I think it would make sense to clear out all django-cached related objects before computing the results.

* Another problem I encountered earlier: With aggressive prefetching, if the objects all contain their cached related objects, these will also be pickled. This means that a TextResult containing a TextAnswer instance will also contain its TextAnswer->Question->Questionnaire->Evaluation->Course->Responsible->Delegates.
This is unnecessary at best and wrong at worst, if these instances are used later after retrieving them from a cache, and they are outdated. We don't have the logic in place to invalidate caches along all hierarchies there. So, I'd argue we don't want to store any related objects in redis.
    * I think this was also a problem before, just not this big, because there were less instances known(?)

* Both combined made me implement `discard_cached_related_objects`, and use it in `_get_results_impl` as well as all the `TextResult`, `RatingResult`, ... classes, on all database instances.
    * Since I'm touching the results anyways, I figured it would make sense to add type annotations. This required reordering them
    * For `HeadingResult` and `TextAnswerVisibility`, adding the conversion-layer means they can no longer be namedtuples. This also requires to adapt one test to use the named accessor (which I'd argue is better anyway)
    * I added `copy()` calls before stripping the related objects, because I think for the average caller, it would be unexpected if the argument was modified.

* For `refresh_results_cache`, we want to prefetch once and use these values (otherwise, it get's ~40% slower). I think the optional kwarg is an okay-ish solution here.
    
* [ ] Tests for `is_m2m_prefetched` (technical debt) and `discard_cached_related_objects` will still be added -- they are necessary, these functions use stuff that's undocumented and not guaranteed to stay available

* [ ] I'll also add a test for https://github.com/e-valuation/EvaP/pull/1738#discussion_r848687453, which hopefully is just a small modification of the existing `TestTextAnswerVisibilityInfo`

Is this too much / are any of these points too cautious?

He3lixxx:
To make code simpler again: With this, `_get_results_impl` takes
* 30ms for getting the results of a single evaluation, including database queries (before: 180ms)
* 850ms to computing results of all evaluations in test_data.json, with prefetched data
* 4,2s for computing all results of all evaluations, without prefetched data (before: 25s)

so we could think about disabling parts of the results caching, if it makes our lives easier.

For example, for the evaluation_detail results page (based on measurements on a single, random evaluation), it doesn't really matter anymore whether the results are cached or computed (280ms vs 310ms time until the browser gets the response)

evap/evaluation/management/commands/refresh_results_cache.py
```diff
@@ -19,11 +19,18 @@ def handle(self, *args, **options):
         self.stdout.write("Calculating results for all evaluations...")
 
         self.stdout.ending = None
-        evaluations = Evaluation.objects.filter(state__in=STATES_WITH_RESULTS_CACHING)
+        evaluations = Evaluation.objects.filter(state__in=STATES_WITH_RESULTS_CACHING).prefetch_related(
```
niklasmohrin, line 22: This could be deduplicated from the call in `_get_results_impl`

evap/results/tools.py
```diff
@@ -85,49 +59,95 @@ def choices(self):
...
     @property
-    def is_published(self):
+    def is_published(self) -> bool:
         return self.counts is not None
 
 
 class TextResult:
-    def __init__(self, question, answers, answers_visible_to=None):
+    def __init__(
+        self, question: Question, answers: Iterable[TextAnswer], answers_visible_to: TextAnswerVisibility = None
```
niklasmohrin, line 97: The `= None` suggests this could be `Optional`

niklasmohrin (approved):

niklasmohrin:
Plus: we talked about introducing some kind of `group_by` utility function for the `defaultdict(list)` dance

He3lixxx:
Regarding `unordered_groupby`: I ended up not using it in
* results/views.py:395 (`questionnaire_max_answers`): The maximum-computation would be a bit much inside a generator expression
* results/tools.py:373 (`grouped_results`): This uses `extend` instead of `append`, I think making it use append would make it worse.
* results/exporters.py:330 (`questionnaires`): I tried, but the code got longer due to the high amount of nested stuff in the generator expression required
* staff/importers.py:564 (`participants_per_evaluation`): Would get longer. Part of #1634 anyways.

He3lixxx:
@janno42 I'd love to have the numbers of
```
time ./manage.py refresh_results_cache
```
from before and after this.

PR #1746, opened 2022-04-28
Django 4.0

### LINKED ISSUE #1745: Upgrade to Django 4.0
and with it, ubuntu18.04 to (at least) ubuntu20.04, and python (to 3.8 or 3.9 or whatever seems reasonable)

He3lixxx:
Actually, quite a simple upgrade. Fixes #1745.

There are a few new(ish) features that I noticed in the changelog that we might want to utilize:
- [ ] Django has [native support for redis](https://docs.djangoproject.com/en/4.0/topics/cache/#redis) now, so we can probably drop our dependency on django-redis.
- [ ] Django provides a [`LoginView`](https://docs.djangoproject.com/en/4.0/topics/auth/default/#django.contrib.auth.views.LoginView) that already has quite some redirect logic. Maybe, if we use this, we could simplify our custom redirect logic?
- [ ] Similar simplifications might be possible when using django's [`FormView`](https://docs.djangoproject.com/en/4.0/ref/class-based-views/generic-editing/#formview), [`DeleteView`](https://docs.djangoproject.com/en/4.0/ref/class-based-views/generic-editing/#deleteview), [`CreateView`](https://docs.djangoproject.com/en/4.0/ref/class-based-views/generic-editing/#createview) or [`UpdateView`](https://docs.djangoproject.com/en/4.0/ref/class-based-views/generic-editing/#updateview).
- [ ] I couldn't fully judge whether the changes in the category [`Forms`](https://docs.djangoproject.com/en/4.0/releases/4.0/#forms) are relevant to us -- especially, whether the new `nonform` class for non-form errors could be handy -- I think we have some custom logic going on for these somewhere?


.github/workflows/tests.yml
```diff
@@ -27,10 +27,10 @@ jobs:
     strategy:
       matrix:
         include:
-          - name: Coverage
+          - name: Test (coverage)
             command: coverage run manage.py test && codecov -X gcov
-          - name: Reverse order
-            command: python manage.py test --reverse
+          - name: Test (shuffled)
+            command: python manage.py test --shuffle
```
He3lixxx, line 33: Shuffle has the advantage that it will converge on finding all inter-test dependencies over time. Its randomness has the disadvantage that tests with inter-test dependencies might not be found out in the PR introducing them, but later at some unrelated point in time.

With reverse, we will not find all dependencies (see the ones fixed in this PR), but it will be deterministic: If the PR run passes, all post-merge runs will pass, too.

Opinions?

He3lixxx:
(I would move the investigation whether django's generic views can reduce our custom code into a new refactoring issue. I feel unable to judge these without actually attempting to do it, and this PR already has enough changes I think)

evap/middleware.py
```diff
@@ -19,7 +21,10 @@ def process_view(request, view_func, _view_args, _view_kwargs):
         if "no_login_required" in view_func.__dict__ and view_func.no_login_required:
             return None
 
-        if view_func.__name__ in ["OIDCAuthenticationRequestView", "OIDCAuthenticationCallbackView"]:
+        if view_func.__name__ in _ALLOWED_WITHOUT_LOGIN:
+            return None
```
He3lixxx, line 25: Thinking about this again: We have a test for this. The behavior of django and mozilla-django-oidc should be deterministic with a fixed version (either view_func.__name__ or view_func.view_class.__name__ should be set). Since the test always runs against the version in requirements.txt, and the code is only ever supposed to run with these versions, I think it's safe to remove the old `if` here.

And if it's safe, we should do it (we don't want to untested dead code lying around)

niklasmohrin: This was an exception for oidc from the beginning, so we can only do the necessary check (as long as we also run the tests on production before deploying :D). I would give the constant a less general name (or inline again).

He3lixxx: I don't think we've ever run the tests on production. So I guess it's time to play the debug game and find out what happened here (tried to circumvent that):

In 3.2.13, django still populated the generated view-method with a `__name__` [here](https://github.com/django/django/blob/3.2.13/django/views/generic/base.py#L74).
Somewhere in between (probably on 4 update), django started intentionally leaving `__name__` unpopulated, arguing that `view_class` should be used. The `view_class` attribute [is documented](https://docs.djangoproject.com/en/3.2/ref/class-based-views/base/#django.views.generic.base.View.as_view), and should have already worked with earlier versions.

So, seems that `view_func.view_class.__name__` should work according to documentation as long as mozilla-django-oidc doesn't add breaking changes (changing from view class to view func, or changing their names).

I'll drop the "old" check, as I think this likely not going to break (ha, I jinxed it)

niklasmohrin (approved):

niklasmohrin:
Ah, should we also pin a major version of node in the script? Node 17 [is about to end](https://nodejs.org/en/about/releases/), we could pin node 18 which is end of life in 2025.

He3lixxx:
> Ah, should we also pin a major version of node in the script? Node 17 [is about to end](https://nodejs.org/en/about/releases/), we could pin node 18 which is end of life in 2025.

In the PR introducing node, I voted against pinning the version (mainly because we weren't able to have the version specified in one single place, but also because node updates happen way too often imo and we don't care about it at all: This is a dev dependency, it shouldn't even be installed on the server. If anything breaks with a new node version, we would notice it, and so far it hasn't happened) -- https://github.com/e-valuation/EvaP/pull/1545#discussion_r551442355

niklasmohrin:
Fair enough

.github/workflows/tests.yml
```diff
@@ -42,14 +65,13 @@ jobs:
...
-
+        run: python manage.py test --shuffle
 
   mypy:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
 
     container:
-      image: python:3.7
+      image: python:3.8
```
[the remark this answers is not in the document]
He3lixxx: Indeed, apache gives the "It works!" apache default page, because reloading the config fails. I wonder why in the year 2022 it's still necessary to point apache so some shared library it should load, but the [current version in the mod_wsgi documentation](https://modwsgi.readthedocs.io/en/develop/user-guides/quick-installation-guide.html) seems to be even worse, asking us to compile the module from source (and then loading it similarly to what we do in wsgi.template.conf).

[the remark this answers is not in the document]
He3lixxx: turns out there is a [`libapache2-mod-wsgi-py3`](https://packages.ubuntu.com/focal/libapache2-mod-wsgi-py3) package that can do all this for us -- but its not linked in any documentation (mod-wsgi, django, interwebz tutorials, ...) ugh.

niklasmohrin (approved):
We need to adjust the [apache section in the wiki](https://github.com/e-valuation/EvaP/wiki/Installation#apache-2-configuration) about the wsgi things afterwards

He3lixxx:
Since there was a test failure due to hardcoded PKs [here](https://github.com/e-valuation/EvaP/runs/6266022730?check_suite_focus=true), I'm currently working on removing all the hardcoded PKs from the tests.

niklasmohrin (approved):

PR #1751, opened 2022-05-02
show display name for hidden questionnaires in result export

### LINKED ISSUE #1487: Questionnaire names in result exports
When hiding a questionnaire and creating a new version, the current date will be added to the title of the old questionnaire. This information will be shown in exports which might confuse users. We should think about only showing the display name instead or handling it in a completely different way. We have to think about exports where multiple old questionnaires of the same name are included though.

dieknolle3333:
closes #1487.
(For more info, read the issue :sweat_smile:)

### DIFF
--- a/evap/results/tests/test_exporters.py
+++ b/evap/results/tests/test_exporters.py
@@ -82,16 +82,16 @@ class TestExporters(TestCase):
         binary_content.seek(0)
         workbook = xlrd.open_workbook(file_contents=binary_content.read())
 
-        self.assertEqual(workbook.sheets()[0].row_values(4)[0], questionnaire_1.name)
+        self.assertEqual(workbook.sheets()[0].row_values(4)[0], questionnaire_1.public_name)
         self.assertEqual(workbook.sheets()[0].row_values(5)[0], question_1.text)
 
-        self.assertEqual(workbook.sheets()[0].row_values(7)[0], questionnaire_2.name)
+        self.assertEqual(workbook.sheets()[0].row_values(7)[0], questionnaire_2.public_name)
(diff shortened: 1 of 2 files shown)


janno42 (changes requested):
it turns out that we don't need the `name` in exports and can always use the `public_name` here.

evap/results/tests/test_exporters.py
```diff
@@ -367,11 +367,40 @@ def test_exclude_used_but_unanswered_questionnaires(self):
         cache_results(evaluation)
 
         sheet = self.get_export_sheet(evaluation.course.semester, degree, [evaluation.course.type.id])
-        self.assertEqual(sheet.row_values(4)[0], used_questionnaire.name)
+        self.assertEqual(sheet.row_values(4)[0], used_questionnaire.public_name)
         self.assertEqual(sheet.row_values(5)[0], used_question.text)
         self.assertNotIn(unused_questionnaire.name, sheet.col_values(0))
         self.assertNotIn(unused_question.text, sheet.col_values(0))
 
+    def test_show_display_name_of_questionnaires(self):
```
He3lixxx, line 375: Since we are now always showing the public name, could this test be removed?

janno42 (approved):

evap/results/tests/test_exporters.py
dieknolle3333: Right – in case the wrong questionnaire title is displayed, it would be pretty clear that this is causing the all the other tests to fail, because all the assertion statements asserting the name would be failing. 
Thus I'll remove this test if that's okay

niklasmohrin (approved):

He3lixxx (approved):

PR #1753, opened 2022-05-02
Fix evaluation counts in importer messages.

jb3rndt:
[Reference](https://github.com/e-valuation/EvaP/pull/1683#issuecomment-1115302197)
Suggestions for a better name than `new_evaluations` are welcome of course

### DIFF
--- a/evap/staff/tests/test_importers.py
+++ b/evap/staff/tests/test_importers.py
@@ -432,14 +432,12 @@ class TestEnrollmentImporter(TestCase):
     def test_existing_course_not_counted_in_created_courses_count(self):
         self.create_existing_course()
 
-        expected_course_count = Course.objects.count() + 22
-
         success_messages, __, __ = EnrollmentImporter.process(
             self.default_excel_content, self.semester, self.vote_start_datetime, self.vote_end_date, test_run=False
         )
 
         self.assertIn(
-            f"The import run will create {expected_course_count} courses/evaluations",
+            "Successfully created 22 courses/evaluations",
             "".join(success_messages),
         )
 
--- a/evap/staff/tests/test_importers.py
+++ b/evap/staff/tests/test_importers.py
@@ -429,6 +429,20 @@ class TestEnrollmentImporter(TestCase):
             existing_course_evaluation.participants.all(),
         )
 
+    def test_existing_course_not_counted_in_created_courses_count(self):
+        self.create_existing_course()
+
+        expected_course_count = Course.objects.count() + 22
+
+        success_messages, __, __ = EnrollmentImporter.process(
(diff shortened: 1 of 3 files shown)


He3lixxx (commented):
Code looks good. Could you add a test?

jb3rndt:
I'm not sure if we should add a test case for `test_run=True` as well, but wouldnt add much value I guess...
But what is with pylint? 
grafik
Should I just add an ignore there? I didnt touch that file so I dont know where its coming from

evap/staff/tests/test_importers.py
```diff
@@ -429,6 +429,18 @@ def test_existing_course_is_not_recreated(self):
...
+        self.create_existing_course()
+
+        success_messages, __, __ = EnrollmentImporter.process(
+            self.default_excel_content, self.semester, self.vote_start_datetime, self.vote_end_date, test_run=False
+        )
+
+        self.assertIn(
+            "Successfully created 22 courses/evaluations",
+            "".join(success_messages),
+        )
```
He3lixxx, line 442: Could this be part of the test above (...is_not_recreated)? Basically, you would only add the assertion then, right? This would then also allow to easily test with `test_run=True` and `test_run=False` (I think we have tests for both cases?)

He3lixxx:
Pylint is simply broken, #1759 should fix that. Just ignore that message for now.

evap/staff/tests/test_importers.py
jb3rndt: Yep I made it part of the test above.
Currently, there is no test regarding existing courses that has `test_run=True`. Should I add one that only tests the success message?

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

evap/staff/tests/test_importers.py
He3lixxx: Yes, there should be a test for the importer test run, and it should be asserted that:
* The merge warning is shown (would be fair if we just assert the the warnings between test run and non-test run are identical, if that works)
* The test success message also has the correct counts. Currently, it doesn't, needs to be fixed here: https://github.com/e-valuation/EvaP/blob/01c1bac1498da374fbea11cfebf2d156d643bfc5/evap/staff/importers.py#L598-L602

He3lixxx (changes requested):
blocking for https://github.com/e-valuation/EvaP/pull/1753#discussion_r873234379, so managing PRs gets a bit easier for us. We have quite a few PRs open at the moment.

evap/staff/tests/test_importers.py
jb3rndt: Ok, I made a test solely for checking the messages because it was too much for the ...is_not_created one

He3lixxx (approved):
thanks :)

niklasmohrin (approved):

PR #1756, opened 2022-05-09
Pylint: Only allow pascal case with Formsets.

### LINKED ISSUE #1748: Pylint: Properly enforce snake_case for variables?
I noticed in a PR that the variable name `snake_case_suddenPascalCase` doesn't get linted, when I think it should.
This is because we currently use custom regexes with pylint that allow this mixture:
https://github.com/e-valuation/EvaP/blob/13e63cb5278ac89c8417c9d79245a187cc81c1f9/pyproject.toml#L28-L37

It seems like we allowed `PascalStyle` for forms and formsets, where we sometimes have local variables that refer to something that behaves like a class (`InlineContributionFormset`)

All of these would default to the `snake_case` style if we removed them. Then, pylint errors for the following names:
1. variable name `InlineContributionFormset` (9 times)
2. variable name `InlineQuestionFormset` (5 times)
3. variable name `DegreeFormset` (1 time)
4. variable name `CourseTypeFormset` (1 time)
5. variable name `TextAnswerWarningFormSet` (1 time) -- note: `FormSet` instead of `Formset`
6. variable name `SectionFormset` (1 time)
7. variable name `InlineQuestionFormset` (1 time)
8. variable name `ContributionFormset` (1 time)
9. argument name `EvaluationFormClass` (1 time)
10. argument name `FormClass` in evaluation/tests/tools.py:119
11. variables `a` and `b` in evaluation_filters.py:78
12. variables `oldFile` and `newFile` in grades/models.py:62

I think by pep-8, local variables and arguments whose values are classes are usually named `something_cls`. We also have this in 13 places in our codebase. All `@classmethod`s take their class as first argument, and this arguments is usually named `cls`.

I propose to:
- Fix 9, 10 by renaming them: `evaluation_form_cls` and `form_cls`
- Fix 11, 12 by using longer names and `old_file`, `new_file`
- Regarding the Formsets (1-8): I'd be fine with either
    - Allowing `PascalCase` if the variable ends with `Formset` (that is: Make the regex: `^(snake_case_regex)|(PascalCaseRegex)Formset$`)
    - Rename them all to snake_case, appending `_cls` to distinguish class and instance (preferred by me)

Note: If we allow them for variables, we'd probably also allow them for arguments.

Removing (and thus defaulting) `function-rgx`, `attr-rgx` and `module-rgx` seems to cause no errors, so I'd do that.

Opinions?

He3lixxx:
Fixes #1748 

I also increased the default VM memory size to 2GB, since we ran out of memory with pylint in some situations.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,8 +27,8 @@ django-settings-module = "evap.settings"
 [tool.pylint.basic]
 # snake_case, or PascalCaseFormset, because django does it that way
 # see https://docs.djangoproject.com/en/4.0/topics/forms/formsets/
-variable-rgx = "(^[a-z0-9_]*)$|(^[A-Za-z]*Formset$)"
-argument-rgx = "(^[a-z0-9_]*)$|(^[A-Za-z]*Formset$)"
+variable-rgx = "(^[a-z0-9_]*$)|(^[A-Za-z]*Formset$)"
+argument-rgx = "(^[a-z0-9_]*$)|(^[A-Za-z]*Formset$)"
 
 good-names = [ "i", "j", "k", "ex", "Run", "_", "__", "e", "logger", "setUpTestData", "setUp", "tearDown"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,16 +25,10 @@ load-plugins = ["pylint_django"]
 django-settings-module = "evap.settings"
 
 [tool.pylint.basic]
-variable-rgx = "[A-Za-z_][A-Za-z0-9_]{2,60}$"
-argument-rgx = "[A-Za-z_][A-Za-z0-9_]{0,40}$"
-function-rgx = "[a-z_][a-z0-9_]{2,60}$"
-attr-rgx = "[a-z_][a-z0-9_]{1,60}$"
-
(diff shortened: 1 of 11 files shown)


pyproject.toml
```diff
@@ -25,16 +25,10 @@ load-plugins = ["pylint_django"]
...
-
-# Allow longer names in tests as test_* methods are usually named a bit more verbose
-method-rgx = "([a-z_][a-z0-9_]{2,60}$)|(test_[a-z0-9_]{2,80}$)"
-
-# allowing exactly four digits at the beginning for migrations
-module-rgx = "([0-9]{4})?([a-z_][a-z0-9_]*)$"
+# snake_case, or PascalCaseFormset, because django does it that way
+# see https://docs.djangoproject.com/en/4.0/topics/forms/formsets/
+variable-rgx = "(^[a-z0-9_]*)$|(^[A-Za-z]*Formset$)"
+argument-rgx = "(^[a-z0-9_]*)$|(^[A-Za-z]*Formset$)"
```
niklasmohrin, line 31: Why capture the `^`, but not the `$`? And why not allow numbers in pascal case?

niklasmohrin (approved):

Issue #1638, opened 2021-10-18
Jumping form on submit


janno42:
When having the cursor in the participant field in an evaluation form and clicking on "Save" on the bottom of the page, the form jumps back to the participant field (now not having the focus anymore). When clicking "Save" again, the form is saved.

The form should be saved when clicking "Save" for the first time, the form should not jump back to the participant field.

dieknolle3333:
This seems to be an issue in select2: https://output.jsbin.com/cosocudiqe.
I opened an issue in their repo regarding this.

janno42:
will be handled in #1461

PR #1758, opened 2022-05-12
Replace select2 with TomSelect

He3lixxx:
* [x] Some TODOs are left in the code
* [x] Licensing? Tomselect uses the Apache2.0 license
* [x] The minimum input length turns out to be harder than initially thought, because the `shouldLoad` function seems to only apply to network loads (using `load`), and not showing options that are already locally available (such as the select options in the html). I wasn't able to come up with a nice solution for that yet.
* [x] ~Frontend tests?~
* [x] ~Move stuff into a .ts file? The base template could use some cleanup.~
* [x] Jumpy bug -- needs minrep example and bugreport with tomselect. It works on their examples.
* [x] For @niklasmohrin, on firefox, the tomselected selects didn't work at all.
* [x] Add another doesn't seem to work properly.
* [x] If converting to typescript is tricky, maybe move into seperate base.js file? (discuss) (edit: I don't feel like it.)
* [x] Tag inputs: Enter can submit (see https://github.com/e-valuation/EvaP/pull/1758#issuecomment-1141396401)

All other parts should behave as expected, so feel free to try it out.


janno42:
my failed experiments with `shouldLoad` a while ago were the reason for me to pause working on the issue :)
i'm looking forward for good ideas.

do you have any concerns with the apache license? is there anything to do besides adding the license info to our license file?

He3lixxx:
I've asked the internet regarding minimum input length here: https://stackoverflow.com/q/72233401/12345551.

About the license: I don't know yet.

evap/evaluation/templates/base.html
```diff
@@ -85,41 +83,70 @@
...
-                        allowClear: this.multiple,
-                        dropdownParent: parent
-                    });
+                        hidePlaceholder: true,
+                        render: {
+                            option_create: function( data, escape ){return '<div class="create">' + escape(data.input) + '</div>';},
+                            no_results: function( data, escape ){return '<div class="no-results">{% trans "No results found" %}</div>';},
+                        },
+                        plugins: {
+                            // "dropdown_input": {}, // # TODO: Do we want this? I think both are a bit weird, with and without
```
He3lixxx, line 124: @janno42 What's your opinion here?
* Without this plugin, the typing to search an entry, the typed text will be visible directly next to the current selected item, even on single-select widgets.
* With this plugin, the typed text is shown in the dropdown (which I think we want), but if no item was selected, the height of the field will break:
</img>

Maybe worth a bugreport with tomselect?

evap/evaluation/templates/base.html
```diff
@@ -85,41 +83,70 @@
...
-            applySelect2($('select#id_delegate_to'), $('#delegateSelectionModal'));
+            applyTomSelect(document.querySelectorAll("select:not(.form-template select):not(.no-tomselect)"));
 
             // don't use the placeholder text on disabled select2 elements
-            applySelect2Disabled = function(elem) {
-                elem.select2();
-            };
-            applySelect2Disabled($('select[disabled]').not('.form-template select'));
+            // # TODO
+            // applySelect2Disabled = function(elem) {
```
He3lixxx, line 146: Where on the page do we have disabled select fields that have no value set and thus would show the placeholder? And why do we want to handle them different? Isn't a disabled "Please select..." pretty clear while a disabled field with no contents looks like a frontend error?

evap/evaluation/templates/base.html
He3lixxx: The resized input field happens because they set the input field with the placeholder inside it to `display: None !important`. Setting it to `visibility: hidden` instead fixes it. We may be able to fix this in our CSS, and Tom Select would probably update their plugin if we issue a bug report.

evap/evaluation/templates/base.html
```diff
@@ -85,41 +83,75 @@
...
+                                }
+                            }
+                        },
+                        plugins: {},
+                    };
+                    if(element.hasAttribute("data-tomselect-fullwidth")) {
+                        // span needed to the "remove this icon" button is right-aligned
+                        baseOptions.render.item = function(data, escape) {return '<div class="w-100"><span class="w-100">' + escape(data.text) + '</span></div>';};
+                    }
+                    if(element.multiple) {
```
niklasmohrin, line 147: ah, yes, a type annotation would have explained that, didn't really know what this was supposed to mean for a moment there :D

evap/evaluation/templates/base.html
```diff
@@ -85,41 +83,75 @@
...
+                                } else {
+                                    return '<div class="no-results">{% trans "No results found" %}</div>';
+                                }
+                            }
+                        },
+                        plugins: {},
+                    };
+                    if(element.hasAttribute("data-tomselect-fullwidth")) {
+                        // span needed to the "remove this icon" button is right-aligned
+                        baseOptions.render.item = function(data, escape) {return '<div class="w-100"><span class="w-100">' + escape(data.text) + '</span></div>';};
```
niklasmohrin, line 145: This (and similar online functions) could use the arrow function syntax. Additionally, this could use template strings, not sure which would be more readable though

niklasmohrin (commented):
strong :+1: from me on porting this to typescript and moving into its own file :D

evap/evaluation/templates/base.html
```diff
@@ -85,41 +83,75 @@
...
+                        baseOptions.plugins.remove_button = {"title": "{% trans 'Remove this item' %}"};
+                    }
+                    new MinimumInputLengthTomSelect(element, Object.assign({}, baseOptions, additionalOptions));
+                });
             };
-            applySelect2Disabled($('select[disabled]').not('.form-template select'));
+            applyTomSelect(document.querySelectorAll("select:not(.form-template select):not(.no-tomselect)"));
 
             // fix Bootstrap bug that selecting toggle buttons via keyboard controls is only visually
             // https://github.com/twbs/bootstrap/issues/26855
```
niklasmohrin, line 157: Is this still needed? The issue is closed :thinking:

He3lixxx: Removed.

He3lixxx:
I gave up trying to convert the js to typescript, because our code there is heavily intertwined with tomselect stuff (e.g. we inherit from it). To use that in typescript means we have to import the tomselect class into the typescript file, which afaik can't work with the `min.js` file they distribute, so we'd have to add tomselect as a npm dependency, and compile/bundle it ourself, then licensing is an issue again.

evap/evaluation/templates/base.html
He3lixxx: Opened issue 383 on tomselect, they committed a fix, so this should work in the next tomselect release. Then, we probably want to use the "dropdown_input" plugin?

He3lixxx: Their issue was fixed and they released 2.0.3. Updated the version in this PR.

evap/evaluation/templates/base.html
```diff
@@ -85,49 +83,73 @@
...
+            class MinimumInputLengthTomSelect extends TomSelect {
+                search(query) {
+                    if(query.length < this.settings.minimumInputLength) {
+                        return this.sifter.prepareSearch(query, {});
+                    }
+                    return super.search(query);
+                }
+            }
+
+            applyTomSelect = function(elements, additionalOptions = {}) {
```
He3lixxx, line 114: (random line of code )

TomSelect has the same issue as we had with select2 in #1638: In all kinds of bad situations, it will make the page scroll to some other location:
* most notably currently: Single select, clearing with backspace
* Also, just like in #1638: When opening the dropdown, scrolling somewhere else, and then closing the dropdown by clicking somewhere, the page will scroll back up.

evap/evaluation/templates/base.html
```diff
@@ -85,49 +83,73 @@
...
-                        minimumResultsForSearch: 15,
-                        minimumInputLength: $(this).children('option').length > 50 ? 3 : 0,
-                        allowClear: this.multiple,
-                        dropdownParent: parent
-                    });
-                });
+            // used in derived templates
+            const tagTomSelectOptions = {
+                create: true,
+                dropdownParent: document.createElement("div"),
```
niklasmohrin, line 102: This reuses the same `div` for all users, I am not sure if this is a problem though :thinking:

evap/evaluation/templates/base.html
```diff
@@ -85,49 +83,73 @@
...
+                        element.classList.add("is-invalid");
+                    }
+                    const baseOptions = {
+                        createOnBlur: true,
+                        placeholder: "{% trans 'Please select...' %}",
+                        hidePlaceholder: true,
+                        minimumInputLength,
+                        render: {
+                            option_create: ( data, escape ) => `<div class="create">${ escape(data.input) }</div>`,
+                            no_results: ( data, escape ) => {
```
niklasmohrin, line 131:
```suggestion
                            option_create: (data, escape) => `<div class="create">${ escape(data.input) }</div>`,
                            no_results: (data, escape) => {
```

niklasmohrin (commented):
haven't checked in browser yet, code is good though

evap/evaluation/templates/base.html
He3lixxx: Doesn't matter, the div is not part of the document. The purpose is to hide the dropdown. Added a comment explaining that.

evap/evaluation/templates/base.html
He3lixxx: > Where on the page do we have disabled select fields that have no value set and thus would show the placeholder? 

@janno42 do you know of such a disabled field?

evap/evaluation/templates/base.html
He3lixxx: Some of the scrolling is caused by our custom `invalid` listener here:
https://github.com/e-valuation/EvaP/blob/81f1d4a0d06347b9bc4ba546e0933f6e02066228/evap/evaluation/templates/base.html#L136-L139

This can be fixed by just using `scroll-padding-top` instead of this custom `invalid` handler.

The scroll-back-on-focus-lose is caused by tomselect's `dropdown_input` plugin. I opened issue 391 with them.

He3lixxx:
@niklasmohrin please try again whether inputs are still broken for you.

evap/evaluation/templates/base.html
janno42: i can't think of such a field right now. it's probably okay to have the placeholder text in disabled fields as well.

niklasmohrin:
Some issues on the degrees page:
- the "no results" is always shown; while it may be true, it is a bit annoying
- when adding a tag a second time (for example, try to add "master" again) and hitting enter, the form submits again

evap/evaluation/templates/base.html
```diff
@@ -85,58 +83,78 @@
...
+                            option_create: (data, escape) => `<div class="create">${ escape(data.input) }</div>`,
+                            no_results: (data, escape) => {
+                                if(data.input.length < minimumInputLength) {
+                                    return `<div class="no-results">{% trans "Please enter ${ minimumInputLength } characters or more..." %}</div>`;
+                                } else {
+                                    return '<div class="no-results">{% trans "No results found" %}</div>';
+                                }
+                            }
+                        },
+                        plugins: {},
```
niklasmohrin, line 143:
```suggestion
```

niklasmohrin (commented):
There is still a mention of select2 in `staff_quesionnaire_form`

evap/evaluation/templates/base.html
He3lixxx: Required to be an object so we can use
```
baseOptions.plugins.clear_button = {"title": "{% trans 'Remove all items' %}"};
```
below -- or is there a smarter way to achieve that?

He3lixxx:
> Some issues on the degrees page:
> 
>     * the "no results" is always shown; while it may be true, it is a bit annoying
> 
>     * when adding a tag a second time (for example, try to add "master" again) and hitting enter, the form submits again

Both essentially come from the same tomselect bug:
* When the dropdown is hidden by making it a child of a `div` that's not in the document, it can not have the keyboard focus. When it doesn't have the keyboard focus, hitting enter will not add the current text as a new tag, but instead submit the form instead. Should be fixed in tomselect issue 389.
    * As a solution for now, I've added the dropdown again -- but, for the tag input, there are no options available, so it would always say "no results"
    * "Enter" is then handled by the "Add [your current text]" entry in the dropdown. However, that entry is removed if you type something that already is a tag -> enter submits again.
    
Possible solutions:
* Wait for tomselect to fix 389, then re-evaluate
* Intercept enter on tomselect inputs, `preventDefault` or issue "tab" instead
* Train staff users to always use Tab instead of enter to add new tags.

niklasmohrin:
When selecting entries, the color is very bold compared to our palette:

image

@janno42 what do you think?

evap/staff/templates/staff_course_type_index.html
```diff
@@ -12,7 +12,7 @@
         <a href="{% url 'staff:course_type_merge_selection' %}" class="btn btn-sm btn-secondary">{% trans 'Merge course types' %}</a>
     </div>
 
-    <form id="course-type-form" method="POST" enctype="multipart/form-data" class="form-horizontal select2form">
```
niklasmohrin, line 15: (not related to this line)

When entering something in the "import names" form on course type index or degree index and hitting "tab" (as we will be teaching our staff users), I accept the suggested tag but also tab to the next input

evap/staff/templates/staff_evaluation_form.html
```diff
@@ -19,7 +19,7 @@ <h3>
             {% include 'course_badges.html' with course=evaluation.course %}
         </div>
     {% endif %}
-    <form id="evaluation-form" method="POST" class="form-horizontal multiselect-form select2form">
```
niklasmohrin, line 22: When selecting entries, the color is very bold compared to our palette:

image
@janno42 what do you think?

evap/staff/templates/staff_questionnaire_form.html
```diff
@@ -89,19 +89,27 @@ <h5 class="card-title">{% trans 'Questions' %}</h5>
...
+                registerSelectChangedHandlers();
             };
             makeFormSortable("question_table", "questions", rowChanged, rowAdded, "", true, true);
 
-            // enable or disable checkbox for additional text answers
-            $('.question-type select').on('select2:select', function (e) {
+            const selectChangedHandler = function(e) {
                 const checkbox = $($(this).closest("td.question-type")).find("input");
-                if (e.params.data.id == 0 || e.params.data.id == 5) {  // 0: Text question; 5: Heading
+                if (e.target.value == 0 || e.target.value == 5) {  // 0: Text question; 5: Heading
```
niklasmohrin, line 98: (We talked about this in person)

Although `e.target` should always equal `selectElement`, but let's be sure. Some options
- Currying: `const selectChangeHandler = selectElement => event => { ... }`.although I normally like this approach best, we need to store the curried functions to keep their identity
- Using `this` in a `function` (not arrow function). Yeah, let's not do `this`
- Using `event.currentTarget`. According to mdn, this is equivalent to `this` without the complicated rules of `this` and allows for arrow functions - so a refactor would not break this 

What do you think?

evap/staff/templates/staff_questionnaire_form.html
```diff
@@ -89,19 +89,27 @@ <h5 class="card-title">{% trans 'Questions' %}</h5>
...
                 } else {
                     checkbox.prop("disabled", false);
                 }
-            });
+            };
+
+            const registerSelectChangedHandlers = function() {
+                document.querySelectorAll(".question-type select").forEach(selectElement => {
+                    selectElement.removeEventListener('change', selectChangedHandler);
+                    selectElement.addEventListener('change', selectChangedHandler);
```
niklasmohrin, line 109: From mdn/removeEventlistener (emphasis mine):

> While addEventListener() will let you add the same listener more than once for the same type _if the options are different_

So the remove should be redundant, right?

evap/staff/templates/staff_course_type_index.html
He3lixxx: Maybe we should just intercept enter submitting the form, but not teach anything specifically? It would do the right thing: Enter only submits if you typed an option that already is a tag.

evap/staff/templates/staff_evaluation_form.html
janno42: i think we should redefine the colors to better match our design

evap/contributor/templates/contributor_evaluation_form.html
```diff
@@ -26,7 +26,7 @@ <h3>
         {{ evaluation.full_name }} ({{ evaluation.course.semester.name }})
```
janno42, line 26: - Dropdown for tags can stay (to add options with Enter)

janno42, line 26: - Comma as auto-separator for tags should be removed (so that a comma can be part of text answer warnings)

janno42, line 26: - The search text entered in a multi-select field should be removed after adding an entry

janno42 (changes requested):

evap/staff/templates/staff_evaluation_form.html
He3lixxx: The color they specify here, #0d6efd, seems to be the bootstrap 5 "active item" default, as it is expanded from `$component-active-bg` here:
https://github.com/orchidjs/tom-select/blob/e466189df876a061df38740c8c02e95daf8dbdbc/src/scss/tom-select.bootstrap5.scss#L24

This is compiled to the CSS rule:
```css
.ts-wrapper.multi .ts-control>div.active {
    background:#0d6efd;
    /// [...]
}
 ```

So it seems we would have to override this. What color would be like? The bg-primary background?

It seems like we use `bg-primary` as defined by bootstrap right now, so I'm a bit confused why the colors mismatch. SCSS can be so confusing...

janno42: please make adjustments (in `static/scss/_adjustments.scss`) to the selection and box-shadow colors to match the current evap colors

janno42 (approved):
looks good so far :)

niklasmohrin (approved):

PR #1759, opened 2022-05-13
Upgrade pylint

He3lixxx:
PyCQA removed the no-init check in issue 2409 / pr 6373 (claiming that it was never issued -- it's failing right now in #1758)

It seems that in #1756, the fact that pylint didn't check the migrations (which violate some of our rules from that PR) was simply a bug in pylint. With the newer version, it now triggers:
* invalid-name on all migration files (because their names start with numbers)
* invalid-name on all models inside migration files (because migrations do stuff like `Questionnaire = apps.get_model('evaluation', 'Questionnaire')`
* cyclic-import (This has always been flaky between single-threaded and multi-threaded) triggers for some of our in-function-imports.

We can't (yet) configure pylint per directory (-> so we can't have a special handling for migrations)
I'd simply allow leading numbers in module names, and PascalCase for variables for now.

For the cyclic imports: If these cause issues, I'd expect tests to fail (and not even start). Thus I'd just disable the pylint check.

### DIFF
--- a/evap/evaluation/migrations/0061_editor_review_reminder_template.py
+++ b/evap/evaluation/migrations/0061_editor_review_reminder_template.py
@@ -16,7 +16,7 @@ def insert_emailtemplates(apps, _schema_editor):
 def remove_emailtemplates(apps, _schema_editor):
     EmailTemplate = apps.get_model("evaluation", "EmailTemplate")
 
-    for name, subject in emailtemplates:
+    for name, __ in emailtemplates:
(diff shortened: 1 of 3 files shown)


janno42 (approved):

He3lixxx:
Merging already with pylint 2.14.0b1.
I couldn't find any release schedule for pylint 2.14. Judging from past releases, the time difference between "b1" and release could be around 30 days, so we can't really wait for the release here (failing on master now, blocking other PRs).

PR #1762, opened 2022-05-16
Remove some more hardcoded PKs from tests

He3lixxx:
The shuffled tests on main failed [here](https://github.com/e-valuation/EvaP/runs/6451237165?check_suite_focus=true) with shuffle 8672148966. I think in #1746, I only searched for `pk=1`, `pk=2` and `id=1`, so I missed these.

There are still some hardcoded ids in results/tests/test_views.py in test classes that use the `minimal_test_data_results` fixture. We should probably try to remove that fixture anyways and create the objects in python? I has 40 objects though, so it's a bit of work.


niklasmohrin (approved):
haven't checked whether there is more, these changes look good though

PR #1764, opened 2022-05-17
Fix error on vote preview without any questionnaires assigned

### LINKED ISSUE #1747: Staff vote preview page is accessible, but crashes when no questionnaires are assigned
We currently have this situation with http://localhost/staff/semester/21/evaluation/1668/preview. There are two contributions for this evaluation, but no questionnaires are assigned.

The staff `evaluation_preview` view calls `get_valid_form_groups_or_render_vote_page`, which in turn will check whether all form groups are valid -- there are no form groups, so this is the case. The method then asserts `not preview`, which fails.

The preview button on the staff semester view that leads to this page is enabled (it is shown if `evaluation.state < evaluation.State.IN_EVALUATION`).

@janno42 how to we want to handle evaluations without any assigned questionnaires, regarding preview as well as starting the evaluation. From what I see, starting an evaluation that has no questionnaires assigned is currently possible?

He3lixxx:
Fixes #1747. The preview will just be empty:

</img>

So far, we've always added the `regression test for issue ####` as a docstring. However, unittest will then use these docstrings when running the tests in verbose, instead of giving the module+class+method name, which is mildly annoying when trying to find out test order dependencies. Sample output with `-v 3`:
```
test_single_locked_questionnaire (evap.contributor.tests.test_views.TestContributorEvaluationEditView) ... ok
test_wrong_state (evap.contributor.tests.test_views.TestContributorEvaluationEditView)
Asserts that a contributor attempting to edit an evaluation ... ok
test_wrong_usergroup (evap.contributor.tests.test_views.TestContributorEvaluationEditView)
Asserts that a user who is not part of the usergroup ... ok
test_check_response_code_200 (evap.contributor.tests.test_views.TestContributorEvaluationPreviewView) ... ok
test_without_questionnaires_assigned (evap.contributor.tests.test_views.TestContributorEvaluationPreviewView) ... ok
Regression test for #846 ... ok
```

Based on that, I'd say we should maybe remove our docstrings, or replace them by usual comments. There is also a [way to tell unittest to not do that](https://stackoverflow.com/questions/12962772/how-to-stop-python-unittest-from-printing-test-docstring), but I think our docstrings often do not really add any value.

### DIFF
--- a/evap/student/views.py
+++ b/evap/student/views.py
@@ -133,6 +133,7 @@ def get_vote_page_form_groups(request, evaluation, preview):
         ]
     return form_groups
 
+
 def render_vote_page(request, evaluation, preview, for_rendering_in_modal=False):
     form_groups = get_vote_page_form_groups(request, evaluation, preview)
 
@@ -182,7 +183,7 @@ def render_vote_page(request, evaluation, preview, for_rendering_in_modal=False)
 
 @participant_required
 def vote(request, evaluation_id):
-    # pylint: disable=too-many-locals,too-many-nested-blocks,too-many-branches
+    # pylint: disable=too-many-nested-blocks,too-many-branches
     evaluation = get_object_or_404(Evaluation, id=evaluation_id)
     if not evaluation.can_be_voted_for_by(request.user):
         raise PermissionDenied
--- a/evap/student/views.py
+++ b/evap/student/views.py
@@ -116,7 +116,7 @@ def index(request):
     return render(request, "student_index.html", template_data)
 
 
-def get_valid_form_groups_or_render_vote_page(request, evaluation, preview, for_rendering_in_modal=False):
+def get_vote_page_form_groups(request, evaluation, preview):
     contributions_to_vote_on = evaluation.contributions.all()
     # prevent a user from voting on themselves
     if not preview:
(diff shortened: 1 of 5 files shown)


janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

evap/student/views.py
```diff
@@ -132,11 +132,10 @@ def get_valid_form_groups_or_render_vote_page(request, evaluation, preview, for_
             for questionnaire in questionnaires
         ]
 
-    if all(all(form.is_valid() for form in form_group) for form_group in form_groups.values()):
-        assert not preview
+    if not preview and all(all(form.is_valid() for form in form_group) for form_group in form_groups.values()):
```
niklasmohrin, line 135: It took me some time to understand the `not preview` condition here - I find it confusing that the `preview` parameter influences the return type. Intuitively, the parameter should change the content of the returned value, not its shape. This whole `get_valid_form_groups_or_render_vote_page` thing is a bit weird anyways - there is only one place that actually uses the `form_group` result, all the others use `[1]` on the returned tuple. I suppose, we could refactor this to one function that builds the formgroups and another one that returns the rendered page, given the formgroups. We could make the form group argument optional (and default to the result of calling the former (pun intended)) to keep caller code short

niklasmohrin (approved):

PR #1770, opened 2022-05-30
Fix #1704: Allow anonymous feedback

### LINKED ISSUE #1704: Allow anonymous feedback messages via settings
When sending a message in the feedback form, the sender field is currently automatically filled and disabled; thus users cannot give anonymous feedback.
This should be made optional via a new setting `ALLOW_ANONYMOUS_FEEDBACK_MESSAGES`. When set to `FALSE`, the current behavior is kept. When set to `TRUE`, a new checkbox is shown in the form, enabling users to send their message anonymously. Checking this checkbox will remove the sender information in the form and in the sent message's text.

Kakadus:
Fixes #1704


Kakadus:
Currently there are multiple options for UI/UX:
-> we could have the buttons in a new line, 100% big (which looks a bit odd)
Screenshot from 2022-05-30 22-28-30

-> we could align them inline-right
Screenshot from 2022-05-30 22-24-17

-> or align them inline (centered?). Then, the subject line would probably also need to be aligned inline.
Screenshot from 2022-05-30 22-29-49

Formatting and functionality will be done in two weeks

He3lixxx:
Just an observation / opinion dump, please wait for @janno42's opinion on the matter before doing any of these.

My 2 cents regarding layout:
* inline align-right or inline centered look misplaced to me.
* I think we want the sender and subject lines to be a single line each, laid out as a grid, with the inputs growing to fill the width. If the disabled input becomes a button group (as it will do in this PR), it should simply be split in half (note: I don't think the "evap" button should be smaller because the word is shorter unless absolutely necessary to prevent wrapping / overflowing)

While we're at it:
* I think the "Message: " label can be removed, and instead, we can add a "Enter your message..." placeholder for the textarea.
* The line between the message and the buttons is there because the buttons are part of the footer. I think it looks a little nicer if the card has no footer and the buttons are part of the body. Probably something to discuss first, as it affects a few modals (@janno42 doesn't a card footer usually have a little darker background? Probably worth discussing in person. I currently know of three places using the footer for action buttons: contact modal, delegation modal, (probably all modals?), staff user index).
* I noticed that we don't have a uniform "Cancel" button placement. The "delegate preparation" modal has the cancel button on the right, the contact modal has it on the left. We might want to fix that.

</img>

package.json
```diff
@@ -30,5 +30,8 @@
         "globalSetup": "jest-environment-puppeteer/setup",
         "globalTeardown": "jest-environment-puppeteer/teardown",
         "testEnvironment": "jest-environment-puppeteer"
+    },
+    "dependencies": {
+        "@types/bootstrap": "^5.1.12"
     }
```
niklasmohrin, line 36: Please move this into `devDependencies`, making sure to keep them sorted

niklasmohrin: Additionally, please rerun `npm install` to update `package-lock.json` afterwards

evap/static/ts/src/csrf-utils.ts
```diff
@@ -12,6 +12,7 @@ function getCookie(name: string): string | null {
     return null;
 }
 const csrftoken = getCookie("csrftoken")!;
+export const csrfheader = { "X-CSRFToken": csrftoken };
```
niklasmohrin, line 15:
```suggestion
export const csrfHeaders = { "X-CSRFToken": csrftoken };
```

evap/static/ts/src/contact_modal.ts
```diff
@@ -0,0 +1,64 @@
+declare const csrf: typeof import("./csrf-utils");
```
niklasmohrin, line 1:
```suggestion
declare const csrfUtils: typeof import("./csrf-utils");
```

evap/static/ts/src/contact_modal.ts
```diff
@@ -0,0 +1,64 @@
...
+declare const bootstrap: typeof import("bootstrap");
+
+class ContactModalLogic {
+    private readonly modal: bootstrap.Modal;
+    private readonly successMessageModal: bootstrap.Modal;
+    private readonly actionButtonElement: HTMLButtonElement;
+    private readonly messageTextElement: HTMLInputElement;
+    private readonly modal_id: string;
```
niklasmohrin, line 9: We want to stick to one naming style, the default in javascript is camel case, so let's change it to
```suggestion
    private readonly messageTextElement: HTMLInputElement;
    private readonly modalId: string;
```

evap/static/ts/src/contact_modal.ts
```diff
@@ -0,0 +1,64 @@
...
+    private readonly title: string;
+
+    constructor(modal_id: string, title: string) {
+        this.modal_id = modal_id;
+        this.title = title;
+        this.modal = new bootstrap.Modal(<HTMLElement>document.getElementById(modal_id));
+        this.successMessageModal = new bootstrap.Modal(
+            <HTMLElement>document.getElementById("successMessageModal_" + modal_id),
```
niklasmohrin, line 17:
```suggestion
            document.getElementById<HTMLElement>("successMessageModal_" + modal_id),
```

evap/static/ts/src/contact_modal.ts
```diff
@@ -0,0 +1,64 @@
...
+        );
+        this.actionButtonElement = document.getElementById(this.modal_id + "ActionButton") as HTMLButtonElement;
```
niklasmohrin, line 19: I think this should work, but I am not 100% sure:
```suggestion
        this.actionButtonElement = document.getElementById<HTMLButtonElement>(this.modal_id + "ActionButton");
```

evap/static/ts/src/contact_modal.ts
```diff
@@ -0,0 +1,64 @@
...
+        this.messageTextElement = document.getElementById(this.modal_id + "MessageText") as HTMLInputElement;
+        this.actionButtonElement.addEventListener("click", event => {
+            this.ModalAction(event);
+        });
```
niklasmohrin, line 23:
```suggestion
        this.actionButtonElement.addEventListener("click", this.ModalAction);
```

evap/static/ts/src/contact_modal.ts
```diff
@@ -0,0 +1,64 @@
...
+        res_promise.then(res => {
+            if (!res.ok) {
+                window.alert("Sending failed, sorry!");
+                return;
+            }
+            this.modal.hide();
+            this.successMessageModal.show();
+            this.messageTextElement.value = "";
+
+            const timeout = 3000;
```
niklasmohrin, line 56: Let's move this to the top of the file and give it a more descriptive name

evap/static/ts/src/contact_modal.ts
```diff
@@ -0,0 +1,64 @@
...
+            <HTMLElement>document.getElementById("successMessageModal_" + modal_id),
+        );
+        this.actionButtonElement = document.getElementById(this.modal_id + "ActionButton") as HTMLButtonElement;
+        this.messageTextElement = document.getElementById(this.modal_id + "MessageText") as HTMLInputElement;
+        this.actionButtonElement.addEventListener("click", event => {
+            this.ModalAction(event);
+        });
+    }
+
+    public ModalShow(): void {
```
niklasmohrin, line 26: The methods should be named in camel case, i.e. start with a lowercase letter

evap/static/ts/src/contact_modal.ts
```diff
@@ -0,0 +1,64 @@
...
+        const message = this.messageTextElement.value;
+        const anonymous = (<HTMLInputElement>document.getElementById(this.modal_id + "AnonymName")).value;
+        if (message.trim() == "") {
+            this.modal.hide();
+            this.actionButtonElement.disabled = false;
+            return;
+        }
+
+        const res_promise = fetch("/contact", {
+            body: JSON.stringify({ message: message, title: this.title, anonymous: anonymous }),
```
niklasmohrin, line 42:
```suggestion
            body: JSON.stringify({ message, title: this.title, anonymous }),
```

evap/static/ts/src/contact_modal.ts
```diff
@@ -0,0 +1,64 @@
...
+            headers: csrf.csrfheader,
+            method: "POST",
+        });
+
+        res_promise.then(res => {
```
niklasmohrin, line 47: I think it is fine if you make the whole surrounding function `async` and turn this into
```javascript
try {
    const response = await fetch(...);
    // ...
} catch (_) {
    window.alert(...);
}
```

evap/static/ts/src/contact_modal.ts
```diff
@@ -0,0 +1,64 @@
...
+            }
+            this.modal.hide();
+            this.successMessageModal.show();
+            this.messageTextElement.value = "";
+
+            const timeout = 3000;
+            setTimeout(() => {
+                this.successMessageModal.hide();
+                this.actionButtonElement.disabled = false;
+            }, timeout);
```
niklasmohrin, line 60: _If_ we make this async, you could introduce a helper that let's you `await` a timeout, like in https://stackoverflow.com/a/33292942/13679671

evap/static/ts/src/contact_modal.ts
niklasmohrin: Hm, actually this makes it so that all the following code is in the `try` too; I suppose it should then be
```javascript
let response;
try {
    response = await fetch(...);
} catch (_) {
    // ...
}
// whatever should be done with `response`

File: evap/static/ts/src/contact_modal.ts (hunk 5 above)
Comment (Kakadus) [reply evap/static/ts/src/contact_modal.ts:17]: Does not work. In this case, we can use the ! syntax

**Comment by janno42** [comment]
re layout discussion above: i agree with He3lixxx in almost all points.
- the field labels should be placed inline and the "message" label can be removed in favor of a placeholder text.
- the sender button group should be left-aligned but not stretched to the full width. the button group could fill about two thirds of the width or so.
- the submit and cancel buttons should be placed outside of the current footer which can be removed. the current order (cancel left, send right) is correct and should be used in the other modals as well - i'll open another issue for that.

File: evap/evaluation/templates/footer.html [hunk 13]
```diff
@@ -22,7 +22,7 @@
             <div class="navbar-nav justify-content-end feedback-button-placeholder">
                 {% if user.is_authenticated %}
                     <span class="feedback-button">
-                        <button type="button" class="btn btn-dark" onclick="feedbackModalShow();">
+                        <button type="button" class="btn btn-dark" onclick="logic.showModal();">
```
niklasmohrin, line 25: It's probably best to add an event listener in javascript right after the declaration of `logic`

evap/static/ts/src/contact_modal.ts
```diff
@@ -0,0 +1,62 @@
+declare const csrfUtils: typeof import("./csrf-utils");
+declare const bootstrap: typeof import("bootstrap");
+declare const utils: typeof import("./utils");
+
+const timeout = 3000;
```
niklasmohrin, line 5: what is this timeout for? (I know, but new readers will not)

evap/static/ts/src/utils.ts
```diff
@@ -0,0 +1,13 @@
+export const selectOrError = <T extends Element>(elementId: string): T => {
+    const elem = document.querySelector<T>(elementId);
+    if (elem === null) throw new Error(`Element with id ${elementId} not found`);
```
niklasmohrin, line 3: I wanted to suggest this change:
```suggestion
    assert(elem !== null, `Element with id ${elementId} not found`);
```
But in the current setup, tsc cannot verify that elem is not null after that, unless we tell it that the assert functions really does assert. For that, we need to make `condition` to type `unknown` and not use arrow functions:
```javascript
export function assert(condition: unknown, message: string = "Assertion Failed"): asserts condition {
    if (!condition) throw new Error(message);
}
```

evap/evaluation/templates/contact_modal.html
```diff
@@ -1,4 +1,6 @@
-<div class="modal fade" id="successMessageModal_{{ modal_id }}" tabindex="-1" role="dialog" aria-labelledby="successMessageModalLabel_{{ modal_id }}" aria-hidden="true">
+{% load static %}
+<div class="modal fade" id="successMessageModal_{{ modal_id }}" tabindex="-1" role="dialog"
+     aria-labelledby="successMessageModalLabel_{{ modal_id }}" aria-hidden="true">
```
He3lixxx, line 3: Adding a linebreak is probably a good idea, but I think indenting with 5 spaces for that is rather odd (literally, har har).

@niklasmohrin what's the current formatting rule here?

evap/evaluation/templates/contact_modal.html
```diff
@@ -21,72 +24,48 @@ <h5 class="modal-title" id="{{ modal_id }}Label">{{ title }}</h5>
...
-                        <input type="text" class="form-control mb-3" id="{{ modal_id }}SenderName" disabled value="{{ user.full_name }}" />
-                    </div>
-                    <div class="mb-3">
-                        <label for="{{ modal_id }}Subject" class="control-label">{% trans 'Subject' %}:</label>
-                        <input type="text" class="form-control" id="{{ modal_id }}Subject" disabled value="{{ title }}" />
-                    </div>
-                    <div class="mb-3">
-                        <label for="{{ modal_id }}MessageText" class="control-label">{% trans 'Message' %}:</label>
-                        <textarea autofocus class="form-control" id="{{ modal_id }}MessageText" style="resize:vertical;min-height:150px;"></textarea>
+                    <br/><br/>
```
He3lixxx, line 27: I think we're using the spaced version rather consistently, so I'd leave this one as it is.

evap/evaluation/templates/contact_modal.html
```diff
@@ -21,72 +24,48 @@ <h5 class="modal-title" id="{{ modal_id }}Label">{{ title }}</h5>
...
+                            </div>
+                        {% else %}
+                            <input type="text" class="form-control" id="{{ modal_id }}SenderName"
+                                   disabled
+                                   value="{{ user.full_name }}"/>
+                        {% endif %}
+                        <label for="{{ modal_id }}Subject"
+                               class="control-label my-auto">{% trans 'Subject' %}:</label>
+                        <input type="text" class="form-control" id="{{ modal_id }}Subject" disabled
+                               value="{{ title }}"/>
```
He3lixxx, line 51: @janno42 the subject submitted here is used without validation by the server, so users can send requests that contain forged titles. I can see this being an opportunity to social-engineer the staff users that get these emails (e.g. someone sending a "Request evaluation changes for XYZ" messages, possibly through the now new anonymous feedback option).

Do we want to do something about that?

evap/evaluation/templates/contact_modal.html
```diff
@@ -21,72 +24,48 @@ <h5 class="modal-title" id="{{ modal_id }}Label">{{ title }}</h5>
...
-                        <label for="{{ modal_id }}MessageText" class="control-label">{% trans 'Message' %}:</label>
-                        <textarea autofocus class="form-control" id="{{ modal_id }}MessageText" style="resize:vertical;min-height:150px;"></textarea>
+                    <br/><br/>
+                    <div class="mb-3 modal-grid">
+                        <label for="{{ modal_id }}SenderName"
+                               class="control-label my-auto">{% trans 'Sender' %}:</label>
+                        {% if modal_id == "feedbackModal" and allow_anonymous_feedback_messages %}
+                            <div class="btn-group w-75" role="group"
+                                 aria-label="Basic radio toggle button group">
+                                <input type="radio" class="btn-check" name="btnradio" id="{{ modal_id }}SenderName"
```
He3lixxx, line 34: I'd rather not use the default `value`s of "on" as I think most people working on evap do not know this. Could we just explicitly specify two `value`s here for the inputs?

evap/evaluation/views.py
```diff
@@ -163,15 +163,18 @@ def legal_notice(request):
 def contact(request):
     message = request.POST.get("message")
     title = request.POST.get("title")
-    email = request.user.email or f"User {request.user.id}"
+    email = (
+        settings.DEFAULT_FROM_EMAIL
+        if request.POST.get("anonymous") == "on"
```
He3lixxx, line 168: Can't do this: Users would be able to send a POST request with `anonymous=on` even if anonymous feedback was disabled in the settings. We have to validate the input here.

Probably best to just throw a SuspiciousOperation if anonymous was requested, but is disabled in the settings.

evap/static/ts/src/contact_modal.ts
```diff
@@ -0,0 +1,61 @@
...
+            const message = this.messageTextElement.value;
+            if (message.trim() == "") {
+                this.modal.hide();
+                this.actionButtonElement.disabled = false;
+                return;
+            }
+            try {
+                const response = await fetch("/contact", {
+                    body: new URLSearchParams({
+                        anonymous: this.anonymousRadioElement.value,
```
He3lixxx, line 37: Wouldn't this always be "on", the default `value` attribute of the radio elements?

He3lixxx (changes requested):

evap/evaluation/templates/contact_modal.html
niklasmohrin: yes, that is quite co-even

in general, indent 4 spaces, everything is else is kind of do as you please. For Python, our lines are max. 120 chars long, but in html we are not as strict. Personally, I always try to do what `prettier` does, which is [this](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAJgSwG4AIwA2AhgM4kC8AOiALYTpEG4BmR6c1um6VIJArmDBwyAWREkiAczij6jAPrBguOgwILuuAL7bOMIgCNMUdgA9eAWgCMnAE4QCcXlkYQp1SlFw+fRO5hElsSGcARO6IYAnrwCQhLiZNKy8gQAMkZhSipqilq6nP6BlgAW3OxQvDB2-BwgAHxevrgYOPjEZLy5BJauBO72js7U6BBg-DQIMNSN3s2+rXiEpBTU3ZaQsFMzTfN7i+0rXamlcGxwdjtzezctJQCsh51rJzCYME6c3LGCwmISyTk6gyoQ0ylUqU06B0egawAApLhqkQoCRcAByRKSGS4EhTdG4eHaVAAeges1uzVJWGwFMpPgOy2etBOhnoUSu9L2CKRdhRaPRAE0IPw7KoATiAO6kXG-CTMfjhKK4qYAOgJRN29OpODp+xJNL1LQNuq8OtpIAANCAIAAHN7QEjIUD+BySgAK-gQTpQjGlUSd1sMfLAAGs4DAAMpESZpExwZCsAh4oMh8NR21EMAmKTIaq1a1wGihdDsdAZKBSfjJABiEDsNCIMDeleQICI-BgECtIBKMBoBAA6mV4CRM8JI973jh3lE22AyD2THi7DB3XypI3E4wUyAAFYkMyRnNOACK-Ag8G3ybg1szdhXbb7A57toCsEH3BgJWQ1gATAADHeDh4oOfK2m2b4iBc2AJtaACOF7wOudo+u2JCWFAcBwGWPZ2HAiGYPh67SFuSBJrueI0JgeY1LefAnnA56Xgm5E7vRBiGJ+6DfsgAAs1rIpgBA5gAwhANBkSAIj3D2-B4gAKkYPoUfR2C1AAkqYUyRmAAT2gAgqYkYwFETjXniuhAA)

evap/evaluation/views.py
niklasmohrin: we could also test this :)

evap/evaluation/templates/contact_modal.html
Kakadus: Apart from [djhtml](https://github.com/rtts/djhtml) (indenter), I found [djlint](https://github.com/Riverside-Healthcare/djLint) (formatter and linter) which seems worth looking into. We might analyze if we can/cannot configure it to your needs. By default, it is a bit different than `prettier`:
```
<div class="modal fade"
     id="successMessageModal_{{ modal_id }}"
     tabindex="-1"
     role="dialog"
     aria-labelledby="successMessageModalLabel_{{ modal_id }}"
     aria-hidden="true">
    <div class="modal-dialog" role="document">
        <div class="modal-content">
            <div class="modal-header">
                <h5 class="modal-title" id="successMessageModalLabel_{{ modal_id }}">{% trans 'Message sent' %}</h5>
            </div>
            <div class="modal-body">{% trans 'Your message was successfully sent.' %}</div>
        </div>
    </div>
</div>
```
In the meanwhile, I just remove the introduced formatting changes

He3lixxx: Made #1777 for the discussion. Resolving this for now.

evap/evaluation/views.py
```diff
@@ -161,17 +162,19 @@ def legal_notice(request):
...
 @login_required
 def contact(request):
+    sent_anonymous = request.POST.get("anonymous") == "true"
+    if sent_anonymous and not settings.ALLOW_ANONYMOUS_FEEDBACK_MESSAGES:
+        raise SuspiciousOperation("Anonymous feedback messages are not allowed, however received one from user!")
     message = request.POST.get("message")
     title = request.POST.get("title")
-    email = request.user.email or f"User {request.user.id}"
+    email = settings.DEFAULT_FROM_EMAIL if sent_anonymous else request.user.email or f"User {request.user.id}"
     subject = f"[EvaP] Message from {email}"
```
He3lixxx, line 171: This would mean that for anonymous messages, the subject would be:
"[EvaP] Message from evap@hpi.de"

-- @janno42 does this match what we want to happen?

He3lixxx (approved):
Code looks good to me. Thanks!

evap/evaluation/templates/contact_modal.html
```diff
@@ -1,3 +1,4 @@
+{% load static %}
```
niklasmohrin, line 1: We usually have an extra line here:

```suggestion
{% load static %}

```

evap/evaluation/templates/contact_modal.html
```diff
@@ -22,71 +23,34 @@ <h5 class="modal-title" id="{{ modal_id }}Label">{{ title }}</h5>
...
+                    <button type="button" class="btn btn-light m-1" data-bs-dismiss="modal">{% trans 'Cancel' %}</button>
+                    <button type="submit" id="{{ modal_id }}ActionButton" class="btn btn-primary m-1">{% trans 'Send Message' %}</button>
                 </div>
             </form>
         </div>
     </div>
 </div>
+<script type="module" src="/static/js/contact_modal.js"></script>
+<script type="module">
+    import {ContactModalLogic} from "{% static 'js/contact_modal.js' %}";
```
niklasmohrin, line 53:
```suggestion
    import { ContactModalLogic } from "{% static 'js/contact_modal.js' %}";
```

evap/evaluation/views.py
janno42: The subject should be "[EvaP] Anonymous message" and the `reply_to` header should be empty.

evap/static/ts/src/contact_modal.ts
```diff
@@ -0,0 +1,61 @@
+declare const bootstrap: typeof import("bootstrap");
+
+import { selectOrError, sleep, assert } from "./utils.js";
+import { csrfHeader } from "./csrf-utils.js";
+
+const success_message_timeout = 3000;
```
niklasmohrin, line 6: Usually, global constants are written as `SCREAMING_SNAKE_CASE`:
```suggestion
const SUCCESS_MESSAGE_TIMEOUT = 3000;
```

niklasmohrin (commented):
Some things I found:

evap/evaluation/templates/contact_modal.html
```diff
@@ -22,71 +23,34 @@ <h5 class="modal-title" id="{{ modal_id }}Label">{{ title }}</h5>
...
-            error : function(json) {
-                actionButton.text('{% trans 'Sending failed, sorry!' %}');
-                setTimeout(function(){
-                    actionButton.text(originalText);
-                    actionButton.prop('disabled', false);
-                }, 4000);
-            }
-        });
-    }
+    new ContactModalLogic("{{ modal_id }}", "{{ title|escapejs }}");
```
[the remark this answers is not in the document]
He3lixxx: Neutral.  I think having the constructor do all the work is also fine if the average use case matches that behavior.

evap/evaluation/templates/contact_modal.html
```diff
@@ -22,18 +24,22 @@ <h5 class="modal-title" id="{{ modal_id }}Label">{{ title }}</h5>
...
-                        <input type="text" class="form-control mb-3" id="{{ modal_id }}SenderName" disabled value="{{ user.full_name }}" />
-                    </div>
-                    <div class="mb-3">
-                        <label for="{{ modal_id }}Subject" class="control-label">{% trans 'Subject' %}:</label>
-                        <input type="text" class="form-control" id="{{ modal_id }}Subject" disabled value="{{ title }}" />
-                    </div>
-                    <div class="mb-3">
-                        <label for="{{ modal_id }}MessageText" class="control-label">{% trans 'Message' %}:</label>
-                        <textarea autofocus class="form-control" id="{{ modal_id }}MessageText" style="resize:vertical;min-height:150px;"></textarea>
+                    <div class="mb-3 modal-grid">
```
janno42, line 27: please use the usual form layout we have e.g. in the evaluation edit form or the delegation modal form with bold labels, no colon and larger margin between label and field.

janno42 (changes requested):

evap/evaluation/tests/test_views.py
```diff
@@ -94,15 +94,36 @@ class TestFAQView(WebTestWith200Check):
...
+        # anonymous email
+        self.app.post(
+            "/contact",
+            params={"message": "feedback message", "title": "some title", "anonymous": "true"},
+            user=user,
+        )
+
+        self.assertEqual(len(mail.outbox), 2)
         self.assertTrue(mail.outbox[0].reply_to == ["user@institution.example.com"])
+        self.assertTrue(mail.outbox[1].reply_to != ["user@institution.example.com"])
```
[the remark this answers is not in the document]
niklasmohrin: Furthermore, there is ˋassertEqualˋ (or ˋassertEqˋ)

package.json
```diff
@@ -5,6 +5,7 @@
         "@types/jquery": "^3.5.14",
         "@types/puppeteer": "^5.4.6",
         "@types/sortablejs": "^1.13.0",
+        "@types/bootstrap": "^5.1.12",
```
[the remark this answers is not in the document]
He3lixxx: This is just a typing package, to be able to use bootstrap types in typescript. A minor version mismatch shouldn't cause any trouble here.

In the bigger picture, we might want to think about synchronizing non-npm-dependencies with their typing package here (but that also applies to sortablejs, and jquery).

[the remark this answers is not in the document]
Kakadus: As stated in [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped#how-do-definitely-typed-package-versions-relate-to-versions-of-the-corresponding-library), that "the type declaration package is unrelated to the library patch version. This allows Definitely Typed to safely update type declarations for the same major/minor version of a library". So, when the major/minor version matches, everything should work fine

evap/evaluation/templates/contact_modal.html
Kakadus: this is now done with .modal-grid .control-label rule in _modal.scss

niklasmohrin (commented):
Code looks good here too :)

niklasmohrin:
This is what I see currently:
image

I think it looks a bit odd that the subject line uses the full width, but the sender row stops in the middle of the modal box. Shouldn't it just split the entire width in two halfs? It would then look like this:
image
... which is also a bit weird. 

Leaving it at 75% width, but centering looks like this:
image

Finally, moving the headings left to the middle as well:
image

I think I like the last best, but I am not settled on either version. What the others think? @He3lixxx @janno42

Kakadus:
An independent customer review came to the result, that the last option is most esthetic, but the heading should be aligned right (only relevant if their length differs).

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

evap/static/ts/src/csrf-utils.ts
```diff
@@ -12,6 +12,7 @@ function getCookie(name: string): string | null {
     return null;
 }
 const csrftoken = getCookie("csrftoken")!;
+export const CSRF_HEADER = { "X-CSRFToken": csrftoken };
```
niklasmohrin, line 15: small nit: As this is an object / a collection of headers (with one element), this should be called `CSRF_HEADERS`

evap/static/scss/components/_modal.scss
```diff
@@ -2,6 +2,28 @@
     max-width: 1120px;
 }
 
+.modal-grid {
+    display: grid;
+    width: 75%;
+    margin: 1.5rem auto 1.5rem auto;
```
niklasmohrin, line 8: I think this is equivalent to
```suggestion
    margin: 1.5rem auto;
```

niklasmohrin (approved):
Looks good, only nits :)

evap/evaluation/templates/contact_modal.html
```diff
@@ -43,50 +48,9 @@ <h5 class="modal-title" id="{{ modal_id }}Label">{{ title }}</h5>
         </div>
     </div>
 </div>
+<script type="module" src="/static/js/contact_modal.js"></script>
```
He3lixxx: I can remove this and my browser still does everything correctly. Is it necessary? If not, we should probably remove it.

(edit: we concluded that it can probably removed here here -- at least it still works in chrome and firefox then)

(also, /static/ here, where the `static` template tag is used below.)

evap/evaluation/tests/test_views.py
```diff
@@ -94,15 +94,36 @@ class TestFAQView(WebTestWith200Check):
...
             "/contact",
-            params={"message": "feedback message", "title": "some title", "sender_email": "unique@mail.de"},
+            params={"message": "feedback message", "title": "some title", "anonymous": "false"},
             user=user,
         )
-        self.assertEqual(len(mail.outbox), 1)
-        self.assertTrue(mail.outbox[0].reply_to == ["user@institution.example.com"])
+        # anonymous email
+        self.app.post(
+            "/contact",
```
He3lixxx, line 108: Would it make sense to have this as `self.url`, maybe from a `reverse()` call? (It's not three times in the tests here).

evap/evaluation/views.py
```diff
@@ -161,17 +162,23 @@ def legal_notice(request):
...
+    sent_anonymous = request.POST.get("anonymous") == "true"
+    if sent_anonymous and not settings.ALLOW_ANONYMOUS_FEEDBACK_MESSAGES:
+        raise SuspiciousOperation("Anonymous feedback messages are not allowed, however received one from user!")
     message = request.POST.get("message")
     title = request.POST.get("title")
-    email = request.user.email or f"User {request.user.id}"
-    subject = f"[EvaP] Message from {email}"
-
+    if sent_anonymous:
+        email = "anonymous user"
```
He3lixxx, line 171: maybe name this "sender" or something (It's not always an email).

evap/static/ts/src/contact_modal.ts
```diff
@@ -0,0 +1,67 @@
...
+    // may be null if anonymous feedback is not enabled
+    private readonly anonymousRadioElement: HTMLInputElement | null;
+
+    constructor(modalId: string, title: string) {
+        this.title = title;
+        this.modal = new bootstrap.Modal(selectOrError("#" + modalId));
+        this.successMessageModal = new bootstrap.Modal(selectOrError("#successMessageModal_" + modalId));
+        this.actionButtonElement = selectOrError("#" + modalId + "ActionButton");
+        this.messageTextElement = selectOrError("#" + modalId + "MessageText");
+        this.anonymousRadioElement = document.querySelector<HTMLInputElement>("#" + modalId + "AnonymName");
```
He3lixxx, line 25: "AnonymousName"?

He3lixxx (approved):
Thanks a lot! :)

Issue #1760, opened 2022-05-13
Add notebook


janno42:
Users should have a place to take notes on EvaP. For example students could use this to save text during the semester which they want to use in later evaluations or which they can use as a reminder about what they wanted to write.
For these notes, a plain text area should be provided. It might be extended with formatting and other features in the future, but this is not in scope of this issue.
The entered text should be saved connected to the user's profile on the server and thus be available on any device the user logs into the platform (a corresponding information about anonymity must be shown).

----
A first design study can be found in 4366a2a07dc1. This is not a complete frontend implementation and many details are still open for discussion.

Floating button to open the notebook:
![image](https://user-images.githubusercontent.com/1781719/168350996-8ed8f8bd-96d6-4710-b6c1-5ba9c35e5e35.png)

The notebook itself in a Bootstrap Offcanvas:
![image](https://user-images.githubusercontent.com/1781719/168351498-b183a149-1501-481c-b6cb-6a86c461e652.png)

fidoriel:
I would take it.

karyon:
Previous discussion: #1041, and a duplicate: #1566

janno42:
We could probably remove the backdrop and not auto-close the offcanvas when clicking somewhere else. If you have additional ideas that work with the existing page or do not make things too complicated, let me know!

This is meant to be a low-level approach for unstructured note-taking that shouldn't add too much complexity. As such, it's not really a duplicate of the above-mentioned issues, there we had more in mind and several of the ideas are still out of scope.

fidoriel:
I would throw in my approach to the issue. It is buggy, but I think you will get the pioint: [https://github.com/fidoriel/EvaP/commit/6826ea735e7f842dd7a65528bd8f5c9428fa458b](url)

PR #1771, opened 2022-06-14
A notebook for EvaP

fidoriel:
This is my approach for the notebook derived from @janno42 design study.
See issue #1760 for reference.


fidoriel:
This still a draft. May @janno42 have a look at the front end implementation to greenlight it?

janno42:
Very nice! There are some (small) things that I want to address:
- The opening animation should be smoother - we probably have to overwrite the Bootstrap animation.
- The input field should not be resizable because it already fills the available space.
- The heading and close button should have a smaller left and right padding so they align with the input field. The top and bottom padding can probably also be reduced.
- The button to open the notebook should be placed on the top left, with the same distance to the top bar as the offcanvas.

evap/evaluation/templates/django/forms/widgets/textarea.html
```diff
@@ -1,2 +1,9 @@
-<textarea class="form-control" name="{{ widget.name }}"{% include 'django/forms/widgets/attrs.html' %}>
+<textarea name="{{ widget.name }}"
+
+{% for name, value in widget.attrs.items %}
```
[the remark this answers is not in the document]
fidoriel: I think I made a suggestion in that direction

evap/evaluation/tests/test_views.py
```diff
@@ -144,3 +145,48 @@ def test_view_settings_as_non_editor(self):
         self.assertIn("Personal information", page)
         self.assertNotIn("Delegates", page)
         self.assertIn(user.email, page)
+
+
+class TestNotebookView(TestCase):
+    url = "/notebook"
```
He3lixxx, line 151: Please use django's `reverse` method to get the url.

evap/evaluation/views.py
```diff
@@ -1,9 +1,9 @@
 import logging
 from datetime import date, timedelta
+from pdb import post_mortem
```
He3lixxx, line 3: leftover import

evap/evaluation/views.py
```diff
@@ -226,3 +224,12 @@ def profile_edit(request):
...
     )
+
+@require_POST
+def notebook(request):
+    if request.method == "POST":
+        form = NotebookForm(request.POST, instance=request.user)
+        if form.is_valid():
+            form.save()
+            return HttpResponse(status=204)
+        return HttpResponseBadRequest()
```
He3lixxx, line 235: missing newline

evap/results/tests/test_views.py
```diff
@@ -459,6 +459,7 @@ def test_heading_question_filtering(self):
         self.assertIn(likert_question.text, page)
         self.assertNotIn(heading_question_2.text, page)
 
+    @patch("django.template.context_processors.get_token", Mock(return_value="predicabletoken"))
```
He3lixxx, line 462:
```suggestion
    @patch("django.template.context_processors.get_token", return_value="predicabletoken")
```

evap/evaluation/views.py
```diff
@@ -159,7 +159,6 @@ def legal_notice(request):
 
 
 @require_POST
-@login_required
```
He3lixxx, line 162: For a nice git history, we probably want to have one, well documented commit that removes the four usages (you have that) as well as the implementation (think it's still in the code), and links the PR where it should have been removed in the past.

evap/static/scss/components/_buttons.scss
```diff
@@ -265,4 +275,4 @@ a.collapse-toggle {
     &.collapsed {
         box-shadow: none;
     }
-}
+}
```
He3lixxx, line 278: missing newline -- probably an editor setting.

evap/static/scss/evap.scss
```diff
@@ -146,3 +146,48 @@ td.icon {
...
+    .notebook-container {
+        width: 25vw;
+        padding-left: $spacer;
+        padding-bottom: $spacer;
+        padding-top: $spacer;
+    }
+
+    .notebook-card {
+        // substracting 3 * Margins 2x card 1x breadcrumb
+        height: calc((100vh - #{$navbar-brand-height}) - (#{$footer-height} + ((#{$spacer} * 1.5) * 3))) !important;
```
He3lixxx, line 169: 1. Why is the spacer multiplied by 1.5?
2. Why do we need the `!important`? (=> what other rule would override the height without `!important`?)

fidoriel: this is multiplied by 1.5 to match the margins. Have a look at the bootsrap docs. I am going to search for a more elegant solution.

evap/static/ts/src/csrf-utils.ts
```diff
@@ -1,35 +0,0 @@
-// based on: https://docs.djangoproject.com/en/3.1/ref/csrf/#ajax
```
niklasmohrin, line 1: Why did you delete this file / move it into `utils`? The `csrftoken` etc. could very well stay here

package.json
```diff
@@ -34,4 +34,4 @@
         "globalTeardown": "jest-environment-puppeteer/teardown",
         "testEnvironment": "jest-environment-puppeteer"
     }
-}
+}
```
niklasmohrin, line 37: Please revert this change

evap/static/ts/tests/unit/utils.ts
```diff
@@ -2,7 +2,7 @@
  * @jest-environment jsdom
  */
 
-Object.defineProperty(document, "cookie", {
+ Object.defineProperty(document, "cookie", {
```
niklasmohrin, line 5: Please revert this line

evap/evaluation/templatetags/evaluation_filters.py
```diff
@@ -221,3 +221,10 @@ def order_by(iterable, attribute):
 @register.filter
 def get(dictionary, key):
     return dictionary.get(key)
+
+
+@register.filter(name="add_class")
```
niklasmohrin, line 226: Since the name is the same as the function, you can use:
```suggestion
@register.filter
```

evap/evaluation/templates/base.html
```diff
@@ -185,6 +202,40 @@
...
+                setCookie("evap_notebook_open", "true");
+            });
+
+            $('#notebook').on('hidden.bs.collapse', function () {
+                onHideNotebook()
+                setCookie("evap_notebook_open", "false");
+            });
+
+            function onShowNotebook()
+            {
```
niklasmohrin, line 227: This is not our codestyle, please format like this:
```suggestion
            function onShowNotebook() {
```

evap/evaluation/templates/base.html
```diff
@@ -185,6 +202,40 @@
...
+
+            import { getCookie, setCookie} from "{% static 'js/utils.js' %}";
+            if (getCookie('evap_notebook_open') == "true") {
+                new bootstrap.Collapse(document.querySelector('#notebook'));
+                new bootstrap.Collapse(document.querySelector('#notebookButton'));  
+                onShowNotebook();
+            };
+
+            // make evap go away for notebook
+            $('#notebook').on('show.bs.collapse', function () {
```
niklasmohrin, line 216: Although there is still quite some amount of jquery code in the evap, please avoid adding more to it :)

evap/evaluation/templates/base.html
```diff
@@ -185,6 +202,40 @@
             };
         </script>
 
+        <script type="module">
+            // Notebook logic
```
niklasmohrin, line 206: Any chance this could be a file `notebook.ts` instead? I haven't looked at the implementation details too close, but for other UI elements we have adopted a pattern of having a class that contains all the logic (for example the contact modal)

niklasmohrin (changes requested):

niklasmohrin (changes requested):
This branch currently has merge conflicts, please resolve those first :)

fidoriel:
I do not know why the tests are not working. this is a bug, it worked before the merge. i will investigate.

fidoriel:
https://github.com/testing-library/jest-dom/issues/457

He3lixxx:
Firefox says:
> Cookie “evap_notebook_open” does not have a proper “SameSite” attribute value. Soon, cookies without the “SameSite” attribute or with an invalid value will be treated as “Lax”. This means that the cookie will no longer be sent in third-party contexts. If your application depends on this cookie being available in such contexts, please add the “SameSite=None“ attribute to it. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite

janno42 (commented):
I didn't have a look at the code yet, but I noticed some things in the UI:
- The notebook doesn't consistently stay open or closed when changing the page.
- The heading line ("Notebook" and close icon) should have the same x-padding as the textarea and can have a smaller top-padding.
- When clicking the save button, there should be visual feedback that the content was saved. You could disable the button for a few seconds and change the text to "Saved".
- On narrow pages, the notebook can't always be placed correctly because there is not enough space. Maybe it should stay at the top for one more breakpoint?
- The changing vertical position on pages with and without breadcrumbs and the space below the notebook for the footer are visually unpleasant. I can't think of a perfect solution right now, maybe always vertically centering the notebook would help.
- There doesn't seem to be a transition when opening or closing.

fidoriel:
we need to update jest because of [https://github.com/swc-project/jest/issues/96](url)

fidoriel:
> I didn't have a look at the code yet, but I noticed some things in the UI:
> 
>     * The notebook doesn't consistently stay open or closed when changing the page.

adressed

> 
>     * The heading line ("Notebook" and close icon) should have the same x-padding as the textarea and can have a smaller top-padding.

I think the margin is OK now, bot I dont know how to fix the top padding; needs investigation on my site.

> 
>     * When clicking the save button, there should be visual feedback that the content was saved. You could disable the button for a few seconds and change the text to "Saved".
How should I translate the Saved in

done

>     * On narrow pages, the notebook can't always be placed correctly because there is not enough space. Maybe it should stay at the top for one more breakpoint?

Yes, but I discussed that with @He3lixxx; At this Page size and Zoomfactor EvaP is not usable with or without Notebook.

> 
>     * The changing vertical position on pages with and without breadcrumbs and the space below the notebook for the footer are visually unpleasant. I can't think of a perfect solution right now, maybe always vertically centering the notebook would help.

If done so, it would overlap with the footer.

> 
>     * There doesn't seem to be a transition when opening or closing.

Yes, initially there was one, but there were issues with the Bootstrap transition. It was not smooth.

fidoriel:
We *need* the resolver; after adding it, the TS tests pass.

fidoriel:
Codacy is failing on lines which were not modified in this PR

He3lixxx:
> > * On narrow pages, the notebook can't always be placed correctly because there is not enough space. Maybe it should stay at the top for one more breakpoint?
> 
> Yes, but I discussed that with @He3lixxx; At this Page size and Zoomfactor EvaP is not usable with or without Notebook.

I don't remember this conversation. Optimally, the existence of the notebook would not make it harder to make pages mobile-friendly in the future. I generally disagree with "It's broken already, so let's add more broken code" reasoning. There might have been other factors that I don't remember though.

> We _need_ the resolver; after adding it, the TS tests pass.

@FranzSw said that if we just remove the `.js` ending from the import, it also works fine. I'd much rather import files without a file extension than adding a new npm dependency. A quick check on the internet indicates that this is also standard-compliant (i.e. we can import from filenames without an extension, the runtime adds the required extension). Is there a reason not to do that?

fidoriel:
@FranzSw what is wrong here. You said removing the .js works fine

FranzSw:
> @FranzSw what is wrong here. You said removing the .js works fine

I said that the tests ran fine without the js, after I tried that on my machine 🙃

fidoriel:
> > @FranzSw what is wrong here. You said removing the .js works fine
> 
> I said that the tests ran fine without the js, after I tried that on my machine 🙃

Have a look here, just removed the js endings and js is broken on the web
https://github.com/fidoriel/EvaP/tree/notebook-broken

evap/static/scss/_variables.scss
```diff
@@ -81,8 +81,11 @@ $link-decoration: none;
...
 
 $small-font-size: 0.8rem;
+$footer-height: 46px;
 
 $breadcrumb-divider-color: $medium-gray;
 
 $input-btn-focus-width: .25rem;
 $input-btn-focus-color-opacity: .25;
+
+$notebook-break: calc(3/4);
```
He3lixxx, line 91: missing newline. What does this value mean/do?

evap/static/scss/components/_buttons.scss
```diff
@@ -151,6 +151,16 @@ a:not([href]):not(.disabled) {
     min-width: 140px;
 }
 
+.button-notebook {
+    @include media-breakpoint-up(md) {
+        position: fixed;
+        bottom: 60px;
```
He3lixxx, line 157: Should this be `$footer-height + some_constant`?

evap/static/scss/evap.scss
```diff
@@ -147,3 +147,67 @@ td.icon {
...
+    .notebook-container {
+        width: 25vw;
+        padding-left: $spacer;
+        padding-bottom: $spacer;
+        padding-top: $spacer;
+    }
+
+    .notebook-card {
+        // substracting 3 * Margins 2x card 1x breadcrumb
+        height: calc((100vh - #{$navbar-brand-height}) - (#{$footer-height} + ((#{$spacer} * 1.5) * 3)));
```
He3lixxx, line 183: I either don't understand the comment or the computation -- there is no `2 * ` in there?

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,74 @@
+/* eslint-disable  @typescript-eslint/no-non-null-assertion */
```
He3lixxx, line 1: This is just to satisfy codacy, right?

@niklasmohrin I think we don't want this in the codebase then, do we?

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,74 @@
...
+    "click",
+    function (event: MouseEvent): void {
+        event.preventDefault();
+        const form = assertDefinedUnwrap(document.getElementById("notebook-form")) as HTMLFormElement;
+        const data = new FormData(form);
+
+        const target = assertDefinedUnwrap(event.target) as HTMLButtonElement;
+        const default_label = assertDefinedUnwrap(target.getAttribute("value"));
+        const sending_label = assertDefinedUnwrap(target.getAttribute("data-label-sending"));
+        const error_label = assertDefinedUnwrap(target.getAttribute("value"));
```
He3lixxx, line 24: Something is wrong here

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,74 @@
...
+        target.disabled = true;
+        target.setAttribute("value", sending_label);
+
+        const form_json = JSON.stringify(Object.fromEntries(data.entries()));
+
+        const cooldown_time = 2000;
+        const success_code = 204;
```
He3lixxx, line 31: I think you can inline these.

For `success_code`, you can just test for `response.ok` (https://developer.mozilla.org/en-US/docs/Web/API/Response/ok)

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,74 @@
...
+            method: "POST",
+        }).then(response => {
+            if (response.status == success_code) {
+                target.setAttribute("value", assertDefinedUnwrap(target.getAttribute("data-label-cooldown")));
+                setTimeout(function (): void {
+                    target.disabled = false;
+                    target.setAttribute("value", default_label);
+                }, cooldown_time);
+            } else {
+                alert(error_label);
```
He3lixxx, line 45: I think we'd want to reset the button so users can try again?

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,74 @@
...
+            }
+        });
+    },
+);
+
+// https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1535
+assertDefinedUnwrap(document.getElementById("notebook")).addEventListener("show.bs.collapse", function (): void {
+    setCookie(NOTEBOOK_COOKIE_NAME, "true");
+    onShowNotebook();
```
He3lixxx, line 54: Why is the `setCookie`-part not a part of `onShowNotebook`?

evap/static/ts/src/utils.ts
```diff
@@ -13,10 +13,36 @@ export function assertDefined<T>(val: T): asserts val is NonNullable<T> {
...
+    return null;
+}
+
+export function setCookie(key: string, value: string): void {
+    document.cookie = key + "=" + value + ";path=/;SameSite=strict";
+}
+
+export const testable = {
+    getCookie,
+};
```
He3lixxx, line 45: @niklasmohrin what is the purpose of this indirection? Do we need it?

evap/static/ts/tests/unit/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+Object.defineProperty(document, "cookie", {
+    get: () =>
+        `foo=${encodeURIComponent("F00")}; ` +
+        `csrftoken=${encodeURIComponent("token")}; ` +
+        `bar=${encodeURIComponent("+{)")}; ` +
+        `baz=${encodeURIComponent("+{`")}`,
+});
+
+// @ts-ignore
+window.$ = require("../../../js/jquery-2.1.3.min");
```
He3lixxx, line 14: What part here needs jquery?

He3lixxx (changes requested):

janno42 (changes requested):
The notebook doesn't work for me. After saving (which—according to the button—was successful) and reloading the page, the notes aren't saved.

evap/static/ts/src/notebook.ts
niklasmohrin: yeet it

evap/static/ts/src/utils.ts
niklasmohrin: we could probably get rid of the `testable` object everywhere, don't really know why it was introduced, `export function` should work just fine. however, I think that it's best to just stick to the existing convention for this PR

evap/evaluation/templates/notebook.html
```diff
@@ -0,0 +1,21 @@
+<div class="collapse position-fixed collapse-horizontal notebook-animation mt-0" tabindex="-1" id="notebook" aria-expanded="false" aria-labelledby="notebook">
+    <div class="card">
+        <div class="card-body notebook-card notebook-div">
+            <div class="offcanvas-header notebook-button mb-2">
+                <h5 class="card-title m-0">
+                    {% trans 'Notebook' %}
+                    <span class="fas fa-circle-info" data-bs-toggle="tooltip" data-bs-placement="right" title="{% blocktrans %}You can put some private notes here that you want to have ready for future evaluations. This data will not be published. But in contrast to answers in evaluations, the content of this field will be saved within your account on the EvaP server and is thus not technically anonymous.{% endblocktrans %}"></span>
```
janno42, line 7:
```suggestion
                    <span class="fas fa-circle-info" data-bs-toggle="tooltip" data-bs-placement="right" title="{% blocktrans %}Here you can store private notes that you want to keep ready for future evaluations. The notes will be stored in plain text in your account on the EvaP server, but will not be shown to anyone but you.{% endblocktrans %}"></span>
```

evap/evaluation/models.py
```diff
@@ -1645,6 +1645,8 @@ class UserProfile(AbstractBaseUser, PermissionsMixin):
 
     is_active = models.BooleanField(default=True, verbose_name=_("active"))
 
+    notes = models.TextField(verbose_name=_("notes"), blank=True, null=False, default="")
```
janno42, line 1648:
```suggestion
    notes = models.TextField(verbose_name=_("notes"), blank=True, default="")
```

evap/evaluation/templates/base.html
```diff
@@ -42,31 +42,43 @@
             {% block breadcrumb_bar %}
             {% endblock %}
         </div>
-
-        <img class="print-brand-image d-none d-print-block" src="{% static 'images/evap.png' %}" alt="{% trans 'Evaluation Platform' %}" />
```
janno42, line 46: this shouldn't be removed

evap/evaluation/tests/test_views.py
```diff
@@ -188,3 +189,20 @@ def test_edit_display_name(self):
         form["first_name_chosen"] = "testdisplayname2"
         form.submit(name="operation", value="illegal", status=400)
         self.assertFalse(UserProfile.objects.filter(first_name_chosen="testdisplayname2").exists())
+
+
+class TestNotebookView(WebTest):
+    url = reverse("evaluation:notebook")
+    page = reverse("evaluation:profile_edit")  # is used, but notebook is accessed from all pages
```
janno42, line 196: do you mean "is used *exemplarily*"?

janno42 (changes requested):
The notebook automatically opens on some pages (e.g., vote, contributor index, rewards) and is closed on others (e.g., student index, results) which doesn't seem to be intended.

It should not be open by default. We could think about one exemption: Opening it on the vote page if there is content saved, but that is up to you.

He3lixxx (changes requested):
What @janno42 described in https://github.com/e-valuation/EvaP/pull/1771#pullrequestreview-1497140741 is still broken: The code currently intents to re-open the notebook whenever it was open before. I discussed this with @janno42, the intended behavior here is fine. But, it currently doesn't work on all pages -- if you go to http://localhost:8000/grades/semester/19, the notebook will always open, regardless of previous state.

fidoriel:
now the notebook cookie now migrated to localstorage

evap/evaluation/templates/notebook.html
```diff
@@ -0,0 +1,21 @@
+<div class="collapse position-fixed collapse-horizontal notebook-animation mt-0" tabindex="-1" id="notebook" aria-expanded="false" aria-labelledby="notebook">
```
janno42, line 1:
```suggestion
<div class="collapse position-fixed collapse-horizontal notebook-animation mt-0 d-print-none" tabindex="-1" id="notebook" aria-expanded="false" aria-labelledby="notebook">
```

janno42 (approved):
after the one proposed change:
✔️ Meets requirements
✔️ UI functionality checked

evap/evaluation/templates/notebook.html
```diff
@@ -0,0 +1,21 @@
...
+    <div class="card">
+        <div class="card-body notebook-card notebook-div">
+            <div class="offcanvas-header notebook-button mb-2">
+                <h5 class="card-title m-0">
+                    {% trans 'Notebook' %}
+                    <span class="fas fa-circle-info" data-bs-toggle="tooltip" data-bs-placement="right" title="{% blocktrans %}Here you can store private notes that you want to keep ready for future evaluations. The notes will be stored in plain text in your account on the EvaP server, but will not be shown to anyone but you.{% endblocktrans %}"></span>
```
janno42, line 7: please move the icon from the heading to the bottom of the notebook. it better fits on the left side of the button (with a `pe-2`)

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,72 @@
...
+        const data = new FormData(form);
+
+        const target = assertDefinedUnwrap(event.target) as HTMLButtonElement;
+        const default_label = assertDefinedUnwrap(target.getAttribute("value"));
+        const sending_label = assertDefinedUnwrap(target.getAttribute("data-label-sending"));
+        const error_label = assertDefinedUnwrap(target.getAttribute("data-label-error"));
+        const cooldown_label = assertDefinedUnwrap(target.getAttribute("data-label-cooldown"));
+        target.disabled = true;
+        target.setAttribute("value", sending_label);
+        const cooldown_time = 2000;
```
He3lixxx, line 27: I don't really think the variable here adds much value:
* It's not clear what cooldown it is about
* It's only used once.
* When reading the setTimeout call, users have to first find the definition of the (constant) variable to interpret what is going on.

-> I'd just inline the 2000 into the `setTimeout` call. What do you think?

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,72 @@
...
+            } else {
+                target.setAttribute("value", default_label);
+                target.disabled = false;
+                alert(error_label);
+            }
+        });
+    },
+);
+
+// https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1535
```
He3lixxx, line 49: I don't get the comment here -- what's this supposed to tell me?

evap/static/ts/src/quick-review-slider.ts
```diff
@@ -20,6 +19,8 @@ import {
     selectOrError,
 } from "./utils.js";
 
+import { CSRF_HEADERS } from "./csrf-utils.js";
+
```
He3lixxx, line 23: Unnecessarily moved line?

evap/static/ts/src/csrf-utils.ts
```diff
@@ -1,16 +1,5 @@
...
```
He3lixxx, line 1: Since this is not based on cookies anymore, can you revert all changes that were related to making `getCookie` available for your code? `csrf-utils.ts` and `utils.ts` should have only minimal changes, then.

He3lixxx (changes requested):

evap/evaluation/templates/base.html
```diff
@@ -42,30 +42,44 @@
             {% block breadcrumb_bar %}
             {% endblock %}
         </div>
-
+
```
He3lixxx, line 45: You resolved the comment but didn't change the code here: You added spaces here, which match the indentation level, but are now also trailing spaces on this line.

For consistency within the file, please remove these spaces again.

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,70 @@
+declare const bootstrap: typeof import("bootstrap");
+import { assertDefinedUnwrap } from "./utils.js";
+import { CSRF_HEADERS } from "./csrf-utils.js";
+
+const NOTEBOOK_LOCALSTORAGE_KEY = "evap_notebook_open";
+
+if (localStorage.getItem(NOTEBOOK_LOCALSTORAGE_KEY) == "true") {
+    new bootstrap.Collapse(assertDefinedUnwrap(document.getElementById("notebook")));
+    new bootstrap.Collapse(assertDefinedUnwrap(document.getElementById("notebookButton")));
```
He3lixxx, line 9: What do these two lines do? I'm asking because I feel they enable some kind of functionality, and they should thus be outside the `if`-block.

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,70 @@
...
+                setTimeout(function (): void {
+                    target.setAttribute("value", default_label);
+                    target.disabled = false;
+                }, 2000);
+            } else {
+                target.setAttribute("value", default_label);
+                target.disabled = false;
+                alert(error_label);
+            }
+        });
```
He3lixxx, line 44: Something we should have seen earlier, and I just now noticed because I used `fetch()` in some other code: This doesn't actually handle a "server is not reachable" case, see https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#checking_that_the_fetch_was_successful.

If the server is not reachable, the promise will reject, so we have to `.catch()` that.

I'd propose so build it like this:
```js
fetch(form.action, {
    // fetch arguments
}).then(response => {
    assert(response.ok);
    // success handling here
}).catch(error => {
    // error handling here
};
```

@niklasmohrin agree / better ideas?

He3lixxx (changes requested):

evap/static/ts/src/notebook.ts
He3lixxx: I think you said you wanted to post a link to some kind of explanation here...

evap/evaluation/templates/base.html
```diff
@@ -81,6 +95,11 @@
             };
             activateTooltips();
 
+
+            for(const offCanvasElement of document.querySelectorAll('.offcanvas')) {
+                new bootstrap.Offcanvas(offcanvasEl);
+            }
```
He3lixxx, line 101: same discussion as with the Collapsible -- just from reading the [bootstrap documentation](https://getbootstrap.com/docs/5.0/components/offcanvas/), this shouldn't be necessary, bootstrap should already do its thing with anything that has `class="offcanvas"`.

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,70 @@
...
+            method: "POST",
+        })
+            .then(response => {
+                assert(response.ok);
+                target.setAttribute("value", cooldown_label);
+                setTimeout(function (): void {
+                    target.setAttribute("value", default_label);
+                    target.disabled = false;
+                }, 2000);
+            })
```
He3lixxx, line 39: @niklasmohrin I find this formatting a bit weird, but it is what prettier currently produces. Is this what we want?

niklasmohrin: Avoiding the discussion about formatting, this should not use `.then` and `.catch` but instead use an async function and `await` the request. It's a bit sus that we are using `fetch` at all, but I suppose it is because we want to reset the UI once the request went through - this "I want my form to tell me when its done" logic could be extracted into its own module, we can also keep it like this for this PR though.

niklasmohrin (changes requested):

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,70 @@
...
+
+const NOTEBOOK_LOCALSTORAGE_KEY = "evap_notebook_open";
+
+if (localStorage.getItem(NOTEBOOK_LOCALSTORAGE_KEY) == "true") {
+    assertDefinedUnwrap(document.getElementById("notebook")).classList.add("show");
+    onShowNotebook();
+}
+
+assertDefinedUnwrap(document.getElementById("notebook-save-button")).addEventListener(
+    "click",
```
niklasmohrin, line 13: This should probably be a submit handler on the form instead

evap/evaluation/templates/base.html
fidoriel: fixed

evap/evaluation/templates/notebook.html
```diff
@@ -0,0 +1,23 @@
+<div class="collapse position-fixed collapse-horizontal notebook-animation mt-0 d-print-none" tabindex="-1" id="notebook" aria-expanded="false" aria-labelledby="notebook">
+    <div class="card">
+        <div class="card-body notebook-card notebook-div">
+            <div class="offcanvas-header notebook-button mb-2">
+                <h5 class="card-title m-0">
+                    {% trans 'Notebook' %}
+                </h5>
+                <button type="button" class="btn-close text-reset" type="button" data-bs-toggle="collapse" data-bs-target="#notebook" aria-controls="collapseNotebook" id="notebookButtonClose"></button>
```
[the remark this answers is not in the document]
fidoriel: fixed

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    public attach = (): void => {
+        unwrap(this.notebook).addEventListener("submit", this.onSubmit);
+    };
+}
+
+export class NotebookLogic {
+    private readonly notebook: HTMLElement;
+    private readonly content: HTMLElement;
+    private form: NotebookFormLogic;
```
niklasmohrin, line 51: I am not sure what these variables are holding exactly, could you give describe them a bit more explicitly?

evap/static/scss/evap.scss
niklasmohrin, line 1: I am not sure what the classes defined here mean, the names `div` and `animation` could be a bit more descriptive. Or you could merge them with the `#notebook` style and use tag selectors instead, which could also clarify the structure in the css already a bit:

```scss
#notebook {
    // we want to cancel bootstrap's animation here
    // (not sure if my comment here is accurate, just a placeholder for your explanation :^) )
    transition: ...;

    form {
        // The form also serves as a container here.
        styles: cool;
    }

    .notebook-container {
        // ..
    }

    // ...    
}
```

I would also move all of the notebook related stuff to `_notebook.scss`. What do you think?

evap/evaluation/templates/base.html
```diff
@@ -182,6 +194,13 @@
             function setSpinnerIcon(id) {
                 $("#" + id).empty().removeClass().addClass('fas fa-spinner fa-spin');
             };
+
+        </script>
+        <script type="module">
+
+            import  {NotebookLogic} from "{% static 'js/notebook.js' %}"
```
niklasmohrin, line 201:
```suggestion
            import  { NotebookLogic } from "{% static 'js/notebook.js' %}"
```

evap/static/scss/evap.scss
```diff
@@ -147,3 +147,73 @@ td.icon {
...
+    }
+
+    .notebook-container {
+        width: 100vw;
+        z-index: $zindex-fixed;
+        padding: $spacer;
+    }
+
+    .notebook-card {
+        height: calc(50vh - (#{$spacer} * 1.5));
```
niklasmohrin, line 204: What does this calculation mean?

niklasmohrin (commented):
I think we are getting there :+1: In general, I think we are in the phase of cleaning things up. I think that in general, the scss is a bit hard to understand, because I can't really tie the names of the classes to what is going on on the screen, maybe you can find a way to make it more expressive in that regard (or add comments)?

evap/static/scss/components/_buttons.scss
```diff
@@ -152,6 +152,16 @@ a:not([href]):not(.disabled) {
     min-width: 140px;
 }
 
+.button-notebook {
+    @include media-breakpoint-up(md) {
```
niklasmohrin, line 156: Currently, when in portrait mode, the button is at the end of the site, which is a bit odd to find. Should we put it at the top? @janno42

evap/evaluation/migrations/0140_userprofile_notes.py
niklasmohrin, line 1: note for later: let's not forget to dump the testdata again :)

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,88 @@
...
+            body: new FormData(this.notebook),
+            method: "POST",
+        })
+            .then(response => {
+                assert(response.ok);
+                this.notebook.setAttribute("data-state", "successful");
+                setTimeout(() => {
+                    this.notebook.setAttribute("data-state", "ready");
+                    submitter.disabled = false;
+                }, 2000);
```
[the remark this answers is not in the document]
fidoriel: @richardebeling said the opposite, because we use it only once

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,88 @@
...
+    private readonly notebook: HTMLElement;
+    private readonly content: HTMLElement;
+    private form: NotebookFormLogic;
+    private readonly localStorageKey: string;
+
+    constructor(notebookId: string) {
+        this.notebook = unwrap(document.getElementById(notebookId));
+        this.form = new NotebookFormLogic(NOTEBOOK_FORM_ID);
+        this.content = unwrap(document.getElementById(WEBSITE_CONTENT_ID));
+        this.localStorageKey = NOTEBOOK_LOCALSTORAGE_KEY;
```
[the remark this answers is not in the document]
fidoriel: why the notebookId? this does not change anything

evap/static/ts/src/notebook.ts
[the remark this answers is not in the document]
richardebeling: (I suggested to inline it when it was a local variable defined 3 lines above. I'd also be fine with a class-constant, but I still stand by the point that this is behavior and not used by anything else, so I think it's fine as an inline number)

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,88 @@
...
+class NotebookFormLogic {
+    private readonly notebook: HTMLFormElement;
+
+    constructor(notebookFormId: string) {
+        this.notebook = unwrap(document.getElementById(notebookFormId)) as HTMLFormElement;
+    }
+
+    private onSubmit = (event: SubmitEvent): void => {
+        event.preventDefault();
+        event.stopPropagation();
```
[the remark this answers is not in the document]
fidoriel: no

evap/static/scss/components/_buttons.scss
fidoriel: if we would fix this somewhere it would overlap with evap content.

janno42: It's probably better to place the icon on the top of the page (above the headline) and not attached to the edge but as a normal button.

evap/evaluation/urls.py
```diff
@@ -12,5 +12,6 @@
     path("contact", views.contact, name="contact"),
     path("key/<int:key>", views.login_key_authentication, name="login_key_authentication"),
     path("profile", views.profile_edit, name="profile_edit"),
+    path("set_notebook", views.notebook, name="notebook_edit"),
```
richardebeling, line 15: three names "set_notebook", "notebook" and "notebook_edit". Can we unify?

evap/evaluation/templates/notebook.html
```diff
@@ -0,0 +1,30 @@
...
+                        {% trans 'Notebook' %}
+                    </h5>
+                    <button type="button" class="btn-close text-reset" data-bs-toggle="collapse" data-bs-target="#notebook" aria-controls="collapseNotebook" id="notebookButtonClose"></button>
+                </div>
+            
+                <form method="POST" action="{% url 'evaluation:notebook_edit' %}" class="notebook-div" id="notebook-form" data-state="ready">
+                    {% csrf_token %}
+                    {{ notebook_form.notes }}
+                    <div class="d-flex justify-content-center p-3">
+                        <button class="btn btn-primary notebook-button" id="notebook-save-button" type="submit" data-errormessage="{% trans 'The server is not responding.' %}">
```
richardebeling, line 16: ID here is unused, please remove

richardebeling, line 16: data-errormessage and `visible-if-error` are duplicated (->data-errormessage should be removed?)

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,89 @@
...
+import { unwrap, assert } from "./utils.js";
+
+const NOTEBOOK_LOCALSTORAGE_KEY = "evap_notebook_open";
+const COLLAPSE_TOGGLE_BUTTON_ID = "notebookButton";
+const WEBSITE_CONTENT_ID = "evapContent";
+const NOTEBOOK_FORM_ID = "notebook-form";
+
+class NotebookFormLogic {
+    private readonly notebook: HTMLFormElement;
+    // add a class constant
```
richardebeling, line 11:
```suggestion
```

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,89 @@
...
+    private readonly notebook_card: HTMLElement;
+    private readonly evap_content: HTMLElement;
+    private form_logic: NotebookFormLogic;
+    private readonly localStorageKey: string;
+
+    constructor(notebookId: string) {
+        this.notebook_card = unwrap(document.getElementById(notebookId));
+        this.form_logic = new NotebookFormLogic(NOTEBOOK_FORM_ID);
+        this.evap_content = unwrap(document.getElementById(WEBSITE_CONTENT_ID));
+        this.localStorageKey = NOTEBOOK_LOCALSTORAGE_KEY + "_" + this.notebook_card.dataset.notebookId;
```
richardebeling, line 59: Inconsistent variable naming scheme - camelCase vs snake_case

(also notebookId)

evap/evaluation/templates/notebook.html
fidoriel: no, this is the message for the alert. discussed that with @niklasmohrin

evap/evaluation/templates/notebook.html
```diff
@@ -0,0 +1,30 @@
+<div class="notebook">
+   <div class="collapse position-fixed collapse-horizontal notebook-animation mt-0 d-print-none" tabindex="-1" id="notebook" aria-expanded="false" aria-labelledby="notebook">
```
richardebeling, line 2: Why do we have "aria-labelledby" here?

evap/static/ts/src/notebook.ts
niklasmohrin: note: in js / ts we use camel case

evap/evaluation/templates/notebook.html
niklasmohrin: Still a fair point though, maybe we don't need the alert if the error message is displayed in the button already

evap/evaluation/templates/notebook.html
fidoriel: dont know, removed

evap/evaluation/templates/notebook.html
fidoriel: ok

fidoriel: I think a alert is more noticable.

evap/evaluation/urls.py
fidoriel: ok

evap/static/scss/components/_buttons.scss
fidoriel: now the position changes with the notebook break

evap/static/ts/src/notebook.ts
fidoriel: removed

evap/static/ts/src/notebook.ts
fidoriel: renamed to camelcase

richardebeling: `notebook_card` still uses snake_case.

evap/evaluation/templates/notebook.html
niklasmohrin: in that case the point still stands and we probably don't want both. @janno42 ?

evap/evaluation/templates/notebook.html
```diff
@@ -0,0 +1,30 @@
+<div class="notebook">
+   <div class="collapse position-fixed collapse-horizontal notebook-animation mt-0 d-print-none" tabindex="-1" id="notebook" aria-expanded="false">
+        <div class="card">
+            <div class="offcanvas-header notebook-button pt-3 ps-3 pe-3">
+                <h5 class="card-title m-0">
+                    {% trans 'Notebook' %}
+                </h5>
+                <button type="button" class="btn-close text-reset" data-bs-toggle="collapse" data-bs-target="#notebook" aria-controls="collapseNotebook" id="notebookButtonClose"></button>
```
richardebeling, line 8: `document.getElementById("collapseNotebook")` returns `null` for me, both when the notebook is open and when it is closed. What is this intended to do?

evap/evaluation/templates/notebook.html
janno42: We'll keep the alert but the button's text should be changed to the original text.

evap/evaluation/templates/notebook.html
```diff
@@ -0,0 +1,29 @@
+<div class="notebook">
+   <div class="collapse position-fixed collapse-horizontal notebook-animation mt-0 d-print-none" tabindex="-1" id="notebook" aria-expanded="false">
+        <div class="card">
+            <div class="offcanvas-header notebook-button pt-3 ps-3 pe-3">
```
janno42, line 4: why is this `notebook-button`?

evap/static/scss/components/_notebook.scss
```diff
@@ -0,0 +1,88 @@
...
+    .visible-if-sending {
+        display: var(--sending-display);
+    }
+    .visible-if-successful {
+        display: var(--successful-display);
+    }
+
+    .notebook-button {
+        flex: 0 0 auto;
+    }
```
janno42, line 33: do we need this at all?

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

(on commit e36704ecc3)

richardebeling (approved):

evap/evaluation/templates/notebook.html
```diff
@@ -0,0 +1,29 @@
...
+        <div class="card">
+            <div class="offcanvas-header pt-3 ps-3 pe-3">
+                <h5 class="card-title m-0">
+                    {% trans 'Notebook' %}
+                </h5>
+                <button type="button" class="btn-close text-reset" data-bs-toggle="collapse" data-bs-target="#notebook" id="notebookButtonClose"></button>
+            </div>
+
+            <form method="POST" action="{% url 'evaluation:notebook_edit' %}" id="notebook-form" data-state="ready">
+                <div class="card-body notebook-card d-flex flex-column" data-state="ready">
```
niklasmohrin, line 12: Why do both of these have a `data-state`?

niklasmohrin (changes requested):
(Final?) comments

evap/static/ts/src/notebook.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    public attach = (): void => {
+        this.notebook.addEventListener("submit", this.onSubmit);
+    };
+}
+
+export class NotebookLogic {
+    private readonly notebookCard: HTMLElement;
+    private readonly evapContent: HTMLElement;
+    private formLogic: NotebookFormLogic;
```
niklasmohrin, line 50:
```suggestion
    private readonly formLogic: NotebookFormLogic;
```

niklasmohrin (approved):

janno42:
Remember, remember, the 10th of ... October. 🎆 

Thank you! :)

PR #1773, opened 2022-06-24
Send review reminders only when needed

janno42:
before, reminders were always sent. they had an empty list if no evaluations required urgent review.

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -1992,15 +1992,7 @@ class EmailTemplate(models.Model):
             template.send_to_user(participant, {}, body_params, use_cc=True)
 
     @classmethod
-    def send_textanswer_reminder(cls):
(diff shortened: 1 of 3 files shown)


evap/evaluation/models.py
```diff
@@ -1992,15 +1992,7 @@ def send_participant_publish_notifications(cls, evaluations, template=None):
             template.send_to_user(participant, {}, body_params, use_cc=True)
 
     @classmethod
-    def send_textanswer_reminder(cls):
+    def send_textanswer_reminder_to_user(cls, user, evaluations):
```
He3lixxx, line 1995:
```suggestion
    def send_textanswer_reminder_to_user(cls, user: UserProfile, evaluations: List[Evaluation]):
        assert len(evaluations) != 0
```

(may require additional imports)

He3lixxx (approved):

PR #1784, opened 2022-07-19
Fix wrong id_for_label usage in rendered evaluation edit form

### LINKED ISSUE #1769: Evaluation Edit: Button Groups don't work, submitting fails with "This field is required".
Since the django4 update (#1746, https://github.com/e-valuation/EvaP/commit/f2ce2c5a686070b153549559ac614c6aad0c8e73), going to any evaluation edit page as a staff user and clicking on "Save" fails with
>  This field is required

on the extra form in the contribution formset. The same is probably true for any formset on our page.

At the same time, the button groups on the page can not be used properly.

I don't remember (and currently also cannot find) any explanation of changed behavior regarding formsets on the django4 [release notes](https://docs.djangoproject.com/en/4.0/releases/4.0/). [This ticket](https://code.djangoproject.com/ticket/33534) also implies django's behavior shouldn't have changed. Maybe we do something to the form so that `has_changed()` evaluates to true.

@niklasmohrin is currently working on a minrep example (for maybe creating a django issue).

Edit: Explanation of what's going on is below in https://github.com/e-valuation/EvaP/issues/1769#issuecomment-1146813897.
The root cause is that django4 removed `id_for_label` from choices of a choice field, but we attempt to use it in `choice_buttom.html`. Now, we always put "None" there in the html, breaking both button groups and sortable.js.

He3lixxx:
Fixes #1769.

Our old code was broken and just happened to work accidentally before django 4: The default widget for a choice field is a [`Select` widget](https://docs.djangoproject.com/en/4.0/ref/forms/widgets/#django.forms.Select). With that, the rendering:
```html
<select id="test" name="test">
   <option>1</option>
   <option>2</option>
</select>
```
wouldn't require any `id_for_label` to be set for the individual values. However, before django 4, this happened to be set, and our code relied on that with no documented guarantee.

This PR changes the button groups to use `RadioSelect` widgets, which would be rendered as:
```html
<input type="radio" id="id_1" name="test" value="1"><label for="id_1">1</label>
<input type="radio" id="id_2" name="test" value="2"><label for="id_2">2</label>
```
i.e. here, every choice gets a seperate ID. This is also guaranteed [by the docs](https://docs.djangoproject.com/en/4.0/ref/forms/widgets/#radioselect):
> To get more granular, you can use each radio button’s tag, choice_label and id_for_label attributes

So this time, if anything changes, we should see an entry in the release notes.

Edit: We can probably make a front end test for this one, it should be detectable just from the rendered html

### DIFF
--- a/evap/static/ts/tests/frontend/staff-evaluation-edit.ts
+++ b/evap/static/ts/tests/frontend/staff-evaluation-edit.ts
@@ -9,7 +9,7 @@ test("changes form data", pageHandler(
     async page => {
         const managerId = await page.evaluate(() => {
             const tomselect = (document.getElementById("id_contributions-0-contributor") as any).tomselect;
-            const options = tomselect!.options;
+            const options = tomselect.options;
             const managerOption = Object.keys(options).find(key => options[key].text == "manager (manager)");
             tomselect.setValue(managerOption);
             return managerOption;
--- a/evap/static/ts/tests/frontend/staff-evaluation-edit.ts
+++ b/evap/static/ts/tests/frontend/staff-evaluation-edit.ts
@@ -1,0 +1,41 @@
+import { test, expect } from "@jest/globals";
+
+import { pageHandler } from "../utils/page";
+
+interface TomSelectedHtmlSelectElement extends HTMLSelectElement {
+    tomselect: any;
+}
+
+test("copies header", pageHandler(
+    "/staff/semester/PK/evaluation/PK/edit/normal.html",
+    async page => {
+        const managerId = await page.evaluate(() => {
+            const tomselect = (document.getElementById("id_contributions-0-contributor") as TomSelectedHtmlSelectElement).tomselect;
+            const options = tomselect!.options;
+            const managerOption = Object.keys(options).find(key => options[key].text == "manager (manager)");
+            tomselect.setValue(managerOption);
+            return managerOption;
+        });
+        console.log("managerID: " + managerId);
+
+        const editorLabels = await page.$x("//label[contains(text(), 'Editor')]");
+        const ownAndGeneralLabels = await page.$x("//label[contains(text(), 'Own and general')]");
+        if (editorLabels.length < 1 || ownAndGeneralLabels.length < 1) {
+          throw new Error("Button group buttons not found.");
+        }
+
+        await (editorLabels[0] as any).click();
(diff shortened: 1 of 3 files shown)


janno42:
Do you want to add a test? (:

janno42 (approved):

niklasmohrin (changes requested):
This seems all correct, blocking until test though (as discussed in person)

niklasmohrin:
Take this:
```typescript
const formData = await page.evaluate(() => {
    return Object.fromEntries(new FormData(document.getElementById("evaluation-form") as HTMLFormElement));
});
```

He3lixxx:
> Take this:
> 
> ```ts
> const formData = await page.evaluate(() => {
>     return Object.fromEntries(new FormData(document.getElementById("evaluation-form") as HTMLFormElement));
> });
> ```

Nice, thanks. Added that, seems to work now.

I've removed the `TomSelectedHtmlSelectElement` interface, because I figured it doesn't really help if we add "This has a .tomselect attribute" to the type system, but then go on to do stuff inside there (-> `.tomselect.options`, access its keys, ...) while only having it declared as `any`. Or, to say it differently: That only hid the `any` behind another interface, but didn't make the code type-safe. I think if we want proper typing, we should just use the proper tomselect types.

---

Codacy complains about magic numbers (in a test, I think they are fine: We _do_ want to click the first button that says "xyz") and `any` / non-null-assertions (Also fine for me: Its a test, I'm fine if it breaks at runtime, I don't want to write additional boilerplate code just to make that explicit.)

niklasmohrin (approved):

evap/static/ts/tests/frontend/staff-evaluation-edit.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import { pageHandler } from "../utils/page";
+
+// regression test for #1769
+test("changes form data", pageHandler(
+    "/staff/semester/PK/evaluation/PK/edit/normal.html",
+    async page => {
+        const managerId = await page.evaluate(() => {
+            const tomselect = (document.getElementById("id_contributions-0-contributor") as any).tomselect;
+            const options = tomselect!.options;
```
niklasmohrin, line 12:
```suggestion
            const options = tomselect.options;
```

Issue #1739, opened 2022-04-21
Merging CourseTypes does not update results (template) cache


He3lixxx:
Staff's `course_type_merge` view updates neither cached results nor the results template cache entry. At least the latter is a bug, as the results index view will then still display the cached row with the outdated course type. (This leads to a javascript error, making the whole datagrid unusable.)

A similar thing might happen for other columns on the results page datagrid as well?

After #1738, we might get away with less caching and/or much more aggressive cache invalidation for the results.

He3lixxx:
Thinking about other data that's not directly a "result" where we might have to invalidate the cache:
* [ ] When changing a `UserProfile`, e.g. the name, we would have to invalidate all cached results where this user is contributor.

These changes are not possible for evaluations where results are cached afaik:
* Changing a questionnaire.
* Changing a contribution (most relevant: its label)

These are used in `results_index_course.html` (-> template cache update required):
* [x] String representation of each degree of the course (updated in DegreeForm)
* [ ] String representation of `course.type` -- this issue. Updated in `CourseTypeForm`, and in `course_edit`, but not on CourseType merge.
* [x] Course name (updated in `course_edit`)
* [x] Short name of the semester the course is in (updated in `SemesterForm`, afaik a course's semester can't be changed)
* [ ] Names of each responsible of the course -- Updated when adding or removing in `course_edit`, but not on UserProfile change.

These are used in `results_index_evaluation.html`:
* [x] Course name (updated in `course_edit` and in `EvaluationForm` when changing the course)
* [ ] Evaluation name and full name
* [ ] Evaluation State -- `Evaluation.save()` will do stuff here, but only if changing from/to a state with caching, and not when saving changes within those states?
* [x] Evaluation's `is_single_result` (can't afaik change?)
* [x] Semester short name (see above)
* [ ] Responsible names

We already have way too much code trying to properly invalidate this -- and I'm pretty certain it would break again in the future. It would be great if we could get updated `test_data.json` from production (@janno42) and check out our current performance when filling the caches there.

On our current test data, `refresh_results_cache` takes around ~4.5s for me, approx ~1s of which is for updating the results cache, and another 1s is for filling the template cache. If we manage to get numbers for a bigger dataset somewhat close to this, I'd argue we should drop the results cache and just recompute the template cache in fixed intervals (once per hour?). It's still a bit tricky, as we would want to keep the site online during that, but a lot less tricky than trying to keep these caches valid.

He3lixxx:
[...]

PR #1787, opened 2022-08-01
Allow running refresh_results_cache during normal operation, with each cache update happening atomically

He3lixxx:
Step 1 to solving #1739. When adding a cronjob that regularly runs ./manage.py refresh_results_cache, this should make invalid cache entries fix themselves the next time the cronjob runs.

(Note: The results page will still be broken during the period after merging courses but before the next cronjob run. We probably want to add manual triggers for such cases.)

### DIFF
--- a/evap/evaluation/management/commands/refresh_results_cache.py
+++ b/evap/evaluation/management/commands/refresh_results_cache.py
@@ -1,4 +1,3 @@
-from django.core.cache import caches
 from django.core.management.base import BaseCommand
 from django.core.serializers.base import ProgressBar
 
@@ -18,9 +17,6 @@ class Command(BaseCommand):
     requires_migrations_checks = True
 
     def handle(self, *args, **options):
-        self.stdout.write("Clearing results cache...")
-        caches["results"].clear()
-
         self.stdout.write("Calculating results for all evaluations...")
 
         self.stdout.ending = None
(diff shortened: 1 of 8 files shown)


janno42 (approved):

evap/evaluation/management/commands/refresh_results_cache.py
```diff
@@ -18,9 +17,6 @@ class Command(BaseCommand):
     requires_migrations_checks = True
 
     def handle(self, *args, **options):
-        self.stdout.write("Clearing results cache...")
-        caches["results"].clear()
```
niklasmohrin, line 22: I don't understand why we wouldn't clear it anymore (even if we are overwriting all keys anyways)

niklasmohrin: Ah, before the caching was lazy and we only let it update by requesting the template. Should we rename `warm_up_template_cache` to `update_template_cache` then?

niklasmohrin (approved):

evap/results/views.py
```diff
@@ -96,19 +97,11 @@ def warm_up_template_cache(evaluations):
...
-
-
 def update_template_cache_of_published_evaluations_in_course(course):
     course_evaluations = course.evaluations.filter(state__in=STATES_WITH_RESULT_TEMPLATE_CACHING)
-    for course_evaluation in course_evaluations:
-        _delete_evaluation_template_cache_impl(course_evaluation)
+    update_template_cache(course_evaluations)
+    # Delete template caches for evaluations that no longer need to be cached (e.g. after unpublishing
     _delete_course_template_cache_impl(course)
-    warm_up_template_cache(course_evaluations)
```
niklasmohrin, line 105: the update should be after the delete (and the comment is missing a paran

niklasmohrin (approved):

PR #1790, opened 2022-08-08
Fix #1785: Prevent unwanted double redemption

### LINKED ISSUE #1785: Add waiting time after reward point redemption
After reward points have been redeemed, the next redemption should not be able for a short amount of time to prevent users from accidentally redeeming the selected number of points twice by sending a double post request.

When a redemption is attempted within 30 seconds of the last one, the redemption should not be performed. Instead, an error should be shown that the user should wait a moment before redeeming more points.

Kakadus:
Closes #1785


evap/rewards/tests/test_views.py
```diff
@@ -80,10 +80,20 @@ def test_redeem_points_for_expired_event(self):
         self.assertEqual(5, reward_points_of_user(self.student))
 
     def test_invalid_post_parameters(self):
-        self.app.post(self.url, params={"points-asd": 2}, user=self.student, status=400)
-        self.app.post(self.url, params={"points-": 2}, user=self.student, status=400)
-        self.app.post(self.url, params={f"points-{self.event1.pk}": ""}, user=self.student, status=400)
-        self.app.post(self.url, params={f"points-{self.event1.pk}": "asd"}, user=self.student, status=400)
+        point_params = {
+            "left-points": reward_points_of_user(self.student),
+            "redeemed-points": redeemed_points_of_user(self.student),
```
niklasmohrin, line 85: Why aren't the two just called `previous_reward_points` and `previous_redeemed_points`?

evap/rewards/views.py
```diff
@@ -22,28 +22,52 @@
...
+    reward_points_of_user,
+    save_redemptions,
+)
 from evap.staff.views import semester_view
 
 
+def redemption_validation(left_points, redeemed_points, user):
+    if left_points is not None and redeemed_points is not None and left_points.isalnum() and redeemed_points.isalnum():
+        return int(left_points) == reward_points_of_user(user) and int(redeemed_points) == redeemed_points_of_user(user)
+    raise BadRequest("Invalid redeemed-points or left-points field in redemption request")
```
niklasmohrin, line 37: I think the flow would be slightly better if it was
```python
if the_data_is_very_wrong():
  raise ...
return is_it_very_correct()
```

evap/rewards/tools.py
```diff
@@ -62,6 +62,10 @@ def reward_points_of_user(user):
     return count
 
 
+def redeemed_points_of_user(user):
+    return RewardPointRedemption.objects.filter(user_profile=user).aggregate(Sum("value"))["value__sum"] or 0
```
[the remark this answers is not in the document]
Kakadus: As there are already function with user as only parameter, such as `reward_points_of_user`, we keep it separated as well.

evap/rewards/views.py
```diff
@@ -62,10 +88,11 @@ def index(request):
     template_data = dict(
         reward_point_actions=reward_point_actions,
         total_points_available=total_points_available,
+        total_points_spent=sum(reward.value for reward in reward_point_redemptions),
```
niklasmohrin, line 91:
```suggestion
        total_points_spent=sum(redemption.value for redemption in reward_point_redemptions),
```

niklasmohrin (commented):
Looks good, haven't tested manually yet

evap/rewards/tests/test_views.py
```diff
@@ -80,10 +80,20 @@ def test_redeem_points_for_expired_event(self):
...
+        self.app.post(self.url, params={"points-asd": 2, **point_params}, user=self.student, status=400)
+        self.app.post(self.url, params={"points-": 2, **point_params}, user=self.student, status=400)
+        self.app.post(self.url, params={f"points-{self.event1.pk}": "", **point_params}, user=self.student, status=400)
+        self.app.post(
+            self.url, params={f"points-{self.event1.pk}": "asd", **point_params}, user=self.student, status=400
+        )
+        # redemption without point parameters
+        self.app.post(self.url, params={f"points-{self.event1.pk}": 1}, user=self.student, status=400)
+        point_params["previous_reward_points"] = 0
+        self.app.post(self.url, params={f"points-{self.event1.pk}": 1, **point_params}, user=self.student, status=409)
```
He3lixxx, line 96: Personally, I would put this in a new test that does what we'd expect users to do:

* Load the page, store it in a variable.
* Load the page again, extract the form, submit it to redeem a point.
* Use the stored page to submit the form again, redeeming another point.
* Assert that 409 is returned.

evap/rewards/views.py
```diff
@@ -22,28 +22,54 @@
...
+        if check_consistent_previous_redemption_counts(request):
+            redeem_reward_points(request)
+        else:
+            messages.warning(
+                request,
+                _(
+                    "Probably your browser sent multiple redemption request. You can see all successful redemptions below."
+                ),
+            )
+            status = 409
```
He3lixxx, line 72: `save_redemptions` has a transaction block where we use ACID guarantees to make sure no parallel / concurrent accesses make us go into inconsistent state.
I think checking the existing redemption points of the user against the ones provided in the request should happen inside this transaction block while the appropriate rows are locked.

Otherwise, we can still get the check to pass while another request redeems the points, bypassing our check here.

He3lixxx (changes requested):

evap/rewards/views.py
```diff
@@ -22,28 +22,54 @@
...
-        except (NoPointsSelected, NotEnoughPoints, RedemptionEventExpired) as error:
-            messages.warning(request, error)
-
+        if check_consistent_previous_redemption_counts(request):
+            redeem_reward_points(request)
+        else:
+            messages.warning(
+                request,
+                _(
+                    "Probably your browser sent multiple redemption request. You can see all successful redemptions below."
```
niklasmohrin, line 69:
```suggestion
                    "It appears that your browser sent multiple redemption requests. You can see all successful redemptions below."
```

evap/rewards/tests/test_views.py
```diff
@@ -80,10 +80,20 @@ def test_redeem_points_for_expired_event(self):
...
     def test_invalid_post_parameters(self):
-        self.app.post(self.url, params={"points-asd": 2}, user=self.student, status=400)
-        self.app.post(self.url, params={"points-": 2}, user=self.student, status=400)
-        self.app.post(self.url, params={f"points-{self.event1.pk}": ""}, user=self.student, status=400)
-        self.app.post(self.url, params={f"points-{self.event1.pk}": "asd"}, user=self.student, status=400)
+        point_params = {
+            "previous_reward_points": reward_points_of_user(self.student),
+            "previous_redeemed_points": redeemed_points_of_user(self.student),
+        }
+        self.app.post(self.url, params={"points-asd": 2, **point_params}, user=self.student, status=400)
```
niklasmohrin, line 87: This is now very lengthy, you could add a helper:
```python
def helper(redemption_params, additional_params=correct_previous_params, status=400):
    self.app.post(self.url, params={**redemption_params, **additional_params}, user=self.student, status=status)
```
You can then refactor all `post` calls to use the helper. I would also suggest not mutating `point_params` below and instead just making a new `incorrect_previous_params`

niklasmohrin (approved):
Some small things, looks good otherwise (plus what Richard said)

karyon:
I'm a little late to the party, but we had this in the past in the voting process. In addition to preventing this at the DB level, we simply disabled the voting button after clicking it: https://github.com/e-valuation/EvaP/pull/852/files#diff-8428cbf7196d5ea3af0a2d1ba803a42f924156e06765c285dd5c56e16d2c6f81R115. I haven't checked whether that's still the current solution, but wouldn't that be enough here? We might not even need to check this in the backend. If a user tries to circumvent this UI-level safety feature, it's their own fault.

He3lixxx:
Good point, but I think it's a bit different with the reward points.

For voting, users can only vote once. Any attempted later votes will be rejected by the server, so nothing bad happens.

With reward point redemption, users can redeem multiple times. This makes the multiple-open-tabs situation relevant: If I have the voting page open in two tabs, submit in one and then in the other tab, the second submission simply fails and no harm is done.
For reward points, it will submit a second redemption

karyon:
Sure. I thought this issue aimed to only fix double submits within the same tab. Personally I would say opening this page in two tabs and submitting them both requires a level of user confusion that is best fixed outside of our code.

evap/rewards/views.py
Kakadus: Also, I turned the warning into an error, as the requested redemption failed

evap/rewards/views.py
Kakadus: I am inexperienced with transactions/locks in django, and did the same as `save_redemptions`. Is this the right way to do it here?

evap/rewards/views.py
```diff
@@ -22,28 +23,58 @@
...
+@transaction.atomic
+def check_consistent_previous_redemption_counts(request):
+    list(request.user.reward_point_grantings.select_for_update())
+    list(request.user.reward_point_redemptions.select_for_update())
+    try:
+        reward = int(request.POST["previous_reward_points"])
+        redeemed = int(request.POST["previous_redeemed_points"])
+    except (KeyError, ValueError, TypeError) as e:
+        raise BadRequest("Invalid redeemed-points or left-points field in redemption request") from e
+    return reward == reward_points_of_user(request.user) and redeemed == redeemed_points_of_user(request.user)
```
He3lixxx, line 44: Could we just add those lines to `save_redemptions`? I think it may reduce code complexity (and we can't forget the call). Or do we have cases where we want to call `save_redemptions` without first checking the validity?

evap/rewards/templates/rewards_index.html
```diff
@@ -22,6 +22,8 @@
                     <form id="reward-redemption-form" action="#" method="POST" class="form-horizontal multiselect-form">
                         {% csrf_token %}
 
+                        <input type="hidden" name="previous_reward_points" value="{{ total_points_available }}">
+                        <input type="hidden" name="previous_redeemed_points" value="{{ total_points_spent }}">
```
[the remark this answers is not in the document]
Kakadus: Good point. However, this would still allow to redeem more points than points that are shown, when reward points are received after page was loaded, which seems rather odd. We could require the previous available points only to be not greater, or do you this edgecase is not relevant?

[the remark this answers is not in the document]
niklasmohrin: I think this isn't a problem, since users can just redeem more points afterwards if they feel like it.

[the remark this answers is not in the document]
Kakadus: I removed the redeemed reward points field now, but reading your answer once again, I think you might have misunderstood me there:
- user has one reward point and loads the redemption page
- user gains 3 reward points (not shown in UI)
- user (mistakenly) types in 4 points and submits the form
- all server-side checks pass, 4 points are redeemed sucessfully
=> the user might have not anticipated that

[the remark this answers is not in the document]
He3lixxx: The user shouldn't be able to enter 4 points in the input / dropdown field, because it should be limited to the available amount of points.

-> For thus, users would have to hand-craft requests -> not a situation where we need to protect users.

evap/rewards/views.py
Kakadus: Refactored it now, introducing an OutdatedRedemptionData-Exception, thrown in save_redemptions and catched in redeem_reward_points. The fact that we showed the message as an error and that we want to change the status to 409, forced me to introduce some changes to the structure of redeem_reward_points (function returns the desired status now)

evap/rewards/tools.py
```diff
@@ -25,11 +26,20 @@
 
 @login_required
 @transaction.atomic
-def save_redemptions(request, redemptions: Dict[int, int]):
+def save_redemptions(request, redemptions: Dict[int, int], previous_reward_points):
+    """Saves the redemptions in the given dict. Returns True if the redemptions were saved successfully, False otherwise."""
```
niklasmohrin, line 30: You could add a type annotation for `previous_reward_points` (for `request`, it is more complicated, so it can stay as is)

Secondly, this always returns `True`, or raises

evap/rewards/views.py
```diff
@@ -26,24 +27,33 @@
...
+        raise BadRequest from e
+
+    try:
+        save_redemptions(request, redemptions, previous_redeemed_points)
+        messages.success(request, _("You successfully redeemed your points."))
+    except (NoPointsSelected, NotEnoughPoints, RedemptionEventExpired) as error:
+        messages.warning(request, error)
+    except OutdatedRedemptionData as error:
+        messages.error(request, error)
+        return 409
```
niklasmohrin, line 48: While this does correctly reproduce the current behavior, it seems a bit odd that we return 200 even if saving the points failed. Let's make a "very small" issue for first semester students to make this return 400 in the first `except` instead.

I noticed that `assertContains` is a bit weird and also asserts that the status code is 200 unless told otherwise which could lead to confusion / requires us to specify this twice - I think they will handle it though :)

evap/rewards/tools.py
Kakadus: Jup. Simply forgot to remove it

evap/rewards/views.py
```diff
@@ -26,24 +27,33 @@
 from evap.staff.views import semester_view
 
 
+def redeem_reward_points(request):
+    redemptions = {}
+    try:
+        for key, value in request.POST.items():
+            if key.startswith("points-"):
```
He3lixxx, line 34: Out of curiosity: Is there a reason why we extract this information manually? Could we instead do it using standard django tools, e.g. using formset logic? I'd then make a new issue for that.

He3lixxx (approved):
@niklasmohrin a few small points where I considered adding an annotation, but if you're fine with them, we can still merge:
* The "this case should fail" tests do not actually assert that the request fails due to the error condition we induce. Personally, I'd put a "and when I remove the wrong input, it doesn't fail" check afterwards. This way, these can never fail for other reasons (input format is wrong, unrelated to what we wanted to test)
* We don't separate concerns, especially with the exception, where the user-string is defined in the innermost technical layer of point redemption.
* `previous_redeemed_points` and `total_points_spent` are two names for the same thing.
* `previous_reward_points` argument name doesn't encode "This is what the request indicated the user thinks is the most up-to-date information". Might be mistaken for the actual past redeemed points in the future.

However, its all nitpicking stuff, so I'm fine with merging.

evap/rewards/tools.py
```diff
@@ -25,11 +26,19 @@
 
 @login_required
 @transaction.atomic
-def save_redemptions(request, redemptions: Dict[int, int]):
+def save_redemptions(request, redemptions: Dict[int, int], previous_reward_points):
```
niklasmohrin, line 29: This should be `previous_redeemed_points`, or? And let's add a type annotation :)

evap/rewards/tools.py
```diff
@@ -25,11 +26,19 @@
...
     # lock these rows to prevent race conditions
     list(request.user.reward_point_grantings.select_for_update())
     list(request.user.reward_point_redemptions.select_for_update())
 
+    # check consistent previous redeemed points
```
niklasmohrin, line 34: Should we extend this comment to explain why we are _not_ also comparing rewarded points? What do you think @He3lixxx ?

evap/rewards/tests/test_views.py
```diff
@@ -79,13 +79,40 @@ def test_redeem_points_for_expired_event(self):
         self.assertContains(response, "event expired already.")
         self.assertEqual(5, reward_points_of_user(self.student))
 
+    def helper_post_redemption_request(self, redemption_params, additional_params=None, status=400):
```
niklasmohrin, line 82: Now, we never use the `status` parameter; should we remove it and instead call the function `post_invalid_redemption_request` or so, to encode that we expect an error?

niklasmohrin: (or better, add some cases where we use this function and the request is successful; then, the default of the helper probably shouldn't be 400, but something positive)

niklasmohrin (changes requested):
I think having both `previous_redeemed_points` and `total_points_spent` is fair, because they are the same value, but in different contexts; we could argue where to correctly place the boundary for the domains, but I think that it is appropriate as currently implemented

evap/rewards/tools.py
niklasmohrin: (this also closes the comment from @He3lixxx about the name)

niklasmohrin:
About the user strings: I agree, we should probably move that to where the exceptions are caught; should we make a small, good first issue about all three exceptions?

evap/rewards/tools.py
He3lixxx: I'm fine with both the current version and an additional comment.

He3lixxx:
> About the user strings: I agree, we should probably move that to where the exceptions are caught; should we make a small, good first issue about all three exceptions?

We could do that. Not sure if it's such a good first issue -- I think to new contributors, refactoring isn't the most rewarding activity -- but it might work out okay.

evap/rewards/views.py
Kakadus: Yeah, probably possible. Maybe the way how we transmit the mapping of event-id and points need to be changed as well

niklasmohrin (approved):
Looks good!

niklasmohrin:
We will merge after #1822 :)

niklasmohrin (approved):

PR #1792, opened 2022-08-15
Add URL to Review Reminder Email

FranzSw:
Adds the url for evaluations to the review reminder and changes the variable in frontend email template


evap/evaluation/management/commands/send_reminders.py
```diff
@@ -49,16 +49,13 @@ def send_student_reminders():
...
-                for evaluation in Evaluation.objects.filter(state=Evaluation.State.EVALUATED)
-                if evaluation.textanswer_review_state == Evaluation.TextAnswerReviewState.REVIEW_URGENT
-            ]
-            if not evaluations:
+            evaluation_url_tuples = Evaluation.get_evaluation_url_tuples_to_review_urgent()
+            if not evaluation_url_tuples:
                 logger.info("no evaluations require a reminder about text answer review.")
                 return
-            evaluations = sorted(evaluations, key=lambda evaluation: evaluation.full_name)
+            evaluation_url_tuples = sorted(evaluation_url_tuples, key=lambda evaluation: evaluation[0].full_name)
```
He3lixxx, line 56: Could / Should the sorting be part of `get_evaluation_url_tuples_to_review_urgent`?

evap/evaluation/management/commands/send_reminders.py
```diff
@@ -49,16 +49,13 @@ def send_student_reminders():
     @staticmethod
     def send_textanswer_reminders():
         if datetime.date.today().weekday() in settings.TEXTANSWER_REVIEW_REMINDER_WEEKDAYS:
-            evaluations = [
-                evaluation
-                for evaluation in Evaluation.objects.filter(state=Evaluation.State.EVALUATED)
-                if evaluation.textanswer_review_state == Evaluation.TextAnswerReviewState.REVIEW_URGENT
-            ]
-            if not evaluations:
+            evaluation_url_tuples = Evaluation.get_evaluation_url_tuples_to_review_urgent()
```
He3lixxx, line 52: The `to_review_urgent` part seems odd to me. How about `with_urgent_review`?

evap/development/fixtures/test_data.json
```diff
@@ -128389,8 +128389,8 @@
   "fields": {
     "name": "Text Answer Review Reminder",
     "subject": "[EvaP] Bitte Textantworten überprüfen / Please review text answers",
-    "plain_content": "(English version below)\r\n\r\n\r\nHallo {{ user.first_name }},\r\n\r\nes gibt noch nicht überprüfte Textantworten für eine oder mehrere Evaluierungen, bei denen der Evaluierungszeitraum abgelaufen ist und nicht mehr auf Notenveröffentlichungen gewartet werden muss. Bitte überprüfe die Textantworten für diese Evaluierungen möglichst bald:\r\n{% for evaluation in evaluations %}    - {{ evaluation.full_name_de }}\r\n{% endfor %}\r\n\r\n(Dies ist eine automatisch versendete E-Mail.)\r\n\r\n\r\n--\r\n\r\n\r\nDear {{ user.first_name }},\r\n\r\nthere are text answers not yet reviewed for one or more evaluations where the evaluation period has ended and there is no need to wait for grade publishing. Please review the text answers for these evaluations as soon as possible:\r\n{% for evaluation in evaluations %}    - {{ evaluation.full_name_en }}\r\n{% endfor %}\r\n\r\n(This is an automated message.)",
```
He3lixxx, line 128392: We lost some whitespace that previously indented the evaluation list.

evap/evaluation/models.py
```diff
@@ -958,6 +959,14 @@ def annotate_with_participant_and_voter_counts(cls, evaluation_query):
             num_voters=Subquery(voter_count_subquery),
         )
 
+    @classmethod
+    def get_evaluation_url_tuples_to_review_urgent(cls) -> List[Tuple['Evaluation', str]]:
+        return [
+            (evaluation, f"{PAGE_URL}/staff/semester/{evaluation.course.semester.id}/evaluation/{evaluation.id}/textanswers")
```
He3lixxx, line 965: I think we should use django's [`reverse`](https://docs.djangoproject.com/en/dev/ref/urlresolvers/#reverse) method here

evap/staff/views.py
```diff
@@ -2217,9 +2217,10 @@ def template_edit(request, template_id):
         EmailTemplate.EDITOR_REVIEW_REMINDER,
         EmailTemplate.PUBLISHING_NOTICE_CONTRIBUTOR,
         EmailTemplate.PUBLISHING_NOTICE_PARTICIPANT,
-        EmailTemplate.TEXT_ANSWER_REVIEW_REMINDER,
     ]:
         available_variables += ["evaluations"]
+    elif template.name == EmailTemplate.TEXT_ANSWER_REVIEW_REMINDER:
+        available_variables += ["evaluation_url_tuples"]
```
He3lixxx, line 2223: Could you add a test for the available variables in this case (codecov says this line is not covered).

He3lixxx (changes requested):

evap/evaluation/models.py
```diff
@@ -958,6 +958,22 @@ def annotate_with_participant_and_voter_counts(cls, evaluation_query):
             num_voters=Subquery(voter_count_subquery),
         )
 
+    @classmethod
+    def get_sorted_evaluation_url_tuples_with_urgent_review(cls) -> List[Tuple["Evaluation", str]]:
+        evaluation_url_tuples = [
+            (
+                evaluation,
+                reverse(
```
He3lixxx, line 966: hmm, sorry, I should have been more clear here: `reverse` only gives the path component of the URL, but not the hostname (i.e. this gives `/semester/1/evaluation/1/textanswers`, without `https://example.com`). So we still need to manually add the host name.

The tests currently pass, so we should probably add a test for the URL generated here.

He3lixxx (changes requested):

evap/evaluation/models.py
```diff
@@ -958,6 +959,23 @@ def annotate_with_participant_and_voter_counts(cls, evaluation_query):
...
         )
 
+    @classmethod
+    def get_sorted_evaluation_url_tuples_with_urgent_review(cls) -> List[Tuple["Evaluation", str]]:
+        evaluation_url_tuples = [
+            (
+                evaluation,
+                PAGE_URL
+                + reverse(
+                    "staff:evaluation_textanswers",
```
niklasmohrin, line 969: Not sure, but maybe this whole method should be a free function in the file where it is used, it feels very specific to the one use case. For example, currently the name suggests that there is one kind of "url" that is associated with an evaluation, the url returned is actually related to textanswer review / staff things though.  The type annotation could then also be changed to remove the quotes :) What do you think?

evap/evaluation/models.py
```diff
@@ -958,6 +959,23 @@ def annotate_with_participant_and_voter_counts(cls, evaluation_query):
...
+                    kwargs={"semester_id": evaluation.course.semester.id, "evaluation_id": evaluation.id},
+                ),
+            )
+            for evaluation in Evaluation.objects.filter(state=Evaluation.State.EVALUATED)
+            if evaluation.textanswer_review_state == Evaluation.TextAnswerReviewState.REVIEW_URGENT
+        ]
+        evaluation_url_tuples = sorted(evaluation_url_tuples, key=lambda evaluation: evaluation[0].full_name)
```
niklasmohrin, line 976:
```suggestion
        evaluation_url_tuples = sorted(evaluation_url_tuples, key=lambda evaluation, __: evaluation.full_name)
```

evap/evaluation/models.py
```diff
@@ -38,6 +38,7 @@
     translate,
     vote_end_datetime,
 )
+from evap.settings import PAGE_URL
```
niklasmohrin, line 41: Please use `dango.conf.settings` instead of `evap.settings` :)

niklasmohrin (commented):
content looks good, I have some small code comments:

He3lixxx (commented):
The code looks good to me if the annotations by @niklasmohrin are resolved. Thanks!

I'm wondering if we want to add a test. We had a broken link generation in between and no test failed. A simple "check that a hardcoded substring is contained in the mail" test would have caught that.

evap/staff/tests/test_views.py
```diff
@@ -3096,6 +3096,12 @@ def test_emailtemplate(self):
         self.assertEqual(self.template.plain_content, "plain_content: mflkd862xmnbo5")
         self.assertEqual(self.template.html_content, "html_content: <p>mflkd862xmnbo5</p>")
 
+    def test_review_reminder_template_tag(self):
+        review_reminder_template = EmailTemplate.objects.get(name=EmailTemplate.TEXT_ANSWER_REVIEW_REMINDER)
+        page = self.app.get(f"/staff/template/{review_reminder_template.pk}", user=self.manager, status=200)
+
+        self.assertContains(page, "evaluation_url_tuples")
```
janno42, line 3103: i think it would be better to check that a review url is included in the generated email to make sure the link generation works.

janno42 (changes requested):

evap/staff/tests/test_views.py
```diff
@@ -874,6 +874,24 @@ def test_form(self):
         self.assertIn("uiae", mail.outbox[0].body)
 
 
+class TestSendReviewReminderTemplate(WebTestStaffMode):
+    @classmethod
+    def setUpTestData(cls):
+        cls.manager = make_manager()
+        cls.evaluation = baker.make(Evaluation, course__responsibles=[], state=Evaluation.State.EVALUATED)
```
He3lixxx, line 881: Is `course__responsibles=[]` required here?

evap/staff/tests/test_views.py
```diff
@@ -874,6 +874,24 @@ def test_form(self):
...
+
+    def test_form(self):
+        evaluation_url_tuples = [(self.evaluation, "http://example.com/evaluation/42")]
+
+        mock_template = MagicMock()
+        with patch.object(EmailTemplate.objects, "get", MagicMock(return_value=mock_template)):
+            with patch.object(mock_template, "send_to_user", MagicMock()) as send_mock:
+                EmailTemplate.send_textanswer_reminder_to_user(self.manager, evaluation_url_tuples)
```
He3lixxx, line 889: Could we just patch `Template.render`?

evap/staff/tests/test_views.py
```diff
@@ -874,6 +874,24 @@ def test_form(self):
...
 
 
+class TestSendReviewReminderTemplate(WebTestStaffMode):
+    @classmethod
+    def setUpTestData(cls):
+        cls.manager = make_manager()
+        cls.evaluation = baker.make(Evaluation, course__responsibles=[], state=Evaluation.State.EVALUATED)
+
+    def test_form(self):
+        evaluation_url_tuples = [(self.evaluation, "http://example.com/evaluation/42")]
```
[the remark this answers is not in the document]
janno42: the email must link to the textanswer review page (as correctly defined in [send_reminders.py](https://github.com/e-valuation/EvaP/pull/1792/files#diff-1830c52762a53b99f7a49a5006f49587f31202f04708849ad12f8a0fdbfce92b))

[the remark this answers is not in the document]
FranzSw: This was just supposed to be any evaluation with url for testing purposes

evap/staff/tests/test_views.py
FranzSw: No, I've removed it

evap/staff/tests/test_views.py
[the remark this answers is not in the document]
He3lixxx: Ah, I see. From `self.assertEqual(evaluation_url_tuples, ...)` below I assumed that this was the _expected_ list of (evaluation, url) pairs passed to the template.

Could we make it so that you don't pass in the tuples, but instead let the system build them, so that we also test that the tuples are built correctly?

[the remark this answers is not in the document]
FranzSw: should be tested now :D

evap/staff/views.py
He3lixxx, line 2223: codecov says this is not covered. I don't think we need to have it covered, though. @niklasmohrin how much do we want those 96%?

He3lixxx (approved):

evap/staff/views.py
niklasmohrin: :shrug: its one line, I guess it's fine to leave it uncovered

anyways, why is codecov saying +0.42% on this PR? Is it against an old `main`? :thinking:

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

janno42:
Thanks :shipit:

PR #1794, opened 2022-08-15
Improve database sorting for user profiles

### LINKED ISSUE #1624: Handle user profile sorting with incomplete names

postmartem:
Closes #1624


niklasmohrin:
The `FieldException` is a bug in django, they fixed it 3 months ago. We will update in the next couple of days

evap/evaluation/models.py
```diff
@@ -356,10 +356,7 @@ def midterm_grade_documents(self):
 
     @cached_property
     def responsibles_names(self):
-        ordered_responsibles = sorted(
-            self.responsibles.all(), key=lambda responsible: (responsible.last_name, responsible.full_name)
-        )
-        return ", ".join([responsible.full_name for responsible in ordered_responsibles])
+        return ", ".join([responsible.full_name for responsible in self.responsibles.all()])
```
niklasmohrin, line 359:
```suggestion
        return ", ".join(responsible.full_name for responsible in self.responsibles.all())
```

evap/evaluation/models.py
```diff
@@ -1547,7 +1544,7 @@ class UserProfile(AbstractBaseUser, PermissionsMixin):
     is_active = models.BooleanField(default=True, verbose_name=_("active"))
 
     class Meta:
-        ordering = ["last_name", "first_name", "email"]
+        ordering = [Lower(Concat("last_name", "first_name", "email"))]
```
niklasmohrin, line 1547: sorry, we only now noticed that we may not want `Concat` afterall; it should just sort for the `Lower` of each individually

evap/evaluation/tests/test_models.py
```diff
@@ -670,6 +670,19 @@ def test_get_sorted_due_evaluations(self):
         sorted_evaluations = student.get_sorted_due_evaluations()
         self.assertEqual(sorted_evaluations, [(evaluations[1], 0), (evaluations[0], 0), (evaluations[2], 1)])
 
+    def test_can_be_sorted(self):
```
niklasmohrin, line 673: I mean, they _can_ be sorted, but _are_ they? And is that sorting correct? Maybe `test_correct_sorting`?

evap/evaluation/tests/test_models.py
```diff
@@ -670,6 +670,19 @@ def test_get_sorted_due_evaluations(self):
...
+        baker.make(
+            UserProfile,
+            last_name=iter(["B", "A", None]),
+            first_name=iter(["X", None, None]),
+            email=iter(["ax@example.com", "b@example.com", "unnamed@example.com"]),
+            _quantity=3,
+            _bulk_create=True,
+        )
+        user_list = list(UserProfile.objects.all())
+        email_list = [user.email for user in user_list]
```
niklasmohrin, line 683:
```suggestion
        email_list = [user.email for user in UserProfile.objects.all()]
```

niklasmohrin (changes requested):

evap/evaluation/migrations/0131_alter_userprofile_options.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 4.0.6 on 2022-08-15 20:12
```
niklasmohrin, line 1: one more thing: Can you rename this file to `0131_userprofile_ordering.py`?

evap/evaluation/tests/test_models.py
```diff
@@ -670,6 +670,18 @@ def test_get_sorted_due_evaluations(self):
         sorted_evaluations = student.get_sorted_due_evaluations()
         self.assertEqual(sorted_evaluations, [(evaluations[1], 0), (evaluations[0], 0), (evaluations[2], 1)])
 
+    def test_correct_sorting(self):
+        baker.make(
+            UserProfile,
+            last_name=iter(["B", "A", None]),
```
niklasmohrin, line 676: Strictly, this doesn't test that a) we `Lower` the fields and b) that there are more sorting keys than just the last name

He3lixxx (approved):

niklasmohrin (approved):

PR #1799, opened 2022-08-22
Django 4.1 and other requirements

### LINKED ISSUE #1796: Update to django 4.1
(#1794 is currently blocked by a bug that is fixed in django 4.1)

[Same procedure as every year](https://github.com/e-valuation/EvaP/issues/1387)

Some points I think might be relevant
* ~Django enables the cached template loader by default now. I think we want to be able to edit a template, reload the page, and instantly see the changes reflected with a development setup. So, we probably want to set a custom, non-cached template loader when `DEBUG=True`?~ **Edit**: As pointed out by @karyon below, the cached template loader has some auto-reload functionality, should work for us.
* ~https://code.djangoproject.com/ticket/33631 was closed, so `blocktranslate asvar` now marks the output as safe. This should allow us to remove all `|safe` filter calls that we have left in our templates after https://github.com/e-valuation/EvaP/pull/1737.~ **Edit**: This will be in 4.2, not 4.1
* ~Django defaults to a new forms renderer using `div` tags. We should check if this breaks some layout because CSS rules do not apply anymore.~ **Edit**: As @karyon points out below, default will be changed in 5.0.

niklasmohrin:
Closes #1796

See commit messages. The ideas about using constraints instead of asserts from the issue can be done in another PR. The cached template things don't seem to be a problem, I get new html after editing files without restarting the server.


He3lixxx (approved):

niklasmohrin:
There is a new version of `pylint` already (2.15.0), but that throws errors when combining `Q` objects with binary operators. Let's stick to 2.14 until that is sorted out

PR #1802, opened 2022-08-29
Port textanswer review quick view to TypeScript

niklasmohrin:
Before working on the textanswer flagging, I wanted to get this done first.
This ports `quick-review.js` to TypeScript and replaces all uses of jQuery with standard DOM APIs.
I will also apply some refactorings and renames where I see fit.


evap/static/ts/src/quick-review-slider.ts
```diff
@@ -0,0 +1,410 @@
+// This file accompanies the "quick" view in the textanswer review site. The quick view presents textanswers in a
+// _slider_ that is made up of three layers - answer, question and questionnaire. One answer is shown at a time, it is
+// _active_. The slider contains a form that has multiple submit buttons, one for each review decision, one for revoking
+// a previous decision and one for editing the answer. If a review decision is made, the next answer slides in from the
+// right. At the end, there is one _alert slide_ that summarises the current state of affairs regarding textanswer
```
janno42, line 5:
```suggestion
// right. At the end, there is one _alert slide_ that summarizes the current state of affairs regarding textanswer
```

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

evap/static/ts/src/quick-review-slider.ts
```diff
@@ -0,0 +1,410 @@
...
+    // Sliding
+    //
+    public startOver = (where: StartOverWhere) => {
+        const decided = this.slider.querySelectorAll<HTMLElement>(`[data-layer="${Layer.Answer}"][data-review]`);
+        const undecided = this.slider.querySelectorAll<HTMLElement>(
+            `[data-layer="${Layer.Answer}"]:not([data-review])`,
+        );
+        this.answerSlides = Array.from(decided).concat(Array.from(undecided));
+
+        const startIndex = where === StartOverWhere.Undecided && undecided.length > 0 ? decided.length : 0;
```
He3lixxx, line 355: I'd prefer some parentheses here (I don't have an operator precedence table in my head).

evap/static/ts/src/quick-review-slider.ts
```diff
@@ -0,0 +1,410 @@
...
+                activeInParentLayer =
+                    findPreviousElementSibling(nextActiveElement, `[data-layer="${layer - 1}"]`) ?? undefined;
+                activeInParentLayer = activeInParentLayer as HTMLElement | undefined;
+            }
+            this.slideLayer(layer - 1, direction, activeInParentLayer);
+        }
+
+        if (nextActiveElement) {
+            // First, translate nextActiveElement to the left or right and
+            // then, in the next frame, move it back to the middle
```
He3lixxx, line 404: I think we want to have a bit more explanation here. Especially: The animation takes more than one browser frame. Why is it okay to change the class in the next frame?

He3lixxx (commented):
Code looks mostly good (but I was very distracted during review, so I might have missed a lot)

evap/staff/templates/quick-review.js
```diff
@@ -1,3 +1,5 @@
+// TODO: DELETE ME
```
He3lixxx, line 1: keep in mind

He3lixxx (approved):
UI seems to work reliably

evap/static/ts/src/quick-review-slider.ts
niklasmohrin: prettier says "but you should" :/ I moved the condition into another variable

PR #1809, opened 2022-09-26
Fix contact modal error (fixes #1803)

### LINKED ISSUE #1803: Modals fail to open
Some modals (at least those in the contributor evaluation form) cannot be opened because the defined `onClick` references a non-existing method. Since 6aa3a5fcd49, the contact modal is opened via a button with a specific id instead of onClick.

He3lixxx:
Allow having 0 or multiple open buttons per modal (as is required by our usages).

I guess a frontend test would be nice. I'll try to build something...

### DIFF
--- a/evap/static/ts/src/contact_modal.ts
+++ b/evap/static/ts/src/contact_modal.ts
@@ -10,7 +10,7 @@ export class ContactModalLogic {
     private readonly successMessageModal: bootstrap.Modal;
     private readonly actionButtonElement: HTMLButtonElement;
     private readonly messageTextElement: HTMLInputElement;
-    private readonly showButtonElements: NodeListOf<HTMLElement>;
+    private readonly showButtonElements: Array<HTMLElement>;
     private readonly title: string;
 
     // may be null if anonymous feedback is not enabled
@@ -23,7 +23,7 @@ export class ContactModalLogic {
         this.actionButtonElement = selectOrError("#" + modalId + "ActionButton");
         this.messageTextElement = selectOrError("#" + modalId + "MessageText");
         this.anonymousRadioElement = document.querySelector<HTMLInputElement>("#" + modalId + "AnonymousName");
-        this.showButtonElements = document.querySelectorAll(`#${modalId}ShowButton, .${modalId}ShowButton`);
+        this.showButtonElements = Array.from(document.querySelectorAll(`#${modalId}ShowButton, .${modalId}ShowButton`));
     }
 
     public attach = (): void => {
--- a/evap/static/ts/src/contact_modal.ts
+++ b/evap/static/ts/src/contact_modal.ts
@@ -10,7 +10,7 @@ export class ContactModalLogic {
     private readonly successMessageModal: bootstrap.Modal;
     private readonly actionButtonElement: HTMLButtonElement;
     private readonly messageTextElement: HTMLInputElement;
-    private readonly showButtonElement: HTMLElement;
+    private readonly showButtonElements: NodeListOf<HTMLElement>;
     private readonly title: string;
 
     // may be null if anonymous feedback is not enabled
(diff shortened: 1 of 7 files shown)


evap/static/ts/src/contact_modal.ts
```diff
@@ -10,7 +10,7 @@ export class ContactModalLogic {
     private readonly successMessageModal: bootstrap.Modal;
     private readonly actionButtonElement: HTMLButtonElement;
     private readonly messageTextElement: HTMLInputElement;
-    private readonly showButtonElement: HTMLElement;
+    private readonly showButtonElements: NodeListOf<HTMLElement>;
```
niklasmohrin, line 13: If I recall correctly, node lists are a bit icky because they don't have some methods I expect (like map, filter), so I just always went with `Array` instead (see `quick-review-slider.ts`) - I am fine with both here though

niklasmohrin (approved):
Looks good, test + CI and we can merge

evap/static/ts/src/contact_modal.ts
He3lixxx: I just used the type that was returned by the API that I used, I think.

evap/static/ts/tests/frontend/contributor-evaluation-edit.ts
```diff
@@ -0,0 +1,95 @@
...
+        *::before {
+            transition-delay: 0s !important;
+            transition-duration: 0s !important;
+        }`;
+
+        page.addStyleTag({ content });
+
+        const modalVisible = async (modalHandle: ElementHandle) =>
+            await page.evaluate(modal => {
+                return window.getComputedStyle(modal).display == "block";
```
niklasmohrin, line 22:
```suggestion
                return window.getComputedStyle(modal).display === "block";
```

evap/static/ts/tests/frontend/contributor-evaluation-edit.ts
```diff
@@ -0,0 +1,95 @@
...
+            }, modalHandle);
+
+        // "Request changes" button
+
+        const changeEvaluationRequestModal = await page.$("#changeEvaluationRequestModal");
+        assert(changeEvaluationRequestModal != null);
```
niklasmohrin, line 28: (import needed)
```suggestion
        assertDefined(changeEvaluationRequestModal);
```

evap/static/ts/tests/frontend/contributor-evaluation-edit.ts
```diff
@@ -0,0 +1,95 @@
...
+    pageHandler("/contributor/evaluation/PK/edit/normal.html", async page => {
+        const content = `
+        *,
+        *::after,
+        *::before {
+            transition-delay: 0s !important;
+            transition-duration: 0s !important;
+        }`;
+
+        page.addStyleTag({ content });
```
niklasmohrin, line 18: my linter says this should be awaited

evap/static/ts/tests/frontend/contributor-evaluation-edit.ts
```diff
@@ -0,0 +1,95 @@
...
+
+        const createAccountRequestModal = await page.$("#createAccountRequestModal");
+        assert(createAccountRequestModal != null);
+        expect(await modalVisible(createAccountRequestModal)).toBe(false);
+
+        const [requestAccountCreateButton] = await page.$x("//button[contains(., 'Request creation of new account')]");
+        assert(requestAccountCreateButton != null);
+        await (requestAccountCreateButton as ElementHandle<Element>).click();
+        await page.waitForSelector("#createAccountRequestModal", { visible: true });
+        expect(await modalVisible(createAccountRequestModal)).not.toBe(null);
```
niklasmohrin, line 52: Shouldn't it be true or false?

evap/static/ts/tests/frontend/contributor-evaluation-edit.ts
```diff
@@ -0,0 +1,95 @@
+import { test, expect } from "@jest/globals";
+import { ElementHandle } from "puppeteer";
+import { assert } from "../../src/utils";
+
+import { pageHandler } from "../utils/page";
+
+test(
+    "contact-modal-opens",
+    pageHandler("/contributor/evaluation/PK/edit/normal.html", async page => {
+        const content = `
```
niklasmohrin, line 10: we can move this outside of the two tests into a constant :)

evap/static/ts/src/contact_modal.ts
niklasmohrin: But the class doesn't care about that. My point was that I personally prefer to have the rich / standard `Array` instead of the odd `NodeList` - even if we don't strictly need it (yet).

niklasmohrin (approved):

PR #1811, opened 2022-09-26
Fix TextAnswer anonymization

### LINKED ISSUE #1798: Anonymize can fail when replacing TextAnswers
The anonymize script replaces TextAnswers with lorem ipsum text. If an original answer exists, it gets replaced as well. In some circumstances this can result in an AssertionError because on saving the answer and original answer must not be equal:

```
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/opt/evap/env/lib/python3.8/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/opt/evap/env/lib/python3.8/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/opt/evap/env/lib/python3.8/site-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/opt/evap/env/lib/python3.8/site-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File "/opt/evap/evap/evaluation/management/commands/anonymize.py", line 51, in handle
    self.anonymize_data()
  File "/opt/evap/evap/evaluation/management/commands/anonymize.py", line 71, in anonymize_data
    self.anonymize_answers(lorem_ipsum)
  File "/opt/evap/evap/evaluation/management/commands/anonymize.py", line 225, in anonymize_answers
    text_answer.save()
  File "/opt/evap/evap/evaluation/models.py", line 1448, in save
    assert self.answer != self.original_answer
AssertionError
```

It has to be made sure that the answer and original_answer are not equal after anonymization.

janno42:
fixes #1798

### DIFF
--- a/evap/evaluation/management/commands/anonymize.py
+++ b/evap/evaluation/management/commands/anonymize.py
@@ -222,6 +222,9 @@ class Command(BaseCommand):
             text_answer.answer = self.lorem(text_answer.answer, lorem_ipsum)
             if text_answer.original_answer:
                 text_answer.original_answer = self.lorem(text_answer.original_answer, lorem_ipsum)
+                # answer and original answer must not be the same (see #1798)
+                if text_answer.answer == text_answer.original_answer:
+                    text_answer.original_answer += " ipsum"
             text_answer.save()
 
         self.stdout.write("Shuffling rating answer counter counts...")


niklasmohrin (approved):
I suppose that works :D I think we should add a comment explaining these lines, because they seem a bit arbitrary otherwise

He3lixxx (approved):
Agree with @niklasmohrin. I'd suggest a simple "These can't be identical, see #issue number" comment.

PR #1814, opened 2022-09-27
Remove remaining uses of `@login_required`

niklasmohrin:
In #1452, we made `login_required` the default and introduced
the `@no_login_required` decorator to opt out instead. Apparently, we
missed some usages of `login_required` or they sneaked back in through
merges or whatnot.

### DIFF
--- a/evap/evaluation/tests/test_views.py
+++ b/evap/evaluation/tests/test_views.py
@@ -150,6 +150,10 @@ class TestProfileView(WebTest):
         result = create_evaluation_with_responsible_and_editor()
         cls.responsible = result["responsible"]
 
+    def test_requires_login(self):
+        response = self.app.get(self.url, user=None, status=302)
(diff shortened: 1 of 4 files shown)


He3lixxx (approved):
Do we have tests for one of these (both: a no-login view can be accessed without logging in, and a undecorated view can not be accessed)? It doesn't look like this was in 1452, and a quick grep doesn't give me such a test. Maybe we want to add something like that.

`test_passworduser_login` tests the first one, to some extent.

niklasmohrin:
I wrote a test for the `/profile` view because it was the only basic one I could find that didn't also have `@reviewer_required` or so; we could also make this a mixin and sprinkle it onto more `Test*` classes

He3lixxx (approved):
Works for me. I'm fine with just one example view, just want to make sure that the generic approach works in general (and e.g. the handler is not bypassed completely)

PR #1815, opened 2022-09-27
`TextAnswer` flagging

### LINKED ISSUE #1118: Text answer flagging

niklasmohrin:
Closes #1118

Open questions:

- [x] ~~Should the "undecided" view show decided, but flagged answers?~~ no
- [x] ~~Should there be a separate "flagged" view?~~ yes
- ...

### DIFF
--- a/evap/staff/templates/staff_evaluation_textanswer_flag_radios.html
+++ b/evap/staff/templates/staff_evaluation_textanswer_flag_radios.html
@@ -1,0 +1,40 @@
+<div class="btn-group icon-buttons" role="group">
+    <!--
+        Note: Firefox caches the state of inputs (not spec compliant), but we
+        can disable that using `autocomplete=off` (not spec compliant). I don't
+        want to think about input caching.
+    -->
+    <input
+        {% if form %}form={{ form }}{% endif %}
+        type="radio"
+        name="is_flagged"
+        value="false"
+        id="radio-flag-false-{{ id }}"
+        class="btn-check"
+        autocomplete="off"
+        {% if not is_initially_flagged %}checked{% endif %}
+    />
+    <label
+        for="radio-flag-false-{{ id }}"
+        class="btn btn-sm btn-light for-above"
+    >
+        <span class="far fa-circle"></span>
+    </label>
+
+    <input
+        {% if form %}form={{ form }}{% endif %}
+        type="radio"
+        name="is_flagged"
+        value="true"
+        id="radio-flag-true-{{ id }}"
+        class="btn-check"
+        autocomplete="off"
+        {% if is_initially_flagged %}checked{% endif %}
+    />
+    <label
+        for="radio-flag-true-{{ id }}"
+        class="btn btn-sm btn-light for-above"
+    >
+        <span class="fas fa-flag"></span>
+    </label>
+</div>
--- a/evap/staff/templates/staff_evaluation_textanswers_full.html
+++ b/evap/staff/templates/staff_evaluation_textanswers_full.html
@@ -37,7 +37,7 @@
 {% block additional_javascript %}
     <script type="text/javascript">
         document
-            .querySelectorAll(".full-update-textanswer-form input[type='radio']")
-            .forEach(radio => radio.addEventListener("change", () => radio.form.submit()));
+            .querySelectorAll(".full-update-textanswer-form input[type='radio'], .full-textanswer-flag-form input[type='radio']")
+            .forEach(input => input.addEventListener("change", () => input.form.requestSubmit()));
     </script>
 {% endblock %}
--- a/evap/staff/templates/staff_evaluation_textanswers_quick.html
+++ b/evap/staff/templates/staff_evaluation_textanswers_quick.html
@@ -31,6 +31,18 @@
         </div>
     </div>
 
+    <form method="POST" action="{% url 'staff:evaluation_textanswers_update_flag' %}" id="shared-full-textanswer-flag-form">
+        {% csrf_token %}
+
+        <!--
+            This form ("the flag form") is closely related to the `textanswer-update-form` form ("the main form"), but
+            separated to avoid overwriting the review decision when changing the flagged status and vice-versa. Since
+            the main form needs to wrap the entire slider and form tags cannot be nested, the flag form lives next to
+            the main form. The inputs for the flag form are linked using the `form` attribute (and thus also excluded
+            from the main form).
+        -->
+    </form>
+
     <form method="POST" action="{% url 'staff:evaluation_textanswers_update_publish' %}" id="textanswer-update-form">
         {% csrf_token %}
 
@@ -63,6 +75,14 @@
                                     value="{{ answer.id }}"
                                     disabled
                                 />
+                                <input
+                                    form="shared-full-textanswer-flag-form"
+                                    type="hidden"
+                                    name="answer_id"
+                                    value="{{ answer.id }}"
+                                    disabled
+                                />
+
                                 <div
                                     class="slider-item card-body"
                                     data-layer="2"
@@ -77,6 +97,7 @@
                                             data-review="delete"
                                         {% endif %}
                                     {% endif %}
+                                    {% if answer.is_flagged %} data-is-flagged {% endif %}
                                 >
                                     {{ answer.answer|linebreaksbr }}
                                     {% if answer.original_answer %}
@@ -114,7 +135,7 @@
             </div>
 
 
-            <div class="card-footer d-flex">
+            <div class="card-footer d-flex align-items-center">
                 <div class="lcr-left">
                     <div data-action-set="reviewing">
                         <button type="submit" name="action" value="unreview" class="btn btn-sm btn-outline-secondary">
@@ -155,8 +176,10 @@
                 <div class="lcr-right">
                     {% if user.is_manager %}
                         <div data-action-set="reviewing">
+                            {% include "staff_evaluation_textanswer_flag_radios.html" with form="shared-full-textanswer-flag-form" id="shared" %}
+
                             <button type="submit" name="action" value="textanswer_edit" class="btn btn-sm btn-outline-secondary">
-                                <span class="fas fa-pencil">
+                                <span class="fas fa-pencil"></span>
                             </button>
                         </div>
                     {% endif %}
@@ -173,6 +196,7 @@
         const slider = new QuickReviewSlider(
             document.querySelector(".slider"),
             document.querySelector("#textanswer-update-form"),
+            document.querySelector("#shared-full-textanswer-flag-form"),
             "{% url 'staff:evaluation_textanswers_skip' %}",
         );
         slider.attach();
--- a/evap/staff/templates/staff_evaluation_textanswers_section.html
+++ b/evap/staff/templates/staff_evaluation_textanswers_section.html
@@ -11,9 +11,10 @@
                 <table class="table table-striped{% if not forloop.last %} mb-4{% endif %}">
                     <thead>
                         <tr>
-                            <th style="width: 80%">{% trans 'Text answer' %}</th>
+                            <th style="width: 75%">{% trans 'Text answer' %}</th>
                             <th style="width: 5%"></th>
                             <th style="width: 15%">{% trans 'Decision' %}</th>
+                            <th style="width: 5%">{% trans 'Flag' %}</th>
                         </tr>
                     </thead>
                     <tbody>
@@ -50,6 +51,15 @@
                                             <input type="radio" class="btn-check" name="action" value="delete" id="{{ answer.id }}-radio-delete" autocomplete="off" {% if answer.will_be_deleted %}checked{% endif %}>
                                             <label class="btn btn-outline-primary" for="{{ answer.id }}-radio-delete">{% trans 'Delete' %}</label>
                                         </div>
+                                    </form>
+                                </td>
+                                <td class="text-center">
+                                    <form class="full-textanswer-flag-form" method="POST" action="{% url 'staff:evaluation_textanswers_update_flag' %}">
+                                        {% csrf_token %}
+
+                                        <input type="hidden" name="answer_id" value="{{ answer.id }}" />
+
+                                        {% include "staff_evaluation_textanswer_flag_radios.html" with is_initially_flagged=answer.is_flagged id=answer.id %}
                                     </form>
                                 </td>
                             </tr>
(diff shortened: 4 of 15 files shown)


janno42:
The "undecided" view should not show flagged decided answers.

niklasmohrin:
It would be nice to have a frontend test for this, but it is impossible with our current test setup as we are submitting forms here :(
We should look into having a server running during tests and what benefits and disadvantages arise from it

evap/static/scss/components/icon-checkbox.scss
```diff
@@ -0,0 +1,12 @@
+.icon-checkbox input[type="checkbox"] {
+    display: none;
+
+    // TODO: transition? (doesn't work on `display`)
```
janno42, line 4: we probably won't use a transition when having different icons, do we?

janno42 (changes requested):

evap/staff/templates/staff_evaluation_textanswers_section.html
```diff
@@ -22,8 +23,8 @@
                                 <td class="text-answer">
                                     {{ answer.answer|linebreaksbr }}
                                     {% if answer.original_answer %}
-                                        <br />
-                                        <span class="textanswer-original">({{ answer.original_answer|linebreaksbr }})</span>
+                                    <br />
+                                    <span class="textanswer-original">({{ answer.original_answer|linebreaksbr }})</span>
```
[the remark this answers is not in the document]
niklasmohrin: hm, dont know what happened here

evap/staff/templates/staff_evaluation_textanswers_full.html
```diff
@@ -37,7 +37,11 @@
 {% block additional_javascript %}
     <script type="text/javascript">
         document
-            .querySelectorAll(".full-update-textanswer-form input[type='radio']")
-            .forEach(radio => radio.addEventListener("change", () => radio.form.submit()));
+            .querySelectorAll(".full-update-textanswer-form input[type='radio'], .full-textanswer-flag-form input[type='radio']")
+            .forEach(input => input.addEventListener("change", () => {
+                input.form.submit();
+                input.disabled = true;
+                setTimeout(() => input.disabled = false, 500);
```
He3lixxx, line 44: Out of curiosity: What happens here on a network error? I'd like it if the input button only was re-enabled once we knew that the server processed our request, but I guess that's a bit hard to do here?

evap/staff/templates/staff_evaluation_textanswers_quick.html
```diff
@@ -155,6 +176,34 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
                 <div class="lcr-right">
                     {% if user.is_manager %}
                         <div data-action-set="reviewing">
+                            <div class="btn-group icon-buttons" role="group">
+                                <!-- Note: The `autocomplete` is not standard compliant, but in practice disables caching -->
```
He3lixxx, line 180: What is the problem case here? The browser checking both checkboxes?

Why do we have two checkboxes? You probably have a good reason, I'd guess we want to have that documented in comment or so.

evap/staff/templates/staff_evaluation_textanswers_section.html
```diff
@@ -52,6 +53,48 @@
                                         </div>
                                     </form>
                                 </td>
+                                <td class="text-center">
+                                    <form class="full-textanswer-flag-form" method="POST" action="{% url 'staff:evaluation_textanswers_update_flag' %}">
+                                        {% csrf_token %}
+
+                                        <input type="hidden" name="answer_id" value="{{ answer.id }}" />
+
+                                        <div class="btn-group icon-buttons" role="group">
```
He3lixxx, line 62: Can we move the (almost) duplicated code to a shared template fragment?

evap/staff/views.py
```diff
@@ -1571,6 +1581,22 @@ def evaluation_textanswers_update_publish(request):
...
+@reviewer_required
+def evaluation_textanswers_update_flag(request):
+    answer = get_object_from_dict_pk_entry_or_logged_40x(TextAnswer, request.POST, "answer_id")
+    assert_textanswer_review_permissions(answer.contribution.evaluation)
+    updated_is_flagged = request.POST.get("is_flagged", None)
+    if updated_is_flagged not in ["true", "false"]:
+        raise SuspiciousOperation
+    updated_is_flagged = updated_is_flagged == "true"
+
+    answer.is_flagged = updated_is_flagged
```
He3lixxx, line 1594:
```suggestion
    answer.is_flagged = updated_is_flagged == "true"
```
or am I missing something?

evap/staff/templates/staff_evaluation_textanswers_full.html
niklasmohrin: yes we could, but that would require submitting through javascript `fetch` again I think

evap/staff/templates/staff_evaluation_textanswers_quick.html
niklasmohrin: I had it for general sanity. This is all firefox things, they are caching the state (not spec compliant) but have this `autocomplete` to disable it (not spec compliant). Other browsers should just ignore the attribute I guess.

The horror scenario in my head is for quick view where the input is shared across answers. The state _should_ be managed by our javascript, but if anything is ever wrong, I don't want to think about mozilla's shenanigans that day

niklasmohrin: see changes

evap/staff/templates/staff_evaluation_textanswers_full.html
He3lixxx: So what happens currently? Is the error simply ignored.

evap/staff/templates/staff_evaluation_textanswer_flag_radios.html
```diff
@@ -0,0 +1,40 @@
...
+    -->
+    <input
+        {% if form %} form={{ form }} {% endif %}
+        type="radio"
+        name="is_flagged"
+        value="false"
+        id="radio-flag-false-{{ id_suffix }}"
+        class="btn-check"
+        autocomplete="off"
+        {% if not is_initially_flagged %} checked {% endif %}
```
He3lixxx, line 15:
```suggestion
        {% if not is_initially_flagged %}checked{% endif %}
```

evap/staff/templates/staff_evaluation_textanswer_flag_radios.html
```diff
@@ -0,0 +1,40 @@
...
+
+    <input
+        {% if form %} form={{ form }} {% endif %}
+        type="radio"
+        name="is_flagged"
+        value="true"
+        id="radio-flag-true-{{ id_suffix }}"
+        class="btn-check"
+        autocomplete="off"
+        {% if is_initially_flagged %} checked {% endif %}
```
He3lixxx, line 32:
```suggestion
        {% if is_initially_flagged %}checked{% endif %}
```

evap/staff/views.py
```diff
@@ -1586,12 +1586,12 @@ def evaluation_textanswers_update_publish(request):
 def evaluation_textanswers_update_flag(request):
     answer = get_object_from_dict_pk_entry_or_logged_40x(TextAnswer, request.POST, "answer_id")
     assert_textanswer_review_permissions(answer.contribution.evaluation)
-    updated_is_flagged = request.POST.get("is_flagged", None)
-    if updated_is_flagged not in ["true", "false"]:
+
+    is_flagged_bool_string = request.POST.get("is_flagged", None)
+    if is_flagged_bool_string not in ["true", "false"]:
         raise SuspiciousOperation
```
He3lixxx, line 1592: codecov says this is not covered, but I feel like I remember seeing a test for invalid inputs here?

evap/staff/templates/staff_evaluation_textanswers_full.html
niklasmohrin: Browser should just timeout and display a sad face, as if one would just submit a form and it never responds

niklasmohrin: If the network is just slow, it doesn't matter if the form is submitted again, because it will just write the same value again

niklasmohrin: But I can agree that it is a bit icky

evap/staff/views.py
niklasmohrin: yep, was broken

evap/staff/templates/staff_evaluation_textanswers_full.html
He3lixxx: Alright, works for me

janno42, line 44: Please remove the timeout

evap/staff/templates/staff_evaluation_textanswer_flag_radios.html
```diff
@@ -0,0 +1,40 @@
...
+    <input
+        {% if form %}form={{ form }}{% endif %}
+        type="radio"
+        name="is_flagged"
+        value="false"
+        id="radio-flag-false-{{ id_suffix }}"
+        class="btn-check"
+        autocomplete="off"
+        {% if not is_initially_flagged %}checked{% endif %}
+    >
```
janno42:
```suggestion
    />
```

evap/staff/templates/staff_evaluation_textanswer_flag_radios.html
```diff
@@ -0,0 +1,40 @@
...
+    <input
+        {% if form %}form={{ form }}{% endif %}
+        type="radio"
+        name="is_flagged"
+        value="true"
+        id="radio-flag-true-{{ id_suffix }}"
+        class="btn-check"
+        autocomplete="off"
+        {% if is_initially_flagged %}checked{% endif %}
+    >
```
janno42:
```suggestion
    />
```

janno42 (changes requested):

evap/staff/templates/staff_evaluation_textanswers_section.html
```diff
@@ -52,6 +53,15 @@
                                         </div>
                                     </form>
                                 </td>
+                                <td class="text-center">
+                                    <form class="full-textanswer-flag-form" method="POST" action="{% url 'staff:evaluation_textanswers_update_flag' %}">
+                                        {% csrf_token %}
+
+                                        <input type="hidden" name="answer_id" value="{{ answer.id }}" />
+
+                                        {% include "staff_evaluation_textanswer_flag_radios.html" with is_initially_flagged=answer.is_flagged id_suffix=answer.id %}
```
He3lixxx, line 62: Maybe an interface consideration: `id_suffix`: The suffix part is kind of an implementation detail, I'd say. For the API, I'd just make it `id` or `textanswer_id` or similar.

evap/staff/tests/test_views.py
```diff
@@ -3090,6 +3090,92 @@ def test_archived(self):
...
+            self.assertFalse(self.answer.is_flagged)
+
+            self.app.post(
+                self.url,
+                user=self.manager,
+                status=400,
+                params={"answer_id": self.answer.pk, "is_flagged": "True"},
+            )
+            self.answer.refresh_from_db()
+            self.assertFalse(self.answer.is_flagged)
```
He3lixxx, line 3149: The blocks here could be a loop of `[("True", False), ("False", False), ("", False), ("true", True)]`, which I think I'd prefer to reduce duplication.

evap/staff/views.py
```diff
@@ -1430,10 +1430,15 @@ def evaluation_login_key_export(_request, semester_id, evaluation_id):
...
-        "TextAnswerSection", ("questionnaire", "contributor", "label", "is_responsible", "results")
-    )
+TextAnswerSection = namedtuple(
+    "TextAnswerSection", ("questionnaire", "contributor", "label", "is_responsible", "results")
+)
+
+
+def get_evaluation_and_contributor_textanswer_sections(
+    evaluation: Evaluation,
+    textanswer_filter: Optional[Q],
```
He3lixxx, line 1440: Could we just default to an always-true object? (Idk how `Q()` evaluates).

He3lixxx (approved):
Code looks good to me. Haven't tested manually.

janno42 (changes requested):
Looks promising so far, but there is something left to do:
- The issue requested an overview page of all flagged textanswers in a semester. I pushed a proposal for that in https://github.com/e-valuation/EvaP/commit/223ea291fbe7446376aa0b60cd6c3cdc2da3da95 :)  It might need some polishing and testing though.
- The suggestions for the next evaluation in quick review need to be adapted (see https://github.com/e-valuation/EvaP/issues/1118#issuecomment-891029989)

niklasmohrin:
The "is flagged"-radios are broken right now, not sure why. When hovering over it, the border turns white / practically disappears

janno42 (commented):
The problem is the `btn-check` class on the radios. `.btn-check + .btn:hover` comes before `.btn:hover` (it's more specific), so the resulting styling is not as we have it at other places where we're using `role="button"`.
You could adapt our `_buttons.scss` to redefine the colors also for `.btn-check`.

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

It's awesome.

evap/staff/views.py
```diff
@@ -1453,11 +1453,15 @@ def evaluation_login_key_export(_request, evaluation_id):
...
-        "TextAnswerSection", ("questionnaire", "contributor", "label", "is_responsible", "results")
-    )
+TextAnswerSection = namedtuple(
+    "TextAnswerSection", ("questionnaire", "contributor", "label", "is_responsible", "results")
+)
+
 
+def get_evaluation_and_contributor_textanswer_sections(
+    evaluation: Evaluation,
+    textanswer_filter: Q,
```
He3lixxx, line 1463: `optional[Q]`? Below the code checks for `if textanswer_filter is not None`

He3lixxx (approved):

evap/staff/views.py
niklasmohrin: ah, actually the check was not needed - weird that mypy / pylint didnt catch it :thinking:

PR #1821, opened 2022-10-10
Importer messages: More intuitive counts / numbers 2

### LINKED ISSUE #1590: Importer message refactoring

FSadrieh:
Fixes #1590


evap/staff/importers/enrollment.py
```diff
@@ -34,6 +35,10 @@
 )
 
 
+def make_count_word(count):
```
niklasmohrin, line 38: Let's add type hints here :)

evap/staff/importers/enrollment.py
```diff
@@ -608,12 +617,23 @@ def import_enrollments(
...
+                ).format(count=make_count_word(new_course_count)),
+                participant_string=ngettext("one participant", "{count} participants", new_participants_count).format(
+                    count=make_count_word(new_participants_count)
+                ),
+                contributor_string=ngettext("one contributor", "{count} contributors", new_responsibles_count).format(
+                    count=make_count_word(new_responsibles_count)
+                ),
             )
+
+            if new_participants_count or new_responsibles_count:
```
niklasmohrin, line 632: I like the idea, but we should also add a "." if there is no list. And we also didn't do this in the other change above

evap/staff/importers/person.py
```diff
@@ -19,21 +20,41 @@ def add_participants_to(
...
 
-    if not test_run:
-        evaluation.participants.add(*users_to_add)
-        msg = format_html(_("{} participants added to the evaluation {}:"), len(users_to_add), evaluation.full_name)
+    if len(users_to_add) == 0:
+        if test_run:
+            msg = format_html(_("No participants would be added to the evaluation {}"), evaluation.full_name)
+        else:
+            msg = format_html(_("No participants added to the evaluation {}"), evaluation.full_name)
+
```
niklasmohrin, line 39:
```suggestion
```

niklasmohrin (changes requested):
You are still missing some strings, for example in `importers/user.py`

He3lixxx (commented):
@janno42 please check whether the UI advantage justifies the code overhead.

@FSadrieh Have you looked at the additional messages that @niklasmohrin referenced?

janno42 (changes requested):

evap/staff/importers/enrollment.py
```diff
@@ -608,12 +621,25 @@ def import_enrollments(
...
+                ),
+                contributor_string=ngettext("1 contributor", "{count} contributors", new_responsibles_count).format(
+                    count=str(new_responsibles_count)
+                ),
             )
+
+            if new_participants_count or new_responsibles_count:
+                msg = format_html(
+                    "{msg}: {list}", msg=msg, list=create_user_list_html_string_for_message(new_user_profiles)
+                )
```
He3lixxx, line 639: This looks like a copy&paste error: different variables in the if-condition vs in the the `create_user_list` call

evap/staff/importers/person.py
```diff
@@ -19,21 +20,40 @@ def add_participants_to(
 
     if already_related:
         msg = format_html(
-            _("The following {} users are already participants in evaluation {}:"),
-            len(already_related),
-            evaluation.full_name,
+            "{sentence}: {list}",
+            sentence=ngettext(
+                "The following user is already participating in evaluation {name}",
+                "The following {user_count} are already participating in evaluation {name}",
```
He3lixxx, line 26:
```suggestion
                "The following {user_count} users are already participating in evaluation {name}",
```

evap/staff/importers/person.py
```diff
@@ -19,21 +20,40 @@ def add_participants_to(
...
         importer_log.add_warning(msg)
 
-    if not test_run:
-        evaluation.participants.add(*users_to_add)
-        msg = format_html(_("{} participants added to the evaluation {}:"), len(users_to_add), evaluation.full_name)
+    if len(users_to_add) == 0:
+        if test_run:
+            msg = format_html(_("0 participants would be added to the evaluation {}."), evaluation.full_name)
+        else:
+            msg = format_html(_("0 participants added to the evaluation {}."), evaluation.full_name)
```
He3lixxx, line 38: Doesn't need special handling

evap/staff/importers/person.py
```diff
@@ -45,25 +65,43 @@ def add_contributors_to(
...
-    if not test_run:
-        for user in users_to_add:
-            order = Contribution.objects.filter(evaluation=evaluation).count()
-            Contribution.objects.create(evaluation=evaluation, contributor=user, order=order)
-        msg = format_html(_("{} contributors added to the evaluation {}:"), len(users_to_add), evaluation.full_name)
+    if len(users_to_add) == 0:
+        if test_run:
+            msg = format_html(_("0 contributors would be added to the evaluation {}."), evaluation.full_name)
+        else:
+            msg = format_html(_("0 contributors added to the evaluation {}."), evaluation.full_name)
```
He3lixxx, line 84: Doesn't need special handling

He3lixxx (changes requested):

He3lixxx (changes requested):
No changes since the last review.

niklasmohrin (commented):
tests look very good, some final remarks:

evap/staff/importers/person.py
```diff
@@ -45,11 +68,14 @@ def add_contributors_to(
...
-            _("The following {} users are already contributing to evaluation {}:"),
-            len(already_related),
-            evaluation.full_name,
+            "{sentence}: {list}",
+            sentence=ngettext(
+                "The following user is already contributing to evaluation {name}",
+                "The following {user_count} users are already contributing to evaluation {name}",
+                len(already_related),
+            ).format(user_count=len(already_related), name=evaluation.full_name),
+            list=create_user_list_html_string_for_message(already_related),
```
niklasmohrin, line 77: and here?

evap/staff/importers/enrollment.py
```diff
@@ -594,12 +595,15 @@ def import_enrollments(
...
             importer_log.add_success(_("The test run showed no errors. No data was imported yet."))
             msg = format_html(
-                _("The import run will create {evaluation_count} courses/evaluations and {user_count} users:{users}"),
-                evaluation_count=new_course_count,
-                user_count=len(new_user_profiles),
-                users=create_user_list_html_string_for_message(new_user_profiles),
+                _("The import run will create {evaluation_string} and {user_string}"),
+                evaluation_string=ngettext(
+                    "1 course/evaluation", "{count} courses/evaluations", new_course_count
+                ).format(count=str(new_course_count)),
```
niklasmohrin, line 601: I don't think you need the explicit `str(...)`, the formatting should do it for you (also below)

evap/staff/importers/person.py
```diff
@@ -1,10 +1,8 @@
+from gettext import ngettext
```
niklasmohrin, line 1: Please import `ngettext` from `django.utils.translation` instead

evap/staff/importers/person.py
```diff
@@ -58,12 +65,21 @@ def add_contributors_to(
         for user in users_to_add:
             order = Contribution.objects.filter(evaluation=evaluation).count()
             Contribution.objects.create(evaluation=evaluation, contributor=user, order=order)
-        msg = format_html(_("{} contributors added to the evaluation {}:"), len(users_to_add), evaluation.full_name)
+        message = ngettext(
+            "1 contributor added to the evaluation {name}",
+            "{user_count} contributors added to the evaluation {name}",
+            len(users_to_add),
+        ).format(user_count=len(users_to_add), name=evaluation.full_name)
+
```
niklasmohrin, line 73:
```suggestion
```

niklasmohrin (changes requested):

evap/staff/importers/enrollment.py
```diff
@@ -594,12 +595,15 @@ def import_enrollments(
         if test_run:
             importer_log.add_success(_("The test run showed no errors. No data was imported yet."))
             msg = format_html(
```
He3lixxx, line 597: This doesn't need to use `format_html`, right? Then, we shouldn't use it. Repeated below.

evap/staff/tools.py
```diff
@@ -76,6 +78,16 @@ def create_user_list_html_string_for_message(users):
     return format_html_join("", "<br />{} {} ({})", ((user.first_name, user.last_name, user.email) for user in users))
 
 
+def append_user_list(message: str, user_profiles: Collection) -> SafeString:
```
He3lixxx, line 81: Maybe name it `append_user_list_if_not_empty` to make clear that an empty list will just end the sentence, which could lead to possibly confusing messages like:
`The following 0 users will be added to the evaluation.` (without an empty list)

He3lixxx (commented):
Looking quite good! :)

evap/staff/tools.py
```diff
@@ -65,6 +66,15 @@ def create_user_list_html_string_for_message(users):
     return format_html_join("", "<br />{} {} ({})", ((user.first_name, user.last_name, user.email) for user in users))
 
 
+def append_user_list_if_not_empty(message: str, user_profiles: Collection) -> str:
+    if not user_profiles:
+        return _("{message}.").format(message=message)
+    return _("{message}: {list}").format(
+        message=message,
+        list=create_user_list_html_string_for_message(user_profiles),
+    )
```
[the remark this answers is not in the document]
niklasmohrin: Richard and I talked about this some more and we came to the conclusion that we generally don't actually always want `format_html` if there is nothing in the format string that has to explicitly not-escaped. So instead, a concatenation of `SafeString` is best. Further, we can ensure that we are not loosing the escaping on the way by using type annotations:
```python
from django.utils.safestring import SafeString
from django.utils.html import conditional_escape, escape


def append_user_list_if_not_empty(message: str, user_profiles: Collection) -> SafeString:
    message = conditional_escape(message)
    assert isinstance(message, SafeString)
    if not user_profiles:
        return message + escape(".")
    return message + escape(": ") + create_user_list_html_string_for_message(user_profiles)
```
Please make sure to also annotate `create_user_list_html_string_for_message` so that it returns `SafeString` (and takes `Iterable[UserProfile]`)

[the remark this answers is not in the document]
niklasmohrin: And actually, we came up with another thing: Django's `conditional_escape` is annotated to only return `str`, not `SafeString`. This stems from the fact that Django tries to be compatible with other escaping libraries. However, we only care about `str` and `SafeString`. Let's write our own `conditional_escape` in `evaluation/tools.py` like so:
```python
def conditional_escape(s: str) -> SafeString:
    """
    Like Django's `conditional_escape`, but only handles `str` and `SafeString`.
    """
    if isinstance(s, SafeString):
        return s
    return escape(s)
```
Using this, we can remove the `assert` from the code in my last comment

niklasmohrin (changes requested):

niklasmohrin (approved):
Looks very good now! :)

You forgot about this earlier note from me:

> Please make sure to also annotate `create_user_list_html_string_for_message` so that it returns `SafeString` (and takes `Iterable[UserProfile]`)

Finally, it would be nice to add a small test for `conditional_escape` for coverage. You could call it with the string `<b>bold</b>`, once like that and once already `escape`d :)

evap/staff/fixtures/excel_files_test_data.py
```diff
@@ -133,6 +133,15 @@
...
 
+test_enrollment_data_empty_filedata = {
+    'MA Belegungen': [
+        ['Degree', 'Student last name', 'Student first name', 'Student email address', 'Course kind', 'Course is graded', 'Course name (de)', 'Course name (en)', 'Responsible title', 'Responsible last name', 'Responsible first name', 'Responsible email address'],
+    ],
+    'BA Belegungen': [
+        ['Degree', 'Student last name', 'Student first name', 'Student email address', 'Course kind', 'Course is graded', 'Course name (de)', 'Course name (en)', 'Responsible title', 'Responsible last name', 'Responsible first name', 'Responsible email address'],
+    ]
+}
+
```
He3lixxx, line 144: @niklasmohrin For maintainability, we should maybe consider reducing the count of "excel file data" struct we have here, and instead maybe rely on a single fixed "correct" input data set that we modify to the needs of each individual test (e.g. inserting duplicate rows, this here could be just removing all rows except for the headers).

evap/staff/tools.py
```diff
@@ -65,6 +67,22 @@ def create_user_list_html_string_for_message(users):
...
+    message = conditional_escape(message)
+    if not user_profiles:
+        return message + escape(".")
+    return message + escape(": ") + create_user_list_html_string_for_message(user_profiles)
+
+
+def conditional_escape(s: str) -> SafeString:
+    """
+    Like Django's `conditional_escape`, but only handles `str` and `SafeString`.
+    """
```
He3lixxx, line 80: I don't get it. Should it just say "`str`" instead of "`str` and `SafeString`"?

evap/staff/tools.py
```diff
@@ -65,6 +67,22 @@ def create_user_list_html_string_for_message(users):
...
+    if isinstance(s, SafeString):
+        return s
```
He3lixxx, line 82: Please add a test for this.

evap/staff/tests/test_importers.py
```diff
@@ -368,12 +369,27 @@ def test_emails_are_cleaned(self):
         self.assertTrue(UserProfile.objects.filter(email="cleaned_diam.synephebos@institution.example.com").exists())
         self.assertTrue(UserProfile.objects.filter(email="cleaned_111@institution.example.com").exists())
 
+    def test_import_with_empty_excel(self):
+        importer_log = import_enrollments(self.empty_excel_content, self.semester, None, None, test_run=True)
+
+        success_messages = [msg.message for msg in importer_log.success_messages()]
+        self.assertIn("The import run will create 0 courses/evaluations and 0 users.", success_messages)
```
He3lixxx, line 376: Please also assert that there were no errors for both assertions in this test (`self.assertEqual(importer_log.errors_by_category(), {})`). In #1855 I came across some tests that still pass, even if the importer has a critical failure, simply because the message was emitted before the fatal failure. I'd like the tests to catch that.

He3lixxx (commented):
Looking good, looking good. I'd be really happy if you could do these 3 things :)

evap/staff/tools.py
niklasmohrin: Ah, the wording could be more precise to better suggest what the difference is (django also wants to be compatible with other types that claim to be safe). I suppose it could be "distincts" instead of "handles" or so. Or another sentence, don't care

evap/staff/fixtures/excel_files_test_data.py
niklasmohrin: I am fine with whatever is simpler. I was wondering if we should try to not generate test data in the tests and instead have functions here that have a name to suggest what kind of modified data they return, but I am not sure if that would lead to too much test specifics bleeding into this one file all shuffled. I suppose there is only one way to find out 😅

I agree to not try it in this PR

janno42 (changes requested):
In the meantime we had some other importer changes. There is at least the message "... participants from the import file already participate in the evaluation." (https://github.com/e-valuation/EvaP/blob/main/evap/staff/importers/enrollment.py#L551) that needs to be updated and maybe also some more.

He3lixxx (approved):
Thanks a lot :-)

evap/staff/importers/enrollment.py
```diff
@@ -548,8 +548,10 @@ def finalize(self) -> None:
 
             if colliding_participant_emails:
                 self.importer_log.add_warning(
-                    _(
-                        "Course {course_name}: {participant_count} participants from the import file already participate in the evaluation."
+                    ngettext(
+                        "Course {course_name}: 1 participant from the import file already participate in the evaluation.",
```
janno42, line 552:
```suggestion
                        "Course {course_name}: 1 participant from the import file already participates in the evaluation.",
```

evap/staff/tests/test_importers.py
```diff
@@ -813,7 +813,7 @@ def test_existing_participation(self):
 
         importer_log = import_enrollments(self.default_excel_content, self.semester, None, None, test_run=True)
 
-        expected_warnings = ["Course Shake: 1 participants from the import file already participate in the evaluation."]
+        expected_warnings = ["Course Shake: 1 participant from the import file already participate in the evaluation."]
```
janno42, line 816:
```suggestion
        expected_warnings = ["Course Shake: 1 participant from the import file already participates in the evaluation."]
```

janno42 (changes requested):
Please also check `staff/importers/user.py` lines 348 and 356

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

He3lixxx:
🎉

PR #1822, opened 2022-10-17
Fix #1804 - Bad request on failed redemption

### LINKED ISSUE #1804: Response status code of failed redemption is 200
As @niklasmohrin remarked in [#1790](https://github.com/e-valuation/EvaP/pull/1790/files#r962983692), in `evap.rewards.views.redeem_reward_points`, the status code of failed redemptions (e.g. due to `NotEnoughPoints` or `RedemptionEventExpired`) is set as 200 OK, even though no redemption points were saved. 

Instead, the status code should be something like 400 Bad Request to underline that something went wrong.
@niklasmohrin added, that `assertContains`, used in some tests in `evap.rewards.tests.test_views.TestIndexView`, needs to adopted, as it asserts that the status code is 200 by default.

florian-str:
Fix for issue #1804

### DIFF
--- a/evap/rewards/tests/test_views.py
+++ b/evap/rewards/tests/test_views.py
@@ -57,7 +57,7 @@ class TestIndexView(WebTest):
         form.set(f"points-{self.event1.pk}", 2)
         form.set(f"points-{self.event2.pk}", 3)
         response = form.submit()
-        self.assertContains(response, "You successfully redeemed your points.")
+        self.assertContains(response, "You successfully redeemed your points.", status_code=200)
         self.assertEqual(0, reward_points_of_user(self.student))
 
     def test_redeem_too_many_points(self):
@@ -65,8 +65,8 @@ class TestIndexView(WebTest):
         form = response.forms["reward-redemption-form"]
         form.set(f"points-{self.event1.pk}", 3)
         form.set(f"points-{self.event2.pk}", 3)
-        response = form.submit()
-        self.assertContains(response, "have enough reward points.")
+        response = form.submit(expect_errors=True)
+        self.assertContains(response, "have enough reward points.", status_code=400)
         self.assertEqual(5, reward_points_of_user(self.student))
 
     def test_redeem_points_for_expired_event(self):
@@ -75,8 +75,8 @@ class TestIndexView(WebTest):
         form = response.forms["reward-redemption-form"]
         form.set(f"points-{self.event2.pk}", 1)
         RewardPointRedemptionEvent.objects.update(redeem_end_date=date.today() - timedelta(days=1))
-        response = form.submit()
-        self.assertContains(response, "event expired already.")
+        response = form.submit(expect_errors=True)
+        self.assertContains(response, "event expired already.", status_code=400)
         self.assertEqual(5, reward_points_of_user(self.student))
 
     def test_invalid_post_parameters(self):
--- a/evap/rewards/tests/test_views.py
+++ b/evap/rewards/tests/test_views.py
@@ -57,25 +57,27 @@ class TestIndexView(WebTest):
         form.set(f"points-{self.event1.pk}", 2)
         form.set(f"points-{self.event2.pk}", 3)
         response = form.submit()
-        self.assertContains(response, "You successfully redeemed your points.", status_code=200)
+        self.assertContains(response, "You successfully redeemed your points.")
         self.assertEqual(0, reward_points_of_user(self.student))
 
+    # pylint: disable=too-many-locals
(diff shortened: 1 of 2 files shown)


evap/rewards/tests/test_views.py
```diff
@@ -57,16 +57,16 @@ def test_redeem_all_points(self):
         form.set(f"points-{self.event1.pk}", 2)
         form.set(f"points-{self.event2.pk}", 3)
         response = form.submit()
-        self.assertContains(response, "You successfully redeemed your points.")
+        self.assertContains(response, "You successfully redeemed your points.", status_code=200)
```
He3lixxx, line 60: I'd expect that `form.submit()` already expects a 200 status code and fails otherwise, since this is convention with the test framework (however, I have to say, it does not seem to be documented at all, I could only find a reference in the actual code [here](https://github.com/Pylons/webtest/blob/561ef780f695c3a602be7b49486eabde0832373c/webtest/app.py#L675-L681)).

evap/rewards/tests/test_views.py
```diff
@@ -57,16 +57,16 @@ def test_redeem_all_points(self):
...
 
     def test_redeem_too_many_points(self):
         response = self.app.get(self.url, user=self.student)
         form = response.forms["reward-redemption-form"]
         form.set(f"points-{self.event1.pk}", 3)
         form.set(f"points-{self.event2.pk}", 3)
-        response = form.submit()
-        self.assertContains(response, "have enough reward points.")
+        response = form.submit(expect_errors=True)
+        self.assertContains(response, "have enough reward points.", status_code=400)
```
He3lixxx, line 69: Have you tried this?
```suggestion
        response = form.submit(status=400)
        self.assertContains(response, "have enough reward points.")
```

I think this would be the most canonical code here.

(repeated below)

He3lixxx (approved):
Thanks a lot! The view code looks very good. For the tests, we think we can use the `status` argument on all requests that we issue, so we only have the expected error in one line instead of two lines.

niklasmohrin:
About the linter error: In the original PR where we noticed this issue, we moved the code into a new function. As this PR is still open, I am fine with ignoring the warning for this PR as #1790 will fix it again.

To silence the warning, add a line with `# pylint: disable=too-many-locals` at the top of the function.

evap/rewards/tests/test_views.py
florian-str: While the parameter `status=400` stops the failure on submit, it still has to be set in `assertContains`, as it also expects a status code of 200.

He3lixxx: Fair enough, the code is fine then :)

evap/rewards/tests/test_views.py
```diff
@@ -60,23 +60,25 @@ def test_redeem_all_points(self):
         self.assertContains(response, "You successfully redeemed your points.")
         self.assertEqual(0, reward_points_of_user(self.student))
 
+    # pylint: disable=too-many-locals
```
He3lixxx, line 63:
```suggestion
```

evap/rewards/tests/test_views.py
```diff
@@ -60,23 +60,25 @@ def test_redeem_all_points(self):
...
         form = response.forms["reward-redemption-form"]
         form.set(f"points-{self.event1.pk}", 3)
         form.set(f"points-{self.event2.pk}", 3)
-        response = form.submit()
-        self.assertContains(response, "have enough reward points.")
+        response = form.submit(status=400)
+        self.assertContains(response, "have enough reward points.", status_code=400)
         self.assertEqual(5, reward_points_of_user(self.student))
 
+    # pylint: disable=too-many-locals
```
He3lixxx, line 73:
```suggestion
```

evap/rewards/views.py
```diff
@@ -28,6 +28,7 @@
 
 @reward_user_required
 def index(request):
+    status = 200
```
He3lixxx, line 31:
```suggestion
def index(request):
    # pylint: disable=too-many-locals
    status = 200
```

He3lixxx (commented):
Our linter (pylint) performs a static analysis of the code without any call-chain inference. We thus have to disable the check in the scope where it fails otherwise (because there are too many locals according to our rules). If you move the `# pylint: disable=too-many-locals` comment into the rewards/views.py `index` view, as shown in the change suggestions below, it should work.

Don't worry about the typescript tests, that failure was unrelated to your changes. We will have to fix that outside of this PR.

He3lixxx (approved):

niklasmohrin (approved):

PR #1828, opened 2022-10-24
Remove some redundant id parameters

niklasmohrin:
It is quite common that we have views that take an id for an evaluation
and a semester such that the evaluation must belong to the semester.
I believe that such requirements should be avoided, because we need to
check that they hold everywhere anyways (or fail to do so). The only
nice thing about this is that users can the /evaluation/foo suffix from
the url to get to the corresponding semester - but in a normal
interaction they can already do that using the breadcrumbs.

I noticed that we are actually missing on such check for the
textanswer-update view (and abused that ourselves in the tests). An evil
reviewer could change the review decision for a textanswer after the
evaluation is published (or similar). Consequently, I removed all other
parameters from urls containing textanswer ids.

### DIFF
--- a/evap/staff/views.py
+++ b/evap/staff/views.py
@@ -1527,7 +1527,7 @@ def evaluation_textanswers_skip(request):
     return HttpResponse()
 
 
-def check_textanswer_review_permissions(evaluation: Evaluation) -> None:
+def assert_textanswer_review_permissions(evaluation: Evaluation) -> None:
     if evaluation.state == Evaluation.State.PUBLISHED:
         raise PermissionDenied
     if evaluation.course.semester.results_are_archived:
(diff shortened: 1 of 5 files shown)


evap/staff/views.py
```diff
@@ -1527,22 +1527,26 @@ def evaluation_textanswers_skip(request):
...
 
 
-@require_POST
-@reviewer_required
-def evaluation_textanswers_update_publish(request):
-    answer = get_object_from_dict_pk_entry_or_logged_40x(TextAnswer, request.POST, "answer_id")
-    evaluation = get_object_from_dict_pk_entry_or_logged_40x(Evaluation, request.POST, "evaluation_id")
-    action = request.POST.get("action", None)
-
+def check_textanswer_review_permissions(evaluation: Evaluation) -> None:
```
He3lixxx, line 1530: what do you think about `assert_textanswer_review_permissions`?

He3lixxx (approved):

Issue #1806, opened 2022-09-12
Importer issues


janno42:
- Collapsibles for import messages have no titles
- When clicking on one card header, all cards collapse
- A course with two different German names (and a unique English name) in the import file results in an "The data of course [...] differs [...]" error message. Instead, we would expect "The English name for course [...] is already used for another course in the import file." (as shown if a course has a unique German name but two different English names) or at least a hint which fields differ.
- Email addresses must be used in lower case everywhere, also for id checks such as in evap/evap/staff/importers/enrollment.py, line 698
- On import, a `duplicate key value violates unique constraint` error (`Key (evaluation_id, userprofile_id)=[...] already exists.`) occurs in `/evap/evap/staff/importers/enrollment.py, line 766, in store_participations_in_db`, possibly because a participation is included twice in the import file.
- After importing the enrollment data, the degree merge warnings are not sorted by evaluation. After testing, they seem to be sorted.

He3lixxx:
> After importing the enrollment data, the degree merge warnings are not sorted by evaluation. After testing, they seem to be sorted.

Do you mean this message?
> Course "{course_name}" already exists. The course will not be created, instead users are imported into the evaluation of the existing course and any additional degrees are added.

There doesn't seem to be explicit sorting for this in the code. I would have assumed that it simply gives out the messages in the order of their first appearance, since that's how they are created and all involved data structures probably keep the insertion order.

Just from looking at the code, the ordering shouldn't change between test run (-> our custom message rendering is used) and the import run (-> django's messages are used. In their source code, I couldn't see anything that I think would change the order, and I think they have an interest to keep the order).

From our code, however, it seems possible that the messages are not grouped by category anymore when shown as django messages (although I'd also expect this happens "by default" because we create our log entries in a category-by-category fashion).

He3lixxx:
> After importing the enrollment data, the degree merge warnings are not sorted by evaluation. After testing, they seem to be sorted.

We discussed this in person, and could not reproduce: The warnings seem to always be sorted by the order of the first occurrence in the import file. We agreed that this is fine for now.

PR #1829, opened 2022-11-07
Fix importers

He3lixxx:
An attempt to fix the stuff in #1806 -- let's see if it works out eventually.

(Possible) todos:
* [x] There is one todo left in the issue regarding message sorting after the test run vs after the import run. Discussion for this is in the issue.
* [ ] Do we want to refactor the message categories? Some of the shorter ones could be a bit more clear.
* [ ] Maybe annotate the tests as regression tests?


niklasmohrin (approved):
Code looks good, haven't verified that all requirements from issue are met, leaving it to @janno42 :)

janno42 (changes requested):
It works, thank you :)

One more thing: The error message `The data of course "ABC" differs from its data in the columns ('name_de',) in a previous row.` has an unnecessary comma in the list.

PR #1832, opened 2022-11-07
Fix for #1813: Refactor error messages for rewards

### LINKED ISSUE #1813: Separate error messages and error detection in reward points redemption
A reward point redemption can fail for several reasons, each of which has its own `Exception` subclass in `evap/rewards/models.py`. These exceptions are created deep in the reward point redemption logic. Currently, error messages that ultimately end up on the user's screen are set at the creation of the exceptions. In order to separate the "error display" and "error detection" logic, the user strings should be moved to where the exceptions are caught.

_Originally came up in https://github.com/e-valuation/EvaP/pull/1790#pullrequestreview-1105770573_

florian-str:
This fixes #1813.

### DIFF
--- a/evap/rewards/views.py
+++ b/evap/rewards/views.py
@@ -52,7 +52,9 @@ def redeem_reward_points(request):
             error_string = _("Sorry, the deadline for this event expired already.")
         elif isinstance(error, OutdatedRedemptionData):
             status_code = 409
-            error_string = _("It appears that your browser sent multiple redemption requests. You can see all successful redemptions below.")
+            error_string = _(
+                "It appears that your browser sent multiple redemption requests. You can see all successful redemptions below."
+            )
         messages.warning(request, error_string)
         return status_code
     return 200
--- a/evap/rewards/views.py
+++ b/evap/rewards/views.py
@@ -42,7 +42,6 @@ def redeem_reward_points(request):
         save_redemptions(request, redemptions, previous_redeemed_points)
         messages.success(request, _("You successfully redeemed your points."))
     except (NoPointsSelected, NotEnoughPoints, RedemptionEventExpired, OutdatedRedemptionData) as error:
-        error_string = ""
         status_code = 400
         if isinstance(error, NoPointsSelected):
             error_string = _("You cannot redeem 0 points.")
--- a/evap/rewards/views.py
+++ b/evap/rewards/views.py
@@ -54,7 +54,7 @@ def redeem_reward_points(request):
             error_string = _(
                 "It appears that your browser sent multiple redemption requests. You can see all successful redemptions below."
             )
-        messages.warning(request, error_string)
+        messages.error(request, error_string)
         return status_code
     return 200
 
--- a/evap/rewards/views.py
+++ b/evap/rewards/views.py
@@ -41,12 +41,20 @@ def redeem_reward_points(request):
     try:
         save_redemptions(request, redemptions, previous_redeemed_points)
         messages.success(request, _("You successfully redeemed your points."))
-    except (NoPointsSelected, NotEnoughPoints, RedemptionEventExpired) as error:
-        messages.warning(request, error)
-        return 400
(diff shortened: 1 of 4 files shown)


evap/rewards/views.py
```diff
@@ -41,12 +41,20 @@ def redeem_reward_points(request):
...
+        if isinstance(error, NoPointsSelected):
+            error_string = "You cannot redeem 0 points."
+        elif isinstance(error, NotEnoughPoints):
+            error_string = "You don't have enough reward points."
+        elif isinstance(error, RedemptionEventExpired):
+            error_string = "Sorry, the deadline for this event expired already."
+        elif isinstance(error, OutdatedRedemptionData):
+            status_code = 409
+            error_string = "It appears that your browser sent multiple redemption requests. You can see all successful redemptions below."
+        messages.warning(request, _(error_string))
```
He3lixxx, line 56: In general, having a single translation call here would be a good thing. However, translation is always a bit tricky. In order to build our translation tables that store the pair <"You cannot redeem 0 points", "Du kannst nicht 0 Punkte einlösen">, we first need to find all strings in the codebase that we want to translate (-> "You cannot redeem 0 points"). Under the hood, django uses the gettext framework for this. In the end, it boils down to using a tool called [xgettext](https://www.gnu.org/software/gettext/manual/gettext.html#Template) which goes through the source code, looks for all function calls to all defined translation methods, and extracts the strings that the method is called with.

In order for that to work, the string needs to be extractable from the source code without actually running it, and thus, the method must directly be called with the string constant. So, in this case, we have to have one function call to `_` for each string literal above.

evap/rewards/views.py
```diff
@@ -41,12 +41,20 @@ def redeem_reward_points(request):
...
-        messages.warning(request, error)
-        return 400
-    except OutdatedRedemptionData as error:
-        messages.error(request, error)
-        return 409
+    except (NoPointsSelected, NotEnoughPoints, RedemptionEventExpired, OutdatedRedemptionData) as error:
+        error_string = ""
+        status_code = 400
+        if isinstance(error, NoPointsSelected):
+            error_string = "You cannot redeem 0 points."
```
He3lixxx, line 48: Codecov indicates that we have no test for this case. Could you add one? It should be fairly easy, check out `evap/rewards/tests/test_views.py` around line 63 for the tests of the other cases.

evap/rewards/views.py
He3lixxx, line 56: @janno42 should these be warnings? I feel like they should be errors.

He3lixxx (changes requested):

evap/rewards/views.py
```diff
@@ -41,12 +41,22 @@ def redeem_reward_points(request):
...
         save_redemptions(request, redemptions, previous_redeemed_points)
         messages.success(request, _("You successfully redeemed your points."))
-    except (NoPointsSelected, NotEnoughPoints, RedemptionEventExpired) as error:
-        messages.warning(request, error)
-        return 400
-    except OutdatedRedemptionData as error:
-        messages.error(request, error)
-        return 409
+    except (NoPointsSelected, NotEnoughPoints, RedemptionEventExpired, OutdatedRedemptionData) as error:
+        error_string = ""
```
niklasmohrin, line 45:
```suggestion
```

niklasmohrin (approved):
Very nice! One nit: we can remove the one line that sets `error_string` to its initial value; this will also make Python complain if we ever catch another kind of error but forget to set `error_string`

evap/rewards/views.py
janno42: Yes, they should be errors

niklasmohrin (changes requested):
Blocking for https://github.com/e-valuation/EvaP/pull/1832#discussion_r1028304449, looks good other than that

niklasmohrin (approved):

janno42:
Thank you 🎆

PR #1833, opened 2022-11-07
Customizable infotexts

### LINKED ISSUE #1768: Customizable Infotext
On the student index page and the contributor index page we currently show hardcoded infotexts (in closable callout boxes) with several entries, linking to the FAQ for more details.
The linked FAQ IDs are as well hardcoded and do not necessarily match to the freely editable FAQ entries. In addition, easy changes and additions to the info texts are not possible.

A new `Infotext` should be implemented, allowing for customized texts editable by managers.
- For each of the three index pages student, contributor, and grades (new) exactly one of these infotexts should exists.
- If an infotext is empty, no callout will be displayed on the respective page. Otherwise it will be shown like now.
- Each infotext should have a title that will be displayed as the heading.
- The infotexts should support HTML formatting just like FAQ entries.
- Both title and text must define an English and a German version.
- The infotexts should be editable on a new page that is only accessible for managers.

Kakadus:
Closes #1768.


evap/staff/templates/staff_edit_infotexts.html
```diff
@@ -0,0 +1,99 @@
...
+
+{% block content %}
+    {{ block.super }}
+
+    <form method="POST" class="form-horizontal" enctype="multipart/form-data">
+        {% csrf_token %}
+
+        <div class="card mb-3">
+            <div class="card-body">
+                <!-- Grid approach
```
He3lixxx, line 16: commented code

evap/staff/urls.py
```diff
@@ -84,6 +84,8 @@
     path("faq/", views.faq_index, name="faq_index"),
     path("faq/<int:section_id>", views.faq_section, name="faq_section"),
 
+    path("edit_infotexts/", views.edit_infotexts, name="edit_infotexts"),
```
He3lixxx, line 87: regarding naming: I think just "infotexts" as the url and `views.infotexts` or maybe `views.infotext_edit` would be more consistent with what we have so far:
* Views are usually named `course_edit`, not `edit_course`.
* The other "smaller" staff pages, e.g. FAQ or text answer warnings, are also "edit" views, but don't necessarily have that in their name.

evap/staff/views.py
```diff
@@ -2263,6 +2264,17 @@ def faq_section(request, section_id):
...
 
+@manager_required
+def edit_infotexts(request):
+    student_form = InfotextForm()
+    contributor_form = InfotextForm()
+    return render(
+        request,
+        "staff_edit_infotexts.html",
+        dict(student_infotext_form=student_form, contributor_infotext_form=contributor_form),
+    )
```
He3lixxx, line 2275: This doesn't (yet) save the changes that the user submitted.

This looks like work in progress, so I'll change the PR to be a draft for now so it's easier for us to keep an overview of the PRs. Please mark it as ready for review when we should review again.

Kakadus:
https://stackoverflow.com/questions/10071742/insert-or-update-data-with-django-formsets

niklasmohrin:
Missing `{{ formset.management_form }}`

Kakadus:
I should probably implement some tests here and there...

evap/evaluation/migrations/0132_add_infotext_model.py
```diff
@@ -0,0 +1,95 @@
...
+<em>Mehr Infos: <a href="/faq#faq-21-q">FAQ/Anonymität</a></em><br>
+<br>
+<b>Verweise auf andere Antworten</b><br>
+Lehrende können sich keine kompletten Fragebögen ansehen. Wenn du in einer Textantwort schreibst "wie oben schon erwähnt", können sie die zugehörige Antwort nicht finden.<br>
+<em>Mehr Infos: <a href="/faq#faq-24-q">FAQ/Verweise</a></em><br>
+<br>
+<b>Evaluierungs-Ergebnisse</b><br>
+Textantworten werden den bewerteten Personen und den Verantwortlichen der Veranstaltung angezeigt. Antworten auf Abstimmungsfragen werden für alle Nutzer·innen der Plattform veröffentlicht, wenn mindestens zwei Personen an der Evaluierung teilgenommen haben. Durchschnittsnoten werden berechnet, wenn die Teilnahmequote mindestens 20 Prozent beträgt.<br>
+<em>Mehr Infos: <a href="/faq#faq-3-s">FAQ/Ergebnisse</a></em>""",
+            "SI",
```
He3lixxx, line 60: As discussed: I'd say we want to use more verbose identifiers here

evap/evaluation/models.py
```diff
@@ -1497,6 +1496,36 @@ class Meta:
         verbose_name_plural = _("questions")
 
 
+class Infotext(models.Model):
+    """Infotext to display at the student index and contributor index pages"""
```
He3lixxx, line 1500:
```suggestion
    """Infotext to display, e.g. at the student index and contributor index pages"""
```

evap/evaluation/templates/infotext.html
```diff
@@ -0,0 +1,13 @@
+{% if infotext.title and infotext.content %}
```
He3lixxx, line 1: Personally, I'd prefer an or here: If one is empty, the other should still be shown.

On the staff edit page, there should probably be a check then that both are filled or both are empty.

evap/grades/views.py
```diff
@@ -22,6 +22,7 @@ def index(request):
     template_data = dict(
         semesters=Semester.objects.filter(grade_documents_are_deleted=False),
         disable_breadcrumb_grades=True,
+        infotext=Infotext.objects.filter(linked_page=Infotext.LinkedPage.GRADES_OVERVIEW).first(),
```
He3lixxx, line 25:
```suggestion
        infotext=Infotext.objects.get(linked_page=Infotext.LinkedPage.GRADES_OVERVIEW),
```

evap/staff/templates/staff_infotexts.html
```diff
@@ -0,0 +1,54 @@
...
+
+{% block content %}
+    {{ block.super }}
+
+    <form method="POST" class="form-horizontal" enctype="multipart/form-data">
+        {% csrf_token %}
+        {{ formset.management_form }}
+        <div class="card mb-3">
+            <div class="card-body">
+                <table id="question_table" class="table">
```
He3lixxx, line 16: question table?

evap/staff/views.py
```diff
@@ -2263,6 +2265,23 @@ def faq_section(request, section_id):
     return render(request, "staff_faq_section.html", template_data)
 
 
+@manager_required
+def infotexts(request):
+    InfoTextFormSet = modelformset_factory(Infotext, form=InfotextForm, extra=0)
+    formset = InfoTextFormSet(request.POST or None, queryset=Infotext.objects.all())
```
He3lixxx, line 2271: codecov says this is not covered -- can we have a simple "can be opened, stuff can be entered, can be submitted, and stuff in the DB is then changed" test?

evap/evaluation/templates/infotext.html
Kakadus: Moreover I'd say, if the content is empty, the box should be closed at all times

Kakadus:
Obviously, the nice edit view needs another test

evap/evaluation/models.py
```diff
@@ -1497,6 +1496,49 @@ class Meta:
...
+
+    def is_empty(self):
+        return not self.title or not self.content
+
+    class LinkedPage(models.TextChoices):
+        STUDENT_INDEX = ("student_index", "Student index page")
+        CONTRIBUTOR_INDEX = ("contributor_index", "Contributor index page")
+        GRADES_PAGES = ("grades_pages", "Grade publishing pages")
+
+    linked_page = models.CharField(
```
janno42, line 1518: the name should probably be changed because the infotext is not linked to a page but placed on it. how do you like something like `position`, `placement` or just `page`?

evap/staff/forms.py
```diff
@@ -1059,6 +1060,19 @@ class Meta:
...
+    class Meta:
+        model = Infotext
+        fields = ("title_de", "title_en", "content_de", "content_en")
+
+    def validate_unique(self):
+        super().validate_unique()
+        if self.errors:
+            for field in self.fields:
+                if not self.cleaned_data.get(field):
+                    self.add_error(field, "")
```
He3lixxx, line 1073: Maybe add a short comment why we do this here.

evap/grades/views.py
```diff
@@ -71,6 +73,7 @@ def course_view(request, semester_id, course_id):
         grade_documents=course.grade_documents.all(),
         disable_if_archived="disabled" if semester.grade_documents_are_deleted else "",
         disable_breadcrumb_course=True,
+        infotext=Infotext.objects.get(linked_page=Infotext.LinkedPage.GRADES_PAGES),
```
janno42, line 76: it is sufficient to place the infobox on the index page and the semester page. please remove it here.

He3lixxx (approved):
Code looks good to me. Haven't tested.

janno42 (changes requested):
very promising! i'm looking forward to having this feature.

can you please update the infobox to span the full page width even if the content is very short?

evap/evaluation/migrations/0132_add_infotext_model.py
```diff
@@ -0,0 +1,113 @@
+from django.db import migrations, models
+from evap.evaluation.models import Infotext
+
+
+def create_infotexts(apps, _schema_editor):
+    infotext = apps.get_model("evaluation", "Infotext")
+
+    infotexts = [
+        (
+            "Information for contributors",
```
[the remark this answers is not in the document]
Kakadus: I removed it, and now the tests cannot find the infotexts :cry: . If I visit the pages manually, they still show up (even after a clean `reload_testdata`)
But in general, what about production? The infotexts need to be in the "global state" of EvaP, so I'd say they should be kept in the migrations but removed in the test data. I'm not sure if I understood it correctly though

Kakadus:
I noticed the hardcoded restriction of callout-contents to be smaller than 1300px. This does not seem to be a clean procedure....

evap/evaluation/templates/infobox.html
```diff
@@ -9,6 +9,11 @@
                     aria-label="Close"></button>
         </div>
         <div class="callout-content small">
+            {# set infobox to hidden, if user closed it before #}
+            <script>
+                if (localStorage["infobox_{{ infotext.page }}"] === "hide")
+                    document.querySelector("#{{ infotext.page }}").classList.add("closed");
+            </script>
```
He3lixxx, line 16: How much earlier than the code in `infobox.ts` does this run?

evap/evaluation/templates/infotext.html
```diff
@@ -1,6 +1,6 @@
 {% load static %}
 
-{% if not infotext.is_empty %}
+{% if infotext and not infotext.is_empty %}
```
He3lixxx: If we don't pass an info text, we simply should not include this template, should we?

evap/static/scss/components/_callouts.scss
```diff
@@ -92,3 +92,17 @@
         }
     }
 }
+
+.infobox.callout {
```
He3lixxx, line 96: I think this means "an element with the class `callout` inside an element with the class `infobox`". Yet, in the commit introducing the change, you change the div in `infotexts.html` to have both the `infotext` and the `callout` class. Was that intended?

evap/static/scss/components/_infobox.scss
```diff
@@ -0,0 +1,13 @@
+.infobox.callout {
```
He3lixxx, line 1: This is completely duplicated from [evap/static/scss/components/_callouts.scss](https://github.com/e-valuation/EvaP/pull/1833/commits/3c5360a2f4aaca044596e2169b53083baf95a639#diff-97a87ac7f4b8ff763d6303984990fb98dbb69f65dbf3b54f8f73ce87a9133d81). Shouldn't be necessary, and even if including the rules in both files was necessary, we'd want to have them only in one .scss file.

evap/static/scss/components/_callouts.scss
```diff
@@ -94,10 +94,19 @@
 }
 
 .infobox.callout {
-  max-width: 100%;
+    min-width: 100%;
 
-  // enlarge infobox to max size
-  .callout-content {
-    width: 1300px;
-  }
```
He3lixxx: Afaik, in CSS, pixels [unintuitively are an absolute unit](https://drafts.csswg.org/css-values/#absolute-lengths), so I think 1300px wasn't really a problem here. But I'm fine with both the old and the new code.

evap/static/scss/components/_callouts.scss
```diff
@@ -94,10 +94,19 @@
...
+    }
+
+    // override transitions to include min-width
+    &.closing {
+        transition: border-radius 1s ease 1s, border-left-width 1s ease 1s, padding 1s ease-in-out, min-width 1s ease 1s;
+    }
+
+    &.opening {
+        transition: border-radius 1s ease, border-left-width 1s ease, padding 1s ease-in-out, min-width 1s ease;
+    }
```
He3lixxx, line 111: This looks like the other transition attributes are copied from somewhere, so we now have an invisible dependency here, and if we change the other place, we'd need to change this here too, which we'd very likely forget. Is that what's happening here?

evap/evaluation/migrations/0132_add_infotext_model.py
```diff
@@ -1,73 +1,4 @@
...
-Textantworten werden den bewerteten Personen und den Verantwortlichen der Veranstaltung angezeigt. Antworten auf Abstimmungsfragen werden für alle Nutzer·innen der Plattform veröffentlicht, wenn mindestens zwei Personen an der Evaluierung teilgenommen haben. Durchschnittsnoten werden berechnet, wenn die Teilnahmequote mindestens 20 Prozent beträgt.<br>
-<em>Mehr Infos: <a href="/faq#faq-3-s">FAQ/Ergebnisse</a></em>""",
-            Infotext.Page.STUDENT_INDEX,
-        ),
-        ("", "", "", "", Infotext.Page.GRADES_PAGES),
-    ]
-
-    for title_en, title_de, content_en, content_de, page in infotexts:
-        infotext.objects.create(title_en=title_en, title_de=title_de, content_en=content_en, content_de=content_de,
-                                page=page)
```
He3lixxx: I think @janno42's comment aimed at removing the actual infotext contents and titles from the migration, but we'd still need to create the three (empty) objects in the database, as our code relies on them existing.

evap/static/scss/components/_callouts.scss
niklasmohrin: No, I think this is correct. What you describe is `.foo .bar` with a space

evap/evaluation/migrations/0132_add_infotext_model.py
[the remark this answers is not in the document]
janno42: The idea is to have the texts within the test data so that we have some plausible texts when testing or using the page during development. To avoid duplicated data, their contents don't need to be added by a migration (but they need to be created empty nevertheless). The initial content on production will be changed manually (we do it like this for all data of this kind, e.g., FAQ texts or email templates).

evap/evaluation/templates/infobox.html
Kakadus: The important part here is, that the browser executes the javascript before the browser parses/displays the content of the content-div, so the infobox is closed via css before the browser displays the content of the infotext. This is only possible with scripts that aren't modules.
(Tested with firefox)

evap/static/scss/components/_callouts.scss
Kakadus: Yes, these transitions are the same from l. 80 & l. 88, and add the min-width property. We can probably create a mixin to reduce duplication.
In general, I tried not to edit ll. 1-94, as the code for these callouts seem to be copied from some github gist

evap/evaluation/migrations/0132_add_infotext_model.py
Kakadus: All right. Can I assert in my tests, that all infotexts are empty per default?

He3lixxx: Yes, if the intention is to have a "before, they are empty, and after page submission, they have the expected value" test. Otherwise, the assertion probably has limited use :D

evap/staff/templates/staff_infotexts.html
```diff
@@ -0,0 +1,64 @@
...
+        {% csrf_token %}
+        {{ formset.management_form }}
+        <div class="card mb-3">
+            <div class="card-body">
+                <table class="table">
+                    <thead>
+                    <tr>
+                        <th>{% trans 'Title/Content (German)' %}</th>
+                        <th>{% trans 'Title/Content (English)' %}</th>
+                    </tr>
```
janno42, line 21: please indent

evap/staff/templates/staff_infotexts.html
```diff
@@ -0,0 +1,64 @@
...
+                            <td colspan="2"
+                                class="spanning-title-row">
+                                {% trans form.instance.get_page_display %}
+                            </td>
+                        </tr>
+
+                        {% for error in form.non_field_errors %}
+                            <tr>
+                                <td colspan="2">
+                                        <span class="error-label">{{ error|escape }}</span>{% if not forloop.last %}<br />{% endif %}
```
janno42, line 38: please check indentation

evap/static/scss/components/_callouts.scss
```diff
@@ -50,6 +50,22 @@
...
+
+@mixin callout-close($extra_transition...) {
+    $transition: border-radius 1s ease 1s, border-left-width 1s ease 1s, padding 1s ease-in-out;
+    @if length($extra_transition) > 0 {
+        $transition: append($transition, $extra_transition);
+    }
+    transition: $transition;
+}
+
 .callout-closable {
```
janno42, line 69: we only ever use the closable callouts for infoboxes.
maybe we can call this `.callout-infobox`, put the min width in here, and get rid off the mixins.

janno42 (changes requested):

evap/staff/views.py
```diff
@@ -2263,6 +2265,27 @@ def faq_section(request, section_id):
...
+
+    if formset.is_valid():
+        formset.save()
+        messages.success(request, _("Successfully updated the infotext entries."))
+        return redirect("staff:infotexts")
+    if formset.errors:
+        messages.error(
+            request, _("The form was not saved and no infotexts were changed. Please resolve the errors shown below.")
+        )
```
He3lixxx, line 2280: I don't think we manually add an error message in any other place, we always get the automatic one by django. Why does that not work here?

evap/static/ts/src/infobox.ts
```diff
@@ -0,0 +1,40 @@
...
+        this.storageKey = "infobox_" + infobox_id;
+    }
+
+    public attach = (): void => {
+        // close the infobox and save state
+        this.closeButton.addEventListener("click", async event => {
+            this.infobox.classList.add("closing");
+            setTimeout(() => {
+                this.infobox.classList.replace("closing", "closed");
+            }, OPEN_CLOSE_TIMEOUT);
```
He3lixxx, line 24: I think we've talked about this, but I've forgotten, so maybe we should document it somewhere: Why do we need seperate `closing` and `closed` classes? Doesn't CSS work that way that it simply animates towards the target state and then stays there? Then, `open` and `closed` should suffice.

evap/static/ts/src/infobox.ts
```diff
@@ -0,0 +1,40 @@
...
+            event.stopPropagation(); // prevent immediate reopening
```
He3lixxx, line 25: Why would it reopen? This is the only event listener that should be active on the close button, right?

(especially consindering that the event handler below does not call this, but instead checks whether the initial state is closed)

evap/static/ts/src/infobox.ts
Kakadus: Now, the infobox can be reopened immediately, even when not completely closed yet. So, the background changes, which might be annoying. What does @janno42 think?

evap/static/ts/src/infobox.ts
Kakadus: I improved the comment

evap/static/ts/src/infobox.ts
He3lixxx: (we discussed in person, and decided that we do not want the "you can click on this"-hover-behavior while the infobox is still closing, so we need an additional class during transition. For javascript reasons, we also can't get the timeout from the element in a nice way here)

niklasmohrin:
In firefox, the selection highlight looks a bit awkward, as it is under the other textbox:
image

evap/contributor/views.py
```diff
@@ -108,6 +109,7 @@ def index(request):
         "semester_list": semester_list,
         "show_delegated": show_delegated,
         "delegate_selection_form": DelegateSelectionForm(),
+        "infotext": Infotext.objects.get(page=Infotext.Page.CONTRIBUTOR_INDEX),
```
niklasmohrin, line 112: I wonder if the infotexts would be better encapsulated if they would be included using [inclusion tags](https://docs.djangoproject.com/en/4.1/howto/custom-template-tags/#inclusion-tags) - there wouldn't be any global `infotext` in the template, instead it would just have `{% include_infobox Infotext.Page.CONTRIBUTOR_INDEX %}` where the infobox should be. The downside would be that it is not clear from just looking at the view function where database accesses happen, however we need to hit the DB anyways, so its not like we would overlook an unnecessary hit. @He3lixxx what do you think?

evap/evaluation/templates/navbar.html
```diff
@@ -73,6 +73,7 @@
                             <a class="dropdown-item" href="{% url 'staff:degree_index' %}">{% trans 'Degrees' %}</a>
                             <a class="dropdown-item" href="{% url 'staff:index' %}">{% trans 'Email templates' %}</a>
                             <a class="dropdown-item" href="{% url 'staff:faq_index' %}">{% trans 'FAQ' %}</a>
+                            <a class="dropdown-item" href="{% url 'staff:infotexts' %}">{% trans 'Infotexts' %}</a>
```
niklasmohrin, line 76: Maybe "Edit infotexts"?

evap/staff/forms.py
```diff
@@ -1063,6 +1064,21 @@ class Meta:
         }
 
 
+class InfotextForm(forms.ModelForm):
+    class Meta:
+        model = Infotext
+        fields = ("title_de", "title_en", "content_de", "content_en")
+
+    def validate_unique(self):
```
niklasmohrin, line 1072: As I understand it, the code below doesn't validate the uniqueness, so this seems like the wrong place to put this logic

evap/staff/forms.py
```diff
@@ -1063,6 +1064,21 @@ class Meta:
...
+        super().validate_unique()
+
+        # add empty error to all empty fields to show the user that they are required
```
niklasmohrin, line 1075: I don't get it, there are some errors, but we need to add more? What kind of error is it that makes `self.errors` true?

evap/staff/templates/staff_infotexts.html
```diff
@@ -0,0 +1,64 @@
...
+
+    <form method="POST" class="form-horizontal" id="infotext-formset">
+        {% csrf_token %}
+        {{ formset.management_form }}
+        <div class="card mb-3">
+            <div class="card-body">
+                <table class="table">
+                    <thead>
+                        <tr>
+                            <th>{% trans 'Title/Content (German)' %}</th>
```
niklasmohrin, line 19: I wonder if it would be better to have headings "Title" / "Content" at the fields themselves, not on top - not sure if it looks better though. I feel like everything looks a bit empty right now

evap/staff/tools.py
```diff
@@ -309,6 +309,13 @@ def merge_users(main_user, other_user, preview=False):
     return merged_user, errors, warnings
 
 
+def fill_infotext_formset(formset, formset_id, title_de="", title_en="", content_de="", content_en=""):
```
niklasmohrin, line 312: To not mess up the ordering, the last four arguments could be kw-only:
```suggestion
def fill_infotext_formset(formset, formset_id, *, title_de="", title_en="", content_de="", content_en=""):
```
Furthermore, is the `formset_id` really the id of the formset, not a particular form in it?

evap/staff/urls.py
```diff
@@ -84,6 +84,8 @@
     path("faq/", views.faq_index, name="faq_index"),
     path("faq/<int:section_id>", views.faq_section, name="faq_section"),
 
+    path("infotexts/", views.infotexts, name="infotexts"),
```
niklasmohrin, line 87: I wonder if it should be `/infotexts/edit` for, we are not super consistent with this, but it feels right to me. @He3lixxx ?

evap/staff/views.py
```diff
@@ -2276,6 +2278,23 @@ def faq_section(request, section_id):
     return render(request, "staff_faq_section.html", template_data)
 
 
+@manager_required
+def infotexts(request):
+    InfoTextFormSet = modelformset_factory(Infotext, form=InfotextForm, edit_only=True, extra=0)
```
niklasmohrin, line 2283:
```suggestion
    InfotextFormSet = modelformset_factory(Infotext, form=InfotextForm, edit_only=True, extra=0)
```

evap/static/ts/src/infobox.ts
```diff
@@ -0,0 +1,40 @@
...
+const OPEN_CLOSE_TIMEOUT = 2000;
+
+export class InfoboxLogic {
+    private readonly infobox: HTMLDivElement;
+    private readonly closeButton: HTMLButtonElement;
+    private readonly storageKey: string;
+
+    constructor(infobox_id: string) {
+        this.infobox = selectOrError("#" + infobox_id);
+        this.closeButton = selectOrError("#" + infobox_id + " .callout-infobox-close");
```
niklasmohrin, line 14: Shouldn't this be something similar to:
```suggestion
        this.closeButton = selectOrError(".callout-infobox-close", this.infobox);
```

niklasmohrin (commented):
Some thoughts:

evap/evaluation/templates/navbar.html
janno42: Almost all of these pages are meant for editing, so I would keep it short. "Infotexts" is fine :)

evap/staff/templates/staff_infotexts.html
janno42: The page would actually better fit the overall layout if the three sections would be placed inside cards with the page names in the cards' headers. The "Title/Content (German)" headings would then be placed on top of the fields within each card.

evap/staff/urls.py
janno42: I would not add the "edit". Like with text answer warnings, degrees and course types, we only have one page acting as overview list and edit page for these things, so there is no need to be more specific.

evap/contributor/views.py
He3lixxx: Yeah sounds good.

evap/staff/forms.py
Kakadus: Yes, you are right. The problem is, that I found no other way to annotate fields with an error. The CheckConstraint only creates a non_field_error, so I add them to each empty field.
Another option is to create a custom CheckConstraint (e. g. by inheriting from CheckConstraint) that add the errors to the empty fields automatically, when validation fails. This seems a bit overkill though.
The reason why e. g. FAQ does not need this special logic, is because use simply `blank=false` and let django do its job. We cannot use that, because Infotext can be empty

evap/staff/forms.py
Kakadus: Its the error of the failed CheckConstraint, which creates an error on the whole model, not individual empty fields

Kakadus:
> In firefox, the selection highlight looks a bit awkward, as it is under the other textbox

Yes. Should we add a gap after each inputs? This is happening at faq_question as well:
image

evap/staff/templates/staff_infotexts.html
Kakadus: you can check out two first versions: `card layout` and `css grid instead of table`

evap/staff/forms.py
Kakadus: Out of boredom, I added it in commit `nothalfempty constraint`

evap/staff/forms.py
He3lixxx: From [the documentation](https://docs.djangoproject.com/en/4.1/ref/forms/validation/):
> The form subclass’s `clean()` method can perform validation that requires access to multiple form fields. This is where you might put in checks such as “if field A is supplied, field B must contain a valid email address”. This method can return a completely different dictionary if it wishes, which will be used as the `cleaned_data`.

And I guess `clean` should work for our purposes here?

janno42:
> > In firefox, the selection highlight looks a bit awkward, as it is under the other textbox
> 
> Yes. Should we add a gap after each inputs? This is happening at faq_section as well: image

Fine for now.

evap/staff/tools.py
janno42, line 1: please move to `tests/utils.py`

evap/staff/templates/staff_infotexts.html
```diff
@@ -0,0 +1,42 @@
...
+                    {% trans form.instance.get_page_display %}
+                </div>
+                <div class="card-body" style="display: grid; grid-template-columns: auto auto; grid-column-gap: 10px">
+                    {% for hidden in form.hidden_fields %}
+                        {{ hidden }}
+                    {% endfor %}
+                    <label class="card-subtitle">{% trans 'Title/Content (German)' %}</label>
+                    <label class="card-subtitle">{% trans 'Title/Content (English)' %}</label>
+
+                    <div class="mb-1 pt-2 border-bottom" style="grid-column: 1/-1;"></div>
```
janno42, line 26: please remove the horizontal line

janno42 (changes requested):
Please change the FAQ card on the staff index page to this:

FAQ and Infotexts
------
- All FAQ Sections
- All Infotexts

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

He3lixxx (approved):

evap/static/ts/src/infobox.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    constructor(infobox_id: string) {
+        this.infobox = selectOrError("#" + infobox_id);
+        this.closeButton = selectOrError(".callout-infobox-close", this.infobox);
+        this.storageKey = "infobox_" + infobox_id;
+    }
+
+    public attach = (): void => {
+        // close the infobox and save state
+        this.closeButton.addEventListener("click", async event => {
```
niklasmohrin, line 18: why are the two event handlers `async`?

niklasmohrin (changes requested):
Very cool already, I really like the changes to the important html files :+1: Some requests:

evap/staff/tests/test_views.py
```diff
@@ -140,6 +142,53 @@ def setUpTestData(cls):
...
+        page = self.app.get(self.url, user=self.manager)
+        formset = page.forms["infotext-formset"]
+
+        # submit invalid data
+        helper_fill_infotext_formset(formset, 0, title_de="abc", title_en="def", content_de="ghi", content_en="jkl")
+        helper_fill_infotext_formset(formset, 1, title_de="invalid infotext", content_en="no translations")
+
+        empty_form_id = formset["form-0-id"]
+        filled_form_id = formset["form-1-id"]
+        formset.submit()
```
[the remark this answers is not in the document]
Kakadus: Django said `wontfix`, so no!!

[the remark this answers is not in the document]
Kakadus: (We agreed, that this is not necessary)

evap/evaluation/templates/infobox.html
```diff
@@ -0,0 +1,26 @@
+{% load static %}
+
+{% if not infotext.is_empty %}
+    <div id="infobox_{{ infotext.page }}" class="infobox callout callout-info callout-infobox">
```
niklasmohrin, line 4: Nit: Usually, we use hyphens instead of underscores for classes and ids

niklasmohrin (approved):
Looks good, final comments:

PR #1837, opened 2022-11-28
Http verbs

florian-str:
It changes the HTTP-Endpoint for a semester activation.

Due to the nature of the confirmation modals, only the query directly effected was changed to a form.

### DIFF
--- a/evap/rewards/tests/test_views.py
+++ b/evap/rewards/tests/test_views.py
@@ -205,21 +205,30 @@ class TestSemesterActivationView(WebTestStaffMode):
             course=course,
         )
 
-        cls.url = f"/rewards/reward_semester_activation/{cls.semester.pk}/"
+        cls.url = f"/rewards/reward_semester_activation/{cls.semester.pk}/edit"
 
     def test_activate(self):
         baker.make(SemesterActivation, semester=self.semester, is_active=False)
-        self.app.post(self.url + "on", user=self.manager)
+        response = self.app.post(self.url, user=self.manager)
+        form = response.forms["form_activation_status"]
+        form.set("activation_status", "on")
+        form.submit()
         self.assertTrue(is_semester_activated(self.semester))
 
     def test_deactivate(self):
         baker.make(SemesterActivation, semester=self.semester, is_active=True)
-        self.app.post(self.url + "off", user=self.manager)
+        response = self.app.post(self.url, user=self.manager)
+        form = response.forms["form_activation_status"]
+        form.set("activation_status", "off")
+        form.submit()
         self.assertFalse(is_semester_activated(self.semester))
 
     def test_activate_after_voting(self):
         baker.make(SemesterActivation, semester=self.semester, is_active=False)
         self.assertEqual(0, reward_points_of_user(self.student))
-        response = self.app.post(self.url + "on", user=self.manager)
+        response = self.app.post(self.url, user=self.manager)
+        form = response.forms["form_activation_status"]
+        form.set("activation_status", "on")
+        response = form.submit()
         self.assertContains(response, "3 reward points were granted")
         self.assertEqual(3, reward_points_of_user(self.student))
--- a/evap/rewards/tests/test_views.py
+++ b/evap/rewards/tests/test_views.py
@@ -205,30 +205,28 @@ class TestSemesterActivationView(WebTestStaffMode):
             course=course,
         )
 
-        cls.url = f"/rewards/reward_semester_activation/{cls.semester.pk}/edit"
+        cls.url = reverse("rewards:semester_activation_edit", args=[cls.semester.pk])
 
+    def test_invalid(self):
+        baker.make(SemesterActivation, semester=self.semester, is_active=False)
+        response = self.app.post(self.url, user=self.manager, params={"activation_status": "invalid"}, status=400)
+        self.assertFalse(is_semester_activated(self.semester))
     def test_activate(self):
         baker.make(SemesterActivation, semester=self.semester, is_active=False)
-        response = self.app.post(self.url, user=self.manager)
-        form = response.forms["form_activation_status"]
-        form.set("activation_status", "on")
-        form.submit()
+        response = self.app.post(self.url, user=self.manager, params={"activation_status": "on"})
         self.assertTrue(is_semester_activated(self.semester))
 
     def test_deactivate(self):
         baker.make(SemesterActivation, semester=self.semester, is_active=True)
-        response = self.app.post(self.url, user=self.manager)
-        form = response.forms["form_activation_status"]
-        form.set("activation_status", "off")
-        form.submit()
+        response = self.app.post(self.url, user=self.manager, params={"activation_status": "off"})
         self.assertFalse(is_semester_activated(self.semester))
 
     def test_activate_after_voting(self):
         baker.make(SemesterActivation, semester=self.semester, is_active=False)
         self.assertEqual(0, reward_points_of_user(self.student))
-        response = self.app.post(self.url, user=self.manager)
-        form = response.forms["form_activation_status"]
-        form.set("activation_status", "on")
-        response = form.submit()
-        self.assertContains(response, "3 reward points were granted")
+        response = self.app.post(self.url, user=self.manager, params={"activation_status": "on"})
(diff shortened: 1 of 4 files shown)


niklasmohrin:
If you want to try at home: The branch is not based on the current version of our `main` branch, likely because you were on your previous feature branch when starting to work on this next feature. You can re-base your new commits by updating your local `main` branch to the newest version from us (`git checkout main; git pull upstream main`) and then running `git rebase main` after switching back to your feature branch (`git checkout http_verbs`).

... or we do it together at the next meeting :)

evap/rewards/views.py
```diff
@@ -137,10 +137,12 @@ def reward_point_redemption_event_export(request, event_id):
...
 
 
+@require_POST
 @manager_required
-def semester_activation(request, semester_id, active):
+def semester_activation_edit(request, semester_id):
     semester = get_object_or_404(Semester, id=semester_id)
-    active = active == "on"
+    status = request.POST.get("activation_status")
+    active = status == "on"
```
niklasmohrin, line 145: I suppose we should respond with 400 Bad Request if `status` is not `on` or `off`

evap/staff/templates/staff_semester_view.html
```diff
@@ -549,23 +553,15 @@ <h3>
...
-            $.ajax({
-                type: "POST",
-                url: "{% url 'rewards:semester_activation' semester.id 'on' %}",
-                data: {},
-                success: function(){ location.reload(); },
-                error: function(){ window.alert("{% trans 'The server is not responding.' %}"); }
-            });
+        function setRewardPointActivationStatus(status) {
+            document.getElementById("activation_status").value = status;
+            document.getElementById("form_activation_status").submit();
```
niklasmohrin, line 558: This is a bit odd, persisting the state in the DOM like this. I think the optimal modeling would link the two submit buttons to the form and give them `name` and `value` (if we have `<button type="submit" name="foo" value="bar">`, clicking the button will submit with the additional data `foo=bar`; using `form=theform` on the button makes it possible for it to be outside of the `form` tag). This works directly with the deactivation-button, however for the activation, the real button is inside the modal and we currently don't have a way to modify this button from the outside (it would also be prone to error). I am not sure what the best solution is here, currently I am think that we should have two buttons with `form`, `name` and `value` where we currently have the button that triggers `deactivateRewardPoints`, make the activation one hidden and let the modal-action click this hidden button. But that is also a bit weird. @He3lixxx Do you a better idea?

niklasmohrin (changes requested):

evap/rewards/tests/test_views.py
```diff
@@ -205,21 +205,30 @@ def setUpTestData(cls):
...
-        cls.url = f"/rewards/reward_semester_activation/{cls.semester.pk}/"
+        cls.url = f"/rewards/reward_semester_activation/{cls.semester.pk}/edit"
 
     def test_activate(self):
         baker.make(SemesterActivation, semester=self.semester, is_active=False)
-        self.app.post(self.url + "on", user=self.manager)
+        response = self.app.post(self.url, user=self.manager)
+        form = response.forms["form_activation_status"]
+        form.set("activation_status", "on")
+        form.submit()
```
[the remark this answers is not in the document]
He3lixxx: I was a bit confused here that this passes the tests, because in my mental model, this view is a REST-like POST-only view that doesn't return any HTML, so the `response.forms["form_activation_status"]` access should fail.

Turns out the view just has `return semester_view(request=request, semester_id=semester_id)` at its end, and thus doesn't indicate any error on the first post request here, but just renders the page.

Two thoughts about that:
1. Should it be a redirect instead of a call to the other view function?
2. If we return 400 Bad Request from the view (as discussed in https://github.com/e-valuation/EvaP/pull/1837#discussion_r1039954115) , these tests should fail and you should _need_ to use the correct post data immediately.

evap/staff/templates/staff_semester_view.html
He3lixxx: Not really.

In general, I think our modals should support adding additional POST data to the submission, we may just want to add that.

evap/rewards/views.py
```diff
@@ -137,13 +136,19 @@ def reward_point_redemption_event_export(request, event_id):
...
+        active = False
+    else:
+        raise SuspiciousOperation("Invalid activation keyword")
     SemesterActivation.objects.update_or_create(semester=semester, defaults={"is_active": active})
     if active:
         grant_eligible_reward_points_for_semester(request, semester)
-
-    return semester_view(request=request, semester_id=semester_id)
+    messages.get_messages(request).used = False
+    return redirect("staff:semester_view", semester_id)  # semester_view(request=request, semester_id=semester_id)
```
He3lixxx, line 154:
```suggestion
    return redirect("staff:semester_view", semester_id)
```

evap/rewards/views.py
```diff
@@ -137,13 +136,19 @@ def reward_point_redemption_event_export(request, event_id):
...
+    elif status == "off":
+        active = False
+    else:
+        raise SuspiciousOperation("Invalid activation keyword")
     SemesterActivation.objects.update_or_create(semester=semester, defaults={"is_active": active})
     if active:
         grant_eligible_reward_points_for_semester(request, semester)
-
-    return semester_view(request=request, semester_id=semester_id)
+    messages.get_messages(request).used = False
```
He3lixxx, line 153: Is this necessary / what does it do? I think we've used redirects before and everything worked fine with messages.

evap/staff/templates/staff_semester_view.html
```diff
@@ -4,8 +4,12 @@
 {% load evaluation_filters %}
 
 {% block content %}
-    {{ block.super }}
 
+    {{ block.super }}
+    <form id="form_activation_status" method="post" action="{% url 'rewards:semester_activation_edit' semester.id%}">
```
He3lixxx, line 9:
```suggestion
    <form id="form_activation_status" method="post" action="{% url 'rewards:semester_activation_edit' semester.id %}">
```

evap/staff/templates/staff_semester_view.html
```diff
@@ -4,8 +4,12 @@
...
+        <input type="hidden" id="activation_status" name="activation_status" value="on"> {# FIXME #}
```
He3lixxx, line 10: I'd oppose having a fixme without any comment here. I would be (hesitantly) okay with a reference to an issue (`FIXME #123456`). Overall, I'm just in favor of adding the issue and not adding the fixme.  TODOs in code [have a tendency to stay there](https://github.com/id-Software/Quake-2/search?q=todo).

@niklasmohrin what do you think? (Follow-up to https://github.com/e-valuation/EvaP/pull/1837#discussion_r1039960441)

(repeated below in the javascript)

He3lixxx (changes requested):
Code looks very nice. I have some minor formatting and some project management / organizational annotations though.

evap/staff/templates/staff_semester_view.html
niklasmohrin: No strong opinion, we can make an issue if you want

evap/rewards/tests/test_views.py
```diff
@@ -205,21 +205,26 @@ def setUpTestData(cls):
             course=course,
         )
 
-        cls.url = f"/rewards/reward_semester_activation/{cls.semester.pk}/"
+        cls.url = reverse("rewards:semester_activation_edit", args=[cls.semester.pk])
+
+    def test_invalid(self):
```
niklasmohrin, line 210: In theory, we could also do one case that checks what happens if the parameter is missing

evap/staff/templates/staff_semester_view.html
```diff
@@ -4,8 +4,12 @@
 {% load evaluation_filters %}
 
 {% block content %}
-    {{ block.super }}
 
+    {{ block.super }}
```
niklasmohrin, line 8: Please revert this little change, so that it again looks like this:
```
{% block content %}
    {{ block.super }}

    <form ...
```

niklasmohrin (commented):
Looks good, final touches:

He3lixxx (approved):

He3lixxx:
Thanks for submitting the PR :)

PR #1841, opened 2022-12-29
Importer performance

### LINKED ISSUE #1840: Openpyxl `max_column` computation bottlenecks importers
The importers currently take seconds up to minutes to run. There's no inherent reason for that, all processing logic inside of them is designed to have O(n) runtime with O(1) database queries. A quick profiling run of the import view shows that we spend most time in openpyxl's `worksheet.py` file in [`max_column`](https://foss.heptapod.net/openpyxl/openpyxl/-/blob/branch/3.0/openpyxl/worksheet/worksheet.py#L365-L375) and a generator expression creating the set `cols`:
```python3
    @property
    def max_column(self):
        """The maximum column index containing data (1-based)

        :type: int
        """
        max_col = 1
        if self._cells:
            cols = set(c[1] for c in self._cells)  ##### line 373 #####
            max_col = max(cols)
        return max_col
```

photo_5289897819194377466_y

I feel like figuring out how many columns a 200kB (12 x 6000 cells) excel file has should be possible in less than 60 seconds, even in a debug-profiling run.
They have an open issue here: https://foss.heptapod.net/openpyxl/openpyxl/-/issues/1763

If I read the output above right, we have 3000 accesses to max_column, just reusing the computed value might be enough for now.

He3lixxx:
Best reviewed by commit. Fixes #1840.

Openpyxl no longer spends half a minute on determining the worksheet dimensions.
We prevent some N+1 query behavior, making the importer and the staff semester page (which the importer redirects to) a bit faster.

On my machine, import times for a 6000 row sheet go from 32-40s to 3-6s.

### DIFF
--- a/evap/staff/importers/base.py
+++ b/evap/staff/importers/base.py
@@ -216,10 +216,9 @@ class ExcelFileRowMapper:
                     )
                 )
 
-            for row_index in range(self.skip_first_n_rows, sheet.max_row):
-                # We use 0-based indexing, openpyxl uses 1-based indexing.
-                row = [cell.value for cell in sheet[row_index + 1]]
-                location = ExcelFileLocation(sheet.title, row_index)
+            # We use 0-based indexing, openpyxl uses 1-based indexing.
+            for row_index, row in enumerate(sheet.iter_rows(min_row=self.skip_first_n_rows + 1, values_only=True)):
+                location = ExcelFileLocation(sheet.title, row_index + 1)
 
                 if not all(isinstance(cell, str) or cell is None for cell in row):
                     self.importer_log.add_error(
--- a/evap/staff/importers/base.py
+++ b/evap/staff/importers/base.py
@@ -216,9 +216,11 @@ class ExcelFileRowMapper:
                     )
                 )
 
-            # We use 0-based indexing, openpyxl uses 1-based indexing.
-            for row_index, row in enumerate(sheet.iter_rows(min_row=self.skip_first_n_rows + 1, values_only=True)):
-                location = ExcelFileLocation(sheet.title, row_index + 1)
+            # openpyxl uses 1-based indexing.
+            for row_number, row in enumerate(
+                sheet.iter_rows(min_row=self.skip_first_n_rows + 1, values_only=True), start=self.skip_first_n_rows
+            ):
+                location = ExcelFileLocation(sheet.title, row_number)
 
                 if not all(isinstance(cell, str) or cell is None for cell in row):
                     self.importer_log.add_error(
--- a/evap/staff/importers/enrollment.py
+++ b/evap/staff/importers/enrollment.py
@@ -339,9 +339,10 @@ class CourseMergeLogic:
         if len(responsibles) != 1 or responsibles[0].email != course_data.responsible_email:
             hindrances.append(_("the responsibles of the course do not match"))
 
-        if len(merge_candidate.evaluations.all()) != 1:
+        merge_candidate_evaluations = merge_candidate.evaluations.all()
+        if len(merge_candidate_evaluations) != 1:
             hindrances.append(_("the existing course does not have exactly one evaluation"))
-        elif merge_candidate.evaluations.all()[0].wait_for_grade_upload_before_publishing != course_data.is_graded:
+        elif merge_candidate_evaluations[0].wait_for_grade_upload_before_publishing != course_data.is_graded:
             hindrances.append(_("the evaluation of the existing course has a mismatching grading specification"))
 
         return hindrances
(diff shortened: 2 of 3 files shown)


evap/evaluation/templatetags/evaluation_filters.py
```diff
@@ -74,6 +74,11 @@
 register = Library()
 
 
+@register.filter(name="len")
+def _len(a):
+    return len(a)
```
He3lixxx, line 79: django has a `length` filter already.

evap/staff/templates/staff_semester_view.html
```diff
@@ -384,11 +384,11 @@ <h3>
                                             {{ course.responsibles_names }}
                                         </div>
                                     </td>
-                                    <td data-col="evaluation-count" data-order="{{ course.evaluations.count }}">
-                                        {% if course.evaluations.count == 0 %}
+                                    <td data-col="evaluation-count" data-order="{{ course.evaluations.all|len }}">
+                                        {% if course.evaluations.all|len == 0 %}
```
He3lixxx, line 388: just stumbled across people on the internet claiming that `prefetch_related` combined with a `.count()` call will use the prefetched object cache instead of hitting the database again. This is not what I experienced in the past, but I will have to check again.

He3lixxx:
moving the staff semester page N+1 query fix to #1842

evap/staff/importers/base.py
```diff
@@ -216,10 +216,9 @@ def map(self, file_content: bytes):
                     )
                 )
 
-            for row_index in range(self.skip_first_n_rows, sheet.max_row):
-                # We use 0-based indexing, openpyxl uses 1-based indexing.
-                row = [cell.value for cell in sheet[row_index + 1]]
-                location = ExcelFileLocation(sheet.title, row_index)
+            # We use 0-based indexing, openpyxl uses 1-based indexing.
+            for row_index, row in enumerate(sheet.iter_rows(min_row=self.skip_first_n_rows + 1, values_only=True)):
+                location = ExcelFileLocation(sheet.title, row_index + 1)
```
niklasmohrin, line 221: If `skip_first_n_rows` is non-zero, `row_index` will be incorrect. To get rid of the `+1` in the second line, I ~~wanted to~~ suggest using `enumerate(..., start=...)`~~, but even better: Let's just not use `values_only` and ask each cell for its `row`.~~ (edit: no, not better, maybe there are no cells in a row)

Should we add a test?

evap/staff/importers/enrollment.py
```diff
@@ -341,7 +341,7 @@ def get_merge_hindrances(course_data: CourseData, merge_candidate: Course) -> Li
 
         if len(merge_candidate.evaluations.all()) != 1:
             hindrances.append(_("the existing course does not have exactly one evaluation"))
-        elif merge_candidate.evaluations.get().wait_for_grade_upload_before_publishing != course_data.is_graded:
+        elif merge_candidate.evaluations.all()[0].wait_for_grade_upload_before_publishing != course_data.is_graded:
```
niklasmohrin, line 344: hm, I guess I understand why this works, but I think the intent of sparing the extra query gets across better if we move the ~~list~~ queryset into its own variable.

While we are at the topic of performance, we _could_ [limit the number of evaluations](https://docs.djangoproject.com/en/4.1/topics/db/queries/#limiting-querysets) returned here as we only need 2 to prove that there are too many, but in our domain there shouldn't be more than, say, 10 evaluations anyways, so probably not worth it.

---

When we get Python 3.10, it will be even nicer:
```python
match list(merge_candidate.evaluations.all()):
    case [existing_evaluation]:
        if existing_evaluation.wait_for_grade_upload_before_publishing != course_data.is_graded:
            hindrances.append(_("the evaluation of the existing course has a mismatching grading specification"))
    case _:
        hindrances.append(_("the existing course does not have exactly one evaluation"))
```

niklasmohrin (commented):
haven't measured or thought about the prefetches yet (more prefetching = more performance, right?); I generally trust you with this, but if you want I can also take a closer look

evap/staff/importers/base.py
He3lixxx: Added a test in https://github.com/e-valuation/EvaP/pull/1841/commits/4a6a44658d1e09a45899c0407ddc57d8a8e1b924, maybe check if you think it's worth it.

We never use the mapper with another `skip_first_n_rows` than one. We could also remove that feature, and have 1 hardcoded until we need something different (which likely isn't going to happen -- you could just remove the excess lines using spreadsheet software before uploading for import). I think I'd like that a bit more.

evap/staff/importers/enrollment.py
He3lixxx: Moved into separate variable.

Limited prefetching is a good idea, but it's also a slippery slope. I wouldn't want to prefetch the limited set into `evaluations`, because other code might (rightfully) assume that it has all evaluations in that attribute. So we'd need a separate attribute, but then we'd have `evaluations` not prefetched for other accesses.
As you said, I don't think there's much to gain here, so I'll leave that for now.

niklasmohrin (approved):

Issue #1839, opened 2022-12-28
Importer doesn't create general contributions


janno42:
When importing enrollment data, no `general_contribution`s are created for the newly added evaluations. This prevents managers from, e.g., using the "assign questionnaires" feature which fails with an `AttributeError`:

```
File "staff/views.py", line 802, in semester_questionnaire_assign
    evaluation.general_contribution.questionnaires.set(form.cleaned_data[evaluation.course.type.name])

Exception Type: AttributeError
Exception Value: 'NoneType' object has no attribute 'questionnaires'
```

The importer should make sure that every new evaluation has a `general_contribution`.

niklasmohrin:
Hm, shouldn't the questionnaire assign view just create it if users tried to use the general contribution? Or, if we dont want to allow that, it shouldnt display an option to assign to the general questionnaire, right?

Do we currently have some invariants about what questionnaires exist for an evaluation? If so, we should probably enforce them with database contraints

He3lixxx:
So the problem is that the importer [uses `bulk_create`](https://github.com/e-valuation/EvaP/blob/e889dce5b2111f42f3cda777649c19d30de165be/evap/staff/importers/enrollment.py#L739) instead of `save` to be faster, but we [use `save` to create additional objects, in this case, the general contribution](https://github.com/e-valuation/EvaP/blob/e889dce5b2111f42f3cda777649c19d30de165be/evap/evaluation/models.py#L456-L462).

I wouldn't want to duplicate the "when a new evaluation is created, we also must create a general contribution for it" logic across the codebase.
I think the approach of overriding `save` and having `pre_save` / `post_save` signals while also having methods like `bulk_create` that completely circumvent them is generally broken in django.

We could defer creation of the general contribution to its first access using the getter for the `general_contribution` property. However, code can also try to access the database object via `Contribution.objects.filter(..., contributor=None)`. This would bypass any getters, causing errors. Searching for `contributor=None` gives me 17 places that look like they might break if the object doesn't exist in the database.

If we use `save` instead of `bulk_*` in the importer, the importer tests will get 4 seconds slower (11s instead of 7s), and the import of somewhat realistically sized file (500 students, 100 courses, 4 enrollments per student) will get ~8s slower (currently 40s vs 48s, but that's due to a performance bug in openpyxl. There is no good reason for the importer to take longer than a second, I think).

[...]

PR #1842, opened 2022-12-30
Prefetching

He3lixxx:
I went through the pages and fixed some N+1 query patterns we had.

Throughout the commits I changed a bit of stuff with the `is_m2m_prefetched` method. I ended up just making it `is_prefetched` and handle both m2m as well as foreign key fields (forward and backward lookup). A given attribute on a model can only refer to a single field type, so it doesn't add much to have separate `is_prefetched` methods.

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -35,7 +35,7 @@ from evap.evaluation.tools import (
     clean_email,
     date_to_datetime,
     is_external_email,
-    is_m2m_prefetched,
+    is_many_prefetched,
     translate,
     vote_end_datetime,
 )
@@ -601,7 +601,7 @@ class Evaluation(LoggedModel):
         if self._participant_count is not None:
             return self._participant_count
 
-        if is_m2m_prefetched(self, "participants"):
+        if is_many_prefetched(self, "participants"):
             return len(self.participants.all())
 
         return self.participants.count()
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -35,7 +35,7 @@ from evap.evaluation.tools import (
     clean_email,
     date_to_datetime,
     is_external_email,
-    is_many_prefetched,
+    is_prefetched,
     translate,
     vote_end_datetime,
 )
@@ -219,8 +219,8 @@ class Questionnaire(models.Model):
 
     @property
     def can_be_edited_by_manager(self):
-        if is_many_prefetched(self, "contributions"):
-            if all(is_single_prefetched(contribution, "evaluation") for contribution in self.contributions.all()):
+        if is_prefetched(self, "contributions"):
+            if all(is_prefetched(contribution, "evaluation") for contribution in self.contributions.all()):
                 return all(
                     contribution.evaluation.state == Evaluation.State.NEW for contribution in self.contributions.all()
                 )
@@ -367,7 +367,7 @@ class Course(LoggedModel):
 
     @property
     def all_evaluations_finished(self):
-        if is_many_prefetched(self, "evaluations"):
+        if is_prefetched(self, "evaluations"):
             return all(evaluation.state >= Evaluation.State.EVALUATED for evaluation in self.evaluations.all())
 
         return not self.evaluations.exclude(state__gte=Evaluation.State.EVALUATED).exists()
@@ -610,7 +610,7 @@ class Evaluation(LoggedModel):
         if self._participant_count is not None:
             return self._participant_count
 
-        if is_many_prefetched(self, "participants"):
+        if is_prefetched(self, "participants"):
             return len(self.participants.all())
 
         return self.participants.count()
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -361,6 +361,9 @@ class Course(LoggedModel):
 
     @property
     def all_evaluations_finished(self):
+        if is_many_prefetched(self, "evaluations"):
+            return all(evaluation.state >= Evaluation.State.EVALUATED for evaluation in self.evaluations.all())
+
         return not self.evaluations.exclude(state__gte=Evaluation.State.EVALUATED).exists()
 
 
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -219,6 +219,12 @@ class Questionnaire(models.Model):
 
     @property
     def can_be_edited_by_manager(self):
+        if is_many_prefetched(self, "contributions"):
+            if all(is_single_prefetched(contribution, "evaluation") for contribution in self.contributions.all()):
+                return all(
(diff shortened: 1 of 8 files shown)


evap/evaluation/models.py
```diff
@@ -550,10 +559,14 @@ def general_contribution_has_questionnaires(self):
 
     @property
     def all_contributions_have_questionnaires(self):
-        return (
-            self.general_contribution
-            and not self.contributions.annotate(Count("questionnaires")).filter(questionnaires__count=0).exists()
-        )
```
He3lixxx, line 556: There is a test (now failing) asserting that we check for the existence of the general_contribution and return False here if it does not yet exist.

Boils down to the broader discussion about how to handle general contributions in #1839

evap/evaluation/tests/test_tools.py
```diff
@@ -62,20 +62,26 @@ def test_log_exceptions_decorator(self, mock_logger, __):
...
 
 class TestHelperMethods(WebTest):
-    def test_is_many_prefetched(self):
-        evaluation = baker.make(Evaluation)
+    def test_is_prefetched(self):
+        evaluation = baker.make(Evaluation, voters=[baker.make(UserProfile)])
         baker.make(Contribution, evaluation=evaluation)
 
-        self.assertFalse(is_many_prefetched(evaluation, "contributions"))
+        def test_logic(cls, pk: int, field: str):
```
niklasmohrin, line 69: I was playing around here a bit, I think the best annotation would be `Type[Model]`, but mypy also accepts just `Type`, because that is `Type[Any]`, but not `Type[object]`. Anyways, I suggest `Type[Model]` :^)

evap/evaluation/tests/test_tools.py
```diff
@@ -62,20 +62,26 @@ def test_log_exceptions_decorator(self, mock_logger, __):
 
 
 class TestHelperMethods(WebTest):
-    def test_is_many_prefetched(self):
-        evaluation = baker.make(Evaluation)
+    def test_is_prefetched(self):
```
niklasmohrin, line 65: Funnily, `test_logic` is typechecked within, but the calls to it are not, as this does not have `-> None`; doesn't matter to me, but if you want to go all the way, this should probably be added

evap/evaluation/models.py
```diff
@@ -559,6 +559,13 @@ def general_contribution_has_questionnaires(self):
 
     @property
     def all_contributions_have_questionnaires(self):
+        if is_prefetched(self, "contributions"):
+            if not self.contributions:
+                return False
```
niklasmohrin, line 564: Not `True`? The line below is `not contributions....exists()`, so if there are no contributions, we should default to `True`, right?

We could also delegate the universal qualifier to the `all` below by also taking this branch if `not self.contributions`

evap/evaluation/models.py
```diff
@@ -559,6 +559,13 @@ def general_contribution_has_questionnaires(self):
...
+
+            if is_prefetched(self.contributions[0], "questionnaires"):
+                return all(len(contribution.questionnaires) > 0 for contribution in self.contributions)
```
niklasmohrin, line 567: Again, maybe we should go for `not any` to make it easier to verify that the two approaches do the same?

evap/grades/views.py
```diff
@@ -27,11 +27,17 @@ def index(request):
...
-        course_data.append((course, course.midterm_grade_documents.count(), course.final_grade_documents.count()))
+        course_data.append(
+            (
+                course,
+                ilen(gd for gd in course.grade_documents.all() if gd.type == GradeDocument.Type.MIDTERM_GRADES),
+                ilen(gd for gd in course.grade_documents.all() if gd.type == GradeDocument.Type.FINAL_GRADES),
+            )
+        )
 
     return course_data
```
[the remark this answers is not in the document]
He3lixxx: Changed name, added type annotations, changed to use list comprehension.

Yes, we could just annotate the count directly in the database, but I'd reserve specialized query crafting for places where we really need it for performance. This page performs fine, so I'd just keep the code simple.
https://github.com/e-valuation/EvaP/blob/e889dce5b2111f42f3cda777649c19d30de165be/evap/staff/tools.py#L337

---
> In grades_semester_view, we include evaluation_badges, but that doesn't do anything really because no evaluation is set in the template environment 🤔

It adds an empty span if I'm not mistaken (`evaluation_badges.html` line 7). @janno42 is this a bug (missing content) or is this just leftover code?

evap/evaluation/models.py
```diff
@@ -222,6 +222,12 @@ def is_below_contributors(self):
 
     @property
     def can_be_edited_by_manager(self):
+        if is_many_prefetched(self, "contributions"):
+            if all(is_single_prefetched(contribution, "evaluation") for contribution in self.contributions.all()):
+                return all(
+                    contribution.evaluation.state == Evaluation.State.NEW for contribution in self.contributions.all()
+                )
```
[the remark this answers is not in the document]
He3lixxx: I thought about this for a bit and in the end I found the python version here much easier to understand, as I find "all equal" much easier to reason about than "not (all without equal) exists". I don't see a way to simplify the query though.

I had `return not any(unequal new)` here first, but it's still significantly different to `.exclude().exists()`, so then I had to think about each one individually to figure out whether they are correct, so I chose the simpler python version.

But yeah, I see the point. We can also use `return not any(unequal new)`.

evap/evaluation/models.py
He3lixxx: Yeah this is weirdness regarding the general contribution and how a "general contribution does not exist" case should be handled when asking for "do all contributions have questionnaires?".

We have a test that asserts that we `return False` here. If the general contribution does not yet exist, it can't have questionnaires assigned, so callers of this method probably expect `False` as a return value that they likely interpret as "can we proceed with this evaluation?"

This method currently fails this test. I'd wait with further changes here until we have a conclusion on how to handle general contributions in #1839.

evap/grades/views.py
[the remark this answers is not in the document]
janno42: It's a bug, the course-related badges should be shown on this page a well. The badges probably need some refactoring to handle both courses and evaluations.

evap/evaluation/models.py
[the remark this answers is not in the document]
niklasmohrin: we discussed in person that one has to read both versions relatively closely either way to verify that they are the same, so having one simpler version is okay

evap/evaluation/models.py
niklasmohrin: Hm, I think if we want to move forward with this PR, we should just stick to the odd `if not self.general_contribution` for now (and leave a note in the other issue)

niklasmohrin:
Looks good, except for the last comment that blocks it

niklasmohrin (approved):
Looks good :)

PR #1843, opened 2022-12-30
Importers: Use .save() instead of `bulk_` operations to fix #1839

### LINKED ISSUE #1839: Importer doesn't create general contributions

He3lixxx:
Quick fix to get the bug resolved before the long architectural discussion takes place. Use `.save()` in the importers so assert side effects of `.save()` happen.

Fixes #1839, but we'll have to discuss how to handle this in the long term. E.g., the user profile still has this comment:
https://github.com/e-valuation/EvaP/blob/e889dce5b2111f42f3cda777649c19d30de165be/evap/evaluation/models.py#L1556-L1560

### DIFF
--- a/evap/staff/importers/user.py
+++ b/evap/staff/importers/user.py
@@ -385,5 +385,8 @@ def update_existing_and_create_new_user_profiles(
     existing_user_profiles: Iterable[UserProfile],
     new_user_profiles: Iterable[UserProfile],
 ):
-    UserProfile.objects.bulk_update(existing_user_profiles, UserData.bulk_update_fields())
-    UserProfile.objects.bulk_create(new_user_profiles)
+    for user_profile in existing_user_profiles:
+        user_profile.save()
+
+    for user_profile in new_user_profiles:
+        user_profile.save()
--- a/evap/staff/importers/enrollment.py
+++ b/evap/staff/importers/enrollment.py
@@ -723,8 +723,9 @@ def update_existing_and_create_new_courses(
         for course_data in course_data_iterable
         if not course_data.merge_into_course
     ]
-    Course.objects.bulk_create(new_course_objects)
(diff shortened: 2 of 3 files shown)


evap/staff/importers/user.py
```diff
@@ -385,5 +385,8 @@ def update_existing_and_create_new_user_profiles(
     existing_user_profiles: Iterable[UserProfile],
     new_user_profiles: Iterable[UserProfile],
 ):
-    UserProfile.objects.bulk_update(existing_user_profiles, UserData.bulk_update_fields())
```
niklasmohrin, line 388: I suppose `bulk_update_fields` is not needed anymore then? The word "bulk" also occurs in the comment of `UserData.get_user_profile_object`

niklasmohrin (approved):
Looks correct, tests are passing, the word `bulk` appears twice, but not in calls to methods, so should be good

evap/staff/importers/user.py
He3lixxx: Yes. This also applies to the whole `LoggedModel.update_log_after_...` stuff, which I added in #1634 to allow for bulk operations here.

For now, I thought it would be best to leave this in, wait for what we think is a good solution to the general save-sideeffects-vs-bulk-operations problem in #1839, and then maybe (or maybe not) remove them.

PR #1849, opened 2023-01-09
Remove navbar cache

### LINKED ISSUE #1645: The navbar cache is at it again: CSRF super rude failure

He3lixxx:
Removes the navbar cache.

Should thus fix #1645. Increases page load times by ~20ms. I tried
1. just fetching semesters from the DB once instead of twice with different filters (still in this PR)
2. prefetching the groups of `request.user` to allow the privilege-checks to happen without database hits.

Both didn't really change much about the performance. At 40-60ms to render a static view, the effects of removing singular db hits is really hard to qualify. I left the semester logic in, as it is contained, and reverted the user profile groups stuff, as prefetching is a bit tricky to do there: We only want it for `request.user` and not for other users, and ideally, we'd want it directly during initial creation of the object (and not later annotated by a middleware).

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -127,10 +127,6 @@ class Semester(models.Model):
         self.save()
 
     @classmethod
-    def get_all_with_unarchived_results(cls):
-        return cls.objects.filter(results_are_archived=False).distinct()
-
-    @classmethod
     def get_all_with_published_unarchived_results(cls):
         return cls.objects.filter(
             courses__evaluations__state=Evaluation.State.PUBLISHED, results_are_archived=False
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -37,9 +37,7 @@
         {% endblock %}
 
         <div class="sticky-top d-print-none">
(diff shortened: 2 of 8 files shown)


niklasmohrin (approved):
Looks good to me

PR #1852, opened 2023-01-16
Update dependencies

He3lixxx:


evap/staff/views.py
```diff
@@ -216,7 +217,9 @@ class Stats:
                 stats.first_start = min(stats.first_start, evaluation.vote_start_datetime)
                 stats.last_end = max(stats.last_end, evaluation.vote_end_date)
     degree_stats = OrderedDict(sorted(degree_stats.items(), key=lambda x: x[0].order))
-    degree_stats["total"] = total_stats
+
+    degree_stats_with_total: Dict[Union[Degree, str], Stats] = cast(Dict[Union[Degree, str], Stats], degree_stats)
```
niklasmohrin, line 221: The type is correctly assigned after the cast:
```suggestion
    degree_stats_with_total = cast(Dict[Union[Degree, str], Stats], degree_stats)
```

niklasmohrin (commented):
Diff looks good, haven't looked at OIDC changelog yet;

Do you think we will spread `StrOrPromise` throughout the repo now? I hope we can keep it somehow contained

He3lixxx:
@niklasmohrin [OIDC changelog](https://github.com/mozilla/mozilla-django-oidc/blob/main/HISTORY.rst#300-2022-11-14) is very small and shouldn't have anything that affects us.

I don't really know about `StrOrPromise`. For now, it seems the impact is manageable, we only need to annotate lazy translations, and theoreteically we can evaluate them to `str` as soon as we know the target language, which I think most often will be the first usage of a global constant of type `_StrPromise`. As [discussed in django-stubs](https://github.com/typeddjango/django-stubs/pull/1139#issuecomment-1232167698), `str` and `_StrPromise` really aren't safely interchangable, so I guess even if this is somewhat invasive, it's the only way to have type annotations that we can rely on.

niklasmohrin (approved):

niklasmohrin:
If I understand you correctly, we could safely evaluate them to str wherever we would also use gettext non-lazy?

PR #1853, opened 2023-01-16
Cache setup steps in Github actions

### LINKED ISSUE #1749: Investigate Github Action caching
One can cache certain directories in github actions. For Python, caching the entire installation is used (including all the `site-packages`), so that wheels don't need to be rebuild. Additionally, the download could be faster. We should investigate how much there is to gain.

Kakadus:
resolves #1749

### DIFF
--- a/.github/setup_python/action.yml
+++ b/.github/setup_python/action.yml
@@ -1,0 +1,27 @@
+name: "Setup Python"
+description: "Sets up Python for use in actions with caching and copy localsettings"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Setup python
+      uses: actions/setup-python@v4
+      with:
+        python-version: 3.8
+
+    - name: Restore venv cache
+      uses: syphar/restore-virtualenv@v1
+      id: cache-virtualenv
+
+    - name: Restore pip download cache
+      uses: syphar/restore-pip-download-cache@v1
+      if: steps.cache-virtualenv.outputs.cache-hit != 'true'
+
+    - name: Install dependencies
+      run: pip install -r requirements-dev.txt
+      shell: bash
+      if: steps.cache-virtualenv.outputs.cache-hit != 'true'
+
+    - name: Add localsettings
+      run: cp evap/settings_test.py evap/localsettings.py
+      shell: bash
--- a/.github/workflows/debug-tests.yml
+++ b/.github/workflows/debug-tests.yml
@@ -25,14 +25,23 @@ jobs:
     steps:
       - name: Check out repository code
         uses: actions/checkout@v2
-      - name: Setup python with caching
+      - name: Setup python
         uses: actions/setup-python@v4
         with:
           python-version: 3.8
-          cache: pip
-          cache-dependency-path: requirements-dev.txt
+
+      - name: Restore venv cache
+        uses: syphar/restore-virtualenv@v1
+        id: cache-virtualenv
+
+      - name: Restore pip download cache
+        uses: syphar/restore-pip-download-cache@v1
+        if: steps.cache-virtualenv.outputs.cache-hit != 'true'
+
       - name: Install dependencies
         run: pip install -r requirements-dev.txt
+        if: steps.cache-virtualenv.outputs.cache-hit != 'true'
+
       - name: Add localsettings
         run: cp evap/settings_test.py evap/localsettings.py
       - name: Run tests
--- a/.github/workflows/debug-tests.yml
+++ b/.github/workflows/debug-tests.yml
@@ -25,24 +25,9 @@ jobs:
     steps:
       - name: Check out repository code
         uses: actions/checkout@v2
+
       - name: Setup python
-        uses: actions/setup-python@v4
-        with:
-          python-version: 3.8
+        uses: ./.github/setup_python
 
-      - name: Restore venv cache
-        uses: syphar/restore-virtualenv@v1
-        id: cache-virtualenv
-
-      - name: Restore pip download cache
-        uses: syphar/restore-pip-download-cache@v1
-        if: steps.cache-virtualenv.outputs.cache-hit != 'true'
-
-      - name: Install dependencies
-        run: pip install -r requirements-dev.txt
-        if: steps.cache-virtualenv.outputs.cache-hit != 'true'
-
-      - name: Add localsettings
-        run: cp evap/settings_test.py evap/localsettings.py
       - name: Run tests
         run: python manage.py test --debug-mode
--- a/.github/workflows/debug-tests.yml
+++ b/.github/workflows/debug-tests.yml
@@ -24,7 +24,7 @@ jobs:
 
     steps:
       - name: Check out repository code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Setup python
         uses: ./.github/setup_python
--- a/.github/workflows/debug-tests.yml
+++ b/.github/workflows/debug-tests.yml
@@ -9,9 +9,6 @@ jobs:
   tests:
     runs-on: ubuntu-20.04
 
-    container:
-      image: python:3.8
-
     services:
       postgres:
         image: postgres
@@ -28,6 +25,12 @@ jobs:
     steps:
       - name: Check out repository code
         uses: actions/checkout@v2
+      - name: Setup python with caching
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
+          cache: pip
+          cache-dependency-path: requirements-dev.txt
       - name: Install dependencies
         run: pip install -r requirements-dev.txt
       - name: Add localsettings
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,9 +12,6 @@ jobs:
 
     runs-on: ubuntu-20.04
 
-    container:
-      image: python:3.8
-
     services:
       postgres:
         image: postgres
@@ -29,6 +26,14 @@ jobs:
     steps:
       - name: Check out repository code
         uses: actions/checkout@v2
+
+      - name: Setup python with caching
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
+          cache: pip
+          cache-dependency-path: requirements-dev.txt
+
       - name: Install dependencies
         run: pip install -r requirements-dev.txt
       - name: Add localsettings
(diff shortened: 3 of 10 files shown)


Kakadus:
in this run, the caches were missed, because they do not exist (yet). Now they do!

He3lixxx:
So when comparing
https://github.com/e-valuation/EvaP/actions/runs/3934364796/usage (no caches yet)
https://github.com/e-valuation/EvaP/actions/runs/3934396439/usage (cached run 1)
https://github.com/e-valuation/EvaP/actions/runs/3934399599/usage (cached run 2)

We save approximately 20s of computation time, out of 12 minutes total, right? When looking at the actual runs, the step [Install Dependencies](https://github.com/e-valuation/EvaP/actions/runs/3934364796/jobs/6729039052#step:5:1) seems to take 12-15s, with and without the cache.

Is there a way to not have the workflow steps duplicated between the jobs (I think I remember there wasn't one a year ago, but maybe now there is)?

Kakadus:
> We save approximately 20s of computation time, out of 12 minutes total, right? When looking at the actual runs, the step [Install Dependencies](https://github.com/e-valuation/EvaP/actions/runs/3934364796/jobs/6729039052#step:5:1) seems to take 12-15s, with and without the cache.

Yes, I see that this is less then anticipated...

> Is there a way to not have the workflow steps duplicated between the jobs (I think I remember there wasn't one a year ago, but maybe now there is)?

No, there is not. However, I tried  to cache an entire venv now...

deployment/update_production.sh
```diff
@@ -44,8 +44,8 @@ sudo -H -u "$USERNAME" "$ENVDIR/bin/python" manage.py dumpdata --natural-foreign
 sudo -H -u "$USERNAME" "$ENVDIR/bin/pip" install -r requirements.txt
 # sometimes, this fails for some random i18n test translation files.
 sudo -H -u "$USERNAME" "$ENVDIR/bin/python" manage.py compilemessages || true
-sudo -H -u "$USERNAME" "$ENVDIR/bin/python" manage.py scss --production
-sudo -H -u "$USERNAME" "$ENVDIR/bin/python" manage.py ts compile --fresh
+sudo -H -u "$USERNAME" "$ENVDIR/bin/python" manage.py scss --production || true
+sudo -H -u "$USERNAME" "$ENVDIR/bin/python" manage.py ts compile --fresh || true
```
Kakadus, line 48: These lines always fail. Note, that even before this PR:
> Could not find sass command
and 
> Could not find npx command
Now, the returncode is actually not zero and more prominent error messages are shown

deployment/load_production_backup.sh
```diff
@@ -47,7 +47,7 @@ sudo -H -u "$USERNAME" "$ENVDIR/bin/pip" install -r requirements.txt
 
 # sometimes, this fails for some random i18n test translation files.
 sudo -H -u "$USERNAME" "$ENVDIR/bin/python" manage.py compilemessages || true
-sudo -H -u "$USERNAME" "$ENVDIR/bin/python" manage.py scss --production
+sudo -H -u "$USERNAME" "$ENVDIR/bin/python" manage.py scss --production || true
```
Kakadus, line 50: The same applies here

deployment/update_production.sh
He3lixxx: I think I'd rather have npm installed here. @niklasmohrin what do you think?

niklasmohrin: Yeah, everything `*production` should fail eagerly

deployment/update_production.sh
```diff
@@ -43,7 +43,7 @@ sudo -H -u "$USERNAME" "$ENVDIR/bin/python" manage.py dumpdata --natural-foreign
 
 sudo -H -u "$USERNAME" "$ENVDIR/bin/pip" install -r requirements.txt
 # sometimes, this fails for some random i18n test translation files.
-sudo -H -u "$USERNAME" "$ENVDIR/bin/python" manage.py compilemessages || true
+sudo -H -u "$USERNAME" "$ENVDIR/bin/python" manage.py compilemessages
 sudo -H -u "$USERNAME" "$ENVDIR/bin/python" manage.py scss --production
 sudo -H -u "$USERNAME" "$ENVDIR/bin/python" manage.py ts compile --fresh
```
He3lixxx, line 48: There's still the discussion about the `|| true`s here, right?

@Kakadus did I understand you correctly that some steps failed here in the past, but the "Test update" workflow still passed?
This feels weird to me: This has `set -e` to abort on error. So, the test workflow should have a step with a non-zero return code, where it should stop and fail. Why did this not happen?

evap/settings_test.py
```diff
@@ -5,22 +5,22 @@
         "NAME": "evap",
         "USER": "postgres",
         "PASSWORD": "postgres",
-        "HOST": "postgres",
+        "HOST": "localhost",
```
He3lixxx, line 8: Was this necessary? I think the dedicated per-service hostname has the advantage that in theory, the workflow could use a "service" hosted in a different container. We don't currently use that, but I wouldn't pessimize without a reason :D

He3lixxx (commented):
Code looks good. Not yet approving for the `|| true` discussion, I'd like to have the CI workflow fail if the script doesn't exit with a zero status code.

.github/workflows/tests.yml
```diff
@@ -2,8 +2,6 @@ name: EvaP Test Suite
 
 on:
   push:
-    branches:
-      - main
```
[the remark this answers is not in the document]
Kakadus: Yes, you are right. I removed these lines to test the actions on the branch in my fork and forgot to restore it. Your point seems reasonable though

deployment/update_production.sh
Kakadus: Yes. when you look at [this backup process run](https://github.com/e-valuation/EvaP/actions/runs/3934713350/jobs/6729702916#step:8:59), the `scss` and `ts complie` command actually "fail", but still the return code must be zero. The previous version use a python container whereas now I use the setup-python action. This is probably due to some github action internal. However, I will try to investigate

evap/settings_test.py
Kakadus: Yes it is. When no container: is specified, the steps are runned directly on the runner, but postgres and redis still inside a docker container, with the defined port mapping. So, the jobs needs to reference localhost to access the docker containers.
(see [here](https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers#running-jobs-directly-on-the-runner-machine))

.github/workflows/debug-tests.yml
```diff
@@ -27,10 +24,10 @@ jobs:
...
       - name: Check out repository code
-        uses: actions/checkout@v2
-      - name: Install dependencies
-        run: pip install -r requirements-dev.txt
-      - name: Add localsettings
-        run: cp evap/settings_test.py evap/localsettings.py
+        uses: actions/checkout@v3
+
+      - name: Setup python
+        uses: ./.github/setup_python
```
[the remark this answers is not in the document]
Kakadus: Yes, that would be very neat. Sadly, this is not the case. If I omit the `name`, then the name is set to `Run /./.github/setup_python`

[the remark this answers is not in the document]
Kakadus: However, the step names inside the composite actions do not matter and are not shown in the logs. Maybe we can therefore remove some names, which follow trivially from the used action (e. g. actions/setup_python)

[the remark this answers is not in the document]
He3lixxx: Ah alright. The current version is fine for me, then.

.github/workflows/tests.yml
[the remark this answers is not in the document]
He3lixxx: I checked again: Translation commits are always pushed to main, so that already triggers a CI run, so we are fine with the old version. So I'd suggest just adding these two lines back.

evap/evaluation/management/commands/ts.py
```diff
@@ -54,8 +54,8 @@ def handle(self, *args, **options):
     def run_command(self, command):
         try:
             subprocess.run(command, check=True)  # nosec
-        except FileNotFoundError:
-            print(f"Could not find {command[0]} command", file=self.stderr)
```
Kakadus, line 58: As it turns out, the issue was these lines in `ts.py` and `scss.py`, as they merely printed the error without changing the return code

Kakadus: I raise a CommandError instead and adopted the tests accordingly

.github/workflows/tests.yml
```diff
@@ -258,24 +245,22 @@ jobs:
 
     steps:
       - name: Check out repository code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           submodules: true
-      - name: Install packages
-        run: sudo apt-get install xvfb libgbm1
```
niklasmohrin, line 265: Does anyone know why these aren't needed anymore? (same for `sudo` above)

.github/setup_python/action.yml
```diff
@@ -0,0 +1,27 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Setup python
+      uses: actions/setup-python@v4
+      with:
+        python-version: 3.8
+
+    - name: Restore venv cache
+      uses: syphar/restore-virtualenv@v1
```
niklasmohrin, line 13: A thought: Should these versions be pinned? I suppose there isn't anything bad that could happen if the author turns malicious here, so its probably fine to leave it at `v1`, but what do the others think?

He3lixxx (approved):

.github/setup_python/action.yml
He3lixxx: I'm fine with them not being pinned

.github/workflows/tests.yml
Kakadus: For `sudo` and `xvfb`, they are included in the ubuntu image, see [here](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md#installed-apt-packages) for a list of "all" installed apt packages. While `libgdm1` is not listed specifically, pupeeter works with the  `libgdm-dev`, which is listed. If I remember, correctly, `libgdm1` was actually installed in the image, as I removed that line

niklasmohrin (approved):
Very nicely done!

PR #1855, opened 2023-01-23
Importer improvements

### LINKED ISSUE #1574: Import data test improvements

He3lixxx:
Fixes #1574. Best served by commit.


He3lixxx:
@niklasmohrin mypy actually caught an error here that tests did not cover and we might have missed: In the checker, I first forgot to ignore rows where parsing the degree had failed (-> InvalidValue). The tests that went this path never checked whether any unexpected errors occured, so they passed because their expected error on degree parsing existed, and they simply ignored the unexpected error.

I've made the tests more assertive regarding unexpected errors in the last commit. Removing the branch in the checker now also causes a few tests to fail.

evap/staff/tests/test_importers.py
```diff
@@ -39,7 +40,27 @@ def test_skip_first_n_rows_handled_correctly(self):
         self.assertEqual(rows[0].value, "3")
 
 
-class TestUserImport(TestCase):
+class ImporterTestCase(TestCase):
+    # pylint: disable=invalid-name
```
He3lixxx, line 44: probably makes more sense to just allow `assertCamelCase` globally

evap/staff/importers/enrollment.py
```diff
@@ -555,6 +588,40 @@ def finalize(self) -> None:
...
 
 
+class UserDegreeMismatchChecker(Checker, RowCheckerMixin):
+    """Assert that a users degree is consistent between rows"""
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+
+        self.degree_by_email: Dict[str, Degree] = {}
+        self.tracker: FirstLocationAndCountTracker = FirstLocationAndCountTracker()
```
niklasmohrin, line 598: Not sure what style we want to prefer, but I would lean towards omitting type annotations that can be inferred:
```suggestion
        self.tracker = FirstLocationAndCountTracker()
```
How do you feel about it?

evap/staff/importers/enrollment.py
```diff
@@ -555,6 +588,40 @@ def finalize(self) -> None:
...
+        degree = next(iter(row.course_data.degrees))
+        stored_degree = self.degree_by_email.setdefault(row.student_data.email, degree)
+
+        if stored_degree != degree:
+            self.tracker.add_location_for_key(row.location, row.student_data.email)
+
+    def finalize(self) -> None:
+        for student_email, location_string in self.tracker.aggregated_keys_and_location_strings():
+            self.importer_log.add_error(
+                _('{location}: The user "{email}" has a different degree than in a previous row.').format(
```
niklasmohrin, line 617: I can't pinpoint it, but this sounds a bit odd. How about the following?

> The degree of user email differs from a previous row.

evap/staff/importers/base.py
```diff
@@ -274,6 +276,9 @@ def aggregated_keys_and_location_strings(self) -> Iterator[Tuple[Any, str]]:
 
             yield key, location_string
 
+    def keys(self):
```
niklasmohrin, line 279:
```suggestion
    def keys(self) -> Iterable[Any]:
```

evap/staff/importers/enrollment.py
```diff
@@ -479,6 +472,46 @@ def finalize(self) -> None:
...
+    """
+    Searches for courses that have names with small edit distance and warns about them to make users aware of possible
+    typos.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.course_en_tracker: FirstLocationAndCountTracker = FirstLocationAndCountTracker()
+        self.course_de_tracker: FirstLocationAndCountTracker = FirstLocationAndCountTracker()
```
niklasmohrin, line 485:
```suggestion
        self.course_en_tracker = FirstLocationAndCountTracker()
        self.course_de_tracker = FirstLocationAndCountTracker()
```

evap/staff/tests/test_importers.py
```diff
@@ -775,8 +819,38 @@ def test_existing_participation(self):
         )
         self.assertFalse(importer_log.has_errors())
 
+    @override_settings(IMPORTER_COURSE_NAME_SIMILARITY_WARNING_THRESHOLD=0.8)
+    def test_course_name_with_typo(self):
+        # Add a typo in one english course name as well
```
niklasmohrin, line 824: Should we test that this doesn't occur without the additional typo? Or is that covered by the other tests?

niklasmohrin (commented):
Looks good, some comments:

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

evap/staff/tests/test_importers.py
He3lixxx: Hmm, good question, idk. This test is the only one A) with the typo in one line and B) with the lowered threshold, so yeah, it probably makes sense to assert that the import passes without the typo.

He3lixxx: However, this test asserts that the warning is triggered for exactly three out of the 35 lines, so at the same time it asserts that the other 32 lines pass without a warning, so I think the current test logic is fine. What do you think?

pyproject.toml
```diff
@@ -31,6 +31,7 @@ django-settings-module = "evap.settings"
 # need to allow that here (their issue 618)
 variable-rgx = "(^[a-z0-9_]+$)|(^[A-Za-z]+$)"
 argument-rgx = "(^[a-z0-9_]+$)|(^[A-Za-z]+$)"
+method-rgx = "(^[a-z0-9_]+$)|(^[A-Za-z]+$)"
```
niklasmohrin, line 34: Hm, this also allows non-assert camel case names, should we not disallow those?

evap/staff/tests/test_importers.py
```diff
@@ -775,8 +818,38 @@ def test_existing_participation(self):
...
+        input_data["MA Belegungen"][1][6] = "BauenWithTypo"
+
+        excel_content = excel_data.create_memory_excel_file(input_data)
+
+        args = (excel_content, self.semester, self.vote_start_datetime, self.vote_end_date)
+        importer_log_test = import_enrollments(*args, test_run=True)
+        importer_log_notest = import_enrollments(*args, test_run=False)
+
+        self.assertEqual(importer_log_test.warnings_by_category(), importer_log_notest.warnings_by_category())
+        self.assertListEqual(
```
niklasmohrin, line 836:
```suggestion
        self.assertEqual(
```

... and this doesn't use `assertErrorsAre` because there could be more that we don't care about, correct?

evap/staff/importers/enrollment.py
```diff
@@ -83,9 +84,9 @@ def all_fields_valid(course_data: CourseData) -> TypeGuard[ValidCourseData]:
 
 
 class DegreeImportMapper:
-    class InvalidDegreeNamesException(Exception):
-        def __init__(self, *args, invalid_degree_names: Set[str], **kwargs):
-            self.invalid_degree_names = invalid_degree_names
+    class InvalidDegreeNameException(Exception):
+        def __init__(self, *args, invalid_degree_name: str, **kwargs):
+            self.invalid_degree_name: str = invalid_degree_name
```
niklasmohrin, line 89:
```suggestion
            self.invalid_degree_name = invalid_degree_name
```

evap/staff/importers/base.py
```diff
@@ -252,7 +254,7 @@ class FirstLocationAndCountTracker:
 
     def __init__(self, *args, **kwargs) -> None:
         self.first_location_by_key: Dict[Any, ExcelFileLocation] = {}
-        self.location_count_by_key: Counter = Counter()
+        self.location_count_by_key = Counter()
```
niklasmohrin, line 257: Mypy fails here, which is odd because `reveal_type` correctly identifies this as `Counter[Any]` :thinking:

niklasmohrin (approved):
Looks good, final remarks:

evap/staff/tests/test_importers.py
niklasmohrin: Fair enough, even within the six names that do collide, there aren't any collisions outside of the pairs, so that should suffice

evap/staff/tests/test_importers.py
He3lixxx: Hmm, yeah, I thought about adding `assertWarningAre`, but I think with warnings the use case is a bit different. With errors, we often want to assert that no other errors happens, especially that no "Import aborted after uncaught exception" error is generated.
With warnings, I'd say we most often don't really care if other warnings trigger. I wouldn't want this test to break if we add a "suspiciously low enrollment count" warning or whatever.

Thoughts / fine?

PR #1858, opened 2023-02-04
Bootstrap 5.3.0

janno42:
Update to Bootstrap 5.3.0-alpha1 and minor improvements to the vote buttons by adding a hover effect on active buttons.
This is a preparation for further changes and should not greatly change the visual appearance.

Please double-check for any unexpected UI impact.

### DIFF
--- a/evap/staff/templates/staff_evaluation_person_management.html
+++ b/evap/staff/templates/staff_evaluation_person_management.html
@@ -115,10 +115,12 @@
                         <p class="card-text">{% trans 'This will create a CSV file containing login keys for all external participants.' %}</p>
                     </div>
                     <div class="card-footer text-center">
-                        <button name="operation" value="export-login-keys" type="submit" class="btn btn-sm btn-primary form-submit-btn"
-                        {% if not evaluation.has_external_participant %} disabled data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans "This evaluation has no external participants." %}"{% endif %}>
-                            {% trans 'Export login keys' %}
-                        </button>
+                        <div{% if not evaluation.has_external_participant %} data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans "This evaluation has no external participants." %}"{% endif %}>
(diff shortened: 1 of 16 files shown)


evap/staff/templates/staff_evaluation_person_management.html
```diff
@@ -115,10 +115,12 @@ <h6 class="card-subtitle mb-2 text-muted">{% trans 'To CSV file' %}</h6>
                         <p class="card-text">{% trans 'This will create a CSV file containing login keys for all external participants.' %}</p>
                     </div>
                     <div class="card-footer text-center">
-                        <button name="operation" value="export-login-keys" type="submit" class="btn btn-sm btn-primary form-submit-btn"
-                        {% if not evaluation.has_external_participant %} disabled data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans "This evaluation has no external participants." %}"{% endif %}>
-                            {% trans 'Export login keys' %}
-                        </button>
+                        <div{% if not evaluation.has_external_participant %} data-bs-toggle="tooltip" data-bs-placement="top" title="{% trans "This evaluation has no external participants." %}"{% endif %}>
+                            <button name="operation" value="export-login-keys" type="submit" class="btn btn-sm btn-primary form-submit-btn"
```
He3lixxx, line 119: Feels to me like this is worse HTML because the tooltip semantically belongs to the button, doesn't it? Why is the additional `div` necessary?

He3lixxx (commented):
Code looks good, haven't tested in a browser.

evap/staff/templates/staff_evaluation_person_management.html
janno42: Tooltips can't be activated on disabled elements.

He3lixxx:
UI seems okay for me after quickly browsing through all the pages. I had a bit of trouble when switching to your branch with scss compilation not working properly even after a the first `vagrant destroy && vagrant up`, but the second time it worked, so :shrug:.

If you rebase on main, CI should pass.

He3lixxx (approved):
(Keep in mind it probably makes sense to update the @types/bootstrap js dependency)

PR #1870, opened 2023-02-06
Work around CI failure due to git repo ownership

He3lixxx:
hopefully...


niklasmohrin (approved):
I suppose this works, should we make a plan to check back on this some time in the future?

PR #1872, opened 2023-02-13
Fix navbar semester ordering

He3lixxx:
In #1849, I added two sets to the `include_navbar` template tag. This messes up the ordering of the semesters shown in the navbar. This commit fixes it to be lists instead.

@niklasmohrin do we want a test here? On one hand, the default ordering is the right one here, so we would be explicitly testing that we're not changing the order. On the other hand, we got it wrong.


janno42 (approved):
✔️ UI functionality checked

i'd say we don't really need a test here. but go ahead if you like :)

PR #1881, opened 2023-02-13
Remove flaky contact modal tests

### LINKED ISSUE #1848: Frontend test `contact-modal-opens` flaky
In #1809, I attempted to add a [frontend test](https://github.com/e-valuation/EvaP/blob/main/evap/static/ts/tests/frontend/contributor-evaluation-edit.ts) to make sure the issue doesn't reappear. Despite best efforts to write the test in such a way that CSS animations do not interfere with the test logic, the test still regularly fails in CI.

I don't know what to do about it apart from disabling/removing the test.

He3lixxx:
Fixes #1848.


niklasmohrin:
Do we not want to skip instead of removing it?

He3lixxx:
I'd say we don't want dead code in the repo. It's not like we're waiting for an update on a dependency or something, we just can't get it to work. If we disable it, it will likely just rot along untested and thus stop working for many other reasons. Would you prefer skipping/disabling?

niklasmohrin:
My thought was that we will see "1 skipped" and not forget about it, but we can also open an issue or so. Dead code is also a point, I am fine with either of the two options

niklasmohrin (approved):

PR #1889, opened 2023-02-27
Use bootstrap grid for staff evaluation

### LINKED ISSUE #1818: Replace textanswer review HTML `table` with CSS grid
HTML tables are a bit icky, for example, you cannot insert forms (or other tags) between table rows, cells and whatnot. Nowadays, table-like layouts can be achieved with [CSS grid](https://developer.mozilla.org/en-US/docs/Glossary/Grid).

To start out, the textanswer review table (defined in `evap/staff/templates/staff_evaluation_textanswers_section.html`) should be adapted.

Afterwards, we can investigate what is needed to port the more complicated tables such as the student index - but in another PR ;)

FranzSw:
This is the ~"low-effort" version~ high-effort approach, using bootstrap classes. This works well, however, aligning the columns to be the same size is not really possible with the default classes (atleast I didnt find the ones with the right width for this table). With dynamic sizing, this only shifts the second title "Entscheidung" to the right.

**Before:**
image
**After:**
image

closes #1818

### DIFF
--- a/evap/static/scss/components/_grid.scss
+++ b/evap/static/scss/components/_grid.scss
@@ -2,6 +2,11 @@
     background-color: $table-striped-bg;
 }
 
+// CSS Grid needs it the exact other way around as the title also has the row class
+.grid-striped .row:nth-of-type(even) {
+    background-color: $table-striped-bg;
+}
+
 .lcr-left {
     flex: 1;
 }
@@ -43,6 +48,10 @@
         padding-top: 0.75rem;
         padding-bottom: 0.75rem;
     }
+}
+
+.col-fit-content {
+    flex: 0 0 fit-content;
 }
 
 %grid-row {
--- a/evap/static/scss/components/_grid.scss
+++ b/evap/static/scss/components/_grid.scss
@@ -132,7 +132,11 @@
 
     grid:
         "answer edit review flag"
-        / auto  3rem 310px  5.5rem;
+        / auto  2rem 285px  4.5rem;
+
+    min-height: 0;
+    padding: 0.5rem;
+    gap: 1rem;
 }
 
 @each $col in name, semester, responsible, participants, result, answer, edit, review, flag {
--- a/evap/static/scss/components/_grid.scss
+++ b/evap/static/scss/components/_grid.scss
@@ -2,9 +2,14 @@
     background-color: $table-striped-bg;
 }
 
-// CSS Grid needs it the exact other way around as the title also has the row class
-.grid-striped .row:nth-of-type(even) {
+.grid-row {
+    display: contents;
+}
+
+.grid-striped .grid-row:nth-of-type(even) > div {
     background-color: $table-striped-bg;
+    border-top: 1px solid $table-border-color;
+    border-bottom: 1px solid $table-border-color;
 }
 
 .lcr-left {
@@ -127,16 +132,16 @@
     }
 }
 
-.textanswer-review-grid-row {
+.textanswer-review-grid {
     @extend %grid-row;
 
+    gap: 0;
     grid:
         "answer edit review flag"
-        / auto  2rem 285px  4.5rem;
+        / auto  min-content min-content min-content;
 
     min-height: 0;
     padding: 0.5rem;
-    gap: 1rem;
 }
 
 @each $col in name, semester, responsible, participants, result, answer, edit, review, flag {
--- a/evap/static/scss/components/_grid.scss
+++ b/evap/static/scss/components/_grid.scss
@@ -50,10 +50,6 @@
     }
 }
 
-.col-fit-content {
-    flex: 0 0 fit-content;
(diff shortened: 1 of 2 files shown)


He3lixxx (approved):
Code looks good to me, screenshot is fine, haven't tested manually.

janno42 (commented):
The problem is that all rows, including the header, are sized independently. If the edit button column would have a header text, the layout would break. Also, if in some rows the button group would be smaller, the edit button would move to the right and also break the layout.
The current approach is probably not what we should use. Do you have any further ideas?

FranzSw:
@janno42 Now its not a bootstrap grid anymore, but the "datagrid", which is also used for e.g. the grades semester view. It should allow for absolute / equal sizing of all columns.

janno42 (commented):
I pushed a commit with suggestions for less padding into your branch. Before there was a lot of empty space.

Is there a way to use `min-content` for the last three columns' widths? It would be perfect if we don't have to specify their individual widths.

janno42 (commented):
As discussed, we could move the padding (and other classes) into the scss, making it easier to work with the template. I pushed a suggestion, what do you think of it?

When testing, I noticed that the original textanswer after editing is not correctly placed. It should be under the new answer, not beside it. Can you please have a look into it?

FranzSw:
@janno42 Original textanswer should now be shown below the new one :)

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

He3lixxx (approved):

evap/static/scss/components/_grid.scss
```diff
@@ -2,6 +2,16 @@
     background-color: $table-striped-bg;
 }
 
+.grid-row {
+    display: contents;
+}
```
niklasmohrin, line 7: This seems a bit more global than I understand it to be, I wonder if we should have it nested in another selector, but there isn't really any particular one that comes to mind. Do you think that in any kind of grid layout (in the context of EvaP) we would want this display type on the row, or is it more incidentally needed here?

evap/static/scss/components/_grid.scss
```diff
@@ -2,6 +2,16 @@
...
+
+.grid-striped .grid-row:nth-of-type(even) > div {
+    background-color: $table-striped-bg;
+    border-top: 1px solid $table-border-color;
+    border-bottom: 1px solid $table-border-color;
```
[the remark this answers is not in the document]
niklasmohrin: Or even clearer: Shouldn't all grid rows just have a bottom border instead of every second one having top and bottom?

evap/static/scss/components/_grid.scss
```diff
@@ -122,7 +132,30 @@
     }
 }
 
-@each $col in name, semester, responsible, participants, result {
+.textanswer-review-grid {
+    @extend %grid-row;
```
niklasmohrin, line 136: This seems a bit sus, the grid container specializes the "grid row" concept :thinking: I suppose it's an artifact of us using one grid container per row before? I am also not sure which rules we need exactly from that, `display: grid` for sure and maybe the `media-breakpoint` business, but the other stuff is overwritten anyways.

Maybe this PR isn't the right place to dig into this and find clarity, but if we are aiming for CSS grid to replace more tables on the site, we should disentangle this at some point.

evap/static/scss/components/_grid.scss
```diff
@@ -122,7 +132,30 @@
...
+
+    .grid-row > div {
+        height: 100%;
+        padding: 0.5rem;
+
+        &:not(:first-child) {
+            display: flex;
+            align-items: center;
+            padding-left: 0.75rem;
+        }
```
niklasmohrin, line 154: The HTML has this information duplicated as the `div` that represent the cells of the table inside a `div.grid-row` have a class of either `justify-content-left` or `justify-content-center`. I do not strongly prefer either place for having this information, but if possible, I would not want it duplicated. Do you think we can achieve this?

niklasmohrin (approved):
Some final thoughts, nothing crucial, but I would prefer to have a short discussion to be on the same page about each of the points before merging

evap/static/scss/components/_grid.scss
FranzSw: I added the selector, so we can just rename it.

The issue is, that a grid does not allow nested elements, e.g. we have the table
```
A B C
D E F
```
Now we want the table to be striped, so DEF should get a gray background.
To do that, we need an element grouping the rows, so that we can select `:nth-child(2)`. but, with
```
<div>A B C</div>
<div>D E F</div>
```
the grid breaks.

The only solution is to apply `display: contents` to the wrapping element.

Therefore, in any table using CSS grid, the row-element should get `display: contents`

niklasmohrin (approved):
Thanks :)

Issue #1820, opened 2022-10-10
Navbar switch button tooltips


janno42:
The language buttons (EN/DE) and mode buttons (regular mode/staff mode) in the navbar should have styled bootstrap tooltips. The tooltips should be shown for enabled and disabled buttons.
Currently, only the enabled mode button has a correctly styled tooltip.

islpriem:
The tooltips are deactivated by default for disabled buttons for a reason. On the one hand, the user cannot press the corresponding buttons anyway, and on the other hand, it seems rather confusing to label them according to the motto "if this button were clickable, you could perform the following action". In my opinion the tooltips for disabled buttons would be rather confusing and are not needed.

janno42:
The current behavior is inconsistent in two ways:
- The language button's tooltips are not correctly styled.
- In other places, we also have tooltips for the currently selected option (see, e.g., the questionnaire list in staff mode).

To me, it seems useful to also explain what the current selection is in case the user isn't sure. So the tooltips should be shown regardless of the state.

PR #1894, opened 2023-03-10
Added tooltip for language slider in navbar

mohamedtimes2:
Added the additional attribute to enable the Bootstrap tooltip on the language slider, this only shows the disabled option similar to the other sliders in the navbar.

DE


janno42 (changes requested):
Hi, thanks for your contribution! I assume this is connected to issue #1820.

You correctly applied the bootstrap styling, but tooltips on the disabled buttons are not working yet. The reason for this is explained in the Bootstrap docs (https://getbootstrap.com/docs/5.3/components/tooltips/#disabled-elements).
Do you want to have another look at it?

janno42:
superseded by #1931

PR #1899, opened 2023-03-13
Isolate node_modules inside development container/VM

### LINKED ISSUE #1648: VM provisioning fails on windows hosts because symlinks can not be created

He3lixxx:
Attempt to fix #1648.

Would be nice if some of you could test it out -- inside the VM, you should have `node_modules`, but outside, the folder should simply be empty (or have whatever contents you want it to have)

I will have to test around with the setup a bit more, especially trying to make it break when having `node_modules` populated both inside and outside the VM. In theory, it should work.

We probably also want to document the docker flags used.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -23,6 +23,8 @@ Vagrant.configure("2") do |config|
     # Docker container really are supposed to be used differently. Hacky way to make it into a "VM".
     d.cmd = ["tail", "-f", "/dev/null"]
 
+    d.create_args = ["--cap-add=SYS_ADMIN", "--security-opt=apparmor:unconfined"]
+
     # Workaround for no SSH server as long as https://github.com/hashicorp/vagrant/issues/8145 is still open
     override.trigger.before :provision do |trigger|
       trigger.ruby do |env, machine| system("vagrant docker-exec -it -- /evap/deployment/provision_vagrant_vm.sh") end
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -23,6 +23,7 @@ Vagrant.configure("2") do |config|
     # Docker container really are supposed to be used differently. Hacky way to make it into a "VM".
     d.cmd = ["tail", "-f", "/dev/null"]
 
+    # Required so we can use mount inside the VM -- see e.g. https://github.com/moby/moby/issues/16429
     d.create_args = ["--cap-add=SYS_ADMIN", "--security-opt=apparmor:unconfined"]
 
     # Workaround for no SSH server as long as https://github.com/hashicorp/vagrant/issues/8145 is still open
--- a/README.md
+++ b/README.md
@@ -27,8 +27,6 @@ The easiest setup using [Vagrant](https://www.vagrantup.com) is shown here.
      git config --global core.autocrlf input
      ```
 
-   * Symlink Privileges: Our setup script for the VM creates symlinks in the repository folder. This requires either [explicitly allowing your user account to create symlinks](https://superuser.com/a/105381) or simply running the commands in step 3 as administrator. Thus, we suggest doing step 3 in a Git Bash that was started using "Run as administrator". Generally, this is only required for the first time executing `vagrant up`.
-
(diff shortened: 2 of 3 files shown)


He3lixxx:
@niklasmohrin suggested to just use npm's `--global` flag, but it doesn't work as documented. I've opened [an issue with npm](https://github.com/npm/cli/issues/6319). So this is now waiting to see how the issue resolves.

Vagrantfile
```diff
@@ -23,6 +23,8 @@ Vagrant.configure("2") do |config|
     # Docker container really are supposed to be used differently. Hacky way to make it into a "VM".
     d.cmd = ["tail", "-f", "/dev/null"]
 
+    d.create_args = ["--cap-add=SYS_ADMIN", "--security-opt=apparmor:unconfined"]
```
niklasmohrin, line 26: Let's add a comment about why this is needed :)

janno42 (approved):

niklasmohrin:
Works on my machine, although its a bit odd that I still get the empty `node_modules` on the host machine. Still, this is definitely better than the setup-as-admin we currently have :+1:

niklasmohrin (approved):

Issue #1901, opened 2023-03-14
Document for EvaP Installation and Set-Up for MacOS Silicon (M1/M2) System Architecture


omrichashper:
I am a student at Drexel University, and took a course that exposes me to the World of FOSS Projects such as this one. In the second half of the term, my group and I decided to make contributions to EvaP, when unexpectedly our group members with newer Macs realized that we cannot install/set-up EvaP on our machines, for several reasons. 

The biggest reasons were:
1) VirtualBox does not have a release that supports MacOS Silicon System Architecture
2) The three Docker commands identified in the readme.md file did not work
3) Gaps in steps/knowledge between information provided in the readme.md and Installation Docs in the Wiki

As a result, we all broke off to find work arounds, and after some time I successfully installed, set-up, and ran the project on my local machine with Docker as the provider. After my experience, I felt that it was necessary to compose a document with easy-to-follow, step-by-step instructions on how to get EvaP up and running on the newer Macs.

You can find the document in my forked repository's wiki: [Installation & Setup for MacOS Silicon (M1 and M2) System Architecture](https://github.com/omrichashper/EvaP/wiki/Installation-&-Setup-for-MacOS-Silicon-(M1-and-M2)-System-Architecture)

Please consider reviewing my document and merging it into your Wiki. I believe that with this document in your deck, you will be able to bring in more contributors with Macs released after the year 2020, as they will have a much easier time getting the software on their machines. I have tested this document with my peers in my group, and have even provided it to the other group in my class that are also looking to contribute to EvaP (who communicated that they also were unable to install/set-up EvaP on their computers). I hope this document serves you well, and thanks for considering!

He3lixxx:
Hi @omrichashper, thanks for opening this issue.

tldr: Could you elaborate on your second point, as this is most likely what we want to fix here:
> 2. The three Docker commands identified in the readme.md file did not work

What exactly did not work, what errors occurred at what step? I'd guess most likely something in `vagrant provision` failed, can you see which command caused the issue there?

---

#### Regarding 1
[...]

PR #1902, opened 2023-03-15
macOS setup in readme

fidoriel:
as discussed in https://github.com/e-valuation/EvaP/issues/1901

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,16 +31,14 @@ Some CPUs like Apples M1/M2 are not supported by VirtualBox, please follow the s
    * Symlink Privileges: Our setup script for the VM creates symlinks in the repository folder. This requires either [explicitly allowing your user account to create symlinks](https://superuser.com/a/105381) or simply running the commands in step 3 as administrator. Thus, we suggest doing step 3 in a Git Bash that was started using "Run as administrator". Generally, this is only required for the first time executing `vagrant up`.
 
 3. Run the following commands on the command line to clone the repository, create the Vagrant VM and run the Django development server:
-```bash
-git clone --recurse-submodules https://github.com/<your_github_username>/EvaP.git
-cd EvaP
-```
-For VirtualBox just use `vagrant up` and for Docker `vagrant up --provider docker && vagrant provision` to build and start the vm.
-Contiune for Docker and VirtualBox with:
-```bash
-vagrant ssh
-./manage.py run
-```
+    Note: For Docker please replace `vagrant up` with `vagrant up --provider docker && vagrant provision`.
+    ```bash
+    git clone --recurse-submodules https://github.com/<your_github_username>/EvaP.git
+    cd EvaP
+    vagrant up
+    vagrant ssh
+    ./manage.py run
+    ```
 
 4. Open your browser at http://localhost:8000/ and login with email ``evap@institution.example.com`` and password ``evap``
 
--- a/README.md
+++ b/README.md
@@ -14,10 +14,9 @@ For the documentation, please see our [wiki](https://github.com/e-valuation/EvaP
 
 ## Installation
 
-The easiest setup using [Vagrant](https://www.vagrantup.com) and [VirtualBox](https://www.virtualbox.org) is shown here. 
-Some CPUs like Apples M1/M2 are not supported by VirtualBox, please follow the steps for docker. For manual installation instructions and production deployment please see the [wiki page on installation](https://github.com/e-valuation/EvaP/wiki/Installation).
+The easiest setup using [Vagrant](https://www.vagrantup.com) and [VirtualBox](https://www.virtualbox.org) is shown here for ARM systems please use Docker.
 
-0. Install [git](https://git-scm.com/downloads), [Vagrant](https://www.vagrantup.com/downloads.html) and one of [VirtualBox](https://www.virtualbox.org/wiki/Downloads) or [Docker Engine](https://docs.docker.com/engine/install/)
+0. Install [git](https://git-scm.com/downloads), [Vagrant](https://www.vagrantup.com/downloads.html) and one of [VirtualBox](https://www.virtualbox.org/wiki/Downloads) or [Docker](https://docs.docker.com/engine/install/)
 
 1. Fork the Evap repository (using the Fork-button in the upper right corner on GitHub)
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -44,14 +45,22 @@ vagrant ssh
 That's it!
 
 
-### Docker
+### Docker Setup
 We provide experimental support to run the development setup in a docker container using
+Please install Docker Desktop, **not** just the CLI docker engine. Vagrant does conflict with used ports from other docker containers, even if they are not in use. If you are encountering this issue during the build process you can change the `host` port in `Vagrantfile`.
```
He3lixxx, line 50: > Please install Docker Desktop, **not** just the CLI docker engine

Do we have any kind of reference / documentation / a vagrant issue for this? If vagrant really doesn't work with the docker CLI engine, it is not documented, and there is no issue, then we should open one. But I really doubt it, I wouldn't expect vagrant to interact with any of the desktop components, it just executes some docker commands. (It certainly is possible that you'd need to restart the terminal or reboot after installing docker to update environment variables / user groups / whatever).

> Vagrant does conflict with used ports from other docker containers, even if they are not in use

I don't think vagrant has anything to do with ports here. Do you mean: "If other docker containers on the system are configured to also use port 8000, this causes conflicts"? In that case, I'd expect users to realize this, since they are probably aware of the containers they have on their system. For a fresh docker installation, port 8000 shouldn't be blocked, right?
-> I'd vote to remove this.

README.md
```diff
@@ -14,7 +14,8 @@ For the documentation, please see our [wiki](https://github.com/e-valuation/EvaP
 
 ## Installation
 
-The easiest setup using [Vagrant](https://www.vagrantup.com) and [VirtualBox](https://www.virtualbox.org) is shown here. For manual installation instructions and production deployment please see the [wiki page on installation](https://github.com/e-valuation/EvaP/wiki/Installation).
+The easiest setup using [Vagrant](https://www.vagrantup.com) and [VirtualBox](https://www.virtualbox.org) is shown here. 
+Some CPUs like Apples M1/M2 are not supported by VirtualBox, please jump down to the docker setup section. For manual installation instructions and production deployment please see the [wiki page on installation](https://github.com/e-valuation/EvaP/wiki/Installation).
```
He3lixxx, line 18: > please jump down to the docker setup section

I think we should just inline the docker-specific parts into this guide instead of having a separate docker section. In the end, it should just be "additionally, run `vagrant provision`". But let's focus on the other parts first.

README.md
```diff
@@ -44,14 +45,22 @@ vagrant ssh
...
 vagrant provision
 vagrant ssh
 ```
 
+bash in container:
+```bash
+export PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
+export PUPPETEER_SKIP_DOWNLOAD=true
+npm install typescript
+```
```
He3lixxx, line 62: I'd expect the downloading of chromium to be a postinstall action of the `puppeteer` npm package. Does `npm` really roll back the whole installation if a postinstall script fails? If it doesn't, an error here should affect the rest of the system.

In the best case, I'd want puppeteer to simply ignore errors during the chromium download here. If we can get that working, we can just include it in `provision_vagrant_vm.sh`. Users would not have to do anything additional then.

In docker, the headless browser is a pain anyways, at least on linux it doesn't start because sandboxing fails. So, I'd also be fine if we simply skipped chromium installation when docker is used. However, for virtualbox, we still want to provide the browser.

Btw, why wasn't this part of #1726? Was it still working back then?

He3lixxx (changes requested):

README.md
fidoriel: I do not know why it did not work one year ago, maybe I did something wrong with docker engine. I will remove it. Until someone complains, I cannot test this on a different mac without braking my setup.

If you have conainer a with host port 8000 and b with 8000, docker will allow that they exist simultaneously, with only running one at the same time. Vagrant does not like that, I thought it would be good to note that down because we have a rather unconventional docker setup.

README.md
fidoriel: If we inline them, we move it from experimental to supported, so we have to support it.

README.md
fidoriel: `apt-get install chromium-browser` is avaliable for arm64. But if we do install it on vagrant/docker, we do install a full chromium on every vm for no reason, takes long uses hdd space and it is useless, because as you said, it does not work the way we like. May we just skip the download on docker?

README.md
He3lixxx: I'd be fine with that. Whether we say that we support it doesn't really make a difference. People want to use it, if its broken they (rightfully) complain, so I'd just try to keep the readme simple.

README.md
He3lixxx: The actual containers running in the end shouldn't be special in any way. I wouldn't expect that vagrant has anything to do with ports other than forwarding this information to docker.

With vagrant, the container will run `tail -f /dev/null` indefinitely, so its running (and occupying) the port at any point between `vagrant up` and `vagrant halt` -- if you do not run `vagrant halt`, it will remain occupied. Is it possible that this was the cause of your problem?

README.md
fidoriel: yes, but we can workaround that by replacing the privision command for docker with `trigger.ruby do |env, machine| system("""vagrant docker-exec -it -- bash -c 'export PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true ; export PUPPETEER_SKIP_DOWNLOAD=true ; /evap/deployment/provision_vagrant_vm.sh'""") end`. Maby it is not the cleanest way, but it works and it is explicit enough to be found

He3lixxx: You'll still have trouble on any future `npm` operations in the container though, as the variables will not be set anymore, right?

Personally, I'd also be fine with this, but again, I think our long-term goal should be to get the frontend tests running in docker, not to further prevent that. @niklasmohrin thoughts?

(If we do this: Please only set one of the two variables, and personally I'd prefer the non-export way of specifying environment variables for a single command: `SOME_VARIABLE=true /evap/deployment/provision_vagrant_vm.sh`)

fidoriel: yes, this is not a long term solution and might create future trouble.

README.md
fidoriel: Create a random docker container using a port of your choice via EXPORT or docker compose. Shut down the container but do not remove it. Use a secondary evap install. Try to use vagrant with docker like the tutorial says on the same port, which now should be usable. Vagrant does not like that.

He3lixxx: I've opened https://github.com/hashicorp/vagrant/issues/13110, let's see what they say. In any case, I think the error is clear enough that (power) users should understand what's going on.

He3lixxx (approved):

niklasmohrin (approved):

PR #1918, opened 2023-04-04
Update to django 4.2

He3lixxx:
Technically, our dev-dependency `django-stubs` doesn't support django4.2 yet. However, I don't see a problem here if CI passes, so I wouldn't wait for them.

The other problematic candidate, `mozilla-django-oidc`, doesn't really give any kind of "we support these versions of django" information. Their tests don't yet run with django4.2. Locally, for me, OIDC login works as expected, and django's changelog doesn't indicate any relevant changes to the authentication mechanism.

---
Running the tests with warnings enabled currently triggers these:
* [x] > /home/evap/venv/lib/python3.8/site-packages/django/contrib/auth/views.py:142: RemovedInDjango50Warning: Log out via GET requests is deprecated and will be removed in Django 5.0. Use POST requests for logging out.

    -> We currently have three tests that `self.app.get(reverse("django-auth-logout")`. Simply changing these to "post" seems to fail. The logout button in the navbar currently simply is a link, not a form that will be submitted. I guess both require a bit of work.

* [x] > /home/evap/venv/lib/python3.8/site-packages/django/db/models/fields/related_lookups.py:52: RemovedInDjango50Warning: Passing unsaved model instances to related filters is deprecated.

    -> A few of our tests currently do this (`test_copy_contribution`, `test_initial_from_original`, `test_no_original_given`, `test_merge_users_changes_data_on_success`. Will have to reevaluate those. It seems like this also happens on some staff views (evaluation_copy, evaluation_create, import-views)

### DIFF
--- a/evap/contributor/templates/contributor_evaluation_form.html
+++ b/evap/contributor/templates/contributor_evaluation_form.html
@@ -108,7 +108,7 @@
                         <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
                     </div>
                     <div class="modal-body">
-                        {{ preview_html|safe }}
+                        {{ preview_html }}
                     </div>
                     <div class="modal-footer">
                         <button type="button" class="btn btn-light" data-bs-dismiss="modal">{% trans 'Close' %}</button>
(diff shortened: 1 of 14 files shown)


He3lixxx:
@niklasmohrin: Filtering anything like: `Contribution.objects.filter(evaluation=some_evaluation)` always behaved identical to `Contribution.objects.filter(evaluation=None)` if `some_evaluation` does not have a `pk` (=> is not stored in the DB). In the places I changed in bcdba1ee, we triggered this behavior.
A) Do we want to have narrower contracts there (=> don't allow input instances that do not have PKs)?
B) Do we want to take any precautions to find more occurrences of this warning? I could see how we trigger it more often in production -- we could try enabling warnings in production and see if this is triggers.

PR #1934, opened 2023-04-24
Dependabot: Don't create PRs that just narrow supported python versions

He3lixxx:
https://github.com/dependabot/dependabot-core/issues/5537 indicates this should work.


niklasmohrin (approved):
Discussed in person that this seems to be the way to go :)

PR #1948, opened 2023-05-28
Ubuntu 22.04 and Python 3.10

janno42:
upgrade vm to ubuntu 22.04 and to python 3.10
require vagrant 2.3.6

a special delivery for @niklasmohrin :)


He3lixxx (approved):
noice 👍

.github/workflows/tests.yml
```diff
@@ -62,10 +62,10 @@ jobs:
         run: python manage.py test --shuffle
 
   mypy:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     container:
-      image: python:3.8
+      image: python:3.10
```
niklasmohrin, line 68: Any reason why we still have the `container` here? We use `setup_python` below and I don't see why mypy would be a reason to use the container now. Maybe removing these lines will fix CI?

niklasmohrin (approved):
awesome

niklasmohrin:
Oh, one thing: Why vagrant 2.3.6? It only came out so recently and is not shipped with most package managers yet, for example, Ubuntu 2204 still has 2.2.19 and Arch has 2.3.6 still in testing. Is there any feature that we need?

niklasmohrin (approved):

PR #1954, opened 2023-06-05
Enable javascript import replacements for ManifestStaticFilesStorage

### LINKED ISSUE #1808: ManifestStaticFilesStorage cache-busting doesn't work for JavaScript imports
We currently use the `ManifestStaticFilesStorage` to have long browser cache times and cache-busting after updates. On `collectstatic`, it creates a copy of all files that also contains the file hash in the name -> `/static/css/evap.css` becomes `/static/css/evap.bfc8f42bfc39.css"`. In django templates, we use the `static` template tag to mark places where this replacement is necessary.
In CSS files, relative file names are automatically changed.

However, this doesn't work for javascript files that import each other. For example, our current `contact_modal.js` imports stuff from `utils.js`. This import statement is missing the hash in the file name.
The access will work, because the files are also available under their original name, but browsers might use the old, cached files.

Currently, production serves files with a one year cache expiring time (3 hours for the raw files without hashes in the name).
So, after an update, users might have broken javascript until that time runs out or they refresh with Ctrl+F5.

Options:
* Reduce the cache time, make the browser re-check more often (~10minutes?). Ensure that a re-check is a simple "nothing has changed" from the webserver (and is fast). Our webserver already sends out ETags with the requests. This would allow us to drop the ManifestStaticFilesStorage (or reduce it to only css and image files). I prefer this option.
* We considered whether we can have a "dear browser, the website was last updated on X, please discard all earlier cached files" header, so that browsers recheck files where they have older versions, but it doesn't seem like there is such a header available in HTTP.

* We could also manually change import-statements in js/ts files, maybe by changing the ManifestStaticFilesStorage, but it would probably be painful and error-prone.
  * We think we can not mark these places, like with the `static` template tag, because it would break javascript tooling.
* JS tooling could probably solve this (webpack, ...). This probably exists somewhere on the internet, collectstatic could then run this. However, in my experience, javascript packaging always is a pain, so I'd vote to not do that.

@niklasmohrin @karyon @janno42 opinions?

He3lixxx:
Fixes #1808.

For local testing, I needed to
1. disable `DEBUG`
2. run `collectstatic`
3. access the page through apache2 at `localhost:8001`

### DIFF
--- a/evap/settings.py
+++ b/evap/settings.py
@@ -13,6 +13,8 @@ import os
 import sys
 from typing import Any, List, Tuple
 
+from django.contrib.staticfiles.storage import ManifestStaticFilesStorage
+
 BASE_DIR = os.path.dirname(os.path.realpath(__file__))
 
 
@@ -133,9 +135,10 @@ CACHES = {
     },
 }
 
-from django.contrib.staticfiles.storage import ManifestStaticFilesStorage
+
 class ManifestStaticFilesStorageWithJsReplacement(ManifestStaticFilesStorage):
     support_js_module_import_aggregation = True
+
 
 STORAGES = {
     "default": {
--- a/evap/settings.py
+++ b/evap/settings.py
@@ -133,12 +133,16 @@ CACHES = {
     },
 }
 
+from django.contrib.staticfiles.storage import ManifestStaticFilesStorage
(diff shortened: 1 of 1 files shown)


niklasmohrin (approved):
Seems like it works, can't see any remaining downloads of js or css files without hashes

Issue #1800, opened 2022-08-24
Replace asserts with contraints


janno42:
A couple of asserts/checks might be better expressed as constraints. I'm not sure there was a reason why we didn't do that.
* [start date must be before end date](https://github.com/e-valuation/EvaP/blob/main/evap/evaluation/models.py#L467)
* [text and heading questions do not allow additional textanswers](https://github.com/e-valuation/EvaP/blob/main/evap/evaluation/models.py#L1121)
* [TextAnswer's answer and original_answer must not be the same](https://github.com/e-valuation/EvaP/blob/main/evap/evaluation/models.py#L1461)
* [Evaluation's _voter_count must be set iff _participant_count is set](https://github.com/e-valuation/EvaP/blob/main/evap/evaluation/models.py#L620)
* [A bunch of validation when redeeming points](https://github.com/e-valuation/EvaP/blob/fc3f5b4390fbaf2aff2633dfbe61bbdd8e80c8b7/evap/rewards/tools.py#L36-L46)

_Originally posted by @karyon in https://github.com/e-valuation/EvaP/issues/1796#issuecomment-1221534890_

florian-str:
Constrains as checklist:

- [ ] [start date must be before end date](https://github.com/e-valuation/EvaP/blob/main/evap/evaluation/models.py#L467)
- [ ] [text and heading questions do not allow additional textanswers](https://github.com/e-valuation/EvaP/blob/main/evap/evaluation/models.py#L1121)
- [ ] [TextAnswer's answer and original_answer must not be the same](https://github.com/e-valuation/EvaP/blob/main/evap/evaluation/models.py#L1461)
- [ ] [Evaluation's _voter_count must be set iff _participant_count is set](https://github.com/e-valuation/EvaP/blob/main/evap/evaluation/models.py#L620)
- [ ] [A bunch of validation when redeeming points](https://github.com/e-valuation/EvaP/blob/fc3f5b4390fbaf2aff2633dfbe61bbdd8e80c8b7/evap/rewards/tools.py#L36-L46)

niklasmohrin:
Closed by #1957

Issue #1846, opened 2023-01-08
Django's `bulk_` database operations circumvent `.save()` side effects


He3lixxx:
#1839 was caused by importers using `bulk_update` and `bulk_create` to create large amounts of instances. This circumvents all logic in the model's `save` method and doesn't send `pre_save`/`post_save` signals. In this case, we created evaluations, but general contributions for those weren't created, causing issues with other code assuming the general contributions exist.

This leads to the main question: How do we want to handle this general problem? Currently, we allow both using `bulk_create` and overriding `save()` to have side effects. However, I'd argue this is not sustainable: We have to duplicate all custom "on-database-store" logic with each `bulk_` operation, and maintain this duplication.

Options I see:
1. We disallow side effects in `.save()`. This makes overriding the method useless. The string `def save` currently occurs 5 times in a `models.py` file. We might get away with this if we manage to model everything as database constraint instead. But there is probably stuff that will be hard to model as a database constraint, like cleaning email addresses.
2. We disallow bulk operations. `bulk_create` currently occurs ~50 times (mostly though model-baker's `_bulk_create` option), `bulk_update` is used 4 times. This would negatively affect performance. For example, the importers are 70% = 5s slower on an import run when saving a single instance at a time.
3. We build infrastructure that allows having both for the 5 models where we currently have custom `save` overrides. I could imagine the following:
    * The model get's a `prepare_save_many(instances: Iterable[Self])` method. This does all the side-effect logic that was in `save` before. An empty implementation could be in a mixin / base class.
    * `save` is overridden and calls `prepare_save_many([self])`, then delegates to super. This could be provided by a mixin.
    * Either we manually call `prepare_save_many` before doing bulk operations, or we add custom managers for those 5 classes that ensure the call happens.
    * To me, it feels like this is possible to do with a reasonable amount of code, but I haven't looked at the overriden `save` methods yet.

Overall, I feel like this is a design flaw in django, and we can't possibly be the first project to hit this, maybe there's smart solutions somewhere on the internet, I haven't checked yet.

@karyon Have you thought about this in the past, and if yes, what were your conclusions?

karyon:
Using DB constraints instead of the asserts in save is handled here: https://github.com/e-valuation/EvaP/issues/1800

[...]

PR #1957, opened 2023-06-12
Use database constraints instead of python asserts

florian-str:
This implements all reasonable constraints as requested in #1800.


evap/evaluation/models.py
```diff
@@ -1452,6 +1471,12 @@ class Meta:
         ordering = ["id"]
         verbose_name = _("text answer")
         verbose_name_plural = _("text answers")
+        constraints = [
+            CheckConstraint(
+                check = ~Q(answer=F('original_answer')),
```
niklasmohrin, line 1476: This could also be written as
```suggestion
                check = Q(answer__ne=F('original_answer')),
```
How do we feel about both versions? @He3lixxx ?

evap/rewards/models.py
```diff
@@ -51,7 +51,7 @@ class RewardPointGranting(models.Model):
     user_profile = models.ForeignKey(UserProfile, models.CASCADE, related_name="reward_point_grantings")
     semester = models.ForeignKey(Semester, models.PROTECT, related_name="reward_point_grantings")
     granting_time = models.DateTimeField(verbose_name=_("granting time"), auto_now_add=True)
-    value = models.IntegerField(verbose_name=_("value"), default=0)
+    value = models.IntegerField(verbose_name=_("value"), default=0) # Might consider enforcing MinValue in the future
```
niklasmohrin, line 54: What is the status on this? Before merging, the comment should probably be more concrete as to why we are not enforcing it, or implement the constraint :)

niklasmohrin (commented):
Make sure to run `manage.py format` :) Secondly, I think the constraints could all use some nicer names, I made a suggestion for one of them below, we can also come up with the others together if you want.

evap/rewards/models.py
florian-str: It used to be the default value, but is apparently not wanted. Therefore, adding this constraint could break the database.

evap/evaluation/models.py
florian-str: It seems to be impossible on text fields.

evap/evaluation/migrations/0136_evaluation_check_evaluation_start_before_end_and_more.py
```diff
@@ -0,0 +1,55 @@
...
+            ),
+        ),
+        migrations.AddConstraint(
+            model_name="textanswer",
+            constraint=models.CheckConstraint(
+                check=models.Q(("answer", models.F("original_answer")), _negated=True),
+                name="check_evaluation_text_answer_is_modified",
+            ),
+        ),
+    ]
```
He3lixxx, line 55: There is 4 constraints here and one in the other migration, but I think this PR only removes 2 of the corresponding `assert` statements. Is this on purpose?

evap/rewards/models.py
He3lixxx: @janno42 Do we have `RewardPointGrantings` with `value==0` on production? I'd expect no, but just to be sure. If we don't have any, we can add the constraint here.

He3lixxx: We don't have any instances with `value = 0` in the production database, so we can add this constraint here. I'd vote in favor of adding it :)

evap/evaluation/models.py
```diff
@@ -1130,9 +1141,18 @@ class Meta:
         ordering = ["order"]
         verbose_name = _("question")
         verbose_name_plural = _("questions")
+        constraints = [
+            CheckConstraint(
+                check=~(Q(type=QuestionType.TEXT) | Q(type=QuestionType.HEADING))
+                | ~Q(
+                    allows_additional_textanswers=True
+                ),  # Value("type") == , #not in [QuestionType.TEXT, QuestionType.HEADING] or not Value("allows_additional_textanswers"),
```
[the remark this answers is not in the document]
florian-str: It is just leftover and has no purpose anymore.

evap/evaluation/migrations/0136_evaluation_check_evaluation_start_before_end_and_more.py
florian-str: It is correct. There are a couple asserts which can't be fully replaced. However, these can and according to the original issue should be enforced using a constraint.

evap/rewards/models.py
```diff
@@ -51,7 +52,7 @@ class RewardPointGranting(models.Model):
     user_profile = models.ForeignKey(UserProfile, models.CASCADE, related_name="reward_point_grantings")
     semester = models.ForeignKey(Semester, models.PROTECT, related_name="reward_point_grantings")
     granting_time = models.DateTimeField(verbose_name=_("granting time"), auto_now_add=True)
-    value = models.IntegerField(verbose_name=_("value"), default=0)
+    value = models.IntegerField(verbose_name=_("value"), validators=[MinValueValidator(1)])
```
He3lixxx, line 55: I'm fine with these for now, but we should keep in mind that this is not actually a database constraint, but something enforced by django on `save()`, and thus adds to the problems #1846.

He3lixxx (approved):
Code looks good. As discussed, please check whether the error message on an evaluation edit/create pages in the error case "end date is before start datetime" is unchanged.

florian-str:
It does in fact still produce the correct error message.

evap/evaluation/migrations/0136_use_more_database_constraints.py
niklasmohrin, line 1: The migration number is out of date again :D

niklasmohrin (approved):
Very nice!

niklasmohrin:
I'll update the migration file and merge it real quick :)

PR #1964, opened 2023-06-19
Refactor to builtin generic views where appropriate

### LINKED ISSUE #1754: Use django's generic views to reduce boilerplate code
Django provides a few generic views. At first glance, it looks like using them could reduce boilerplate code. The [`LoginView`](https://docs.djangoproject.com/en/4.0/topics/auth/default/#django.contrib.auth.views.LoginView), [`FormView`](https://docs.djangoproject.com/en/4.0/ref/class-based-views/generic-editing/#formview), [`DeleteView`](https://docs.djangoproject.com/en/4.0/ref/class-based-views/generic-editing/#deleteview), [`CreateView`](https://docs.djangoproject.com/en/4.0/ref/class-based-views/generic-editing/#createview) and [`UpdateView`](https://docs.djangoproject.com/en/4.0/ref/class-based-views/generic-editing/#updateview) seem to be the most relevant for us.

The login view implements some kind of redirect logic which might allow reducing or removing our custom redirect logic, but I don't know whether it would play well with multiple login methods (key, pw, oidc). The other views should remove overhead for delete/create/update views, iirc some of our views are boilerplate-only, so they should be applicable.

niklasmohrin:
Closes #1754


niklasmohrin:
Linting is currently failing for some complicated views like `UploadGradesView`, I think I will just revert this one instead of trying to make the class based one work

evap/staff/tests/test_views.py
```diff
@@ -1914,20 +1914,20 @@ def test_edit_course(self):
         self.course = Course.objects.get(pk=self.course.pk)
         self.assertEqual(self.course.name_en, "A different name")
 
-    @patch("evap.staff.views.redirect")
-    def test_operation_redirects(self, mock_redirect):
-        mock_redirect.side_effect = lambda *_args: HttpResponse()
+    @patch("evap.staff.views.reverse")
+    def test_operation_redirects(self, mock_reverse):
```
He3lixxx, line 1918: This now no longer tests that the correct status code is actually returned to the browser. Did you change this because the old mock didn't work anymore? Is it possible that if we simply change [where we patch](https://docs.python.org/3/library/unittest.mock.html#where-to-patch) that we could get stuff to work agian? Otherwise, could we maybe mock the constructor of `HttpResponse` too or something?

Idk, may also not be worth it.

Edit: I see that we simply specify where to redirect to in `def get_success_url` but don't actually do anything about the redirect ourselves. I we argue that we don't want to test django logic, I'd say we should remove the patching completely and simply call `get_success_url` here in the test...

evap/staff/views.py
```diff
@@ -679,27 +682,37 @@ def semester_import(request, semester_id):
...
+class SemesterExportView(SingleObjectMixin, FormsetView):
+    model = Semester
+    pk_url_kwarg = "semester_id"
+    form_class = formset_factory(form=ExportSheetForm, can_delete=True, extra=0, min_num=1, validate_min=True)
+    template_name = "staff_semester_export.html"
 
-    ExportSheetFormset = formset_factory(form=ExportSheetForm, can_delete=True, extra=0, min_num=1, validate_min=True)
-    formset = ExportSheetFormset(request.POST or None, form_kwargs={"semester": semester})
+    def dispatch(self, *args, **kwargs):
+        self.object = self.semester = self.get_object()
```
He3lixxx, line 692: We could just always refer to `self.object`, I think I'd be fine with that? Might backfire though. But having two different names might also backfire...

Edit: We use `object` in the `CourseEditView` below, I think having it explicitly declared here as `object: Semester` should be fine.

evap/staff/views.py
```diff
@@ -1002,41 +1015,46 @@ def course_copy(request, course_id):
...
+    def get_success_url(self):
+        match self.request.POST.get("operation"):
+            case "save":
+                return reverse("staff:semester_view", args=[self.object.semester.id])
+            case "save_create_evaluation":
+                return reverse("staff:evaluation_create_for_course", args=[self.object.id])
+            case "save_create_single_result":
+                return reverse("staff:single_result_create_for_course", args=[self.object.id])
+            case _:
+                raise SuspiciousOperation("Invalid POST operation")
```
He3lixxx, line 1057: This would still have changed the course, and then return the 400 after that, right? Shouldn't be a bit deal, but might be a bit weird.

evap/staff/views.py
```diff
@@ -1941,37 +1959,33 @@ def questionnaire_set_locked(request):
...
 
 @manager_required
-def degree_index(request):
-    degrees = Degree.objects.all()
-
-    DegreeFormset = modelformset_factory(
-        Degree, form=DegreeForm, formset=ModelWithImportNamesFormset, can_delete=True, extra=1
+class DegreeIndexView(SuccessMessageMixin, SaveValidFormMixin, FormsetView):
+    model = Degree
+    form_class = modelformset_factory(
```
He3lixxx, line 1964: This is used by `FormsetView`, which is our code, right? If so, do we want to name this `formset_class` instead?

evap/staff/views.py
```diff
@@ -2223,15 +2234,16 @@ def user_bulk_update(request):
...
+    template_name = "staff_user_merge_selection.html"
 
-    if form.is_valid():
-        main_user = form.cleaned_data["main_user"]
-        other_user = form.cleaned_data["other_user"]
-        return redirect("staff:user_merge", main_user.id, other_user.id)
-
-    return render(request, "staff_user_merge_selection.html", {"form": form})
+    def get_success_url(self):
+        return redirect(
```
He3lixxx, line 2242: codecov says untested -- I'd make an issue for these? I'll just annotate "codecov says untested" below to keep track of this.

evap/staff/views.py
```diff
@@ -2262,61 +2274,59 @@ def user_merge(request, main_user_id, other_user_id):
 
 
 @manager_required
-def template_edit(request, template_id):
-    template = get_object_or_404(EmailTemplate, id=template_id)
-    form = EmailTemplateForm(request.POST or None, request.FILES or None, instance=template)
+class TemplateEditView(SuccessMessageMixin, UpdateView):
+    model = EmailTemplate
+    pk_url_kwarg = "template_id"
```
He3lixxx, line 2279: codecov says untested

evap/staff/views.py
```diff
@@ -2262,61 +2274,59 @@ def user_merge(request, main_user_id, other_user_id):
...
+        return context
 
-    if formset.is_valid():
-        formset.save()
-        messages.success(request, _("Successfully updated the FAQ sections."))
-        return redirect("staff:faq_index")
 
-    return render(request, "staff_faq_index.html", {"formset": formset, "sections": sections})
+@manager_required
+class FaqIndexView(SuccessMessageMixin, SaveValidFormMixin, FormsetView):
```
He3lixxx, line 2321: codecov says untested on the class definition line? This got to be wrong, right?

He3lixxx (commented):
Good stuff

He3lixxx:
> Linting is currently failing for some complicated views like `UploadGradesView`, I think I will just revert this one instead of trying to make the class based one work

Overall, I'd be fine with having grade upload remain a function-based view. From the current CI linting output:

> evap/staff/views.py:692:8: W0201: Attribute 'object' defined outside __init__ (attribute-defined-outside-init)

I'd be fine with adding exceptions for these, i.e. telling pylint that `dispatch` (and, if necessary, `form_valid`) may define new attributes.

> evap/grades/views.py:116:4: W0221: Number of parameters was 4 in 'View.dispatch' and is now 3 in overriding 'UploadGradesView.dispatch' method (arguments-differ)

Why is that? What additional two arguments does `View.dispatch` offer? Couldn't we just also take `*args` and `**kwargs` and simply pass them on?
(similar thing for `get_object`)

evap/staff/views.py
niklasmohrin: Yes, we could also check it in `form_valid` before calling `super().form_valid` where it would be saved

evap/evaluation/auth.py
```diff
@@ -45,160 +48,110 @@ def authenticate(self, request, email=None, password=None):  # pylint: disable=a
...
     """
     Decorator for views that checks that the user is logged in and has edit
     right for at least one evaluation.
     """
-
-    def check_user(user):
-        return user.is_editor
-
-    return user_passes_test(check_user)(view_func)
+    return user.is_editor
```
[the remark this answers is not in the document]
niklasmohrin: yup, unused

evap/grades/views.py
```diff
@@ -43,42 +47,51 @@ def course_grade_document_count_tuples(courses: QuerySet[Course]) -> List[Tuple[
...
+    def get_object(self, *args, **kwargs):
+        course = super().get_object(*args, **kwargs)
+        if course.semester.grade_documents_are_deleted:
+            raise PermissionDenied
+        return course
+
+    def get_context_data(self, **kwargs):
+        return super().get_context_data(**kwargs) | {
+            "semester": self.object.semester,
+            "grade_documents": self.object.grade_documents.all(),
```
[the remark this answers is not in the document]
niklasmohrin: Yes, it gets added as "course" by the `DetailView`

evap/grades/views.py
```diff
@@ -43,42 +47,51 @@ def course_grade_document_count_tuples(courses: QuerySet[Course]) -> List[Tuple[
...
+            "disable_if_archived": "disabled",
```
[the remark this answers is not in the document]
niklasmohrin: ah yes, I had it the wrong way around

evap/staff/views.py
niklasmohrin: Actually, this is used by `FormView` which is Django, but we could add this renaming functionality for `FormSetView` too. I was thinking about overriding `form_valid` and similar too so that subclasses can use `formset` as the argument name, but I wanted to wait until we decide which views we want to keep

evap/staff/tests/test_views.py
niklasmohrin: Hm, I feel that using `get_success_url` would be too specific to the implementation, best case I would like to get the response and look at it, but by then the urls are resolved already, so I mocked `reverse`. So to include your suggestion, I would say we could either a) keep the current check and also assert that the response redirects (anywhere) or b) only look at the response and compare with the raw url (which we know) or compare with the result of a `reverse(...)`

niklasmohrin: What do you think about the current version?

niklasmohrin:
Lints are all fixed, reverted some views to functions, `FormSetView` now has more niceties

evap/evaluation/auth.py
```diff
@@ -45,160 +48,101 @@ def authenticate(self, request, email=None, password=None):  # pylint: disable=a
...
-    """
-    Decorator for views that checks whether a user passes a given test
-    (raising 403 if not). The test should be a callable that takes the
-    user object and returns True if the user passes.
-    """
-
-    def decorator(view_func):
-        @wraps(view_func)
-        def _wrapped_view(request, *args, **kwargs):
+def class_or_function_check_decorator(test_func: Callable[[UserProfile], bool]):
```
He3lixxx, line 51: Given that this is more complex now: I'd vote to add 4 unit tests here (function passed, function failed, class passed, class failed) if we don't already have something similar.

He3lixxx (commented):
Looks good at a quick glance, but I'll have to go through it carefully once more. You can rebase / prepare a merge if you want.

evap/evaluation/auth.py
```diff
@@ -45,160 +48,101 @@ def authenticate(self, request, email=None, password=None):  # pylint: disable=a
...
+                raise PermissionDenied
+            return func(request, *args, **kwargs)
 
-        return _wrapped_view
+        return wrapped
 
-    return decorator
+    def decorator(class_or_function):
+        if inspect.isclass(class_or_function):
+            return method_decorator(function_decorator, name="dispatch")(class_or_function)
```
richardebeling, line 63: Maybe add a comment pointing to https://docs.djangoproject.com/en/4.2/topics/class-based-views/intro/#decorating-the-class, and possibly explain what we're doing here, something like:
> Using django's `method_decorator` and hardcoding `name="dispatch"` so it can be applied to view classes. See https://docs.djangoproject.com/en/4.2/topics/class-based-views/intro/#decorating-the-class

evap/grades/views.py
```diff
@@ -41,42 +44,51 @@ def course_grade_document_count_tuples(courses: QuerySet[Course]) -> list[tuple[
...
+        course = super().get_object(*args, **kwargs)
+        if course.semester.grade_documents_are_deleted:
+            raise PermissionDenied
+        return course
+
+    def get_context_data(self, **kwargs):
+        return super().get_context_data(**kwargs) | {
+            "semester": self.object.semester,
+            "grade_documents": self.object.grade_documents.all(),
+            "disable_if_archived": "",
```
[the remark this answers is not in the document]
richardebeling: Passing it as explicitly empty seems more intention-conveying to me than just leaving it out, so I'd prefer that.

I'd be fine with a comment.

I'd even say the check above is business logic, not technical necessity, so it could change, so we might even want to bring back the old expression here (`disabled if blah else blah`)

evap/staff/views.py
```diff
@@ -681,27 +684,39 @@ def semester_import(request, semester_id):
...
-    semester = get_object_or_404(Semester, id=semester_id)
+class SemesterExportView(SingleObjectMixin, FormsetView):
+    model = Semester
+    pk_url_kwarg = "semester_id"
+    formset_class = formset_factory(form=ExportSheetForm, can_delete=True, extra=0, min_num=1, validate_min=True)
+    template_name = "staff_semester_export.html"
 
-    ExportSheetFormset = formset_factory(form=ExportSheetForm, can_delete=True, extra=0, min_num=1, validate_min=True)
-    formset = ExportSheetFormset(request.POST or None, form_kwargs={"semester": semester})
+    object: Semester
```
richardebeling, line 693: We talked about this and got confused. Maybe it's better to keep the old view here?

Kakadus (commented):
I think the forms are definitely more readable now

richardebeling (approved):

PR #1967, opened 2023-06-19
Redesign profile page

janno42:
- add help texts to form fields
- use only a single form on the user profile page
- redesign the user profile page


evap/evaluation/models.py
```diff
@@ -1621,7 +1621,13 @@ class UserProfile(AbstractBaseUser, PermissionsMixin):
 
     title = models.CharField(max_length=255, blank=True, default="", verbose_name=_("Title"))
     first_name_given = models.CharField(max_length=255, blank=True, verbose_name=_("given first name"))
-    first_name_chosen = models.CharField(max_length=255, blank=True, default="", verbose_name=_("display name"))
+    first_name_chosen = models.CharField(
+        max_length=255,
+        blank=True,
+        default="",
+        verbose_name=_("display name"),
+        help_text=_("This will replace your given first name."),
```
niklasmohrin, line 1629: "given first name" is a bit odd, maybe it should be "given name" instead?

evap/evaluation/templates/profile.html
```diff
@@ -6,17 +6,44 @@
...
+                        {% endfor %}
+                    </div>
+                </div>
+            </div>
+            {% if user.is_editor %}
+                <div class="col">
+                    <div class="card h-100">
+                        <div class="card-body">
+                            <h4 class="card-title mb-4">{% trans 'Delegates and emails' %}</h4>
+                            <div class="form-text">{% trans 'Your delegates will receive emails sent to you in CC, will be able to prepare evaluations on your behalf, and will see the results of your evaluations.' %}</div>
```
niklasmohrin, line 29: I found it a bit odd that the help text for the display name is below the field, but this help text is below a heading / nothing. From a quick glance, maybe moving it below the select element would be more consistent? What do you think?

evap/evaluation/migrations/0136_alter_userprofile_first_name_chosen.py
```diff
@@ -0,0 +1,23 @@
...
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="userprofile",
+            name="first_name_chosen",
+            field=models.CharField(
+                blank=True,
+                default="",
+                help_text="This will replace your given first name.",
```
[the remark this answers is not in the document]
janno42: We'll keep it in the (quite unnecessary) own migration.

evap/evaluation/models.py
janno42: We decided that the form field should be named "given first name" to make clear that it is about the first name. "Given name" might be ambiguous.

niklasmohrin (approved):

PR #1968, opened 2023-06-26
Merge suggestions

### LINKED ISSUE #1780: User profile merge suggestions
On the user merge page, two user profiles can be selected and then be merged. The view should be extended by a second part (placed in a second card on the right, the two cards should be placed like those on the user profile index page) which lists user profiles that should probably be merged.

The suggestion list should show all user profile pairs where
- both users are internal and
- the users' email addresses first part (before the `@`) matches.

For each of the entries, a button should be shown which links to the user profile merge preview page of the suggested merge of the two user profiles. The main user profile to be kept is the one with the higher ID.

olewoy:
closes #1780

@janno42 any tipps/ideas to improve the frontend?

Backend in Python is planned to be replaced by using the database.


janno42:
I pushed a suggestion for the UI design into your branch. If you want a detailed explanation about my thoughts, let's talk at a meeting :)

evap/staff/views.py
```diff
@@ -2226,12 +2227,24 @@ def user_bulk_update(request):
 def user_merge_selection(request):
     form = UserMergeSelectionForm(request.POST or None)
 
+    all_users = list(UserProfile.objects.all())
+    suggested_merges = []
+    for user1 in all_users:
+        if user1 != None and user1.is_external == False:
+            for user2 in all_users:
```
He3lixxx, line 2234: As you already indicated, we'll have to do something a bit more efficient here.

@niklasmohrin Postgres has string functions that would allow us the email-to-username-portion conversion (e.g. `split_part`), we could then use database joining to annotate users with a merge candidate, and only return those users that actually have one. However, the database function is unique to postgres. We could also manually do a hash join in python, this would be portable across databases. Preferences?

evap/development/fixtures/test_data.json
```diff
@@ -130235,6 +130235,69 @@
...
+    "is_proxy_user": false,
+    "login_key": null,
+    "login_key_valid_until": null,
+    "is_active": true,
+    "groups": [],
+    "user_permissions": [],
+    "delegates": [],
+    "cc_users": []
+  }
+},
```
He3lixxx, line 130300: The idea here probably is to have some content on that page when loading it? I think for that, a single new user suffices. Could you undo all other changes to this file? (We can help you with that at a meeting)

evap/development/fixtures/test_data.json
```diff
@@ -130235,6 +130235,69 @@
...
+{
+  "model": "evaluation.userprofile",
+  "fields": {
+    "password": "",
+    "last_login": null,
+    "is_superuser": false,
+    "email": "echo.andre@institution.duplicate.com",
```
He3lixxx, line 130286: The usage of `example.com` is deliberate, as this domain is reserved for such purposes. We don't have the same guarantee for `duplicate.com`, so I'd try to stick to `example.com`. Maybe we can use `duplicate.example.com`?

evap/staff/views.py
```diff
@@ -2226,12 +2227,24 @@ def user_bulk_update(request):
 def user_merge_selection(request):
     form = UserMergeSelectionForm(request.POST or None)
 
+    all_users = list(UserProfile.objects.all())
+    suggested_merges = []
+    for user1 in all_users:
+        if user1 != None and user1.is_external == False:
```
He3lixxx, line 2233: We'll still need to add a test for the logic here, currently it's untested.

He3lixxx (commented):
Not sure if / how much you wanted to have feedback for this PR already. To dismiss my notification in good conscience, I quickly went through the code. Feel free to ignore stuff you already knew, fixed, or that is outdated.

evap/staff/views.py
He3lixxx: django code that finds a merge candidate in the database:
```python3
    from django.db.models import Func, Value, CharField, Subquery
    from evap.evaluation.models import UserProfile

    def email_to_local_part(field):
        return Func(field, Value('"@"'), Value(1), output_field=CharField(), function="split_part")

    query = UserProfile.objects.annotate(email_user_part=email_to_local_part("email"))
    query = query.annotate(
        merge_candidate_pk=Subquery(
            query.filter(pk__gt=OuterRef("pk"), email_user_part=email_to_local_part(OuterRef("email"))).values("pk")[:1]
        )
    )
    
    # -- untested from here --
    # not sure if we can make django directly annotate "into" a foreign key field so we don't have to manually do this
    users_with_merge_candidates = list(query.filter(merge_candidate_pk__isnull=False))
    merge_candidate_by_pk = {(user.pk, user) for user in UserProfile.objects.filter(
        pk__in=[inner_user.merge_candidate_pk for inner_user in users_with_merge_candidates]
    )}
    
    for user in users_with_merge_candidate:
        user.merge_candidate = merge_candidates_by_pk[user.merge_candidate_pk].
        
    # and then, it should be very easy, as you can just use
    print(user.merge_candidate)
    # to access a the merge candidate of each user.
```

richardebeling:
@olewoy Do you want to continue working on this or should we take over?

janno42:
Hey @olewoy, this is a very nice PR and it would be great if you would continue to work on it. Any chance you'll be able to get to it soon?

olewoy:
Yes, I will be there at 18.12. and finish it.

evap/staff/templates/staff_user_merge_selection.html
```diff
@@ -8,20 +8,50 @@
...
             </div>
         </div>
-        <div class="card card-submit-area text-center mb-3">
-            <div class="card-body">
-                <button type="submit" class="btn btn-primary">{% trans 'Show merge info' %}</button>
+        <div class="col">
+            <div class="card">
+                <div class="card-body">
+                    <h4 class="card-title">{% trans 'Merge suggestions' %}</h4>
+                    <table id="merge-suggestion-table" class="table table-striped table-narrow table-vertically-aligned">
```
richardebeling, line 31: The `id` seems to be unused here.

```suggestion
                    <table class="table table-striped table-narrow table-vertically-aligned">
```

evap/staff/templates/staff_user_merge_selection.html
```diff
@@ -8,20 +8,50 @@
...
+                        <tbody>
+                            {% for users in suggested_merges %}
```
richardebeling, line 33: In python, you could use tuple unpacking/destructuring and write something like
```python3
for main_user, merge_candidate in suggested_merges
```
with the advantage of being able to name these two. I think this also works here, but not too sure about it.

evap/staff/views.py
```diff
@@ -2294,6 +2306,35 @@ class UserMergeSelectionView(FormView):
...
+            # django docs support our usage here:
+            # https://docs.djangoproject.com/en/5.0/ref/models/expressions/#func-expressions
+            # pylint: disable=abstract-method
+            template = "split_part(%(expressions)s, '@', 1)"
+
+        query = UserProfile.objects.annotate(username_part_of_email=UserNameFromEmail("email"))
+
+        users_with_merge_candidates = query.annotate(
+            merge_candidate_pk=query.filter(username_part_of_email=UserNameFromEmail(OuterRef("email")))
+            .exclude(email__gte=OuterRef("email"))
```
richardebeling, line 2322: While thinking about the "don't match user with themselves" and "we only want each pair once", I didn't put enough thought into _which_ pair we want to have here. In the issue, the requirement given is that
> the main user profile to be kept is the one with the higher ID.

But we should be able to incorporate this here by simply using
```suggestion
            .exclude(pk__gte=OuterRef("pk"))
```

Or, maybe it's easier to think about the set here with `filter` and `lt`, not sure about that.

richardebeling (commented):
Looking mostly good from my side, thank you.

We might incorporate small final changes ourselves to get this merged before the next release. Would this be fine with you?

evap/staff/views.py
niklasmohrin: I usually find `filter` to be easier to think about

evap/staff/templates/staff_user_merge_selection.html
```diff
@@ -8,20 +8,50 @@
...
     {{ block.super }}
-    <h3>{% trans 'Merge users' %}</h3>
-
-    <form id="user-selection-form" method="POST" class="form-horizontal">
-        {% csrf_token %}
-        <div class="card mb-3">
-            <div class="card-body">
-                <p>{% trans 'Select the users you want to merge.' %}</p>
-                {% include 'bootstrap_form.html' with form=form wide=True %}
+    <div class="row row-cols-1 row-cols-md-2">
```
janno42, line 11:
```suggestion
    <div class="row row-cols-1 row-cols-xl-2">
```

janno42 (changes requested):

janno42 (approved):

richardebeling (approved):
Code is fine with me. We should probably add a test, but we can make a new issue for that.

Issue #1854, opened 2023-01-23
Use `LiveServerTestCase` for frontend tests


niklasmohrin:
With our current setup for frontend tests, we generate static html files that are then opened with selenium by a nodejs script. This comes with the disadvantage, that we cannot (in all cases) test the behavior when submitting forms, which is something we want to do more often. There is https://docs.djangoproject.com/en/4.1/topics/testing/tools/#liveservertestcase which could be used as a driver for the frontend tests. Even further, it turns out the linked documentation even provides an example of a selenium driven test, so we should investigate porting our frontend tests to Python.

PR #1973, opened 2023-07-03
Add keyboard shortcut for toggling flag state of textanswer

### LINKED ISSUE #1974: Hotkey for flagging Textanswers
When reviewing Textanswers in quick view, there are hotkeys for various operations. The newly implemented flagging (#1815) does not yet have a hotkey. The keys `F` and `D` should be added to set or remove a flag and should be added to the hotkey overview.

niklasmohrin:
Closes #1974


He3lixxx (approved):

niklasmohrin:
Any opinion on clicking on `input` vs `label`? Users click on the label, but clicking the input should be just as fine, unless we ever move the input under another clickable thing, but that would be insane. Not impossible though. Any more thoughts?

He3lixxx:
> Any opinion on clicking on `input` vs `label`? Users click on the label, but clicking the input should be just as fine, unless we ever move the input under another clickable thing, but that would be insane. Not impossible though. Any more thoughts?

I'm fine with both, I think both are equally likely to break. If we want to prevent that, we'd need to add a test.

niklasmohrin:
Pinging #1854

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

PR #1978, opened 2023-07-10
Importers: Disallow course merging for published evaluations and single results

### LINKED ISSUE #1976: Importer must not change data of published evaluations
When importing enrollment data, the importer checks if there are already courses/evaluations with the same name. If there is exactly one evaluation for the course name specified in the import file, all participants of this course who are not yet in the participant list of this existing evaluation will be added to the import file. This should not happen with already published evaluations, because then `participants` will be changed, although `_participant_count` has already been set at publication and is therefore stored in the cache. The numbers would then differ, leading to incorrectly displayed values and making operations impossible, since the equality of these numbers is asserted when unpublishing.
*Edit: We don't want any courses that have a state >= `IN_EVALUATION` to be modified.*

If a midterm evaluation was manually created, finished, and published, and this is the only existing evaluation for a course, it should still not be touched by the importer during enrollment import. Instead, an error message should be displayed stating that a published evaluation already exists for this course and that it cannot be imported.

He3lixxx:
Fixes #1976 (I hope)

@janno42 does simply checking for `_participant_count` cover all relevant cases?

We thought about preventing more changes to `participants` as soon as `_participant_count` is set, but
* deletions are an expected thing (if a `UserProfile` is deleted, the user is removed from `participants`, but kept in `_participant_count`. This is intentional to keep showing a correct count, even if users are deleted. Users are kept as `participants` so that we can still show what evaluations they participated in in the past).
* Only additions could be limited -- here, we could use a m2m signal handler on additions -- an addition is unexpected if the `_count` field is set, but we'd again bump into #1843 as soon as stuff uses bulk_inserts or similar operations.

### DIFF
--- a/evap/staff/importers/enrollment.py
+++ b/evap/staff/importers/enrollment.py
@@ -337,17 +337,26 @@ class CourseMergeLogic:
         merge_candidate_evaluations = merge_candidate.evaluations.all()
         if len(merge_candidate_evaluations) != 1:
             hindrances.append(_("the existing course does not have exactly one evaluation"))
-        else:
-            merge_candidate_evaluation: Evaluation = merge_candidate_evaluations[0]
-            if merge_candidate_evaluation.wait_for_grade_upload_before_publishing != course_data.is_graded:
-                hindrances.append(_("the evaluation of the existing course has a mismatching grading specification"))
+            return hindrances
 
-            if merge_candidate_evaluation._participant_count is not None:
-                hindrances.append(
-                    _(
-                        "the evaluation of the existing course has unmodifiable participants (is it published or a single result?)"
-                    )
+        merge_candidate_evaluation: Evaluation = merge_candidate_evaluations[0]
+
+        if merge_candidate_evaluation.wait_for_grade_upload_before_publishing != course_data.is_graded:
+            hindrances.append(_("the evaluation of the existing course has a mismatching grading specification"))
+
+        if merge_candidate_evaluation.is_single_result:
+            hindrances.append(_("the evaluation of the existing course is a single result"))
+            return hindrances
+
+        if merge_candidate_evaluation.state >= Evaluation.State.PUBLISHED:
+            hindrances.append(
+                _(
+                    "the import would add participants to the existing evaluation but the participants can't be modified "
+                    "because the evaluation is already published"
                 )
+            )
+        else:
+            assert merge_candidate_evaluation._participant_count is None
 
         return hindrances
 
--- a/evap/staff/importers/enrollment.py
+++ b/evap/staff/importers/enrollment.py
@@ -326,6 +326,7 @@ class CourseMergeLogic:
     @staticmethod
     def get_merge_hindrances(course_data: CourseData, merge_candidate: Course) -> list[str]:
         hindrances = []
+
         if merge_candidate.type != course_data.course_type:
             hindrances.append(_("the course type does not match"))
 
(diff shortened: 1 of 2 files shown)


evap/staff/importers/enrollment.py
```diff
@@ -336,8 +337,17 @@ def get_merge_hindrances(course_data: CourseData, merge_candidate: Course) -> Li
...
-            hindrances.append(_("the evaluation of the existing course has a mismatching grading specification"))
+        else:
+            merge_candidate_evaluation: Evaluation = merge_candidate_evaluations[0]
+            if merge_candidate_evaluation.wait_for_grade_upload_before_publishing != course_data.is_graded:
+                hindrances.append(_("the evaluation of the existing course has a mismatching grading specification"))
+
+            if merge_candidate_evaluation._participant_count is not None:
+                hindrances.append(
+                    _(
+                        "the evaluation of the existing course has unmodifiable participants (is it published or a single result?)"
```
niklasmohrin, line 348: I think something like "but this row would add a new participant" would improve the error message

janno42: maybe `the import would add participants to the existing evaluation but the participants can't be modified (because the evaluation is already published or is a single result)`

janno42 (commented):
checking for `_participant_count` should be fine but it seems a bit like working around the symptom rather than modeling it by use case. did you have any reasons not to check for the state?

He3lixxx:
> checking for `_participant_count` should be fine but it seems a bit like working around the symptom rather than modeling it by use case. did you have any reasons not to check for the state?

Hmm. To me, checking for the evaluation state felt more indirect / working-around-the-symptom-y. In my mind, the issue here is that the importer wants to change participants while participants semantically are immutable. I was thinking we may change the correlation between immutability of participants and state of an evaluation in the future, then a check for evaluation state here would become wrong, while this check would still be valid.

niklasmohrin:
I have been trying to come up with an opinion on this matter for some time now, but I think it boils down to what "modeling by use case" refers to. Maybe we can find a fair middle ground by adding `participations_are_mutable` to `Evaluation` which removes the knowledge about `_participant_count` here but also is specific enough for the _enrollment_-importer

evap/staff/tests/test_importers.py
```diff
@@ -724,6 +726,66 @@ def test_existing_course_different_attributes(self):
...
+        existing_evaluation = Evaluation.objects.get(pk=existing_evaluation.pk)
+        existing_evaluation._participant_count = existing_evaluation.participants.count()
+        existing_evaluation._voter_count = existing_evaluation.voters.count()
+        existing_evaluation.save()
+
+        with override_settings(DEBUG=False):
+            importer_log = import_enrollments(
+                self.default_excel_content, self.semester, self.vote_start_datetime, self.vote_end_date, test_run=False
+            )
+        self.assertCountEqual(
```
niklasmohrin, line 760: Could probably be just `assertEqual`

niklasmohrin (approved):

evap/staff/importers/enrollment.py
```diff
@@ -336,9 +337,27 @@ def get_merge_hindrances(course_data: CourseData, merge_candidate: Course) -> Li
...
+        merge_candidate_evaluation: Evaluation = merge_candidate_evaluations[0]
+
+        if merge_candidate_evaluation.wait_for_grade_upload_before_publishing != course_data.is_graded:
             hindrances.append(_("the evaluation of the existing course has a mismatching grading specification"))
 
+        if merge_candidate_evaluation.is_single_result:
+            hindrances.append(_("the evaluation of the existing course is a single result"))
+            return hindrances
+
+        if merge_candidate_evaluation.state >= Evaluation.State.PUBLISHED:
```
janno42, line 351:
```suggestion
        if merge_candidate_evaluation.state >= Evaluation.State.IN_EVALUATION:
```

we shouldn't allow changes on evaluations that are already running

evap/staff/importers/enrollment.py
```diff
@@ -336,9 +337,27 @@ def get_merge_hindrances(course_data: CourseData, merge_candidate: Course) -> Li
...
+            hindrances.append(
+                _(
+                    "the import would add participants to the existing evaluation but the participants can't be modified "
+                    "because the evaluation is already published"
```
janno42, line 355: `the import would add participants to the existing evaluation but the evaluation is already running`

janno42 (changes requested):

janno42 (commented):
In case you get bored (doesn't need to be changed imho): When the error about an existing unmodifiable evaluation is displayed, the warning about a course of the same name into which the participants will be merged is still displayed.

niklasmohrin (approved):

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

PR #1979, opened 2023-07-10
Fixed duplicate through email cleaning #1956

### LINKED ISSUE #1956: Missing form error for duplicate email address
When changing a user's email address to an address that does not yet exist, but results in an existing address after applying the replacements in `evaluation.tools.clean_email`, an internal server error is thrown.

`staff.forms.clean_email` should call `evaluation.tools.clean_email` to prevent this.

Test case:
- replace the `INSTITUTION_EMAIL_REPLACEMENTS` definition in the settings to `INSTITUTION_EMAIL_REPLACEMENTS: List[Tuple[str, str]] = [("institution.example.com", "institution.com")]`
- change Oma Abner's email address to `oma.abner@institution.com`
- change any other user's email address to `oma.abner@institution.example.com`

florian-str:
Emails now get transformed before being validated / checked for duplicates in the edit form. Therefor, you can avoid accidentally triggering a database constraint.

Fixes #1956

### DIFF
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -987,11 +987,11 @@ class UserForm(forms.ModelForm):
         return evaluations_participating_in
 
     def clean_email(self):
-        email = self.cleaned_data.get("email")
+        email = clean_email(self.cleaned_data.get("email"))
         if email is None:
             return None
 
-        user_with_same_email = UserProfile.objects.filter(email__iexact=clean_email(email))
+        user_with_same_email = UserProfile.objects.filter(email__iexact=email)
 
         # make sure we don't take the instance itself into account
         if self.instance and self.instance.pk:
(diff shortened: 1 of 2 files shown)


niklasmohrin (changes requested):
I don't think is exactly what we want, it feels a bit strange that we use our helper `clean_email` to compare to existing data, but still clean the email of this form here differently. I think it would be better if we would use `clean_email` directly on the email retrieved from `cleaned_data` and just have all the email cleaning in this one place - the call to `lower` later can then be removed too

He3lixxx (approved):
nice, thanks :)

niklasmohrin (approved):

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

PR #1990, opened 2023-07-31
Add configurable startpage

### LINKED ISSUE #1742: Configurable start page
Users who are both participant and contributor should be able to define which index page (student index or contributor index) is their start page (default: contributor index). This information should be saved within the users' profiles.

On top of the page which is currently not the start page of the user, a button should be added: "Make this my start page". Clicking this button changes the user's start page to this page.

janno42:
fix #1742; make start page configurable


evap/evaluation/templates/startpage_button.html
```diff
@@ -0,0 +1,25 @@
...
+    </div>
+{% else %}
+    <button id="makeStartPageButton" type="button" class="btn btn-sm btn-light" onclick="makeStartPage();" data-bs-toggle="tooltip" data-bs-placement="left" title="{% trans 'Make this my startpage' %}"><span id="span-set-startpage" class="fas fa-house"></span></button>
+{% endif %}
+
+
+{% block additional_javascript %}
+    <script type="text/javascript">
+        function makeStartPage() {
+            var makeStartPageButton = $('#makeStartPageButton');
```
niklasmohrin, line 13: In spirit of our current motto of preferring browser behavior over javascript and trimming down on jQuery in particular, I would suggest to replace the current logic with a form. The button becomes a submit with `name="page" value="{{ page }}"` and the view redirects to the users start page - which, if updating the startpage worked correctly, should be equivalent to reloading the current site

He3lixxx (approved):

evap/evaluation/templates/startpage_button.html
```diff
@@ -0,0 +1,13 @@
...
+    </div>
+{% else %}
+    <form class="d-flex" method="post" action="{% url 'evaluation:set_startpage' %}">
+        {% csrf_token %}
+        <input name="page" type="hidden" value="{{ page }}">
+        <button id="makeStartPageButton" type="submit" class="btn btn-sm btn-light" onclick="setSpinnerIcon('span-set-startpage')" data-bs-toggle="tooltip" data-bs-placement="left" title="{% trans 'Make this my startpage' %}">
+            <span id="span-set-startpage" class="fas fa-house"></span>
+        </button>
+    </form>
+{% endif %}
```
niklasmohrin, line 13: How do you feel about the following, to make sure that the buttons look the same in both cases (I don't like the duplicated html)?

```html
<form class="d-flex" method="post" action="{% url 'evaluation:set_startpage' %}" onsubmit="setSpinnerIcon('span-set-startpage')">
    {% csrf_token %}
    <input name="page" type="hidden" value="{{ page }}">
    <span
        data-bs-toggle="tooltip"
        data-bs-placement="left"
        {% if user.startpage == page %}
            title="{% trans 'This is your startpage' %}"
        {% else %}
            title="{% trans 'Make this my startpage' %}"
        {% endif %}
    >
        <button
            id="makeStartPageButton"
            type="submit"
            class="btn btn-sm btn-light"
            {% if user.startpage == page %}disabled{% endif %}
        >
            <span id="span-set-startpage" class="fas fa-house"></span>
        </button>
    </span>
</form>
```

niklasmohrin (changes requested):

evap/evaluation/views.py
```diff
@@ -36,6 +36,11 @@ def redirect_user_to_start_page(user):
             return redirect("grades:semester_view", active_semester.id)
         return redirect("grades:index")
 
+    if user.startpage == UserProfile.StartPage.STUDENT and user.is_student:
+        return redirect("student:index")
```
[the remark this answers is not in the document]
janno42: I added the grades page as a possible start page (only the grades page of the active semester) and renamed `Undefined` to `Default`.

You can only be a reviewer when in staff mode, the start pages don't apply there.

evap/evaluation/templates/startpage_button.html
janno42: i prefer not to have a form at all if not necessary. i see your point, but find the approach with two separate definitions more readable. in theory, the icon on the disabled button could also be different (e.g., a person inside the house) - i don't think of it as "the same button".

evap/evaluation/views.py
```diff
@@ -31,11 +31,17 @@ def redirect_user_to_start_page(user):
             return redirect("staff:semester_view", active_semester.id)
         return redirect("staff:index")
 
+    if user.startpage == UserProfile.StartPage.STUDENT and user.is_student:
```
niklasmohrin, line 34: Should this not be `is_participant`?

evap/evaluation/models.py
```diff
@@ -1827,6 +1840,12 @@ def is_editor_or_delegate(self):
     def is_responsible_or_contributor_or_delegate(self):
         return self.is_responsible or self.is_contributor or self.is_delegate
 
+    @cached_property
+    def show_startpage_button(self):
+        return [self.is_student, self.is_responsible_or_contributor_or_delegate, self.is_grade_publisher].count(
+            True
+        ) > 1
```
niklasmohrin, line 1847: Could a user loose their status as for example "student" and then be unable to update their startpage? In addition I think this should also be "participant" instead of student (and maybe the other thing is not a problem anymore then?)

In general, maybe we should have `can_visit_student_index` etc to not have this knowledge duplicated? (Sorry for the general scope bleed :/ )

evap/evaluation/views.py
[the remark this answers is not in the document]
niklasmohrin: Ah, okay then

evap/evaluation/templates/startpage_button.html
niklasmohrin: I am not sure I agree, but we can do it your way :)

He3lixxx (commented):
Lgtm after the conflict of `is_student` vs `is_participant` is resolved.

evap/evaluation/models.py
janno42: i didn't implement anything like `can_visit_student_index` because this might get even more confusing when thinking about staff mode.

He3lixxx (approved):

evap/evaluation/templates/startpage_button.html
```diff
@@ -0,0 +1,13 @@
+{% if user.startpage == page %}
+    <div data-bs-toggle="tooltip" data-bs-placement="left" title="{% trans 'This is your startpage' %}">
+        <button type="button" class="btn {{ btnClass }} btn-light" disabled><span class="fas fa-house"></span></button>
+    </div>
+{% else %}
+    <form id="startpage-form" class="d-flex" method="post" action="{% url 'evaluation:set_startpage' %}">
+        {% csrf_token %}
+        <input name="page" type="hidden" value="{{ page }}">
+        <button id="makeStartPageButton" type="submit" class="btn {{ btnClass }} btn-light" onclick="setSpinnerIcon('span-set-startpage')" data-bs-toggle="tooltip" data-bs-placement="left" title="{% trans 'Make this my startpage' %}">
```
niklasmohrin, line 9: This seems to be unused:
```suggestion
        <button type="submit" class="btn {{ btnClass }} btn-light" onclick="setSpinnerIcon('span-set-startpage')" data-bs-toggle="tooltip" data-bs-placement="left" title="{% trans 'Make this my startpage' %}">
```

niklasmohrin (approved):

PR #1998, opened 2023-08-21
Control name change on import

### LINKED ISSUE #1490: Name change control on import
When an import would change a user's name (title, first name, or last name) currently a warning is shown, informing about the change. This should be changed to a process in which managers can decide for each individual user whether the respective change should be applied on import or not.
If the managers decide not to apply the name change, the original name will be kept in the database. In all other points the importer will behave as before.

Kakadus:
fix #1490 

This is a first approach to this issue. Some parts need refactoring, or renaming at least.

Ah, and there are no tests (yet).

### DIFF
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -1,5 +1,7 @@
 import logging
 from collections.abc import Iterable
+from dataclasses import dataclass
+from enum import Enum
 
 from django import forms
 from django.contrib.auth.models import Group
@@ -34,6 +36,8 @@ from evap.evaluation.models import (
 from evap.evaluation.tools import clean_email, date_to_datetime
 from evap.results.tools import STATES_WITH_RESULT_TEMPLATE_CACHING, STATES_WITH_RESULTS_CACHING, cache_results
 from evap.results.views import update_template_cache, update_template_cache_of_published_evaluations_in_course
+from evap.staff.importers import ImporterLogEntry
+from evap.staff.importers.base import ImporterDecisionEntry
 from evap.staff.tools import remove_user_from_represented_and_ccing_users
 from evap.student.models import TextAnswerWarning
 
@@ -118,6 +122,45 @@ class UserImportForm(forms.Form):
         required=False,
         widget=forms.FileInput(attrs={"accept": "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"}),
     )
+
+
+@dataclass
+class DecisionFormEntry:
+    imported: str
+    existing: str | None = None
+
+    @classmethod
+    def from_importer_decision_entry(cls, entry: ImporterDecisionEntry):
+        return cls(entry.imported, entry.existing)
+
+    def __eq__(self, other):
+        return self.imported == other.imported and self.existing is None or self.existing == other.existing
+
+
+class MyChoiceField(forms.TypedMultipleChoiceField):
+    def __init__(self, *, category: ImporterLogEntry.Category, **kwargs):
+        super().__init__(**kwargs)
+        self.category = category
+
+    def validate(self, value):
+        """Validate, but do not check if value in choices."""
+        if value == self.empty_value and self.required:
+            raise ValidationError(self.error_messages["required"], code="required")
+
+
+class UserDecisionForm(forms.Form):
+    name_mismatches = MyChoiceField(
+        category=ImporterLogEntry.Category.NAME,
+        coerce=DecisionFormEntry,
+        label=_("Name Mismatches"),
+        required=False,
+    )
+
+    def by_category(self):
+        return {field.category: self.cleaned_data[name] for name, field in self.fields.items()}
+
+    def empty(self):
+        return not any(field.choices for field in self.fields.values())
 
 
 class EvaluationParticipantCopyForm(forms.Form):
@@ -1134,3 +1177,8 @@ class ExportSheetForm(forms.Form):
         self.fields["selected_course_types"] = forms.MultipleChoiceField(
             choices=course_type_tuples, required=True, widget=forms.CheckboxSelectMultiple(), label=_("Course types")
         )
+
+
+class Action(Enum):
+    KEEP = 0
+    CHANGE = 1
--- a/evap/staff/importers/base.py
+++ b/evap/staff/importers/base.py
@@ -1,4 +1,5 @@
 import itertools
+import typing
 from abc import ABC, abstractmethod
 from collections import Counter, namedtuple
 from collections.abc import Iterable, Iterator
@@ -12,6 +13,11 @@ from django.conf import settings
 from django.contrib import messages
 from django.utils.translation import gettext as _
 from django.utils.translation import gettext_lazy, ngettext
+
+from evap.evaluation.models import UserProfile
+
+if typing.TYPE_CHECKING:
+    from evap.staff.importers.user import UserData
 
 
 @dataclass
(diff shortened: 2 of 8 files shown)


evap/staff/templates/staff_user_import.html
```diff
@@ -12,48 +12,69 @@
...
             <div class="card-body">
                 <p>
                     {% trans 'Upload Excel file' %}
-                    (<a href="{% url 'staff:download_sample_file' 'sample_user.xlsx' %}">{% trans 'Download sample file' %}</a>,
-                    <button type="button" class="btn btn-link" onclick="copyHeaders(['Title', 'First name', 'Last name', 'Email'])">{% trans 'Copy headers to clipboard' %}</button>).
+                    (<a
+                    href="{% url 'staff:download_sample_file' 'sample_user.xlsx' %}">{% trans 'Download sample file' %}</a>,
+                    <button type="button" class="btn btn-link"
+                            onclick="copyHeaders(['Title', 'First name', 'Last name', 'Email'])">{% trans 'Copy headers to clipboard' %}</button>
+                    ).
```
Kakadus, line 28: These reformats will be removed

evap/staff/templates/staff_user_importer.html
```diff
@@ -0,0 +1,114 @@
...
+                                </div>
+                            </div>
+                        </div>
+                    </div>
+                {% endif %}
+            {% endfor %}
+        {% endfor %}
+        <button type="submit" class="btn btn-primary mb-3">
+            {% trans 'Resolve and download excel' %}
+        </button>
```
Kakadus, line 107: Currently, the submit button is not next to the `Upload and test`. This may look weird. The problem is, that this is a different form.

evap/staff/views.py
```diff
@@ -2147,7 +2154,14 @@ def user_import(request):
...
                 __, importer_log = import_users(file_content, test_run=True)
-                if not importer_log.has_errors():
+                decision_form.fields["name_mismatches"].choices = (
+                    (str(i), DecisionFormEntry(decision.imported, decision.existing))
+                    for i, decision in enumerate(
+                        importer_log.decisions_by_category().get(ImporterDecisionEntry.Category.NAME, [])
+                    )
+                )
+
+                if not importer_log.has_errors() and (not importer_log.has_decisions() or operation == "test"):
```
He3lixxx, line 2164: `or operator == "test"` inside a `if operator == "test"` block

evap/staff/forms.py
```diff
@@ -34,6 +36,8 @@
 from evap.evaluation.tools import clean_email, date_to_datetime
 from evap.results.tools import STATES_WITH_RESULT_TEMPLATE_CACHING, STATES_WITH_RESULTS_CACHING, cache_results
 from evap.results.views import update_template_cache, update_template_cache_of_published_evaluations_in_course
+from evap.staff.importers import ImporterLogEntry
+from evap.staff.importers.base import ImporterDecisionEntry
```
He3lixxx, line 40: I don't like the coupling and all the related problems this has with the importers while not being fully integrated into the process. Essentially, we get the worst of both worlds (code is coupled and UX is cumbersome)

@janno42 Do we want this only for user imports, or should it also work with enrollment imports?

evap/staff/importers/base.py
```diff
@@ -13,6 +14,11 @@
 from django.utils.translation import gettext as _
 from django.utils.translation import gettext_lazy, ngettext
 
+from evap.evaluation.models import UserProfile
+
+if typing.TYPE_CHECKING:
+    from evap.staff.importers.user import UserData
```
He3lixxx, line 20: I'd prefer preventing circular imports by putting all the new stuff in `importers/decisions.py` or similar

evap/staff/importers/base.py
```diff
@@ -56,11 +62,29 @@ class Category(Enum):
     message: str
 
 
+@dataclass
+class ImporterDecisionEntry(ImporterLogEntry):
```
He3lixxx, line 66: Not a big fan of having this in the log, I think the log and the decisions are semantically different. If we have both in the same structure, we should rename it, as it is no more a "log" of things that happened.

evap/staff/importers/base.py
```diff
@@ -56,11 +62,29 @@ class Category(Enum):
...
+    existing: str
+    imported: str
+    user_object: UserProfile
+    user_data: "UserData"
+    resolved: bool = False
+
+    def apply(self):
```
He3lixxx, line 73: "Decision" and "apply" do not match.

When thinking about how we could structure this, my mind went through compiler errors / messages and corresponding autofix features. I wonder how clangd manages this internally.

I'd not be opposed of changing our model to that of a compiler. We could rename "Log" to "Diagnostics", where each entry is associated with a checker, and checkers have a "autofix" feature. I guess the most annoying part is that checkers by design work on a heavily preprocessed representation to make them as simple as possible. We'd always have to re-build the `xlsx` file from scratch in the end instead of modifying an existing one. I wonder how clangd and similar do this internally.

evap/staff/templates/staff_user_import.html
```diff
@@ -12,48 +12,69 @@
...
             <div class="card-body">
-                {% if not test_passed %}
-                    <button name="operation" value="test" type="submit" class="btn btn-light form-submit-btn">{% trans 'Upload and test' %}</button>
-                {% else %}
-                    <button name="operation" value="test" type="submit" class="btn btn-light form-submit-btn">{% trans 'Upload and test' %}</button>
-                    <button name="operation" value="import" type="button" onclick="importUserModalShow('import');" class="btn btn-primary form-submit-btn">{% trans 'Import previously uploaded file' %}</button>
+                <button name="operation" value="test" type="submit"
+                        class="btn btn-light form-submit-btn">{% trans 'Upload and test' %}</button>
+                {% if test_passed %}
+                    {% if not importer_log.has_decisions %}
```
He3lixxx, line 39: If I opt out of a "autofix" in the xlsx file, the name conflict will remain in the file, so I will always get the decision again in future "test" runs, right? This means that this is not a valid UI condition here, right?

We'd probably have to have some javascript that dynamically tests whether any decision/autofix is set to "please fix my xlsx file", and in that case, only enable the "download fixed file". When all autofixes are set to "no, the file is correct", the "import" button be enabled.

evap/staff/importers/base.py
He3lixxx: Another thought, even more annoying: Compiler autofixes sometimes fail / do weird stuff. In an IDE, that's usually not a problem, as I can observe the change right on my screen, and git / PR review will also notice a shady transformation.

With our current approach, users will not get a nice diff view, so they'll likely not notice if anything went wrong.

evap/staff/views.py
```diff
@@ -2164,12 +2178,72 @@ def user_import(request):
...
         {
             "excel_form": excel_form,
+            "decision_form": decision_form,
             "importer_log": importer_log,
             "test_passed": test_passed,
         },
     )
 
 
+def export_users(file, users):
```
He3lixxx, line 2188: Maybe the same component that initially transforms the `xlsx` file to our internal representation should also be responsible for building a `xlsx` file again, so we just have the logic encapsulated in one place.

evap/staff/views.py
```diff
@@ -2164,12 +2178,72 @@ def user_import(request):
...
+    file_content = get_import_file_content_or_raise(request.user.id, import_type=ImportType.USER)
+
+    imported_users, importer_log = import_users(file_content, test_run=True)
+
+    open_decisions = importer_log.decisions_by_category().get(ImporterLogEntry.Category.NAME, [])
+    user_decisions = {decision.imported for decision in decision_form.cleaned_data["name_mismatches"]}
+
+    to_resolve = []
+    for decision in open_decisions:
+        if decision.imported in user_decisions:
```
He3lixxx, line 2222: N^2 run time

PR #2011, opened 2023-08-30
Add enrollment preprocessing CLI tool

### LINKED ISSUE #1490: Name change control on import

Kakadus:
fix #1490 

Upload my progress on the cli tool from monday.

- [x] There needs to be a link to the user export route somewhere.
- [x] There is probably a test missing for the cli tool. 
- [x] The django-pylint plugin has a problem with linting the preprocessor (I manually linted it though)
- [x] codecov does not see the tools


tools/enrolment_preprocessor.py
```diff
@@ -0,0 +1,60 @@
...
+
+def fix_users(users: dict[str, User], title: Cell | None, last_name: Cell, first_name: Cell, email: Cell):
+    imported = User(last_name=last_name.value, first_name=first_name.value, email=email.value, title=title.value or "" if title else "")  # type: ignore[arg-type]  # if schema is correct, all values are strings
+    existing = users.setdefault(imported.email, imported)
+    if existing != imported:
+        print("There is a conflict in the user data.")
+        print(f"existing: {existing}.")
+        print(f"imported: {imported}.")
+        if input("Do you want to keep the existing user? (y/n)")[0].lower() == "n":
+            return
```
He3lixxx, line 27: I think the decisions would have to be based on each individual attribute (title, first name, last name).

tools/enrolment_preprocessor.py
```diff
@@ -0,0 +1,60 @@
...
+    if "python" in args[0]:
+        args = args[1:]
+    parser = ArgumentParser(description="Commandline tool to preprocess enrollment xlsx files.")
+    parser.add_argument(
+        "-ud", "--user-data", help="Path to a csv file containing an export of all existing users.", required=True
+    )
+    parser.add_argument(
+        "-ed", "--enrolment-data", help="Path to the enrolment data in xlsx format for import.", required=True
+    )
+    ns = parser.parse_args(args[1:])
```
He3lixxx, line 47: I've never seen anyone mess with `sys.argv` manually here. For me, locally, `python3` is stripped from `argv` if I run a script as `python3 test.py some_args`.

tools/enrolment_preprocessor.py
```diff
@@ -0,0 +1,60 @@
...
+if __name__ == "__main__":
+    args = sys.argv
+    if "python" in args[0]:
+        args = args[1:]
+    parser = ArgumentParser(description="Commandline tool to preprocess enrollment xlsx files.")
+    parser.add_argument(
+        "-ud", "--user-data", help="Path to a csv file containing an export of all existing users.", required=True
+    )
+    parser.add_argument(
+        "-ed", "--enrolment-data", help="Path to the enrolment data in xlsx format for import.", required=True
```
He3lixxx, line 45: we use the american variant, "enrollment".

evap/staff/views.py
```diff
@@ -2116,6 +2118,25 @@ def user_list(request):
     return render(request, "staff_user_list.html", {"users": users, "filter_users": filter_users})
 
 
+@manager_required
+def user_export(request):
+    filter_users = get_parameter_from_url_or_session(request, "filter_users")
+
+    users = UserProfile.objects.all()
+    if filter_users:
+        users = users.exclude(is_active=False)
```
He3lixxx, line 2127: I don't think we need this filter?

evap/staff/tests/test_views.py
```diff
@@ -529,6 +531,48 @@ def test_wrong_files_dont_crash(self):
...
+    def setUpTestData(cls) -> None:
+        cls.manager = make_manager()
+        baker.make(UserProfile, _quantity=5, _fill_optional=["first_name_given", "last_name", "email", "title"])
+        baker.make(
+            UserProfile,
+            _quantity=5,
+            _fill_optional=["first_name_given", "last_name", "email", "title"],
+            is_active=False,
+        )
+        # the titles are not filled by baker
```
He3lixxx, line 547: They should be if passed as `_fill_optional`. If that doesn't work, we should probably open an issue with them. You can pass an iterator to `baker.make` as the `title` kwarg, it will then use these values.

Kakadus: It had an default value, so it is not filled

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,82 @@
...
+        imported_cells.title.value = user_decision("title", existing.title, imported.title)
+    imported_cells.last_name.value = user_decision("last name", existing.last_name, imported.last_name)
+    imported_cells.first_name.value = user_decision("first name", existing.first_name, imported.first_name)
+    imported_cells.email.value = user_decision("email", existing.email, imported.email)
+    print()
+
+
+def run_preprocessor(enrollment_data: str | BytesIO, user_data: TextIO):
+    workbook = load_workbook(enrollment_data)
+    users = {}
```
richardebeling, line 61: Shouldn't this require type annotations?

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,82 @@
...
+def user_decision(field: str, existing: str, imported: str) -> str:
+    if existing == imported:
+        return existing
+    decision = input(f"Do you want to keep the existing user {field}? (y/n) ")
+    if decision and decision[0].lower() == "n":
+        return imported
+    return existing
+
+
+def fix_users(users: dict[str, User], imported_cells: UserCells):
```
richardebeling, line 43: should be singular, as it only fixes one user, right?

richardebeling (commented):
Looking good! :)

tools/enrollment_preprocessor.py
Kakadus: no, because mypy is able to infer the type the way data is filled into the dict

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,86 @@
...
+            str(self.first_name.value),
+            str(self.email.value),
+        )
+
+
+def user_decision(field: str, existing: str, imported: str) -> str:
+    if existing == imported:
+        return existing
+    decision = input(f"Do you want to keep the existing user {field}? (y/n) ")
+    if decision and decision[0].lower() == "n":
```
richardebeling, line 41: This makes "y" the default on any other input that "n", right? I'd say we want stricter input validation by repeatedly asking the user until the input is either empty (=default), "y", or "n". If someone entered "x", they likely made a mistake and possibly intended to type "y".

If we keep a "default for empty input", this should be shown on the prompt by using "(Y/n)"

@janno42 opinions? Do we want a "no input == yes" policy? I'd think we wouldn't want that.

(side note: I think we have similar "ask the user for yes/no" logic in multiple places. Could be deduplicated/shared)

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,86 @@
...
+def fix_user(users: dict[str, User], imported_cells: UserCells) -> None:
+    imported = imported_cells.value()
+    existing = users.setdefault(imported.email, imported)
+    if imported == existing:
+        return
+    print("There is a conflict in the user data.")
+    print(f"existing: {existing}.")
+    print(f"imported: {imported}.")
+    if imported_cells.title is not None:
+        imported_cells.title.value = user_decision("title", existing.title, imported.title)
```
richardebeling, line 55: Maybe add a comment "None is passed exclusively for participants since they have no title column"? Took me quite a while to understand, at first I thought this was some (wrong) auto-update logic for titles

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,86 @@
...
+    imported_cells.last_name.value = user_decision("last name", existing.last_name, imported.last_name)
+    imported_cells.first_name.value = user_decision("first name", existing.first_name, imported.first_name)
+    imported_cells.email.value = user_decision("email", existing.email, imported.email)
+    print()
+
+
+def run_preprocessor(enrollment_data: str | BytesIO, user_data: TextIO) -> None:
```
richardebeling, line 62: Can `enrollment_data` really be a `str`? I'd expect `load_workbook` to call `read` on it. Maybe [`typing.IO`](https://docs.python.org/3/library/typing.html#typing.IO) is the right annotation here (since I'm in favor of not modifying the argument but returning a new `BytesIO` instance).

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+if __name__ == "__main__":  # pragma: nocover
+    parser = ArgumentParser(description="Commandline tool to preprocess enrollment xlsx files.")
+    parser.add_argument(
+        "-u", "--user-data", help="Path to a csv file containing an export of all existing users.", required=True
+    )
+    parser.add_argument(
+        "-e", "--enrollment-data", help="Path to the enrollment data in xlsx format for import.", required=True
+    )
```
richardebeling, line 83: This makes it so that a user _has_ to at least specify `-e` and `-u`, you can't just paste two paths as arguments, right?

If we wanted to keep it simple, we could just make them both positional. I'd say this would be common practice if you have two required arguments that you can't really accidentally switch up (as a .csv file will not pass xlsx loading). But I'm also fine with the current named approach.

richardebeling (commented):
More in-depth this time, so a few more comments, but it's all very detail-y.

tools/enrollment_preprocessor.py
Kakadus: yes, it is the PATH to the excel file. Why? Idk. Otherwise, I would need to wrap TextIO/BinaryIO to BytesIO for the cli

tools/enrollment_preprocessor.py
Kakadus: I added it in [ce8c35e](https://github.com/e-valuation/EvaP/pull/2011/commits/ce8c35e12081ea3211987938c34944d691dd239d), we can revert it, if needed

evap/staff/views.py
```diff
@@ -2120,6 +2122,21 @@ def user_list(request):
     return render(request, "staff_user_list.html", {"users": users, "filter_users": filter_users})
 
 
+@manager_required
+def user_export(request):
+    io = BytesIO()
+    Writer = codecs.getwriter("utf-8")
```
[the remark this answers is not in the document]
Kakadus: encoding has no default here :(
the problem is that I need a `TextIIO` for csv AND `BytesIO` for the `FileResponse`

tools/enrollment_preprocessor.py
janno42: I wouldn't set a default for now and force the user to make a choice. If this turns out to be too annoying with real data, we can change it later.

evap/staff/views.py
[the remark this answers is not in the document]
niklasmohrin: We also have `AttachmentResponse` which seems to work without `codecs`

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,89 @@
...
+    wb_out = BytesIO()
+    workbook.save(wb_out)
+    return wb_out
+
+
+if __name__ == "__main__":  # pragma: nocover
+    parser = ArgumentParser(description="Commandline tool to preprocess enrollment xlsx files.")
+    parser.add_argument("user-data", help="Path to a csv file containing an export of all existing users.")
+    parser.add_argument("enrollment-data", help="Path to the enrollment data in xlsx format for import.")
+    ns = parser.parse_args(sys.argv)
```
niklasmohrin, line 85:
```suggestion
    ns = parser.parse_args()
```

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,89 @@
...
+    with open(ns.user_data, encoding="utf-8") as csvfile:
```
niklasmohrin, line 86:
```suggestion
    with open(ns.user_data) as csvfile:
```

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,89 @@
...
+        wb = run_preprocessor(ns.enrollment_data, csvfile)
+    with open(ns.enrollment_data, "wb", encoding="utf-8") as out:
+        out.write(wb.read())
```
niklasmohrin, line 89: I thought about this for a minute and wasn't sure whether this needs another seek of `wb` and I wouldn't bet my life on `openpyxl` not messing this up in the future. Should we add an integration test that starts this script and inspects the generated files?

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+if __name__ == "__main__":  # pragma: nocover
+    parser = ArgumentParser(description="Commandline tool to preprocess enrollment xlsx files.")
+    parser.add_argument("user-data", help="Path to a csv file containing an export of all existing users.")
+    parser.add_argument("enrollment-data", help="Path to the enrollment data in xlsx format for import.")
+    ns = parser.parse_args(sys.argv)
+    with open(ns.user_data, encoding="utf-8") as csvfile:
+        wb = run_preprocessor(ns.enrollment_data, csvfile)
+    with open(ns.enrollment_data, "wb", encoding="utf-8") as out:
```
niklasmohrin, line 88: If I understand the documentation of `open` correctly, the encoding is ignored in binary (`b`) mode

tools/enrollment_preprocessor.py
richardebeling: (I think `run_preprocessor` should just seek to 0)

tools/enrollment_preprocessor.py
Kakadus: this is required by pylint: [W1514](https://pylint.pycqa.org/en/latest/user_guide/messages/warning/unspecified-encoding.html)

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,89 @@
...
+    print()
+
+
+def run_preprocessor(enrollment_data: str | BytesIO, user_data: TextIO) -> BytesIO:
+    workbook = load_workbook(enrollment_data)
+    users = {}
+    reader = csv.reader(user_data, delimiter=";", lineterminator="\n")
+    for row in reader:
+        email = row[-1]
+        users[email] = User(*row)
```
richardebeling, line 70: Maybe, to not duplicate the "last column is email" knowledge:
```suggestion
        user = User(*row)
        users[user.email] = user
```

richardebeling (approved):

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,89 @@
...
+    wb_out = BytesIO()
+    workbook.save(wb_out)
+    wb_out.seek(0)
+    return wb_out
+
+
+if __name__ == "__main__":  # pragma: nocover
+    parser = ArgumentParser(description="Commandline tool to preprocess enrollment xlsx files.")
+    parser.add_argument("user-data", help="Path to a csv file containing an export of all existing users.")
+    parser.add_argument("enrollment-data", help="Path to the enrollment data in xlsx format for import.")
```
janno42, line 84:
```suggestion
    parser.add_argument("user_data", help="Path to a csv file containing an export of all existing users.")
    parser.add_argument("enrollment_data", help="Path to the enrollment data in xlsx format for import.")
```
unfortunately, the parser doesn't handle the conversion from `-` to `_`

Kakadus: :cry:

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,90 @@
...
+            str(self.email.value),
+        )
+
+
+def user_decision(field: str, existing: str, imported: str) -> str:
+    if existing == imported:
+        return existing
+    decision: str = ""
+    while decision not in ("y", "n"):
+        decision = input(f"Do you want to keep the existing user {field}? (y/n) ")
```
richardebeling, line 41: For me as a first time user of the script, I had to think a bit about what "yes" and "no" imply here, because the question is a bit negated ("Do you want to keep the existing" == "Do you want to not apply the change that is in the data when importing"). Intuitively, I'd expect "yes" to be "Yes, the new data/change is correct" and "no" to mean "No, discard this / please keep the old data". Might be a me-problem though. @janno42 should probably say what the exact prompt should be here.

It might help if we list the value that the user would accept with "y" ->
> Do you want to keep the existing user first_name "Bastius"? (y/n)

.github/workflows/tests.yml
```diff
@@ -89,7 +89,7 @@ jobs:
         uses: ./.github/setup_python
 
       - name: Run linter
-        run: pylint evap -j 0
+        run: pylint evap tools -j 0
```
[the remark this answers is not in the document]
Kakadus: removed the argument in the actions

evap/staff/tests/test_views.py
```diff
@@ -553,6 +555,33 @@ def test_wrong_files_dont_crash(self):
...
+        user_objects = {
+            (user.title or "", user.last_name or "", user.first_name or "", user.email or "")
+            for user in UserProfile.objects.iterator()
+        }
+        response = self.app.get(self.url, user=self.manager)
+
+        reader = csv.reader(response.text.strip().split("\n"), delimiter=";", lineterminator="\n")
+        # skip header
+        next(reader)
+        self.assertSetEqual({tuple(row) for row in reader}, user_objects)
```
niklasmohrin, line 582: If the inputs are sets, it uses `assertSetEqual` on its own:
```suggestion
        self.assertEqual({tuple(row) for row in reader}, user_objects)
```

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python3
```
niklasmohrin, line 1:
```suggestion
#!/usr/bin/env python3

```

evap/staff/views.py
```diff
@@ -2124,6 +2124,18 @@ def user_list(request):
     return render(request, "staff_user_list.html", {"users": users, "filter_users": filter_users})
 
 
+@manager_required
+def user_export(request):
+    response = AttachmentResponse("exported_users.csv")
+    writer = csv.writer(response, delimiter=";", lineterminator="\n")
+    row = (_("Title"), _("Last name"), _("First name"), _("Email"))
+    writer.writerow(row)
```
niklasmohrin, line 2132:
```suggestion
    header_row = (_("Title"), _("Last name"), _("First name"), _("Email"))
    writer.writerow(header_row)
```

evap/staff/tests/test_views.py
```diff
@@ -553,6 +555,33 @@ def test_wrong_files_dont_crash(self):
...
+
+    @classmethod
+    def setUpTestData(cls) -> None:
+        cls.manager = make_manager()
+        # titles are not filled by baker because it has a default, see https://github.com/model-bakers/model_bakery/discussions/346
+        baker.make(
+            UserProfile,
+            _quantity=5,
+            _fill_optional=["first_name_given", "last_name", "email"],
+            title=cycle(("", "Some", "Custom", "Titles")),
```
niklasmohrin, line 569: I slightly prefer this version, what do you think?
```suggestion
            title=iter(["", "Some", "Custom", "Titles", ""]),
```

evap/staff/tests/test_views.py
```diff
@@ -553,6 +555,33 @@ def test_wrong_files_dont_crash(self):
         self.assertIn("An error happened when processing the file", reply)
 
 
+class TestUserExportView(WebTestStaffMode):
+    url = "/staff/user/export"
```
niklasmohrin, line 559: Let's use `reverse`?

niklasmohrin (approved):
code looks good, just some nits

janno42 (changes requested):
- Please add trimming so that a change from `Test name` to ` Test name` won't be shown as a prompt.
- Users from the import file with empty email addresses should be skipped. They will result in errors during the import and should not be handled here (because they conflict with all existing accounts with empty email addresses).

I will continue real-world data tests after these changes :)

janno42 (changes requested):
As discussed:
- For each user (identified by distinct email) changes should be shown only once (per attribute)
- Please order the changes by attribute
- The attribute with changes should be shown at the front (and highlighted in bold), followed by the full user name with email (without the attribute names)
- The choice should be e/i for existing/imported
- The result should be written in a new file (original file name plus timestamp) and the name of this file should be printed at the end

It could look like this (with missing bold highlighting):
```
First name
----------

existing: Oma (Oma Abner, oma.abner@institution.example.com)
imported: Omar (Dr. Omar Abner, oma.abner@institution.example.com)
Which should be used? (e/i):
```

Kakadus:
I've added single quotes to the conflict output, because the title might be empty.
Currently, a renamed file is created, even if there are no changes. Maybe in this case, no excel file should be created?

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,123 @@
...
+    parser = ArgumentParser(description="Commandline tool to preprocess enrollment xlsx files.")
+    parser.add_argument("user_data", help="Path to a csv file containing an export of all existing users.")
+    parser.add_argument("enrollment_data", help="Path to the enrollment data in xlsx format for import.")
+    ns = parser.parse_args()
+
+    target = Path(ns.enrollment_data)
+
+    with open(ns.user_data, encoding="utf-8") as csvfile:
+        wb = run_preprocessor(target, csvfile)
+    with open(target.with_stem(f"{target.stem}_{datetime.now().isoformat()}"), "wb") as out:
```
janno42, line 122: Fails with `FileNotFoundError: [Errno 2] No such file or directory: '/opt/evap/enrollment_data_2024-01-22T17:38:56.614843.xlsx'` for me on saving the new file; same when using a relative path for the enrollment data

janno42, line 122: > I've added single quotes to the conflict output, because the title might be empty. Currently, a renamed file is created, even if there are no changes. Maybe in this case, no excel file should be created?

Not creating a file if there are no changes would make it even better :)

tools/enrollment_preprocessor.py
```diff
@@ -16,63 +20,89 @@ class User:
     first_name: str
     email: str
 
+    def full_name(self) -> str:
+        prefix = ""
+        if self.title:
+            prefix = f"{self.title} "
+        return f"{prefix}{self.first_name} {self.last_name}, {self.email}"
```
richardebeling, line 27: Canonical representation on importers, as far as I know, is
> firstname lastname (email)

with email in parentheses. But no strong opinion.

tools/enrollment_preprocessor.py
```diff
@@ -16,63 +20,89 @@ class User:
...
+        for field in iter(self):
+            if not field or not field.value:
+                yield ""
+                continue
+            field.value = str(field.value or "").strip()
+            yield field.value
+
+
+def make_bold(text: str) -> str:
+    return f"\x1b[1m{text}\x1b[0m"
```
richardebeling, line 46: I think I'm more used to the octal representation of the escape code (`\033[1m`), but no strong opinion either

tools/enrollment_preprocessor.py
richardebeling: Windows (or NTFS?) file names can't contain `:`, I guess that's the issue.

tools/enrollment_preprocessor.py
Kakadus: This would lead to ` Oma (Oma Abner (oma.abner@institution.example.com))` or so. The tool already separates the conflicting field value from the full name with parenthesis

tools/enrollment_preprocessor.py
Kakadus: done :smiley:

janno42 (changes requested):
Scenario: A user is listed twice in the import file. Once with 'Dr. Eva' as first name and once with 'Eva' as first name. The database holds 'Dr. Eva'.
The tool will not find the first to be a conflict and will not ask about it. It will ask about the second and when choosing 'Eva' (imported), this is no change in the file, so nothing happens.
When now importing, the importer will fail with an error, because the user's data is inconsistent.

What instead should happen: Both occurrences in the import file are set to 'Eva' because this is the chosen value.
So the tool should probably make two runs: In a first run collect all choices and in the second run set the values everywhere.

Kakadus:
> Scenario: A user is listed twice in the import file. Once with 'Dr. Eva' as first name and once with 'Eva' as first name. The database holds 'Dr. Eva'. The tool will not find the first to be a conflict and will not ask about it. It will ask about the second and when choosing 'Eva' (imported), this is no change in the file, so nothing happens. When now importing, the importer will fail with an error, because the user's data is inconsistent.
> 
> What instead should happen: Both occurrences in the import file are set to 'Eva' because this is the chosen value. So the tool should probably make two runs: In a first run collect all choices and in the second run set the values everywhere.

The general problem is, that the conflict is only resolved locally, not everywhere in the excel file

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,163 @@
...
+                continue
+
+            for field in ["title", "last_name", "first_name"]:
+                field_value = getattr(cells, field)
+                if field_value is not None and getattr(existing, field) != getattr(imported, field):
+                    groups[field].append((cells, existing.email))
+    return groups
+
+
+def parse_existing(user_data: TextIO) -> dict[str, User]:
```
richardebeling, line 76: maybe `csv_users_by_email` (possibly with `get_` or `parse_` prefix)?

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,163 @@
...
+    users = {}
+    reader = csv.reader(user_data, delimiter=";", lineterminator="\n")
+    next(reader)  # skip header
+    for row in reader:
+        user = User(*row)
+        users[user.email] = user
+    return users
+
+
+def parse_imported(enrollment_data: Workbook):
```
richardebeling, line 86: maybe `user_cells_by_email` (possibly with `get_` or `parse_` prefix)?

Also: Annotate return type

richardebeling: Does it make sense to keep the "cells" for such a long time, or could we very early just turn them into a `User` early on? Then, we wouldn't need `UserCells` (which I find a bit unintuitive anyway), and we'd just have the users from the csv file (maybe call them database_users or whatever) and the users from the xls import file (call them import users or whatever).

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,163 @@
...
+            imported = cells.clean_user()
+            existing = decisions[existing_email]  # existing is current user decision
+
+            if getattr(existing, field) == getattr(imported, field):
+                continue
+
+            print(f"existing: '{make_bold(getattr(existing, field))}' ({existing.full_name()})")
+            print(f"imported: '{make_bold(getattr(imported, field))}' ({imported.full_name()})")
+
+            decision: str = ""
```
richardebeling, line 112: Do we need `: str` here? I'd expect any type checker to be able to infer that

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,163 @@
...
+
+    import_data = parse_imported(workbook)
+    decisions = get_user_decisions(
+        parse_existing(user_data),
+        import_data,
+    )
+
+    # apply decisions
+    changed = False
+    for email, user in decisions.items():
```
richardebeling, line 133: Currently O(n^2). I'd just
* iterate over the rows in the import (xlsx) file
* For each row, look up the values from `decisions` (maybe call it `correct_user_data_by_email`?) and replace mismatches.


Could we just replace the cell values (and possibly overwrite with already existing data) and just check equality of the workbook we modified with the workbook we were passed to determine `changed`?

Kakadus: I doubt the O(n^2) thing: In total, each cell can be in `outdated_user` at most once throughout the first loop. So, the innermost loop is executed at most the number of user cells times, which should be optimal.
Initially, I just did not want to iterate over the file again. You are right though, this might be cleaner now

Kakadus: I don't think you can do that with the workbook, as the file is completely rewritten when saved. As I can see, there is no proper equality operator defined on the `Workbook` class. Also, the file changes when the identical workbook is saved again :cry:

tools/enrollment_preprocessor.py
Kakadus: Yes and no. My idea before https://github.com/e-valuation/EvaP/pull/2011/commits/46481147f456a6a7f267f7e08a2e37880c0bef4e was to hold a mapping between user (email) and cells, so that they can later be easily updated with the changed userdata.
Current (46481147f456a6a7f267f7e08a2e37880c0bef4e) parses the excel file again and writes thereby the data.

Kakadus:
I were ably to combine initial excel parsing and conflict grouping, because of direct conversion from Cells to Users, but the update logic needed more complexity (Codacy is not very happy)

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,160 @@
...
+        for imported in chain.from_iterable(map(user_from_row, sheet.iter_rows(min_row=2, min_col=2))):
+            if not imported.email:
+                continue
+            existing = database_users.setdefault(imported.email, imported)
+            if imported == existing:
+                continue
+            for field in ["title", "last_name", "first_name"]:
+                field_value = getattr(imported, field)
+                if field_value is not None and getattr(existing, field) != getattr(imported, field):
+                    conflicts_by_field[field].append(imported)
```
richardebeling, line 83: The first two lines are superfluous and basically just repeat the individual tests from below, right? I would remove them then:
```suggestion
            for field in ["title", "last_name", "first_name"]:
                field_value = getattr(imported, field)
                if field_value is not None and getattr(existing, field) != getattr(imported, field):
                    conflicts_by_field[field].append(imported)
```

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # apply decisions
+    changed = False
+    for sheet in workbook.worksheets:
+        for wb_row in sheet.iter_rows(min_row=2, min_col=2):
+            if wb_row[2] and wb_row[2].value:
+                changed = (
+                    update_cells(UserCells(None, *wb_row[:3]), correct_user_data_by_email[wb_row[2].value.strip()])
+                    or changed
+                )
```
richardebeling, line 132:
```suggestion
                changed |= update_cells(UserCells(None, *wb_row[:3]), correct_user_data_by_email[wb_row[2].value.strip()]
```

(repeated below)

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,160 @@
...
+
+            setattr(database_users[choice.email], field, getattr(choice, field))
+        print()
+    return database_users
+
+
+def update_cells(cells: UserCells, user: User) -> bool:
+    changed = False
+    for cell, field_value in zip(cells, [user.title, user.last_name, user.first_name, user.email], strict=True):
+        if cell and cell.value != field_value:
```
richardebeling, line 110: Could we do
```suggestion
        if cell is not None and cell.value != field_value:
```
to make our intention more explicit

tools/enrollment_preprocessor.py
```diff
@@ -0,0 +1,160 @@
...
+            while decision not in ("e", "i"):
+                decision = input("Which one should be used? (e/i):\n")
+            choice = existing if decision == "e" else imported
+
+            setattr(database_users[choice.email], field, getattr(choice, field))
+        print()
+    return database_users
+
+
+def update_cells(cells: UserCells, user: User) -> bool:
```
richardebeling, line 107: Maybe make a method on `UserCells`, called `update_from` or something?

richardebeling (approved):

janno42 (commented):
As discussed: The tool asks for the same user multiple times.

janno42 (approved):
Seems to work now :shipit:

Kakadus:
Just to be sure: Currently the whole userdata from the conflict is shown, so this question is possible:
> First_name
> \---------
> existing: 'Bastius' (Bastius Prorsus, same@email.com)
imported: 'Lucilia' (Lucilia Manilium, same@email.com)

Selecting imported only applies the conflicting field, here the first name (resulting data: Lucilia Prorsus, same@email.com)
Of cause, you are then asked about Prorsus vs Manilium later.  
Instead, we could show the resulting data always, so 
> existing: 'Bastius' (Bastius Prorsus, same@email.com)
imported: 'Lucilia' (Lucilia Prorsus, same@email.com)

is asked instead.

janno42:
We can leave that as it is now. I'd say the second way makes interpretation harder.

PR #2012, opened 2023-08-31
don't remove selected values from multiselect lists

janno42:
Participant lists can get quite long for some evaluations. When manually adding participants from a list, it is sometimes easier to just type in a few names and click to add them, rather than checking to see if the name already exists in the list. In these cases, if a name doesn't show up in the list, the editor may think that the account doesn't exist yet and request that it be created.
This can be avoided by showing all names in the list, whether selected or not.

### DIFF
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -145,6 +145,7 @@
                         },
                         closeAfterSelect: true, // also clears search input on enter.
                         plugins: {},
+                        hideSelected: false,
                     };
                     if(element.hasAttribute("data-tomselect-fullwidth")) {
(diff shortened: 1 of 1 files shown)


He3lixxx (approved):

niklasmohrin (approved):
Hm, a bit sad that the preview doesn't have any indication that the option is already selected (it might be confusing when nothing happens after typing a name and pressing enter). Should we maybe ask tomselect for such a feature?

janno42:
Feel free if you want to :)
I don't think this has a very large impact in our case where the problem mainly occurs because the list of selected users is already quite long and editors don't want to scan the whole list.

Issue #2001, opened 2023-08-21
Add averages to results exports


janno42:
In all Excel results exports, an additional column should be inserted after the questions (in column B), showing the average of all answers given for this question in all evaluations.

The new column should have the header "Average for this question". The lines "degrees", "course type", and the summarizing lines on the bottom should be empty.

fidoriel:
Agreed with @janno42 to only use published evaluations from all semesters.

janno42:
Update: We only want to use Evaluations from non-archived semesters (`results_are_archived == False`) for the average.

PR #2035, opened 2023-10-16
type results exporter

fidoriel:
this PR makes work for #2001 easier

### DIFF
--- a/evap/results/exporters.py
+++ b/evap/results/exporters.py
@@ -1,14 +1,19 @@
 import warnings
 from collections import OrderedDict, defaultdict
+from collections.abc import Iterable, Sequence
 from itertools import chain, repeat
+from typing import Any, TypeVar
 
 import xlwt
-from django.db.models import Q
+from django.db.models import Q, QuerySet
+from django.db.models.base import Model
 from django.utils.translation import gettext as _
 
-from evap.evaluation.models import CourseType, Degree, Evaluation, Questionnaire
+from evap.evaluation.models import CourseType, Degree, Evaluation, Question, Questionnaire, Semester, UserProfile
 from evap.evaluation.tools import ExcelExporter
 from evap.results.tools import (
+    AnsweredRatingResult,
+    QuestionResult,
     RatingResult,
     calculate_average_course_distribution,
     calculate_average_distribution,
@@ -16,6 +21,10 @@ from evap.results.tools import (
     get_grade_color,
     get_results,
 )
+
+T = TypeVar("T", bound=Model)
+QuerySetOrSequence = QuerySet[T] | Sequence[T]
+AnnotatedEvaluation = Any
 
 
 class ResultsExporter(ExcelExporter):
@@ -38,24 +47,24 @@ class ResultsExporter(ExcelExporter):
         **ExcelExporter.styles,
     }
 
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__()
 
         for index, color in self.COLOR_MAPPINGS.items():
             self.workbook.set_colour_RGB(index, *color)
 
     @classmethod
-    def grade_to_style(cls, grade):
+    def grade_to_style(cls, grade: float) -> str:
         return "grade_" + str(cls.normalize_number(grade))
 
     @classmethod
-    def normalize_number(cls, number):
+    def normalize_number(cls, number: float) -> float:
         """floors 'number' to a multiply of cls.STEP"""
         rounded_number = round(number, 1)  # see #302
         return round(int(rounded_number / cls.STEP + 0.0001) * cls.STEP, 1)
 
     @classmethod
-    def init_grade_styles(cls):
+    def init_grade_styles(cls) -> None:
         """
         Adds the grade styles to cls.styles and as a xlwt identifier.
         This also notes all registered colors in cls.COLOR_MAPPINGS for the instances.
@@ -88,7 +97,7 @@ class ResultsExporter(ExcelExporter):
             cls.styles[style_name] = xlwt.easyxf(grade_base_style.format(color_name), num_format_str="0.0")
 
     @staticmethod
-    def filter_text_and_heading_questions(questions):
+    def filter_text_and_heading_questions(questions: Iterable[Question]) -> list[Question]:
         questions = [question for question in questions if not question.is_text_question]
 
         # remove heading questions if they have no "content" below them
@@ -103,16 +112,23 @@ class ResultsExporter(ExcelExporter):
         return filtered_questions
 
     @staticmethod
-    def filter_evaluations(semesters, evaluation_states, degrees, course_types, contributor, include_not_enough_voters):
+    def filter_evaluations(
+        semesters: Iterable[Semester],
+        evaluation_states: Iterable[Evaluation.State],
+        degree_ids: Iterable[int],
+        course_type_ids: Iterable[int],
+        contributor: UserProfile | None,
+        include_not_enough_voters: bool,
+    ) -> tuple[list[tuple[Evaluation, OrderedDict[int, list[QuestionResult]]]], list[Questionnaire], bool]:
         # pylint: disable=too-many-locals
         course_results_exist = False
         evaluations_with_results = []
-        used_questionnaires = set()
+        used_questionnaires: set[Questionnaire] = set()
         evaluations_filter = Q(
             course__semester__in=semesters,
             state__in=evaluation_states,
-            course__degrees__in=degrees,
-            course__type__in=course_types,
+            course__degrees__in=degree_ids,
+            course__type__in=course_type_ids,
         )
         if contributor:
             evaluations_filter = evaluations_filter & (
@@ -124,13 +140,13 @@ class ResultsExporter(ExcelExporter):
                 continue
             if not evaluation.can_publish_rating_results and not include_not_enough_voters:
                 continue
-            results = OrderedDict()
+            results: OrderedDict[int, list[QuestionResult]] = OrderedDict()
             for contribution_result in get_results(evaluation).contribution_results:
                 for questionnaire_result in contribution_result.questionnaire_results:
                     # RatingQuestion.counts is a tuple of integers or None, if this tuple is all zero, we want to exclude it
-                    if all(
-                        not question_result.question.is_rating_question or not RatingResult.has_answers(question_result)
-                        for question_result in questionnaire_result.question_results
+                    question_results = questionnaire_result.question_results
+                    if not any(
+                        isinstance(question_result, AnsweredRatingResult) for question_result in question_results
                     ):
                         continue
                     if (
@@ -138,30 +154,35 @@ class ResultsExporter(ExcelExporter):
                         or contribution_result.contributor is None
                         or contribution_result.contributor == contributor
                     ):
-                        results.setdefault(questionnaire_result.questionnaire.id, []).extend(
-                            questionnaire_result.question_results
-                        )
+                        results.setdefault(questionnaire_result.questionnaire.id, []).extend(question_results)
                         used_questionnaires.add(questionnaire_result.questionnaire)
-            evaluation.course_evaluations_count = evaluation.course.evaluations.count()
-            if evaluation.course_evaluations_count > 1:
+            annotated_evaluation: AnnotatedEvaluation = evaluation
+            annotated_evaluation.course_evaluations_count = annotated_evaluation.course.evaluations.count()
+            if annotated_evaluation.course_evaluations_count > 1:
                 course_results_exist = True
-                weight_sum = sum(evaluation.weight for evaluation in evaluation.course.evaluations.all())
-                evaluation.weight_percentage = int((evaluation.weight / weight_sum) * 100)
-                evaluation.course.avg_grade = distribution_to_grade(
-                    calculate_average_course_distribution(evaluation.course)
+                weight_sum = sum(evaluation.weight for evaluation in annotated_evaluation.course.evaluations.all())
+                annotated_evaluation.weight_percentage = int((evaluation.weight / weight_sum) * 100)
+                annotated_evaluation.course.avg_grade = distribution_to_grade(
+                    calculate_average_course_distribution(annotated_evaluation.course)
                 )
-            evaluations_with_results.append((evaluation, results))
+            evaluations_with_results.append((annotated_evaluation, results))
 
         evaluations_with_results.sort(
             key=lambda cr: (cr[0].course.semester.id, cr[0].course.type.order, cr[0].full_name)
         )
-        used_questionnaires = sorted(used_questionnaires)
+        sorted_questionnaires = sorted(used_questionnaires)
 
-        return evaluations_with_results, used_questionnaires, course_results_exist
+        return evaluations_with_results, sorted_questionnaires, course_results_exist
 
     def write_headings_and_evaluation_info(
-        self, evaluations_with_results, semesters, contributor, degrees, course_types, verbose_heading
-    ):
+        self,
+        evaluations_with_results: list[tuple[Evaluation, OrderedDict[int, list[QuestionResult]]]],
+        semesters: QuerySetOrSequence[Semester],
+        contributor: UserProfile | None,
+        degrees: Iterable[int],
+        course_types: Iterable[int],
+        verbose_heading: bool,
+    ) -> None:
         export_name = _("Evaluation")
         if contributor:
             export_name += f"\n{contributor.full_name}"
@@ -199,27 +220,32 @@ class ResultsExporter(ExcelExporter):
         # One more cell is needed for the question column
         self.write_empty_row_with_styles(["default"] + ["border_left_right"] * len(evaluations_with_results))
 
-    def write_overall_results(self, evaluations_with_results, course_results_exist):
-        evaluations = [e for e, __ in evaluations_with_results]
+    def write_overall_results(
+        self,
+        evaluations_with_results: list[tuple[AnnotatedEvaluation, OrderedDict[int, list[QuestionResult]]]],
+        course_results_exist: bool,
+    ) -> None:
+        annotated_evaluations = [e for e, __ in evaluations_with_results]
 
         self.write_cell(_("Overall Average Grade"), "bold")
-        averages = (distribution_to_grade(calculate_average_distribution(e)) for e in evaluations)
+        averages = (distribution_to_grade(calculate_average_distribution(e)) for e in annotated_evaluations)
         self.write_row(averages, lambda avg: self.grade_to_style(avg) if avg else "border_left_right")
 
         self.write_cell(_("Total voters/Total participants"), "bold")
-        voter_ratios = (f"{e.num_voters}/{e.num_participants}" for e in evaluations)
+        voter_ratios = (f"{e.num_voters}/{e.num_participants}" for e in annotated_evaluations)
         self.write_row(voter_ratios, style="total_voters")
 
         self.write_cell(_("Evaluation rate"), "bold")
         # round down like in progress bar
         participant_percentages = (
-            f"{int((e.num_voters / e.num_participants) * 100) if e.num_participants > 0 else 0}%" for e in evaluations
+            f"{int((e.num_voters / e.num_participants) * 100) if e.num_participants > 0 else 0}%"
+            for e in annotated_evaluations
         )
         self.write_row(participant_percentages, style="evaluation_rate")
 
         if course_results_exist:
             # Only query the number of evaluations once and keep track of it here.
-            count_gt_1 = [e.course_evaluations_count > 1 for e in evaluations]
+            count_gt_1: list[bool] = [e.course_evaluations_count > 1 for e in annotated_evaluations]
 
             # Borders only if there is a course grade below. Offset by one column
             self.write_empty_row_with_styles(
@@ -228,12 +254,13 @@ class ResultsExporter(ExcelExporter):
 
             self.write_cell(_("Evaluation weight"), "bold")
             weight_percentages = (
-                f"{e.weight_percentage}%" if gt1 else None for e, gt1 in zip(evaluations, count_gt_1, strict=True)
+                f"{e.weight_percentage}%" if gt1 else None
+                for e, gt1 in zip(annotated_evaluations, count_gt_1, strict=True)
             )
             self.write_row(weight_percentages, lambda s: "evaluation_weight" if s is not None else "default")
 
             self.write_cell(_("Course Grade"), "bold")
-            for evaluation, gt1 in zip(evaluations, count_gt_1, strict=True):
+            for evaluation, gt1 in zip(annotated_evaluations, count_gt_1, strict=True):
                 if not gt1:
                     self.write_cell()
                     continue
@@ -246,7 +273,12 @@ class ResultsExporter(ExcelExporter):
             # Same reasoning as above.
             self.write_empty_row_with_styles(["default"] + ["border_top" if gt1 else "default" for gt1 in count_gt_1])
 
-    def write_questionnaire(self, questionnaire, evaluations_with_results, contributor):
+    def write_questionnaire(
+        self,
+        questionnaire: Questionnaire,
+        evaluations_with_results: list[tuple[Evaluation, OrderedDict[int, list[QuestionResult]]]],
+        contributor: UserProfile | None,
+    ) -> None:
         if contributor and questionnaire.type == Questionnaire.Type.CONTRIBUTOR:
             self.write_cell(f"{questionnaire.public_name} ({contributor.full_name})", "bold")
         else:
@@ -270,6 +302,7 @@ class ResultsExporter(ExcelExporter):
                 for grade_result in results[questionnaire.id]:
                     if grade_result.question.id != question.id or not RatingResult.has_answers(grade_result):
                         continue
+
                     values.append(grade_result.average * grade_result.count_sum)
                     count_sum += grade_result.count_sum
                     if grade_result.question.is_yes_no_question:
@@ -292,17 +325,17 @@ class ResultsExporter(ExcelExporter):
     # pylint: disable=arguments-differ
     def export_impl(
         self,
-        semesters,
-        selection_list,
-        include_not_enough_voters=False,
-        include_unpublished=False,
-        contributor=None,
-        verbose_heading=True,
+        semesters: QuerySetOrSequence[Semester],
+        selection_list: Sequence[tuple[Iterable[int], Iterable[int]]],
+        include_not_enough_voters: bool = False,
+        include_unpublished: bool = False,
+        contributor: UserProfile | None = None,
+        verbose_heading: bool = True,
     ):
         # We want to throw early here, since workbook.save() will throw an IndexError otherwise.
         assert len(selection_list) > 0
 
-        for sheet_counter, (degrees, course_types) in enumerate(selection_list, 1):
+        for sheet_counter, (degree_ids, course_type_ids) in enumerate(selection_list, 1):
             self.cur_sheet = self.workbook.add_sheet("Sheet " + str(sheet_counter))
             self.cur_row = 0
             self.cur_col = 0
@@ -314,14 +347,14 @@ class ResultsExporter(ExcelExporter):
             evaluations_with_results, used_questionnaires, course_results_exist = self.filter_evaluations(
                 semesters,
                 evaluation_states,
-                degrees,
-                course_types,
+                degree_ids,
+                course_type_ids,
                 contributor,
                 include_not_enough_voters,
             )
 
             self.write_headings_and_evaluation_info(
-                evaluations_with_results, semesters, contributor, degrees, course_types, verbose_heading
+                evaluations_with_results, semesters, contributor, degree_ids, course_type_ids, verbose_heading
             )
 
             for questionnaire in used_questionnaires:
--- a/evap/results/tools.py
+++ b/evap/results/tools.py
@@ -179,7 +179,7 @@ def get_single_result_rating_result(evaluation):
     return create_rating_result(question, answer_counters)
 
 
-def get_results_cache_key(evaluation):
+def get_results_cache_key(evaluation: Evaluation) -> str:
     return f"evap.staff.results.tools.get_results-{evaluation.id:d}"
 
 
@@ -189,7 +189,7 @@ def cache_results(evaluation, *, refetch_related_objects=True):
     caches["results"].set(cache_key, _get_results_impl(evaluation, refetch_related_objects=refetch_related_objects))
 
 
-def get_results(evaluation):
+def get_results(evaluation: Evaluation) -> EvaluationResult:
     assert evaluation.state in STATES_WITH_RESULTS_CACHING | {Evaluation.State.IN_EVALUATION}
 
     if evaluation.state == Evaluation.State.IN_EVALUATION:
@@ -197,7 +197,7 @@ def get_results(evaluation):
 
     cache_key = get_results_cache_key(evaluation)
     result = caches["results"].get(cache_key)
-    assert result is not None
+    assert isinstance(result, EvaluationResult)
     return result
 
 
@@ -210,7 +210,7 @@ GET_RESULTS_PREFETCH_LOOKUPS = [
 ]
 
 
-def _get_results_impl(evaluation: Evaluation, *, refetch_related_objects: bool = True):
+def _get_results_impl(evaluation: Evaluation, *, refetch_related_objects: bool = True) -> EvaluationResult:
     if refetch_related_objects:
         discard_cached_related_objects(evaluation)
 


evap/results/exporters.py
```diff
@@ -101,11 +106,18 @@ def filter_text_and_heading_questions(questions):
         return filtered_questions
 
     @staticmethod
-    def filter_evaluations(semesters, evaluation_states, degrees, course_types, contributor, include_not_enough_voters):
+    def filter_evaluations(
+        semesters: Iterable[Semester],
+        evaluation_states: Iterable[int],
```
richardebeling, line 111: `Iterable[Evaluation.State]`?

evap/results/exporters.py
```diff
@@ -101,11 +106,18 @@ def filter_text_and_heading_questions(questions):
...
+        degrees: Iterable[int],
+        course_types: CourseType,
```
richardebeling, line 113: Can you rename these to `degree_ids` and `course_type_ids` throughout the exporters?

richardebeling, line 113: This is wrong, course_types is an Iterable (the code here should work for `Union[CourseType, int]`, but it seems like we only ever pass `int`s)

evap/results/exporters.py
```diff
@@ -101,11 +106,18 @@ def filter_text_and_heading_questions(questions):
...
+        contributor: UserProfile | None,
```
richardebeling, line 114: Should be `Optional[UserProfile]`, right?

evap/results/exporters.py
```diff
@@ -122,46 +134,49 @@ def filter_evaluations(semesters, evaluation_states, degrees, course_types, cont
                 continue
             if not evaluation.can_publish_rating_results and not include_not_enough_voters:
                 continue
-            results = OrderedDict()
+            results: OrderedDict[EvaluationResult, list[RatingResult]] = OrderedDict()
             for contribution_result in get_results(evaluation).contribution_results:
                 for questionnaire_result in contribution_result.questionnaire_results:
                     # RatingQuestion.counts is a tuple of integers or None, if this tuple is all zero, we want to exclude it
+                    question_results: list[RatingResult] = questionnaire_result.question_results  # type: ignore[assignment]
```
richardebeling, line 141: Why is the narrowing to `list[RatingResult]` correct here instead of just `list[QuestionResult]`?

evap/results/exporters.py
```diff
@@ -122,46 +134,49 @@ def filter_evaluations(semesters, evaluation_states, degrees, course_types, cont
...
                         or contribution_result.contributor == contributor
                     ):
-                        results.setdefault(questionnaire_result.questionnaire.id, []).extend(
-                            questionnaire_result.question_results
-                        )
+                        results.setdefault(questionnaire_result.questionnaire.id, []).extend(question_results)
                         used_questionnaires.add(questionnaire_result.questionnaire)
-            evaluation.course_evaluations_count = evaluation.course.evaluations.count()
-            if evaluation.course_evaluations_count > 1:
+            evaluation.course_evaluations_count = evaluation.course.evaluations.count()  # type: ignore[attr-defined]
```
richardebeling, line 156: Can we declare `evaluation` to be `Any` once inside this function that does the monkey patching, instead of adding multiple `type ignore` comments?

evap/results/exporters.py
```diff
@@ -122,46 +134,49 @@ def filter_evaluations(semesters, evaluation_states, degrees, course_types, cont
...
-        return evaluations_with_results, used_questionnaires, course_results_exist
+        return evaluations_with_results, sorted(used_questionnaires), course_results_exist
 
     def write_headings_and_evaluation_info(
-        self, evaluations_with_results, semesters, contributor, degrees, course_types
-    ):
+        self,
+        evaluations_with_results: list[tuple[Evaluation, Any]],
+        semesters: QuerySet[Semester],
+        contributor: None | UserProfile,
```
richardebeling, line 176: Optional?

evap/results/exporters.py
```diff
@@ -122,46 +134,49 @@ def filter_evaluations(semesters, evaluation_states, degrees, course_types, cont
...
 
-        return evaluations_with_results, used_questionnaires, course_results_exist
+        return evaluations_with_results, sorted(used_questionnaires), course_results_exist
 
     def write_headings_and_evaluation_info(
-        self, evaluations_with_results, semesters, contributor, degrees, course_types
-    ):
+        self,
+        evaluations_with_results: list[tuple[Evaluation, Any]],
+        semesters: QuerySet[Semester],
```
richardebeling, line 175: I think we only require `Collection[Semester]` here?

evap/results/exporters.py
```diff
@@ -122,46 +134,49 @@ def filter_evaluations(semesters, evaluation_states, degrees, course_types, cont
...
+        contributor: None | UserProfile,
+        degrees: Iterable[int],
+        course_types: Iterable[int],
```
richardebeling, line 178: Again, can you rename to `degree_ids` and `course_type_ids`?

evap/results/exporters.py
```diff
@@ -195,7 +210,9 @@ def write_headings_and_evaluation_info(
         # One more cell is needed for the question column
         self.write_empty_row_with_styles(["default"] + ["border_left_right"] * len(evaluations_with_results))
 
-    def write_overall_results(self, evaluations_with_results, course_results_exist):
+    def write_overall_results(
+        self, evaluations_with_results: list[tuple[Evaluation, None]], course_results_exist: bool
```
richardebeling, line 214: Does `None` allow us here to pass something in that is not `None` there? I think it should be `Any` if we want to express that we don't care?

evap/results/exporters.py
```diff
@@ -215,15 +232,18 @@ def write_overall_results(self, evaluations_with_results, course_results_exist):
 
         if course_results_exist:
             # Only query the number of evaluations once and keep track of it here.
-            count_gt_1 = [e.course_evaluations_count > 1 for e in evaluations]
+            count_gt_1 = [e.course_evaluations_count > 1 for e in evaluations]  # type: ignore[attr-defined]
```
richardebeling, line 235: Again, can we annotate `evaluations` above once to make the code using the monkey-patched attributes work without additional type annotations?

evap/results/exporters.py
```diff
@@ -215,15 +232,18 @@ def write_overall_results(self, evaluations_with_results, course_results_exist):
...
 
             # Borders only if there is a course grade below. Offset by one column
             self.write_empty_row_with_styles(
                 ["default"] + ["border_left_right" if gt1 else "default" for gt1 in count_gt_1]
             )
 
             self.write_cell(_("Evaluation weight"), "bold")
-            weight_percentages = (f"{e.weight_percentage}%" if gt1 else None for e, gt1 in zip(evaluations, count_gt_1))
+            weight_percentages: Generator[str | None, None, None] = (
```
richardebeling, line 243: Same thing with the `None`s and `Optional`s as in earlier comments

evap/results/exporters.py
```diff
@@ -232,15 +252,20 @@ def write_overall_results(self, evaluations_with_results, course_results_exist):
...
 
             # Same reasoning as above.
             self.write_empty_row_with_styles(["default"] + ["border_top" if gt1 else "default" for gt1 in count_gt_1])
 
-    def write_questionnaire(self, questionnaire, evaluations_with_results, contributor):
+    def write_questionnaire(
+        self,
+        questionnaire: Questionnaire,
+        evaluations_with_results: list[tuple[Evaluation, Any]],
+        contributor: None | UserProfile,
```
richardebeling, line 267: Optional

evap/results/exporters.py
```diff
@@ -285,8 +310,13 @@ def write_questionnaire(self, questionnaire, evaluations_with_results, contribut
...
     # pylint: disable=arguments-differ
     def export_impl(
-        self, semesters, selection_list, include_not_enough_voters=False, include_unpublished=False, contributor=None
-    ):
+        self,
+        semesters: QuerySet[Semester],
+        selection_list,
+        include_not_enough_voters: bool = False,
+        include_unpublished: bool = False,
+        contributor: None | UserProfile = None,
```
richardebeling, line 318: Optional

niklasmohrin: Agreed, also applies in other places

fidoriel: https://peps.python.org/pep-0604/

richardebeling: My point was to swap the order here, so that `None` comes last since that's the order we usually use. If there's a way for us to also add a linting check for this, you can also add that

evap/results/exporters.py
fidoriel: We do get RatingResult by convention, and the Typechecker does not like the usage of a more broad Class.

evap/results/exporters.py
fidoriel: Mypy says `Collection[Semester]` is not iterabel. this seems a valid argument

evap/results/exporters.py
fidoriel: Yes we could do that. But I think some type ignores are better than `Any` to cover a little bit and think about what and why is monkeypatched.

evap/results/exporters.py
fidoriel: I dont have any idea

evap/results/exporters.py
fidoriel: Yeah, nice try, but `Evaluation.State.X` is an integer and so we cannot type this as a list because it looses all context information.

evap/results/exporters.py
richardebeling: > We do get RatingResult by convention
I don't see why that is true. Other result types were not excluded in this method up to this point. The variable is called `question_result`. The very next statement (line 143) goes through all question results and checks whether they are coming from a rating question (-> they are a `RatingResult`) or not. In the end, all results are returned, possibly including other result types.

Where do you see the convention that these are all `RatingResult`s?

evap/results/exporters.py
richardebeling: In that case, we should make `Evaluation.State` an `IntEnum` in the future ([the fsm field supports that](https://github.com/viewflow/django-fsm/issues/204), doesn't require a migration, but juts quickly patching it breaks some UI, e.g. the staff semester index view). For now, I'd still like to have something else than `int` here because the type very clearly is not `int`. Passing `1337` is never correct. We can use `Any` and add a follow-up issue to make it an `IntEnum` and use the correct type here. We can also use some placeholder type for now, which can be equal to `int` or `Any`, but add semantical clearness with a different name (`StateT` or whatever)

@niklasmohrin thoughts?

evap/results/exporters.py
richardebeling: To me, `Any` says that this variable currently isn't typechecked, so it tells you that you can not rely on typechecking. Annotating as `Evaluation`, on the other hand, would imply that it is treated according to static typing rules and typechecked, which is a lie here.

To make this code typecheck in the future
* with `Any`, we'd have to change one place (replace `Any` with a type that also allows the monkeypatched attributes)
* with lots of `type: ignore`, we have to go through lots of code, and find out for each annotation again whether it was due to a bug/failure in mypy, problem with the stubs, or because we lied to the typechecker.

I have a strong opinion here. The code doesn't treat `evaluation` according to static type checking rules for `Evaluation`, so we shouldn't make the type checker assume that it does and then go on and ignore all violations.

@niklasmohrin opinions?

evap/results/exporters.py
richardebeling: You resolved the discussion item but you didn't respond or fix the code.

Surely, the API here isn't that users have to pass in a list in the form of `[(evaluation1, None), (evaluation2, None), ...]`, but instead, it seems to me that you simply don't care about the second tuple element in each tuple. In that case, it seems to me that `None` is a wrong annotation and `Any` should be used here instead.

evap/results/exporters.py
```diff
@@ -215,15 +231,18 @@ def write_overall_results(self, evaluations_with_results, course_results_exist):
...
+            count_gt_1 = [e.course_evaluations_count > 1 for e in evaluations]  # type: ignore[attr-defined]
 
             # Borders only if there is a course grade below. Offset by one column
             self.write_empty_row_with_styles(
                 ["default"] + ["border_left_right" if gt1 else "default" for gt1 in count_gt_1]
             )
 
             self.write_cell(_("Evaluation weight"), "bold")
-            weight_percentages = (f"{e.weight_percentage}%" if gt1 else None for e, gt1 in zip(evaluations, count_gt_1))
+            weight_percentages: Generator[str | Any, Any, Any] = (
```
richardebeling, line 242: Should be `Generator[str | None, None, None]` or `Iterable[str | None]` (see [the docs](https://docs.python.org/3/library/typing.html#typing.Generator))

evap/results/exporters.py
fidoriel: Decided to use `Any`

evap/results/exporters.py
```diff
@@ -101,16 +107,23 @@ def filter_text_and_heading_questions(questions):
         return filtered_questions
 
     @staticmethod
-    def filter_evaluations(semesters, evaluation_states, degrees, course_types, contributor, include_not_enough_voters):
+    def filter_evaluations(
+        semesters: Iterable[Semester],
+        evaluation_states: Any,
+        degree_ids: Iterable[int],
+        course_type_ids: CourseType,
```
richardebeling, line 114: `course_type_ids` is plural, `CourseType` is singular. Should this be `Iterable[int]` or `Iterable[CourseType]`?

evap/results/exporters.py
```diff
@@ -122,53 +135,58 @@ def filter_evaluations(semesters, evaluation_states, degrees, course_types, cont
...
                     if all(
                         not question_result.question.is_rating_question
-                        or question_result.counts is None
-                        or sum(question_result.counts) == 0
-                        for question_result in questionnaire_result.question_results
+                        or (
+                            isinstance(question_result, RatingResult)
+                            and (question_result.counts is None or sum(question_result.counts) == 0)
+                        )
+                        for question_result in question_results
```
richardebeling, line 149: When we were talking about this, I had something like this in my mind:
```python3
if all(
    not isinstance(question_result, RatingResult)
    or question_results.counts is None
    or sum (question_result.counts) == 0
    for question_result in question_results
):
```

evap/results/exporters.py
```diff
@@ -232,15 +255,20 @@ def write_overall_results(self, evaluations_with_results, course_results_exist):
...
             self.next_row()
 
             # Same reasoning as above.
             self.write_empty_row_with_styles(["default"] + ["border_top" if gt1 else "default" for gt1 in count_gt_1])
 
-    def write_questionnaire(self, questionnaire, evaluations_with_results, contributor):
+    def write_questionnaire(
+        self,
+        questionnaire: Questionnaire,
+        evaluations_with_results: list[tuple[Evaluation, Any]],
```
richardebeling, line 269: The code below looks to me as if `Any` here should be `dict[int, GradeResult]`. Is that correct? If yes, I'd rather that annotated.

evap/results/exporters.py
richardebeling: (The comment was resolved but the code was not changed. I unresolved.)

evap/results/exporters.py
niklasmohrin: Yeah, let's use `Any`

evap/results/exporters.py
richardebeling: (Note: `isinstance(question_result, RatingResult)` is 'wider' than the old `question_result.question.is_rating_question` since we can now have text answers to rating questions. I believe that my suggestion here is what the intention behind this code was, but this might be wrong)

richardebeling: Not happy with the current change here since it modifies behavior. Before, if we had a text answer to a rating question, this code crashed since `question_result.counts` undefined. Now, it would simply pass and include that result.

From a quick look at `_get_results_impl`, it looks to me as if additional text answers to rating questions are always included in a rating result. If those never existed as a TextAnswer, then the "text answer to rating question" situation shouldn't be a problem here. This would explain why we never observed a crash here.

I currently believe that the `not question_result.question.is_rating_question` part was initially intended to filter out all non-rating-results and is from before we had the `RatingResult` class, so that's why it doesn't use `isinstance`.

evap/results/exporters.py
```diff
@@ -262,12 +291,22 @@ def write_questionnaire(self, questionnaire, evaluations_with_results, contribut
...
+                    # `filter_text_and_heading_questions` filters all text questions
+                    # `grade_result.question.id != question.id` filters the remaining HeadingResults because `question.is_heading_question` filters all heading questions
+                    # so here all `grade_result` from the OrderedDict values are only RatingResults with the same question as the current question which is not a heading question or a text question so it must be a rating question
+                    rating_result = cast(RatingResult, grade_result)
+                    assert rating_result.count_sum
+                    assert rating_result.average
+                    values.append(rating_result.average * rating_result.count_sum)
+                    count_sum += rating_result.count_sum
+                    if rating_result.question.is_yes_no_question:
+                        assert rating_result.approval_count
```
richardebeling, line 308: I don't think these asserts do what you intended them to do. I think you wanted to do this:
```suggestion
                    assert hasattr(rating_result, "count_sum")
                    assert hasattr(rating_result, "average")
                    values.append(rating_result.average * rating_result.count_sum)
                    count_sum += rating_result.count_sum
                    if rating_result.question.is_yes_no_question:
                        assert hasattr(rating_result, "approval_count")
```

evap/results/exporters.py
```diff
@@ -262,12 +291,22 @@ def write_questionnaire(self, questionnaire, evaluations_with_results, contribut
                 approval_count = 0
 
                 for grade_result in results[questionnaire.id]:
-                    if grade_result.question.id != question.id or not grade_result.has_answers:
+                    if grade_result.question.id != question.id or (
+                        isinstance(grade_result, RatingResult) and not grade_result.has_answers
+                    ):
```
richardebeling, line 296: Again, not happy: This modifies behavior and thus widens the contract of this snippet here. A for "add type annotations" PR, I'd like to have no functional changes.
```suggestion
                    assert isinstance(grade_result, RatingResult)
                    if grade_result.question.id != question.id or not grade_result.has_answers:
```

richardebeling (changes requested):

fidoriel:
typechecker fails => this is fixed in #2064

evap/results/exporters.py
```diff
@@ -122,53 +133,56 @@ def filter_evaluations(semesters, evaluation_states, degrees, course_types, cont
...
                 for questionnaire_result in contribution_result.questionnaire_results:
                     # RatingQuestion.counts is a tuple of integers or None, if this tuple is all zero, we want to exclude it
+                    question_results: list[QuestionResult] = questionnaire_result.question_results
                     if all(
-                        not question_result.question.is_rating_question
-                        or question_result.counts is None
-                        or sum(question_result.counts) == 0
-                        for question_result in questionnaire_result.question_results
+                        not isinstance(question_result, RatingResult)
+                        or (question_result.counts is None or sum(question_result.counts) == 0)
```
richardebeling, line 143: `or` should evaluate left to right here, so the old code without parenthesis should still be fine, right? In that case, I'd prefer that. With the parenthesis, the last `or sum(question_result.counts) == 0` is more hidden and might be overlooked by readers.

evap/results/exporters.py
```diff
@@ -122,53 +133,56 @@ def filter_evaluations(semesters, evaluation_states, degrees, course_types, cont
...
                     ):
-                        results.setdefault(questionnaire_result.questionnaire.id, []).extend(
-                            questionnaire_result.question_results
-                        )
+                        results.setdefault(questionnaire_result.questionnaire.id, []).extend(question_results)
                         used_questionnaires.add(questionnaire_result.questionnaire)
-            evaluation.course_evaluations_count = evaluation.course.evaluations.count()
-            if evaluation.course_evaluations_count > 1:
+            monkeypatched_evaluation: Any = evaluation
+            monkeypatched_evaluation.course_evaluations_count = monkeypatched_evaluation.course.evaluations.count()  # type: ignore[attr-defined]
```
richardebeling, line 155: My idea behind having a variable cast as `Any` was to allow now having the `  # type: ignore[attr-defined]` comments -- these aren't necessary anymore now, are they?

evap/results/exporters.py
```diff
@@ -122,53 +133,56 @@ def filter_evaluations(semesters, evaluation_states, degrees, course_types, cont
...
                         or contribution_result.contributor == contributor
                     ):
-                        results.setdefault(questionnaire_result.questionnaire.id, []).extend(
-                            questionnaire_result.question_results
-                        )
+                        results.setdefault(questionnaire_result.questionnaire.id, []).extend(question_results)
                         used_questionnaires.add(questionnaire_result.questionnaire)
-            evaluation.course_evaluations_count = evaluation.course.evaluations.count()
-            if evaluation.course_evaluations_count > 1:
+            monkeypatched_evaluation: Any = evaluation
```
richardebeling, line 154: ["Monkeypatched"](https://en.wikipedia.org/wiki/Monkey_patch) probably isn't the most precise word for what we're doing here. What do you think about `evaluation_as_any`?

evap/results/exporters.py
```diff
@@ -122,53 +133,56 @@ def filter_evaluations(semesters, evaluation_states, degrees, course_types, cont
...
-            evaluations_with_results.append((evaluation, results))
+            evaluations_with_results.append((monkeypatched_evaluation, results))
 
         evaluations_with_results.sort(
             key=lambda cr: (cr[0].course.semester.id, cr[0].course.type.order, cr[0].full_name)
         )
-        used_questionnaires = sorted(used_questionnaires)
 
-        return evaluations_with_results, used_questionnaires, course_results_exist
+        return evaluations_with_results, sorted(used_questionnaires), course_results_exist
```
richardebeling, line 169: Not related to typing. I liked the old code a bit more. The `sorted` here can easily be overlooked, and before, we had a single block that took care of sorting, now the sorting is more distributed / entagled in the surrounding code.

evap/results/exporters.py
```diff
@@ -215,15 +233,18 @@ def write_overall_results(self, evaluations_with_results, course_results_exist):
 
         if course_results_exist:
             # Only query the number of evaluations once and keep track of it here.
-            count_gt_1 = [e.course_evaluations_count > 1 for e in evaluations]
+            count_gt_1 = [e.course_evaluations_count > 1 for e in evaluations]  # type: ignore[attr-defined]
```
richardebeling, line 236: cast-to-any logic applies here (and in all other places that use `# type: ignore[attr-defined]`) as well.

evap/results/exporters.py
```diff
@@ -262,12 +288,24 @@ def write_questionnaire(self, questionnaire, evaluations_with_results, contribut
...
-                    count_sum += grade_result.count_sum
-                    if grade_result.question.is_yes_no_question:
-                        approval_count += grade_result.approval_count
+
+                    # `filter_text_and_heading_questions` filters all text questions
+                    # `grade_result.question.id != question.id` filters the remaining HeadingResults because `question.is_heading_question` filters all heading questions
+                    # so here all `grade_result` from the OrderedDict values are only RatingResults with the same question as the current question which is not a heading question or a text question so it must be a rating question
+                    rating_result = cast(RatingResult, grade_result)
+                    assert hasattr(rating_result, "count_sum")
+                    assert hasattr(rating_result, "average")
```
richardebeling, line 303: We have `assert isinstance(grade_result, RatingResult)` four lines above in 294, I don't understand the intention here. I think in the best case this PR doesn't have to touch any lines after the `continue` in 296.

evap/results/exporters.py
```diff
@@ -285,12 +323,17 @@ def write_questionnaire(self, questionnaire, evaluations_with_results, contribut
 
     # pylint: disable=arguments-differ
     def export_impl(
-        self, semesters, selection_list, include_not_enough_voters=False, include_unpublished=False, contributor=None
-    ):
+        self,
+        semesters: Sequence[Semester],
+        selection_list,
```
richardebeling, line 328: Seems to be a sequence of `(degree_ids, course_type_ids)` pairs in the code below -> `Sequence[tuple[Iterable[int], Iterable[int]]]`?

evap/results/exporters.py
```diff
@@ -285,12 +323,17 @@ def write_questionnaire(self, questionnaire, evaluations_with_results, contribut
...
+        include_not_enough_voters: bool = False,
+        include_unpublished: bool = False,
+        contributor: UserProfile | None = None,
+    ) -> None:
         # We want to throw early here, since workbook.save() will throw an IndexError otherwise.
         assert len(selection_list) > 0
 
-        for sheet_counter, (degrees, course_types) in enumerate(selection_list, 1):
+        for sheet_counter, (degrees, course_type_ids) in enumerate(selection_list, 1):
```
richardebeling, line 336: Both seem to be `_ids` semantically from the usage below, at least if we annotated `filter_evaluations` and `write_headings_and_evaluation_info` below correctly.

richardebeling (changes requested):

evap/results/exporters.py
fidoriel: Type checker complains about redefinition

niklasmohrin: I think that doesn't void the argument ;)

richardebeling: Suggestion that should satisfy the type checker and us:
```
# [...] other sorting line here
sorted_used_questionnaires = sorted(used_questionnaires)

return evaluations_with_results, sorted_used_questionnaires, course_results_exist
```

evap/results/exporters.py
fidoriel: if grade_result.question.id != question.id: and if grade_result.question.is_yes_no_question: implicitly enforce the type so we assert to tell the typechecker

evap/results/exporters.py
```diff
@@ -262,11 +290,17 @@ def write_questionnaire(self, questionnaire, evaluations_with_results, contribut
...
+                        continue
+
+                    assert isinstance(grade_result, RatingResult)
+                    if not grade_result.has_answers:
                         continue
+
                     values.append(grade_result.average * grade_result.count_sum)
                     count_sum += grade_result.count_sum
                     if grade_result.question.is_yes_no_question:
+                        assert hasattr(grade_result, "approval_count")
```
richardebeling, line 303: this is already covered by `assert isinstance(grade_result, RatingResult)` above (same as https://github.com/e-valuation/EvaP/pull/2035#discussion_r1400950031)
```suggestion
```

richardebeling (dismissed):
thanks :)

evap/evaluation/tools.py
```diff
@@ -276,7 +277,11 @@ def next_row(self) -> None:
         self.cur_col = 0
         self.cur_row += 1
 
-    def write_row(self, vals: Iterable[str], style: str = "default") -> None:
+    def write_row(
+        self,
+        vals: Iterable[OptionalStrOrOptionalFloat],
```
niklasmohrin, line 282: This has already been typed in #2073 (and I prefer the version from that). Could you rebase your branch to incorporate the changes?

niklasmohrin (changes requested):

niklasmohrin:
This seems to be running into the same stuff that is solved in #2091 now, I would say we wait until that is done?

evap/results/exporters.py
```diff
@@ -123,52 +132,58 @@ def filter_evaluations(semesters, evaluation_states, degrees, course_types, cont
...
+        contributor: UserProfile | None,
+        degree_ids: Iterable[int],
+        course_type_ids: Iterable[int],
+        verbose_heading: bool,
+    ) -> None:
         export_name = _("Evaluation")
         if contributor:
             export_name += f"\n{contributor.full_name}"
         elif len(semesters) == 1:
             export_name += f"\n{semesters[0].name}"
```
niklasmohrin, line 183: If you do
```suggestion
            export_name += f"\n{next(iter(semesters)).name}"
```
you can type `semesters` as `Collection[Semester]`

niklasmohrin (changes requested):
Please make sure to also review your code yourself before requesting the a review, you yourself are most familiar with the code and can sort out low hanging fruit most quickly

evap/results/tools.py
```diff
@@ -179,7 +179,7 @@ def cache_results(evaluation, *, refetch_related_objects=True):
     caches["results"].set(cache_key, _get_results_impl(evaluation, refetch_related_objects=refetch_related_objects))
 
 
-def get_results(evaluation):
+def get_results(evaluation: Evaluation) -> EvaluationResult:
```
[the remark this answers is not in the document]
fidoriel: Because we get it from cache and only put the correct type in the cache.

evap/results/exporters.py
fidoriel: To sum this up: Django is not consistent with `QuerySet`. `QuerySet` implements `contains` but not `__contains__`. This cannot be annotated as a Container because `__contains__` is missing. So using a common base type is not possible. This could be typed as an `Sequence` (with this suggestion), but then typeguard complains about not having the same type, which is obviously true. MyPy does not complain because everything used is implemented and Django Stubs implements `__contains__`. It is possible to be a Container that does not provide `__contains__` this trigger's fallback, so `__iter__` with comparisons is used. Because `__iter__` is defined, we could use `Iterable` but we deemed this as too generic.

fidoriel: Please see https://github.com/typeddjango/django-stubs/issues/1924 for the QuerySet Typing Issue

fidoriel: The correct type is a class which inherits `Sized` and `Iterable`

evap/results/exporters.py
```diff
@@ -170,7 +173,10 @@ def filter_evaluations(
     def write_headings_and_evaluation_info(
         self,
         evaluations_with_results: list[tuple[Evaluation, OrderedDict[int, list[QuestionResult]]]],
-        semesters: Sequence[Semester] | QuerySet[Semester],
+        semesters: Sequence[Semester]
+        | QuerySet[
+            Semester
+        ],  # see https://github.com/typeddjango/django-stubs/issues/1924 why this is not a Collection
```
richardebeling, line 179: 1. I'd rather have your django issue linked here (https://code.djangoproject.com/ticket/35154), as the django-stubs issue is not about the core problem (they only document what django does/has).
2. I'd prefer to not have this comment (and those 4 "ugly" lines in general) copied around the codebase. I'd suggest we add a new generic helper type (`QuerySetLike` or `QuerySetOrContainer` or whatever) and use it whenever we want to express "QuerySet[x] or Container[x]". The comment would then go with this type definition.

richardebeling (changes requested):

evap/results/exporters.py
```diff
@@ -103,16 +110,23 @@ def filter_text_and_heading_questions(questions):
         return filtered_questions
 
     @staticmethod
-    def filter_evaluations(semesters, evaluation_states, degrees, course_types, contributor, include_not_enough_voters):
+    def filter_evaluations(
+        semesters: Iterable[Semester],
+        evaluation_states: Iterable[Any],
```
niklasmohrin, line 115:
```suggestion
        evaluation_states: Iterable[int],
```

evap/results/exporters.py
```diff
@@ -124,52 +138,58 @@ def filter_evaluations(semesters, evaluation_states, degrees, course_types, cont
...
-            results = OrderedDict()
+            results: OrderedDict[int, list[QuestionResult]] = OrderedDict()
             for contribution_result in get_results(evaluation).contribution_results:
                 for questionnaire_result in contribution_result.questionnaire_results:
                     # RatingQuestion.counts is a tuple of integers or None, if this tuple is all zero, we want to exclude it
+                    question_results: list[QuestionResult] = questionnaire_result.question_results
                     if all(
-                        not question_result.question.is_rating_question or not RatingResult.has_answers(question_result)
-                        for question_result in questionnaire_result.question_results
+                        not isinstance(question_result, RatingResult) or not RatingResult.has_answers(question_result)
```
niklasmohrin, line 147: Shouldn't this be the same now?
```suggestion
                        not isinstance(question_result, AnsweredRatingResult)
```
If so, I think `not any(isinstance(...))` would be nicer to read now :)

evap/results/exporters.py
```diff
@@ -199,7 +219,11 @@ def write_headings_and_evaluation_info(
         # One more cell is needed for the question column
         self.write_empty_row_with_styles(["default"] + ["border_left_right"] * len(evaluations_with_results))
 
-    def write_overall_results(self, evaluations_with_results, course_results_exist):
+    def write_overall_results(
+        self,
+        evaluations_with_results: list[tuple[Evaluation, OrderedDict[int, list[QuestionResult]]]],
```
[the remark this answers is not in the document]
niklasmohrin: Strong agree, the current version is not typed correctly. I would also prefer `annotated_evaluation` over `evaluation_as_any` as a naming scheme everywhere, because the former is what happens in the code, while the latter only happens in the typing and is only incidental complexity

Whether or not we end up with `Protocol` in the end, having `AnnotatedEvaluation` already typed out seems like a good idea

niklasmohrin:
Apart from the open discussions, this looks good to me :)

richardebeling (approved):

niklasmohrin:
I unresolved some comments that still seem unanswered to me, could you give a quick response to them?

evap/results/exporters.py
fidoriel: did include it

evap/results/exporters.py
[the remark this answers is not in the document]
fidoriel: annotated as any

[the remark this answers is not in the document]
niklasmohrin: Yes, I saw that. What I meant was my suggestion to rename `evaluation_as_any` to `annotated_evaluation` - or do you have arguments against that?

The line `evaluation_as_any: Any = evaluation` also still does not use `AnnotatedEvaluation`

evap/results/exporters.py
niklasmohrin: Again, I saw that you committed the initial suggestion, but you did not respond to my suggestion of using `not any` instead. I still think "none of the them are answered" is nicer than "all of them are un-answered"

I wouldn't have wrote a comment asking you to re-check the discussions if there wouldn't be something still left in those :)

evap/results/exporters.py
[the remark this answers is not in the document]
fidoriel: Changed that.

evap/results/exporters.py
niklasmohrin: After #2174, this should now be possible :)
```suggestion
        evaluation_states: Iterable[Evaluation.State],
```

niklasmohrin (commented):
Almost there :) Please remember to perform self-reviews, it's usually worth it (that is also my experience with code I write myself)

niklasmohrin (approved):
Perfection!

richardebeling (approved):
Thanks. There's still some minor issues left (`T` should probably be `ModelT`, `degrees` vs `degree_ids`, `course_types` vs `course_types_ids`, a `-> None` is missing, I think one `: list[bool]` annotation is superfluous), but I think we can already merge this.

PR #2036, opened 2023-10-16
Improve translation handling in JavaScript and TypeScript

### LINKED ISSUE #2027: Translations in Javascript and Typescript
When writing Javascript and Typescript in separate, non-HTML files, we can't use the Django template functions `trans`, `blocktrans`, etc. anymore. We have worked around this by putting translated strings into the DOM and accessing them via Javascript then.

Instead of doing this, we want to have a unified approach where the use-site can just write `trans("The server is not responding.")` or so. There are two possible approaches:

1. DIY: We have a function `trans(english: string, to: Language = window.LANGUAGE): string` with `type Language = "English" | "German"`. This function looks up the string in a global dictionary (for example `window.translationDictionary` or so). I am not sure what it should do if the string is not present, probably return the English string and emit a warning? This dictionary would be defined in a script tag in a HTML file, something like (possibly with an implementation that doesn't repeat the strings a little less):
```html
    window.translationDictionary = {
        "de": {
            {% language 'de' %}
                "The server is not responding": "{% trans 'The server is not responding' %}",
            {% endlanguage %}
        }
    };
</script>
```
2. Use Django's builtin functionality: There is a builtin way that configures an extra endpoint to make all translations available (https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#internationalization-in-javascript-code). A plus is that it also supports `ngettext` and so on. It seems like it can also detect all strings used in translations, but the setup may be a bit tricky with Typescript thrown into the mix.

I think I prefer the first approach, but maybe we encounter difficulties with it or decide that we will need `ngettext` etc. in the future and go with the Django versions directly.

hansegucker:
Closes #2027


niklasmohrin:
For Typescript and modules:
1. It seems that using a global function from a module just works, so we can keep using the import as it currently is
2. To inform Typescript about the existence of these functions, a line like the following can be added and imported in all files using it:
```typescript
export declare function gettext(s: string): string;
```
I would suggest that these declarations live either in `utils.ts` or in a separate file like `translation.ts`, slightly preferring the latter.

hansegucker:
@niklasmohrin Ready for review

evap/evaluation/templates/sortable_form_js.html
niklasmohrin, line 1: I think you forgot to add the new `sortable_form.js` to the commit :)

evap/locale/de/LC_MESSAGES/djangojs.po
```diff
@@ -0,0 +1,71 @@
...
+msgid "Delete"
+msgstr ""
+
+#: evap/static/js/sortable_form.js:20
+msgid "add another"
+msgstr ""
+
+#: evap/static/js/tom-select.complete.min.js:132
+msgid "<div>"
+msgstr ""
```
niklasmohrin, line 35: It seems that tomselect uses something like `_(...)` somewhere, we should probably ignore our external library files in the `makemessage` call

evap/static/ts/src/translation.ts
```diff
@@ -0,0 +1,13 @@
+export {};
+declare global {
```
niklasmohrin, line 2: Let's add a short comment here explaining where these come from :)

evap/urls.py
```diff
@@ -13,6 +14,8 @@
 
     path("logout", django.contrib.auth.views.LogoutView.as_view(next_page="/"), name="django-auth-logout"),
     path("oidc/", include('mozilla_django_oidc.urls')),
+
+    path("i18n.js", JavaScriptCatalog.as_view(), name="javascript-catalog"),
```
niklasmohrin, line 18: I would prefer if this would be called something with "catalog" in it, so that it is consistent with the other names. Is there anything that?

niklasmohrin (commented):
Very cool!

evap/locale/de/LC_MESSAGES/djangojs.po
hansegucker: I am not sure which rule would apply to ignore only the external libraries and not our own files.

evap/locale/de/LC_MESSAGES/djangojs.po
```diff
@@ -0,0 +1,71 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+#
```
richardebeling, line 5: weird default copyright header -- I'd just replace it with the header from `django.po`?


```suggestion
# EvaP translation
# This file is distributed under the same license as the EvaP project.
#
```

evap/locale/de/LC_MESSAGES/djangojs.po
```diff
@@ -0,0 +1,71 @@
...
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2023-10-16 20:14+0200\n"
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: LANGUAGE <LL@li.org>\n"
+"Language: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
```
richardebeling, line 19: Also taking in the fields from the existing django.po, this would be:

```suggestion
"Project-Id-Version: EvaP\n"
"Report-Msgid-Bugs-To: \n"
"POT-Creation-Date: 2023-10-16 20:14+0200\n"
"PO-Revision-Date: 2023-10-16 20:14+0200\n"
"Last-Translator: Johannes Wolf <janno42@posteo.de>\n"
"Language-Team: Johannes Wolf (janno42)\n"
"Language: de\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=UTF-8\n"
"Content-Transfer-Encoding: 8bit\n"
"Plural-Forms: nplurals=2; plural=(n != 1);\n"
"X-Generator: Poedit 3.3.2\n"
```

richardebeling (approved):
Looking good, thanks :)

evap/locale/de/LC_MESSAGES/djangojs.po
richardebeling: From the documentation, I think adding something like [`--ignore=evap/static/js/*.min.js`](https://docs.djangoproject.com/en/4.2/ref/django-admin/#cmdoption-makemessages-ignore) to the [`makemessages` command line](https://github.com/e-valuation/EvaP/pull/2036/files#diff-80bbdb2f9832c97a33b4cebdad86c8bc7d2e3463f4db47b2bde21067b38f6154R12) should already do the trick.

evap/locale/de/LC_MESSAGES/django.po
richardebeling, line 1: I think all the changes here are created by running `makemessages`, but the changes here are not relevant to the javascript translations, right? In this case, could you revert changes to this file?

evap/static/js/.gitignore
```diff
@@ -1,3 +1,4 @@
 # ignore the typescript output, but still track the libraries
 *.js
 !*.min.js
+!sortable_form.js
```
richardebeling, line 4: missing newline at the end of file

evap/static/js/locale/moment_de.js
richardebeling, line 1: You removed three javascript files here -- is that intentional?

richardebeling (changes requested):

hansegucker:
@richardebeling

richardebeling (approved):

niklasmohrin (approved):
Thanks :)

niklasmohrin (approved):

richardebeling:
Thanks :)

PR #2037, opened 2023-10-16
fix arrow alignment

### LINKED ISSUE #2033: Ordering icons misplaced in table headers
On the grades semester page and the staff semester page, the order icons (arrows) overlap the heading text as shown below because `padding-left` defined by `.col-order` is overwritten by Bootstrap's `.table`.
image

Instead, they should be placed next to the text, as on the results page.
image

jolanda19B:
fix #2033

### DIFF
--- a/evap/static/scss/components/_grid.scss
+++ b/evap/static/scss/components/_grid.scss
@@ -130,7 +130,7 @@
 
 .col-order {
     position: relative;
-    padding-left: 1em;
+    padding-left: 1.4em;
     cursor: pointer;
 
     &::before, &::after {
@@ -155,7 +155,7 @@
 }
 
 th.col-order {
-    padding-left: 1.4em;
+    padding-left: 1.4em !important;
 
     &:empty {
         padding-right: 0;


evap/static/scss/components/_grid.scss
```diff
@@ -155,7 +155,7 @@
 }
 
 th.col-order {
-    padding-left: 1.4em;
+    padding-left: 1.4em !important;
```
niklasmohrin, line 158: @richardebeling This is now the same as in the rule for just `.col-order`, so we should be able to just remove this rule altogether, right? I am not sure why exactly we had explicitly different values for the `th` and non-`th` rules, maybe @janno42 can remember?

janno42: we probably needed `th.col-order` to be more specific than something else to avoid using `!important`. removing this line would require us to put `!important` behind the rule in `.col-order`.
i would prefer a solution without `!important` by adding the other classes of the object to the selector for higher specificity (https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity)

janno42:
@jolanda19B Do you want to continue working on this or should we take over?

PR #2040, opened 2023-10-16
Replace toggle_no_grades with an idempotent alternative

### LINKED ISSUE #1965: Replace `toggle_no_grades` with an idempotent alternative
Currently, there is a URL for toggling grade status of a course (`grades:toggle_no_grades`). Instead of having a toggle, we would like an idempotent operation, that is, the endpoint should be something like `grades:set_gets_grade_documents` where the desired boolean value is sent with the request. This way, two people who want to change the status don't revert each others action but instead set the same value.

till2:
Fixes #1965

### DIFF
--- a/evap/grades/templates/grades_semester_view.html
+++ b/evap/grades/templates/grades_semester_view.html
@@ -108,8 +108,12 @@
     {% include 'confirmation_modal.html' with modal_id='confirmNouploadModal' title=title question=question action_text=action_text btn_type='primary' %}
     <script type="text/javascript">
         function confirmNouploadModalAction(dataId) {
-            fetch("{% url 'grades:toggle_no_grades' %}", {
-                body: new URLSearchParams({course_id: dataId}),
+            confirmUploadModalAction(dataId, true);
+        }
+
+        function confirmUploadModalAction(dataId, status) {
+            fetch("{% url 'grades:set_no_grades'%}", {
+                body: new URLSearchParams({course_id: dataId, status: `${Number(status)}` }),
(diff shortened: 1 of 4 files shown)


evap/grades/tests.py
```diff
@@ -135,7 +135,7 @@ def test_upload_final_grades(self):
         evaluation.save()
         self.helper_check_final_grade_upload(course, 0)
 
-    def test_toggle_no_grades(self):
```
niklasmohrin, line 138: Let's also check that sending the same request again doesn't change anything by adding another request at the end of the test

evap/grades/templates/grades_semester_view.html
```diff
@@ -108,8 +108,12 @@ <h3 class="col-8 mb-0">
     {% include 'confirmation_modal.html' with modal_id='confirmNouploadModal' title=title question=question action_text=action_text btn_type='primary' %}
     <script type="text/javascript">
         function confirmNouploadModalAction(dataId) {
-            fetch("{% url 'grades:toggle_no_grades' %}", {
-                body: new URLSearchParams({course_id: dataId}),
+            confirmUploadModalAction(dataId, true);
+        }
+
+        function confirmUploadModalAction(dataId, status) {
```
niklasmohrin, line 114: How do you feel about renaming this to something like `updateGradeStatusOfCourse(courseId, status)`?

evap/grades/tests.py
LinqLover: Done, please close

PR #2042, opened 2023-10-22
Fix typecheck failure introduced with `django-stubs==4.2.5`

richardebeling:
With their 4.2.5 release (coming from 4.2.4), django-stubs did two things that broke type checking for us

* They improved their typing on ManyToManyFields, but didn't add functionality to correctly resolve lazy model names (`"UserProfile`) without an app label (`"evaluation.UserProfile"` works).
* They added `AbstractBaseUser.REQUIRED_FIELDS` as a `ClassVar`, we had it defined as an instance variable.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,8 +2,8 @@
 black~=23.9.1
 coverage[toml]
 django-debug-toolbar~=4.0
-django-stubs~=4.2.4
-django-stubs-ext~=4.2.0
+django-stubs~=4.2.5
+django-stubs-ext~=4.2.5
 django-webtest~=1.9.10
 isort~=5.12.0
 model-bakery~=1.16.0
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,8 +2,8 @@
 black~=23.9.1
 coverage[toml]
 django-debug-toolbar~=4.0
-django-stubs~=4.2.5
-django-stubs-ext~=4.2.5
+django-stubs==4.2.5
+django-stubs-ext==4.2.5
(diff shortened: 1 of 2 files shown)


requirements-dev.txt
```diff
@@ -2,8 +2,8 @@
 black~=23.9.1
 coverage[toml]
 django-debug-toolbar~=4.0
-django-stubs~=4.2.4
-django-stubs-ext~=4.2.0
+django-stubs==4.2.5
+django-stubs-ext==4.2.5
```
richardebeling, line 6: @niklasmohrin since they follow the django versioning scheme, they don't really have semantic versioning, i.e. we can expect that future `patch` changes will break our CI again. I'm not a big fan of that, so I'd propose fixing `django-stubs` down to its patch version in `requirements-dev.txt`.

However, this will bring two annoyances:
* More depenabot PRs. We might want to give it a higher cooldown for these?
* We need to keep these two numbers here in sync. Even more annoying since I think dependabot will not do this. In pip's reference for requirements files, I don't see a way to specify that we want the same versions for `django-stubs` and `django-stubs-ext`. Should we suggest to them to allow installing `django-stubs-ext` as an extra -> `django-stubs[ext]==4.2.5`? Currently, they only have the [`compatible-mypy` extra](https://github.com/typeddjango/django-stubs/blob/master/setup.py#L34).

niklasmohrin: Sure, if they break with patch, we will have to deal with all of their breaking releases separately :+1:

niklasmohrin (approved):

PR #2045, opened 2023-10-23
Fix #2002 remove degrees and course types from contributor export header

### LINKED ISSUE #2002: Remove degrees and course types from contributor export header
In the contributor results export file, the header includes "Evaluation", the contributor's name, and all existing degrees and course types.
Instead, only "Evaluation" and the name should be exported.

Exports other than the contributor export should not be modified.

Tillbtn:
closes #2002

### DIFF
--- a/evap/results/exporters.py
+++ b/evap/results/exporters.py
@@ -160,7 +160,7 @@ class ResultsExporter(ExcelExporter):
         return evaluations_with_results, used_questionnaires, course_results_exist
 
     def write_headings_and_evaluation_info(
-        self, evaluations_with_results, semesters, contributor, degrees, course_types, verbose_heading=True
+        self, evaluations_with_results, semesters, contributor, degrees, course_types, verbose_heading
     ):
         export_name = "Evaluation"
         if contributor:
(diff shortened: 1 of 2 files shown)


janno42:
As discussed, this removes the list of degrees and course types globally from the exporter. Instead, you should provide a new parameter that decides whether these values are written to the Excel file or not. The filtering should still be applied.

evap/results/exporters.py
```diff
@@ -169,9 +169,13 @@ def write_headings_and_evaluation_info(
             export_name += f"\n{semesters[0].name}"
         degree_names = [degree.name for degree in Degree.objects.filter(pk__in=degrees)]
         course_type_names = [course_type.name for course_type in CourseType.objects.filter(pk__in=course_types)]
-        self.write_cell(
-            _("{}\n\n{}\n\n{}").format(export_name, ", ".join(degree_names), ", ".join(course_type_names)), "headline"
-        )
+        if verbose_heading:
+            self.write_cell(
```
richardebeling, line 173: The two paths here are not tested. Testing this should be relatively simple, can you add a small test for the two different headers? (Either at the view level by using two views that use the exporter here, or directly testing the exporter. Not quire sure which makes more sense, it feels like the view level would be a bit more stable)

richardebeling (commented):
Looking good, thanks :)

janno42:
Continued in #2057

Issue #1816, opened 2022-10-10
Position view selection buttons on top


janno42:
The pages `results_publication_detail` and `staff_evaluation_textanswers` have buttons to select different views on the top right.
When evaluations have long names that span over more than one line, the buttons are vertically positioned in the middle of the evaluation name.
Instead, they should stay on the top of the page.

florian-str:
All sides where an input field of some sort is next to a textfield are effected by this issue.

janno42:
On a quick glance I found only one other occurrence on `grades_semester_view`. Did you find more?

richardebeling:
@janno42 The issue text above says `results_publication_detail` where it should be `results_evaluation_detail`, correct?

fkoch-tgm:
Here are some Screenshots, I would like some input:

`grades_semester_view`:
<details><summary>Screenshots</summary>
<p>


Vorher:
![Screen Shot 2023-10-16 at 20 52 57](https://github.com/e-valuation/EvaP/assets/45557700/9920e9ad-227a-4342-8fe2-0b94485ff215)
Top-Aligned:
![Screen Shot 2023-10-16 at 20 52 25](https://github.com/e-valuation/EvaP/assets/45557700/9a9f3871-be15-4933-a899-d96a1859dd0d)

</p>
</details> 

`results_evaluation_details`:
<details><summary>Screenshots</summary>
<p>
Top:
![image](https://github.com/e-valuation/EvaP/assets/45557700/a715dc88-cc83-4d87-b1d4-c5698b51efa1)

Bottom:
![image](https://github.com/e-valuation/EvaP/assets/45557700/dc08c150-4fb5-4053-b0ae-b919e7c0cf65)

Center:
![image](https://github.com/e-valuation/EvaP/assets/45557700/fd55e211-af09-415d-944e-3bcb6b77b846)

</p>
</details> 


For `grades_semester_view` I personally think the centered version looks cleaner, but I am unsure.

For `results_evaluation_details` I would say that the bottom aligned version looks the best.

What do you think?

fkoch-tgm:
Here also Screenshots for `staff_evaluation_textanswers`:
<details><summary>Screenshots</summary>
<p>
Top:
![image](https://github.com/e-valuation/EvaP/assets/45557700/8fca65de-9f09-4491-971b-dce5cfffbdff)


Bottom:
[...]

PR #2047, opened 2023-10-23
Reposition View Selection Buttons

fkoch-tgm:
Addresses #1816

Current orientation for the Buttons: end or bottom

<details><summary>Screenshots</summary>

http://localhost:8000/results/semester/19/evaluation/1463
image

http://localhost:8000/staff/evaluation/1600/textanswers
image

http://localhost:8000/grades/semester/19
image

</details>


evap/results/templates/results_evaluation_detail.html
```diff
@@ -26,43 +26,55 @@
...
+                    <div class="btn-switch-label">{% trans 'View' %}</div>
+                    <div class="btn-switch btn-group">
+                        {% if user.is_staff and view == 'export' or is_contributor %}
+                            <a href="
+                                    {% if is_contributor %}{% url 'results:evaluation_detail' evaluation.course.semester.id evaluation.id %}?view=export{% endif %}"
+                               role="button" class="btn btn-sm btn-light{% if view == 'export' %} active{% endif %}"
+                               data-bs-toggle="tooltip" data-bs-placement="bottom"
+                               title="{% blocktrans %}Shows filtered view meant for personal export. Other contributors' results and private answers are hidden.{% endblocktrans %}">
+                                {% trans 'Export' context 'view mode' %}
+                            </a>
```
richardebeling, line 45: (Wait for @niklasmohrin to confirm before doing new formatting)

We don't use a html formatter (yet) because we're unaware of one that handles django's templating language well. Currently, the code base uses either a very packed style (pretty much all attributes on one line, this is what this file used before, with the exception that the long title was put on a separate line), or we use what [Prettier](https://prettier.io/) suggests for HTML with additional manual adjustments for the django template language constructs.

For consistency, I'd like to stick to one of these styles. See [the prettier.io playground for this snippet](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAhgAgBYCc4BmAvADojACkmAlgTQM4D6ksu1ARgK4wS6YUBfSpk64ANpgDk+epzEx6SOADd0YzuhjVojACZwY6amMmYVajVugA6SKPpxrDgLZx68XNeq6zq9Zu0oLx9BAH5lajgAdyI4AA8AB14YYQRdWn4BMhIoTDz8gsKi4vzcCDE4UhAuGB4oMkwwMXR6eir2GFyOqABaemdMbp6xagBzbBSqDIjozCIiKXik3BhTQUx0MC1lOFSodLpBbNyS07P83U10HvZ6Hp5R0YqqnnKtBIbLwxu7hOawOCuWDtCC1CDOY7nKElLQwZ5kYTsMQQMAAaxguHQUHomQAytgIFEcQRjB44D4ZlFMK4sTBMAReJgEnBcPRoGozIlktZMAB5GDYFmNaAYjjcXj0UwyOQKDb7JlsVTwOX0KIsnHofA4bz6IJ7XRIlHozHYzJkAB8OWh1sKwgxWJxkgAolyVqYWPA4nTJJTqRB9GssicbedUAB6dDmkAAGhAEASVmxyFAmrKUQACpqEPRkCA1FF0ABPHOx9iYtEGXHoVwAGWoUDgyAIagcpfLqMrCU29dGyAxnDgsZc1D7uAHQ57FQAipxQY2kM2xK2QAArehxXGTuAzudNluDkAAR1n8HTZQSOZQLR6DfJ5JjIHtxh7AGFwc50Lm1GIH-QtwBBWo2BqOB0xZOsGz3JcDwmZwxAAdWwah4HoLsAVxbNkOoCIYELXMwFaB9lAHABJfYEBgXEwDYBN-32XFcIqKDlwSMoHHgzEElzVi3BZHYH3rBwVjPdBRg-ZiDy7VkWVzWCf1jfBj2ofARLEz8F33WNDHYeDvAFZAABZY04BwABV0FuCTY0BdhyX0XQayxUYNFGOAADFeA-Woey-cUH1Y+sYF03R9KQABGAAmAAGAQBCAA) for the prettier.io formatting.

Also applies to other changes in HTML files

evap/results/templates/results_evaluation_detail.html
```diff
@@ -26,43 +26,55 @@
...
-                                {% trans 'Participant' %}
-                        {% else %}
-                            {% trans 'Shows results available for everyone logged in.' %}">
-                                {% trans 'Public' %}
+                                   {% trans 'Participant' %}
+                               {% else %}
+                                   {% trans 'Shows results available for everyone logged in.' %}">
+                                   {% trans 'Public' %}
+                               {% endif %}
+                            </a>
```
richardebeling, line 75: I think the indentation of the django tags should be "correct" following their django semantics, i.e. the `if` here needs to be indented more and the `else` and `endif` should be on the same indentation level as the `if`. The current indentation confused me a lot, I initially thought the `</a>` was moved inside the `if` accidentally.

I do see that the code here is confusing in any formatting because we have control logic while opening the anchor tag. The intention here was to not duplicate all the other attributes. This might be cleaner if expressed using [`blocktranslate`](https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#blocktranslate-template-tag) with the `asvar` attribute to create two new "local variables" in the template that can then be used as title and content of the anchor. If you want, you can change that.

evap/results/tests/test_views.py
```diff
@@ -471,20 +471,20 @@ def test_heading_question_filtering(self):
...
 
-        # the view=public button should have class "active". The rest in-between is just noise.
-        expected_button = (
-            f'<a href="/results/semester/{self.evaluation.course.semester.pk}/evaluation/{self.evaluation.pk}'
-            + '?view=public" role="button" class="btn btn-sm btn-light active"'
-        )
-
         page_without_get_parameter = self.app.get(self.url, user=self.manager)
-        self.assertContains(page_without_get_parameter, expected_button)
+        self.assertTrue(page_without_get_parameter.context["view"] == "public")
```
richardebeling, line 475: You can use
```suggestion
        self.assertEqual(page_without_get_parameter.context["view"], "public")
```
here and below, which will give a more descriptive error message if the assertion fails

evap/results/tests/test_views.py
```diff
@@ -471,20 +471,20 @@ def test_heading_question_filtering(self):
...
 
         page_with_random_get_parameter = self.app.get(self.url + "?view=asdf", user=self.manager)
-        self.assertContains(page_with_random_get_parameter, expected_button)
+        self.assertTrue(page_with_random_get_parameter.context["view"] == "public")
+
+        page_with_full_get_parameter = self.app.get(f"{self.url}?view=full", user=self.manager)
+        self.assertFalse(page_with_full_get_parameter.context["view"] == "public")
+
+        page_with_export_get_parameter = self.app.get(f"{self.url}?view=export", user=self.manager)
+        self.assertFalse(page_with_export_get_parameter.context["view"] == "public")
```
richardebeling, line 487: I'd go with the assertion that the `view` variable is `"full"` or `"export"`, instead of asserting that it is `not "public"`.

evap/static/bootstrap
richardebeling, line 1: Is the bootstrap submodule version change intentional here?

(Updating bootstrap likely introduces some front-end bugs, we'll have to manually check all pages when updating bootstrap. Typically, we do this in isolated PRs.)

richardebeling (changes requested):
Works nicely, thanks :)

evap/static/bootstrap
fkoch-tgm: That was not intentional

evap/results/templates/results_evaluation_detail.html
fkoch-tgm: Sorry, I wasn't aware that there is no tooling set up to format/validate the HTML/Django templates. (I just relied on my IDEs auto-formatting and assumed the precommit script would complain if it's against the project style)

I reformatted all my changes to use the "prettier" version.

Should I also reformat the rest of the file to fit this style?

richardebeling: I think the current state is fine, thanks 👍

richardebeling (approved):
Code looks good to me

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

PR #2049, opened 2023-10-23
Fix user selection form for user merge view

hansegucker:
including regression test

### DIFF
--- a/evap/staff/tests/test_views.py
+++ b/evap/staff/tests/test_views.py
@@ -328,18 +328,15 @@ class TestUserDeleteView(DeleteViewTestMixin, WebTestStaffMode):
         return {"user_id": cls.instance.pk}
 
 
-class TestUserMergeSelectionView(WebTestStaffModeWith200Check):
+class TestUserMergeSelectionView(WebTestStaffMode):
     url = "/staff/user/merge"
 
     @classmethod
     def setUpTestData(cls):
         cls.manager = make_manager()
-        cls.test_users = [cls.manager]
 
-        baker.make(UserProfile)
-
-        cls.main_user = baker.make(UserProfile)
-        cls.other_user = baker.make(UserProfile)
+        cls.main_user = baker.make(UserProfile, _fill_optional=["email"])
+        cls.other_user = baker.make(UserProfile, _fill_optional=["email"])
 
     def test_redirection_user_merge_view(self):
         page = self.app.get(self.url, user=self.manager)
@@ -348,9 +345,10 @@ class TestUserMergeSelectionView(WebTestStaffModeWith200Check):
         form["main_user"] = self.main_user.pk
         form["other_user"] = self.other_user.pk
 
-        page = form.submit(name="operation", value="test")
+        page = form.submit().follow()
 
-        self.assertContains(page, "", status_code=302)
+        self.assertContains(page, self.main_user.email)
+        self.assertContains(page, self.other_user.email)
 
 
 class TestUserMergeView(WebTestStaffModeWith200Check):
--- a/evap/staff/tests/test_views.py
+++ b/evap/staff/tests/test_views.py
@@ -333,9 +333,24 @@ class TestUserMergeSelectionView(WebTestStaffModeWith200Check):
 
     @classmethod
     def setUpTestData(cls):
-        cls.test_users = [make_manager()]
+        cls.manager = make_manager()
+        cls.test_users = [cls.manager]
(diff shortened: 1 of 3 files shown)


evap/staff/tests/test_views.py
```diff
@@ -333,10 +333,25 @@ class TestUserMergeSelectionView(WebTestStaffModeWith200Check):
...
+    def test_redirection_user_merge_view(self):
+        page = self.app.get(self.url, user=self.manager)
+
+        form = page.forms["user-selection-form"]
+        form["main_user"] = self.main_user.pk
+        form["other_user"] = self.other_user.pk
+
+        page = form.submit(name="operation", value="test")
+
+        self.assertContains(page, "", status_code=302)
```
janno42, line 353: you could test that after the redirect (get the page by `.follow()`) the page contains the names or email addresses of the two users.

janno42 (changes requested):
✔️ Meets requirements
✔️ UI functionality checked

two additional comments though :)

evap/staff/tests/test_views.py
```diff
@@ -333,10 +333,25 @@ class TestUserMergeSelectionView(WebTestStaffModeWith200Check):
...
     @classmethod
     def setUpTestData(cls):
-        cls.test_users = [make_manager()]
+        cls.manager = make_manager()
+        cls.test_users = [cls.manager]
 
         baker.make(UserProfile)
 
+        cls.main_user = baker.make(UserProfile)
+        cls.other_user = baker.make(UserProfile)
```
richardebeling, line 342: The previous `baker.make(UserProfile)` call was probably required to that our very basic "When you load the page it doesn't crash" check from `WebTestStaffModeWith200Check` has some data available. Now that you create additional users, I would think that this can be removed:

```suggestion
        cls.main_user = baker.make(UserProfile)
        cls.other_user = baker.make(UserProfile)
```

It might not even make sense to have the 200 check anymore, since the test case below also requires that (implicitly in the `self.app.get()`, which will fail for non-200 response codes). In that case we can also go back to inheriting from `WebTestStaffMode` here.

evap/staff/views.py
```diff
@@ -2289,11 +2289,11 @@ class UserMergeSelectionView(FormView):
...
     template_name = "staff_user_merge_selection.html"
 
-    def get_success_url(self):
+    def form_valid(self, form):
         return redirect(
             "staff:user_merge",
-            self.form.cleaned_data["main_user"].id,
-            self.form.cleaned_data["other_user"].id,
+            form.cleaned_data["main_user"].id,
+            form.cleaned_data["other_user"].id,
```
richardebeling, line 2296: @niklasmohrin from what I currently understand, `self.form` was not defined here (at least I can not find why it should be defined in django's docs), and `get_success_url` should probably return an url, not a HttpResponse (as redirect does). Is there a chance that we messed this up with other views as well?

(The code form this PR seems to be ok to me according to the docs)

niklasmohrin: Oh no I think we were literally talking about this code and agreed that it is trivial enough to not test or something 😂: https://github.com/e-valuation/EvaP/pull/1964#discussion_r1265804150

niklasmohrin: But in all honesty, we should probably go over the generic views PR again and maybe add types or so? Its mostly just classes with variables, those are fine, but the rest we should check to be sure

richardebeling: Created #2055 for this. Resolving here.

hansegucker:
@niklasmohrin @richardebeling Please review again.

evap/staff/tests/test_views.py
```diff
@@ -328,14 +328,29 @@ def get_post_params(cls):
...
+    def test_redirection_user_merge_view(self):
+        page = self.app.get(self.url, user=self.manager)
+
+        form = page.forms["user-selection-form"]
+        form["main_user"] = self.main_user.pk
+        form["other_user"] = self.other_user.pk
+
+        page = form.submit()
+
+        page = page.follow()
```
richardebeling, line 350:
```suggestion
        page = form.submit().follow()
```
is what we usually use, the current code is also fine though.

richardebeling (approved):

richardebeling (approved):

janno42 (approved):
✔️ UI functionality checked

niklasmohrin (approved):
Thanks :)

Issue #2008, opened 2023-08-28
Shortcut button for creating exam evaluation


janno42:
A new button should be added to the actions column of the Evaluations tab of the staff semester page.
This button (with a `fa-file-pen` icon and the tooltip `Create exam evaluation`) should open a modal titled `Create exam evaluation` with a date input field (labeled `Exam date`) below the following text: `Create an exam evaluation based on this evaluation. This will copy all the participants and contributors from the original evaluation. It will set the weight of the original evaluation to 9 and its end date will be set to the day before the exam.`

When submitting the modal form (`Create exam evaluation` button), the following should happen:
- The weight of the evaluation at which the button was pressed (original evaluation) will be set to 9. If the evaluation has already been published, an error message is displayed (and nothing happens).
- If the original evaluation has not yet ended, the evaluation period end date will be set to the day before the one selected in the modal. If this date is before the start date, an error message is displayed (and nothing happens).
- A new evaluation will be created for the same course as the original evaluation (exam evaluation), will be **named "Klausur/Exam"**, have a **weight of 1**, and be **unrewarded**.
- The start datetime of the exam evaluation is 08:00 on the day after the exam date defined in the modal.
- The end date of the exam evaluation is the third day after the exam date defined in the modal.
- All participants from the original evaluation will be copied to the exam evaluation.
- All contributors from the original evaluation will be copied to the exam evaluation without any assigned questionnaires.
- All questionnaires listed in a new settings variable `EXAM_QUESTIONNAIRES` (initially set to contain the id of the "Miscellaneous" questionnaire) will be added to the exam evaluation as general questionnaires.

PR #2050, opened 2023-10-23
Adds shortcut button to add exam evaluation

FSadrieh:
#2008

### DIFF
--- a/deployment/localsettings.template.py
+++ b/deployment/localsettings.template.py
@@ -53,3 +53,5 @@ GLOBAL_EVALUATION_PROGRESS_INFO_TEXT = {
     "de": mark_safe("Deine Teilnahme am Evaluationsprojekt wird helfen. Evaluiere also <b>jetzt</b>!"),
     "en": mark_safe("Your participation in the evaluation helps, so evaluate <b>now</b>!"),
 }
+# Questionnaires automatically added to exam evaluations
+EXAM_QUESTIONNAIRE_IDS = [83]
--- a/deployment/localsettings.template.py
+++ b/deployment/localsettings.template.py
@@ -54,4 +54,4 @@ GLOBAL_EVALUATION_PROGRESS_INFO_TEXT = {
     "en": mark_safe("Your participation in the evaluation helps, so evaluate <b>now</b>!"),
 }
 # Questionnaires automatically added to exam evaluations
-EXAM_QUESTIONNAIRE_IDS = [83]
+EXAM_QUESTIONNAIRE_IDS = [111]
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -471,11 +471,11 @@ class Evaluation(LoggedModel):
     )
 
     @property
-    def has_exam(self):
+    def has_exam_evaluation(self):
         return self.course.evaluations.filter(name_de="Klausur", name_en="Exam").exists()
 
     @property
-    def get_earliest_legal_exam_date(self):
+    def earliest_possible_exam_date(self):
         return self.vote_start_datetime.date() + timedelta(days=1)
 
     @transaction.atomic
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -22,7 +22,6 @@ from django.core.mail import EmailMessage, EmailMultiAlternatives
 from django.db import IntegrityError, models, transaction
 from django.db.models import CheckConstraint, Count, Exists, F, Manager, OuterRef, Q, Subquery, Value
 from django.db.models.functions import Coalesce, Lower, NullIf, TruncDate
-from django.db.models.query import QuerySet
 from django.dispatch import Signal, receiver
 from django.http import HttpRequest
 from django.template import Context, Template
@@ -496,7 +495,6 @@ class Evaluation(LoggedModel):
             exam_evaluation.contributions.create(contributor=contribution.contributor)
         exam_evaluation.general_contribution.questionnaires.set(settings.EXAM_QUESTIONNAIRE_IDS)
         exam_evaluation.save()
-
 
     class TextAnswerReviewState(Enum):
         NO_TEXTANSWERS = auto()
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -20,8 +20,9 @@ from django.core.cache import caches
 from django.core.exceptions import ValidationError
 from django.core.mail import EmailMessage, EmailMultiAlternatives
 from django.db import IntegrityError, models, transaction
-from django.db.models import CheckConstraint, Count, Exists, F, Manager, OuterRef, Q, Subquery, Value, query
+from django.db.models import CheckConstraint, Count, Exists, F, Manager, OuterRef, Q, Subquery, Value
 from django.db.models.functions import Coalesce, Lower, NullIf, TruncDate
+from django.db.models.query import QuerySet
 from django.dispatch import Signal, receiver
 from django.http import HttpRequest
 from django.template import Context, Template
@@ -474,33 +475,35 @@ class Evaluation(LoggedModel):
     def has_exam(self):
         return self.course.evaluations.filter(name_de="Klausur", name_en="Exam").exists()
 
-    def make_exam_evaluation(
+    @transaction.atomic
+    def create_exam_evaluation(
         self,
         exam_date: datetime,
         evaluation_end_date: datetime,
-        participants: query.QuerySet["UserProfile"],
-        eval_contributions: query.QuerySet["Contribution"],
     ):
+        def _set_exam_evaluation_attributes(
+            exam_evaluation: Evaluation,
+            exam_date: date,
+            participants: QuerySet["UserProfile"],
+            eval_contributions: QuerySet["Contribution"],
+        ):
+            exam_evaluation.vote_start_datetime = datetime.combine(exam_date + timedelta(days=1), time(8, 0))
+            exam_evaluation.vote_end_date = exam_date + timedelta(days=3)
+            exam_evaluation.save()
+            exam_evaluation.participants.set(participants)
+            for contribution in eval_contributions:
+                exam_evaluation.contributions.create(contributor=contribution.contributor)
+            exam_evaluation.general_contribution.questionnaires.set(settings.EXAM_QUESTIONNAIRE_IDS)
+            exam_evaluation.save()
+
         self.weight = 9
         self.vote_end_date = evaluation_end_date
         self.save()
+        participants = self.participants.all()
+        eval_contributions = self.contributions.exclude(contributor=None)
         exam_evaluation = Evaluation(course=self.course, name_de="Klausur", name_en="Exam", weight=1, is_rewarded=False)
-        exam_evaluation.set_exam_evaluation_attributes(exam_date, participants, eval_contributions)
-        exam_evaluation.save()
+        _set_exam_evaluation_attributes(exam_evaluation, exam_date, participants, eval_contributions)
 
-    def set_exam_evaluation_attributes(
-        self,
-        exam_date: date,
-        participants: query.QuerySet["UserProfile"],
-        eval_contributions: query.QuerySet["Contribution"],
-    ):
-        self.vote_start_datetime = datetime.combine(exam_date + timedelta(days=1), time(8, 0))
-        self.vote_end_date = exam_date + timedelta(days=3)
-        self.save()
-        self.participants.set(participants)
-        for contribution in eval_contributions:
-            self.contributions.create(contributor=contribution.contributor)
-        self.general_contribution.questionnaires.set(settings.EXAM_QUESTIONNAIRE_IDS)
 
     class TextAnswerReviewState(Enum):
         NO_TEXTANSWERS = auto()
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -20,7 +20,7 @@ from django.core.cache import caches
 from django.core.exceptions import ValidationError
 from django.core.mail import EmailMessage, EmailMultiAlternatives
 from django.db import IntegrityError, models, transaction
-from django.db.models import CheckConstraint, Count, Exists, F, Manager, OuterRef, Q, Subquery, Value
+from django.db.models import CheckConstraint, Count, Exists, F, Manager, OuterRef, Q, Subquery, Value, query
 from django.db.models.functions import Coalesce, Lower, NullIf, TruncDate
 from django.dispatch import Signal, receiver
 from django.http import HttpRequest
@@ -474,7 +474,26 @@ class Evaluation(LoggedModel):
     def has_exam(self):
         return self.course.evaluations.filter(name_de="Klausur", name_en="Exam").exists()
 
-    def make_exam_evaluation(self, exam_date: date, participants, eval_contributions):
+    def make_exam_evaluation(
+        self,
+        exam_date: datetime,
+        evaluation_end_date: datetime,
+        participants: query.QuerySet["UserProfile"],
+        eval_contributions: query.QuerySet["Contribution"],
+    ):
+        self.weight = 9
+        self.vote_end_date = evaluation_end_date
+        self.save()
+        exam_evaluation = Evaluation(course=self.course, name_de="Klausur", name_en="Exam", weight=1, is_rewarded=False)
+        exam_evaluation.set_exam_evaluation_attributes(exam_date, participants, eval_contributions)
+        exam_evaluation.save()
+
+    def set_exam_evaluation_attributes(
+        self,
+        exam_date: date,
+        participants: query.QuerySet["UserProfile"],
+        eval_contributions: query.QuerySet["Contribution"],
+    ):
         self.vote_start_datetime = datetime.combine(exam_date + timedelta(days=1), time(8, 0))
         self.vote_end_date = exam_date + timedelta(days=3)
         self.save()
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -4,7 +4,7 @@ import uuid
 from collections import defaultdict
 from collections.abc import Collection, Container, Iterable, Sequence
 from dataclasses import dataclass
-from datetime import date, datetime, timedelta
+from datetime import date, datetime, time, timedelta
 from enum import Enum, auto
 from functools import partial
 from numbers import Real
@@ -469,6 +469,19 @@ class Evaluation(LoggedModel):
     wait_for_grade_upload_before_publishing = models.BooleanField(
         verbose_name=_("wait for grade upload before publishing"), default=True
     )
+
+    @property
+    def has_exam(self):
+        return self.course.evaluations.filter(name_de="Klausur", name_en="Exam").exists()
+
+    def make_exam_evaluation(self, exam_date: date, participants, eval_contributions):
+        self.vote_start_datetime = datetime.combine(exam_date + timedelta(days=1), time(8, 0))
+        self.vote_end_date = exam_date + timedelta(days=3)
+        self.save()
+        self.participants.set(participants)
+        for contribution in eval_contributions:
+            self.contributions.create(contributor=contribution.contributor)
+        self.general_contribution.questionnaires.set(settings.EXAM_QUESTIONNAIRES)
 
     class TextAnswerReviewState(Enum):
         NO_TEXTANSWERS = auto()
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -479,30 +479,21 @@ class Evaluation(LoggedModel):
     def create_exam_evaluation(
         self,
         exam_date: datetime,
-        evaluation_end_date: datetime,
+        evaluation_end_date: date,
     ):
-        def _set_exam_evaluation_attributes(
-            exam_evaluation: Evaluation,
-            exam_date: date,
-            participants: QuerySet["UserProfile"],
-            eval_contributions: QuerySet["Contribution"],
-        ):
-            exam_evaluation.vote_start_datetime = datetime.combine(exam_date + timedelta(days=1), time(8, 0))
-            exam_evaluation.vote_end_date = exam_date + timedelta(days=3)
-            exam_evaluation.save()
-            exam_evaluation.participants.set(participants)
-            for contribution in eval_contributions:
-                exam_evaluation.contributions.create(contributor=contribution.contributor)
-            exam_evaluation.general_contribution.questionnaires.set(settings.EXAM_QUESTIONNAIRE_IDS)
-            exam_evaluation.save()
-
         self.weight = 9
         self.vote_end_date = evaluation_end_date
         self.save()
         participants = self.participants.all()
         eval_contributions = self.contributions.exclude(contributor=None)
         exam_evaluation = Evaluation(course=self.course, name_de="Klausur", name_en="Exam", weight=1, is_rewarded=False)
-        _set_exam_evaluation_attributes(exam_evaluation, exam_date, participants, eval_contributions)
+        exam_evaluation.vote_start_datetime = datetime.combine(exam_date + timedelta(days=1), time(8, 0))
+        exam_evaluation.vote_end_date = exam_date + timedelta(days=3)
+        exam_evaluation.participants.set(participants)
+        for contribution in eval_contributions:
+            exam_evaluation.contributions.create(contributor=contribution.contributor)
+        exam_evaluation.general_contribution.questionnaires.set(settings.EXAM_QUESTIONNAIRE_IDS)
+        exam_evaluation.save()
 
 
     class TextAnswerReviewState(Enum):
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -498,7 +498,6 @@ class Evaluation(LoggedModel):
         for contribution in self.contributions.exclude(contributor=None):
             exam_evaluation.contributions.create(contributor=contribution.contributor)
         exam_evaluation.general_contribution.questionnaires.set(settings.EXAM_QUESTIONNAIRE_IDS)
-        exam_evaluation.save()
 
     class TextAnswerReviewState(Enum):
         NO_TEXTANSWERS = auto()
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -474,11 +474,12 @@ class Evaluation(LoggedModel):
     def has_exam(self):
         return self.course.evaluations.filter(name_de="Klausur", name_en="Exam").exists()
 
+    @property
+    def get_earliest_legal_exam_date(self):
+        return self.vote_start_datetime.date() + timedelta(days=1)
+
     @transaction.atomic
-    def create_exam_evaluation(
-        self,
-        exam_date: date,
-    ):
+    def create_exam_evaluation(self, exam_date: date):
         self.weight = 9
         self.vote_end_date = exam_date - timedelta(days=1)
         self.save()
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -483,9 +483,15 @@ class Evaluation(LoggedModel):
         self.weight = 9
         self.vote_end_date = exam_date - timedelta(days=1)
         self.save()
-        exam_evaluation = Evaluation(course=self.course, name_de="Klausur", name_en="Exam", weight=1, is_rewarded=False)
-        exam_evaluation.vote_start_datetime = datetime.combine(exam_date + timedelta(days=1), time(8, 0))
-        exam_evaluation.vote_end_date = exam_date + timedelta(days=3)
+        exam_evaluation = Evaluation(
+            course=self.course,
+            name_de="Klausur",
+            name_en="Exam",
+            weight=1,
+            is_rewarded=False,
+            vote_start_datetime=datetime.combine(exam_date + timedelta(days=1), time(8, 0)),
+            vote_end_date=exam_date + timedelta(days=3),
+        )
         exam_evaluation.save()
 
         exam_evaluation.participants.set(self.participants.all())
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -477,21 +477,18 @@ class Evaluation(LoggedModel):
     @transaction.atomic
     def create_exam_evaluation(
         self,
-        exam_date: datetime,
-        evaluation_end_date: date,
+        exam_datetime: datetime,
     ):
         self.weight = 9
-        self.vote_end_date = evaluation_end_date
+        self.vote_end_date = exam_datetime - timedelta(days=1)
         self.save()
-        participants = self.participants.all()
-        eval_contributions = self.contributions.exclude(contributor=None)
         exam_evaluation = Evaluation(course=self.course, name_de="Klausur", name_en="Exam", weight=1, is_rewarded=False)
-        exam_evaluation.vote_start_datetime = datetime.combine(exam_date + timedelta(days=1), time(8, 0))
-        exam_evaluation.vote_end_date = exam_date + timedelta(days=3)
+        exam_evaluation.vote_start_datetime = datetime.combine(exam_datetime + timedelta(days=1), time(8, 0))
+        exam_evaluation.vote_end_date = exam_datetime + timedelta(days=3)
         exam_evaluation.save()
 
-        exam_evaluation.participants.set(participants)
-        for contribution in eval_contributions:
+        exam_evaluation.participants.set(self.participants.all())
+        for contribution in self.contributions.exclude(contributor=None):
             exam_evaluation.contributions.create(contributor=contribution.contributor)
         exam_evaluation.general_contribution.questionnaires.set(settings.EXAM_QUESTIONNAIRE_IDS)
         exam_evaluation.save()
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -481,7 +481,7 @@ class Evaluation(LoggedModel):
         self.participants.set(participants)
         for contribution in eval_contributions:
             self.contributions.create(contributor=contribution.contributor)
-        self.general_contribution.questionnaires.set(settings.EXAM_QUESTIONNAIRES)
+        self.general_contribution.questionnaires.set(settings.EXAM_QUESTIONNAIRE_IDS)
 
     class TextAnswerReviewState(Enum):
         NO_TEXTANSWERS = auto()
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -489,6 +489,8 @@ class Evaluation(LoggedModel):
         exam_evaluation = Evaluation(course=self.course, name_de="Klausur", name_en="Exam", weight=1, is_rewarded=False)
         exam_evaluation.vote_start_datetime = datetime.combine(exam_date + timedelta(days=1), time(8, 0))
         exam_evaluation.vote_end_date = exam_date + timedelta(days=3)
+        exam_evaluation.save()
+
         exam_evaluation.participants.set(participants)
         for contribution in eval_contributions:
             exam_evaluation.contributions.create(contributor=contribution.contributor)
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -477,14 +477,14 @@ class Evaluation(LoggedModel):
     @transaction.atomic
     def create_exam_evaluation(
         self,
-        exam_datetime: datetime,
+        exam_date: date,
     ):
         self.weight = 9
-        self.vote_end_date = exam_datetime - timedelta(days=1)
+        self.vote_end_date = exam_date - timedelta(days=1)
         self.save()
         exam_evaluation = Evaluation(course=self.course, name_de="Klausur", name_en="Exam", weight=1, is_rewarded=False)
-        exam_evaluation.vote_start_datetime = datetime.combine(exam_datetime + timedelta(days=1), time(8, 0))
-        exam_evaluation.vote_end_date = exam_datetime + timedelta(days=3)
+        exam_evaluation.vote_start_datetime = datetime.combine(exam_date + timedelta(days=1), time(8, 0))
+        exam_evaluation.vote_end_date = exam_date + timedelta(days=3)
         exam_evaluation.save()
 
         exam_evaluation.participants.set(self.participants.all())
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -211,5 +211,18 @@
         </script>
 
         {% block additional_javascript %}{% endblock %}
+
+        <script type="text/javascript"> 
+            document.querySelectorAll("[data-earliest-legal-date]").forEach(element => {
+                element.addEventListener("change", () => {
+                    if (element.value < element.dataset.earliestLegalDate) {
+                        element.setCustomValidity(element.dataset.earliestDateReason);
+                    } else {
+                        element.setCustomValidity("");
+                    }
+                });
+                element.dispatchEvent(new Event("change"));
+            });
+        </script>
     </body>
 </html>
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -210,8 +210,6 @@
             setupForms();
         </script>
 
-        {% block additional_javascript %}{% endblock %}
-
         <script type="text/javascript"> 
             document.querySelectorAll("[data-earliest-possible-date]").forEach(element => {
                 element.addEventListener("change", () => {
@@ -224,5 +222,7 @@
                 element.dispatchEvent(new Event("change"));
             });
         </script>
+
+        {% block additional_javascript %}{% endblock %}
     </body>
 </html>
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -213,9 +213,9 @@
         {% block additional_javascript %}{% endblock %}
 
         <script type="text/javascript"> 
-            document.querySelectorAll("[data-earliest-legal-date]").forEach(element => {
+            document.querySelectorAll("[data-earliest-possible-exam-date]").forEach(element => {
                 element.addEventListener("change", () => {
-                    if (element.value < element.dataset.earliestLegalDate) {
+                    if (element.value < element.dataset.earliestPossibleDate) {
                         element.setCustomValidity(element.dataset.earliestDateReason);
                     } else {
                         element.setCustomValidity("");
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -213,7 +213,7 @@
         {% block additional_javascript %}{% endblock %}
 
         <script type="text/javascript"> 
-            document.querySelectorAll("[data-earliest-possible-exam-date]").forEach(element => {
+            document.querySelectorAll("[data-earliest-possible-date]").forEach(element => {
                 element.addEventListener("change", () => {
                     if (element.value < element.dataset.earliestPossibleDate) {
                         element.setCustomValidity(element.dataset.earliestDateReason);
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -210,19 +210,6 @@
             setupForms();
         </script>
 
-        <script type="text/javascript"> 
-            document.querySelectorAll("[data-earliest-possible-date]").forEach(element => {
-                element.addEventListener("change", () => {
-                    if (element.value < element.dataset.earliestPossibleDate) {
-                        element.setCustomValidity(element.dataset.earliestPossibleDateReason);
-                    } else {
-                        element.setCustomValidity("");
-                    }
-                });
-                element.dispatchEvent(new Event("change"));
-            });
-        </script>
-
         {% block additional_javascript %}{% endblock %}
     </body>
 </html>
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -216,7 +216,7 @@
             document.querySelectorAll("[data-earliest-possible-date]").forEach(element => {
                 element.addEventListener("change", () => {
                     if (element.value < element.dataset.earliestPossibleDate) {
-                        element.setCustomValidity(element.dataset.earliestDateReason);
+                        element.setCustomValidity(element.dataset.earliestPossibleDateReason);
                     } else {
                         element.setCustomValidity("");
                     }
--- a/evap/staff/templates/staff_semester_view.html
+++ b/evap/staff/templates/staff_semester_view.html
@@ -546,10 +546,11 @@
                                             href="{% url 'staff:course_copy' course.id %}"
                                             title="{% translate 'Copy course' %}">
                                             <span class="fas fa-copy"></span>
-                                            <a class="btn btn-sm btn-dark" data-bs-toggle="tooltip"
-                                            onclick="createExamEvaluationShow();"
-                                            title="{% trans 'Create exam evaluation' %}">
-                                            <span class="fas fa-file-pen fa-fw"></span>
+                                        </a>
+                                        <a class="btn btn-sm btn-dark" data-bs-toggle="tooltip"
+                                        onclick="createExamEvaluationShow({{ course.id }}, '{{ course.name|escapejs }}');"
+                                        title="{% trans 'Create exam evaluation' %}">
+                                        <span class="fas fa-file-pen fa-fw"></span>
                                         </a>
                                         {% endif %}
                                         {% if course.can_be_deleted_by_manager %}
@@ -591,8 +592,16 @@
     {% trans 'Create exam evaluation' as action_text %}
     {% include 'confirmation_modal.html' with modal_id='createExamEvaluation' title=title question=question action_text=action_text btn_type='danger' %}
     <script type="text/javascript">
-        function createExamEvaluationAction() {
-            pass
+        function createExamEvaluationAction(dataId) {
+            alert(dataId);
+            fetch("{% url 'staff:create_exam_evaluation' %}", {
+                body: new URLSearchParams({course_id: dataId}),
+                headers: CSRF_HEADERS,
+                method: "POST",
+            }).then(response => {
+                assert(response.ok);
+                location.reload();
+            }).catch(error => {window.alert("{% trans 'The server is not responding.' %}");});
         };
     </script>
 {% endblock %}
--- a/evap/staff/templates/staff_semester_view.html
+++ b/evap/staff/templates/staff_semester_view.html
@@ -585,7 +585,7 @@
     {% trans 'Create exam evaluation' as title %}
     {% blocktrans asvar question %}Create an exam evaluation based on this evaluation. This will copy all the participants and contributors from the original evaluation. It will set the weight of the original evaluation to 9 and its end date will be set to the day before the exam. {% endblocktrans %}
     {% trans 'Create exam evaluation' as action_text %}
-    {% include 'confirmation_modal.html' with modal_id='createExamEvaluation' title=title question=question action_text=action_text btn_type='danger' %}
+    {% include 'confirmation_modal.html' with modal_id='createExamEvaluation' title=title question=question action_text=action_text btn_type='primary' %}
     <script type="text/javascript">
         function createExamEvaluationAction(dataId) {
             fetch("{% url 'staff:create_exam_evaluation' %}", {
--- a/evap/staff/templates/staff_semester_view.html
+++ b/evap/staff/templates/staff_semester_view.html
@@ -374,6 +374,7 @@
                     </div>
                 </div>
 
+                <!-- Each exam_creation_modal needs its own exam date input field. -->
                 {% for evaluation in evaluations %}
                     <form id="exam_creation_form_{{ evaluation.id }}" reload-on-success method="post" action="{% url 'staff:create_exam_evaluation' %}">
                         {% csrf_token %}
--- a/evap/staff/templates/staff_semester_view.html
+++ b/evap/staff/templates/staff_semester_view.html
@@ -499,7 +499,7 @@
                             <col />
                             <col style="width: 14em" />
                             <col style="width: 12em" />
-                            <col style="width: 12.9em" />
+                            <col style="width: 9em" />
                         </colgroup>
                         <thead>
                             <tr>
@@ -547,11 +547,6 @@
                                             title="{% translate 'Copy course' %}">
                                             <span class="fas fa-copy"></span>
                                         </a>
-                                        <a class="btn btn-sm btn-dark" data-bs-toggle="tooltip"
-                                        onclick="createExamEvaluationShow({{ course.id }}, '{{ course.name|escapejs }}');"
-                                        title="{% trans 'Create exam evaluation' %}">
-                                        <span class="fas fa-file-pen fa-fw"></span>
-                                        </a>
                                         {% endif %}
                                         {% if course.can_be_deleted_by_manager %}
                                             <confirmation-modal type="submit" form="course-deletion-form" name="course_id" value="{{ course.id }}" confirm-button-class="btn-danger">
@@ -593,9 +588,8 @@
     {% include 'confirmation_modal.html' with modal_id='createExamEvaluation' title=title question=question action_text=action_text btn_type='danger' %}
     <script type="text/javascript">
         function createExamEvaluationAction(dataId) {
-            alert(dataId);
             fetch("{% url 'staff:create_exam_evaluation' %}", {
-                body: new URLSearchParams({course_id: dataId}),
+                body: new URLSearchParams({evaluation_id: dataId}),
                 headers: CSRF_HEADERS,
                 method: "POST",
             }).then(response => {
--- a/evap/staff/templates/staff_semester_view.html
+++ b/evap/staff/templates/staff_semester_view.html
@@ -374,6 +374,12 @@
                     </div>
                 </div>
 
+                {% for evaluation in evaluations %}
+                    <form id="exam_creation_form_{{ evaluation.id }}" reload-on-success method="post" action="{% url 'staff:create_exam_evaluation' %}">
+                        {% csrf_token %}
+                    </form>
+                {% endfor %}
+
                 <form id="evaluation-deletion-form" custom-success method="POST" action="{% url 'staff:evaluation_delete' %}">
                     {% csrf_token %}
                 </form>
@@ -383,25 +389,6 @@
                         fadeOutThenRemove(document.getElementById(`evaluation-row-${event.detail.body.get("evaluation_id")}`))
                     });
                 </script>
-
-                <form reload-on-success method="post" action="{% url 'staff:create_exam_evaluation' %}">
-                    {% csrf_token %}
-                    <confirmation-modal type="submit" id="exam_creation_modal" name="evaluation_id" value="{{ evaluation.id }}" confirm-button-class="btn-primary" title="{% translate 'Create exam evaluations' %}">
-                        <span slot="title">{% translate 'Create exam evaluations' %}</span>
-                        <span slot="action-text">{% translate 'Create exam evaluation' %}</span>
-                        <span slot="question">
-                            {% blocktranslate %}
-                            Create an exam evaluation based on this evaluation. This will copy all the participants and contributors from the original evaluation. It will set the weight of the original evaluation to 9 and its end date will be set to the day before the exam.
-                            {% endblocktranslate %}
-                        </span>
-                        <span slot="extra-inputs">
-                            <label for="exam_date">{% translate 'Exam Date:' %}</label>
-                            <input type="date" id="exam_date" name="exam_date" required>
-                        </span>
-
-                        <button slot="show-button" type="button" class="btn btn-sm btn-light"><span class="fas fa-file-pen fa-fw"></span></button>
-                    </confirmation-modal>
-                </form>
 
                 <form id="evaluation_operation_form" class="table-responsive" method="GET" action="{% url 'staff:evaluation_operation' semester.pk %}">
                     {% if num_evaluations > 0 %}
--- a/evap/staff/templates/staff_semester_view.html
+++ b/evap/staff/templates/staff_semester_view.html
@@ -384,6 +384,20 @@
                     });
                 </script>
 
+                <form reload-on-success method="post" action="{% url 'staff:create_exam_evaluation' %}">
+                    {% csrf_token %}
+                    <confirmation-modal type="submit" id="exam_creation_modal" name="evaluation_id" value="{{ evaluation.id }}" confirm-button-class="btn-primary" title="{% translate 'Create exam evaluations' %}">
+                        <span slot="title">{% translate 'Create exam evaluations' %}</span>
+                        <span slot="action-text">{% translate 'Create exam evaluation' %}</span>
+                        <span slot="question">
+                            {% blocktranslate %}
+                            Create an exam evaluation based on this evaluation. This will copy all the participants and contributors from the original evaluation. It will set the weight of the original evaluation to 9 and its end date will be set to the day before the exam.
+                            {% endblocktranslate %}
+                        </span>
+                        <button slot="show-button" type="button" class="btn btn-sm btn-light"><span class="fas fa-file-pen fa-fw"></span></button>
+                    </confirmation-modal>
+                </form>
+
                 <form id="evaluation_operation_form" class="table-responsive" method="GET" action="{% url 'staff:evaluation_operation' semester.pk %}">
                     {% if num_evaluations > 0 %}
                         <table id="evaluation-table" class="table table-striped table-narrow table-vertically-aligned">
@@ -466,6 +480,7 @@
                     {% endif %}
                 </form>
             </div>
+
             <div class="tab-pane" id="courses" role="tabpanel">
                 <div class="row align-items-center mb-3">
                     <div class="col-9">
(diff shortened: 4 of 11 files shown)


evap/staff/templates/staff_semester_view.html
```diff
@@ -371,7 +371,7 @@ <h3>
                             <col />
                             <col style="width: 14em" />
                             <col style="width: 12em" />
-                            <col style="width: 9em" />
+                            <col style="width: 12.9em" />
```
richardebeling, line 374: @janno42 do we have a mechanic to make a column "shrink-to-fit"?

janno42: @FranzSw is working on the first implementation in #1889

evap/staff/templates/staff_semester_view.html
```diff
@@ -417,7 +417,8 @@ <h3>
                                         <a class="btn btn-sm btn-dark" data-bs-toggle="tooltip"
                                             href="{% url 'staff:course_copy' course.id %}"
                                             title="{% trans 'Copy course' %}">
-                                            <span class="fas fa-copy"></span>                                            </a>
+                                            <span class="fas fa-copy"></span>
+                                            </a>
```
richardebeling:
```suggestion
                                        </a>
```

evap/staff/views.py
```diff
@@ -1040,6 +1040,41 @@ def course_copy(request, course_id):
...
+@manager_required
+def create_exam_evaluation(request):
+    evaluation = get_object_from_dict_pk_entry_or_logged_40x(Evaluation, request.POST, "evaluation_id")
+    exam_date = datetime.today()  # request.POST.get("date")
+    if evaluation.is_single_result:
+        raise SuspiciousOperation("Creating an exam evaluation for a single result evaluation is not allowed")
+
+    if evaluation.course.evaluations.filter(name_de="Klausur", name_en="Exam").exists():
+        messages.error(request, _("An exam evaluation already exists for this course."))
+        return HttpResponse()  # 200 OK
```
richardebeling, line 1053: I think we should return a `SuspiciousOperation` here and the frontend should disable the button and add a tooltip about the exam evaluation already existing.

evap/staff/views.py
```diff
@@ -1040,6 +1040,41 @@ def course_copy(request, course_id):
...
+    )
+    exam_evaluation.vote_start_datetime = datetime.combine(exam_date + timedelta(days=1), time(8, 0))
+    exam_evaluation.vote_end_date = exam_date + timedelta(days=3)
+    exam_evaluation.save()
+    exam_evaluation.participants.set(evaluation.participants.all())
+    for contribution in evaluation.contributions.all().filter(contributor__isnull=False):
+        exam_evaluation.contributions.create(contributor=contribution.contributor)
+    exam_evaluation.general_contribution.questionnaires.set([Questionnaire.objects.get(id=1)])
+
+    evaluation.save()
```
richardebeling, line 1073: Please group the three modifications to `evaluation` and the final `save()` call together.

This should also be in a database transaction. You should be able to just decorate the method with `@transaction.atomic`

evap/staff/views.py
```diff
@@ -1040,6 +1040,41 @@ def course_copy(request, course_id):
...
+    if evaluation.is_single_result:
+        raise SuspiciousOperation("Creating an exam evaluation for a single result evaluation is not allowed")
+
+    if evaluation.course.evaluations.filter(name_de="Klausur", name_en="Exam").exists():
+        messages.error(request, _("An exam evaluation already exists for this course."))
+        return HttpResponse()  # 200 OK
+
+    evaluation.weight = 9
+
+    course_evaluation_end_date = exam_date - timedelta(days=1)
```
janno42, line 1057: it's a bit confusing that the evaluation object is called `evaluation` but for some variables you use the prefix `course_evaluation_`. please use consistent names :)

evap/staff/views.py
```diff
@@ -1040,6 +1040,41 @@ def course_copy(request, course_id):
...
+    if evaluation.vote_start_datetime > course_evaluation_end_date:
+        raise SuspiciousOperation("The selected date is before the start date")
```
janno42, line 1059: please make this more clear, such as "The exam date is before the start date of the main evaluation"

evap/staff/views.py
```diff
@@ -1040,6 +1040,41 @@ def course_copy(request, course_id):
...
+    evaluation.vote_end_date = course_evaluation_end_date
+
+    exam_evaluation = Evaluation(
+        course=evaluation.course, name_de="Klausur", name_en="Exam", weight=1, is_rewarded=False
+    )
+    exam_evaluation.vote_start_datetime = datetime.combine(exam_date + timedelta(days=1), time(8, 0))
+    exam_evaluation.vote_end_date = exam_date + timedelta(days=3)
+    exam_evaluation.save()
+    exam_evaluation.participants.set(evaluation.participants.all())
+    for contribution in evaluation.contributions.all().filter(contributor__isnull=False):
```
janno42, line 1069:
```suggestion
    for contribution in evaluation.contributions.exclude(contributor=None):
```

evap/staff/views.py
```diff
@@ -1040,6 +1040,41 @@ def course_copy(request, course_id):
...
+        exam_evaluation.contributions.create(contributor=contribution.contributor)
+    exam_evaluation.general_contribution.questionnaires.set([Questionnaire.objects.get(id=1)])
```
janno42, line 1071: We can't assume that the questionnaire with `id=1` is the correct questionnaire, it may not even be valid. Please read the specification in the issue:
```
All questionnaires listed in a new settings variable EXAM_QUESTIONNAIRES (initially set to contain the id of the "Miscellaneous" questionnaire) will be added to the exam evaluation as general questionnaires.
```

janno42 (changes requested):

evap/staff/views.py
```diff
@@ -231,6 +231,9 @@ class Stats:
     degree_stats_with_total = cast(dict[Degree | str, Stats], degree_stats)
     degree_stats_with_total["total"] = total_stats
 
+    for evaluation in evaluations:
+        evaluation.has_exam = evaluation.course.evaluations.filter(name_de="Klausur", name_en="Exam").exists()
```
niklasmohrin, line 235: I wonder if this is the right place to keep the knowledge of "what is an exam evaluation". Maybe `Evaluation` should have `exam_evaluation` helper that returns the queryset of all / the one exam evaluation(s)?

niklasmohrin (commented):
Looks good from a quick skim :+1: I will take a closer look once the modals are done and in use here

evap/staff/views.py
richardebeling: Sounds good. Optimally, we'd also have the same knowledge about how to _make_ an exam evaluation in the same place, but that might not be so easy

richardebeling:
@FSadrieh Modals are merged so you should be un-blocked here now

evap/staff/templates/staff_semester_view.html
```diff
@@ -387,6 +387,25 @@ <h3 class="m-0 me-1">{{ semester.name }}</h3>
...
+                            {% blocktranslate %}
+                            Create an exam evaluation based on this evaluation. This will copy all the participants and contributors from the original evaluation. It will set the weight of the original evaluation to 9 and its end date will be set to the day before the exam.
+                            {% endblocktranslate %}
+                        </span>
+                        <span slot="extra-inputs">
+                            <label for="exam_date">{% translate 'Exam Date:' %}</label>
+                            <input type="date" id="exam_date" name="exam_date" required>
+                        </span>
+
+                        <button slot="show-button" type="button" class="btn btn-sm btn-light"><span class="fas fa-file-pen fa-fw"></span></button>
```
Kakadus, line 405: This button does not have a tooltip text. Maybe something like "Create exam evaluation" is appropriate?

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -225,4 +228,17 @@
     {% endif %}
 </td>
 
+<script>
+    var exam_creation_modal = document.getElementById("exam_creation_modal");
+    var btn = document.querySelectorAll('button[evaluation_id="{{ evaluation.id }}"]')[0];
+    var span = document.getElementsByClassName("close")[0];
```
Kakadus, line 234: btn is only available, if `not evaluation.has_exam`. This causes a javascript error if there are evaluations with exams visible

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -188,6 +188,9 @@
             <a href="{% url 'staff:evaluation_copy' evaluation.id %}" class="btn btn-sm btn-light" data-bs-toggle="tooltip" title="{% translate "Copy" %}">
                 <span class="fas fa-copy"></span>
             </a>
+            {% if not evaluation.has_exam %}
+            <button id="openModalBtn" type="button" class="btn btn-sm btn-light" evaluation_id="{{ evaluation.id }}"><span class="fas fa-file-pen fa-fw"></span></button>
```
Kakadus, line 192: What is the difference to the `show-button` in the form. Is this the right way to use the new modals @niklasmohrin ?

Kakadus (changes requested):

evap/staff/templates/staff_semester_view_evaluation.html
niklasmohrin: No, the `confirmation-modal` element should be placed wherever the button to open that modal is. Then, only the `slot=show-button` element will be shown on the normal page and clicking it automatically triggers the modal to open. You can see this in the other usages throughout the codebase

niklasmohrin: If the button to trigger the modal is far away from the form it triggers, you can use `form=the-id-of-the-other-form` on `confirmation-modal` to link it to the correct form

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -188,6 +188,23 @@
             <a href="{% url 'staff:evaluation_copy' evaluation.id %}" class="btn btn-sm btn-light" data-bs-toggle="tooltip" title="{% translate "Copy" %}">
                 <span class="fas fa-copy"></span>
             </a>
+            {% if not evaluation.has_exam %}
+            <confirmation-modal type="submit" id="exam_creation_modal" name="evaluation_id" value="{{ evaluation.id }}" confirm-button-class="btn-primary" title="{% translate 'Create exam evaluations' %}" form="exam_creation_form_{{ evaluation.id }}">
+                <span slot="title">{% translate 'Create exam evaluations' %}</span>
```
janno42, line 193:
```suggestion
                <span slot="title">{% translate 'Create exam evaluation' %}</span>
```

evap/staff/views.py
```diff
@@ -1051,6 +1051,40 @@ def course_copy(request, course_id):
...
+def create_exam_evaluation(request):
+    evaluation = get_object_from_dict_pk_entry_or_logged_40x(Evaluation, request.POST, "evaluation_id")
+    exam_date = request.POST.get("exam_date")
+    exam_date = datetime.combine(datetime.strptime(exam_date, "%Y-%m-%d"), datetime.min.time())
+
+    if evaluation.is_single_result:
+        raise SuspiciousOperation("Creating an exam evaluation for a single result evaluation is not allowed")
+
+    if evaluation.has_exam:
+        raise SuspiciousOperation("An exam evaluation already exists for this course.")
```
janno42, line 1066:
```suggestion
        raise SuspiciousOperation("An exam evaluation already exists for this course")
```

evap/staff/views.py
```diff
@@ -1051,6 +1051,40 @@ def course_copy(request, course_id):
...
+
+    evaluation_end_date = exam_date - timedelta(days=1)
+    if evaluation.vote_start_datetime > evaluation_end_date:
+        raise SuspiciousOperation("The exam date is before the start date of the main evaluation")
```
janno42, line 1070: this should be a user facing error message and not a `SuspiciousOperation`

janno42 (commented):
not tested yet, will submit another review

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -188,6 +188,23 @@
...
+                <span slot="title">{% translate 'Create exam evaluation' %}</span>
+                <span slot="action-text">{% translate 'Create exam evaluation' %}</span>
+                <span slot="question">
+                    {% blocktranslate %}
+                    Create an exam evaluation based on this evaluation. This will copy all the participants and contributors from the original evaluation. It will set the weight of the original evaluation to 9 and its end date will be set to the day before the exam.
+                    {% endblocktranslate %}
+                </span>
+                <span slot="extra-inputs">
+                    <label for="exam_date">{% translate 'Exam Date:' %}</label>
+                    <input type="date" id="exam_date" name="exam_date" required form="exam_creation_form_{{ evaluation.id }}">
```
janno42, line 202: can we use the bootstrap datetimepicker here?

deployment/localsettings.template.py
```diff
@@ -15,3 +15,6 @@
 
 # Make apache work when DEBUG == False
 ALLOWED_HOSTS = ["localhost", "127.0.0.1"]
+
+# Questionnaires automatically added to exam evaluations
+EXAM_QUESTIONNAIRE_IDS = [83]
```
janno42, line 20: It would be better to create a new "Exam" questionnaire in the test data that is used here.

janno42 (changes requested):

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -188,6 +188,23 @@
...
+                </span>
+
+                <button slot="show-button" type="button" class="btn btn-sm btn-light" title="{% translate 'Create exam evaluation' %}" data-bs-placement="top" data-bs-toggle="tooltip"><span class="fas fa-file-pen fa-fw"></span></button>
```
Kakadus, line 205: Still no tooltip :cry:

Kakadus (changes requested):
looking good :+1:

deployment/localsettings.template.py
niklasmohrin: Shouldn't we come up with something nicer than hardcoding ids in the settings? The best idea I have so far would be to go via the title of the questionnaire, which still feels a bit hacky but better than the id to me

richardebeling: I think I'd prefer the id as its constant. Any other field might change (if only hypothetically in the future after adding forms for that), so using them for identifing objects seems more brittle to me.

niklasmohrin: Fair enough, my point was more about the UX of looking up database ids. We can revisit this later / in another issue though - or not at all if you disagree

richardebeling: I see that having to look up IDs from browser urls is a bit awkward, but then again, editing the localsettings.py requires SSHing into the server and editing the right text file in the right location, so for me, they both rank at "for expert admin who knows what they're doing". I'm also fine with other fields in general, but I'd expect them to cause more problems if they are 1. not unique or 2. mutable. Afaik, the pk is the only unique constant field.

niklasmohrin: I was thinking about something on the website, completely disconnected from the settings. I agree that for our current audience we are fine like this though

evap/staff/tests/test_views.py
```diff
@@ -1924,6 +1924,79 @@ def test_evaluation_copy(self):
         self.assertEqual(copied_evaluation.contributions.count(), 4)
 
 
+class TestEvaluationExamCreation(WebTestStaffMode):
+    csrf_checks = False
+
+    @classmethod
+    def setUpTestData(cls):
+        cls.manager = make_manager()
+        cls.semester = baker.make(Semester)
```
richardebeling, line 1933: not used

evap/staff/tests/test_views.py
```diff
@@ -1924,6 +1924,79 @@ def test_evaluation_copy(self):
...
+        cls.course = baker.make(Course, semester=cls.semester)
+        vote_start_datetime = datetime.datetime.now() - datetime.timedelta(days=50)
+        cls.evaluation = baker.make(Evaluation, course=cls.course, vote_start_datetime=vote_start_datetime)
+        cls.contributions = baker.make(
+            Contribution, evaluation=cls.evaluation, _fill_optional=["contributor"], _quantity=3, _bulk_create=True
+        )
+        cls.questionnaire = baker.make(Questionnaire, pk=83)
```
richardebeling, line 1940: not used

evap/staff/tests/test_views.py
```diff
@@ -1924,6 +1924,79 @@ def test_evaluation_copy(self):
...
+        cls.url = reverse("staff:create_exam_evaluation")
```
richardebeling, line 1941: not dependent on any data in `setUpTestData`, so this can be a class variable

evap/staff/tests/test_views.py
```diff
@@ -1924,6 +1924,79 @@ def test_evaluation_copy(self):
...
+        self.assertEqual(exam_evaluation.contributions.count(), self.evaluation.contributions.count())
+        self.assertEqual(
+            exam_evaluation.vote_start_datetime,
+            datetime.datetime.combine(self.exam_date + datetime.timedelta(days=1), datetime.time(8, 0)),
+        )
+        self.assertEqual(exam_evaluation.vote_end_date, self.exam_date + datetime.timedelta(days=3))
+        self.assertEqual(exam_evaluation.name_de, "Klausur")
+        self.assertEqual(exam_evaluation.name_en, "Exam")
+        self.assertEqual(exam_evaluation.course, self.evaluation.course)
+        self.assertEqual(exam_evaluation.participants.count(), self.evaluation.participants.count())
```
richardebeling, line 1962: `assertQuerySetEqual`?

Also, with the current test setup, I don't think the evaluation has any participants, does it?

evap/staff/tests/test_views.py
```diff
@@ -1924,6 +1924,79 @@ def test_evaluation_copy(self):
...
+        self.assertEqual(exam_evaluation.weight, 1)
```
richardebeling, line 1963: I'd also assert the weight of the existing evaluation

evap/staff/tests/test_views.py
```diff
@@ -1924,6 +1924,79 @@ def test_evaluation_copy(self):
...
+        cls.semester = baker.make(Semester)
+        cls.course = baker.make(Course, semester=cls.semester)
+        vote_start_datetime = datetime.datetime.now() - datetime.timedelta(days=50)
+        cls.evaluation = baker.make(Evaluation, course=cls.course, vote_start_datetime=vote_start_datetime)
+        cls.contributions = baker.make(
+            Contribution, evaluation=cls.evaluation, _fill_optional=["contributor"], _quantity=3, _bulk_create=True
+        )
+        cls.questionnaire = baker.make(Questionnaire, pk=83)
+        cls.url = reverse("staff:create_exam_evaluation")
+        cls.exam_date = (datetime.date.today()) + datetime.timedelta(days=10)
```
richardebeling, line 1942: `datetime` if it's a datetime...

evap/staff/tests/test_views.py
```diff
@@ -1924,6 +1924,79 @@ def test_evaluation_copy(self):
...
+        self.evaluation.vote_start_datetime = datetime.datetime.now() + datetime.timedelta(days=100)
+        self.evaluation.vote_end_date = datetime.date.today() + datetime.timedelta(days=150)
+        self.evaluation.save()
+
+        self.app.post(
+            self.url,
+            user=self.manager,
+            status=200,
+            params={"evaluation_id": self.evaluation.pk, "exam_date": self.exam_date},
+        )
```
richardebeling, line 1996: I think we have a helper to assert that no database modifications were made (`assert_no_database_modifications`). It would probably make sense to use that here

evap/staff/tests/test_views.py
```diff
@@ -1924,6 +1924,79 @@ def test_evaluation_copy(self):
...
+        self.assertFalse(self.evaluation.has_exam)
```
richardebeling, line 1997: Please also assert that the error message is shown

evap/staff/views.py
```diff
@@ -1051,6 +1051,43 @@ def course_copy(request, course_id):
     )
 
 
+@require_POST
+@manager_required
+@transaction.atomic
+def create_exam_evaluation(request):
+    evaluation = get_object_from_dict_pk_entry_or_logged_40x(Evaluation, request.POST, "evaluation_id")
+    exam_date = request.POST.get("exam_date")
```
richardebeling, line 1059: This will cause a server error if the parameter is missing, right? It should be a 4XX (client) error, probably 400 bad request.

evap/staff/views.py
```diff
@@ -1051,6 +1051,43 @@ def course_copy(request, course_id):
...
+    exam_date = datetime.combine(datetime.strptime(exam_date, "%Y-%m-%d"), datetime.min.time())
```
richardebeling, line 1060: If it's a datetime, I'd prefer we also name it `datetime`.

evap/staff/views.py
```diff
@@ -1051,6 +1051,43 @@ def course_copy(request, course_id):
...
+
+    if evaluation.is_single_result:
+        raise SuspiciousOperation("Creating an exam evaluation for a single result evaluation is not allowed")
+
+    if evaluation.has_exam:
+        raise SuspiciousOperation("An exam evaluation already exists for this course")
```
richardebeling, line 1066: Please move these up to the earliest possible place

evap/staff/views.py
```diff
@@ -1051,6 +1051,43 @@ def course_copy(request, course_id):
...
+
+    evaluation_end_date = exam_date - timedelta(days=1)
+    if evaluation.vote_start_datetime > evaluation_end_date:
+        messages.error(
+            request, _("The exam date is before the start date of the main evaluation. No exam evaluation created.")
```
richardebeling, line 1071:
```suggestion
            request, _("The exam date is before the start date of the main evaluation. No exam evaluation was created.")
```

evap/staff/views.py
```diff
@@ -1051,6 +1051,43 @@ def course_copy(request, course_id):
...
+    evaluation.save()
+
+    exam_evaluation = Evaluation(
+        course=evaluation.course, name_de="Klausur", name_en="Exam", weight=1, is_rewarded=False
+    )
+    exam_evaluation.make_exam_evaluation(
+        exam_date=exam_date,
+        participants=evaluation.participants.all(),
+        eval_contributions=evaluation.contributions.exclude(contributor=None),
+    )
```
richardebeling, line 1086: `make` in our codebase is, coming from model_bakery, connotated with "create some new instance".

Have you considered putting all this stuff into a static method `Evaluation::make_exam_evaluation()` that does these steps?

evap/staff/views.py
Kakadus: I think l. 1058  was fine and @richardebeling meant l. 1059, right?

richardebeling: Yup, `get_object_from_dict_pk_entry_or_logged_40x(Evaluation, request.POST, "evaluation_id")` is fine I think.

Keep in mind that `strptime` below will also throw errors, this should also be converted into a bad request.

evap/staff/views.py
```diff
@@ -1051,6 +1051,39 @@ def course_copy(request, course_id):
...
+        exam_datetime = datetime.combine(datetime.strptime(exam_datetime, "%Y-%m-%d"), datetime.min.time())
+    except ValueError:
+        return HttpResponseBadRequest("Exam date missing or invalid.")
+
+    evaluation_end_date = exam_datetime - timedelta(days=1)
+    if evaluation.vote_start_datetime > evaluation_end_date:
+        messages.error(
+            request, _("The exam date is before the start date of the main evaluation. No exam evaluation was created.")
+        )
+        return HttpResponse()
```
richardebeling, line 1075: I don't think the message flow makes sense here (as is reflected in the test for this): The post request will be answered with an empty HTTP 200 OK -- how would a user get that error? To my understanding, they would only see that once they go to a new page, right?

If this error is possible to be triggered by a user, they should get a meaningful error immediately.

(A post-only REST handler setting an error in the messages framework generall seems not really useful to me)

evap/evaluation/models.py
```diff
@@ -443,6 +443,38 @@ class State:
...
 
+    @property
+    def has_exam(self):
+        return self.course.evaluations.filter(name_de="Klausur", name_en="Exam").exists()
+
+    def make_exam_evaluation(
+        self,
+        exam_date: datetime,
+        evaluation_end_date: datetime,
+        participants: query.QuerySet["UserProfile"],
```
richardebeling, line 454: please import and use `QuerySet` directly

evap/evaluation/models.py
```diff
@@ -443,6 +443,38 @@ class State:
         verbose_name=_("wait for grade upload before publishing"), default=True
     )
 
+    @property
+    def has_exam(self):
+        return self.course.evaluations.filter(name_de="Klausur", name_en="Exam").exists()
+
+    def make_exam_evaluation(
```
richardebeling, line 450: What do you think of `create_exam_evaluation`? I think that follows django wording (of Manager.create()) more.

evap/evaluation/models.py
```diff
@@ -443,6 +443,38 @@ class State:
...
+        self,
+        exam_date: datetime,
+        evaluation_end_date: datetime,
+        participants: query.QuerySet["UserProfile"],
+        eval_contributions: query.QuerySet["Contribution"],
```
richardebeling, line 455: why the parameters `participants` and `eval_contributions`? I'd say they are functionally dependent on `exam_evaluation` and defining how they are deduced makes more sense inside this function than in the view (where the argument is currently calculated).

evap/evaluation/models.py
```diff
@@ -443,6 +443,38 @@ class State:
...
+    ):
+        self.weight = 9
+        self.vote_end_date = evaluation_end_date
+        self.save()
+        exam_evaluation = Evaluation(course=self.course, name_de="Klausur", name_en="Exam", weight=1, is_rewarded=False)
+        exam_evaluation.set_exam_evaluation_attributes(exam_date, participants, eval_contributions)
+        exam_evaluation.save()
+
+    def set_exam_evaluation_attributes(
```
richardebeling, line 464: I'd vote to inline, to me it doesn't seem to be a helpful operation on its own

evap/staff/views.py
```diff
@@ -1051,6 +1051,39 @@ def course_copy(request, course_id):
     )
 
 
+@require_POST
+@manager_required
+@transaction.atomic
```
richardebeling, line 1056: I think this can (and probably should) now be moved to the method on the model, since all the database updates happen in there.

richardebeling (changes requested):

evap/evaluation/models.py
```diff
@@ -443,6 +444,40 @@ class State:
...
+    def has_exam(self):
+        return self.course.evaluations.filter(name_de="Klausur", name_en="Exam").exists()
+
+    @transaction.atomic
+    def create_exam_evaluation(
+        self,
+        exam_date: datetime,
+        evaluation_end_date: datetime,
+    ):
+        def _set_exam_evaluation_attributes(
```
richardebeling, line 457: When I said that I'd vote to inline the function, I didn't mean just copying the function definition here, but actually [inline-expanding the function call](https://en.wikipedia.org/wiki/Inline_expansion) :D

evap/staff/templates/staff_semester_view.html
```diff
@@ -377,6 +377,12 @@ <h3 class="m-0 me-1">{{ semester.name }}</h3>
                     </div>
                 </div>
 
+                {% for evaluation in evaluations %}
+                    <form id="exam_creation_form_{{ evaluation.id }}" reload-on-success method="post" action="{% url 'staff:create_exam_evaluation' %}">
```
niklasmohrin, line 381: Given that except for the `id`, all these forms are equal, we really only need one

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -188,6 +188,23 @@
             <a href="{% url 'staff:evaluation_copy' evaluation.id %}" class="btn btn-sm btn-light" data-bs-toggle="tooltip" title="{% translate "Copy" %}">
                 <span class="fas fa-copy"></span>
             </a>
+            {% if not evaluation.has_exam %}
+            <confirmation-modal type="submit" id="exam_creation_modal" name="evaluation_id" value="{{ evaluation.id }}" confirm-button-class="btn-primary" form="exam_creation_form_{{ evaluation.id }}">
```
niklasmohrin, line 192: Please indent the content after the `{% if ... %}`

niklasmohrin, line 192: The `id` of these modals will be duplicated when there is more than one applicable evaluation - I think we don't need the id anyways, in that case let's just remove it

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -188,6 +188,23 @@
...
+                <span slot="title">{% translate 'Create exam evaluation' %}</span>
+                <span slot="action-text">{% translate 'Create exam evaluation' %}</span>
+                <span slot="question">
+                    {% blocktranslate %}
```
niklasmohrin, line 196: We should go with `blocktranslate trimmed`, otherwise the whitespace will be included in the string for the translation

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -188,6 +188,23 @@
...
+                    Create an exam evaluation based on this evaluation. This will copy all the participants and contributors from the original evaluation. It will set the weight of the original evaluation to 9 and its end date will be set to the day before the exam.
```
niklasmohrin, line 197: Also please indent :)

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -188,6 +188,23 @@
...
+                    {% endblocktranslate %}
+                </span>
+                <span slot="extra-inputs">
+                    <label for="exam_date">{% translate 'Exam Date:' %}</label>
+                    <input type="date" id="exam_date" name="exam_date" required form="exam_creation_form_{{ evaluation.id }}">
```
niklasmohrin, line 202: This id will also be duplicated. We could include the evaluation's id, or it should also work to move the `input` element into the `label` element, then we don't need the `id` and `for` attributes (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label). I think I prefer the second approach, we would have to make sure it also works with `type="date"` though :)

evap/evaluation/models.py
```diff
@@ -443,6 +444,40 @@ class State:
...
 
+    @property
+    def has_exam(self):
+        return self.course.evaluations.filter(name_de="Klausur", name_en="Exam").exists()
+
+    @transaction.atomic
+    def create_exam_evaluation(
+        self,
+        exam_date: datetime,
+        evaluation_end_date: datetime,
```
niklasmohrin, line 455: Shouldn't this just be a date instead of a datetime?

niklasmohrin (commented):
some comments before I take a closer look at the view code; confirmation modal usage is mostly correct, only some subtleties :+1:

evap/staff/templates/staff_semester_view_evaluation.html
FSadrieh: I do not think we can remove it. Without id it throws: `ValueError: "<Evaluation: Database Systems I – Exam>" needs to have a value for field "id" before this many-to-many relationship can be used.`

evap/staff/views.py
FSadrieh: @niklasmohrin we settled to check the date when the user submits the modal, similar to the semester deletion modal (See the new commit). One problem we currently face is the error message disappearing fast as the modal closes on click. How would this feature work best with the confirmation modal?

evap/staff/templates/staff_semester_view_evaluation.html
niklasmohrin: Sorry, that wasnt very clear - I was referring to the id of the html element, that is, the `id=exam_evaluation_modal` part

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -188,6 +188,56 @@
...
+                        <label for="exam_date">{% translate 'Exam Date:' %}</label>
+                        <input type="date" id="exam_date" name="exam_date" data-must-before="{{ evaluation.vote_start_datetime }}" class="form-control" required form="exam_creation_form_{{ evaluation.id }}"/>
+                    </div>
+
+                    <button slot="show-button" type="button" class="btn btn-sm btn-light" title="{% translate 'Create exam evaluation' %}" data-bs-placement="top" data-bs-toggle="tooltip"><span class="fas fa-file-pen fa-fw"></span></button>
+
+                    <script type="text/javascript">
+                        document.getElementById('submit_button').addEventListener('click', function(event) {
+                            var inputDate = new Date(document.getElementById('exam_date').value);
+                            var evaluationDate = new Date(document.getElementById('exam_date').dataset.mustBefore);
```
niklasmohrin, line 210: Maybe you have already considered this, but I would think that this validation usually happens in a [`change` event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event) handler on the element itself. So something like

```javascript
document.querySelectorAll("input[type=date][data-must-before]").forEach(input => {
    input.addEventListener("change", () => {
        // check stuff with input.value and input.dataset.mustBefore
        // and call setCustomValidity
    });
});
```

This code should then be placed outside the Django `{% for ... %}` loop so it is not duplicated - the `additional_javascript` block seems appropriate

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -188,6 +188,56 @@
...
+                    <span slot="title">{% translate 'Create exam evaluation' %}</span>
+                    <span slot="action-text" id="submit_button">{% translate 'Create exam evaluation' %}</span>
+                    <span slot="question">
+                        {% blocktranslate trimmed %}
+                            Create an exam evaluation based on this evaluation. This will copy all the participants and contributors from the original evaluation. It will set the weight of the original evaluation to 9 and its end date will be set to the day before the exam.
+                        {% endblocktranslate %}
+                    </span>
+                    <div slot="extra-inputs">
+                        <label for="exam_date">{% translate 'Exam Date:' %}</label>
+                        <input type="date" id="exam_date" name="exam_date" data-must-before="{{ evaluation.vote_start_datetime }}" class="form-control" required form="exam_creation_form_{{ evaluation.id }}"/>
```
niklasmohrin, line 202: The `id="exam_date"` attribute is problematic, because this id will occur multiple times in the HTML, which is forbidden. After the change to the event handler from my last comment, the id is only needed for linking the `input` and `label` elements together - this can also be achieved by moving the input element into the label element, like `<label>Exam Date: <input ...></label>` (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label).

Alternatively, we could mix the evaluation id into it (`exam_date_{{ evaluation.id }}`), but I would prefer the first approach

evap/staff/views.py
niklasmohrin: I opened #2253 to cover this case. There is also another approach described in the comments there, however I am not sure which of the two solutions should be preferred. If you want to move forward here, you can copy the solution as in the semester deletion modal

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -189,6 +189,42 @@
...
+                        </label>
+                    </div>
+
+                    <button slot="show-button" type="button" class="btn btn-sm btn-light" title="{% translate 'Create exam evaluation' %}" data-bs-placement="top" data-bs-toggle="tooltip"><span class="fas fa-file-pen fa-fw"></span></button>
+
+                    <script type="text/javascript">
+                        document.querySelectorAll("[data-evaluation-date]").forEach(element => {
+                            element.addEventListener("change", () => {
+                                console.log("exam_date",element.value);
+                                console.log("evaluation_date",element.dataset.evaluationDate);
```
Kakadus, line 214: do we want to keep these log messages?

evap/staff/tests/test_views.py
```diff
@@ -1923,6 +1923,102 @@ def test_evaluation_copy(self):
...
+    def test_exam_evaluation_for_single_result(self):
+        self.evaluation.is_single_result = True
+        self.evaluation.save()
+        self.app.post(
+            self.url,
+            user=self.manager,
+            status=400,
+            params={"evaluation_id": self.evaluation.pk, "exam_date": self.exam_datetime},
+        )
+        self.assertFalse(self.evaluation.has_exam)
```
richardebeling, line 1977: Have you considered using `with assert_no_database_modifications():`? I think it would fit here (and the tests below)

evap/staff/tests/test_views.py
```diff
@@ -1923,6 +1923,102 @@ def test_evaluation_copy(self):
...
+        self.evaluation.save()
+        self.app.get("", user=self.manager)  # Needed to not get a last login database update
+
+        with assert_no_database_modifications():
+            self.app.post(
+                self.url,
+                user=self.manager,
+                status=400,
+                params={"evaluation_id": self.evaluation.pk, "exam_date": self.exam_datetime},
+            )
```
richardebeling, line 2001: The parameters are duplicated across the tests -- this could also all just be one tests that defined all those "base" arguments once -- or maybe have these parameters stored in the class or something

richardebeling (changes requested):

evap/staff/templates/staff_semester_view.html
niklasmohrin: We concluded in person that this is needed because the date-selection inputs in all the modals should be tied to different forms. Please document this decision in a comment here though :)

evap/staff/tests/test_views.py
```diff
@@ -1923,6 +1923,102 @@ def test_evaluation_copy(self):
...
+        cls.manager = make_manager()
+        cls.course = baker.make(Course)
+        vote_start_datetime = datetime.datetime.now() - datetime.timedelta(days=50)
+        cls.evaluation = baker.make(Evaluation, course=cls.course, vote_start_datetime=vote_start_datetime)
+        cls.contributions = baker.make(
+            Contribution, evaluation=cls.evaluation, _fill_optional=["contributor"], _quantity=3, _bulk_create=True
+        )
+        cls.exam_datetime = (datetime.date.today()) + datetime.timedelta(days=10)
+        cls.semester_overview_url = reverse("staff:semester_view", args=[cls.course.semester.pk])
+        # Default the evaluation does not have any participants, so we need to add some
```
janno42, line 1941:
```suggestion
        # By default the evaluation does not have any participants, so we need to add some
```

evap/staff/views.py
```diff
@@ -1051,6 +1051,38 @@ def course_copy(request, course_id):
...
+        return HttpResponseBadRequest("Exam date missing or invalid.")
+
+    evaluation_end_date = exam_datetime - timedelta(days=1)
+    if evaluation.vote_start_datetime > evaluation_end_date:
+        messages.error(
+            request, _("The exam date is before the start date of the main evaluation. No exam evaluation was created.")
+        )
+        raise SuspiciousOperation(
+            "The exam date is before the start date of the main evaluation. No exam evaluation was created."
+        )
```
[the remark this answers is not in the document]
janno42: in this case probably `SuspiciousOperation` because the UI doesn't allow invalid values

janno42 (changes requested):

evap/staff/templates/staff_semester_view.html
richardebeling: Unresolved, don't see a comment

> Please document this decision in a comment here though :)

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -189,6 +189,40 @@
...
+                    <span slot="action-text" id="submit_button">{% translate 'Create exam evaluation' %}</span>
+                    <span slot="question">
+                        {% blocktranslate trimmed %}
+                            Create an exam evaluation based on this evaluation. This will copy all the participants and contributors from the original evaluation. It will set the weight of the original evaluation to 9 and its end date will be set to the day before the exam.
+                        {% endblocktranslate %}
+                    </span>
+                    <div slot="extra-inputs">
+                        <label>
+                            {% translate 'Exam Date:' %}
+                            <input type="date" name="exam_date" data-earliest-legal-exam-evaluation-date="{{ evaluation.vote_start_datetime }}" class="form-control" required form="exam_creation_form_{{ evaluation.id }}"/>
```
richardebeling, line 204: `data-earliest-legal-exam-evaluation-date="{{ evaluation.vote_start_datetime }}"` seems semantically incorrect: When creating the exam evaluation, the vote_end_date is set to the exam_date minus 1 day, so we would get an evaluation that ends before it starts here.

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -189,6 +189,40 @@
...
+                        </label>
+                    </div>
+
+                    <button slot="show-button" type="button" class="btn btn-sm btn-light" title="{% translate 'Create exam evaluation' %}" data-bs-placement="top" data-bs-toggle="tooltip"><span class="fas fa-file-pen fa-fw"></span></button>
+
+                    <script type="text/javascript">
+                        document.querySelectorAll("[data-evaluation-date]").forEach(element => {
+                            element.addEventListener("change", () => {
+                                if (element.value <= element.dataset.earliestLegalExamEvaluationDate) {
+                                    element.setCustomValidity("{% translate 'The exam date is before the start date of the main evaluation.' %}");
```
richardebeling, line 214: This isn't the semantics I intended with the naming "earliest legal".

If "earliest legal day" is 01.01.2010, then the user is allowed to choose 01.01.2010.

evap/staff/tests/test_views.py
```diff
@@ -1923,6 +1923,99 @@ def test_evaluation_copy(self):
...
+        cls.manager = make_manager()
+        cls.course = baker.make(Course)
+        vote_start_datetime = datetime.datetime.now() - datetime.timedelta(days=50)
+        cls.evaluation = baker.make(Evaluation, course=cls.course, vote_start_datetime=vote_start_datetime)
+        cls.contributions = baker.make(
+            Contribution, evaluation=cls.evaluation, _fill_optional=["contributor"], _quantity=3, _bulk_create=True
+        )
+        cls.exam_datetime = datetime.date.today() + datetime.timedelta(days=10)
+        cls.semester_overview_url = reverse("staff:semester_view", args=[cls.course.semester.pk])
+        # By default the evaluation does not have any participants, so we need to add some
```
richardebeling, line 1941:
```suggestion
```

evap/staff/tests/test_views.py
```diff
@@ -1923,6 +1923,99 @@ def test_evaluation_copy(self):
...
+    def setUpTestData(cls):
+        cls.manager = make_manager()
+        cls.course = baker.make(Course)
+        vote_start_datetime = datetime.datetime.now() - datetime.timedelta(days=50)
+        cls.evaluation = baker.make(Evaluation, course=cls.course, vote_start_datetime=vote_start_datetime)
+        cls.contributions = baker.make(
+            Contribution, evaluation=cls.evaluation, _fill_optional=["contributor"], _quantity=3, _bulk_create=True
+        )
+        cls.exam_datetime = datetime.date.today() + datetime.timedelta(days=10)
+        cls.semester_overview_url = reverse("staff:semester_view", args=[cls.course.semester.pk])
```
richardebeling, line 1940: unused

evap/staff/tests/test_views.py
```diff
@@ -1923,6 +1923,99 @@ def test_evaluation_copy(self):
...
+    @classmethod
+    def setUpTestData(cls):
+        cls.manager = make_manager()
+        cls.course = baker.make(Course)
+        vote_start_datetime = datetime.datetime.now() - datetime.timedelta(days=50)
+        cls.evaluation = baker.make(Evaluation, course=cls.course, vote_start_datetime=vote_start_datetime)
+        cls.contributions = baker.make(
+            Contribution, evaluation=cls.evaluation, _fill_optional=["contributor"], _quantity=3, _bulk_create=True
+        )
+        cls.exam_datetime = datetime.date.today() + datetime.timedelta(days=10)
```
richardebeling, line 1939: `datetime` and `date` are different, please name correctly.

evap/staff/tests/test_views.py
```diff
@@ -1923,6 +1923,99 @@ def test_evaluation_copy(self):
...
+        self.evaluation.is_single_result = True
+        self.evaluation.save()
+        self.app.get("", user=self.manager)  # Needed to not get a last login database update
+        with assert_no_database_modifications():
+            self.app.post(
+                self.url,
+                user=self.manager,
+                status=400,
+                params=self.params,
+            )
```
richardebeling, line 1979:
```suggestion
            self.app.post(self.url, user=self.manager, status=400, params=self.params)
```
(the trailing comma makes black format it in one-line-per-argument fashion)
(repeated below)

evap/staff/tests/test_views.py
```diff
@@ -1923,6 +1923,99 @@ def test_evaluation_copy(self):
...
+
+    def test_exam_evaluation_for_already_existing_exam_evaluation(self):
+        baker.make(Evaluation, course=self.course, name_en="Exam", name_de="Klausur")
+        self.assertTrue(self.evaluation.has_exam)
+        self.app.get("", user=self.manager)  # Needed to not get a last login database update
```
richardebeling, line 1984: Instead of duplicating this line 5 times, can we maybe simply allow the one SQL statements that updates the last login in `assert_no_database_modifications`?

richardebeling (changes requested):

evap/staff/templates/staff_semester_view.html
FSadrieh: As Niklas said the code is needed and his initial comment is not relevant anymore

evap/evaluation/models.py
```diff
@@ -445,6 +445,29 @@ class State(models.IntegerChoices):
...
 
+    @property
+    def has_exam(self):
+        return self.course.evaluations.filter(name_de="Klausur", name_en="Exam").exists()
+
+    @transaction.atomic
+    def create_exam_evaluation(
+        self,
+        exam_date: date,
+    ):
```
richardebeling, line 456:
```suggestion
    def create_exam_evaluation(self, exam_date: date):
```

If you leave a trailing comma, our code formatter interprets that as "please always list these vertically", but we generall don't want to do that when everything fits in one line.

evap/staff/templates/staff_semester_view.html
richardebeling: Please add a short comment in the code, something like
```django
{# separate forms for each modal since we want separate date-selection inputs because REASON #}
```

evap/staff/urls.py
```diff
@@ -63,7 +64,6 @@
     path("questionnaire/update_indices", views.questionnaire_update_indices, name="questionnaire_update_indices"),
     path("questionnaire/questionnaire_visibility", views.questionnaire_visibility, name="questionnaire_visibility"),
     path("questionnaire/questionnaire_set_locked", views.questionnaire_set_locked, name="questionnaire_set_locked"),
-
```
richardebeling, line 66: Please undo

evap/staff/tests/test_views.py
```diff
@@ -1923,6 +1923,71 @@ def test_evaluation_copy(self):
...
+    def setUpTestData(cls):
+        cls.manager = make_manager()
+        cls.course = baker.make(Course)
+        vote_start_datetime = datetime.datetime.now() - datetime.timedelta(days=50)
+        cls.evaluation = baker.make(Evaluation, course=cls.course, vote_start_datetime=vote_start_datetime)
+        cls.contributions = baker.make(
+            Contribution, evaluation=cls.evaluation, _fill_optional=["contributor"], _quantity=3, _bulk_create=True
+        )
+        cls.exam_date = datetime.date.today() + datetime.timedelta(days=10)
+        cls.evaluation.participants.set(baker.make(UserProfile, _quantity=3))
```
richardebeling, line 1940: For readability, I'd move this line below the `baker.make(Evaluation, ...)` line. To me, this is a modification of the evaluation, while the contributions stuff is (more) separate from the evaluation.

evap/staff/tests/test_views.py
```diff
@@ -1923,6 +1923,71 @@ def test_evaluation_copy(self):
...
+        cls.params = {"evaluation_id": cls.evaluation.pk, "exam_date": cls.exam_date}
+
+    def test_create_exam_evaluation(self):
+        self.app.post(
+            self.url,
+            user=self.manager,
+            status=200,
+            params=self.params,
+        )
```
richardebeling, line 1949:
```suggestion
        self.app.post(self.url, user=self.manager, status=200, params=self.params)
```

evap/staff/tests/test_views.py
```diff
@@ -1923,6 +1923,71 @@ def test_evaluation_copy(self):
...
+    def test_exam_evaluation_with_wrong_date(self):
+        self.evaluation.vote_start_datetime = datetime.datetime.now() + datetime.timedelta(days=100)
+        self.evaluation.vote_end_date = datetime.date.today() + datetime.timedelta(days=150)
+        self.evaluation.save()
+        with assert_no_database_modifications():
+            self.app.post(self.url, user=self.manager, status=400, params=self.params)
+
+    def test_exam_evaluation_with_missing_date(self):
+        with assert_no_database_modifications():
+            self.app.post(self.url, user=self.manager, status=400, params=self.params)
```
richardebeling, line 1988: This seems wrong, `self.params` includes a date.

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -189,6 +189,40 @@
...
+                                } else {
+                                    element.setCustomValidity("");
+                                }
+
+                            });
+                            element.dispatchEvent(new Event("change"));
+                        });
+                    </script>
+
+                </confirmation-modal>
```
richardebeling, line 224:
```suggestion
                                }
                            });
                            element.dispatchEvent(new Event("change"));
                        });
                    </script>
                </confirmation-modal>
```

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -189,6 +189,40 @@
...
+                        <label>
+                            {% translate 'Exam Date:' %}
+                            <input type="date" name="exam_date" data-earliest-legal-exam-date="{{ evaluation.vote_start_datetime.date() + datetime.timedelta(days=1) }}" class="form-control" required form="exam_creation_form_{{ evaluation.id }}"/>
+                        </label>
+                    </div>
+
+                    <button slot="show-button" type="button" class="btn btn-sm btn-light" title="{% translate 'Create exam evaluation' %}" data-bs-placement="top" data-bs-toggle="tooltip"><span class="fas fa-file-pen fa-fw"></span></button>
+
+                    <script type="text/javascript">
+                        document.querySelectorAll("[data-evaluation-date]").forEach(element => {
```
richardebeling, line 211: Is this encapsulated in some kind of shadow root, or does this find all modals on the page, and thus create O(n^2) event listeners?

(Maybe, if we have to think about it, just make it `query` just inside the current `confirmation-modal` anyway?)

(`data-evaluation-date` seems outdated, should probably be `data-earliest-lega-exam-date` now?)

richardebeling (changes requested):

evap/evaluation/models.py
```diff
@@ -468,6 +468,30 @@ class State(models.IntegerChoices):
         verbose_name=_("wait for grade upload before publishing"), default=True
     )
 
+    @property
+    def has_exam(self):
+        return self.course.evaluations.filter(name_de="Klausur", name_en="Exam").exists()
+
+    @property
+    def get_earliest_legal_exam_date(self):
```
niklasmohrin, line 476: Let's remove "get". Is there anything "legal" to our definition here? Otherwise, I would prefer
```suggestion
    def earliest_possible_exam_date(self):
```

evap/evaluation/models.py
```diff
@@ -468,6 +468,30 @@ class State(models.IntegerChoices):
         verbose_name=_("wait for grade upload before publishing"), default=True
     )
 
+    @property
+    def has_exam(self):
```
niklasmohrin, line 472: @richardebeling Do we prefer this over `has_exam_evaluation`? (same for `earliest_possible...`); for creation, we also didn't call it `create_exam`. I think I prefer `exam_evaluation`

evap/evaluation/templates/base.html
```diff
@@ -211,5 +211,18 @@
         </script>
 
         {% block additional_javascript %}{% endblock %}
+
+        <script type="text/javascript"> 
+            document.querySelectorAll("[data-earliest-legal-date]").forEach(element => {
```
niklasmohrin, line 216: I would refrain from "legal" again, how about something like `data-minimum-date` and `data-minimum-date-reason`?

evap/evaluation/models.py
FSadrieh: I am ok with using has_exam_evaluation. I like it more than exam_evaluation as you do not get the evaluation object

evap/staff/views.py
```diff
@@ -1093,6 +1093,31 @@ def course_copy(request, course_id):
...
+        exam_date = datetime.strptime(exam_date, "%Y-%m-%d").date()
+    except TypeError:
+        return HttpResponseBadRequest("Exam date missing or invalid.")
+
+    if exam_date < evaluation.vote_start_datetime.date() + timedelta(days=1):
+        raise SuspiciousOperation(
+            "The end date of the main evaluation would be before its start date. No exam evaluation was created."
+        )
+
+    evaluation.create_exam_evaluation(exam_date)
```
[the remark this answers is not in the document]
FSadrieh: I do not have a strong opinion on where the check should be, but kinda like it here as this case should never happen from user interaction and it is really not an interal server error.

evap/evaluation/models.py
niklasmohrin: Ah yeah, sorry, by "I think I prefer `exam_evaluation`, I meant that I prefer to use "exam_evaluation" over just "exam" for everything, but still with "has_" and "create_".

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -187,6 +187,26 @@
             <a href="{% url 'staff:evaluation_copy' evaluation.id %}" class="btn btn-sm btn-light" data-bs-toggle="tooltip" title="{% translate 'Copy' %}">
                 <span class="fas fa-copy"></span>
             </a>
+            {% if not evaluation.has_exam %}
+                <confirmation-modal type="submit" name="evaluation_id" value="{{ evaluation.id }}" confirm-button-class="btn-primary" form="exam_creation_form_{{ evaluation.id }}">
+                    <span slot="title">{% translate 'Create exam evaluation' %}</span>
+                    <span slot="action-text" id="submit_button">{% translate 'Create exam evaluation' %}</span>
+                    <span slot="question">
+                        {% blocktranslate trimmed %}
+                            Create an exam evaluation based on this evaluation. This will copy all the participants and contributors from the original evaluation. It will set the weight of the original evaluation to 9 and its end date will be set to the day before the exam.
```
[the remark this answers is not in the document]
niklasmohrin: Fair, I think I agree; I would still like to hear if @janno42 and @richardebeling have opinions on this to wrap this up

evap/staff/views.py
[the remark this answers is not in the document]
niklasmohrin: Yeah looking at it today again, I also think that internal server error would be wrong. One argument for checking it inside `create_exam_evaluation` is that we explicitly made everything inside `create_exam_evaluation` in a `transaction.atomic`, but the check here is outside of that. I think the correct-est way would be something like
```python
# Model code:
def create_exam_evaluation(self, exam_date):
    if exam_date < self.earliest_possible_exam_date:
        raise ValueError("...")
    ...

# View code:
def create_exam_evaluation(request):
    ...
    try:
        exam_evaluation = evaluation.create_exam_evaluation(...)
    except ValueError as e:
        raise SuspiciousOperation from e
```
What do you think? Also, thoughts @richardebeling @Kakadus ?

evap/staff/templates/staff_semester_view_evaluation.html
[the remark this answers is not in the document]
richardebeling: Hardcoded until we need to change it the first time is fine with me.

evap/staff/views.py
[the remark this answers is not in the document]
richardebeling: View-Validation here is fine for me. The field has a [constraint](https://github.com/e-valuation/EvaP/blob/6d7970e23b4c51eea2063b8ceb681ec9398edc2f/evap/evaluation/models.py#L489) that ensures we can not store "wrong" data. Attempting to do that would correctly cause an internal server error. A complying view checks/handles this, and instead returns a client error.

I wouldn't do `raise ValueError` in `create_exam_evaluation`, that would be a third check, seems overly complicated to me. The view could in theory try to catch a constraint validation error, but I think I like the current approach most.

Kakadus (approved):

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -187,6 +187,26 @@
             <a href="{% url 'staff:evaluation_copy' evaluation.id %}" class="btn btn-sm btn-light" data-bs-toggle="tooltip" title="{% translate 'Copy' %}">
                 <span class="fas fa-copy"></span>
             </a>
+            {% if not evaluation.has_exam %}
```
niklasmohrin, line 190:
```suggestion
            {% if not evaluation.has_exam_evaluation %}
```

evap/evaluation/models.py
```diff
@@ -470,6 +470,30 @@ class State(models.IntegerChoices):
...
+
+    @transaction.atomic
+    def create_exam_evaluation(self, exam_date: date):
+        self.weight = 9
+        self.vote_end_date = exam_date - timedelta(days=1)
+        self.save()
+        exam_evaluation = Evaluation(course=self.course, name_de="Klausur", name_en="Exam", weight=1, is_rewarded=False)
+        exam_evaluation.vote_start_datetime = datetime.combine(exam_date + timedelta(days=1), time(8, 0))
+        exam_evaluation.vote_end_date = exam_date + timedelta(days=3)
+        exam_evaluation.save()
```
niklasmohrin, line 489:
```suggestion
        exam_evaluation = Evaluation(
            course=self.course,
            name_de="Klausur",
            name_en="Exam",
            weight=1,
            is_rewarded=False,
            vote_start_datetime=datetime.combine(exam_date + timedelta(days=1), time(8, 0)),
            vote_end_date=exam_date + timedelta(days=3),
        )
        exam_evaluation.save()
```

niklasmohrin (changes requested):

evap/staff/views.py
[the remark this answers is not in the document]
niklasmohrin: I don't mean to add a third check, I would just move all checks into the transaction and close to the creation. Main motivation for me is "parse, don't validate". It would then look like this:
```python
class Evaluation:
    # ...

    @transaction.atomic
    def create_exam_evaluation(self, exam_date: date):
        if self.is_single_result:
            raise ValueError("Creating an exam evaluation for a single result evaluation is not allowed")
        if self.has_exam_evaluation:
            raise ValueError("An exam evaluation already exists for this course")
        if exam_date < self.earliest_possible_exam_date:
            raise ValueError("The end date of the main evaluation would be before its start date.")

        self.weight = 9
        self.vote_end_date = exam_date - timedelta(days=1)
        self.save()
        exam_evaluation = Evaluation(
            course=self.course,
            name_de="Klausur",
            name_en="Exam",
            weight=1,
            is_rewarded=False,
            vote_start_datetime=datetime.combine(exam_date + timedelta(days=1), time(8, 0)),
            vote_end_date=exam_date + timedelta(days=3),
        )
        exam_evaluation.save()

        exam_evaluation.participants.set(self.participants.all())
        for contribution in self.contributions.exclude(contributor=None):
            exam_evaluation.contributions.create(contributor=contribution.contributor)
        exam_evaluation.general_contribution.questionnaires.set(settings.EXAM_QUESTIONNAIRE_IDS)
        exam_evaluation.save()

# views.py
@require_POST
@manager_required
def create_exam_evaluation(request: HttpRequest) -> HttpResponse:
    evaluation = get_object_from_dict_pk_entry_or_logged_40x(Evaluation, request.POST, "evaluation_id")
    exam_date_string = request.POST.get("exam_date", "")

    try:
        exam_date = datetime.strptime(exam_date_string, "%Y-%m-%d").date()
        evaluation.create_exam_evaluation(exam_date)
    except ValueError as e:
        raise SuspiciousOperation("Failed to create exam evaluation") from e

    messages.success(request, _("Successfully created exam evaluation."))
    return HttpResponse()  # 200 OK
```
@richardebeling Which do you prefer? I like the above, but if you are still with the current one, I can accept also accept that

evap/staff/tests/test_views.py
```diff
@@ -1960,6 +1960,71 @@ def test_evaluation_copy(self):
...
+
+    def test_create_exam_evaluation(self):
+        self.app.post(self.url, user=self.manager, status=200, params=self.params)
+        self.assertEqual(Evaluation.objects.count(), 2)
+        exam_evaluation = Evaluation.objects.exclude(pk=self.evaluation.pk).get()
+        self.assertEqual(exam_evaluation.contributions.count(), self.evaluation.contributions.count())
+        self.assertEqual(
+            exam_evaluation.vote_start_datetime,
+            datetime.datetime.combine(self.exam_date + datetime.timedelta(days=1), datetime.time(8, 0)),
+        )
```
richardebeling, line 1992: Can you add the analogous assertion about `self.evaluation.vote_end_date`?

evap/staff/views.py
```diff
@@ -1093,6 +1093,34 @@ def course_copy(request, course_id):
...
+    if evaluation.has_exam_evaluation:
+        raise SuspiciousOperation("An exam evaluation already exists for this course")
+
+    exam_date_string = request.POST.get("exam_date")
+    if not exam_date_string:
+        return HttpResponseBadRequest("Exam date missing.")
+    try:
+        exam_date = datetime.strptime(exam_date_string, "%Y-%m-%d").date()
+    except ValueError:
+        return HttpResponseBadRequest("Exam date invalid.")
```
richardebeling, line 1112: Currently not covered. You could combine the two error cases here, a missing exam date is also an invalid exam date, so I'd just use `request.POST.get("exam_date", "")` above

evap/evaluation/templates/base.html
```diff
@@ -211,5 +211,18 @@
...
+                element.addEventListener("change", () => {
+                    if (element.value < element.dataset.earliestPossibleDate) {
+                        element.setCustomValidity(element.dataset.earliestPossibleDateReason);
+                    } else {
+                        element.setCustomValidity("");
+                    }
+                });
+                element.dispatchEvent(new Event("change"));
+            });
+        </script>
```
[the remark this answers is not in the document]
FSadrieh: @niklasmohrin and I talked about putting the check into the base.html to not replicate it for each evaluation and to use it for future date selection modals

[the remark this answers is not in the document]
niklasmohrin: Yes, I thought that if we need to move it away from the use-site, we might as well decouple the names from the "exam date" concept and make it a generic tool. As such, base.html seemed to be the right place. However, I think it could be in the same javascript block as the `setupForms`, as the two procedures are similar.

I don't think a comment is needed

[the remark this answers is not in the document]
richardebeling: Hmm, ok, fine with me

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -187,6 +187,25 @@
...
+                    <span slot="action-text">{% translate 'Create exam evaluation' %}</span>
+                    <span slot="question">
+                        {% blocktranslate trimmed %}
+                            Create an exam evaluation based on this evaluation. This will copy all the participants and contributors from the original evaluation. It will set the weight of the original evaluation to 9 and its end date will be set to the day before the exam.
+                        {% endblocktranslate %}
+                    </span>
+                    <div slot="extra-inputs">
+                        <label>
+                            {% translate 'Exam Date:' %}
+                            <input type="date" name="exam_date" data-earliest-possible-date="{{ evaluation.earliest_possible_exam_date }}" data-earliest-possible-date-reason="{% translate 'The end date of the main evaluation would be before its start date.' %}" class="form-control" required form="exam_creation_form_{{ evaluation.id }}"/>
```
Kakadus, line 202: I just found out: We can directly add a `min` field here, so that the user is not even tempted to select invalid values, as described in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date. `data-earliest-possible-date-reason` would then become maybe obsolete

Kakadus (approved):
image

I personally find the date constraint good, so that the user is not lost in years back and so forth. We also validate it in the backend.

janno42 (approved):

evap/staff/views.py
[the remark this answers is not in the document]
FSadrieh: @niklasmohrin is this resolved for you?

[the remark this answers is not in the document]
niklasmohrin: Yeah, you can close it

evap/evaluation/models.py
```diff
@@ -470,6 +470,35 @@ class State(models.IntegerChoices):
...
+            weight=1,
+            is_rewarded=False,
+            vote_start_datetime=datetime.combine(exam_date + timedelta(days=1), time(8, 0)),
+            vote_end_date=exam_date + timedelta(days=3),
+        )
+        exam_evaluation.save()
+
+        exam_evaluation.participants.set(self.participants.all())
+        for contribution in self.contributions.exclude(contributor=None):
+            exam_evaluation.contributions.create(contributor=contribution.contributor)
```
niklasmohrin, line 499: @richardebeling Shouldn't we also copy things like `role`, `textanswer_visibility`, etc. here, or should this just be adjusted by managers afterwards?

niklasmohrin (commented):
Looks all good, only one concern left before I want to approve

evap/evaluation/models.py
janno42: These values should not be copied; editors have to decide about the permissions here as well.

niklasmohrin (approved):

richardebeling:
🎉

evap/staff/templates/staff_semester_view_evaluation.html
niklasmohrin: @jooooosef noticed that this is still the default date picker

PR #2051, opened 2023-10-23
Introduce ruff to ci

Kakadus:


evap/locale/de/formats.py
```diff
@@ -1 +1 @@
-from evap.locale.en.formats import *  # pylint: disable=unused-wildcard-import,wildcard-import
+from evap.locale.en.formats import *  # pylint: disable=unused-wildcard-import,wildcard-import  # noqa: F403
```
richardebeling, line 1: I'd tend to not put (and thus maintain) multiple ignores for "the same thing" in the codebase. Initially, my idea was to completely remove pylint (with the intention of getting rid of the long runtime).

If ruff is still lacking features, we can run both, but then I'd try to configure them in such a way that pylint doesn't check for anything that ruff also checks for. (I don't know how much configuration that requires though. I'm not interested in maintaining hundreds of lines of linter configuration either).

README.md
```diff
@@ -53,7 +53,7 @@ Before committing, run the following commands:
 
 or, to combine all three, simply run `./manage.py precommit`.
 
-You can also set up `pylint`, `isort`, `black` and `prettier` in your IDE to avoid doing this manually all the time.
+You can also set up `ruff`, `isort`, `black` and `prettier` in your IDE to avoid doing this manually all the time.
```
richardebeling, line 56: The changes here still assume a replacement of pylint, but I think you went with running both now.

test.py
richardebeling, line 1: accidentally added

evap/evaluation/models.py
```diff
@@ -1245,6 +1245,9 @@ class Choices(NamedTuple):
     names: list[StrOrPromise]
     is_inverted: bool
 
+    def zipped(self):
```
richardebeling, line 1248: maybe `as_name_color_value_tuples`?

Kakadus: tracked in #2064

Kakadus:
rebased and cleaned the commits a bit. #2063 and #2064 are no longer part of this

evap/locale/de/formats.py
Kakadus: in [3de4519](https://github.com/e-valuation/EvaP/pull/2051/commits/3de451916f54fda5139a303efe7fc3d099244777), I disabled three pylint checks in favor of ruff. There should be no cases where of both ruff and pylint is disabled

README.md
Kakadus: yes, but I'm still undecided. What do you think?

richardebeling: Proposal for migration:
* Add necessary configuration for ruff
* Additionally run ruff in CI, but keep pylint as the "official" linter (-> revert changes to the README here)
* monitor what they find in CI and check if we'd miss stuff if we disabled pylint
* -> If we don't, we can disable pylint some time later.

evap/evaluation/migrations/0100_clear_evaluation_names.py
```diff
@@ -22,7 +22,7 @@ def name_evaluations(apps, _schema_editor):
             evaluation.name_en = course.name_en
             evaluation.save()
         else:
-            for i in range(0, course.evaluations.count()):
+            for i in range(course.evaluations.count()):
                 evaluation = Evaluation.objects.get(pk=course.evaluations.all()[i].pk)
```
richardebeling, line 26: Never expected to see such cursed code in EvaP. Probably doesn't really matter since its a rather old migration.

evap/evaluation/models.py
```diff
@@ -444,7 +444,7 @@ class State:
     )
 
     class TextAnswerReviewState(Enum):
-        do_not_call_in_templates = True  # pylint: disable=invalid-name
+        do_not_call_in_templates = True
```
richardebeling, line 447: Was the "disable" statement effectless? I would have expected pylint to fail on lines with a disable comment where the listed check does not actually trigger, so this should have made CI fail.

evap/evaluation/models_logging.py
```diff
@@ -128,6 +128,19 @@ class LoggedModel(models.Model):
     class Meta:
         abstract = True
 
+    def save(self, *args, **kw):
```
richardebeling, line 131: Why was the order changed here? Does ruff expect "public" methods before "private" methods, or what is the criterion?

evap/locale/de/formats.py
```diff
@@ -1 +1 @@
-from evap.locale.en.formats import *  # pylint: disable=unused-wildcard-import,wildcard-import
+from evap.locale.en.formats import *  # noqa: F403
```
richardebeling, line 1: Is there any way we can have more readable ignored with ruff? `unused-wildcard-import` says a lot more to me than F403. I guess `undefined-local-with-import-star` should work here?

evap/results/exporters.py
```diff
@@ -71,11 +71,12 @@ def init_grade_styles(cls):
                 "This can happen, if the file is imported / run multiple "
                 "times in one application run.",
                 ImportWarning,
+                stacklevel=2,
```
richardebeling, line 74: I guess this triggered [no-explicit-stacklevel (B028)](https://docs.astral.sh/ruff/rules/no-explicit-stacklevel/).

I'm a bit confused about this rule. The rule description (linked above) says:
> The warnings.warn method uses a stacklevel of 1 by default, which limits the rendered stack trace to that of the line on which the warn method is called.
> 
> It's recommended to use a stacklevel of 2 or higher, give the caller more context about the warning.

which sounds like "this is the backtrace depth that you will get printed out", while [the documentation of warnings.warn](https://docs.python.org/3/library/warnings.html#warnings.warn) says:
> The stacklevel argument can be used by wrapper functions written in Python, like this:
> 
> ```
> def deprecated_api(message):
>   warnings.warn(message, DeprecationWarning, stacklevel=2)
> ```
> This makes the warning refer to deprecated_api’s caller, rather than to the source of deprecated_api itself (since the latter would defeat the purpose of the warning message).

which sounds to me as if "the warning trigger point will be reported as if the warning was raised N levels higher in the stacktrace".

evap/settings.py
```diff
@@ -407,12 +407,10 @@ def CHARACTER_ALLOWED_IN_NAME(character):  # pylint: disable=invalid-name
 # Create a localsettings.py if you want to locally override settings
 # and don't want the changes to appear in 'git status'.
 try:
-    # if a localsettings file exists (vagrant), this will cause wildcard-import errors
-    # if it does not, (GitHub), it would cause useless-suppression
-    # pylint: disable=unused-wildcard-import,wildcard-import,useless-suppression
+    # localsettings file may exist (vagrant) or do not (GitHub)
```
richardebeling, line 410: This comment may now have confusing grammar or do not.

```suggestion
    # localsettings file may (vagrant) or may not (CI run) exist
```

pyproject.toml
```diff
@@ -22,6 +22,33 @@ extend_skip_glob = ["**/migrations/*"]
...
+    "COM812",                # incompatible with formatter
+    "PERF203",
+]
+
+ignore-init-module-imports = true
+pep8-naming.extend-ignore-names = ["assert*", "*Formset"]  # custom assert methods use camelCase; Formsets use PascalCase
+
+[tool.ruff.per-file-ignores]
+"**/migrations/*.py" = ["N806"]  # migrations have model classes as local variables, we use PascalCase for these
+"**/commands/*.py" = ["T201"]  # printing in commands is ok
```
richardebeling, line 43: Why does github render this with full-red background? :D

pyproject.toml
```diff
@@ -22,6 +22,33 @@ extend_skip_glob = ["**/migrations/*"]
...
+"deployment/*.py" = ["INP001"]  # not intendet for direct import
```
richardebeling, line 44:
```suggestion
"deployment/*.py" = ["INP001"]  # not intended for direct import
```

pyproject.toml
```diff
@@ -22,6 +22,33 @@ extend_skip_glob = ["**/migrations/*"]
 
 ##############################################
 
+[tool.ruff]
+target-version = "py310"
+line-length = 120
+select = ["F", "E", "B", "W", "N", "UP", "YTT", "FIX", "ASYNC", "A", "DJ", "EXE", "ICN", "G", "SLOT", "TID", "TCH", "INT", "C4", "ISC", "INP", "PIE", "RSE", "RET", "COM","PGH","FLY", "PERF", "PLE"]  # PYI,Q,SIM,PTH,PL,TRY,RUF
+ignore = [
+    "E501",                  # line-too-long: black does code formatting for us
```
richardebeling, line 30: Is the `line-length` setting relevant then? (I could imagine it being relevant for auto-fixes)

pyproject.toml
```diff
@@ -22,6 +22,33 @@ extend_skip_glob = ["**/migrations/*"]
...
+    "FIX004",                # hacks should be possible
+    "A003",                  # we shadow some builtins sometime
+    "DJ008",                 # do not use __str__ where it is not required
+    "COM812",                # incompatible with formatter
+    "PERF203",
```
richardebeling, line 35: I wouldn't expect us to hit this one often -- can we enable it?

richardebeling (commented):
Looking good! What's still left to do to get this ready to review?

evap/evaluation/models_logging.py
Kakadus: Yes, it's [DJ102](https://docs.astral.sh/ruff/rules/django-unordered-body-content-in-model/). Also suggested by the [django style guide](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#model-style)

evap/locale/de/formats.py
Kakadus: Sadly, nope. There is [astral-sh/ruff#1773](https://www.github.com/astral-sh/ruff/issues/1773) but yeah....

pyproject.toml
Kakadus: Not really. I don't know about auto-fixes, but I added `line-length` so that the formatter works (although black is still the way to go of course)

evap/evaluation/models.py
Kakadus: well, ruff _should_ have catched this...

pyproject.toml
Kakadus: there were some occurrences, but I can try

pyproject.toml
richardebeling: Ah, I forgot that they want to do linting and formatting. Well, fine with me, since they claim to be black-compatible.

pyproject.toml
richardebeling: If there's multiple occurrences or something annoying about them, I'm also fine to keep the exclusion and maybe add a comment why we weren't able to enable it. Just off the top of my head I just didn't see why we would often violate this.

evap/evaluation/models.py
richardebeling: Ah, so the pylint considers this a constant, which would have to be UPPER_CASE in pylint's interpretation of pep8 (since it's a [constant](https://peps.python.org/pep-0008/#constants)). Pylint probably would warn about a useless suppression. You removed the suppression because `invalid-name` is now disabled globally for pylint since ruff handles that part.

Quickly searching through the ruff rules, I couldn't find any that catches this. This implies to me that any name would be allowed here? I think pylint's interpretation of pep8 makes sense, and this should be a UPPER_CASE variable (if we could choose its name freely).

evap/locale/de/formats.py
richardebeling: Hmm, sad, but I can live with it for now.

evap/evaluation/models.py
niklasmohrin: And if I recall correctly, we weren't really happy with other parts of ruff variable name validation, right? So if we keep pylint anyways, we could also leave the naming stuff to pylint for now

pyproject.toml
Kakadus: Okay, _some_ occurrences mean one, in 
https://github.com/e-valuation/EvaP/blob/75bee4190544aeab13861560536f8939f6965356/evap/evaluation/models.py#L929-L956
only used as a management command. Here, the try-except covers invalid state transitions. We can add a `noqa` here

pyproject.toml
```diff
@@ -22,6 +22,35 @@ extend_skip_glob = ["**/migrations/*"]
...
+    "FIX004",                # hacks should be possible
+    "A003",                  # we shadow some builtins sometime
+    "DJ008",                 # do not use __str__ where it is not required
+    "COM812",                # incompatible with formatter
+    "N802",                  # not as mighty as pylint's invalid-name https://github.com/astral-sh/ruff/issues/7660
+    "PLR0913",               # we can't determine a good limit for arguements. reviews should spot bad cases of this.
+    "PLR2004", "PLW2901",    # not activated pylint extensions
+]
+
+ignore-init-module-imports = true
```
richardebeling, line 40: GitHub's toml parsing breaks here -- can we un-break it?

richardebeling (commented):
Code is fine for me

evap/evaluation/tests/test_commands.py
```diff
@@ -336,7 +336,7 @@ def test_dont_remind_already_voted(self):
         self.assertEqual(mock.call_count, 0)
         self.assertEqual(len(mail.outbox), 0)
 
-    @override_settings(TEXTANSWER_REVIEW_REMINDER_WEEKDAYS=list(range(0, 8)))
+    @override_settings(TEXTANSWER_REVIEW_REMINDER_WEEKDAYS=list(range(8)))
```
niklasmohrin, line 339: Any idea why we have the words "weekdays" and "8" in this combination here?

pyproject.toml
Kakadus: seems to break here but is ok in the rest of the file

evap/evaluation/tests/test_commands.py
Kakadus: should this be changed to `range(7)`? The comment in the settings indicate this should be `range(7)`:
https://github.com/e-valuation/EvaP/blob/main/evap/settings.py#L64-L67
```suggestion
    @override_settings(TEXTANSWER_REVIEW_REMINDER_WEEKDAYS=list(range(7)))
```

pyproject.toml
```diff
@@ -22,6 +22,35 @@ extend_skip_glob = ["**/migrations/*"]
 
 ##############################################
 
+[tool.ruff]
+target-version = "py310"
+line-length = 120
+select = ["F", "E", "B", "W", "N", "UP", "YTT", "FIX", "ASYNC", "A", "DJ", "EXE", "ICN", "G", "SLOT", "TID", "TCH", "INT", "C4", "ISC", "INP", "PIE", "RSE", "RET", "COM","PGH","FLY", "PERF", "PL", "BLE"]  # PYI,Q,SIM,PTH,TRY,RUF
```
richardebeling, line 28: What are `# PYI,Q,SIM,PTH,TRY,RUF`?

pyproject.toml
```diff
@@ -22,6 +22,35 @@ extend_skip_glob = ["**/migrations/*"]
...
+ignore = [
+    "E501",                  # line-too-long: black does code formatting for us
+    "FIX004",                # hacks should be possible
+    "A003",                  # we shadow some builtins sometime
+    "DJ008",                 # do not use __str__ where it is not required
+    "COM812",                # incompatible with formatter
+    "N802",                  # not as mighty as pylint's invalid-name https://github.com/astral-sh/ruff/issues/7660
+    "PLR0913",               # we can't determine a good limit for arguements. reviews should spot bad cases of this.
```
richardebeling, line 36:
```suggestion
    "PLR0913",               # we can't determine a good limit for arguments. reviews should spot bad cases of this.
```

pyproject.toml
```diff
@@ -61,12 +89,12 @@ disable = [
     "missing-docstring",              # yeah... we don't have those
     "protected-access",               # for us that means "be careful", not "don't do it"
     "too-many-lines",                 # we don't currently think that splitting up views.py or test_views.py creates any value
-    "too-many-arguments",             # we can't determine a good limit here. reviews should spot bad cases of this.
     "duplicate-code",                 # Mostly imports and test setup.
     "cyclic-import",                  # We use these inside methods that require models from multiple apps. Tests will catch actual errors.
     "unsupported-binary-operation",   # broken in pylint: https://github.com/PyCQA/pylint/issues/7381
     "use-implicit-booleaness-not-comparison-to-string",  # forces us to use less expressive code
     "use-implicit-booleaness-not-comparison-to-zero",    # forces us to use less expressive code
+    "broad-exception-caught", "line-too-long","unused-wildcard-import", "wildcard-import","too-many-arguments", "too-many-statements", "too-many-return-statements"
```
richardebeling, line 97: I guess there are covered by ruff now -- maybe add a comment about that?

pyproject.toml
Kakadus: We will add an issue to track these rules

richardebeling (approved):
Nice, thanks :)

evap/results/exporters.py
```diff
@@ -72,11 +72,12 @@ def init_grade_styles(cls):
                 "This can happen, if the file is imported / run multiple "
                 "times in one application run.",
                 ImportWarning,
+                stacklevel=3,
```
niklasmohrin, line 75: To pinpoint this to the callsite, it should be 2 right? If I duplicate the `ResultsExporter.init_grade_styles()` on line 334, I currently get a warning pointing me to `from evap.results.exporters import TextAnswerExporter`, with 2 it points to my faulty `ResultsExporter.init_grade_styles()`

niklasmohrin (approved):
The switch from rule names to numbers for `noqa` and the explicit `__all__` is a bit sad, but maybe ruff will figure it out in the future. Overall very nice!

evap/results/exporters.py
richardebeling: (related: https://github.com/e-valuation/EvaP/pull/2051#discussion_r1433284641 -- we there assumed that an additional execution of this file would be the "root" location, so we wanted to know where it got imported/executed from)

niklasmohrin: Oh okay, sure

evap/staff/fixtures/excel_files_test_data.py
```diff
@@ -208,7 +208,7 @@
     ['Prof. Dr.', 'Prorsus', 'Christoph', '123@external.com']
 ]
 
-random_file_content = b"Hallo Welt\n"
```
Kakadus, line 211: Just found out, why this was necessary: `fixtures` wasn't a python module before, so pylint ignored it

PR #2054, opened 2023-10-23
Mobile improvements

LinqLover:
Several CSS changes that avoid horizontal overflows or truncated texts on mobile.

### DIFF
--- a/evap/staff/templates/staff_semester_view.html
+++ b/evap/staff/templates/staff_semester_view.html
@@ -80,7 +80,7 @@
             {% endif %}
         </div>
         <div class="collapse" id="overviewCard">
-            <div class="card-body">
+            <div class="card-body table-responsive">
                 <table class="table">
                     <thead>
                         <tr>
@@ -259,7 +259,7 @@
                         </div>
                     </div>
                 </div>
-                <form id="evaluation_operation_form" method="GET" action="{% url 'staff:evaluation_operation' semester.pk %}">
+                <form id="evaluation_operation_form" class="table-responsive" method="GET" action="{% url 'staff:evaluation_operation' semester.pk %}">
                     {% if num_evaluations > 0 %}
                         <table id="evaluation-table" class="table table-striped table-narrow table-vertically-aligned">
                             <colgroup>
--- a/evap/static/scss/components/_buttons.scss
+++ b/evap/static/scss/components/_buttons.scss
@@ -158,7 +158,7 @@ a:not([href]):not(.disabled) {
     .button-open-notebook {
         position: fixed;
         bottom: $footer-height + 20px;
-        z-index: $zindex-sticky - 1;
+        z-index: $zindex-fixed;
         border-top-left-radius: 0;
         border-bottom-left-radius: 0;
     }
--- a/evap/static/scss/components/_buttons.scss
+++ b/evap/static/scss/components/_buttons.scss
@@ -158,7 +158,7 @@ a:not([href]):not(.disabled) {
     .button-open-notebook {
         position: fixed;
         bottom: $footer-height + 20px;
-        z-index: $zindex-fixed;
+        z-index: $zindex-sticky - 1;
         border-top-left-radius: 0;
         border-bottom-left-radius: 0;
     }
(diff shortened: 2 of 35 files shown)


evap/static/scss/components/_buttons.scss
```diff
@@ -158,7 +158,7 @@ a:not([href]):not(.disabled) {
     .button-open-notebook {
         position: fixed;
         bottom: $footer-height + 20px;
-        z-index: $zindex-fixed;
+        z-index: $zindex-sticky - 1;
```
richardebeling, line 161: What is the rationale here: Why sticky instead of fixed, why -1? Can we add a comment?

richardebeling (commented):
Thanks :)

evap/static/scss/components/_buttons.scss
LinqLover:
```suggestion
        z-index: $zindex-sticky - 1;  /* be behind sticky navbar menus */
```

richardebeling: Yup, sounds good :+1:

richardebeling (approved):
Code looks good to me :)

evap/staff/templates/staff_semester_view.html
```diff
@@ -259,7 +259,7 @@ <h3>
                         </div>
                     </div>
                 </div>
-                <form id="evaluation_operation_form" method="GET" action="{% url 'staff:evaluation_operation' semester.pk %}">
+                <form id="evaluation_operation_form" class="table-responsive" method="GET" action="{% url 'staff:evaluation_operation' semester.pk %}">
```
niklasmohrin, line 262: Not sure we want the `table-*` class on the form here, shouldn't we instead apply it to the nested tables?

evap/static/scss/components/_staff-index.scss
```diff
@@ -0,0 +1,17 @@
...
+
+        .card {
+            display: inline-block;
+            width: 100%;
+        }
+    }
+
+    .card {
+        margin-bottom: 1rem;
+    }
```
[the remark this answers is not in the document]
LinqLover: Sure, can commit this if it actually works :)

evap/staff/templates/staff_semester_view.html
LinqLover: > Make any table responsive across all viewports by wrapping a .table with .table-responsive.

https://getbootstrap.com/docs/5.3/content/tables/#responsive-tables

So the `table-responsive` is actually applied to the (or any) parent element of the table. :)

evap/static/scss/components/_distribution-bar.scss
```diff
@@ -62,6 +62,10 @@
         margin: 0 0.5rem;
     }
 
+    &:not(:empty) {
+        min-width: calc(3ch + 2*3px);
```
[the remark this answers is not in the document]
LinqLover: It's a while ago, but I guess for the following reason:

```suggestion
        min-width: calc(
            3ch /* up to 3 characters (e.g., 42%). This constraint is not relevant for 4 characters (100%) when there will be a single bar only. */
            + 2*3px /* left and right border */
        );
```

Feel free to commit this suggestion if we want to have the explanation in the sources :)

janno42, line 66:
```suggestion
        min-width: calc(3ch + 2*3px); // 3 characters + left and right border
```

evap/static/scss/components/_buttons.scss
```diff
@@ -158,7 +158,7 @@ a:not([href]):not(.disabled) {
     .button-open-notebook {
         position: fixed;
         bottom: $footer-height + 20px;
-        z-index: $zindex-fixed;
+        z-index: $zindex-sticky - 1;  /* be behind sticky navbar menus */
```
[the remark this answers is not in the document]
richardebeling: I tried getting the current z-index (on current main) of the button to cause some broken rendering but couldn't get it to do that :/ I'm especially unable to get the button to be in some place where a navbar dropdown is rendered. @LinqLover could you look into this again?

richardebeling (approved):

LinqLover:
Sorry for ghosting you, this had somehow escaped my attention in my cluttered inbox! Thanks for merging!

PR #2060, opened 2023-10-30
Type hints for class based views (and other tools)

### LINKED ISSUE #2055: Type-check class based views from #1964
In #1964, we made a mistake with the instance variables and return types inside [`UserMergeSelectionView::get_success_url`](https://github.com/e-valuation/EvaP/pull/1964/files#diff-94e55a2578328edb5cc0dd3b57b2ceae72f5e824fff0742ae770d59451f046d7R2292). Mypy didn't complain about these mistakes. I would have expected that the function signature of `get_success_url` from the base class is applied, but [mypy doesn't do this](https://mypy.readthedocs.io/en/stable/class_basics.html#overriding-statically-typed-methods).

To make mypy detect such errors, I think we want to type-annotate the class based views introduced in #1964 and possibly also mark them with the `@override` decorator.

richardebeling:
Fixes #2055

### DIFF
--- a/evap/evaluation/tools.py
+++ b/evap/evaluation/tools.py
@@ -2,7 +2,7 @@ import datetime
 from abc import ABC, abstractmethod
 from collections import defaultdict
 from collections.abc import Iterable, Mapping
-from typing import Any, TypeVar
+from typing import Any, Protocol, TypeVar
 from urllib.parse import quote
 
 import xlwt
@@ -14,7 +14,6 @@ from django.http import HttpRequest, HttpResponse
 from django.shortcuts import get_object_or_404
 from django.utils.translation import get_language
 from django.views.generic import FormView
-from typing_extensions import Protocol
 
 M = TypeVar("M", bound=Model)
 T = TypeVar("T")
@@ -117,7 +116,10 @@ def translate(**kwargs):
     return property(lambda self: getattr(self, kwargs[get_language() or "en"]))
 
 
-def clean_email(email: str) -> str:
+EmailT = TypeVar("EmailT", str, None)
+
+
+def clean_email(email: EmailT) -> EmailT:
     if email:
         email = email.strip().lower()
         # Replace email domains in case there are multiple alias domains used in the organisation and all emails should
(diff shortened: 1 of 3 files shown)


evap/evaluation/tools.py
```diff
@@ -9,10 +9,12 @@
 from django.conf import settings
 from django.core.exceptions import SuspiciousOperation, ValidationError
 from django.db.models import Model
-from django.http import HttpResponse
+from django.forms.formsets import BaseFormSet
+from django.http import HttpRequest, HttpResponse
 from django.shortcuts import get_object_or_404
 from django.utils.translation import get_language
 from django.views.generic import FormView
+from typing_extensions import Protocol
```
niklasmohrin, line 17: `Protocol` exists in `typing` since 3.8 :)

evap/evaluation/tools.py
```diff
@@ -58,52 +60,52 @@ def is_prefetched(instance, attribute_name: str):
...
 
-def is_external_email(email):
+def is_external_email(email: str) -> bool:
     return not any(email.endswith("@" + domain) for domain in settings.INSTITUTION_EMAIL_DOMAINS)
 
 
-def sort_formset(request, formset):
+def sort_formset(request: HttpRequest, formset: BaseFormSet) -> None:
     if request.POST:  # if not, there will be no cleaned_data and the models should already be sorted anyways
         formset.is_valid()  # make sure all forms have cleaned_data
```
niklasmohrin, line 90: (not yours), but?
```suggestion
        assert formset.is_valid()  # make sure all forms have cleaned_data
```

evap/evaluation/tools.py
```diff
@@ -115,7 +117,7 @@ def translate(**kwargs):
     return property(lambda self: getattr(self, kwargs[get_language() or "en"]))
 
 
-def clean_email(email):
+def clean_email(email: str) -> str:
```
niklasmohrin, line 120: This is actually `str | None`, see for example `UserForm.clean_email`. Not sure if we can model a `None` iff `None` relationship with generics or so, would be cool

niklasmohrin (changes requested):
Everything else looks good to me

evap/evaluation/tools.py
richardebeling: One test fails then. The method seems to be intended to also be callable with invalid formsets (via the default value in the line below).

niklasmohrin (approved):

Issue #1975, opened 2023-07-06
Filter results by participants and sort by voters


janno42:
On the results page, evaluations can currently be filtered by name, degree, course type, and semester.
They can be sorted by name, semester, name of the responsible person, and average grade.

An additional filter for the number of participants should be added, allowing filtering for small or large courses. It should use a multi-handle slider to set the lower and upper boundaries which should be placed in the empty space below the name filter field. This component doesn't exist in Bootstrap and must first be created. The minimum value for the slider should be 0, and the maximum value should be the highest existing number of participants.

Additionally, the "Voters" column should be made sortable by the percentage of voters.

janno42:
https://github.com/e-valuation/EvaP/commit/67ec739976ff00b3072dffd857fc712a02c0a5b3 adds a range slider to the results page. this can be used when implementing the data grid filter.

PR #2061, opened 2023-10-30
Sorting for participants in Issue #1975

constantin-kuehne:
This implements sorting on the participants on the evaluation result page. It is only part of the issue #1975. 
Co-authored-by: @till2


evap/evaluation/models.py
```diff
@@ -692,6 +692,11 @@ def can_publish_rating_results(self):
         # the rating results are only published if at least the configured number of participants voted during the evaluation for anonymity reasons
         return self.num_voters >= settings.VOTER_COUNT_NEEDED_FOR_PUBLISHING_RATING_RESULTS
 
+    @cached_property
+    def percent_voted(self):
+        return -(self.num_voters / self.num_participants)
```
niklasmohrin, line 697: The `-` here seems wrong - I would say that the ordering is also the wrong way around right now

evap/results/templates/results_index_evaluation_impl.html
```diff
@@ -50,10 +50,13 @@
         </div>
     {% endif %}
     {% if evaluation.is_single_result %}
-        <div data-col="participants" class="text-center"><span class="fas fa-user"></span>&nbsp;{{ evaluation.single_result_rating_result.count_sum }}</div>
+    <!-- data-order="{{ 1 }}" is used to show the singe result ratings on the bottom when sorting desc / top when asc -->
+    <div data-col="participants" class="text-center" data-order="{{ 1 }}"><span class="fas fa-user"></span>&nbsp;{{ evaluation.single_result_rating_result.count_sum }}</div>
```
niklasmohrin, line 54:
```suggestion
        {# data-order="-1" sorts single results before all other results #}
        <div data-col="participants" class="text-center" data-order="-1"><span class="fas fa-user"></span>&nbsp;{{ evaluation.single_result_rating_result.count_sum }}</div>
```

niklasmohrin (changes requested):
I wonder if sorting by the percentage is really what we want here. @janno42 ?

evap/results/templates/results_index_evaluation_impl.html
richardebeling: (It may also be desirable that single results are still ordered within their "category", currently their order would be random I guess?)

niklasmohrin: Agree. This may change completely if we would change to "total participants" instead of percentage anyways

janno42:
Yes, ordering by percentage of voters is specified in the issue and what we want here.
Finding courses of a certain size will be implemented via the participant count filter.

evap/results/templates/results_index_evaluation_impl.html
janno42: We can keep the order as is for now.

richardebeling:
Hey @constantin-kuehne and @till2 do you want to continue working on this or should we take over?

richardebeling:
@janno42 is this order fine: Courses with multiple evaluations first, then single results, then evaluations with ascending voter percentage (or the other way around if sorted descending)? Especially asking for courses with multiple evaluations, I could see that we'd want to average out the voter percentages of all their evaluations.

image

evap/results/templates/results_index.html
```diff
@@ -77,7 +77,7 @@ <h5 class="card-title mb-lg-0">
                 <div data-col="name" class="col-order">{% trans 'Evaluation' %}</div>
                 <div data-col="semester" class="col-order">{% trans 'Semester' %}</div>
                 <div data-col="responsible" class="col-order">{% trans 'Responsible' %}</div>
-                <div data-col="participants">{% trans 'Voters' %}</div>
+                <div data-col="participants" class="col-order">{% trans 'Voters' %}</div>
```
richardebeling, line 80: Is there any reason we intermingle `participants` and `voters` here (@janno42 @niklasmohrin)? :D Seems like this could be cleaned up

richardebeling (approved):

evap/results/templates/results_index.html
janno42: Good catch! Should be called `voters`

richardebeling: offline discussion: -> unify to "voters"

richardebeling:
Offline discussion results:
* We don't really care if single results are above or below the other results. They should however we grouped together.
* For courses with multiple evaluations, it probably makes sense to take the maximum voter participation ratio as a sorting criterion. An evaluation voter ratio gives more information here than a single result voter count, so it should be preferred for sorting

-> proposal:
* Voter ratios are in [0, 1]
* single results get a score of 0 (as they currently do)
* A course's data-order score is the max of the scores of its evaluations

janno42 (approved):
great stuff

evap/evaluation/models.py
```diff
@@ -813,6 +813,12 @@ def num_voters(self):
             return self._voter_count
         return self.voters.count()
 
+    @property
+    def voter_ratio(self):
+        if self.is_single_result or self.num_participants == 0:
+            return 0
+        return self.num_voters / self.num_participants
```
niklasmohrin, line 820: We picked 0 here because we want this to be the ordering, but is this also accurate for `voter_ratio`? I would think that in a single result, the ratio of voters is 100% rather than 0 - or maybe this is just a feeling I have right now

evap/static/ts/src/datagrid.ts
```diff
@@ -158,7 +158,12 @@ abstract class DataGrid {
             header.classList.remove("col-order-asc", "col-order-desc");
         }
         for (const [column, ordering] of this.state.order) {
-            this.sortableHeaders.get(column)!.classList.add(`col-order-${ordering}`);
+            const header = this.sortableHeaders.get(column);
+            if (header === undefined) {
+                // Silently ignore non-existing columns: They were probably renamed, users will have to rebuild their localstorage entry
```
niklasmohrin, line 163: And this will happen when? Can we rephrase the comment to rather emphasize why this is not a problem?

niklasmohrin (approved):
Works nicely, some last questions

evap/evaluation/models.py
richardebeling: Offline discussion: We could probably choose "0" or "1" here. I slightly favor "0" as "1" could indicate "everyone who could have voted voted", and that would be a wrong message e.g. with single results.

Thus, 0 is probably fine for now, but this is basically open for future change.

evap/static/ts/src/datagrid.ts
richardebeling: Rephrased to make clear that this will fix itself "automatically" the next time the user sorts by a column.

PR #2064, opened 2023-11-06
Replace `zip_choices` with strict `zip`

Kakadus:
Another commit extracted for #2051. The `zip_choices` filter zipped choices' `counts` and `names`. This asserts, that the no answer option is always the last option.

### DIFF
--- a/evap/evaluation/templatetags/evaluation_filters.py
+++ b/evap/evaluation/templatetags/evaluation_filters.py
@@ -5,6 +5,7 @@ from django.template import Library
 from django.utils.translation import gettext_lazy as _
 
 from evap.evaluation.models import BASE_UNIPOLAR_CHOICES, Contribution, Evaluation
+from evap.results.tools import RatingResult
 from evap.rewards.tools import can_reward_points_be_used_by
 from evap.student.forms import HeadingField
 
@@ -76,12 +77,7 @@ register = Library()
 
 @register.filter(name="zip")
 def _zip(a, b):
-    return zip(a, b)
-
-
-@register.filter()
-def zip_choices(counts, choices):
-    return zip(counts, choices.names, choices.colors, choices.values)
+    return zip(a, b, strict=True)
 
 
 @register.filter
@@ -115,12 +111,12 @@ def percentage_one_decimal(fraction, population):
 
 
 @register.filter
-def to_colors(choices):
-    if not choices:
+def to_colors(question_result: RatingResult | None):
+    if question_result is None:
         # When displaying the course distribution, there are no associated voting choices.
         # In that case, we just use the colors of a unipolar scale.
-        return BASE_UNIPOLAR_CHOICES["colors"]
-    return choices.colors
+        return BASE_UNIPOLAR_CHOICES["colors"][:-1]
+    return question_result.colors
 
 
 @register.filter
--- a/evap/results/templates/distribution_with_grade.html
+++ b/evap/results/templates/distribution_with_grade.html
@@ -5,7 +5,7 @@
     <div class="distribution-bar {{ question_result.choices.css_class }}"
         {% if question_result.question.is_bipolar_likert_question %} style="left: {{ question_result.minus_balance_count|percentage_one_decimal:question_result.count_sum }}"{% endif %}
     >
-        {% with colors=question_result.choices|to_colors %}
+        {% with colors=question_result|to_colors %}
             {% for value, color in distribution|zip:colors %}
                 {% if value != 0 %}
                     <div class="vote-bg-{{ color }}" style="width: {{ value|percentage_one_decimal:1 }};">&nbsp;</div>
--- a/evap/results/tools.py
+++ b/evap/results/tools.py
@@ -58,10 +58,13 @@ class RatingResult:
     def has_answers(cls, rating_result) -> TypeGuard["AnsweredRatingResult"]:
         return isinstance(rating_result, AnsweredRatingResult)
 
-    def __init__(self, question, additional_text_result=None):
+    def __init__(self, question, additional_text_result=None) -> None:
         assert question.is_rating_question
         self.question = discard_cached_related_objects(copy(question))
         self.additional_text_result = additional_text_result
+        self.colors = tuple(
+            color for _, color, value in self.choices.as_name_color_value_tuples() if value != NO_ANSWER
+        )
 
     @property
     def choices(self):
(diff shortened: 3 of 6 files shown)


evap/results/tools.py
```diff
@@ -48,10 +48,14 @@ def __init__(self, question, answer_counters, additional_text_result=None):
...
+            counts = OrderedDict(
+                (value, [0, name, color, value]) for (name, color, value) in self.choices.as_name_color_value_tuples()
+            )
+            counts.pop(NO_ANSWER)
             for answer_counter in answer_counters:
-                counts[answer_counter.answer] = answer_counter.count
-            self.counts = tuple(counts.values())
+                counts[answer_counter.answer][0] = answer_counter.count
+            self.zipped_choices = tuple(counts.values())
+            self.counts = tuple((count for count, _, _, _ in counts.values()))
```
niklasmohrin, line 58: This feels overly complicated, maybe something like the following is better?
```python
            answer_indices = {
                answer: i for i, answer in enumerate(self.choices.values) if answer != NO_ANSWER
            }
            counts = [0] * len(answer_indices)
            for counter in answer_counters:
                counts[answer_indices[counter.answer]] = counter.count
            self.counts = tuple(counts)
            self.zipped_choices = self.choices.as_name_color_value_tuples()
```
We should probably refactor the whole choices business to use dataclasses instead of named tuples and increase typing with these result classes here, but we don't need to do it now

Kakadus: The problem is, that `self.choices.as_name_color_value_tuples()` contains a `NO_ANSWER`-element as the last element. This needs to be removed from the result of `as_name_color_value_tuples`. Also, the counts need to be prepended

evap/evaluation/templatetags/evaluation_filters.py
```diff
@@ -115,12 +111,12 @@ def percentage_one_decimal(fraction, population):
...
 @register.filter
-def to_colors(choices):
-    if not choices:
+def to_colors(question_result: RatingResult | None):
+    if not question_result:
         # When displaying the course distribution, there are no associated voting choices.
         # In that case, we just use the colors of a unipolar scale.
-        return BASE_UNIPOLAR_CHOICES["colors"]
-    return choices.colors
+        return BASE_UNIPOLAR_CHOICES["colors"][:-1]
```
niklasmohrin, line 118: Was this wrong before? Do we need a test for it?

evap/results/tools.py
```diff
@@ -42,55 +45,76 @@ def __init__(self, visible_by_contribution: Iterable[UserProfile], visible_by_de
...
+class _CountedRatingResult(RatingResult):
+    counts: tuple[int, ...]
+    zipped_choices: list[list[Real | StrOrPromise | str | Real]]
+
+    count_sum: int
+    average: float
+    approval_count: int
+    minus_balance_count: float
+
+    is_published = True
```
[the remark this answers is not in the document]
Kakadus: you're right. If the `is_published` property would have remained in `RatingResult` it would make a difference though

evap/results/templates/distribution_with_grade.html
```diff
@@ -5,7 +5,7 @@
     <div class="distribution-bar {{ question_result.choices.css_class }}"
         {% if question_result.question.is_bipolar_likert_question %} style="left: {{ question_result.minus_balance_count|percentage_one_decimal:question_result.count_sum }}"{% endif %}
     >
-        {% with colors=question_result.choices|to_colors %}
+        {% with colors=question_result|to_colors %}
             {% for value, color in distribution|zip:colors %}
```
Kakadus, line 9: to your other comment: before, distribution values (w/o NO_ANSWER) may have been zipped with the colors with NO_ANSWER. 
Now, I explicitly remove NO_ANSWER from the colors and zip them strictly. I don't think a test is necessary, because this is already covered and only does some narrowing on the zipable tuples

evap/evaluation/templatetags/evaluation_filters.py
Kakadus: see comment below in distribution_with_grade

evap/results/tools.py
```diff
@@ -42,55 +45,75 @@ def __init__(self, visible_by_contribution: Iterable[UserProfile], visible_by_de
...
+        if not is_published(self):
             return None
         return sum(self.counts)
 
     @property
-    def minus_balance_count(self) -> float:
+    def minus_balance_count(self) -> float | None:
         assert self.question.is_bipolar_likert_question
+        if not is_published(self):
+            return None
```
niklasmohrin, line 79: In the tests, this never happens. Should this be an assert instead? If not, why?

evap/results/tools.py
```diff
@@ -42,55 +45,75 @@ def __init__(self, visible_by_contribution: Iterable[UserProfile], visible_by_de
...
+    zipped_choices: list[list[Real | StrOrPromise | str | Real]]
+
+    count_sum: int
+    average: float
+    approval_count: int
+    minus_balance_count: float
+
+
+def is_published(result: RatingResult) -> TypeGuard[_CountedRatingResult]:
+    return result.counts is not None
```
niklasmohrin, line 112: I don't know, this all feels a bit brittle, we only have this situation because we are passing `None` for `answer_counters` if we don't want to publish them. We should probably refactor this (at some point) to have actual typing describe the state of the system and not work around it with `TypeGuard`. Why is this needed here again?

evap/results/tools.py
```diff
@@ -42,55 +45,75 @@ def __init__(self, visible_by_contribution: Iterable[UserProfile], visible_by_de
...
-        return self.is_published and any(count != 0 for count in self.counts)
+        return has_answers(self)
 
-    @property
-    def is_published(self) -> bool:
-        return self.counts is not None
+
+class _CountedRatingResult(RatingResult):
+    counts: tuple[int, int, int, int]
+    zipped_choices: list[list[Real | StrOrPromise | str | Real]]
```
niklasmohrin, line 103:
```suggestion
    zipped_choices: list[list[Real | StrOrPromise]]
```

evap/results/tools.py
Kakadus: The check/assert is needed because `self.counts` and `self.count_sum` is accessed. I just followed the style of the other classes. We can use the assert here instead if it's better

evap/results/tools.py
Kakadus: You are right, the typeguards are not the the actual target of this PR and are not required here

richardebeling (commented):
Looking good so far, but will have to recheck changes to the Result classes once #2091 is merged

richardebeling (approved):

niklasmohrin (approved):

PR #2065, opened 2023-11-06
 Suggest user merge if user with same email address exists 

### LINKED ISSUE #1958: Suggest user merge if email address already exists
As a manager, when changing a user's email address to an existing email address of another user, an error is shown in the user form.
In addition, a warning message should be shown on top of the page, suggesting to merge the two users with a link to the merge page (where the user currently edited will be the main user).

hansegucker:
Close #1958

### DIFF
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -999,8 +999,8 @@ class UserForm(forms.ModelForm):
         if self.instance and self.instance.pk:
             user_with_same_email = user_with_same_email.exclude(pk=self.instance.pk)
 
-        if user_with_same_email.exists():
-            self.user_with_same_email = user_with_same_email.get()
+        if user_with_same_email:
+            self.user_with_same_email = user_with_same_email.first()
             raise forms.ValidationError(_("A user with the email '%s' already exists") % email)
         return email
 
--- a/evap/staff/forms.py
+++ b/evap/staff/forms.py
@@ -957,6 +957,7 @@ class UserForm(forms.ModelForm):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.user_with_same_email = None
         evaluations_in_active_semester = Evaluation.objects.filter(course__semester=Semester.active_semester())
         excludes = [x.id for x in evaluations_in_active_semester if x.is_single_result]
         evaluations_in_active_semester = evaluations_in_active_semester.exclude(id__in=excludes)
@@ -999,6 +1000,7 @@ class UserForm(forms.ModelForm):
             user_with_same_email = user_with_same_email.exclude(pk=self.instance.pk)
 
         if user_with_same_email.exists():
+            self.user_with_same_email = user_with_same_email.get()
             raise forms.ValidationError(_("A user with the email '%s' already exists") % email)
         return email
 
--- a/evap/staff/templates/staff_user_form.html
+++ b/evap/staff/templates/staff_user_form.html
@@ -32,6 +32,16 @@
         {% endif %}
     </div>
 
+    {% if user_with_same_email %}
+        <div class="alert alert-warning alert-dismissible">
+            <p>
+                {% trans "A user with this email address already exists. You probably want to merge the users." %}
+            </p>
+            <a type="button" class="btn btn-primary btn-sm" href="{% url 'staff:user_merge' user_with_same_email.id form.instance.id %}">{% trans "Merge both users" %}</a>
+            <button type="button" class="btn-close" data-bs-dismiss="alert"></button>
+        </div>
+    {% endif %}
+
 
     <form id="user-form" method="POST" class="form-horizontal multiselect-form">
         {% csrf_token %}
--- a/evap/staff/tests/test_views.py
+++ b/evap/staff/tests/test_views.py
@@ -272,12 +272,24 @@ class TestUserEditView(WebTestStaffMode):
         cls.testuser = baker.make(UserProfile)
         cls.url = f"/staff/user/{cls.testuser.pk}/edit"
 
-    def test_questionnaire_edit(self):
+    def test_user_edit(self):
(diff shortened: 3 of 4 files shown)


evap/staff/forms.py
```diff
@@ -999,6 +1000,7 @@ def clean_email(self):
             user_with_same_email = user_with_same_email.exclude(pk=self.instance.pk)
 
         if user_with_same_email.exists():
+            self.user_with_same_email = user_with_same_email.get()
```
niklasmohrin, line 1003: This now does two database queries, one for `exists` and one for `get`. In general, could it be the case that multiple users exist? @janno42

evap/staff/templates/staff_user_form.html
```diff
@@ -32,6 +32,16 @@ <h3>{% if form.instance.id %}{% trans 'Edit user' %}{% else %}{% trans 'Create u
         {% endif %}
     </div>
 
+    {% if user_with_same_email %}
+        <div class="alert alert-warning alert-dismissible">
+            <p>
+                {% trans "There is already another user with this email address. You probably want to merge the users." %}
+            </p>
+            <a type="button" class="btn btn-primary btn-sm" href="{% url 'staff:user_merge' form.instance.id user_with_same_email.id %}">{% trans "Merge both users" %}</a>
```
janno42, line 40:
```suggestion
            <a type="button" class="btn btn-primary btn-sm" href="{% url 'staff:user_merge' user_with_same_email.id form.instance.id %}">{% trans "Merge both users" %}</a>
```

In the specification, I wrote it the wrong way round - please make the existing user the main user of the merge.

evap/staff/forms.py
hansegucker: No, because `email` is unique on database level.

evap/staff/templates/staff_user_form.html
```diff
@@ -32,6 +32,16 @@ <h3>{% if form.instance.id %}{% trans 'Edit user' %}{% else %}{% trans 'Create u
         {% endif %}
     </div>
 
+    {% if user_with_same_email %}
+        <div class="alert alert-warning alert-dismissible">
+            <p>
+                {% trans "There is already another user with this email address. You probably want to merge the users." %}
```
janno42, line 38:
```suggestion
                {% trans "A user with this email address already exists. You probably want to merge the users." %}
```

janno42 (changes requested):

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

evap/staff/forms.py
niklasmohrin: Except that we are matching with `iexact` above. I think we can usually assume that all emails in the database are lowercased, but I wonder if we should handle the other case here (or assert that at most one matches?)

richardebeling: I think it's fair if we just take the first email that matches. In the (unlikely) case there are multiple matching emails, you'd simply get another merge suggestion afterwards.

evap/staff/tests/test_views.py
```diff
@@ -272,13 +272,25 @@ def setUpTestData(cls):
...
         page = self.app.get(self.url, user=self.manager, status=200)
         form = page.forms["user-form"]
         form["email"] = "lfo9e7bmxp1xi@institution.example.com"
         form.submit()
         self.assertTrue(UserProfile.objects.filter(email="lfo9e7bmxp1xi@institution.example.com").exists())
 
+    def test_user_edit_duplicate_email(self):
+        second_user = baker.make(UserProfile)
+        second_user.email = "lfo9e7bmxp1xi@institution.example.com"
+        second_user.save()
```
richardebeling, line 285:
```suggestion
        second_user = baker.make(UserProfile, email="lfo9e7bmxp1xi@institution.example.com")
```

evap/staff/tests/test_views.py
```diff
@@ -272,13 +272,25 @@ def setUpTestData(cls):
...
+        page = self.app.get(self.url, user=self.manager, status=200)
+        form = page.forms["user-form"]
+        form["email"] = "lfo9e7bmxp1xi@institution.example.com"
```
richardebeling, line 288: If you reuse the value, I as a reader don't have to spend mental energy on comparing the value:
```suggestion
        form["email"] = second_user.email
```

evap/staff/tests/test_views.py
richardebeling: (we could also use a less cryptic looking mailbox-part than "lfo9e7bmxp1xi" here. I don't know why the existing test used that, may have been about accidental collisions, but that shouldn't be a problem anymore)

evap/staff/forms.py
hansegucker: Done

hansegucker:
@richardebeling Please review again.

evap/staff/tests/test_views.py
```diff
@@ -272,12 +272,24 @@ def setUpTestData(cls):
...
-        form["email"] = "lfo9e7bmxp1xi@institution.example.com"
+        form["email"] = "test@institution.example.com"
         form.submit()
-        self.assertTrue(UserProfile.objects.filter(email="lfo9e7bmxp1xi@institution.example.com").exists())
+        self.assertTrue(UserProfile.objects.filter(email="test@institution.example.com").exists())
+
+    def test_user_edit_duplicate_email(self):
+        second_user = baker.make(UserProfile)
+        second_user.email = "test@institution.example.com"
+        second_user.save()
```
richardebeling, line 285:
```suggestion
        second_user = baker.make(UserProfile, email="test@institution.example.com")
```

richardebeling (approved):

niklasmohrin (approved):
(lgtm, once the other comment is fixed)

richardebeling:
thanks :)

PR #2071, opened 2023-11-13
Translate evaluation state names in log display

### LINKED ISSUE #1693: Translate evaluation state names in log display
When displaying a logentry for a change of the `state` field of an `evaluation`, the state names should not only be mapped from their integer representation to string values (what we currently do), but also be translated into the display language.

For example, although I am using the site in German, the logs at the end of the evaluation edit site for "Fachspezifisches Englisch (Level 2)" are saying: "State: reviewed → evaluated"

DieKautz:
This closes #1693.


richardebeling (approved):
Code looks good to me -- it would be nice if you could add a small test.

janno42 (commented):
It seems that the label "State" is not yet translated in the log, although I don't know why. Can you take a look at this?

DieKautz:
Ah. I forgot to add that a pipeline fail caused by this would be expected, as we already discovered that "State" (capitalized) would still have to be (fuzzy) translated by you(?).
Also "editor_approved" would need translation

evap/evaluation/tests/test_models.py
```diff
@@ -537,6 +538,21 @@ def test_textanswer_review_state(self):
             evaluation.TextAnswerReviewState.REVIEWED,
         )
 
+    def test_state_change_log_translated(self):
+        evaluation = baker.make(
+            Evaluation,
+            state=Evaluation.State.NEW,
+        )
+        evaluation.save()
```
richardebeling, line 546:
```suggestion

File: evap/evaluation/tests/test_models.py [hunk 2]
```diff
@@ -537,6 +538,21 @@ def test_textanswer_review_state(self):
...
+        with translation.override("de"):
+            # Refresh reference to evaluation, to force new log entry
+            evaluation = Evaluation.objects.get(pk=evaluation.pk)
+            evaluation.ready_for_editors()
+            evaluation.save()
+            log_entry = evaluation.related_logentries().order_by("id").last()
+            self.assertEqual(Evaluation.state_to_str(log_entry.data["state"]["change"][0]), "neu")
```
richardebeling, line 553: (Semi-answering to https://github.com/e-valuation/EvaP/pull/2071#issuecomment-1819740914) The CI run currently fails here because "new" isn't translated in the test. You said that the translation for this should already be available, right? If so, this might be a bit harder to debug, feel free to ask me/us to look into what's going on here in that case.

evap/evaluation/tests/test_models.py
```diff
@@ -537,6 +538,21 @@ def test_textanswer_review_state(self):
...
+            self.assertEqual(Evaluation.state_to_str(log_entry.data["state"]["change"][1]), "vorbereitet")
```
richardebeling, line 554: I was kind of hoping that we could have a bit of a larger frame in the test -- this test now is very low-level and wouldn't catch if we someday stopped calling `state_to_str` in `transform_log_action`. A view level test might be more appropriate here. Sorry, when I said a model test would be fine I was hoping that there is a higher-level method on the model that could be used, something like "give me the list of log entries as strings to be shown on the page".

I can take over the testing from here if you want, ping me in case you'd prefer that.

evap/evaluation/tests/test_models.py
DieKautz: Thats probably true :+1: . The docs are also implying this

evap/evaluation/tests/test_models.py
DieKautz: Would this test fit in `TestEvaluationEditView`? My idea would be to use similar logic as in in the model test and use some kind of `self.assertIn("neu", page)` or `self.assertIn("neu", log_div)` and also the negation if "en" is selected.
If that would be the path you indended, I would try doing it myself, and reach out again when the next roadblock appears

richardebeling: Yes, that sounds good 👍 . "neu" itself might appear in other places on the page, so `self.assertIn("neu", page)` might pass even if your translation was broken. Searching in the log-div would probably be fine if the word doesn't appear there otherwise (should be asserted in the test then), or you could add a bit more context around, i.e. search for "Zustand geändert: X --> neu" or something (don't currently know the exact formatting / wording used in the log entry).

evap/evaluation/tests/test_models.py
richardebeling: Yup, the translation entry you linked to should work here. I don't see why this fails from the code, will debug and get back to you.

richardebeling: Turns out we've never used the hardcoded translation in any test so far. Tests only use translations stored in the database, such as english/german names of evaluations. We don't even compile the messages (`manage.py compilemessages`) on the CI runner, so the machine-readable translation file `django.mo` is missing on the CI runner. Thus, the translation entry linked above is not available in the test.

We could compile the message catalog on the test runner, but that would add a dependency between this test and the translations, and it would increase test run time.

Instead, I'd propose to just [patch](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) the `gettext` / `_` method for this test to "hardcode" the expected translation. Then, the test is independent of the actual translated string and works without the compiled messages available. You can do this by decorating the test as follows:

```python3
@patch("evap.evaluation.models._", new=(lambda string: "vorbereitet" if string == "prepared" else string))
def test_state_change_log_translated(self):
    ....
    # now only check for "prepared"/"vorbereitet", since we only included
    # that value in our mocked translation method
```
The same approach should also work with a view test.

evap/evaluation/tests/test_models.py
DieKautz: See [44cb0c7](https://github.com/e-valuation/EvaP/pull/2071/commits/44cb0c7a314ebcf9867b47a28aae07f05d00fbda)

evap/evaluation/tests/test_models.py
DieKautz: See [97983e8](https://github.com/e-valuation/EvaP/pull/2071/commits/97983e80e0dd1cc39808cc82cde64a0893ead3fa)

DieKautz:
I cannot attend in-person today. But hopefully it should now be ready for the final review :) @richardebeling

richardebeling:
> edit: nvm, for some reason overriding the language works different in the ci? (all 575 pass locally)

The issue was with the path that you patched -- see the [Where to patch?](https://docs.python.org/3/library/unittest.mock.html#where-to-patch) part of the documentation for more details. In short, if a name is already imported in some location, patching the source-location afterwards doesn't work.

For your test, it should work if you patch the name `evap.evaluation.models._` -- the following test passes for me:

```python3
    @patch("evap.evaluation.models._", new=(lambda string: "vorbereitet" if string == "prepared" else string))
    def test_state_change_log_translated(self):
        log_ul = self.app.get(self.url, user=self.manager).html.find(class_="list-group")
        self.assertNotIn("vorbereitet", str(log_ul))

        self.evaluation.ready_for_editors()
        self.evaluation.save()

        log_ul = self.app.get(self.url, user=self.manager).html.find(class_="list-group")
        self.assertIn("vorbereitet", str(log_ul))
```

To get the same behavior locally as with the CI run, you can simply `rm evap/locale/de/LC_MESSAGES/django.mo`

evap/evaluation/models.py
```diff
@@ -791,7 +791,7 @@ def unpublish(self):
 
     @classmethod
     def state_to_str(cls, state):
-        return cls.STATE_STR_CONVERSION[state]
```
janno42, line 794: Instead of translating it here, please add the underscore call to all strings above in `STATE_STR_CONVERSION`. Then "editor_approved" and "in_evaluation" will be added to the translated strings.

DieKautz: Now included with 545a75e :+1:

DieKautz:
> For your test, it should work if you patch the name `evap.evaluation.models._` -- the following test passes for me:

When executing just this single test, it does not pass for me if `self.manager`s lanaguage is not overwritten.
Could this be because language is not reset in an earlier test, maybe [here](https://github.com/e-valuation/EvaP/blob/18f91943b1cfb4b023547b725238866043ac2c80/evap/staff/tests/test_views.py#L704)?

richardebeling:
Ah, sorry. When we moved the `_(...)` call out of the body of `state_to_str`, we changed the time when the `_` method is evaluated. The call inside the function is evaluated during test runtime. The class variable is evaluated before the test runs, so by the time we arrive at the test, the value is already a string marked to be lazily translated. Thus, at that point, patching `evap.evaluation.models._` does not change anything anymore, the strings will still be translated by the default django mechanics. These then use the language of `self.manager`, which is english by default. (The whole idea behind the patching was that the language of `self.manager` doesn't matter, because we overwrite the translation result for all languages.)

Moving the `_()` call to the values of the `STATE_STR_CONVERSION` certainly is the right thing to do here. However, I don't see a nice way to overwrite the translation result for the test with this setup. Django has an [`override_settings`](https://docs.djangoproject.com/en/4.2/topics/testing/tools/#django.test.override_settings) helper to dynamically change setting values for tests. Optimally, we'd have an equivalent for translation, but they don't seem to have that.

My concern before, when I asked you to write a test, was that the code could (in future) not call `state_to_str`, but use `STATE_STR_CONVERSION` directly and thus fail to translate. Considering that `STATE_STRING_CONVERSION` is now translated, "failing to translate" can not really be a problem anymore. We could thus just test that the values of `STATE_STRING_CONVERSION` are used. This test passes for me:

```python3
    @patch.dict(Evaluation.STATE_STR_CONVERSION, {Evaluation.State.PREPARED: "mock-translated-prepared"})
    def test_state_change_log_translated(self):
        log_ul = self.app.get(self.url, user=self.manager).html.find(class_="list-group")
        self.assertNotIn("mock-translated-prepared", str(log_ul))

        self.evaluation.ready_for_editors()
        self.evaluation.save()

        log_ul = self.app.get(self.url, user=self.manager).html.find(class_="list-group")
        self.assertIn("mock-translated-prepared", str(log_ul))
```

niklasmohrin (approved):
Nice, thanks!

janno42 (approved):
Thanks

PR #2073, opened 2023-11-20
Introduce typeguard in tests

### LINKED ISSUE #2066: Check correctness of type annotations at test runtime
Plausible but wrong type annotations could give a false sense of security and are therefore (imho) more harmful than no type annotations. When we go through old code and annotate it (as we currently do in #2035), we might make mistakes that mypy does not catch, e.g. because calling code that uses other argument types simply isn't typed yet.

There is https://typeguard.readthedocs.io/en/latest/, which allows to inject run-time type checking into the program. I'm thinking of running the tests with this injected so that we can catch cases where we annotated stuff wrongly or too strictly.

One issue I can see is that this disallows passing mocked arguments in tests (especially if the validation uses `isinstance`). We might have to add manual exclusions for such cases. I wouldn't want this to be too invasive though, a best case scenario for me would be a handful of lines to inject the checking and maybe a handful more lines for manual exclusions where we use mocks.

@niklasmohrin @Kakadus thoughts?

Kakadus:
close #2066

### DIFF
--- a/evap/evaluation/tools.py
+++ b/evap/evaluation/tools.py
@@ -1,4 +1,5 @@
 import datetime
+import typing
 from abc import ABC, abstractmethod
 from collections import defaultdict
 from collections.abc import Iterable, Mapping
@@ -12,13 +13,16 @@ from django.db.models import Model
 from django.forms.formsets import BaseFormSet
 from django.http import HttpRequest, HttpResponse
 from django.shortcuts import get_object_or_404
+from django.utils.datastructures import MultiValueDict
 from django.utils.translation import get_language
 from django.views.generic import FormView
+from django_stubs_ext import StrOrPromise
 
 M = TypeVar("M", bound=Model)
 T = TypeVar("T")
 Key = TypeVar("Key")
 Value = TypeVar("Value")
+CellValue: typing.TypeAlias = str | int | float
 
 
 def unordered_groupby(key_value_pairs: Iterable[tuple[Key, Value]]) -> dict[Key, list[Value]]:
@@ -35,7 +39,9 @@ def unordered_groupby(key_value_pairs: Iterable[tuple[Key, Value]]) -> dict[Key,
     return dict(result)
 
 
-def get_object_from_dict_pk_entry_or_logged_40x(model_cls: type[M], dict_obj: Mapping[str, Any], key: str) -> M:
+def get_object_from_dict_pk_entry_or_logged_40x(
+    model_cls: type[M], dict_obj: MultiValueDict[str, Any] | Mapping[str, Any], key: str
+) -> M:
     try:
         return get_object_or_404(model_cls, pk=dict_obj[key])
     # ValidationError happens for UUID id fields when passing invalid arguments
@@ -130,7 +136,7 @@ def clean_email(email: EmailT) -> EmailT:
     return email
 
 
-def capitalize_first(string: str) -> str:
+def capitalize_first(string: StrOrPromise) -> str:
     return string[0].upper() + string[1:]
 
 
@@ -174,6 +180,7 @@ class FormsetView(FormView):
         return super().form_valid(formset)
 
 
+@typing.runtime_checkable
 class HasFormValid(Protocol):
     def form_valid(self, form):
         pass
@@ -262,7 +269,7 @@ class ExcelExporter(ABC):
         else:
             self.cur_sheet = None
 
-    def write_cell(self, label: str | None = "", style: str = "default") -> None:
+    def write_cell(self, label: CellValue | None = "", style: str = "default") -> None:
         """Write a single cell and move to the next column."""
         self.cur_sheet.write(
             self.cur_row,
@@ -276,7 +283,9 @@ class ExcelExporter(ABC):
         self.cur_col = 0
         self.cur_row += 1
 
-    def write_row(self, vals: Iterable[str], style: str = "default") -> None:
+    def write_row(
+        self, vals: Iterable[CellValue | None], style: str | typing.Callable[[CellValue | None], str] = "default"
+    ) -> None:
         """
         Write a cell for every value and go to the next row.
         Styling can be chosen
--- a/evap/evaluation/tools.py
+++ b/evap/evaluation/tools.py
@@ -137,6 +137,7 @@ def clean_email(email: EmailT) -> EmailT:
 
 
 def capitalize_first(string: StrOrPromise) -> str:
+    """Realize lazy promise objects and capitalize first letter."""
     return string[0].upper() + string[1:]
 
 
--- a/evap/evaluation/tools.py
+++ b/evap/evaluation/tools.py
@@ -285,9 +285,7 @@ class ExcelExporter(ABC):
         self.cur_col = 0
         self.cur_row += 1
 
-    def write_row(
-        self, vals: Iterable[CV], style: str | typing.Callable[[CV], str] = "default"
-    ) -> None:
+    def write_row(self, vals: Iterable[CV], style: str | typing.Callable[[CV], str] = "default") -> None:
         """
         Write a cell for every value and go to the next row.
         Styling can be chosen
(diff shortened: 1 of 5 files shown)


evap/staff/importers/enrollment.py
```diff
@@ -73,7 +73,14 @@ def differing_fields(self, other) -> set[str]:
...
 
-class ValidCourseData(CourseData):
+class Meta(type):
+    def __instancecheck__(cls, instance: object) -> TypeGuard["ValidCourseData"]:
+        if not isinstance(instance, CourseData):
+            return False
+        return all_fields_valid(instance)
+
+
+class ValidCourseData(CourseData, metaclass=Meta):
```
richardebeling, line 83: Wow, didn't know about `__instancecheck__`. On one hand this is so cool, but on the other hand it allows programmers to do so many things that are just wrong :D

Two notes:
1. should probably be named `ValidCourseDataMeta` instead of just `Meta`?
2. does this allow us to just use `isinstance` in the importer when we currently check `all_fields_valid`? If yes, is that desirable? Feels consistent to me, but I'm unsure.

manage.py
```diff
@@ -5,6 +5,11 @@
 if __name__ == "__main__":
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "evap.settings")
 
+    if "test" in sys.argv:
+        from typeguard import install_import_hook
+
+        install_import_hook(("evap", "tools"))
+
```
richardebeling, line 12: We have an implementation of "are we testing?" in `settings.py` and also have a block of things we do in this case: https://github.com/e-valuation/EvaP/blob/cdbdc9d632413f0e8ecdf3cfceb23cdd9eb86a29/evap/settings.py#L419-L422

Could this stuff also be done there?

evap/staff/importers/enrollment.py
Kakadus: if 2., should `all_fields_valid` be merged with this instancecheck? What do you think @niklasmohrin

Kakadus:
Github does not update to the last changes somehow... Maybe my changes will come up soon...

evap/staff/importers/enrollment.py
niklasmohrin: If I understand correctly, we are overwriting `isinstance(x, ValidCourseData)` here. Do you know if this has any implications internally? Afaik, all Python objects have their own `__class__` field to look up methods etc. So if we had a method on `ValidCourseData`, we could have the situation that an object passes `isinstance`, but still uses a method from a parent, right? This is odd to think about, can't we just have `def as_valid(self) -> Optional[ValidCourseData]` method where we construct the object of the child class?

(If the answer is no, I would be fine with merging the two methods)

Kakadus: As far as I understood it, `isinstance(x, cls)` will use the implementation of `__instancecheck__` of the metaclass of `cls`, for all and every `x`. The metaclass is also "inherited" when subclassing `ValidCourseData` in this case. So, the answer should be no I guess?

Context: typeguard uses `isinstance` to determine if the annotated return type matches the type of the returned object. typeguard does not support `TypeGuard`s though (originally intended for static typechecking only) and complains that valid `CourseData` objects aren't actually `ValidCourseData` objects at runtime. This is just a limitation of typeguard.

Since metaclasses always cause some headache (see this thread), maybe it is still better to use `as_valid` all along? Then I feel like we may be loosing some conveniences of `TypeGuard` though.

We can also just don't do 2., i.e. revert [123af29](https://github.com/e-valuation/EvaP/pull/2073/commits/123af290212505100e0895b6bbb66100b57c0a23) and leave the `__instancecheck__` meta thingy only for typeguard and continue to use `all_fields_valid`

Kakadus: After reading your question again: You mean when we override a function of `CourseData` in `ValidCourseData`? Then indeed, the implementation in `CourseData` should be used, because the `type` of the object did not change. Since we are talking about `TypeGuard`s here, this is not really relevant because you won't override a function in `ValidCourseData`, if I understood `TypeGuard` correctly (The "feature" is that the runtime type remains the same while). You would only change the function parameter types / return types

evap/settings.py
```diff
@@ -420,6 +420,10 @@ def CHARACTER_ALLOWED_IN_NAME(character):  # pylint: disable=invalid-name
 
 # speed up tests
```
niklasmohrin, line 421: This comment seems out of place now that more stuff is going on in the branch below

evap/evaluation/tools.py
```diff
@@ -130,7 +136,7 @@ def clean_email(email: EmailT) -> EmailT:
     return email
 
 
-def capitalize_first(string: str) -> str:
+def capitalize_first(string: StrOrPromise) -> str:
```
niklasmohrin, line 139: Ah interesting; should we write a comment that clarifies that this realizes the lazy promise object?

evap/evaluation/tools.py
```diff
@@ -276,7 +283,9 @@ def next_row(self) -> None:
         self.cur_col = 0
         self.cur_row += 1
 
-    def write_row(self, vals: Iterable[str], style: str = "default") -> None:
+    def write_row(
+        self, vals: Iterable[CellValue | None], style: str | typing.Callable[[CellValue | None], str] = "default"
```
niklasmohrin, line 287: Shouldn't this be generic over the type of the iterator item? Because if I give in some numbers and a function that expects numbers, I don't want typing to fail because I didn't handle the case where `style` is called with a string. Isn't it something like `TypeVar(..., bound=CellValue)`? We can also postpone this until it actually occurs somewhere though.

niklasmohrin (commented):
Have we verified that this catches the kinds of errors we expect it to / are there any examples yet?

Some final remarks:

evap/staff/importers/enrollment.py
niklasmohrin: I am not really convinced to be honest, I feel that we should generally avoid playing with `isinstance`, because it is a function that programmers have strong assumptions about and we are breaking them here. In particular, this typing-related change has implications in non-typing situations, because we actually changed the code.

In the case here, the change is probably harmless and the easiest we will get. If you two prefer the current approach here, I am fine with that, still, I would vote against `__instancecheck__` though :D

richardebeling: I'd be ok with it here since `ValidCourseData` is a dataclass with no methods, but I see your point, so I'm also open to other workarounds / solutions.

evap/settings.py
Kakadus: altered the comment

evap/evaluation/tools.py
Kakadus: added a docstring

evap/evaluation/tools.py
Kakadus: added a typevar for this

evap/staff/importers/enrollment.py
Kakadus: okay, reverted 123af29 first of all

niklasmohrin (approved):

evap/staff/importers/enrollment.py
niklasmohrin: After taking another look at the usage of `ValidCourseData`, I am also fine with moving forward with the current version. Constructing a true instance of `ValidCourseData` would mean that we need to inject this new instance back into the process, adding another `Mapper`. Even worse, we would have to add `ValidEnrollmentParsedRow` etc. which have `ValidCourseData` instead of `CouseData`.

Maybe, as a fun exercise, we could come up with a class decorator that adds a "validated" version of a class into the class namespace, so we would have
```python
@partial_invalid
@dataclass
class CourseData:
    """Holds information about a course, retrieved from an import file."""

    name_de: str
    name_en: str
    degrees: MaybeInvalid[set[Degree]]
    course_type: MaybeInvalid[CourseType]
    is_graded: MaybeInvalid[bool]
    responsible_email: str

def foo(course_data: CourseData):
    bar: CourseData.Validated = course_data.validate()
    # ...
```
... and it recurses into member variables that are annotated with `partial_invalid`. It is clearly overkill, but sounds like a fun project :^)

PR #2077, opened 2023-11-20
Support live server tests with Selenium

### LINKED ISSUE #1854: Use `LiveServerTestCase` for frontend tests

hansegucker:
Close #1854


deployment/provision_vagrant_vm.sh
```diff
@@ -93,6 +93,27 @@ cp $REPO_FOLDER/deployment/manage_autocompletion.sh /etc/bash_completion.d/
 # install chrome, see: https://github.com/puppeteer/puppeteer/issues/7740
 apt-get -q install -y chromium-browser
 
+# install firefox and geckodriver
```
niklasmohrin, line 96: Can you elaborate a bit further? I don't know anything about debian packaging, what is the config below doing?

evap/evaluation/tests/tools.py
```diff
@@ -245,3 +249,53 @@ def assert_no_database_modifications(*args, **kwargs):
...
+class CustomSeleniumTestCaseBase(SeleniumTestCaseBase):
+    external_host = os.environ.get("TEST_HOST", "") or None
+    browsers = ["firefox"]
+    selenium_hub = os.environ.get("TEST_SELENIUM_HUB", "") or None
+    headless = True
+
+    def create_options(self):  # pylint: disable=bad-mcs-method-argument
+        options = super().create_options()
+
+        if self.browser == "chrome":
```
niklasmohrin, line 263: This should never be the case, right? Then let's leave it out of the code so that it doesn't get outdated there (we can still look it up from the PR again if we need it)

evap/evaluation/tests/tools.py
```diff
@@ -245,3 +249,53 @@ def assert_no_database_modifications(*args, **kwargs):
...
+
+
+class LiveServerTest(SeleniumTestCase, metaclass=CustomSeleniumTestCaseBase):
+    def _screenshot(self, name):
+        self.selenium.save_screenshot(os.path.join(settings.BASE_DIR, f"{name}.png"))
+
+    def _create_test_user(self):
+        self.test_user = baker.make(  # pylint: disable=attribute-defined-outside-init
+            UserProfile, email="evap@institution.example.com", groups=[Group.objects.get(name="Manager")]
+        )
```
niklasmohrin, line 281: We also have `make_manager`

evap/evaluation/tests/tools.py
```diff
@@ -245,3 +249,53 @@ def assert_no_database_modifications(*args, **kwargs):
             lower_sql = query["sql"].lower()
             if not any(lower_sql.startswith(prefix) for prefix in allowed_prefixes):
                 raise AssertionError("Unexpected modifying query found: " + query["sql"])
+
+
+class CustomSeleniumTestCaseBase(SeleniumTestCaseBase):
+    external_host = os.environ.get("TEST_HOST", "") or None
+    browsers = ["firefox"]
+    selenium_hub = os.environ.get("TEST_SELENIUM_HUB", "") or None
+    headless = True
```
niklasmohrin, line 258: We are always passing `headless` in `create_options` below; as far as I understand, this option here should take care of this for us, right?

niklasmohrin (commented):
Very cool, I feel that there are some parts that could be cleaned up or need some more explanation otherwise:

niklasmohrin:
Manager group was not present, but only if all tests are run. We found that django will first run `django.test.TestCase` instances, then `django.test.TransactionTestCase` and then `unittest.TestCase`. After the first, the database is cleared. If only running the live server test, the database is still there when running the test.

Link dump if we try to debug this again:

- https://code.djangoproject.com/ticket/23640
- https://code.djangoproject.com/ticket/23727
- https://docs.djangoproject.com/en/5.0/howto/initial-data/#provide-initial-data-with-migrations
- https://docs.djangoproject.com/en/5.0/topics/testing/overview/#test-case-serialized-rollback
- http://atodorov.org/blog/2017/12/26/on-pytest-django-and-liveservertestcase-with-initial-data/

We concluded that we want to migrate an empty database, dump it, and use that dump as a `fixture` in the live server test - at least for now, to continue working on this.

deployment/provision_vagrant_vm.sh
richardebeling: Since January, Mozilla provides their own PPA. We should be able to get firefox up using just:
```
# https://support.mozilla.org/en-US/kb/install-firefox-linux#w_install-firefox-deb-package-for-debian-based-distributions
sudo install -d -m 0755 /etc/apt/keyrings
wget -q https://packages.mozilla.org/apt/repo-signing-key.gpg -O- | sudo tee /etc/apt/keyrings/packages.mozilla.org.asc > /dev/null
echo "deb [signed-by=/etc/apt/keyrings/packages.mozilla.org.asc] https://packages.mozilla.org/apt mozilla main" | sudo tee -a /etc/apt/sources.list.d/mozilla.list > /dev/null
echo '
Package: *
Pin: origin packages.mozilla.org
Pin-Priority: 1000
' | sudo tee /etc/apt/preferences.d/mozilla 
sudo apt-get update && sudo apt-get install firefox
```

I don't know if their repo also contains a matching geckodriver package though. I would hope it does.

(The current code here still has the geckodriver vs firefox version mismatch problem, right?)

evap/evaluation/tests/tools.py
```diff
@@ -245,3 +250,84 @@ def assert_no_database_modifications(*args, **kwargs):
...
                 raise AssertionError("Unexpected modifying query found: " + query["sql"])
+
+
+class LiveServerTest(SeleniumTestCase):
+    external_host = os.environ.get("TEST_HOST", "") or None
+    browser = "firefox"
+    selenium_hub = os.environ.get("TEST_SELENIUM_HUB", "") or None
+    headless = True
+
+    fixtures = [os.path.join(settings.BASE_DIR, "development", "fixtures", "test_with_migrations.json")]
```
richardebeling, line 261: From the documentation. I'd expect [`serialized_rollback`](https://docs.djangoproject.com/en/5.0/topics/testing/overview/#test-case-serialized-rollback) to solve our problems here -- what's the problem with that?

deployment/provision_vagrant_vm.sh
Kakadus: Well no, geckodriver is not jet a part of it. I found https://bugzilla.mozilla.org/show_bug.cgi?id=1876704 tracking it

evap/evaluation/tests/tools.py
hansegucker: Seems to work

deployment/provision_vagrant_vm.sh
hansegucker: From the linked bug report:
> For us it will be important that we are not bound to Firefox release cycles. Compared to Chrome and chromedriver, which require a version match for each release, geckodriver can be used with a variety of Firefox releases at maximum back to the latest ESR release. If there are issues that need to be fixed and released quickly we shouldn't be blocked on getting beta/release approval to uplift patches.

So, for firefox and geckodriver, there is no such problem.

hansegucker:
@Kakadus @niklasmohrin @richardebeling Finally, the last has been migrated and I'm ready for a final review. 🥳

evap/student/tests/test_live.py
```diff
@@ -0,0 +1,157 @@
...
+        self.selenium.get(self.live_server_url + reverse("student:vote", args=[self.evaluation.pk]))
+        WebDriverWait(self.selenium, 10).until(
+            expected_conditions.presence_of_element_located((By.ID, "vote-submit-btn"))
+        ).click()
+
+        for row in self.selenium.find_elements(By.CSS_SELECTOR, "#student-vote-form .row"):
+            try:
+                checkbox = row.find_element(By.CSS_SELECTOR, "input[type=radio][value='2'] + label.choice-error")
+            except NoSuchElementException:
+                continue
```
richardebeling, line 112: This would silently allow this test to "skip" all rows and not assert anything. For what cases is this necessary?

evap/student/tests/test_live.py
```diff
@@ -0,0 +1,157 @@
...
+            checkbox.click()
+
+            assert len(row.find_elements(By.CSS_SELECTOR, ".choice-error")) == 0
+
+    def test_skip_contributor(self):
```
richardebeling, line 117: This check was asserting that the card collapsed before in the TS implementation. I think the check should be possible with the new setup as well, or is there some issue with that?

richardebeling (commented):
Not yet fully reviewed, will have to look through most of the ported tests still, but generally this is looking very nice. Thanks!

evap/staff/tests/test_live.py
```diff
@@ -0,0 +1,111 @@
...
+        contribution1.questionnaires.set([contributor_questionnaire])
+        contribution2.questionnaires.set([contributor_questionnaire])
+
+        self._login(manager)
+
+        self._enter_staff_mode()
+
+        self.selenium.get(self.live_server_url + reverse("staff:evaluation_edit", args=[evaluation.pk]))
+
+        self._screenshot("changes_form_data")
```
richardebeling, line 79: leftover(?) screenshot command -- we don't currently do anything with this, right?

niklasmohrin:
Getting chromium or firefox should be a lot easier now that we have nix, we can work something out together some time if you want :)

Kakadus:
Heey, there is a green check now! :tada: 

@hansegucker do you want to take over and resolve the open threads? If not, I can also finish this : )

Things I fixed:
- we no longer supported `django.contrib.auth.backends.ModelBackend`, I switched to our `evap.evaluation.auth.RequestAuthUserBackend` (since https://github.com/e-valuation/EvaP/pull/2249)
- increased window size so that all elements should be visible now
- compile scss&ts before testing the frontend
- separate firefox into a separate nix shell / CI job

evap/staff/tests/test_live.py
```diff
@@ -0,0 +1,111 @@
...
+        manager_id = self.selenium.execute_script(
+            """
+            const tomselect = document.getElementById("id_contributions-0-contributor").tomselect;
+            const options = tomselect.options;
+            const managerOption = Object.keys(options).find(
+                key => options[key].text == "manager (manager@institution.example.com)",
+            );
+            tomselect.setValue(managerOption);
+            return managerOption;
+        """
```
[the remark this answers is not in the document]
Kakadus: it accepts both :shrug:

evap/staff/tests/test_live.py
Kakadus: not really, we upload them as an artifact. Screenshot diffing would be the next step. Should I remove the screenshot commands and the artifact upload? I think I added the upload to debug the ci previously. I could imagine that the screenshots may make reviews easier?

evap/student/tests/test_live.py
Kakadus: a bit of a problem: this could be a race condition because we check for `collapsing`

evap/student/tests/test_live.py
Kakadus: for the textanswers. I can add a `has:(.btn-check)` in the loop to skip those. If there weren't any, the test still passes silently

evap/staff/tests/test_live.py
richardebeling: Well at least this screenshot currently is not viable for automated screenshot comparison because it includes all kind of random data generated by model_bakery.

While I don't see a big problem of "useless" screenshot commands in there, thinking about devs who stumble across this and then wonder why it is there, I would remove it until we actually need it. It's not like we're losing huge amounts of knowledge here.

evap/student/tests/test_live.py
richardebeling: How about adding the `has .btn-check` condition, (optionally asserting that there is at least one row), then working on the first row only. No for-loop, no try. Seems simpler to me?

evap/student/tests/test_live.py
richardebeling: If this really is an issue (Wouldn't it be optimal if the browser was halted whenever test driver code was executed? Idk) we could check for (`collapsing` or `collapsed`) so that if we missed collapsing we'd still succeed

evap/student/tests/test_live.py
Kakadus: i think this was actually the ts test

evap/evaluation/tests/tools.py
```diff
@@ -29,6 +36,13 @@
 )
 
 
+class SkipLiveServerTestsRunner(DiscoverRunner):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        if not self.tags and not self.exclude_tags:
+            self.exclude_tags = {"live-server"}
```
richardebeling, line 43: If someone uses the `SkipLiveServerTestsRunner` and passes `tags` or `exclude_tags` here, it doesn't skip live server tests anymore. I'd either expect us to just add "live-server" to `exclude_tags`, or at least assert that neither `tags` or `exclude_tags` was set (if we never expect this to happen). The current approach of continuing but with behavior that doesn't match the class name, to me, is just a trap we set.

evap/evaluation/tests/tools.py
```diff
@@ -273,3 +257,75 @@ def assert_no_database_modifications(*args, **kwargs):
...
                 raise AssertionError("Unexpected modifying query found: " + query["sql"])
+
+
+@tag("live-server")
+class LiveServerTest(SeleniumTestCase):
+    external_host = os.environ.get("TEST_HOST", "") or None
+    browser = "firefox"
+    selenium_hub = None
+    headless = True
+    window_size = (1920, 3080)
```
richardebeling, line 268: 3080 intended? Or should it be 1080?

evap/evaluation/tests/tools.py
```diff
@@ -273,3 +257,75 @@ def assert_no_database_modifications(*args, **kwargs):
...
+    serialized_rollback = True
+
+    def _create_test_user(self):
+        """Create a default test user."""
+        test_user = baker.make(
+            UserProfile, email="evap@institution.example.com", groups=[Group.objects.get(name="Manager")]
+        )
+        test_user_password = "evap"  # nosec
+        test_user.set_password(test_user_password)
+        test_user.save()
```
richardebeling, line 279:
```suggestion
        """Create a default test user."""
        test_user = baker.prepare(
            UserProfile, email="evap@institution.example.com", groups=[Group.objects.get(name="Manager")]
        )
        test_user.set_password("evap")  # nosec
        test_user.save()
```
doesn't hit the database twice

evap/staff/tests/test_live.py
```diff
@@ -0,0 +1,88 @@
...
+
+        form_data = self.selenium.execute_script(
+            """
+                return Object.fromEntries(new FormData(document.getElementById("evaluation-form")));
+            """
+        )
+        self.assertEqual(form_data["contributions-0-contributor"], manager_id)
+        self.assertEqual(form_data["contributions-0-order"], "0")
+        self.assertEqual(form_data["contributions-0-role"], "1")
+        self.assertEqual(form_data["contributions-0-textanswer_visibility"], "GENERAL")
```
richardebeling, line 88: Could we just submit and then assert on the database values? Would make the test more stable against field field ID changes (that don't need to be stable)

richardebeling (commented):
Looking good! I especially like the changes to package.json and package-lock.json.

Haven't re-checked yet whether all ported tests assert the same things as the old tests, will need to re-do that in a final review.

evap/evaluation/tests/tools.py
Kakadus: I think adding `--tags live-server` to start the live servers is reasonable. To mitigate the other stuff, I think this is most intuitive
```suggestion
        if "live-server" not in self.tags:
            self.exclude_tags.add("live-server")
```

Kakadus: Maybe also a global for the `"live-server"` tag?

evap/evaluation/tests/tools.py
Kakadus: yes, the height should be large enough so that there is no scrolling needed. Maybe a comment?

evap/staff/tests/test_live.py
```diff
@@ -36,16 +37,16 @@ def test_changes_form_data(self):
             evaluation=evaluation,
             contributor=responsible,
             order=0,
-            role=Contribution.Role.EDITOR,
-            textanswer_visibility=Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS,
+            role=Contribution.Role.CONTRIBUTOR,
+            textanswer_visibility=Contribution.TextAnswerVisibility.OWN_TEXTANSWERS,
```
Kakadus, line 41: I changed this so that the values selected in the UI are not already selected. I found out this was intended to be a test for #1769 but I could not find there that this is required for this

Kakadus:
I'm going to rebase this and sort the changes

evap/evaluation/tests/tools.py
richardebeling: Yeah, I'd add a comment, since 1080->3080 could easily be a typo. Note that we're never guaranteed that we don't need any scrolling, the result pages can easily be dozens of screens long. But I see that something around 4k makes our lives easier here

evap/evaluation/tests/tools.py
richardebeling: > I think adding --tags live-server to start the live servers is reasonable

We're in a wonky world then, though, aren't we? Using the `SkipLiveServerTestsRunner` to run tests tagged with `live-server`... I see that this is what you currently do on CI (using ` manage.py test --tag live-server`). Is that also the invocation that we expect developers to use locally?

In the end fine with me if there is no other way to have django run specific. Maybe rename this runner then to just `EvapTestRunner`? The "add the exclude tag if it isn't explicitly included" behavior looks fine to me, then.

I don't think we need a constant for the tag, using the string is fine.

evap/static/ts/src/copy-to-clipboard.ts
```diff
@@ -1,19 +1,3 @@
-function copyToClipboard(text: string) {
```
richardebeling, line 1: Don't we still have the [usage in staff_template_form](https://github.com/e-valuation/EvaP/blob/70434ef3391e643cdc4ae4ad4bfd04c4147bdaa0/evap/staff/templates/staff_template_form.html#L47)?

evap/results/tests/test_live.py
```diff
@@ -0,0 +1,179 @@
...
+        def make_winter_semester(year):
+            return baker.make(
+                Semester,
+                name_de=f"Wintersemester {year}/{year + 1}",
+                name_en=f"Winter term {year}/{year + 1}",
+                short_name_de=f"WS {year % 1000}/{year % 1000 + 1}",
+                short_name_en=f"WT {year % 1000}/{year % 1000 + 1}",
+            )
+
+        def make_summer_semester(year):
```
richardebeling, line 26: maybe add type annotations

evap/static/ts/tests/frontend/staff-evaluation-edit.ts
```diff
@@ -1,40 +0,0 @@
...
-test(
-    "changes form data",
-    pageHandler("/staff/semester/PK/evaluation/PK/edit/normal.html", async page => {
-        const managerId = await page.evaluate(() => {
-            const tomselect = (document.getElementById("id_contributions-0-contributor") as any).tomselect;
-            const options = tomselect.options;
-            const managerOption = Object.keys(options).find(
-                key => options[key].text == "manager (manager@institution.example.com)",
-            );
-            tomselect.setValue(managerOption);
```
niklasmohrin, line 17: Is this selection important? I can't find it in the Python version

evap/evaluation/tests/tools.py
```diff
@@ -265,11 +265,12 @@ class LiveServerTest(SeleniumTestCase):
     selenium: WebDriver
     headless = True
     window_size = (1920, 3080)
-    serialized_rollback = True
+    serialized_rollback = False
```
niklasmohrin, line 268: Why do we explicitly disable this again? Is there anything we should write down for future reference?

niklasmohrin (commented):
Very cool! After ensuring that I follow the steps from CI, they also passed locally :)

evap/evaluation/tests/tools.py
niklasmohrin: I am also fine with odd dimensions (and a comment) for now. We should revisit scrolling at some point though

evap/static/ts/tests/frontend/staff-evaluation-edit.ts
niklasmohrin: Ah, I suppose this is due to https://github.com/e-valuation/EvaP/pull/2077#discussion_r1922778892 ?

evap/evaluation/tests/tools.py
Kakadus: django said that serialized rollbacks may slow down tests up to 3x. I tried that one out, but honestly I did not see a performance difference. The drawback of `serialized_rollback = False` is that we need to add the manager group ourselves (and maybe more stuff in the future).

I think we can even enable serialized rollbacks until we see a performance hit.

evap/static/ts/tests/frontend/staff-evaluation-edit.ts
Kakadus: no, the comment is unrelated. I'm looking into it..

evap/evaluation/tests/tools.py
niklasmohrin: I think that we either a) don't care about the setting, in which case it shouldn't be mentioned, or b) do care, in which case we should add a short comment why

evap/static/ts/tests/frontend/staff-evaluation-edit.ts
Kakadus: I think this got replaced by a simple managerId update. Reading the referenced PR, I think that it may be important. I check if we can have something similar in python with selenium.

evap/evaluation/tests/tools.py
richardebeling: For reference: It seems like `serialized_rollback` is what I dug up from the docs: https://github.com/e-valuation/EvaP/pull/2077#discussion_r1486675190, although I didn't remember that either. Short comment in the code sure would help. Proposal:
> `TransactionTestCase` drops any data created by migrations, which we need for groups etc., so we need to fall back to `TestCase` and use `serialized_rollback`. See https://docs.djangoproject.com/en/5.0/topics/testing/overview/#test-case-serialized-rollback for details.

richardebeling (commented):
@Kakadus you re-requested review but some of the older review discussion points are still open. I'm generally fine with the code once these are resolved.

evap/evaluation/tests/tools.py
Kakadus: I now added 
> SeleniumTestCase is a TransactionTestCase, which drops migration data. This keeps fixture data but may slow down tests, see https://docs.djangoproject.com/en/5.0/topics/testing/overview/#test-case-serialized-rollback

evap/static/ts/tests/frontend/staff-evaluation-edit.ts
Kakadus: done something similar now that should be equivalent :+1:

richardebeling (approved):

.github/workflows/tests.yml
```diff
@@ -175,62 +203,6 @@ jobs:
...
-        with:
-          flags: render-pages
-          token: ${{ secrets.CODECOV_TOKEN }}
-      - name: Store rendered pages
-        uses: actions/upload-artifact@v4
-        with:
-          name: rendered-pages
-          path: evap/static/ts/rendered
-
-  typescript:
```
niklasmohrin, line 202: Don't we still have unit tests in Typescript?

niklasmohrin (commented):
One last thing, everything else looks good :)

.github/workflows/tests.yml
Kakadus: done

niklasmohrin (approved):
Nice!

Kakadus (approved):

PR #2078, opened 2023-11-20
Fix generation of JS translations by dropping old locale folder

hansegucker:


evap/static/js/.gitignore
```diff
@@ -2,3 +2,4 @@
 *.js
 !*.min.js
 !sortable_form.js
+!moment_de.js
```
richardebeling, line 5: Unless there' value in these two entries here (which I currently don't see), I'd consider this duplication and thus rather have these removed
```suggestion
```

evap/locale/de/LC_MESSAGES/djangojs.po
```diff
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: EvaP\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-10-30 17:26+0100\n"
+"POT-Creation-Date: 2023-11-20 21:26+0100\n"
```
richardebeling, line 9: Not a relevant change -- but also doesn't really matter

```suggestion
"POT-Creation-Date: 2023-10-30 17:26+0100\n"
```

evap/development/management/commands/translate.py
```diff
@@ -9,6 +9,10 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
         self.stdout.write('Executing "manage.py makemessages --locale=de --ignore=node_modules/*"')
         call_command("makemessages", "--locale=de", "--ignore=node_modules/*")
+        self.stdout.write(
```
richardebeling, line 12: Can you add a `logged_call_command` to `commands/tools.py` and use that here, something along the lines of
```python3
def logged_call_command(stdout, *args):
    stdout.write("Executing " + " ".join(args))
    call_command(*args)
```
so we don't have the arguments duplicated (and thus possibly diverging in the future)?

richardebeling (commented):
Looking good, thanks!

hansegucker:
🥳

Issue #2072, opened 2023-11-20
Clarifying wording in installation readme


CurtisHolick:
I am a student, taking a class where we are exposed to open source projects, and encouraged to attempt to contribute to one. While attempting to setup my development environment, I ran in to a few errors, one of which I believe could be remedied with a change to the wording in the readme's installation section.

I was using git bash as directed to change the setting for line endings, but when I continued, I spent a large amount of time troubleshooting why I was unable to get the next steps of starting vagrant and, SSHing, and running the manage.py file. I had interpreted the term 'command line' to be a generic term, implying you would continue imputing the commands into git bash, or an alternative you had used.

I have seen from past issues there is a desire to keep the onboarding readme as short and simple as possible. However, I feel that rewording or adding a word or two to step 3 about using specifically the windows command _prompt_ would add clarity and help others avoid this pitfall. 

My specific suggestion is to switch the term 'command line' for 'command prompt' or 'windows command prompt', as windows refers to it.

niklasmohrin:
Hi! I am not sure I understand the motivation for your suggested change. After changing the setting in Git bash, you should be able to (and are supposed to) continue entering the next commands in this Git bash. Can you describe a particular problem that occurred in your setup?

---

The term "command line" is actually somewhat generic here - these commands should work from any of the standard methods to operate your computer by entering text commands (Windows `cmd`, Windows Powershell, Windows Git bash, Macos Terminal, any terminal on Linux, even in a non-graphical Linux session). Since the readme should actually cover all of these platforms, changing the wording to something like "windows command prompt" is a definite no-go and I don't see why we would emphasize that the windows program in particular is called "command prompt".

If there is anything tricky that we could add as a hint for a specific platform, we could think about that. However, I feel that if the only hurdle is having to understand what it means to "run a command on the command line", then my gut feeling would be that we leave this for the reader to figure out - they will have to understand it anyways once they start to work in the development VM.

richardebeling:
We tried the setup procedure from the readme using Git Bash on Windows on two machines today. On one of the machines, we encountered an issue where the `vagrant ssh` command would not correctly give an interactive pseudo-terminal with a command prompt, i.e., `(venv) evap@ubuntu-jammy:/evap $ ` would not appear and it thus looks like the execution hangs:
```
user@machine MINGW64 ~/Desktop/EvaP (main)
$ vagrant ssh
[...]

PR #2079, opened 2023-11-21
Improve windows setup process

richardebeling:
* Don't require manual fiddling with `autocrlf`, instead just enforce it for the repo using a `.gitattributes` file. This means that users can no longer easily clone to CRLF if they intend to, but I think it's fair for us to do.
* Enable shallow cloning for the bootstrap submodule (saves ~200MB of network traffic)
* Add a workaround for ssh not allocating a tty in Git Bash (run ssh with `-tt`)
* Don't recommend Git Bash as it behaves weirdly with the vagrant ssh connection with that workaround
     * It doesn't allow auto completion with TAB
     * Pressing Ctrl+C terminates the SSH connection
     * Colors do not work

Open for suggestions regarding the readme content and/or phrasing. You can find the rendered version at https://github.com/richardebeling/EvaP/blob/main/README.md.

TODO:
* [x] Re-test setup on linux (make sure that nothing was broken)
* [x] Do we want to replace the clone uri template (currently https) with a ssh template?
    * Https allows quick and easy access to the source code and people can get started with EvaP quickly
    * However, in order to push, I think it makes most sense to just setup an ssh key since GitHub doesn't allow password login anymore
    * Properly setting up an ssh key requires a few considerations that might overwhelm new contributors (passphrase, where to store, ssh-agent), so I think it makes sense to defer that ssh setup to when it is required. People will have to update their remote, but that seems tolerable to me.
    * We could hint at cloning with ssh for users that already have a key set up.
    
related to #2072


richardebeling:
* Bootstrap: I think nothing really changes here. I made a [main-updated-boostrap](https://github.com/richardebeling/EvaP/tree/main-updated-bootstrap) branch to play around with:
    After switching from main to main-with-updated-bootstrap, you get
    ```
    $ git status
    [...]
    modified:   evap/static/bootstrap (new commits)
    ```
    And you can update using
    ```
    $ git submodule update
    Submodule path 'evap/static/bootstrap': checked out '344e912d04b5b6a04482113eff20ab416ff01048'
    ```
    Since we usually update to newer versions of bootstrap, the client will have to fetch new information from the bootstrap repo anyways. "Shallow" only affects (for the average developer useless) information about past versions.
    
* Git Bash: I created issues with git-for-windows about broken behavior with `vagrant ssh` both with their bundled OpenSSH as well as the Windows-provided OpenSSH. I don't think the `-tt` will realistically cause harm though. You'd have to pipe input through `vagrant ssh` in order for it to break things. I'd slightly tend towards just leaving it in for now and seeing how the issues play out. Not a strong opinion though.

* Forking, Pushing, SSH: I like the idea of splitting it up and removing the forking from the initial quick setup. We make people always clone from e-valuation with HTTPS. Further down in the Readme, we can have a new section "Submitting your changes" or something, which would include these steps and explain them:
    * Create the fork through GitHub's UI
    * Set up SSH keys
    * `git remote set-url origin git@github.com:<username>/EvaP.git`
    * (maybe) `git remote add upstream git@github.com:e-valuation/EvaP.git` (and possibly update main?)
    * `git switch -c <branch_name>`
    * git add/commit
    * `git push -u origin <branch_name>`
    * Go to GitHub, click "Create Pull-Request" button on the huge warning-color banner

    Nice about this:
    1. We have documented the command to create the `upstream` remote.
    2. We tell people to create a new branch so they don't clobber their main branch.
    3. It moves lots of the Git explanation away from the introduction, so we can skip it there and save time. People are only "bothered" by this once they have successfully improved something. Doing this interactively with first time git users makes sense anyway, as we can spot errors much better.
    
    I'm not sure how in-depth I want our explanations about the whole fork-pr-workflow and especially the SSH key setup to be. I think I'd rather link to external resources. People will have to figure out their handling of private key passphrases, ssh-agents, integration with password managers, etc. depending on their systems and where they want to be on the security/usability spectrum.

Kakadus (approved):
Some parts in the readme are very introductory, but this may be helpful in addition to our presentation for first semesters

README.md
```diff
@@ -55,6 +53,22 @@ or, to combine all three, simply run `./manage.py precommit`.
...
 
+### Creating a Pull Request (Workflow Suggestion)
+1. (once) [Fork](https://github.com/e-valuation/EvaP/fork) the repository so you have a GitHub Repo that you have write access to.
+
+2. (once) Set up some Authentication for GitHub that allows push access. A common option is using [SSH keys](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/about-ssh), the remaining instructions assume an SSH key setup. An alternative is using the [GitHub CLI tool](https://cli.github.com/).
+
+3. (once) Ensure your [git remotes](https://git-scm.com/book/en/v2/Git-Basics-Working-with-Remotes) are setup to use SSH. To fetch the up-to-date state of the official repo, it's always useful to have an "upstream" remote configured:
+   ```bash
+   git remote set-url origin git@github.com:<your-username>/EvaP.git
+   git remote add upstream git@github.com:e-valuation/EvaP.git
```
niklasmohrin, line 64: Huh, when did they change that `set-url` actually sets both fetch and push urls? I thought I remembered that it was odd that you always had to run it once without and once with `--push` to actually set both urls, but this here works also for me :D

I wonder if we should maybe provide a more idempotent set of commands here:
```bash
git remote remove origin
git remote add origin git@github.com:<your-username>/EvaP.git
git remote add upstream git@github.com:e-valuation/EvaP.git
```

README.md
```diff
@@ -55,6 +53,22 @@ or, to combine all three, simply run `./manage.py precommit`.
...
+   ```
+
+4. Create a branch (`git branch -c <your-branch-name>`), commit your changes (`git add` and `git commit`), and push them (`git push`). "Push" will ask you to specify an upstream branch (`git push -u origin <your-branch-name>`).
```
niklasmohrin, line 67: Did you mean this?
```suggestion
4. Create a branch (`git switch -c <your-branch-name>`), commit your changes (`git add` and `git commit`), and push them (`git push`). "Push" will ask you to specify an upstream branch (`git push -u origin <your-branch-name>`).
```

README.md
```diff
@@ -55,6 +53,22 @@ or, to combine all three, simply run `./manage.py precommit`.
...
+
+5. GitHub should now ask you whether you want to open a Pull Request when reloading the page. If the PR solves an issue, use one of GitHub's [magic keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) (like "fixes") in the commit message or pull request text to create a link between your PR and the issue.
```
niklasmohrin, line 69: "If necessary, please also provide a short summary of your changes in the description of your Pull Request."? Or do we not want to omit this in favor of brevity?

niklasmohrin, line 69: "when reloading the page" -> which page are we talking about here? How about "on the Github page of your fork" or so?

"pull request text" -> "pull request description"? (I would also prefer to not encourage users to mention issue numbers in commits)

niklasmohrin (commented):
Nice

README.md
richardebeling: > Huh, when did they change that set-url actually sets both fetch and push urls? I thought I remembered that it was odd that you always had to run it once without and once with --push to actually set both urls, but this here works also for me :D

Fair point. The way I understand the documentation, `set-url` without any further arguments will simply change the "first" url configured for the remote, and perhaps this is the push and pull url with the trivial setup after cloning? That would mean you'd have to manually set the push url once you specified a separate push url -- that would make sense to me.

I didn't want to mess with the origin branch in any way that could mess up the upstream-branch configuration for main, that's why I wanted to keep origin around but re-point it, but I see the argument about idempotency. How does a branch "remember" its upstream branch, is it just by name, or is it messed up after removing the remote and adding a new remote with the same name?

README.md
richardebeling: The sentence starts with "GitHub", so that's probably fine? We can also completely omit the "when reloading the page", users will probably figure that out sooner or later.

Yeah sounds good. Updated version:
```suggestion
5. GitHub should now ask you whether you want to open a Pull Request. If the PR solves an issue, use one of GitHub's [magic keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) (like "fixes") in the pull request description to create a link between your PR and the issue.
```

README.md
```diff
@@ -55,6 +53,22 @@ or, to combine all three, simply run `./manage.py precommit`.
 
 You can also set up `pylint`, `isort`, `black` and `prettier` in your IDE to avoid doing this manually all the time.
 
+### Creating a Pull Request (Workflow Suggestion)
+1. (once) [Fork](https://github.com/e-valuation/EvaP/fork) the repository so you have a GitHub Repo that you have write access to.
+
+2. (once) Set up some Authentication for GitHub that allows push access. A common option is using [SSH keys](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/about-ssh), the remaining instructions assume an SSH key setup. An alternative is using the [GitHub CLI tool](https://cli.github.com/).
```
richardebeling, line 59: Note to self: Recheck capitalization for the whole file
```suggestion
1. (once) [Fork](https://github.com/e-valuation/EvaP/fork) the repository so you have a GitHub repo that you have write access to.

2. (once) Set up some authentication for GitHub that allows push access. A common option is using [SSH keys](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/about-ssh), the remaining instructions assume an SSH key setup. An alternative is using the [GitHub CLI tool](https://cli.github.com/).
```

README.md
richardebeling: Removing and re-adding the remote breaks the branch upstream configuration of existing branches. `set-url` should be as idempotent as remove-add. The main disadvantage of `set-url` is if multiple URLs are configured, but we don't expect that here (otherwise the user probably knows how remotes work).

Left as `set-url` for now. We'll see if we run into any trouble with it.

Issue #1945, opened 2023-05-22
Fix button group borders


janno42:
In several places we have button groups where not all `btn`s are next to each other in the HTML code, so the CSS styling gets a bit weird. Either the border radii don't match (when setting the border on the grouping element) or the inner borders, where two buttons touch each other, are doubled (when setting the border on the buttons).
Some examples:
- language selection buttons in the navbar (border on grouping element)
- decision buttons and flag buttons on the text answer review page (border on buttons)

The expected styling would make sure that the radii match and that the inner borders are not doubled.

richardebeling:
As discussed online, all to the best of my knowledge: Yes, that button group is one that the issue refers to.
In the current version, afaik, we only have the "border radius" issue, i.e. the button is not fully "filled" to the border. If that is fixed, there shouldn't be any need for changes to the border, I think the border-issue only occurred in another attempt to fix this border-radius issue.

(For reference: This is what is currently rendered on main for me, the borders look fine, but there is a very subtle border radius filling issue):
![image](https://github.com/e-valuation/EvaP/assets/13838962/f69722f7-c45d-44d6-b06c-28e9225b1525)

fkoch-tgm:
There are two cases for this issue:

## 1. The buttons are nested inside form-tags (Navbar)
I fixed those buttons by removing all borders and border radii from the .btn and moving it into the .btn-group. This is done:

![image](https://github.com/e-valuation/EvaP/assets/45557700/a3de2489-b1d9-4477-8feb-806ae4dc420f)



## 2. General `box-shadow`s inside rounded corners:
It seems like if you have a inset `box-shadow` inside a button with rounded corners, it will not be rendered correctly in Firefox:
![image](https://github.com/e-valuation/EvaP/assets/45557700/a157d0ba-40e3-4750-89e6-a537e435ff80)


It seems like this is not an issue with Chromium:
![image](https://github.com/e-valuation/EvaP/assets/45557700/c319dccf-64a6-4acc-8353-371a8f29d8eb)

As this issue is with the rendering of the HTML itself, the only solution I can think of, is to switch to `background-image: radial-gradient(...)` instead.

janno42:
It looks like parts of the original issue don't occur anymore. Sorry that we didn't include screenshots. The button's "filling" didn't reach the border in the corners, visually impacting even when not zooming in too much. This now looks fine for me in both Firefox and Chrome.

[...]

PR #2085, opened 2023-12-04
Fix Borders in Outlined Button Group in textanswers view

fkoch-tgm:
Fixes double borders referenced [here](https://github.com/e-valuation/EvaP/issues/1945#issuecomment-1819526342).

This needs the `:has()` pseudoselector, which will be supported in the next firefox release [121](https://bugzilla.mozilla.org/show_bug.cgi?id=1858743) (planned release date: 2023-12-19)

grafik

### DIFF
--- a/evap/static/scss/_adjustments.scss
+++ b/evap/static/scss/_adjustments.scss
@@ -140,7 +140,7 @@ $alert-colors: (
 
     // If .btn is not the last element in the group, remove the right border.
     // Also uses :has to not remove the right border if the next element is a disabled input
-    .btn:not(:last-child):not(.dropdown-toggle):has(+ :not(.btn-check[disabled])), .btn-group:not(:last-child) > .btn {
+    .btn:not(:last-child):has(+ :not(.btn-check[disabled])) {
         border-right: 0;
     }
 
@@ -149,9 +149,8 @@ $alert-colors: (
         border-left: 0;
     }
 
-    // Revert Bootstrap (Prevent double borders when buttons are next to each other)
-    > :not(.btn-check:first-child) + .btn,
-    > .btn-group:not(:first-child) {
+    // Revert Bootstrap's hack, that uses negative margin which would appear as double borders
+    :not(.btn-check:first-child) + .btn {
         margin-left: unset;
     }
 }
--- a/evap/static/scss/_adjustments.scss
+++ b/evap/static/scss/_adjustments.scss
@@ -133,3 +133,25 @@ $alert-colors: (
     border-color: $input-focus-border-color;
     box-shadow: $input-focus-box-shadow;
 }
+
+// Change .btn-group outlined border style to fix double borders with outlined button style
+// see https://github.com/twbs/bootstrap/issues/25888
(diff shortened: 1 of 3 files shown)


evap/static/scss/components/_button-group.scss
```diff
@@ -17,3 +17,25 @@ div.vote-inputs > textarea {
...
+    .btn-check[disabled] + .btn {
+        border-left: 0;
+    }
+
+    // Revert Bootstrap (Prevent double borders when buttons are next to each other)
+    > :not(.btn-check:first-child) + .btn,
+    > .btn-group:not(:first-child) {
+        margin-left: unset;
+    }
+}
```
richardebeling, line 41: missing newline

richardebeling (commented):
Code looks good to me, will have to test with firefox once 121 is released.

evap/static/scss/components/_button-group.scss
janno42, line 1: we should probably put this into `_adjustments.scss` to the other bootstrap styling fixes

fkoch-tgm:
@richardebeling I tried it using the new firefox and could not reproduce the issue anymore.

janno42 (approved):
Beautiful :)

richardebeling (approved):

niklasmohrin:
Just so we are on the same page: This is just a special case of #1945, right? Otherwise, we should link the PR and issue

evap/static/scss/_adjustments.scss
```diff
@@ -133,3 +133,25 @@ $alert-colors: (
...
     box-shadow: $input-focus-box-shadow;
 }
+
+// Change .btn-group outlined border style to fix double borders with outlined button style
+// see https://github.com/twbs/bootstrap/issues/25888
+.btn-group.outline-fix {
+
+    // If .btn is not the last element in the group, remove the right border.
+    // Also uses :has to not remove the right border if the next element is a disabled input
+    .btn:not(:last-child):not(.dropdown-toggle):has(+ :not(.btn-check[disabled])), .btn-group:not(:last-child) > .btn {
```
niklasmohrin, line 143: As far as I can tell, `.outline-fix .btn.dropdown-toggle` never matches on evap. I also don't know what it is supposed to fix

niklasmohrin, line 143: Similar story: We don't have any `.outline-fix .btn-group`, so the part after the comma never matches either

niklasmohrin (changes requested):
The scss is very long, I still have questions :D

fkoch-tgm:
@niklasmohrin I removed all unused CSS selectors.

I also checked again, if the issue still occurs on main and I could verify that there are double borders that happen on main but not on this branch.

niklasmohrin (approved):

niklasmohrin:
@janno42 Can you confirm that this closes everything from #1945?

janno42 (approved):
works for me and i'd say we can close #1945 then

richardebeling (approved):

PR #2086, opened 2023-12-04
fix: vagrant permissions

fidoriel:


deployment/provision_vagrant_vm.sh
```diff
@@ -37,11 +37,10 @@ service redis-server restart
 apt-get -q install -y apache2 apache2-dev libapache2-mod-wsgi-py3
 
 # With docker for mac, root will own the mount point (uid=0). chmod does not touch the host file system in these cases.
-OWNER=$(stat -c %U "$MOUNTPOINT/evap")
-if [ "$OWNER" == "root" ]; then chown -R 1042 "$MOUNTPOINT"; fi
+if [ $(stat -c "%u" "$MOUNTPOINT/evap") == "root" ]; then chown -R 1042 "$MOUNTPOINT"; fi
 
 # make user, create home folder, set uid to the same set in the Vagrantfile (required for becoming the synced folder owner), set default shell to bash
-useradd -m -u $(stat -c "%u" "$MOUNTPOINT/evap") -s /bin/bash evap
+useradd -m -u 1042 -s /bin/bash evap
```
richardebeling, line 43: breaks ubuntu docker setup for me. With native docker, the folder is owned by the user with id 1000. For me, with your setup:
```
$ stat -c "%u" "/evap"
1000  # vagrant mount did not use the `mount_options` we specified
$ sudo cat /etc/passwd
-> root has user id 0  # as expected
```
but, due to this change, the user `evap` doesn't own `/evap` anymore, and thus all file accesses to that folder fail.


I think the `$(stat -c "%u" "$MOUNTPOINT/evap")` here was added because vagrant doesn't correctly apply the mount options of uid and gid that we provide in the Vagrantfile. In any case, whether this folder is mounted for uid 1000 or 1042 or any other value, we want the evap user to have that user id. Going back to hardcoding this value doesn't seem to be an option to be, considering this background.

richardebeling (changes requested):

niklasmohrin:
In addition, I don't understand what problem we are solving in the first place :sweat_smile:

deployment/provision_vagrant_vm.sh
```diff
@@ -37,11 +37,10 @@ service redis-server restart
 apt-get -q install -y apache2 apache2-dev libapache2-mod-wsgi-py3
 
 # With docker for mac, root will own the mount point (uid=0). chmod does not touch the host file system in these cases.
-OWNER=$(stat -c %U "$MOUNTPOINT/evap")
-if [ "$OWNER" == "root" ]; then chown -R 1042 "$MOUNTPOINT"; fi
+if [ $(stat -c "%u" "$MOUNTPOINT/evap") == "root" ]; then chown -R 1042 "$MOUNTPOINT"; fi
```
richardebeling, line 40: For your setup where this doesn't work,
* what does `stat -c "%u" "$MOUNTPOINT/evap"` return here? If it doesn't return "root", maybe we could instead check if the owning-user-id is 0?
* does `chown -R 1042 "$MOUNTPOINT"` work, so that the mounting point is afterwards owned by user id 1042 when viewed from inside the docker container?

Also: What vagrant and docker versions are you using?

Kakadus:
Since the previous check in this line is also relevant for docker rootless (unsupported for EvaP), I'll throw in my experience with it. Feel free to skip to the last paragraph if you are only interested in a possible solution. This may not apply to plain docker on Mac.

I've investigated my (unsupported) setup with docker rootless, which does some uid mapping very similar to [`userns-remap` mode](https://docs.docker.com/engine/security/userns-remap/) and the mapping specified by Vargrant seems to be ignored. That means, that _only_ uid `0` in the container is mapped to my host user uid (let's say `1000`). Every other conainer uid is mapped to host uid with an offset of 99999. So, if I `chown evap` the mountpoint in the container with evap having `1042`, everything is mapped to `101041`. [This bug is a](https://rootlesscontaine.rs/how-it-works/userns/) [feature of docker rootless](https://rootlesscontaine.rs/getting-started/common/subuid/). 

It would be interesting to know, what happens on mac. What happens after you `chown`ed the files in the container? What is the output of `ls -lan` both in the container and on the host? The comment `chmod does not touch the host file system in these cases` implies it cannot be the same behaviour as with docker rootless.

The simplest fix I just came up with, is giving `evap` the uid `0` in the container. This can be achieved by adding `--non-unique` to useradd (did not know about that flag before), which does not break the flow with different providers and seems to work fine with docker rootless at l(e)ast. Feel free to try out my branch at https://github.com/e-valuation/EvaP/compare/main...Kakadus:EvaP:provision-docker-rootless

richardebeling:
> The simplest fix I just came up with, is giving evap the uid 0 in the container.

This would effectively make our user `root`, right? One point of having the additional non-root user in the VM is that we can test deployment scripts, and one aspect of that is finding potentially missing `sudo` calls or other permission errors. If the VM user becomes root, we wouldn't have that anymore.

Kakadus:
Next year, next try: We can use [bindfs](https://github.com/mpartel/bindfs), which can map uids! Sadly, it uses [FUSE](https://github.com/libfuse/libfuse). This is possible in docker containers, if the `/dev/fuse` device is forwarded. This would essentially require fuse to be installed for all users for the docker provider. @fidoriel, does this work? https://github.com/e-valuation/EvaP/compare/main...Kakadus:EvaP:provision-docker-rootless-2

fkoch-tgm:
It seems to me that we want to check if the id of the owner of the mountpoint and if it is *not* our user id `1042`, then we have a problem.

Using a more generalized version of @Kakadus fix, I got it to work using Libvirt/QEMU as a Vagrant provider:

```bash
OWNER=$(stat -c %u "$MOUNTPOINT/evap")

# if, for any reason, vagrant failed to mount evap under the correct uid, use bindfs to link it to /opt/evap using the wanted user id
if [ $OWNER != 1042 ]; then
  apt-get -q install -y bindfs && \
  mkdir -p "$REPO_FOLDER"
  bindfs --map=$OWNER/1042:@$OWNER/@1042 "$MOUNTPOINT" "$REPO_FOLDER"
  echo "sudo bindfs --map=$OWNER/1042:@$OWNER/@1042 '$MOUNTPOINT' '$REPO_FOLDER'" >> /home/$USER/.bashrc
else
  ln -s "$MOUNTPOINT" "$REPO_FOLDER"
fi
```

fkoch-tgm:
I opened a new PR #2105 , @fidoriel does this also fix your mounting issue?

PR #2091, opened 2023-12-11
decompose RatingResult

Kakadus:
small rewrite of RatingResult, so that it is split up to PublishedRatingResult and AnsweredRatingResult

### DIFF
--- a/evap/results/tools.py
+++ b/evap/results/tools.py
@@ -43,7 +43,7 @@ class TextAnswerVisibility:
 
 def create_rating_result(question, answer_counters, additional_text_result=None):
     if answer_counters is None:
-        return RatingResult(question, answer_counters, additional_text_result)
+        return RatingResult(question, additional_text_result)
     if any(counter.count != 0 for counter in answer_counters):
         return AnsweredRatingResult(question, answer_counters, additional_text_result)
     return PublishedRatingResult(question, answer_counters, additional_text_result)
@@ -70,7 +70,7 @@ class RatingResult:
 
 class PublishedRatingResult(RatingResult):
     def __init__(self, question, answer_counters, additional_text_result=None):
-        super().__init__(question, answer_counters, additional_text_result)
+        super().__init__(question, additional_text_result)
         counts = OrderedDict((value, 0) for value in self.choices.values if value != NO_ANSWER)
         for answer_counter in answer_counters:
             counts[answer_counter.answer] = answer_counter.count
--- a/evap/results/tools.py
+++ b/evap/results/tools.py
@@ -2,7 +2,7 @@ from collections import OrderedDict, defaultdict
 from collections.abc import Iterable
 from copy import copy
 from math import ceil, modf
-from typing import cast
+from typing import TypeGuard, cast
 
 from django.conf import settings
 from django.core.cache import caches
@@ -50,8 +50,13 @@ def create_rating_result(question, answer_counters, additional_text_result=None)
 
 
 class RatingResult:
-    is_published = False
-    has_answers = False
+    @classmethod
+    def is_published(cls, rating_result) -> TypeGuard["PublishedRatingResult"]:
+        return isinstance(rating_result, PublishedRatingResult)
+
+    @classmethod
+    def has_answers(cls, rating_result) -> TypeGuard["AnsweredRatingResult"]:
+        return isinstance(rating_result, AnsweredRatingResult)
 
     def __init__(self, question, _, additional_text_result=None):
         assert question.is_rating_question
@@ -64,8 +69,6 @@ class RatingResult:
 
 
 class PublishedRatingResult(RatingResult):
-    is_published = True
-
     def __init__(self, question, answer_counters, additional_text_result=None):
         super().__init__(question, answer_counters, additional_text_result)
         counts = OrderedDict((value, 0) for value in self.choices.values if value != NO_ANSWER)
@@ -90,8 +93,6 @@ class PublishedRatingResult(RatingResult):
 
 
 class AnsweredRatingResult(PublishedRatingResult):
-    has_answers = True
-
     @property
     def average(self) -> float:
         return sum(grade * count for count, grade in zip(self.counts, self.choices.grades)) / self.count_sum
@@ -141,7 +142,7 @@ class ContributionResult:
                     return True
                 if question.is_rating_question:
                     assert isinstance(question_result, RatingResult)
-                    return question_result.has_answers
+                    return RatingResult.has_answers(question_result)
         return False
 
 
--- a/evap/results/tools.py
+++ b/evap/results/tools.py
@@ -67,9 +67,7 @@ class PublishedRatingResult(RatingResult):
     is_published = True
 
     def __init__(self, question, answer_counters, additional_text_result=None):
-        assert not any(counter.count != 0 for counter in answer_counters) or self.__class__ != PublishedRatingResult
         super().__init__(question, answer_counters, additional_text_result)
-
         counts = OrderedDict((value, 0) for value in self.choices.values if value != NO_ANSWER)
         for answer_counter in answer_counters:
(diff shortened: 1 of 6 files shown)


evap/results/tools.py
```diff
@@ -41,28 +41,42 @@ def __init__(self, visible_by_contribution: Iterable[UserProfile], visible_by_de
...
+    if answer_counters is None:
+        return RatingResult(question, answer_counters, additional_text_result)
+    if any(counter.count != 0 for counter in answer_counters):
+        return AnsweredRatingResult(question, answer_counters, additional_text_result)
+    return PublishedRatingResult(question, answer_counters, additional_text_result)
+
+
 class RatingResult:
+    is_published = False
+    has_answers = False
```
niklasmohrin, line 54: This might be a funky idea, but maybe the following is possible:

```python
class RatingResult:
    @property
    def is_published(self) -> TypeGuard["PublishedRatingResult"]:
        return isinstance(self, PublishedRatingResult)
```
I would consider this a faithful use of `TypeGuard` and it would allow the results to be used exactly the same in caller code, but with the nice underlying structure of this PR (I like in particular that the methods are not even available on `RatingResult` until you have checked the type!). What do you think?

evap/results/tools.py
```diff
@@ -41,28 +41,42 @@ def __init__(self, visible_by_contribution: Iterable[UserProfile], visible_by_de
...
     @property
     def choices(self):
         return CHOICES[self.question.type]
 
+
+class PublishedRatingResult(RatingResult):
+    is_published = True
+
+    def __init__(self, question, answer_counters, additional_text_result=None):
+        assert not any(counter.count != 0 for counter in answer_counters) or self.__class__ != PublishedRatingResult
```
richardebeling, line 70: To me, it is not clear (just from code in this PR, maybe I'm just missing context) why we have `PublishedRatingResult` and `AnsweredRatingResult` and why this one here asserts that all counts are != 0. Surely we can publish results to a rating question where one of the answer options was never used? Is the assumption here simply that counters will only be lazily created on their first increment? If yes, is that in any way important to the code? Seems to me that this should just be an implementation detail of the voting code.

---

I think the `or self.__class__ != PublishedRatingResult` breaks Liskov substitution. For two classes A and B with B inheriting from A, you would assume that any invariants that A's constructor establishes should also hold for any instance of B (since a "subobject" of A was constructed). Without the invariant enforced here, we possibly have instances of B that are not valid As, or in specific, we have instances of `AnsweredRatingResult` that are not valid `PublishedRatingResult`s. This can become potentially dangerous in user code along the lines of
```python3
if isinstance(my_result, PublishedRatingResult):
    # code assuming that all counters have non-zero count
```

It may well be that this is just a design quirk that we currently already have in the results code.

evap/results/tools.py
Kakadus: sadly, the auther of the PEP decided not support this. Only the first explicitly passed parameter can be guarded :cry: 
https://github.com/microsoft/pyright/discussions/3125#discussioncomment-2266438
https://github.com/python/mypy/issues/16618#issuecomment-1840650889

So, as far as I can see, we either have to use `isinstance` directly or via `is_published(...)`, but not via an instance method, especially not with a property

evap/results/tools.py
Kakadus: Currently, if a `RatingResult` `is_published`, the count properties are numbers, and `None` otherwise.
If it also `has_answers`, the `average` can be computed (`None` otherwise). 
This PR just moves these three types into python classes (`PublishedRatingResult`, `AnsweredRatingResult` respectively). I added this assert statement just to prevent the manual creation of a `PublishedRatingResult`, which actually can be an `AnsweredRatingResult`.
---
Actually with the double negation, it asserts that all counters have zero counts (will fix definitely)

Kakadus: Since I don't currently check the counters in `RatingResult`, we might just remove this assertion

richardebeling: Ok. I'd say we should remove the assertion, then each subclass should be more strict about the invariants it enforces, client code can use that, but Liskov-Substitution still holds at all times.

evap/results/tools.py
richardebeling: I'd be fine with making this a freestanding or static method if we leave a comment to the docs (I'd link https://peps.python.org/pep-0647/#narrowing-of-implicit-self-and-cls-parameters).

Kakadus: the problem is, that the templates need to access those too. With that in mind, there are no use cases for a typeguarded `has_answers` function without #2035. `is_published` is only used inside templates (once).

niklasmohrin: That's too bad, I really thought I was onto something here :D

For templates, we could also register `has_answers` etc. as template filters (we can still define them next to the classes and only register them in `evaluation_filters.py`).

We need a type guard in some way though (I am still convinced that splitting this up on the type level is a good move). I suppose `assert RatingResult.has_answers(rating_result)` isn't too bad, just kind of annoying :/

evap/results/tools.py
```diff
@@ -41,28 +41,43 @@ def __init__(self, visible_by_contribution: Iterable[UserProfile], visible_by_de
...
-            self.counts = tuple(counts.values())
-        else:
-            self.counts = None
-
     @property
     def choices(self):
         return CHOICES[self.question.type]
 
+
+class PublishedRatingResult(RatingResult):
```
richardebeling, line 71: The class names are not very intuitive to me. Can we find a short comment explaining these names that we can add here? I think the last state was "This is what the code currently has, we don't really have a good argument why" :sweat_smile:

richardebeling (approved):
Code looks good to me

evap/results/tools.py
```diff
@@ -41,28 +41,43 @@ def __init__(self, visible_by_contribution: Iterable[UserProfile], visible_by_de
...
-    def __init__(self, question, answer_counters, additional_text_result=None):
+    @classmethod
+    def is_published(cls, rating_result) -> TypeGuard["PublishedRatingResult"]:
+        return isinstance(rating_result, PublishedRatingResult)
+
+    @classmethod
+    def has_answers(cls, rating_result) -> TypeGuard["AnsweredRatingResult"]:
+        return isinstance(rating_result, AnsweredRatingResult)
+
+    def __init__(self, question, _, additional_text_result=None):
```
niklasmohrin, line 61:
```suggestion
    def __init__(self, question, additional_text_result=None):
```

evap/results/tools.py
```diff
@@ -41,28 +41,43 @@ def __init__(self, visible_by_contribution: Iterable[UserProfile], visible_by_de
         self.visible_by_delegation_count = visible_by_delegation_count
 
 
+def create_rating_result(question, answer_counters, additional_text_result=None):
+    if answer_counters is None:
+        return RatingResult(question, answer_counters, additional_text_result)
+    if any(counter.count != 0 for counter in answer_counters):
+        return AnsweredRatingResult(question, answer_counters, additional_text_result)
+    return PublishedRatingResult(question, answer_counters, additional_text_result)
```
niklasmohrin, line 49: I wonder if these invariants would be better kept in the `__init__` of the respective classes. And then, we could `try` to make the most concrete one first, but I am not sure if it would be nicer. But we could add a helper to `PublishedRatingResult` (which the child class would also inherit):
```python
@classmethod
def try_to_create(cls, *args, **kwargs) -> Optional[Self]:
    try:
        return cls(*args, **kwargs)
    except ValueError:
        return None
```
Or maybe it's too weird. @richardebeling ?

richardebeling: I think the current code is fine -- your suggestion also would still need the factory method, right? In that case, I'd try to stick to the simpler code (which I think is the current code)

niklasmohrin (approved):
Thanks!

PR #2094, opened 2023-12-18
fix: notebook logic

fidoriel:
Currently, the javascript that wants to attach the notebook logic to the HTML elements triggers even if the notebook wasn't rendered into the page (when the user is not logged in). As a consequence, on the login page, there is a javascript error in the browser console.

This PR unifies creation of the HTML elements and execution of the corresponding JS code.

### DIFF
--- a/evap/evaluation/templates/notebook.html
+++ b/evap/evaluation/templates/notebook.html
@@ -1,3 +1,4 @@
+{% load static %}
 <div class="notebook">
    <div class="collapse position-fixed collapse-horizontal notebook-animation mt-0 d-print-none" tabindex="-1" id="notebook" aria-expanded="false">
         <div class="card">
@@ -35,7 +36,7 @@
 </div>
 
 <script type="module">
-    import  { initNotebook } from "{% static 'js/notebook.js' %}"
-    initNotebook("#notebook", "#notebook-form", "#evapContent", "evap_notebook_open", "#notebookButton");
+    import { NotebookLogic } from "{% static 'js/notebook.js' %}";
+    import { selectOrError } from "{% static 'js/utils.js' %}";
+    new NotebookLogic(selectOrError("#notebook"), selectOrError("#notebook-form"), selectOrError("#evapContent"), selectOrError("#notebookButton"), "evap_notebook_open").attach();
 </script>
-
--- a/evap/evaluation/templates/notebook.html
+++ b/evap/evaluation/templates/notebook.html
@@ -27,3 +27,15 @@
         </div>
     </div>
 </div>
+
+<div class="collapse show" tabindex="-1" id="notebookButton">
+    <button class="btn btn-outline-secondary button-open-notebook" type="button" data-bs-toggle="collapse" data-bs-target="#notebook">
+        <span class="fas fa-pen-to-square"></span>
+    </button>
+</div>
+
+<script type="module">
+    import  { initNotebook } from "{% static 'js/notebook.js' %}"
+    initNotebook("#notebook", "#notebook-form", "#evapContent", "evap_notebook_open", "#notebookButton");
+</script>
+
--- a/evap/static/ts/src/notebook.ts
+++ b/evap/static/ts/src/notebook.ts
@@ -1,11 +1,6 @@
 import "./translation.js";
 import { unwrap, assert, selectOrError } from "./utils.js";
 
-const NOTEBOOK_LOCALSTORAGE_KEY = "evap_notebook_open";
-const COLLAPSE_TOGGLE_BUTTON_SELECTOR = "#notebookButton";
-const WEBSITE_CONTENT_SELECTOR = "#evapContent";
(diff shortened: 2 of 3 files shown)


evap/static/ts/src/notebook.ts
```diff
@@ -1,11 +1,6 @@
 import "./translation.js";
 import { unwrap, assert, selectOrError } from "./utils.js";
 
-const NOTEBOOK_LOCALSTORAGE_KEY = "evap_notebook_open";
```
richardebeling, line 4: For the localstorage key, we have two options:

1. The notebook assumes it is only ever loaded once into a page. Then, the key can be encapsulated within the .ts file, no other component needs to know about it.
2. The notebook might be loaded multiple times. We must prevent collisions then. One way to do this would be to have the calling code pass a unique id. The current code passes the generic term "evap_notebook_open", the notebook will then concatenate the ID (#notebook) in line 54 below. This is somewhat overcomplicated.

I'm fine with any of the two approaches, but I'd like it to be straightforward. If the calling code passes a localstorage key, just use that.

evap/evaluation/templates/notebook.html
```diff
@@ -27,3 +27,15 @@ <h5 class="card-title m-0">
...
+
+<div class="collapse show" tabindex="-1" id="notebookButton">
+    <button class="btn btn-outline-secondary button-open-notebook" type="button" data-bs-toggle="collapse" data-bs-target="#notebook">
+        <span class="fas fa-pen-to-square"></span>
+    </button>
+</div>
+
+<script type="module">
+    import  { initNotebook } from "{% static 'js/notebook.js' %}"
+    initNotebook("#notebook", "#notebook-form", "#evapContent", "evap_notebook_open", "#notebookButton");
```
richardebeling, line 39: Can we pass `Element`s/`HTMLElement`s instead of selectors? I'd say that would be a more canonical interface.

evap/static/ts/src/notebook.ts
```diff
@@ -86,3 +84,10 @@ export class NotebookLogic {
         this.formLogic.attach();
     };
 }
+
+
+export function initNotebook(notebookSelector: string, noteBookFormSelector: string, evapContentSelector: string, localStorageKey: string, collapseNotebookButtonSelector: string): void {
+    if (document.querySelector(notebookSelector) && document.querySelector(noteBookFormSelector) && document.querySelector(evapContentSelector) && localStorageKey && collapseNotebookButtonSelector) {
```
richardebeling, line 90: No `if` required here after we've moved the JS into notebook.html: If someone calls this functions with invalid selectors/arguments, hard error to the console to make it debuggable.

After that, I think there is no point in the function anymore and we can go back to just using the class / constructor directly.

evap/static/ts/src/notebook.ts
```diff
@@ -50,28 +45,36 @@ export class NotebookLogic {
...
-    constructor(notebookSelector: string) {
-        this.notebookCard = selectOrError(notebookSelector);
-        this.formLogic = new NotebookFormLogic(NOTEBOOK_FORM_SELECTOR);
-        this.evapContent = selectOrError(WEBSITE_CONTENT_SELECTOR);
-        this.localStorageKey = NOTEBOOK_LOCALSTORAGE_KEY + "_" + notebookSelector;
+    private readonly collapseNotebookButton: HTMLElement;
+
+    constructor(
+        notebook: HTMLElement,
+        noteBookForm: HTMLFormElement,
```
richardebeling, line 52: inconsistent capitalization compared to all other instances of "notebook" in names

richardebeling (approved):

evap/static/ts/src/notebook.ts
```diff
@@ -1,17 +1,12 @@
...
 
 class NotebookFormLogic {
     private readonly notebook: HTMLFormElement;
     private readonly updateCooldown = 2000;
 
-    constructor(notebookFormId: string) {
-        this.notebook = selectOrError(notebookFormId);
+    constructor(notebook: HTMLFormElement) {
+        this.notebook = notebook;
     }
```
niklasmohrin, line 10: You can now remove the variable declaration and move it here:
```suggestion
    constructor(private readonly notebook: HTMLFormElement) {}
```

evap/static/ts/src/notebook.ts
```diff
@@ -50,28 +45,36 @@ export class NotebookLogic {
...
+        notebookForm: HTMLFormElement,
+        evapContent: HTMLElement,
+        collapseNotebookButton: HTMLElement,
+        localStorageKey: string,
+    ) {
+        this.notebookCard = notebook;
+        this.formLogic = new NotebookFormLogic(notebookForm);
+        this.evapContent = evapContent;
+        this.localStorageKey = localStorageKey;
+        this.collapseNotebookButton = collapseNotebookButton;
```
niklasmohrin, line 61: Same thing here

niklasmohrin (approved):

PR #2105, opened 2024-01-22
Fix incorrect permissions on `/opt/evap`

fekoch:
This PR checks if the mounted `/evap` has **not** the correct owner and group (`1042:1042`).
If not, it uses `bindfs` instead of a symlink to create the `/opt/evap` link with correct permissions.

### DIFF
--- a/deployment/provision_vagrant_vm.sh
+++ b/deployment/provision_vagrant_vm.sh
@@ -40,9 +40,10 @@ apt-get -q install -y apache2 apache2-dev libapache2-mod-wsgi-py3
 OWNER=$(stat -c %u "$MOUNTPOINT/evap")
 if [ "$OWNER" != 1042 ]; then
   #  if, for any reason, vagrant failed to mount evap under the correct uid, use bindfs to link it to /opt/evap using the wanted user id
+  #  If you are using docker: Make sure /dev/fuse of your host is forwarded into the container
   apt-get -q install -y bindfs
   mkdir -p "$REPO_FOLDER"
-  bindfs --map="$OWNER/1042:@$OWNER/@1042" "$MOUNTPOINT" "$REPO_FOLDER"
+  bindfs --map="$OWNER/1042:@$OWNER/@1042" "$MOUNTPOINT" "$REPO_FOLDER" || exit 1
   echo "sudo bindfs --map=$OWNER/1042:@$OWNER/@1042 '$MOUNTPOINT' '$REPO_FOLDER'" >> /home/$USER/.bashrc
 else
   ln -s "$MOUNTPOINT" "$REPO_FOLDER"
--- a/deployment/provision_vagrant_vm.sh
+++ b/deployment/provision_vagrant_vm.sh
@@ -36,20 +36,25 @@ service redis-server restart
 # install apache
 apt-get -q install -y apache2 apache2-dev libapache2-mod-wsgi-py3
 
-# With docker for mac, root will own the mount point (uid=0). chmod does not touch the host file system in these cases.
-OWNER=$(stat -c %U "$MOUNTPOINT/evap")
-if [ "$OWNER" == "root" ]; then chown -R 1042 "$MOUNTPOINT"; fi
+# link the mounted evap folder from the home directory:
+OWNER=$(stat -c %u "$MOUNTPOINT/evap")
+if [ "$OWNER" != 1042 ]; then
+  #  if, for any reason, vagrant failed to mount evap under the correct uid, use bindfs to link it to /opt/evap using the wanted user id
+  apt-get -q install -y bindfs
+  mkdir -p "$REPO_FOLDER"
+  bindfs --map="$OWNER/1042:@$OWNER/@1042" "$MOUNTPOINT" "$REPO_FOLDER"
+  echo "sudo bindfs --map=$OWNER/1042:@$OWNER/@1042 '$MOUNTPOINT' '$REPO_FOLDER'" >> /home/$USER/.bashrc
+else
+  ln -s "$MOUNTPOINT" "$REPO_FOLDER"
+fi
 
 # make user, create home folder, set uid to the same set in the Vagrantfile (required for becoming the synced folder owner), set default shell to bash
-useradd -m -u $(stat -c "%u" "$MOUNTPOINT/evap") -s /bin/bash evap
+useradd -m -u "$(stat -c "%u" "$MOUNTPOINT/evap")" -s /bin/bash evap
 # allow ssh login
 cp -r /home/vagrant/.ssh /home/$USER/.ssh
 chown -R $USER:$USER /home/$USER/.ssh
 # allow sudo without password
 echo "$USER ALL=(ALL) NOPASSWD:ALL" | tee /etc/sudoers.d/evap
-
-# link the mounted evap folder from the home directory
-ln -s "$MOUNTPOINT" "$REPO_FOLDER"
 
 sudo -H -u $USER $EVAP_PYTHON -m venv $ENV_FOLDER
 # venv will use ensurepip to install a new version of pip. We need to update that version.
@@ -100,7 +105,7 @@ apt-get -q install -y libasound2 libgconf-2-4 libgbm1 libgtk-3-0 libnss3 libx11-
 wget https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh --no-verbose --output-document - | sudo -H -u $USER bash
 
 # setup evap
-cd "$MOUNTPOINT"
+cd "$MOUNTPOINT" || exit 1;
 sudo -H -u $USER git submodule update --init
 
 sudo -H -u $USER mkdir node_modules
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,6 @@ todo
 
 # flake8 configuration
 setup.cfg
+
+# User specific vagrant overrides (needs extra config)
+.Vagrantfile.local


Vagrantfile
```diff
@@ -41,6 +41,13 @@ Vagrant.configure("2") do |config|
     end
   end
 
+  config.vm.provider :libvirt do |l,override|
+    override.vm.box = "generic/ubuntu2204" # un
+    l.memory = 2048
+    override.vm.provision "shell", path: "deployment/provision_vagrant_vm.sh"
+    override.vm.allowed_synced_folder_types = ["sshfs"]
+  end
+
```
richardebeling, line 50: I think we wouldn't want to merge libvirt support (for now) -- can we exclude this part from the PR?

(You can keep using it locally, but I wouldn't want unsupported code in the main repo)

deployment/provision_vagrant_vm.sh
```diff
@@ -36,9 +36,17 @@ service redis-server restart
...
+OWNER=$(stat -c %u "$MOUNTPOINT/evap")
+
+# if, for any reason, vagrant failed to mount evap under the correct uid, use bindfs to link it to /opt/evap using the wanted user id
+if [ $OWNER != 1042 ]; then
+  apt-get -q install -y bindfs && \
+  mkdir -p "$REPO_FOLDER"
+  bindfs --map=$OWNER/1042:@$OWNER/@1042 "$MOUNTPOINT" "$REPO_FOLDER"
+  echo "sudo bindfs --map=$OWNER/1042:@$OWNER/@1042 '$MOUNTPOINT' '$REPO_FOLDER'" >> /home/$USER/.bashrc
+else
+  ln -s "$MOUNTPOINT" "$REPO_FOLDER"
```
richardebeling, line 48: If we move this here, we should also remove the copy down in line 60

(and I'd also prefer to move the comment up to this part)

deployment/provision_vagrant_vm.sh
```diff
@@ -36,9 +36,17 @@ service redis-server restart
...
 
-# With docker for mac, root will own the mount point (uid=0). chmod does not touch the host file system in these cases.
-OWNER=$(stat -c %U "$MOUNTPOINT/evap")
-if [ "$OWNER" == "root" ]; then chown -R 1042 "$MOUNTPOINT"; fi
+OWNER=$(stat -c %u "$MOUNTPOINT/evap")
+
+# if, for any reason, vagrant failed to mount evap under the correct uid, use bindfs to link it to /opt/evap using the wanted user id
+if [ $OWNER != 1042 ]; then
+  apt-get -q install -y bindfs && \
+  mkdir -p "$REPO_FOLDER"
```
richardebeling, line 44: why `&& \`? These two can just be separate commands, right?

richardebeling (changes requested):

deployment/provision_vagrant_vm.sh
fkoch-tgm: like this?

richardebeling: Yes :+1:

deployment/provision_vagrant_vm.sh
```diff
@@ -36,21 +36,26 @@ service redis-server restart
...
-# With docker for mac, root will own the mount point (uid=0). chmod does not touch the host file system in these cases.
-OWNER=$(stat -c %U "$MOUNTPOINT/evap")
-if [ "$OWNER" == "root" ]; then chown -R 1042 "$MOUNTPOINT"; fi
+# link the mounted evap folder from the home directory:
+OWNER=$(stat -c %u "$MOUNTPOINT/evap")
+if [ "$OWNER" != 1042 ]; then
+  #  if, for any reason, vagrant failed to mount evap under the correct uid, use bindfs to link it to /opt/evap using the wanted user id
+  apt-get -q install -y bindfs
+  mkdir -p "$REPO_FOLDER"
+  bindfs --map="$OWNER/1042:@$OWNER/@1042" "$MOUNTPOINT" "$REPO_FOLDER"
```
Kakadus, line 45: This will fail, if `/dev/fuse` is not explicitly forwarded to the docker container:
https://stackoverflow.com/a/49021109

fkoch-tgm: Mounting `/dev/fuse` is the responsibility of the user, as this only seems to apply to users who have a very nonstandard setup.

If you want to add local changes to your vagrantfile: https://github.com/fkoch-tgm/EvaP/wiki/Vagrant-with-libvirt#setup-for-fedora

fkoch-tgm: But you are right, it may be useful to note this somewhere :+1:

Kakadus: adding an appropriate line in the `Vargentfile` may have following implications:
1. requires all docker users to install fuse
2. may require quirks on macos
Especially with 2. I am uncertain as to if it is a problem

fkoch-tgm: I thought it was only an issue for some docker users?

Kakadus: okay, so in order to activate the quirks in the `Vagrantfile`, you need to manually add a `Vagrantfile.local`. Maybe we can echo a suggestion adding this, if the bindfs call fails here?

richardebeling: It could be hard to spot that echo. Maybe we just want to `set -e` before `bindfs` so the script execution aborts there if the mount fails?

(We don't have `set -e` for the whole script so that, if a single niche setup aspect fails, it doesn't prevent the general setup to succeed, so new contributors can get a setup running even if e.g. the browser installation fails. However, if the mounting fails, I don't think we can provide an environment that's good for anything, so exiting would be fair)

niklasmohrin: (or `mount-the-stuff || exit 1` instead of `set -e`)

deployment/provision_vagrant_vm.sh
```diff
@@ -36,21 +36,27 @@ service redis-server restart
...
-OWNER=$(stat -c %U "$MOUNTPOINT/evap")
-if [ "$OWNER" == "root" ]; then chown -R 1042 "$MOUNTPOINT"; fi
+# link the mounted evap folder from the home directory:
+OWNER=$(stat -c %u "$MOUNTPOINT/evap")
+if [ "$OWNER" != 1042 ]; then
+  #  if, for any reason, vagrant failed to mount evap under the correct uid, use bindfs to link it to /opt/evap using the wanted user id
+  #  If you are using docker: Make sure /dev/fuse of your host is forwarded into the container
+  apt-get -q install -y bindfs
+  mkdir -p "$REPO_FOLDER"
+  bindfs --map="$OWNER/1042:@$OWNER/@1042" "$MOUNTPOINT" "$REPO_FOLDER" || exit 1
```
richardebeling, line 46: Coming back to https://github.com/e-valuation/EvaP/pull/2105#discussion_r1462468031:
For me, with a (I'd say) pretty standard Ubuntu 22.04 installation, I'll need to have the additional `--device=/dev/fuse` docker argument in my Vagrantfile.

Since we currently list the docker installation as supported, I'd just add this to our Vagrantfile?
->
```
    d.create_args = ["--cap-add=SYS_ADMIN", "--security-opt=apparmor:unconfined", "--device=/dev/fuse"]
```

richardebeling (changes requested):

deployment/provision_vagrant_vm.sh
```diff
@@ -36,21 +36,27 @@ service redis-server restart
...
+  echo "sudo bindfs --map=$OWNER/1042:@$OWNER/@1042 '$MOUNTPOINT' '$REPO_FOLDER'" >> /home/$USER/.bashrc
```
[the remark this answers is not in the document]
Kakadus: for docker rootless, `stat -c "%u" /evap/evap` returns 0, so the creation of the user fails directly. Its uid should be 1042 instead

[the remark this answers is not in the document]
fkoch-tgm: @richardebeling I am very confused. Who do you mean with "my user"? `vagrant`, `evap` or your host user?

fkoch-tgm:
I just tested the set-up again, the mounting permissions are not a problem any more for 
 - Vagrant 2.3.4
   + Plugin vagrant-sshfs (1.3.7, system)
   + Plugin vagrant-libvirt (0.11.2, system)
 - libvirtd 9.7.0
 - using the custom override `override.vm.allowed_synced_folder_types = ["sshfs"]`

deployment/provision_vagrant_vm.sh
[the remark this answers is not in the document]
richardebeling: I mean the "evap" user inside docker. In line 53 of the provisioning script, the user is created with the user id that is the owner of `$MOUNTPOINT/evap", which is the vagrant-caused mountpoint of the repo folder, and which is owned by uid 1000 at that point (with docker-ce on ubuntu 22.04 after running `vagrant up --provider=docker`).

Note that there is currently (latest commit in this PR) an inconsistency in the provision script: We have a owner-changing bindfs-mount from `$MOUNTPOINT` to `$REPO_FOLDER` to fix ownership, but the rest of the script works in `$MOUNTPOINT`. Before, this didn't matter as we fixed ownership for both, but now they have different ownership, so working inside `$MOUNTPOINT` doesn't seem reasonable to me. The current effect of this is that the provisioning script finishes without error (since everything is setup for the user with id 1000), but attempting to use the environment after `vagrant ssh` fails since evap with uid 1000 doesn't own any of the 1042-owned files in `/evap/evap`. It would work in `/evap/evap`.

Kakadus, line 47: This will be overridden by the line https://github.com/fekoch/EvaP/blob/libvirt-support/deployment/provision_vagrant_vm.sh#L81
> cp /etc/skel/.bashrc /home/$USER/
below. This have to be moved up

deployment/provision_vagrant_vm.sh
```diff
@@ -36,21 +36,27 @@ service redis-server restart
...
+else
+  ln -s "$MOUNTPOINT" "$REPO_FOLDER"
+fi
 
 # make user, create home folder, set uid to the same set in the Vagrantfile (required for becoming the synced folder owner), set default shell to bash
-useradd -m -u $(stat -c "%u" "$MOUNTPOINT/evap") -s /bin/bash evap
+useradd -m -u "$(stat -c "%u" "$MOUNTPOINT/evap")" -s /bin/bash evap
```
Kakadus, line 53: To solve the first part of https://github.com/e-valuation/EvaP/pull/2105/files#r1488291372, this should be 
```suggestion
useradd -m -u 1042 -s /bin/bash evap
```

deployment/provision_vagrant_vm.sh
```diff
@@ -100,7 +106,7 @@ apt-get -q install -y libasound2 libgconf-2-4 libgbm1 libgtk-3-0 libnss3 libx11-
 wget https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh --no-verbose --output-document - | sudo -H -u $USER bash
 
 # setup evap
-cd "$MOUNTPOINT"
+cd "$MOUNTPOINT" || exit 1;
```
Kakadus, line 109: To solve the second part of https://github.com/e-valuation/EvaP/pull/2105/files#r1488291372, this should be
```suggestion
cd "$REPO_FOLDER" || exit 1;
```
Not `$MOUNTPOINT` but `$REPO_FOLDER` is affected by bindfs.

Kakadus (changes requested):
I think it would be really nice to have this. [#2105 (comment)](https://github.com/e-valuation/EvaP/pull/2105#discussion_r1462468031) is definitely needed for docker

Kakadus:
As it turns out, @fekoch does not use this with libvirt, but uses `sshfs` to mount the code, so he cannot really test this. 
In regard to this issue, `sshfs` is no solution for docker because sshfs does not seem to support docker. I will close this and open a new PR with specific docker fixes (we're getting there!!)

PR #2108, opened 2024-01-29
Javascript import maps

### LINKED ISSUE #2031: Use import maps
Javascript / Typescript imports are a bit weird sometimes, for example you have to import from `.js` files when in a `.ts` file. Additionally, we currently use the `support_js_module_import_aggregation` option of Django's `ManifestStaticFilesStorage` to change the imports of javascript files we serve so that they use a filename with a hash in it (this avoids fetching a cached, outdated file).

However, there is also a builtin feature: [Import maps](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap). This way, we could switch out `utils.js` for `utils` and also ensure that we are using the same version of a file everywhere.

A nice bonus: Currently, all files including `foo.js` need to be updated when the hash of `foo.js` changes and therefore also get a new hash and will not be cached anymore. This will not happen anymore with import maps.

niklasmohrin:
Closes #2031

I haven't touched the typescript files yet, because I haven't decided between
- trying to change the imports in the typescript files from `./csrf-utils.js` to `csrf-utils` and teaching typescript how to resolve these modules, and
- just using `./csrf-utils.js` in the import map too.

I think I am slightly preferring the latter. What do you think?

### DIFF
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -23,6 +23,14 @@
                 <link rel="icon" type="image/png" sizes="32x32" href="{% static 'images/favicon_32.png' %}" />
                 <link rel="icon" type="image/png" sizes="16x16" href="{% static 'images/favicon_16.png' %}" />
             {% endif %}
+
+            <script type="importmap">
+                {
+                    "imports": {
+                        "contact_modal": "{% static 'js/contact_modal.js' %}",
+                    }
+                }
+            </script>
         {% endblock %}
     </head>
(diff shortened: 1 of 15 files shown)


richardebeling:
Both fine with me. Using `./csrf-utils.js` also in inline scripts might look a bit more natural, so I'd even slightly prefer that

niklasmohrin:
Closing this for now, as it does not seem too important and I would rather be around if this breaks something :D

PR #2110, opened 2024-02-05
Fix django-stubs-ext missing breaking a production setup

richardebeling:
As django-stubs-ext is required so that `from django_stubs_ext import StrOrPromise` doesn't fail on production, but I also don't want to duplicate the dependency version number across two files.

Opinions? Is there a better way? Maybe by specifying dependencies in pyproject.toml or something?

### DIFF
--- a/evap/evaluation/tools.py
+++ b/evap/evaluation/tools.py
@@ -3,7 +3,7 @@ import typing
 from abc import ABC, abstractmethod
 from collections import defaultdict
 from collections.abc import Iterable, Mapping
-from typing import Any, Protocol, TypeVar
+from typing import TYPE_CHECKING, Any, Protocol, TypeVar
 from urllib.parse import quote
 
 import xlwt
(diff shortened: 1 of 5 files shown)


janno42 (approved):
fine for me

Kakadus (approved):

niklasmohrin:
Couldn't we just do 
```python
from typing import TYPE_CHECKING, Any

if TYPE_CHECKING:
    from django_stubs_ext import StrOrPromise
else:
    StrOrPromise = Any
```

niklasmohrin (approved):

evap/evaluation/tools.py
```diff
@@ -16,7 +16,11 @@
 from django.utils.datastructures import MultiValueDict
 from django.utils.translation import get_language
 from django.views.generic import FormView
-from django_stubs_ext import StrOrPromise
+
+if TYPE_CHECKING:
+    from django_stubs_ext import StrOrPromise
+else:
+    StrOrPromise = Any
```
richardebeling, line 23: @niklasmohrin this means that at test runtime, even if `StrOrPromise` would be available with the correct type, we will have it being equal to `Any`. Run-Time Type Checking (typeguard) thus wouldn't catch violations. Maybe it's better to use
```suggestion
try:
    from django_stubs_ext import StrOrPromise
except ImportError:
    assert not TYPE_CHECKING
    StrOrPromise = Any
```
?

niklasmohrin: Thats better, yes

niklasmohrin (approved):

PR #2112, opened 2024-02-05
Make merge suggestion test independent of localsettings

richardebeling:


evap/staff/tests/test_views.py
```diff
@@ -364,15 +360,19 @@ def test_redirection_user_merge_view(self):
...
         self.assertContains(page, self.other_user.email)
 
+    @override_settings(INSTITUTION_EMAIL_DOMAINS=["institution.example.com", "student.institution.example.com"])
     def test_suggested_merge(self):
+        suggested_merge_candidate = baker.make(UserProfile, email="user@student.institution.example.com")
+        suggested_main_user = baker.make(UserProfile, email="user@institution.example.com")
+
+        # Ensure that the merge candidate has the lower pk / seems older
+        if suggested_merge_candidate.pk > suggested_main_user.pk:
+            suggested_merge_candidate, suggested_main_user = suggested_main_user, suggested_merge_candidate
```
niklasmohrin, line 370: Was this flaky before? I thought that what we currently have also checks for us that the user who was created last is always the main user and we get a test error if that ever changes

richardebeling: Replace with an assertLess -- background was simply that I first assumed that this might be the cause why the test fails for me locally

janno42 (approved):

Kakadus (approved):

PR #2114, opened 2024-02-05
Use `translate` and `blocktranslate` template tags

richardebeling:
They were [introduced in django3.1](https://docs.djangoproject.com/en/5.0/releases/3.1/#templates) and, while it says the old tags  are still supported, the documentation [does not mention them anymore](https://docs.djangoproject.com/en/5.0/topics/i18n/translation/#translate-template-tag).

Replacement was done using the commands in the commit messages. I quickly looked through the results and they seem meaningful to me. It should be easy to rebase/redo, so we can wait (e.g. for #1985)

### DIFF
--- a/evap/contributor/templates/contributor_evaluation_form.html
+++ b/evap/contributor/templates/contributor_evaluation_form.html
@@ -2,7 +2,7 @@
 
 {% load evaluation_filters %}
 
-{% block title %}{% trans 'Your EvaP' %} - {{ block.super }}{% endblock %}
+{% block title %}{% translate 'Your EvaP' %} - {{ block.super }}{% endblock %}
 
 {% block breadcrumb_bar %}
     <div class="breadcrumb-bar">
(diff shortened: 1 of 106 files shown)


janno42 (approved):

Kakadus (approved):

niklasmohrin:
I wonder if we should introduce a Github action workflow that greps for `{% trans ` and `{% blocktrans ` to ensure we only use the verbose names. What do you think?

richardebeling:
> I wonder if we should introduce a Github action workflow that greps for `{% trans ` and `{% blocktrans ` to ensure we only use the verbose names. What do you think?

IMO it's not worth the effort now since we don't have the infrastructure. It doesn't really hurt if the old tags creep in, we can just replace any that we come by.

If we ever setup template linting / semgrep / whatever, I'd be fine with adding a rule for this, but I'd regard other stuff they check for as more important.

Issue #2115, opened 2024-02-12
Remove underlines from results evaluation lines


janno42:
On the results index and results detail pages, the text in evaluation lines is underlined when hovering. The text should not be underlined on hover state.

![image](https://github.com/e-valuation/EvaP/assets/1781719/cd01793a-0679-4c8c-bbbf-365ea6df6c03)

richardebeling:
This broke in #1771 / baaa7dd5c55749a9fa17f5ad1f14e2e2bd4fec53 on main

Bootstrap applies `text-decoration: underline` to `a:hover` in `_reboot.scss:247`. Before that commit, our override in `_grid.scss:55` to have `text-decoration: none` was more important. After the commit, the bootstrap rule has become more important. Seems like both rules have identical specificity (one class, one type), and we unintentionally changed their order due to some additional includes in that PR.

PR #2117, opened 2024-02-12
Use click handlers instead of inserting anchor tags in `.table-seamless-links`

niklasmohrin:
This way we don't need to think about the DOM elements being moved around and we can stop click events in nested elements.

This came up in #1985

### DIFF
--- a/evap/static/scss/components/_tables.scss
+++ b/evap/static/scss/components/_tables.scss
@@ -35,19 +35,8 @@ $table-colors: (
     padding-bottom: 0.4rem;
 }
 
-// Seamless links in tables
-//
-// -1- Expand links to fill the cell
 .table-seamless-links td {
     overflow: hidden;
-
-    a:not(.btn) {
-        color: $black;
-        text-decoration: none;
-        display: block; // -1-
-        margin: -10em;  // -1-
-        padding: 10em;  // -1-
-    }
 }
 
 // Hoverable rows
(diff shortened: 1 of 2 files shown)


evap/static/scss/components/_tables.scss
```diff
@@ -35,19 +35,8 @@ $table-colors: (
...
 
-// Seamless links in tables
-//
-// -1- Expand links to fill the cell
 .table-seamless-links td {
     overflow: hidden;
-
-    a:not(.btn) {
-        color: $black;
-        text-decoration: none;
```
richardebeling, line 46: Tangential to #2115 -- `text-decoration: none` intended to remove the underlining on hovering (and successfully did that some time ago). We now have no equivalent of that, right?

richardebeling: Ah, the results index page doesn't use `.table-seamless-links`. Never mind.

richardebeling (approved):
Code is fine, works nicely in Firefox 121 on Ubuntu

Kakadus (approved):
Also looks good on firefox 122.0.1 :smile:

niklasmohrin:
Working on #1985, I noticed that this breaks the delegation modal on the contributor index: Although the `onclick=` uses `return false` at the end, the click handler on the row is still run. In that PR, I changed the `onclick=` to `addEventListener` with `event.stopPropagation`

PR #2125, opened 2024-02-26
Suppress stdout output in test_dumpdata_called

richardebeling:


Kakadus (approved):
Nice! The precommit command still uses (an untested) plain `print`. Maybe this can be changed in this PR as well?

Kakadus (approved):

PR #2128, opened 2024-02-26
ci: run backup process without dev dependencies

### LINKED ISSUE #2113: Run a CI workflow where dev dependencies are not installed
After #2110, I think that at least one of our CI workflows should run without the dev dependencies installed. @Kakadus said that it is probably possible, we may want to get rid of the dependency of the virtualenv cache action and use `actions/cache` ourselves.

I think the backup process workflow is suited for this, so we don't need to add an extra workflow.

The goal I have in mind is the possibility of the following:
```yaml
- name: Setup python
  uses: ./.github/setup_python
  with:
    install-dev-requirements: false
```
and our `setup_python` does the right thing :^)

Kakadus:
fix #2113

does not (yet?) remove dependency to external venv caching action.

### DIFF
--- a/.github/setup_python/action.yml
+++ b/.github/setup_python/action.yml
@@ -21,7 +21,7 @@ runs:
       uses: actions/cache@v4
       with:
         path: .venv
-        key: venv-${{ runner.os }}-${{ steps.python.outputs.python-version }}-${{ hashFiles( 'requirements*.txt' ) }}
+        key: venv-${{ runner.os }}-${{ steps.python.outputs.python-version }}-${{ hashFiles( inputs.requirements-file ) }}
 
     - name: Install dependencies
       run: |
--- a/.github/setup_python/action.yml
+++ b/.github/setup_python/action.yml
@@ -21,7 +21,8 @@ runs:
       uses: actions/cache@v4
       with:
         path: .venv
-        key: venv-${{ runner.os }}-${{ steps.python.outputs.python-version }}-${{ hashFiles( inputs.requirements-file ) }}
+        # hashFiles( 'requirements*.txt' ) because -dev.txt includes requirements.txt
+        key: venv-${{ runner.os }}-${{ steps.python.outputs.python-version }}-${{ hashFiles( 'requirements*.txt' ) }}-${{ inputs.requirements-file }}
 
     - name: Install dependencies
       run: |
--- a/.github/setup_python/action.yml
+++ b/.github/setup_python/action.yml
@@ -20,9 +20,8 @@ runs:
         install_dev: ${{ inputs.install-dev-requirements }}
       run: |
         if [[ $install_dev == 'true' ]]; then \
-          REQUIREMENTS_TXT='requirements-dev.txt' \
-        else REQUIREMENTS_TXT='requirements.txt' fi
-        
+          set REQUIREMENTS_TXT 'requirements-dev.txt' \
+        else set REQUIREMENTS_TXT 'requirements.txt' fi
         echo "requirements_txt=$REQUIREMENTS_TXT" >> $GITHUB_ENV
       shell: bash
 
--- a/.github/setup_python/action.yml
+++ b/.github/setup_python/action.yml
@@ -21,7 +21,7 @@ runs:
       run: |
         if [[ $install_dev == 'true' ]]; then \
           set REQUIREMENTS_TXT 'requirements-dev.txt' \
-        else set REQUIREMENTS_TXT 'requirements.txt' fi
+        else set REQUIREMENTS_TXT 'requirements.txt'; fi
         echo "requirements_txt=$REQUIREMENTS_TXT" >> $GITHUB_ENV
       shell: bash
 
--- a/.github/setup_python/action.yml
+++ b/.github/setup_python/action.yml
@@ -20,8 +20,9 @@ runs:
         install_dev: ${{ inputs.install-dev-requirements }}
       run: |
         if [[ $install_dev == 'true' ]]; then \
-          set REQUIREMENTS_TXT 'requirements-dev.txt' \
-        else set REQUIREMENTS_TXT 'requirements.txt'; fi
+          REQUIREMENTS_TXT='requirements-dev.txt' \
+        else REQUIREMENTS_TXT='requirements.txt'; fi
+
         echo "requirements_txt=$REQUIREMENTS_TXT" >> $GITHUB_ENV
       shell: bash
 
--- a/.github/setup_python/action.yml
+++ b/.github/setup_python/action.yml
@@ -16,24 +16,40 @@ runs:
         python-version: "3.10"
 
     - name: Restore venv cache
-      uses: syphar/restore-virtualenv@v1
       id: cache-virtualenv
+      uses: actions/cache@v3
       with:
-        requirement_files: ${{ inputs.requirements-file }}
+        path: .venv
+        key: venv-${{ runner.os }}-${{ hashFiles( inputs.requirements-file ) }}-${{ inputs.requirements-file }}
+        restore-keys: venv-${{ runner.os }}-${{ hashFiles( inputs.requirements-file ) }}-${{ inputs.requirements-file }}
 
     - name: Restore pip download cache
-      uses: syphar/restore-pip-download-cache@v1
-      if: steps.cache-virtualenv.outputs.cache-hit != 'true'
+      uses: actions/cache@v3
       with:
-        requirement_files: ${{ inputs.requirements-file }}
+        path: ~/.cache/pip
+        key: pip-${{ runner.os }}-${{ hashFiles( inputs.requirements-file ) }}-${{ inputs.requirements-file }}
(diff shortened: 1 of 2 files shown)


Kakadus:
[74dc1f5](https://github.com/e-valuation/EvaP/pull/2128/commits/74dc1f58267b3f5b68b6744302e6cf2f1809d237) substitutes unmaintained `syphar/restore-virtualenv` action with github's `actions/cache` action and is completely optional

.github/setup_python/action.yml
```diff
@@ -1,27 +1,55 @@
...
-        python-version: '3.10'
+        python-version: "3.10"
 
     - name: Restore venv cache
-      uses: syphar/restore-virtualenv@v1
       id: cache-virtualenv
+      uses: actions/cache@v3
+      with:
+        path: .venv
+        key: venv-${{ runner.os }}-${{ hashFiles( inputs.requirements-file ) }}-${{ inputs.requirements-file }}
```
niklasmohrin, line 23: I think this doesn't suffice: In `requirements-dev.txt` we have `-r requirements.txt`, so the hash of `requirements-dev.txt` doesn't change if we change the contents of `requirements.txt`. Maybe the
```
with:
  requirements-file: requirements.txt
```
interface is too broad afterall and a boolean argument `install-dev-dependencies` is better to clearly state that we are always talking about `requirements.txt` and `requirements-dev.txt` and nothing else. The easiest cache key would then be to always hash both files

.github/setup_python/action.yml
```diff
@@ -1,27 +1,55 @@
...
       if: steps.cache-virtualenv.outputs.cache-hit != 'true'
 
     - name: Install dependencies
-      run: pip install -r requirements-dev.txt
+      env:
+        requirements_txt: ${{ inputs.requirements-file }}
+      run: |
+        python -m venv .venv
+        source .venv/bin/activate
+        pip install -r $requirements_txt
```
niklasmohrin, line 40: Would the following also work?
```suggestion
      run: |
        python -m venv .venv
        source .venv/bin/activate
        pip install -r "${{ inputs.requirements-file }}"
```

.github/setup_python/action.yml
Kakadus: nicely spotted. Will do so...

.github/setup_python/action.yml
Kakadus: seems to work: https://github.com/e-valuation/EvaP/actions/runs/8143458817/job/22255467262#step:3:61

.github/setup_python/action.yml
```diff
@@ -1,27 +1,55 @@
...
+      with:
+        path: .venv
+        key: venv-${{ runner.os }}-${{ hashFiles( inputs.requirements-file ) }}-${{ inputs.requirements-file }}
+        restore-keys: venv-${{ runner.os }}-${{ hashFiles( inputs.requirements-file ) }}-${{ inputs.requirements-file }}
 
     - name: Restore pip download cache
-      uses: syphar/restore-pip-download-cache@v1
+      uses: actions/cache@v3
+      with:
+        path: ~/.cache/pip
```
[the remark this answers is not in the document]
Kakadus: You are right! I think the 3rd party used this to speed up cache misses but this will definitely require more logic now (only "worked" in https://github.com/e-valuation/EvaP/actions/runs/8143458817/job/22255467262#step:3:61 because of error below)

.github/setup_python/action.yml
Kakadus: -dev.txt only: 0c4dd49f29fc31a89010887dc58c0ec0c4b0253d7d12bc8cce03363be6176822
both: 3884987a67ff81e209abc66c7d6d8b7c8e79818064cabd67ec7fa78cb9edb069

.github/setup_python/action.yml
```diff
@@ -1,27 +1,56 @@
...
+      run: |
+        cache_files="requirements.txt"
+        requirements_file="requirements.txt"
+        if [[ ${{ inputs.install-dev-dependencies }} == 'true' ]]; then
+          cache_files="requirements*.txt"
+          requirements_file="requirements-dev.txt"
+        fi
+        echo cache-files=$cache_files >> $GITHUB_ENV
+        echo requirements-file=$requirements_file >> $GITHUB_ENV
+      shell: bash
```
richardebeling, line 23: I vote in favor of just always hashing both files. For a dev-dependency-only update, we will have to rebuild the pip packages unnecessarily, but I think that happens rarely enough.

.github/workflows/tests.yml
```diff
@@ -145,6 +145,8 @@ jobs:
 
       - name: Setup python
         uses: ./.github/setup_python
+        with:
+          install-dev-dependencies: true
```
richardebeling, line 149:
```suggestion
          install-dev-dependencies: false
```

richardebeling (changes requested):

richardebeling (approved):
thanks

niklasmohrin (approved):
Awesome (under the assumption that it works like expected :D)

.github/setup_python/action.yml
```diff
@@ -1,27 +1,44 @@
...
-      uses: syphar/restore-virtualenv@v1
       id: cache-virtualenv
-
-    - name: Restore pip download cache
-      uses: syphar/restore-pip-download-cache@v1
-      if: steps.cache-virtualenv.outputs.cache-hit != 'true'
+      uses: actions/cache@v4
+      with:
+        path: .venv
+        key: venv-${{ runner.os }}-${{ hashFiles( 'requirements*.txt' ) }}
```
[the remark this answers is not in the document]
niklasmohrin: Bold idea: Add the Python binary to `hashFiles`?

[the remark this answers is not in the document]
richardebeling: Fine with me, but I also wouldn't expect much trouble as our python version is minor pinned (we never really needed some specific patch as far as I remember), assuming that we can purge the caches when changing the workflow definition.

.github/setup_python/action.yml
```diff
@@ -1,27 +1,44 @@
...
       run: cp evap/settings_test.py evap/localsettings.py
       shell: bash
+
+    - name: Activate venv
+      run: |
+        source .venv/bin/activate
+        echo $PATH >> $GITHUB_PATH
+        echo PATH=$PATH >> $GITHUB_ENV
+        echo VIRTUAL_ENV=$VIRTUAL_ENV >> $GITHUB_ENV
+        echo PYTHONHOME=$PYTHONHOME >> $GITHUB_ENV
```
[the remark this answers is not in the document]
Kakadus: Just experimented a bit with `A="A B"` in bash:
```bash
echo P=$A
# P=A B
echo P="$A"
# P=A B
```
Maybe your concern is the output of the echo, so something like this:
```bash
echo "P='$A'"
# P='A B'
```

.github/setup_python/action.yml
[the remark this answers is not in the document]
Kakadus: I added the python version to the cache key

.github/setup_python/action.yml
```diff
@@ -1,27 +1,45 @@
...
+        key: venv-${{ runner.os }}-${{ steps.python.outputs.python-version }}-${{ hashFiles( 'requirements*.txt' ) }}
 
     - name: Install dependencies
-      run: pip install -r requirements-dev.txt
+      run: |
+        python -m venv .venv
+        source .venv/bin/activate
+        pip install -r "${{ inputs.requirements-file }}"
       shell: bash
       if: steps.cache-virtualenv.outputs.cache-hit != 'true'
```
niklasmohrin, line 32: Not sure if this is a problem or not, but does this mean that the cache key is the same regardless of which requirements file is passed via inputs.requirements-file? If I am not mistaken, we probably want both the hash of all files, but still the name of the file that is used for pip in the following step. Otherwise, we might use a cache for requirements-Dev when we really only wanted requirements.txt - or is that prevented elsewhere?

niklasmohrin (commented):
Sorry for the stream of questions, they are just coming up like this for me as well 😄

.github/setup_python/action.yml
Kakadus: you are right, I forgot this when reverting to the requirements-file input. Should now be fixed

niklasmohrin: The current version has the problem from my first comment again. The resulting venv directory generated by pip is a function of
- the OS,
- the python version,
- what the contents of _all_ requirements files are (because they include each other), and
- which requirements file is actually used.

So I would think that we would use something like `venv-${{ runner.os }}-${{ steps.python.outputs.python-version }}-${{ hashFiles( 'requirements*.txt' ) }}-${{ inputs.requirements-file }}`

niklasmohrin (changes requested):
Blocking on https://github.com/e-valuation/EvaP/pull/2128#discussion_r1541213626

.github/setup_python/action.yml
Kakadus: Ah well, I did not understand the initial problem... Your line works and fixes the problem :+1:  
As a drawback, the "requirements.txt" cache is cleared when only -dev dependencies change. I think this should be fine however.

niklasmohrin (approved):

PR #2134, opened 2024-03-04
Fix misplaced order icons: Make css selector for order icon padding more selective

### LINKED ISSUE #2033: Ordering icons misplaced in table headers

richardebeling:
fixes #2033

### DIFF
--- a/evap/static/scss/components/_grid.scss
+++ b/evap/static/scss/components/_grid.scss
@@ -180,7 +180,7 @@
     &::before, &::after {
         position: absolute;
         bottom: 0;
-        opacity: 0.3;
+        color: $light-gray;
     }
 
     &::before {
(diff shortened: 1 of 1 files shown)


Kakadus (approved):

Kakadus (changes requested):
Overlapping is fixed but still looks weired for me: 
image
The arrows should not be transparent but just gray, right?

janno42 (commented):
Works for me but we should replace the opacity with gray and black font color to fix @Kakadus's issue

janno42 (approved):

Kakadus (approved):

PR #2136, opened 2024-03-05
reset evaluation after it started

### LINKED ISSUE #1991: Reset evaluation after it started
It is not currently possible to reset a running evaluation. If it turns out that the start of the evaluation period has been set too early, or if the evaluation has to be stopped for other reasons, a reset can currently only be done in the database.

The evaluation operation `revert_to_new` ("Revert to preparation" in the frontend), which resets the selected evaluation to state `new`, should be extended to other states. The operation should be allowed for all states from `prepared` to `reviewed` (all states except `new` and `published`) and should be called `reset_to_new` ("Reset evaluation" in the frontend).
When performing this operation, the user should be able to select a checkbox (default: checked) on the operation confirmation page to delete previously received answers.
If checked 
- all received answers (`TextAnswer` and `TextAnswerCounter`) for this evaluation will be deleted, and
- the `voters` list is cleared.

To prevent unintentional deletion of data, an additional confirmation modal should be displayed if the checkbox for the data to be deleted is set.

fekoch:
Closes #1991 
Replaces the `revert_to_new` evaluation operation with `reset_to_new`. It now allows resetting from any State except `NEW` and `PUBLISHED`. It also allows deleting all previously received answers.

Also, it makes `Evaluation.State` an `IntEnum`.

- [x] Only show confirmation modal when deleting previous answers
- [x] Finish Unit-Tests
- [x] Cleanup leftover TODO's

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -711,9 +711,24 @@ class Evaluation(LoggedModel):
     def manager_approve(self):
         pass
 
-    @transition(field=state, source=[State.PREPARED, State.EDITOR_APPROVED, State.APPROVED], target=State.NEW)
-    def revert_to_new(self):
-        pass
+    @transition(
+        field=state,
+        source=[
+            State.PREPARED,
+            State.EDITOR_APPROVED,
+            State.APPROVED,
+            State.IN_EVALUATION,
+            State.EVALUATED,
+            State.REVIEWED,
+        ],
+        target=State.NEW,
+    )
+    def reset_to_new(self, delete_previous_answers: bool | None = False):
+        """Reset an Evaluation after it started (#1991)"""
+        if delete_previous_answers:
+            for answer_class in Answer.__subclasses__():
+                answer_class._default_manager.filter(contribution__evaluation_id=self.id).delete()
+            self.voters.clear()
 
     @transition(
         field=state,
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -617,6 +617,12 @@ class Evaluation(LoggedModel):
             return self.is_user_responsible_or_contributor_or_delegate(user)
         return self.can_be_seen_by(user)
 
+    def can_reset_to_new(self):
+        """Is it possible to execute .reset_to_new() for this evaluation?"""
+        return any(
+            state_transition.name == "reset_to_new" for state_transition in self.get_available_state_transitions()
+        )  # get_available_<fieldname>_transitions() is available for all fsm-fields on a class
+
     @property
     def can_be_edited_by_manager(self):
         return not self.participations_are_archived and self.state < Evaluation.State.PUBLISHED
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -617,6 +617,7 @@ class Evaluation(LoggedModel):
             return self.is_user_responsible_or_contributor_or_delegate(user)
         return self.can_be_seen_by(user)
 
+    @property
     def can_reset_to_new(self):
         allowed_sources = [
             Evaluation.State.PREPARED,
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -729,7 +729,7 @@ class Evaluation(LoggedModel):
         target=State.NEW,
         conditions=[lambda self: not self.is_single_result],
     )
-    def reset_to_new(self, *, delete_previous_answers: bool | None):
+    def reset_to_new(self, *, delete_previous_answers: bool):
         if delete_previous_answers:
             for answer_class in Answer.__subclasses__():
                 answer_class._default_manager.filter(contribution__evaluation_id=self.id).delete()
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -618,9 +618,15 @@ class Evaluation(LoggedModel):
         return self.can_be_seen_by(user)
 
     def can_reset_to_new(self):
-        return any(
-            state_transition.name == "reset_to_new" for state_transition in self.get_available_state_transitions()
-        )  # get_available_<fieldname>_transitions() is available for all fsm-fields on a class
+        allowed_sources = [
+            Evaluation.State.PREPARED,
+            Evaluation.State.EDITOR_APPROVED,
+            Evaluation.State.APPROVED,
+            Evaluation.State.IN_EVALUATION,
+            Evaluation.State.EVALUATED,
+            Evaluation.State.REVIEWED,
+        ]
+        return self.state in allowed_sources and not self.is_single_result
 
     @property
     def can_be_edited_by_manager(self):
@@ -716,19 +722,7 @@ class Evaluation(LoggedModel):
     def manager_approve(self):
         pass
 
-    @transition(
-        field=state,
-        source=[
-            State.PREPARED,
-            State.EDITOR_APPROVED,
-            State.APPROVED,
-            State.IN_EVALUATION,
-            State.EVALUATED,
-            State.REVIEWED,
-        ],
-        target=State.NEW,
-        conditions=[lambda self: not self.is_single_result],
-    )
+    @transition(field=state, target=State.NEW, conditions=[lambda self: self.can_reset_to_new])
     def reset_to_new(self, *, delete_previous_answers: bool):
         if delete_previous_answers:
             for answer_class in Answer.__subclasses__():
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -728,7 +728,7 @@ class Evaluation(LoggedModel):
         ],
         target=State.NEW,
     )
-    def reset_to_new(self, delete_previous_answers: bool):
+    def reset_to_new(self, *, delete_previous_answers: bool):
         if delete_previous_answers:
             for answer_class in Answer.__subclasses__():
                 answer_class._default_manager.filter(contribution__evaluation_id=self.id).delete()
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -728,7 +728,7 @@ class Evaluation(LoggedModel):
         ],
         target=State.NEW,
     )
-    def reset_to_new(self, *, delete_previous_answers: bool):
+    def reset_to_new(self, *, delete_previous_answers: bool | None):
         if delete_previous_answers:
             for answer_class in Answer.__subclasses__():
                 answer_class._default_manager.filter(contribution__evaluation_id=self.id).delete()
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -618,7 +618,6 @@ class Evaluation(LoggedModel):
         return self.can_be_seen_by(user)
 
     def can_reset_to_new(self):
-        """Is it possible to execute .reset_to_new() for this evaluation?"""
         return any(
             state_transition.name == "reset_to_new" for state_transition in self.get_available_state_transitions()
         )  # get_available_<fieldname>_transitions() is available for all fsm-fields on a class
@@ -729,8 +728,7 @@ class Evaluation(LoggedModel):
         ],
         target=State.NEW,
     )
-    def reset_to_new(self, delete_previous_answers: bool | None = False):
-        """Reset an Evaluation after it started (#1991)"""
+    def reset_to_new(self, delete_previous_answers: bool):
         if delete_previous_answers:
             for answer_class in Answer.__subclasses__():
                 answer_class._default_manager.filter(contribution__evaluation_id=self.id).delete()
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -619,15 +619,7 @@ class Evaluation(LoggedModel):
 
     @property
     def can_reset_to_new(self):
-        allowed_sources = [
-            Evaluation.State.PREPARED,
-            Evaluation.State.EDITOR_APPROVED,
-            Evaluation.State.APPROVED,
-            Evaluation.State.IN_EVALUATION,
-            Evaluation.State.EVALUATED,
-            Evaluation.State.REVIEWED,
-        ]
-        return self.state in allowed_sources and not self.is_single_result
+        return Evaluation.State.PREPARED <= self.state <= Evaluation.State.REVIEWED and not self.is_single_result
 
     @property
     def can_be_edited_by_manager(self):
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -727,6 +727,7 @@ class Evaluation(LoggedModel):
             State.REVIEWED,
         ],
         target=State.NEW,
+        conditions=[lambda self: not self.is_single_result],
     )
     def reset_to_new(self, *, delete_previous_answers: bool | None):
         if delete_previous_answers:
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -719,7 +719,7 @@ class Evaluation(LoggedModel):
     def reset_to_new(self, *, delete_previous_answers: bool):
         if delete_previous_answers:
             for answer_class in Answer.__subclasses__():
-                answer_class._default_manager.filter(contribution__evaluation_id=self.id).delete()
+                answer_class._default_manager.filter(contribution__evaluation=self).delete()
             self.voters.clear()
 
     @transition(
--- a/evap/staff/templates/staff_evaluation_operation.html
+++ b/evap/staff/templates/staff_evaluation_operation.html
@@ -55,9 +55,53 @@
         {% endif %}
         <input type="hidden" name="target_state" value="{{ target_state }}" />
 
+        {% if show_delete_answers_checkbox %}
+            <div class="card mb-3">
+                <div class="card-body">
+                    <div class="row">
+                        <div class="col">
+                            <div class="form-check">
+                                <input class="form-check-input" type="checkbox" name="delete-previous-answers"
+                                       id="delete-previous-answers" checked>
+                                <label class="form-check-label" for="delete-previous-answers">
+                                    {% trans "Delete previous answers" %}
+                                </label>
+                            </div>
+                        </div>
+                    </div>
+                </div>
+            </div>
+        {% endif %}
+
         <div class="card card-submit-area card-submit-area-2 text-center mb-3">
             <div class="card-body">
-                <button type="submit" class="btn btn-primary">{% translate 'Confirm' %}</button>
+                <confirmation-modal type="submit" form="evaluation-operation-form" id="confirmation-modal">
+                    <span slot="title">{% translate 'Reset Evaluation' %}</span>
+                    <span slot="action-text">{% translate 'Reset Evaluation' %}</span>
+                    <span slot="question">
+                        {% blocktranslate trimmed %}
+                            This will delete all previously received answers. Evaluation participants will have to
+                            resubmit their answers.
+                        {% endblocktranslate %}
+                        <span slot="show-button" class="d-none"></span>
+                    </span>
+                </confirmation-modal>
+                <button id="form-confirmation-btn" type="button"
+                        class="btn btn-primary">{% translate 'Confirm' %}</button>
+                <script>
+                    /** @type ConfirmationModal */
+                    const modal = document.getElementById("confirmation-modal");
+                    /** @type HTMLFormElement */
+                    const form = document.getElementById("evaluation-operation-form");
+
+                    document.getElementById("form-confirmation-btn").addEventListener("click", () => {
+                        if (document.getElementById("delete-previous-answers").checked) {
+                            modal.dialog.showModal();
+                        } else {
+                            form.requestSubmit();
+                        }
+                    });
+                </script>
                 <a href="{% url 'staff:semester_view' semester.id %}" class="btn btn-light">{% translate 'Cancel' %}</a>
             </div>
         </div>
--- a/evap/staff/templates/staff_evaluation_operation.html
+++ b/evap/staff/templates/staff_evaluation_operation.html
@@ -6,7 +6,7 @@
     <form id="evaluation-operation-form" method="POST" class="form-horizontal">
         {% csrf_token %}
         {% for evaluation in evaluations %}
-            <input type="hidden" name="evaluation" value="{{ evaluation.id }}"/>
+            <input type="hidden" name="evaluation" value="{{ evaluation.id }}">
         {% endfor %}
 
         <div class="card mb-3">
@@ -14,15 +14,15 @@
                 <p>{{ confirmation_message }}</p>
                 <table class="table table-striped table-vertically-aligned">
                     <colgroup>
-                        <col style="width: 2.75em"/>
-                        <col style="width: 1.25em"/>
-                        <col style="width: 1.25em"/>
-                        <col style="width: 1.25em"/>
-                        <col style="width: 1.25em"/>
-                        <col style=""/>
-                        <col style="width: 12.25em"/>
-                        <col style="width: 12em"/>
-                        <col style="width: 9em"/>
+                        <col style="width: 2.75em">
+                        <col style="width: 1.25em">
+                        <col style="width: 1.25em">
+                        <col style="width: 1.25em">
+                        <col style="width: 1.25em">
+                        <col style="">
+                        <col style="width: 12.25em">
+                        <col style="width: 12em">
+                        <col style="width: 9em">
                     </colgroup>
                     <tbody>
                     {% for evaluation in evaluations %}
@@ -53,7 +53,7 @@
                 {% endif %}
             </div>
         {% endif %}
-        <input type="hidden" name="target_state" value="{{ target_state }}"/>
+        <input type="hidden" name="target_state" value="{{ target_state }}">
 
         {% if show_delete_answers_checkbox %}
             <div class="card mb-3">
--- a/evap/staff/templates/staff_evaluation_operation.html
+++ b/evap/staff/templates/staff_evaluation_operation.html
@@ -61,8 +61,7 @@
                     <div class="row">
                         <div class="col">
                             <div class="form-check">
-                                <input class="form-check-input" type="checkbox" name="delete-previous-answers"
-                                       id="delete-previous-answers" checked>
+                                <input class="form-check-input" type="checkbox" name="delete-previous-answers" id="delete-previous-answers" checked>
                                 <label class="form-check-label" for="delete-previous-answers">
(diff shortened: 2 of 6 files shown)


evap/staff/templates/staff_evaluation_operation.html
```diff
@@ -55,9 +56,39 @@
...
+                    <span slot="action-text">{% translate 'Reset Evaluation' %}</span>
+                    <!-- TODO: Brainstorm better modal-message -->
+                    <span slot="question">
+                        {% blocktranslate trimmed %}
+                            This will delete all previously received answers.
+                        {% endblocktranslate %}
+                    </span>
+                    <!-- TODO: when should one use type="button" vs type="submit" while using the confirmation-model dialogue? -->
+                    <button slot="show-button" type="button" class="btn btn-primary">{% translate 'Confirm' %}</button>
+                </confirmation-modal>
```
fekoch, line 91: @niklasmohrin As I understand it, you worked on the new `<confirmation-modal>`. Do you have an idea for a nice way to skip the modal or only show it when certain conditions are met?

niklasmohrin: The modal display is triggered here: https://github.com/e-valuation/EvaP/blob/4d998159bff118e5ac14529201cc8da9695e1e80/evap/static/ts/src/confirmation-modal.ts#L30

You could make this single line available as a method on `ConfirmationModal` and then use JavaScript to dynamically use the modal or `form.requestSubmit()` directly. Your JavaScript can then use an arbitrary condition for when to use which of the two.

Does this help?

evap/staff/templates/staff_evaluation_operation.html
```diff
@@ -55,9 +56,39 @@
...
+                <confirmation-modal type="submit" form="evaluation-operation-form">
+                    <span slot="title">{% translate 'Reset Evaluation' %}</span>
+                    <span slot="action-text">{% translate 'Reset Evaluation' %}</span>
+                    <!-- TODO: Brainstorm better modal-message -->
+                    <span slot="question">
+                        {% blocktranslate trimmed %}
+                            This will delete all previously received answers.
+                        {% endblocktranslate %}
+                    </span>
+                    <!-- TODO: when should one use type="button" vs type="submit" while using the confirmation-model dialogue? -->
```
niklasmohrin, line 89: The show-button should probably never be a submit button, because form elements defined here (in the users of the confirmation modal) will bind to a form outside of the modal, not within it. The show button should not submit any forms and therefore not be a submit type.

However, it may be practical to overwrite the submit row with buttons that belong to forms outside of the modal, so there it could make sense.

evap/staff/templates/staff_evaluation_operation.html
fekoch: I added the option to set `skipShowingModal`, a generic `()=>boolean`-function, on the modal. With this one can easily specify the logic next to the modal in question:

```html
<script>
    const modal = document.getElementById("confirmation-modal");
                                                                           
    function isDeleteAnswersUnchecked() {
        return !document.getElementById("delete-previous-answers").checked;
    }
                                                                           
    modal.skipShowingModal = isDeleteAnswersUnchecked;
</script>
```

What do you think?

niklasmohrin: After thinking about it again, maybe it would be better if "submit without modal" would not be the default. Instead, the extra button could just be ˋtype=buttonˋ and either trigger the modal, or do ˋform.requestSubmit() otherwise

fekoch:
Two Unit Tests are missing. I do not have any idea on how to practically test for ui popups without selenium tests

evap/evaluation/migrations/0143_alter_evaluation_state.py
```diff
@@ -0,0 +1,34 @@
...
+        ("evaluation", "0142_alter_evaluation_state"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="evaluation",
+            name="state",
+            field=django_fsm.FSMIntegerField(
+                choices=[
+                    (evap.evaluation.models.Evaluation.State["NEW"], "new"),
```
niklasmohrin, line 20: Is this what came out of `makemigrations`? Usually it would emit `10` instead of `Evaluation.State["NEW"]` here, with a good reason: The imported `Evaluation` class always reflects the current state of the codebase, so if we change `Evaluation.State` in the future, this migration would also use the changed version. We want the migration to make changes according to what `Evaluation.State` meant at the of writing though. See for example the migration `0123_evaluation_state_fsm_int` where I changed evaluation states to be numbers instead of strings. There, because the migration should work regardless of the current definition of `Evaluation`, all constants are copied in the migration

niklasmohrin (commented):
I started reviewing your changes and I find it a bit tricky to separate the "revert -> reset" and "make State an IntEnum" changes. Do you think you could split these changes into two PRs? If not, we will have to see how we manage :D 

The extra PR would also be a good place to discuss the motivation for the move to `IntEnum` - it sounds like a good idea, but it's hard to see the benefits when they are mixed in with the other changes here

fekoch:
@fekoch some TODO's:

- [ ] type annotation in test helper function
- [ ] remove empty newline

evap/evaluation/models.py
```diff
@@ -617,6 +617,12 @@ def can_results_page_be_seen_by(self, user):
             return self.is_user_responsible_or_contributor_or_delegate(user)
         return self.can_be_seen_by(user)
 
+    def can_reset_to_new(self):
+        """Is it possible to execute .reset_to_new() for this evaluation?"""
```
richardebeling, line 621: I'd say this is probably fine without the comment, and generally, I'll vote to remove any comments that don't add value.

evap/evaluation/models.py
```diff
@@ -711,9 +717,24 @@ def editor_approve(self):
...
+            State.PREPARED,
+            State.EDITOR_APPROVED,
+            State.APPROVED,
+            State.IN_EVALUATION,
+            State.EVALUATED,
+            State.REVIEWED,
+        ],
+        target=State.NEW,
+    )
+    def reset_to_new(self, delete_previous_answers: bool | None = False):
```
richardebeling, line 732: Why `| None`?

evap/evaluation/models.py
```diff
@@ -711,9 +717,24 @@ def editor_approve(self):
...
+        """Reset an Evaluation after it started (#1991)"""
```
richardebeling, line 733: Maybe also remove? The history/background/issue/PR are available in the introducing commits.

evap/staff/templates/staff_evaluation_operation.html
```diff
@@ -55,9 +55,53 @@
...
-                <button type="submit" class="btn btn-primary">{% translate 'Confirm' %}</button>
+                <confirmation-modal type="submit" form="evaluation-operation-form" id="confirmation-modal">
+                    <span slot="title">{% translate 'Reset Evaluation' %}</span>
+                    <span slot="action-text">{% translate 'Reset Evaluation' %}</span>
+                    <span slot="question">
+                        {% blocktranslate trimmed %}
+                            This will delete all previously received answers. Evaluation participants will have to
+                            resubmit their answers.
+                        {% endblocktranslate %}
+                        <span slot="show-button" class="d-none"></span>
```
richardebeling, line 86: Should probably not be in the `question` slow. Can you add a short comment that the modal always expects this element to exist, but we don't need it on this page?

(I'd also be fine with modifying the confirmation modal to have nicely if an element for this slot does not exist)

evap/evaluation/models.py
fekoch: So that one doesn't have to specify `delete_previous_answers=False` for every `.reset_to_new()`. But as every other current usage is in the tests, it will be clearer when this is removed

evap/evaluation/models.py
```diff
@@ -711,9 +716,23 @@ def editor_approve(self):
...
+            State.PREPARED,
+            State.EDITOR_APPROVED,
+            State.APPROVED,
+            State.IN_EVALUATION,
+            State.EVALUATED,
+            State.REVIEWED,
+        ],
+        target=State.NEW,
+    )
+    def reset_to_new(self, delete_previous_answers: bool):
```
richardebeling, line 731:
```suggestion
    def reset_to_new(self, *, delete_previous_answers: bool):
```

richardebeling (approved):

evap/staff/templates/staff_evaluation_operation.html
```diff
@@ -53,11 +53,59 @@
...
+                        {% endblocktranslate %}
+                    </span>
+                    {% comment %}
+                        The modal needs an element in the slot, otherwise it throws an error.
+                        No button is needed, as the modal is conditionally shown via JS.
+                    {% endcomment %}
+                    <span slot="show-button"></span>
+                </confirmation-modal>
+                <button id="form-confirmation-btn" type="button"
+                        class="btn btn-primary">{% translate 'Confirm' %}</button>
```
janno42:
```suggestion
                <button id="form-confirmation-btn" type="button"
                        class="btn btn-danger">{% translate 'Confirm' %}</button>
```

evap/staff/templates/staff_evaluation_operation.html
```diff
@@ -53,11 +53,59 @@
...
         <div class="card card-submit-area card-submit-area-2 text-center mb-3">
             <div class="card-body">
-                <button type="submit" class="btn btn-primary">{% translate 'Confirm' %}</button>
+                <confirmation-modal type="submit" form="evaluation-operation-form" id="confirmation-modal">
+                    <span slot="title">{% translate 'Reset Evaluation' %}</span>
+                    <span slot="action-text">{% translate 'Reset Evaluation' %}</span>
+                    <span slot="question">
+                        {% blocktranslate trimmed %}
+                            This will delete all previously received answers. Evaluation participants will have to
+                            resubmit their answers.
```
janno42, line 83:
```suggestion
                            This will delete all previously received answers. Voters will have to resubmit the
                            evaluation.
```

evap/staff/views.py
```diff
@@ -276,41 +276,51 @@ def warning_for_inapplicables(amount):
...
+    def applicable_to(evaluation: Evaluation):
+        return evaluation.can_reset_to_new()
 
     @staticmethod
     def warning_for_inapplicables(amount):
         return ngettext(
-            "{} evaluation cannot be reverted, because it already started. It was removed from the selection.",
-            "{} evaluations cannot be reverted, because they already started. They were removed from the selection.",
+            "{} evaluation cannot be reset, because it is already published. It was removed from the selection",
+            "{} evaluations cannot be reset, because they were already published. They were removed from the selection.",
```
janno42, line 300:
```suggestion
            "{} evaluation cannot be reset, because it is already in preparation or published. It was removed from the selection",
            "{} evaluations cannot be reset, because they were already in preparation or published. They were removed from the selection.",
```

janno42 (changes requested):

janno42 (changes requested):
- Forgot to specify before: It should not be possible to reset single results. They should be removed from the selection and the operation should not be `applicable_to` them.
- Submitting other operations (e.g., start evaluation, publish, unpublish) is not possible anymore. The submit button does not send a request in these cases (`document.getElementById("delete-previous-answers")` fails).

evap/staff/views.py
```diff
@@ -276,41 +276,51 @@ def warning_for_inapplicables(amount):
...
+        email_template_contributor=None,
+        email_template_participant=None,
+        delete_previous_answers=None,
     ):
         assert email_template_contributor is None
         assert email_template_participant is None
 
         for evaluation in evaluations:
-            evaluation.revert_to_new()
+            evaluation.reset_to_new(delete_previous_answers=delete_previous_answers)
```
Kakadus, line 317: as noted, it is a BIT sus that we pass `bool | None` to `reset_to_new`, but I think it's okayish here

Kakadus (approved):

evap/staff/templates/staff_evaluation_operation.html
fekoch: image

Ist das wirklich der Button den du meinst?

fekoch: Oder meinst du nicht eher den Bestätigungsbutton?

fekoch: image

evap/staff/views.py
richardebeling: I'd prefer the narrower contract on the function, i.e. annotating the function as only taking in a `bool` and then explicitly casting the possibly None value to bool on the caller site. What do you think about that?

fekoch: Yeah, we can just do it narrower. About casting on the caller site:

The only caller site is currently the `ResetToNewOperation`. We could do it like:

```python
evaluation.reset_to_new(delete_previous_answers=bool(delete_previous_answers))
```

But as `delete_previous_answers` is not currently typed inside `ResetToNewOperation.apply()`, nothing complains if we still just do:
```python
evaluation.reset_to_new(delete_previous_answers=delete_previous_answers)
```

fekoch: @richardebeling @Kakadus I completely removed the inheritance and instead use types to validate that the correct method is called.

I am also thinking of adding a test like 
```python
class TestWrongOperationArgs(TestCase):
    def test_wrong_args_throw_typeerror(self):
        evaluation = baker.make(
            Evaluation,
            state=Evaluation.State.PREPARED
        )
        _, operation = evap.staff.views.target_state_and_operation_from_str(str(Evaluation.State.NEW))

        args = {"email_template": EmailTemplate()}

        self.assertRaisesRegex(
            TypeError,
            "got an unexpected keyword argument",
            operation.apply,
            Mock(), [evaluation], **args
        )
```

Do you think this makes sense? Is this not needed? Or should I cover more/all of the operations?

janno42 (changes requested):

evap/staff/templates/staff_evaluation_operation.html
```diff
@@ -53,11 +53,59 @@
...
+                    <span slot="show-button"></span>
+                </confirmation-modal>
+                <button id="form-confirmation-btn" type="button" class="btn btn-primary">{% translate 'Confirm' %}</button>
+                <script>
+                    /** @type ConfirmationModal */
+                    const modal = document.getElementById("confirmation-modal");
+                    /** @type HTMLFormElement */
+                    const form = document.getElementById("evaluation-operation-form");
+
+                    document.getElementById("form-confirmation-btn").addEventListener("click", () => {
```
niklasmohrin, line 98: I think we can clean this up a bit, I imagine the following structure:

```html
<form id="the-form">
    {% if something %}
        <input type="checkbox" name="delete-previous-answers" checked />
    {% endif %}

    <confirmation-modal type="submit" data-allow-answer-deletion ...>
        ...
    </confirmation-modal>

    <button type="submit" ...></button>
</form>

<script>
    const form = document.getElementById("the-form");
    form.addEventListener("submit", event => {
        const data = new FormData(form);
        if (data.get("delete-previous-answers") && !event.submitter?.hasAttribute("data-allow-answer-deletion")) {
            event.preventDefault();
            form.querySelector("confirmation-modal[data-allow-answer-deletion]").dialog.showModal();
        }
    });
</script>
```

This way, we avoid more `id`s and don't need to include different buttons depending on what value `show_delete_answers_checkbox` has. What do you think?

evap/staff/templates/staff_evaluation_operation.html
```diff
@@ -53,11 +53,59 @@
...
+        <input type="hidden" name="target_state" value="{{ target_state }}"/>
+
+        {% if show_delete_answers_checkbox %}
+            <div class="card mb-3">
+                <div class="card-body">
+                    <div class="form-check">
+                        <input class="form-check-input" type="checkbox" name="delete-previous-answers" id="delete-previous-answers" checked>
+                        <label class="form-check-label" for="delete-previous-answers">
+                            {% trans "Delete previous answers" %}
+                        </label>
```
niklasmohrin, line 65: If possible, I favor nesting inputs and labels instead of giving out `id`s. By putting the `input` inside the `label`, they also become associated without having to use `id` and `for`. (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label)

evap/staff/templates/staff_evaluation_operation.html
```diff
@@ -53,11 +53,59 @@
...
+                    </div>
+                </div>
+            </div>
+        {% endif %}
 
         <div class="card card-submit-area card-submit-area-2 text-center mb-3">
             <div class="card-body">
-                <button type="submit" class="btn btn-primary">{% translate 'Confirm' %}</button>
+            {% if show_delete_answers_checkbox %}
+                <confirmation-modal type="submit" form="evaluation-operation-form" id="confirmation-modal">
```
niklasmohrin, line 74: Since this element is already nested in the `evaluation-operation-form`, the `form=` attribute is not needed :)

niklasmohrin (changes requested):
Some notes:

evap/evaluation/models.py
```diff
@@ -617,6 +617,11 @@ def can_results_page_be_seen_by(self, user):
             return self.is_user_responsible_or_contributor_or_delegate(user)
         return self.can_be_seen_by(user)
 
+    def can_reset_to_new(self):
```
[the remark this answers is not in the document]
fekoch: Seems good, I used it the other way around because the logic was simple enough. But it will be probably more efficient the other way around.

evap/staff/templates/staff_evaluation_operation.html
```diff
@@ -6,23 +6,23 @@
     <form id="evaluation-operation-form" method="POST" class="form-horizontal">
         {% csrf_token %}
         {% for evaluation in evaluations %}
-            <input type="hidden" name="evaluation" value="{{ evaluation.id }}" />
+            <input type="hidden" name="evaluation" value="{{ evaluation.id }}"/>
```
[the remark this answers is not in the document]
fekoch: What do you think about removing the slash altogether? For the `<input>` and `<col>` it is not needed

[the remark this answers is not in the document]
richardebeling: I think for general parsing as a human, it's a bit nicer to have the slashes there, indicating that the tag immediately closes, so I'd leave them.

evap/staff/templates/staff_evaluation_operation.html
fekoch: @niklasmohrin I changed it to this. We need to put some kind of marker on the button inside the modal as this element will be used the `.submitter` value.

```html
<form id="evaluation-operation-form">
	<confirmation-modal type="submit" data-allow-answer-deletion>
	...
	    <button slot="submit-group" class="btn ms-2 btn-danger" data-event-type="confirm" data-confirm-answer-deletion>
		{% translate 'Reset Evaluation' %}
	    </button>

	    <span slot="show-button"></span>
	</confirmation-modal>

	<button type="submit" class="btn btn-primary">{% translate 'Confirm' %}</button>
	<script>
	    document.getElementById("evaluation-operation-form").addEventListener("submit", submitEvent => {
		const data = new FormData(submitEvent.target);
		if (data.get("delete-previous-answers") === "on" && !submitEvent.submitter?.hasAttribute("data-confirm-answer-deletion")) {
		    submitEvent.preventDefault()
		    submitEvent.target.querySelector("confirmation-modal[data-allow-answer-deletion]").dialog.showModal();
		}
	    })
	</script>
</form>
```

But I am not sure if I like it, because now the modal and javascript is used on every evaluation operation.

niklasmohrin: Ough I forgot about this: https://github.com/e-valuation/EvaP/blob/a9e417180b6a8e380bbcb9885b301f13e9524bc0/evap/static/ts/src/confirmation-modal.ts#L53-L55
(https://github.com/WICG/webcomponents/issues/814)

I wrote the above thinking that the modal itself would be the submitter :/ 

> But I am not sure if I like it, because now the modal and javascript is used on every evaluation operation.

I am not sure I understand this point - the javascript runs on every submit, but that is not noticable if the condition in the handler is `false`. I prefer handling things in the `submit` events over `click` events because it marks the form as the central actor of the process. The modal is only shown if the delete-answers checkbox is checked, so the modal should only get involved when we really want to ask for additional confirmation - or does it appear more often than needed now? Or am I not getting something else here?

---

I think your solution with the explicit `submit-group` is the only correct way if we want a solution where all _allowed_ submitters are marked with a data attribute. We could also do it the other way around and mark the other submit button as (for example) `data-prompt-confirmation-for-delete-answers` - and we leave the modal just as is. This approach is slightly less nice beacuse the default submission is "unsafe". However, there are only two possible submission sources for the operation form, so this should be fine.

What do you think?

fekoch: re 2: Yeah, I would also go for the current way. It just seems more explicit & readable that way.

re 1: My point was that this way we always send and execute additional javascript to the client, when it is only used for one of the operations. But this was only a minor thought.

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

richardebeling (approved):
thanks :)

Kakadus (approved):

evap/evaluation/models.py
```diff
@@ -617,6 +617,18 @@ def can_results_page_be_seen_by(self, user):
...
+        allowed_sources = [
+            Evaluation.State.PREPARED,
+            Evaluation.State.EDITOR_APPROVED,
+            Evaluation.State.APPROVED,
+            Evaluation.State.IN_EVALUATION,
+            Evaluation.State.EVALUATED,
+            Evaluation.State.REVIEWED,
+        ]
+        return self.state in allowed_sources and not self.is_single_result
```
niklasmohrin, line 630: Any reason to prefer this over the following?
```suggestion
        return Evaluation.State.NEW < self.state < Evaluation.State.PUBLISHED and not self.is_single_result
```

evap/evaluation/models.py
```diff
@@ -711,9 +723,12 @@ def editor_approve(self):
...
         pass
 
-    @transition(field=state, source=[State.PREPARED, State.EDITOR_APPROVED, State.APPROVED], target=State.NEW)
-    def revert_to_new(self):
-        pass
+    @transition(field=state, target=State.NEW, conditions=[lambda self: self.can_reset_to_new])
+    def reset_to_new(self, *, delete_previous_answers: bool):
+        if delete_previous_answers:
+            for answer_class in Answer.__subclasses__():
+                answer_class._default_manager.filter(contribution__evaluation_id=self.id).delete()
```
niklasmohrin, line 730: Does the following also work?
```suggestion
                answer_class._default_manager.filter(contribution__evaluation=self).delete()
```

evap/staff/views.py
```diff
@@ -545,8 +580,17 @@ def evaluation_operation(request, semester_id):
                 plain_content=request.POST["email_plain_participant"],
                 html_content=request.POST["email_html_participant"],
             )
+        if request.POST.get("delete-previous-answers") == "on":
+            delete_previous_answers = True
```
niklasmohrin, line 584: Couldn't we do `delete_previous_answers = request.POST.get("delete-previous-answers") == "on"` from the start? I think this would also fix the `bool | None` "issue", although I don't care so much about that

evap/staff/templates/staff_evaluation_operation.html
```diff
@@ -55,9 +55,47 @@
...
+                    <span slot="question">
+                        {% blocktranslate trimmed %}
+                            This will delete all previously received answers. Voters will have to resubmit the evaluation.
+                        {% endblocktranslate %}
+                    </span>
+                    {% comment %}
+                        The modal needs an element in the slot, otherwise it throws an error.
+                        No button is needed, as the modal is conditionally shown via JS.
+                    {% endcomment %}
+                    <span slot="show-button"></span>
```
niklasmohrin, line 87: Good catch, we should change the modal so that this is not needed, you or me can fix it in a follow up :)

evap/staff/templates/staff_evaluation_operation.html
```diff
@@ -55,9 +55,47 @@
...
+                </confirmation-modal>
                 <button type="submit" class="btn btn-primary">{% translate 'Confirm' %}</button>
+                <script>
+                    document.getElementById("evaluation-operation-form").addEventListener("submit", submitEvent => {
+                        const data = new FormData(submitEvent.target);
+                        if (data.get("delete-previous-answers") === "on" && !submitEvent.submitter?.hasAttribute("data-confirm-answer-deletion")) {
+                            submitEvent.preventDefault()
+                            submitEvent.target.querySelector("confirmation-modal[data-allow-answer-deletion]").dialog.showModal();
```
niklasmohrin, line 95: With the following, we can get rid of the `data-allow-answer-deletion` on the modal:
```suggestion
                            submitEvent.target.querySelector("confirmation-modal:has([data-confirm-answer-deletion])").dialog.showModal();
```
You can decide which version you prefer, I think both are fine

niklasmohrin (approved):
Nice job, I have some final remarks below.

Since all reviews are green now, but you still may want to change something, just write a quick comment once you are okay with merging

evap/staff/views.py
fekoch: We could do it like that, but then we would need to change the `apply` methods of all other operations to expect `False` instead of `None`. This is doable, but as all the other parameters use `None` I do not think it will contribute to readability.

```python
    @staticmethod
    def apply(
        request,
        evaluations,
        email_template=None,
        email_template_contributor=None,
        email_template_participant=None,
        delete_previous_answers=None,
    ):
        assert email_template_contributor is None
        assert email_template_participant is None
        assert delete_previous_answers is None
        
        # ....
```

 So I would argue for keeping it as it is now.

evap/staff/templates/staff_evaluation_operation.html
fekoch: this seems very neat

evap/staff/views.py
richardebeling: I think the other `apply` methods requiring `None` is fine (not type-annotated in the current version, and subject to change anyway?), but we can still use `delete_previous_answers = request.POST.get("delete-previous-answers") == "on"` here.

richardebeling (approved):

evap/staff/views.py
niklasmohrin: either way is fine - if we are divided on this, we can also leave it as is for now

niklasmohrin (approved):
Very cool, thanks!

PR #2149, opened 2024-03-25
CSP preparations: Use less inline scripts and styles

richardebeling:
Commits from #2099 adapted to current main, without CSP nonces, with the intention of possibly reducing merge conflicts for when #2099 can eventually be merged (although it feels like with the new modals and the `trans` -> `translation` change we already have the worst possible merge conflicts :D)


evap/evaluation/templates/navbar.html
```diff
@@ -91,7 +91,7 @@
                         <form class="d-flex" method="post" action="{% url 'evaluation:set_lang' %}">
                             {% csrf_token %}
                             <input name="language" type="hidden" value="{{ language_code }}">
-                            <button data-bs-toggle="tooltip" data-bs-placement="bottom" title="{{ language_name }}" type="submit" onclick="setSpinnerIcon('span-set-language-{{ language_code }}')"
+                            <button data-bs-toggle="tooltip" data-bs-placement="bottom" id="button-set-language-{{ language_code }}" title="{{ language_name }}" type="submit"
```
niklasmohrin, line 94: This comes up a lot, what do you think about `data-set-spinner-icon="span-set-language-{{ language-code }}"` and one `[data-set-spinner-icon]` selector in some JavaScript in `base.html`?

evap/staff/templates/staff_semester_view.html
```diff
@@ -163,7 +163,7 @@ <h3 class="m-0 me-1">{{ semester.name }}</h3>
 
     <div class="card collapsible mb-3">
         <div class="card-header d-flex">
-            <a class="collapse-toggle collapsed" data-bs-toggle="collapse" href="#overviewCard" aria-expanded="false" aria-controls="overviewCard" onclick="saveCollapseState()">{% translate 'Overview' %}</a>
+            <a id="collapseButton" class="collapse-toggle collapsed" data-bs-toggle="collapse" href="#overviewCard" aria-expanded="false" aria-controls="overviewCard">{% translate 'Overview' %}</a>
```
niklasmohrin, line 166: I would prefer something more precise to avoid accidental duplication of IDs

evap/staff/templates/staff_template_form.html
```diff
@@ -21,7 +21,7 @@
                 <div class="col-3">
                     <p>{% translate 'The following variables are available for this email template:' %}</p>
                     {% for variable in available_variables %}
-                        <button type="button" class="btn btn-sm btn-link text-dark pe-1" data-bs-toggle="tooltip" data-bs-placement="top" title="{% translate 'Copy to clipboard' %}" onClick="copyToClipboard('{{ variable }}')">
+                        <button type="button" class="btn btn-sm btn-link text-dark pe-1 copy-to-clipboard-btn" data-variable="{{ variable }}" data-bs-toggle="tooltip" data-bs-placement="top" title="{% translate 'Copy to clipboard' %}">
```
niklasmohrin, line 24:
```suggestion
                        <button type="button" class="btn btn-sm btn-link text-dark pe-1 copy-variable-to-clipboard-btn" data-variable="{{ variable }}" data-bs-toggle="tooltip" data-bs-placement="top" title="{% translate 'Copy to clipboard' %}">
```

niklasmohrin (approved):
Code looks good, haven't tested

janno42 (changes requested):

evap/staff/templates/staff_email_preview_form.html
```diff
@@ -34,3 +33,6 @@
         </div>
     </div>
 </div>
+<script type="text/javascript">
+    document.getElementById("send_email{{ id_suffix }}").addEventListener("click", () => document.getElementById("email_form").classList.toggle("d-none"));
```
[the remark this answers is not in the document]
richardebeling: Urgh, never trust your own code... https://github.com/e-valuation/EvaP/pull/2132. Good catch!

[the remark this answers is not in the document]
niklasmohrin: Go team!

evap/static/scss/_utilities.scss
```diff
@@ -79,3 +79,9 @@ a.no-underline:hover {
 .break-spaces {
     white-space: break-spaces;
 }
+
+@for $i from 0 through 100 {
+    .width-percent-#{$i} {
+        width: $i * 1% !important;
+    }
+}
```
Kakadus, line 87: can we reuse bootstraps `w-...` class naming (like `w-100`)?

evap/student/templates/student_vote.html
```diff
@@ -285,7 +285,7 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
                     }
                 }).catch(error => {
                     // show a warning if the post isn't successful
-                    document.getElementById('submit-error-warning').style.display = 'block';
+                    document.getElementById('submit-error-warning').classList.remove("d-none");
```
Kakadus, line 288: The code before set display to `block`. Is this still the case if just `d-none` is removed?

evap/static/scss/_utilities.scss
richardebeling: @niklasmohrin also suggested this before. I intentionally did not use bootstrap's scheme to
* make sure the names don't collide if bootstrap ever adds w-X classes
* indicate that our unit is "percent", which is rarely the case for other bootstrap helper classes

Not too strong of an opinion on that, if you both prefer it, we can also change it.

evap/student/templates/student_vote.html
richardebeling: Yes, but now only implicitly/by default (since a paragraph element simply is a block element and thus has `display: block` by default). I initially considered this to be a win, not too sure about this now.

janno42 (approved):
🏁

PR #2151, opened 2024-03-25
revert overflow-x-scroll on footer from #2054

janno42:
In #2054 `overflow-x-scroll` was added to the footer ("make footer horizontally scrollable on small screens; The previous behavior was the footer became wider than the screen and the entire page width was increased.")

This leads to a disabled scrollbar (only visible in some browsers on some systems, depending on the scrollbar rendering) in normal resolution, which is worse :)

image

This PR reverts that change, but future commits are welcome. I won't open a new issue, though, because the effect only occurs on very small screens (works on my phone) and doesn't seem worth the effort.

### DIFF
--- a/evap/evaluation/templates/footer.html
+++ b/evap/evaluation/templates/footer.html
@@ -6,7 +6,7 @@
         <div class="vote-bg-orange" style="width: 6%;"></div>
         <div class="vote-bg-red" style="width: 3%;"></div>
     </div>
-    <nav class="navbar navbar-expand overflow-x-scroll">
+    <nav class="navbar navbar-expand">
         <div class="collapse navbar-collapse justify-content-between">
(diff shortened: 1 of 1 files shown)


richardebeling:
> the effect only occurs on very small screens (works on my phone)

@LinqLover is there something we are missing here? I also need to reduce my page size to a point where the rest of the page is unusable anyway in order to get overflow on the footer.

richardebeling (approved):

Kakadus (approved):
Another "bug" I observed with `overflow-x-scroll`: When I leave the zoom at 100%, do not change the height and make the window very thin, I can somehow get the navbar to scroll horizontally for a few pixels. I think it's the "Legal Notice / Data Privacy" button. Maybe this is the symptom on linux devices? It stops if there is no `overflow-x-scroll`

PR #2153, opened 2024-03-27
Fix confirmation modal show button flicker

### LINKED ISSUE #2150: "Delete Semester" (modal submit) button flicker-shows on staff semester page
@niklasmohrin we have an anti-flicker css rule in #1985:

https://github.com/e-valuation/EvaP/blob/e93b9f25227b078bf62d85115edf2979db315341/evap/static/scss/components/_confirmation-modal.scss#L1-L6

However, the selector here excludes the submit group. For me, on the staff semester index page, the red "Delete Semester" button thus shows for a short period during page load. Removing the ` > :not([slot="submit-group"])` part of the selector fixes the flickering for me, but I guess there was a reason for this exclusion?

niklasmohrin:
Closes #2150

It seems that my suspicion from the issue thread is correct and I messed up the CSS rule or renamed the slots at some point 🤷

### DIFF
--- a/evap/static/scss/components/_confirmation-modal.scss
+++ b/evap/static/scss/components/_confirmation-modal.scss
@@ -1,4 +1,4 @@
-confirmation-modal:not(:defined) > :not([slot="submit-group"]) {
+confirmation-modal:not(:defined) > :not([slot="show-button"]) {
   // Without this, the elements that are slotted into the dialog element show up as if they are just normal child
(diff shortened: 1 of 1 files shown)


richardebeling (approved):
👍  (haven't tested, but explanation and code seem reasonable)

Kakadus (approved):
Just tested with firefox 124.0.1, with the semester index page :+1:

PR #2154, opened 2024-03-29
Don't sort course_evaluations in results_evaluation_detail.html again

richardebeling:
It is already sorted in the view:
https://github.com/e-valuation/EvaP/blob/b57ee63d84c88d13664630f758312b10c141caa3/evap/results/views.py#L183

### DIFF
--- a/evap/results/templates/results_evaluation_detail.html
+++ b/evap/results/templates/results_evaluation_detail.html
@@ -134,7 +134,7 @@
                 </div>
                 {% if course_evaluations %}
                     {% include 'results_index_course.html' with evaluations=course_evaluations %}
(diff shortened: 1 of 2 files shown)


niklasmohrin (approved):

niklasmohrin:
Removing the `.sort` line does not fail any tests, should we add one?

richardebeling:
> Removing the `.sort` line does not fail any tests, should we add one?

Done. It's a bit annoying, the test runs for ~400ms. 70ms for results computation/caching, 280ms for rendering the page. I guess its a performance problem with the view, though, the test itself is fine I think.

janno42 (approved):

PR #2155, opened 2024-03-29
Password login improvements

richardebeling:
* disallow password login while OIDC login is active
* aggressively assert that the used login mechanism is intended to be active
* update password login test
* warn if we happen to have users in the database with usable passwords (which we don't expect to happen)

In the long run, it might make sense to completely ditch the password field from the database, and also enforce some kind of mocked OIDC login on development setups


evap/evaluation/models.py
```diff
@@ -1738,6 +1738,13 @@ class Meta:
     def save(self, *args, **kwargs):
         # This is not guaranteed to be called on every insert. For example, the importers use bulk insertion.
 
+        from evap.evaluation.auth import password_login_is_active
```
richardebeling, line 1741: @niklasmohrin this is a bit annoying: When defining the helper methods in auth.py, we can not include `auth` in `models` at the beginning of the file because `auth` already includes `models` at the top of the file. I guess we want to move the method to `tools`, then? We could also implement it in `auth.py` before importing the models, so the file would have generic imports, then generic definitions, then model imports, and then model-specific definitions.

I will have to look through the django documentation, maybe they have some standard structuring for this...

janno42 (approved):

evap/evaluation/tests/test_auth.py
```diff
@@ -128,15 +134,44 @@ def test_oidc_login(self):
...
+            password_form = page.forms["email-login-form"]
+            password_form["email"] = user.email
+            password_form["password"] = "asd"  # nosec
+            response = password_form.submit()
+            self.assertIsInstance(response.context["user"], AnonymousUser)
+            self.assertNotContains(response, "Logout")
+
+        # correct password while password login is disabled
+        with override_settings(ACTIVATE_OPEN_ID_LOGIN=True):
+            self.assertEqual(auth.password_login_is_active(), False)
```
niklasmohrin, line 162:
```suggestion
            self.assertFalse(auth.password_login_is_active())
```

evap/evaluation/tests/test_auth.py
```diff
@@ -128,15 +134,44 @@ def test_oidc_login(self):
...
+
+            password_form["password"] = "evap"  # nosec
+            response = password_form.submit(status=400)
+            self.assertFalse(response.context["user"])
+
+        # correct password while password login is enabled
+        with override_settings(ACTIVATE_OPEN_ID_LOGIN=False):
+            self.assertEqual(auth.password_login_is_active(), True)
```
niklasmohrin, line 170:
```suggestion
            self.assertTrue(auth.password_login_is_active())
```

niklasmohrin (approved):

niklasmohrin:
> In the long run, it might make sense to completely ditch the password field from the database, and also enforce some kind of mocked OIDC login on development setups

that would be cool!

evap/evaluation/models.py
niklasmohrin: Moving to `tools`, while annoying and unsatisfying, seems fair to me. If Django suggests something else, I am also fine with that

Kakadus (approved):

evap/evaluation/models.py
richardebeling: Would you prefer the inline import in the model method, or to have the methods moved to tools? Not quite a fan of both options, but it probably also doesn't really matter. Moving to `tools` would ensure that there's no performance penalty for importing here, but realistically the import should only make a difference the first time we come across the imports.

niklasmohrin: I think I slightly prefer moving the functions to `tools`

Issue #2095, opened 2023-12-18
Discussion Dev Setup Docker


fidoriel:
There are several vagrant issues on my setup.

1. Every Monday I have to reprovision the container. Vagrant throws an OCI Shim error.
2. There is this permission issue.
3. Starting from today, I even cannot rebuild the setup:
```
+ MOUNTPOINT=/evap
+ USER=evap
+ REPO_FOLDER=/opt/evap
+ ENV_FOLDER=/home/evap/venv
+ NODE_MODULES_FOLDER=/home/evap/node_modules
+ EVAP_PYTHON=python3.10
+ export DEBIAN_FRONTEND=noninteractive
+ DEBIAN_FRONTEND=noninteractive
+ apt-get -q update
Hit:1 http://ports.ubuntu.com/ubuntu-ports jammy-updates InRelease
Hit:2 http://ports.ubuntu.com/ubuntu-ports jammy-backports InRelease
Hit:3 http://ports.ubuntu.com/ubuntu-ports jammy-security InRelease
Get:4 http://ports.ubuntu.com/ubuntu-ports jammy InRelease [270 kB]
Err:4 http://ports.ubuntu.com/ubuntu-ports jammy InRelease
  The following signatures were invalid: BADSIG 871920D1991BC93C Ubuntu Archive Automatic Signing Key (2018) <ftpmaster@ubuntu.com>
Reading package lists...
W: GPG error: http://ports.ubuntu.com/ubuntu-ports jammy InRelease: The following signatures were invalid: BADSIG 871920D1991BC93C Ubuntu Archive Automatic Signing Key (2018) <ftpmaster@ubuntu.com>
E: The repository 'http://ports.ubuntu.com/ubuntu-ports jammy InRelease' is not signed.
```

VBox even discontinued the arm version for macOS: https://www.virtualbox.org/ticket/21771
So we can assume, arm will not be supported for Arm Guest/Arm Host for Win and Linux hosts.
Also, using docker with vagrant abuses the container idea.

We also had some discussion about testing setups. If we would fully change to a container setup we would reduce the dev setup complexity. We are currently using containers in CI. All provisioning could go into an 15 line Dockerfile. We could have an idiot-proof, reproducible docker-compose file. For new members, we could provide a prebuild development docker container to speedup setup. If we do not want to be responsible for a container. We do not need this option. Docker benefits from docker-layer-caching, to speedup rebuilds and save storage.

[...]

PR #2162, opened 2024-04-15
docker mount ownership fixes

Kakadus:
This should finally close https://github.com/e-valuation/EvaP/issues/2095 using bindfs to map the ownership to 1042. This basically fixes all open points from #2105.


richardebeling (approved):

deployment/provision_vagrant_vm.sh
```diff
@@ -100,7 +109,7 @@ apt-get -q install -y libasound2 libgconf-2-4 libgbm1 libgtk-3-0 libnss3 libx11-
 wget https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh --no-verbose --output-document - | sudo -H -u $USER bash
 
 # setup evap
-cd "$MOUNTPOINT"
+cd "$REPO_FOLDER" || exit 1
```
niklasmohrin, line 112: Why the change to `REPO_FOLDER` here? Is it important?

deployment/provision_vagrant_vm.sh
```diff
@@ -36,12 +36,22 @@ service redis-server restart
...
+if [ "$OWNER" != 1042 ]; then
+  #  if, for any reason, vagrant failed to mount evap under the correct uid, use bindfs to link it to /opt/evap using the wanted user id
+  #  If you are using docker: Make sure /dev/fuse of your host is forwarded into the container
+  apt-get -q install -y bindfs
+  mkdir -p "$REPO_FOLDER"
+  bindfs --map="$OWNER/1042:@$OWNER/@1042" "$MOUNTPOINT" "$REPO_FOLDER" || exit 1
+  echo "sudo bindfs --map=$OWNER/1042:@$OWNER/@1042 '$MOUNTPOINT' '$REPO_FOLDER'" >> /home/$USER/.bashrc
+else
+  ln -s "$MOUNTPOINT" "$REPO_FOLDER"
+fi
```
niklasmohrin, line 51: Should we even keep this? It might make debugging more complicated if we have two possible setups. Or does bindfs break if we map 1042:1042 or so?

deployment/provision_vagrant_vm.sh
```diff
@@ -36,12 +36,22 @@ service redis-server restart
...
 
-# With docker for mac, root will own the mount point (uid=0). chmod does not touch the host file system in these cases.
-OWNER=$(stat -c %U "$MOUNTPOINT/evap")
-if [ "$OWNER" == "root" ]; then chown -R 1042 "$MOUNTPOINT"; fi
+cp /etc/skel/.bashrc /home/$USER/
+# link the mounted evap folder from the home directory:
+OWNER=$(stat -c %u "$MOUNTPOINT/evap")
+if [ "$OWNER" != 1042 ]; then
+  #  if, for any reason, vagrant failed to mount evap under the correct uid, use bindfs to link it to /opt/evap using the wanted user id
+  #  If you are using docker: Make sure /dev/fuse of your host is forwarded into the container
```
niklasmohrin, line 44: Shouldn't this always be the case according to our Vagrantfile?

In VirtualBox, we should have `/dev/fuse` from the guest OS, or would we also have to forward it?

niklasmohrin (commented):
If this works for everyone needing it, I am fine with merging

What does migrating to this new provision script look like? Does `vagrant provision` on an existing VM just work, or is `vagrant destroy` needed? (Just asking preemptively because this might come up after merging)

deployment/provision_vagrant_vm.sh
Kakadus: `MOUNTPOINT` may have incorrect ownership, $REPO_FOLDER is guaranteed to be correct because of bindfs

deployment/provision_vagrant_vm.sh
Kakadus:
```suggestion
```

deployment/provision_vagrant_vm.sh
Kakadus: bindfs certainly supports this. My idea was to keep the 'old' (simpler?) behavior for 'standard' environments

deployment/provision_vagrant_vm.sh
Kakadus: Specially, all calls with `sudo -H -u $USER ...` will fail then

deployment/provision_vagrant_vm.sh
niklasmohrin: I would prefer if we would be confident that the bindfs solution works in all cases and only have the one setup, but if you prefer having working setups definitely keep working, that is also fair :+1:

niklasmohrin (changes requested):
Did a fresh vagrant `up` and `provision` with Docker. I took the bindfs route, because `OWNER` was `1000` (my uid on the host). Some errors occurred, I listed them in the review comments.

After fixing them, I did another fresh run and it seems to work as expected. I get this warning on every `vagrant ssh` though:
```plain
fuse: mountpoint is not empty
fuse: if you are sure this is safe, use the 'nonempty' mount option
```
I suppose this is because we re-mount in bashrc right? If so, I am fine with leaving it

---

(not this PR) One thing that was annoying was that `npm ci` took a couple of minutes because the puppeteer chrome download failed. For future reference:

```plain
npm ERR! code 1
npm ERR! path /opt/evap/node_modules/puppeteer
npm ERR! command failed
npm ERR! command sh -c node install.js
npm ERR! ERROR: Failed to set up Chrome r115.0.5790.102! Set "PUPPETEER_SKIP_DOWNLOAD" env variable to skip download.
npm ERR! Error: read ECONNRESET
npm ERR!     at TLSWrap.onStreamRead (node:internal/stream_base_commons:218:20) {
npm ERR!   errno: -104,
npm ERR!   code: 'ECONNRESET',
npm ERR!   syscall: 'read'
npm ERR! }
```
This also left me without tsc and scss, so the first `./manage.py run` prompted me to install those. Aborting and running `npm ci` again gives the same error. With `PUPPETEER_SKIP_DOWNLOAD=1 npm ci`, I could work again

deployment/provision_vagrant_vm.sh
richardebeling: Does the ubuntu virtualbox base image already contain bindfs (-> is apt install a noop)? If yes, I'd agree that always using bindfs simplifies stuff without adding bloat/requirements and I'd vote in favor of just always using it.

(If it has to be installed first, I'm neutral.)

Kakadus:
> I suppose this is because we re-mount in bashrc right? If so, I am fine with leaving it

Yes. `-o nonempty` silence this error, but it not what we want (bindfs mounts it several times). We need this so that it is mounted after a `vagrant halt`

deployment/provision_vagrant_vm.sh
Kakadus: `bindfs` is not installed by default, so I will switch to bindfs. I'm only a bit concerned with fuse and other "exotic" setups, like wsl on windows (even though wsl 2 should support fuse (https://unix.stackexchange.com/a/639820))

fekoch:
@Kakadus are there any additional requirements for this setup to work (besides what is specified in the readme)? I could not get it to work by simply using `vagrant up --provider docker && vagrant provision`. 

If this is just an issue with my setup, I am also okay with that. It does not break my own dev setup using libvirt/sshfs. But if there are some additional requirements, we should document those.

fekoch (approved):

niklasmohrin (approved):
If it works for everyone involved, we are good :)

richardebeling (approved):
Works for me on Ubuntu 22.04 both with docker and virtualbox. I don't like the error message about the mounting point not being empty (wrong errors are useless, spammy, send you off in the wrong direction when debugging -- imagine someone new where the setup doesn't work trying to find the actual error when they are presented with a wrong error on every SSH), but I won't block for it.

deployment/provision_vagrant_vm.sh
```diff
@@ -36,21 +36,23 @@ service redis-server restart
...
-useradd -m -u $(stat -c "%u" "$MOUNTPOINT/evap") -s /bin/bash evap
+useradd -m -u 1042 -s /bin/bash evap
+cp /etc/skel/.bashrc /home/$USER/
+
+# remount the mounted evap folder from the home directory. Use bindfs to map the owner in all setups.
+OWNER=$(stat -c %u "$MOUNTPOINT/evap")
+apt-get -q install -y bindfs
+mkdir -p "$REPO_FOLDER"
+bindfs --map="$OWNER/1042:@$OWNER/@1042" "$MOUNTPOINT" "$REPO_FOLDER" || exit 1
+echo "sudo bindfs --map=$OWNER/1042:@$OWNER/@1042 '$MOUNTPOINT' '$REPO_FOLDER'" >> /home/$USER/.bashrc
```
Kakadus, line 48: The best I can think of in order to skip mounting if folder is not empty
```suggestion
echo "[[ -z \$(ls -A '$REPO_FOLDER') ]] && sudo bindfs --map='$OWNER/1042:@$OWNER/@1042' '$MOUNTPOINT' '$REPO_FOLDER'  #  remount iff folder empty" >> /home/$USER/.bashrc
```

richardebeling: fine with me (haven't tested)

richardebeling (approved):
works for me

PR #2163, opened 2024-04-22
Disable DATA_UPLOAD_MAX_NUMBER_FIELDS

janno42:
In evaluations forms we can have large numbers of participants. If this number exceeds 1000 (default for `settings.DATA_UPLOAD_MAX_NUMBER_FIELDS`), a `SuspiciousOperation` is raised.

We don't really need this check but want to support large numbers of participants. This PR disables the check by changing the setting to `None`.


richardebeling (approved):
(In case we ever need to block requests: It would probably make more sense to block via https://httpd.apache.org/docs/2.4/mod/core.html#limitrequestbody, but even then, it's probably trivial for attackers to DoS EvaP by just exhausting network bandwidth)

PR #2167, opened 2024-04-22
Reraise exceptions when debugging

Kakadus:
This was raised during debugging https://github.com/e-valuation/EvaP/pull/2136, concerning `update_evaluations`. There, errors due to typeguard were silenced in the tests because of the raw except clause leading to weird test failures.
This adds checks to reraise the exceptions when debugging.

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -2062,14 +2062,15 @@ class EmailTemplate(models.Model):
     def send_to_user(self, user, subject_params, body_params, use_cc, additional_cc_users=(), request=None):
         if not user.email:
             message = gettext_noop("{} has no email address defined. Could not send email.")
-            level = logging.ERROR
+            log_message = message.format(user.full_name_with_additional_info)
             # If this method is triggered by a cronjob changing evaluation states, the request is None.
             # In this case warnings should be sent to the admins via email (configured in the settings for logger.error).
             # If a request exists, the page is displayed in the browser and the message can be shown on the page (messages.warning).
             if request is not None:
-                level = logging.WARNING
+                logger.warning(log_message)
                 messages.warning(request, _(message).format(user.full_name_with_additional_info))
-            logger.log(level, message.format(user.full_name_with_additional_info))
+            else:
+                logger.error(log_message)
             return
 
         cc_users = set(additional_cc_users)
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -27,7 +27,8 @@ from django.urls import reverse
 from django.utils.functional import cached_property
 from django.utils.safestring import SafeData
 from django.utils.timezone import now
-from django.utils.translation import gettext_lazy as _, gettext_noop
+from django.utils.translation import gettext_lazy as _
+from django.utils.translation import gettext_noop
 from django_fsm import FSMIntegerField, transition
 from django_fsm.signals import post_transition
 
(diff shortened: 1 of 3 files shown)


evap/evaluation/models.py
```diff
@@ -2059,17 +2061,15 @@ def send_to_users_in_evaluations(self, evaluations, recipient_groups, use_cc, re
...
             # If this method is triggered by a cronjob changing evaluation states, the request is None.
             # In this case warnings should be sent to the admins via email (configured in the settings for logger.error).
             # If a request exists, the page is displayed in the browser and the message can be shown on the page (messages.warning).
             if request is not None:
-                logger.warning(warning_message)
-                messages.warning(request, _(warning_message))
-            else:
-                logger.error(warning_message)
+                level = logging.WARNING
+                messages.warning(request, _(message))
```
niklasmohrin, line 2071: Does `_(message)` even work with the user data already formatted into `message`? Don't we usually do `_("look at this data: {}").format(the_data)`?

niklasmohrin (approved):
re-raising looks good to me

evap/evaluation/models.py
Kakadus: reverted the 'simplification' and fixed the issue. Only did it in the first place to keep under ruff's `too-many-branches`. This is now only covered with pylint

evap/evaluation/models.py
```diff
@@ -2059,17 +2061,17 @@ def send_to_users_in_evaluations(self, evaluations, recipient_groups, use_cc, re
...
+            message = _("{} has no email address defined. Could not send email.").format(
+                user.full_name_with_additional_info
             )
             # If this method is triggered by a cronjob changing evaluation states, the request is None.
             # In this case warnings should be sent to the admins via email (configured in the settings for logger.error).
             # If a request exists, the page is displayed in the browser and the message can be shown on the page (messages.warning).
             if request is not None:
-                logger.warning(warning_message)
-                messages.warning(request, _(warning_message))
+                logger.warning(message)
```
richardebeling, line 2071: It could be weird if we log translated messages here, because we might only find ~half of the messages if we ever decide to search/grep for the error message. Not a strong issue for me, but for consistency, it probably makes sense to strictly always log in english...

richardebeling (approved):

richardebeling (approved):

niklasmohrin (approved):

evap/evaluation/models.py
niklasmohrin: Do you know why ruff raised a pylint-based lint while pylint itself did not raise it? If the two check the same, we should probably prefer ruff for speed, right?

richardebeling: There was disagreement between pylint and ruff about how many branches there really are in the function. I think ruff counted one more than pylint, so the pylint check doesn't trigger while the ruff check does. Don't know if it's a ruff bug or if ruff thinks it's a pylint bug.

In general, I'd agree with @niklasmohrin to go with ruff for speed.

niklasmohrin: Okay, maybe lets add this as a comment so we can revisit this later

Kakadus: Ah well no. ruff has a bug there. It does not count branches inside `with` statements... [astral-sh/ruff#11313](https://www.github.com/astral-sh/ruff/issues/11313)

PR #2168, opened 2024-04-22
ci: track github actions with dependabot

Kakadus:
Just saw some warnings in the [latest](https://github.com/e-valuation/EvaP/actions/runs/8790652276) ci runs, github [is deprecating](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/) another node version. I never actually saw any practical incompatibilities of actions, so I told dependabot to group all ci PRs for less noise.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,3 +22,13 @@ updates:
         update-types: ["version-update:semver-minor"]
     labels:
       - "[T] Dependencies"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      actions:
+        patterns: ["*"]
+    labels:
+      - "[T] Dependencies"
+      - "[C] Tests and CI"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -22,3 +22,13 @@ updates:
...
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      actions:
+        patterns: ["*"]
+    labels:
+      - "[T] Dependencies"
+      - "[C] Tests and CI"
```
Kakadus, line 34: maybe overkill?
```suggestion
```

richardebeling:
I'm a bit confused, I don't see a deprecation warning (I checked out the output of the "Compile SCSS / Setup nodejs" step).
Our workflow definition file in [.github/setup_nodejs/action.yml](https://github.com/e-valuation/EvaP/blob/main/.github/setup_nodejs/action.yml) doesn't seem to specify a node version. Where exactly can I find the warning and what version specification are we talking about?

In general, I'm fine with just always running on the latest stable node version. After #2077, we're basically only using node for scss and typescript, both are big enough for that to work I guess.

richardebeling (approved):
Ah sorry, didn't see those. The changes are fine with me then

PR #2174, opened 2024-04-29
Make Evaluation.State an IntEnum

fekoch:
Originally a part of PR #2136. Like @niklasmohrin suggested, I pulled the specific changes required to make `Evaluation.State` an enum.

In the process I discovered Django's own `models.IntegerChoices`, which works even better than the python `IntEnum`. It also generates a migration that does not depend on the `Evaluation` class (also an issue @niklasmohrin mentioned)


evap/evaluation/tools.py
```diff
@@ -324,3 +324,11 @@ def export(self, response, *args, **kwargs) -> None:
         """Convenience method to avoid some boilerplate."""
         self.export_impl(*args, **kwargs)
         self.workbook.save(response)
+
+
+def enum_for_django_template(cls):
```
niklasmohrin, line 329: This is not used anywhere anymore :)

evap/staff/views.py
```diff
@@ -494,7 +494,7 @@ def apply(
 
 def target_state_and_operation_from_str(target_state_str: str) -> tuple[int, type[EvaluationOperation]]:
     try:
-        target_state = int(target_state_str)
+        target_state = Evaluation.State(int(target_state_str))
```
niklasmohrin, line 497: We can also mark the return type as `Evaluation.State` instead of `int` then :)

evap/staff/tests/test_views.py
```diff
@@ -2229,16 +2230,33 @@ def test_questionnaire_with_answers_warning(self):
             '<label class="form-check-label badge bg-danger" for="id_contributions-1-questionnaires_0">', page
         )
 
-    @patch.dict(Evaluation.STATE_STR_CONVERSION, {Evaluation.State.PREPARED: "mock-translated-prepared"})
     def test_state_change_log_translated(self):
```
niklasmohrin, line 2233: The new test does not contain any `assertNotIn`, so it is very well possible that "prepared", "new", etc. come up accidentally. Do you think you can come up with a way to avoid this, or is it too annoying now that Django handles the translation?

evap/staff/tests/test_views.py
```diff
@@ -2229,16 +2230,33 @@ def test_questionnaire_with_answers_warning(self):
...
-        page = self.app.get(self.url, user=self.manager)
-        self.assertIn("mock-translated-prepared", page)
+        translation.activate("en")
+        self.manager.language = "en"
+        self.manager.save()
+
+        page_en = self.app.get(self.url, user=self.manager)
+
+        new_en = translation.gettext("new")
+        prepared_en = translation.gettext("prepared")
```
niklasmohrin, line 2244: I think we can lean a bit more towards integration-testing and use "neu" etc. directly

evap/evaluation/tests/test_views.py
```diff
@@ -246,3 +248,51 @@ def test_notebook(self):
...
+        invalid_start_states = [Evaluation.State.NEW, Evaluation.State.PUBLISHED]
+
+        valid_start_states = [
+            Evaluation.State.PREPARED,
+            Evaluation.State.EDITOR_APPROVED,
+            Evaluation.State.APPROVED,
+            # TODO: allow those states:
+            # Evaluation.State.IN_EVALUATION,
+            # Evaluation.State.EVALUATED,
+            # Evaluation.State.REVIEWED
```
niklasmohrin, line 285: CI complains about "TODO" - let's remove these for now and keep this PR unrelated to your other one :)

evap/evaluation/models.py
```diff
@@ -999,14 +988,6 @@ def unlogged_fields(self):
             "_participant_count",
         ]
 
-    @classmethod
-    def transform_log_action(cls, field_action):
```
niklasmohrin, line 1003: Nice, we can get rid of `transform_log_action` everywhere then! If you want, you can try it here, otherwise you or someone else can also do it in another PR later

evap/evaluation/models.py
```diff
@@ -779,24 +779,13 @@ def unpublish(self):
...
-    }
-
     @classmethod
-    def state_to_str(cls, state):
-        return cls.STATE_STR_CONVERSION[state]
+    def state_to_str(cls, state: "Evaluation.State | int") -> StrOrPromise:
+        return Evaluation.State(state).label
 
     @property
     def state_str(self):
```
niklasmohrin, line 787: I suppose we don't need this anymore, the one usage can just do `evaluation.state.label` itself

evap/evaluation/tests/test_views.py
```diff
@@ -246,3 +248,51 @@ def test_notebook(self):
...
+
+        form = semester_overview_page.forms["evaluation_operation_form"]
+
+        form["evaluation"] = [evaluation.pk]
+
+        confirmation_page = form.submit(name="target_state", value=str(Evaluation.State.NEW))
+        confirmation_form = confirmation_page.forms["evaluation-operation-form"]
+        confirmation_form.submit()
+
+        assertion(Evaluation.objects.get(pk=evaluation.pk))
```
niklasmohrin, line 273: Maybe you can refactor this to take `success_expected: bool` instead of `assertion` and do the `assertEqual` etc. here?

niklasmohrin (commented):
Very cool, some comments:

evap/staff/tests/test_views.py
niklasmohrin: (I think the better translation is worth more than the test here, so I am not expecting any changes outside of tests for this comment)

evap/staff/tests/test_views.py
richardebeling: I'm not sure if we've ever done that (make tests rely on the translation stored in messages.po). For some other tests, I think we mock `gettext` or other translation functions to make them return some unique string. The tests then
* look for the existence of the unique string on the page
* assert that the mock was called with the right input

This might help with https://github.com/e-valuation/EvaP/pull/2174#discussion_r1585330486 above, but it might also be overkill. If using the translations stored in messages.mo helps to keep it simple here, that's also fine with me

evap/evaluation/models.py
fekoch: I believe we can't. The field `state` can be just an integer so we need to convert it into an enum before calling .label.

fekoch: We could of course call `Evaluation.State(evaluation.state).label` but I believe this is verbose enough to warrant it's own function.

But if you have a good way to make sure that `state` is always set to an enum member, this would be even better

evap/evaluation/models.py
fekoch: we should probably move this into a different PR. When I quickly tested it just now, it threw multiple issues.

evap/staff/tests/test_views.py
fekoch: nevermind. It seems like this mocking persists past the end of the test.

fekoch: I could not get translation with mocking to work and decided to just match for the full html fragment instead (like you @niklasmohrin thought, the response of course contained the string "new" :) )

evap/evaluation/models.py
niklasmohrin: We could rename the database field to `_state` and add

```python
@property
def state(self):
    return Evaluation.State(self._state)
```

Not sure if we would run into some problems though. @richardebeling What do you think? It sounds like a fair change to me

evap/evaluation/tests/test_views.py
```diff
@@ -246,3 +248,43 @@ def test_notebook(self):
...
+        form = semester_overview_page.forms["evaluation_operation_form"]
+
+        form["evaluation"] = [evaluation.pk]
+
+        confirmation_page = form.submit(name="target_state", value=str(Evaluation.State.NEW))
+        confirmation_form = confirmation_page.forms["evaluation-operation-form"]
+        confirmation_form.submit()
+
+        self.assertTrue(success_expected, "Should have raised exception by now")
+        self.assertEqual(Evaluation.objects.get(pk=evaluation.pk).state, Evaluation.State.NEW)
```
niklasmohrin, line 274: What do you think about the following?

```suggestion
        confirmation_form.submit(expect_errors=not success_expected)

        if success_expected:
            self.assertEqual(Evaluation.objects.get(pk=evaluation.pk).state, Evaluation.State.NEW)
```

niklasmohrin: This also leaves the question whether we want to do anything in the `else` case - I think it would also be fine like this

evap/evaluation/models.py
Kakadus: I also used this pattern and it worked quite well :+1:

richardebeling: As pointed out by @fekoch today: The property has the disadvantage that using the field in django db field specifications (`evaluation__state`) becomes weird (triple underscore? Does that work with the parsing?) when using django's ORM or model bakery.

I'd be fine with all proposed solutions as long as you can still use the field with django DB field lookups.

evap/evaluation/tests/test_views.py
```diff
@@ -246,3 +248,43 @@ def test_notebook(self):
 
         user.refresh_from_db()
         self.assertEqual(user.notes, self.note)
+
+
+class TestResetEvaluation(WebTestStaffMode):
+
+    @classmethod
+    def setUpTestData(cls):
```
richardebeling, line 256: A bit annoying, but pep-484 defaults to "Any" instead of "None" for methods that return nothing
```suggestion
    def setUpTestData(cls) -> None:
```

evap/evaluation/tests/test_views.py
```diff
@@ -246,3 +248,43 @@ def test_notebook(self):
...
+        cls.manager = make_manager()
+        cls.semester = baker.make(Semester, results_are_archived=True)
+
+    def reset_from_x_to_new(self, x, success_expected: bool):
```
richardebeling, line 260:
```suggestion
    def reset_from_x_to_new(self, x, success_expected: bool) -> None:
```

evap/staff/tests/test_views.py
```diff
@@ -2229,16 +2230,18 @@ def test_questionnaire_with_answers_warning(self):
...
+        trans.gettext.side_effect = lambda key: f"TRANSLATED-{key}"
 
         self.evaluation.ready_for_editors()
         self.evaluation.save()
 
-        page = self.app.get(self.url, user=self.manager)
-        self.assertIn("mock-translated-prepared", page)
+        response = self.app.get(self.url, user=self.manager)
+        self.assertContains(
+            response, "<li> TRANSLATED-state: TRANSLATED-new &#8594; TRANSLATED-prepared </li>", html=True
```
richardebeling, line 2243: I've observed `html=True` to be annoyingly slow in the past -- do we need it here? What's the average test runtime with and without the flag?

evap/evaluation/tests/test_views.py
richardebeling: (Agreed. `"Should have raised exception by now"` seems subtly wrong to me, exceptions wouldn't propagate into the test context here. I would have expected this to fail -- maybe the `assertRaises` below is even triggered by some unintended error)

evap/staff/tests/test_views.py
fekoch: comparing:
| Header | Header |
|--------|--------|
| `self.assertContains(response, "<li> TRANSLATED-state: TRANSLATED-new &#8594; TRANSLATED-prepared </li>", html=True)` | ~0.140s |
| `self.assertContains(response, "TRANSLATED-state: TRANSLATED-new &#8594; TRANSLATED-prepared", html=True)` | ~0.154s |
| `self.assertContains(response, "TRANSLATED-state: TRANSLATED-new &#8594; TRANSLATED-prepared")` | ~0.132s |

Yeah, I think without `html=True` it works just as well

evap/evaluation/models.py
fekoch: I would argue for just keeping the `state` field as is. I decided for keeping the `state_str()` property which already existed for the one instance where it is used.

evap/evaluation/tests/test_views.py
fekoch: I changed it to:
```python
        confirmation_page = form.submit(name="target_state", value=str(Evaluation.State.NEW))

        if success_expected:
            confirmation_form = confirmation_page.forms["evaluation-operation-form"]
            confirmation_form.submit()
            self.assertEqual(Evaluation.objects.get(pk=evaluation.pk).state, Evaluation.State.NEW)
        else:  # no confirmation form should exist
            self.assertNotIn("evaluation-operation-form", confirmation_page.forms)

```

evap/evaluation/templatetags/evaluation_filters.py
```diff
@@ -21,8 +21,6 @@
     Evaluation.State.PUBLISHED: _("published"),
 }
 
-STR_TO_STATE = {s: i for i, s in Evaluation.STATE_STR_CONVERSION.items()}
-
```
fekoch, line 25: I could not find any usages of this dict and removing it did not break any tests, is there a reason to keep it around?

richardebeling: I don't think so, probably left unused after some past change, removing should be fine. Git's pickaxe functions (`git log -S STR_TO_STATE`) might show you whether/when the last usage was removed.

evap/evaluation/models.py
niklasmohrin: Ah, the ORM story is a good point. Okay then, let's keep it as is :)

evap/evaluation/tests/test_views.py
```diff
@@ -246,3 +248,45 @@ def test_notebook(self):
...
+        form["evaluation"] = [evaluation.pk]
+
+        confirmation_page = form.submit(name="target_state", value=str(Evaluation.State.NEW))
+
+        if success_expected:
+            confirmation_form = confirmation_page.forms["evaluation-operation-form"]
+            confirmation_form.submit()
+            self.assertEqual(Evaluation.objects.get(pk=evaluation.pk).state, Evaluation.State.NEW)
+        else:  # no confirmation form should exist
+            self.assertNotIn("evaluation-operation-form", confirmation_page.forms)
```
niklasmohrin, line 278: (Nit) I think we can be more succinct here:
```suggestion
        if success_expected:
            confirmation_page.forms["evaluation-operation-form"].submit()
            self.assertEqual(Evaluation.objects.get(pk=evaluation.pk).state, Evaluation.State.NEW)
        else:
            self.assertNotIn("evaluation-operation-form", confirmation_page.forms)
```

evap/staff/tests/test_views.py
```diff
@@ -2229,16 +2230,16 @@ def test_questionnaire_with_answers_warning(self):
             '<label class="form-check-label badge bg-danger" for="id_contributions-1-questionnaires_0">', page
         )
 
-    @patch.dict(Evaluation.STATE_STR_CONVERSION, {Evaluation.State.PREPARED: "mock-translated-prepared"})
-    def test_state_change_log_translated(self):
-        page = self.app.get(self.url, user=self.manager)
-        self.assertNotIn("mock-translated-prepared", page)
```
niklasmohrin, line 2235: Can we keep the `assertNotIn` check before the call to `ready_for_editors`?

evap/evaluation/models.py
```diff
@@ -381,17 +381,17 @@ def all_evaluations_finished(self):
...
-        EDITOR_APPROVED = 30
-        APPROVED = 40
-        IN_EVALUATION = 50
-        EVALUATED = 60
-        REVIEWED = 70
-        PUBLISHED = 80
+    class State(models.IntegerChoices):
+        NEW = 10, _("new")
+        PREPARED = 20, _("prepared")
+        EDITOR_APPROVED = 30, _("editor_approved")
```
niklasmohrin, line 387: (Maybe follow up work)

@richardebeling @Kakadus 

The string `editor_approved` is still not intended for display on the website (we have the `statename` template filter for that) - maybe we can get rid of it at some point and just encode it here? We have other `*Choices` where we use whitespace in the string id (for example `GradeDocument.Type`). I suppose we should come up with a getter that always returns `Evaluation.State` then (not like our current `evaluation.state` which might return `int`)

Is there any reason to not replace `_` with ` ` already now? If not, we could do it now to possibly avoid an additional migration later

niklasmohrin (approved):
This turned out really nicely :)

I haven't checked whether the migration is up to date and leave that to the final reviewer :^)

evap/evaluation/models.py
richardebeling: > Is there any reason to not replace _ with  already now?

None that I know of, putting in the user-facing string is fine with me 👍

evap/evaluation/templatetags/evaluation_filters.py
fekoch: @niklasmohrin you added this in 2021, but even at that commit (https://github.com/e-valuation/EvaP/tree/37508eb2c426d1e61437e24832a12dbad0d95d12) I do not find any usages

niklasmohrin: :thinking: :thinking: :thinking: 
image

Huh, yeah, I guess it's fair to delete it then :D

Kakadus (approved):

evap/evaluation/models.py
Kakadus: searching for `_\("[\w_ ]*_[\w_ ]*"\)` does not yield any other translations, so I'm fine with fixing this here

richardebeling: Note: The same is probably true for `in_evaluation`. Not having another migration doesn't seem so important to me that this PR needs to be blocked by that. I'd say this can be part of another PR.

evap/evaluation/tests/test_views.py
```diff
@@ -246,3 +248,44 @@ def test_notebook(self):
...
+
+        semester_overview_page = self.app.get(
+            reverse("staff:semester_view", args=[self.semester.pk]), user=self.manager, status=200
+        )
+
+        form = semester_overview_page.forms["evaluation_operation_form"]
+
+        form["evaluation"] = [evaluation.pk]
+
+        confirmation_page = form.submit(name="target_state", value=str(Evaluation.State.NEW))
```
richardebeling, line 271: (opinion): If you add `cls.url = reverse("staff:semester_view", args=[self.semester.pk])` in `setUpTestData`, this can be compacted to:

```suggestion
    def reset_from_x_to_new(self, x, success_expected: bool) -> None:
        evaluation = baker.make(Evaluation, state=x, course__semester=self.semester)

        semester_overview_page = self.app.get(self.url, user=self.manager, status=200 )
        form = semester_overview_page.forms["evaluation_operation_form"]
        form["evaluation"] = [evaluation.pk]
        confirmation_page = form.submit(name="target_state", value=str(Evaluation.State.NEW))
```

which I like more -- imo multiple single-statement "blocks" don't really help with readability as they don't provide semantical grouping.

evap/evaluation/tests/test_views.py
```diff
@@ -246,3 +248,44 @@ def test_notebook(self):
...
+
+        if success_expected:
+            confirmation_page.forms["evaluation-operation-form"].submit()
+            self.assertEqual(Evaluation.objects.get(pk=evaluation.pk).state, Evaluation.State.NEW)
+        else:
+            self.assertNotIn("evaluation-operation-form", confirmation_page.forms)
```
richardebeling, line 277: To me, it would be natural to assert that the evaluation state hasn't changed in the else case -- is there a reason you didn't do that?

At the same time, the fact that the form doesn't exist anymore seems more like a accidental implementation detail (I guess we redirect somewhere?) that might change.

evap/evaluation/tests/test_views.py
```diff
@@ -246,3 +248,44 @@ def test_notebook(self):
...
         user.refresh_from_db()
         self.assertEqual(user.notes, self.note)
+
+
+class TestResetEvaluation(WebTestStaffMode):
+
+    @classmethod
+    def setUpTestData(cls) -> None:
+        cls.manager = make_manager()
+        cls.semester = baker.make(Semester, results_are_archived=True)
```
richardebeling, line 258: is `results_are_archived` relevant? If so, maybe add a short comment why? I find it a bit confusing because on a real instance, you wouldn't want to change evaluation states of archived evaluations / in archived semesters

evap/evaluation/tests/test_views.py
```diff
@@ -246,3 +248,44 @@ def test_notebook(self):
...
+        valid_start_states = [
+            Evaluation.State.PREPARED,
+            Evaluation.State.EDITOR_APPROVED,
+            Evaluation.State.APPROVED,
+        ]
+
+        for s in valid_start_states:
+            self.reset_from_x_to_new(s, True)
+        for s in invalid_start_states:
+            self.reset_from_x_to_new(s, False)
```
richardebeling, line 291: I'd add the parameter names here to make reading easier:

```suggestion
        for s in valid_start_states:
            self.reset_from_x_to_new(s, success_expected=True)
        for s in invalid_start_states:
            self.reset_from_x_to_new(s, success_expected=False)
```

evap/evaluation/tests/test_views.py
fekoch: I replaced the check for the nonexisting form with a comparison of the HTTP Status code and assert that must not be 200. Or is it better to not do that check?

evap/evaluation/tests/test_views.py
fekoch: I do not think this is relevant. The original commit was probably in January or something

PR #2178, opened 2024-05-06
Remove reverted scale warning icon

janno42:


richardebeling (approved):
One could argue that the warning, by now, causes more harm than it helps. "are inverted" in the tooltip might imply to some users that they should change/invert their answer, which is not our intention at all. With yes/no questions, the message makes little sense.

We've had reports that people felt more confused by the warning than they would have been if it hadn't been there.

Maybe we should drop the warning completely? The colors very clearly indicate whether the clicked answer is "good" or "bad", that may be enough visual feedback...

niklasmohrin (approved):
I guess this icon is less threatening

I agree with @richardebeling, maybe the text could be longer, like

> Note: As with the other questions, options on the left are positive feedback and options on the right are negative feedback. Hence, for this question, the order of the answer options is the reverse of the usual order.

or even

> The answer scale is inverted for this question: Left = Disagreement = positive

Alternatively, we could display a long disclaimer at the top

> Note: This evaluation contains questions with inverted answer scales. These questions include statements that participants should ideally disagree with (for example: "The teacher was rarely available"), in contrast to most other questions. As with all questions, the positive answer options are on the left and highlighted as green, but the answer texts are turned around. All applicable questions are marked with the following symbol: (the symbol)
>
> _Hide this message in the future_

and clicking the symbol at a question scrolls to the disclaimer.

Kakadus:
> Note: This evaluation contains questions with inverted answer scales. These questions include statements that participants should ideally disagree with (for example: "The teacher was rarely available"), in contrast to most other questions. As with all questions, the positive answer options are on the left and highlighted as green, but the answer texts are turned around. All applicable questions are marked with the following symbol: (the symbol)
> 
> _Hide this message in the future_

Maybe this can become an infotext? I can also imagine that the colors could be already enough and no warning suffices.

niklasmohrin:
I keep wondering whether removing the warning completely would harm accessibility - but maybe users with vision impairments have better ways of dealing with color coded information already and don't need the warning either? 

I have no idea, to be honest

janno42:
I will remove the warning icon. It was originally intended to help avoid confusion, and any additional text would be counterproductive. I currently don't see a way to make the scale more intuitive than it already is.

richardebeling (approved):

niklasmohrin (approved):
I guess most users have seen the warning by now and thus know what do expect anyways

Kakadus (approved):

PR #2179, opened 2024-05-06
Assert that users have no passwords in anonymize command

richardebeling:


niklasmohrin (approved):
Okay with me, but can you elaborate on the motivation for this change? Is it just to avoid leaking a password hash accidentally, or to avoid having a user identifiable by them having a password?

richardebeling:
> Okay with me, but can you elaborate on the motivation for this change? Is it just to avoid leaking a password hash accidentally, or to avoid having a user identifiable by them having a password?

Anonymized data (possibly intended for publishing) should not contain any real password hashes. If we expected users to have passwords stored in our database, we would have to replace them. Since we don't expect users to have any password set, it felt more natural to me to just assert that this expectation holds (and we thus don't leak any critical data).

PR #2182, opened 2024-05-06
track all composite actions with dependabot

Kakadus:
There are still actions which are not covered by dependabot... As it turns out, these need to be mentioned explicitly [dependabot/dependabot-core#4178 (comment)](https://www.github.com/dependabot/dependabot-core/issues/4178#issuecomment-1118492006) as long as there is still no wildcard support [dependabot/dependabot-core#2178](https://github.com/dependabot/dependabot-core/issues/2178). Meh

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -31,3 +31,21 @@ updates:
         patterns: ["*"]
     labels:
       - "[T] Dependencies"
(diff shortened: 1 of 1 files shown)


richardebeling (approved):
I love software that can be configured in a DRY fashion...

niklasmohrin (approved):

PR #2186, opened 2024-05-13
Fix django stubs 5.0.0

Kakadus:
This would require us to add `django-stubs-ext` as a runtime requirement, in order to define an itemgetter for these generic classes via `monkeypatch`

The new annotations do not improve much actually, because `Form.fields` is de-facto untyped ([typeddjango/django-stubs#1514](https://www.github.com/typeddjango/django-stubs/issues/1514))


richardebeling (approved):
thanks :+1:

PR #2187, opened 2024-05-13
 JSON enrollment importer

### LINKED ISSUE #2131: JSON enrollment data importer (v1)
*Edits:*
*- 2024-03-25: Added "scale" which defines whether the Course is graded or not*
*- 2024-03-25: Added "relatedevents" which defines that two events belong to the same Course, changed previous logic with "lvnr"*
*- 2024-05-27: Removed that "All changes should be made by an 'import user' who will be the `user` for `LogEntry`s." Other automatic changes have no user set in the `LogEntry`, so we don't need it here either.*

----

A new enrollment data importer must be implemented. This importer will process JSON files containing account and course/evaluation data.
*This is a first version of the importer, it will be extended in future issues. The JSON files will contain additional information that is omitted here (e.g. "students" in "events" will have additional attributes, hence the bulky format).*

The importer will be called by a management command (which will be called by a cron job) and will process a single file specified by a parameter. A second parameter specifies the `Semester` for which the data is to be imported.
Unlike the current importer, there won't be a GUI, so error messages will have to be sent by email.
In addition, a log should be created and also sent by email after a successful import. This log should contain
- a list of name changes in `UserProfile`s
- a list of new `Course`s/`Evaluation`s created
- a list of updated `Course`s/`Evaluation`s
- a list of removed `Course`s/`Evaluation`s
- a list of all attempted changes for evaluations that are in states >= `APPROVED`

The JSON file is formatted as described below.
- The first part contains information about all the students who are participating in the courses of this semester ("students").
  - Each entry refers to a `UserProfile` which can be identified by the `email`.
  - The `gguid` is a unique ID that will be used to reference the UserProfile in a later step.
  - The data is mapped like this:
    - `name` -> `last_name`
    - `christianname` -> `first_name_given`
- The second part of the file contains information about all the contributors of the courses of this semester ("lecturers").
  - Each entry refers to a `UserProfile` which can be identified by the `email`.
  - The `gguid` is a unique ID that will be used to reference the UserProfile in a later step.
  - The data is mapped like this:
    - `name` -> `last_name`
    - `christianname` -> `first_name_given`
    - `titlefront` -> `title`
- The third part contains information about the courses/evaluations ("events").
  - The `gguid` should be stored as `cms_id` (Campus Management System ID) as a unique identifier for a `Course` to reference later changes. If no `Course` with this `gguid` exists, it needs to be created, otherwise it needs to be updated if there are changes in the metadata. The data is mapped like this:
    - `title` -> `name_de`
    - `title_en` -> `name_en`
    - `type` -> `CourseType` (`name_de`) - If there is no `CourseType` with this name, it should be created automatically.
    - `courses.cprid` -> `Degree` (`name_de`) - This should be mapped via `import_names`. If no `Degree` with a matching `import_name` exists, it should be created automatically.
    - `lecturers` -> `responsibles` - The corresponding user is referenced in `lecturers` via the above-mentioned `gguid`.
  - The `gguid` should be stored as `cms_id` as a unique identifier for an `Evaluation` to reference later changes (the same `gguid` will be set for the Course and its main Evaluation, they are only unique within classes). If no `Evaluation` with this `cms_id` exists, it must be created, otherwise it must be updated when the metadata changes. The data is mapped like this:
    - `relatedevents.gguid`: If the evaluation is not the main evaluation (so if `isexam` is `true`), this links the Evaluation to its Course via `course.cms_id`.
    - `isexam`: If `true`, `name_de` should be "Klausur" and `name_en` should be "Exam", otherwise both are empty (main Evaluation).
    - `appointments.end` -> If `isexam` is true, `vote_start_datetime` is 08:00 on the day after the given date, `vote_end_date` is three days after the given date. Otherwise, `vote_start_datetime` is 08:00 on Monday of the week before the given date, `vote_end_date` is Sunday of the week of the given date (using Monday as the first day of the week).
    - `lecturers` -> `contributors` (referenced by `gguid` in "lecturers") -- **Contributors should only be added by the importer, not removed. In case contributors are removed in a later import, this should be ignored.**
    - `students` -> `participants` (referenced by `gguid` in "students")
    - `courses.scale`: If set, `wait_for_grade_upload_before_publishing` is `True`, otherwise its `False`

Changes to existing evaluations should only be made when the evaluation is in a state < `APPROVED`. Attempted changes in later states should be logged.

----
Format
```
{
    "students": [
        {
            "gguid": 34 character ID String,
            "email": String,
            "name": String,
            "christianname": String
        }
    ],
    "lecturers": [
        {
            "gguid": 34 character ID String,
            "email": String,
            "name": String,
            "christianname": String,
            "titlefront": String
        }
    ],
    "events": [
        {
            "gguid": 34 character ID String,
            "lvnr": double,
            "title": String,
            "title_en": String,
            "type": String,
            "isexam": boolean,
            "courses": [
                {
                    "cprid": String,
                    "scale": String
                },
                {
                    "cprid": String,
                    "scale": String
                }
            ],
            "relatedevents": {
                "gguid": 34 character ID String
            },
            "appointments": [
                {
                    "begin": String (Datetime),
                    "end": String (Datetime)
                }
            ],
            "lecturers": [
                {
                    "gguid": 34 character ID String
                }
            ],
            "students": [
                {
                    "gguid": 34 character ID String
                }
            ]
        }
    ]
}

```

----

Example
```
{
    "students": [
        {
            "gguid": "0x1",
            "email": "1@example.com",
            "name": "1",
            "christianname": "1"
        },
        {
            "gguid": "0x2",
            "email": "2@example.com",
            "name": "2",
            "christianname": "2"
        }
    ],
    "lecturers": [
        {
            "gguid": "0x3",
            "email": "3@example.com",
            "name": "3",
            "christianname": "3",
            "titlefront": "Prof. Dr."
        },
        {
            "gguid": "0x4",
            "email": "4@example.com",
            "name": "4",
            "christianname": "4",
            "titlefront": "Dr."
        }
    ],
    "events": [
        {
            "gguid": "0x5",
            "lvnr": "1",
            "title": "Prozessorientierte Informationssysteme",
            "title_en": "Process-oriented information systems",
            "type": "Vorlesung",
            "isexam": false,
            "courses": [
                {
                    "cprid": "BA-Inf",
                    "scale": "GRADE_PARTICIPATION"
                },
                {
                    "cprid": "MA-Inf",
                    "scale": "GRADE_PARTICIPATION"
                }
            ],
            "relatedevents": {
                "gguid": "0x6"
            },
            "appointments": [
                {
                    "begin": "15.04.2024 10:15",
                    "end": "15.07.2024 11:45"
                }
            ],
            "lecturers": [
                {
                    "gguid": "0x3"
                }
            ],
            "students": [
                {
                    "gguid": "0x1"
                },
                {
                    "gguid": "0x2"
                }
            ]
        },
        {
            "gguid": "0x6",
            "lvnr": "2",
            "title": "Prozessorientierte Informationssysteme",
            "title_en": "Process-oriented information systems",
            "type": "Klausur",
            "isexam": true,
            "courses": [
                {
                    "cprid": "BA-Inf",
                    "scale": "GRADE_TO_A_THIRD"
                },
                {
                    "cprid": "MA-Inf",
                    "scale": "GRADE_TO_A_THIRD"
                }
            ],
            "relatedevents": {
                "gguid": "0x5"
            },
            "appointments": [
                {
                    "begin": "29.07.2024 10:15",
                    "end": "29.07.2024 11:45"
                }
            ],
            "lecturers": [
                {
                    "gguid": "0x3"
                },
                {
                    "gguid": "0x4"
                }
            ],
            "students": [
                {
                    "gguid": "0x1"
                },
                {
                    "gguid": "0x2"
                }
            ]
        }
    ]
}
```

hansegucker:
Close #2131


evap/staff/importers/json.py
```diff
@@ -0,0 +1,204 @@
...
+class ImportDict(TypedDict):
+    students: list[ImportStudent]
+    lecturers: list[ImportLecturer]
+    events: list[ImportEvent]
+
+
+class JSONImporter:
+    DATETIME_FORMAT = "%d.%m.%Y %H:%M"
+
+    def __init__(self, semester: Semester):
```
richardebeling, line 62: Can you annotate all functions/methods that return `None` as `-> None`? Otherwise, PEP-484/Mypy default to `Any`.

evap/staff/importers/json.py
```diff
@@ -0,0 +1,204 @@
...
+        self.course_map[data["gguid"]] = course
+
+        return course
+
+    def _import_evaluation(self, course: Course, data: ImportEvent) -> Evaluation:
+        course_end = datetime.strptime(data["appointments"][0]["end"], self.DATETIME_FORMAT)
+
+        if data["isexam"]:
+            # Set evaluation time frame of three days for exam evaluations:
+            evaluation_start_datetime = course_end.replace(hour=8, minute=0) + timedelta(days=1)
```
richardebeling, line 133: Maybe add `assert course_end.second == course_end.microsecond == 0` in case the input format changes to include seconds?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,204 @@
...
+
+        for lecturer in data["lecturers"]:
+            self._import_contribution(evaluation, lecturer)
+
+        return evaluation
+
+    def _import_contribution(self, evaluation: Evaluation, data: ImportRelated):
+        user_profile = self.user_profile_map[data["gguid"]]
+
+        contribution, __ = Contribution.objects.update_or_create(
```
richardebeling, line 175: `update_or_create` without `defaults` is just a `get_or_create`, right?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,204 @@
...
+            evaluation=evaluation,
+            contributor=user_profile,
+        )
+        return contribution
+
+    def _import_events(self, data: list[ImportEvent]):
+        # Divide in two lists so corresponding courses are imported before their exams
+        normal_events = filter(lambda e: not e["isexam"], data)
+        exam_events = filter(lambda e: e["isexam"], data)
```
richardebeling, line 184: I think our linting will ask you to convert these to list comprehensions (same for `any(filter(lambda grade: grade["scale"], data["courses"]))` above in line 150).

I think `event for event in data if not event["isexam"]` is a bit more natural to read.

evap/staff/importers/json.py
```diff
@@ -0,0 +1,204 @@
...
+
+        for event in normal_events:
+            event: ImportEvent
```
richardebeling, line 187: If we _need_ to do this, I'd say we should annotate `normal_events` and `exam_events` above. However, I'd expect mypy to be able to deduce these types -- are you sure this is requried?

evap/staff/tests/test_json_importer.py
```diff
@@ -0,0 +1,202 @@
...
+    def test_import_students(self):
+        self.assertEqual(UserProfile.objects.all().count(), 0)
+
+        importer = JSONImporter(self.semester)
+        importer._import_students(self.students)
+
+        user_profiles = UserProfile.objects.all()
+        self.assertEqual(user_profiles.count(), 2)
+
+        for i, user_profile in enumerate(user_profiles):
```
richardebeling, line 73: * The loop already checks that there are no more user profiles (because it triggers an out-of-bound access on self.students otherwise), so we don't necessarily need the count assertion. The pattern is repeated below in other tests, I think I'd favor shorter/less test code if it covers the same aspects.

* I don't think you can assume that the user profiles will always be iterated in the same order. so I'd say the iterate-and-compare approach here is flaky. We could explicitly sort by `id` (which would rely on the creation order in the importer), or, possibly nicer, we could try to get some kind of set-comparison (i.e., assert that the set of user profiles is equal to a hardcoded expected set)

evap/staff/tests/test_json_importer.py
```diff
@@ -0,0 +1,202 @@
...
+            self.assertEqual(user_profile.email, self.students[i]["email"])
+            self.assertEqual(user_profile.last_name, self.students[i]["name"])
+            self.assertEqual(user_profile.first_name_given, self.students[i]["christianname"])
+
+    def test_import_existing_students(self):
+
+        user_profile = baker.make(UserProfile, email=self.students[0]["email"])
+        print(user_profile.email)
```
richardebeling, line 81: empty first line and leftover print

evap/staff/tests/test_json_importer.py
```diff
@@ -0,0 +1,202 @@
...
+
+        importer = JSONImporter(self.semester)
+        importer._import_students(self.students)
+
+        assert UserProfile.objects.all().count() == 2
```
richardebeling, line 86: `self.assertEqual`

evap/staff/tests/test_json_importer.py
```diff
@@ -0,0 +1,202 @@
...
+        for i, user_profile in enumerate(user_profiles):
+            self.assertEqual(user_profile.email, self.lecturers[i]["email"])
+            self.assertEqual(user_profile.last_name, self.lecturers[i]["name"])
+            self.assertEqual(user_profile.first_name_given, self.lecturers[i]["christianname"])
+            self.assertEqual(user_profile.title, self.lecturers[i]["titlefront"])
+
+    def test_import_existing_lecturers(self):
+
+        user_profile = baker.make(UserProfile, email=self.lecturers[0]["email"])
+        print(user_profile.email)
```
richardebeling, line 112: empty first line, leftover print

evap/staff/tests/test_json_importer.py
```diff
@@ -0,0 +1,202 @@
...
+
+        importer = JSONImporter(self.semester)
+        importer._import_lecturers(self.lecturers)
+
+        assert UserProfile.objects.all().count() == 2
```
richardebeling, line 117: `self.assertEqual`

evap/staff/importers/json.py
```diff
@@ -0,0 +1,204 @@
...
+
+            course = self.course_map[event["relatedevents"]["gguid"]]
+
+            self._import_evaluation(course, event)
+
+    @transaction.atomic
+    def import_json(self, data: ImportDict):
+        self._import_students(data["students"])
+        self._import_lecturers(data["lecturers"])
+        self._import_events(data["events"])
```
richardebeling, line 204: If it's a "JsonImporter" with an `import_json` method, I would expect it to take json as input. I think that also makes sense, both for the tests and other users of the class.

evap/staff/tests/test_json_importer.py
```diff
@@ -0,0 +1,202 @@
...
+        self.students = EXAMPLE_DATA["students"]
+        self.lecturers = EXAMPLE_DATA["lecturers"]
+
+        self.semester = baker.make(Semester)
+
+    def test_import_students(self):
+        self.assertEqual(UserProfile.objects.all().count(), 0)
+
+        importer = JSONImporter(self.semester)
+        importer._import_students(self.students)
```
richardebeling, line 68: If we have this repeated over multiple tests, I'd vote in favor of `JsonImporter(self.semester)._import_students(self.students)` (no unused variable left over, one line less)

evap/staff/tests/test_json_importer.py
```diff
@@ -0,0 +1,202 @@
...
+        self.assertEqual(course.cms_id, EXAMPLE_DATA["events"][0]["gguid"])
+        self.assertEqual(course.name_de, EXAMPLE_DATA["events"][0]["title"])
+        self.assertEqual(course.name_en, EXAMPLE_DATA["events"][0]["title_en"])
+        self.assertEqual(course.type.name_de, EXAMPLE_DATA["events"][0]["type"])
+        self.assertListEqual(
+            [d.name_de for d in course.degrees.all()], [d["cprid"] for d in EXAMPLE_DATA["events"][0]["courses"]]
+        )
+        self.assertListEqual(
+            list(course.responsibles.all()),
+            [importer.user_profile_map[lecturer["gguid"]] for lecturer in EXAMPLE_DATA["events"][0]["lecturers"]],
```
richardebeling, line 152: I'd prefer to not have tests rely on the logic that is being tested: If the `user_profile_map` somehow is wrong, the importer would associate the wrong user profiles with the course, but the test wouldn't catch it because it uses the same (wrong) map. I think it's fine if you just hardcode the values we'd expect based on the input data.

evap/staff/tests/test_json_importer.py
hansegucker: Doesn't work because importer object is used later

evap/staff/tests/test_json_importer.py
```diff
@@ -0,0 +1,263 @@
...
+                    new_first_name_given=self.students[0]["christianname"],
+                )
+            ],
+        )
+
+    def test_import_lecturers(self):
+        self.assertEqual(UserProfile.objects.all().count(), 0)
+
+        importer = JSONImporter(self.semester)
+        importer._import_lecturers(self.lecturers)
```
richardebeling, line 112: In theory, you could always just inline these. I don't know if it's more readable though.

```suggestion
        JSONImporter(self.semester)._import_lecturers(self.lecturers)
```

evap/staff/tests/test_json_importer.py
```diff
@@ -0,0 +1,263 @@
...
+        self.assertEqual(main_evaluation.course, course)
+        self.assertEqual(main_evaluation.name_de, "")
+        self.assertEqual(main_evaluation.name_en, "")
+        # [{"begin": "15.04.2024 10:15", "end": "15.07.2024 11:45"}]
+        self.assertEqual(main_evaluation.vote_start_datetime, datetime(2024, 7, 8, 8, 0))
+        self.assertEqual(main_evaluation.vote_end_date, date(2024, 7, 21))
+        self.assertListEqual(
+            list(main_evaluation.participants.all()),
+            [importer.user_profile_map[student["gguid"]] for student in EXAMPLE_DATA["events"][0]["students"]],
+        )
```
richardebeling, line 177: Could be a problem due to order mismatch -- I think you can use assertSetEqual

test_data.json
```diff
@@ -0,0 +1,117 @@
+{
```
richardebeling, line 1: unused?

Personally, I favor more end-to-endy tests (allowing us to detect errors in calling the importer from the management command), but in general, I dislike having test data separately, it always gets messy to maintain those. I'd probably define this as a python dict in the corresponding test file, then.

evap/staff/importers/json.py
```diff
@@ -0,0 +1,325 @@
...
+        heading += "\n" + "".join(["-" for i in heading]) + "\n"
+        return heading
+
+    @staticmethod
+    def _make_total(total: int) -> str:
+        return f"({total} in total)\n\n"
+
+    def get_log(self) -> str:
+        log = "JSON IMPORTER REPORT\n"
+        log += "====================\n\n"
```
richardebeling, line 94: Maybe make `make_heading` above take a separator character, and simply pass `=` here then?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,325 @@
...
+        log += f"Import finished at {now()}\n\n"
+        log += self._make_heading("Name Changes")
+        for name_change in self.name_changes:
+            log += f"- {name_change.old_first_name_given} {name_change.old_last_name} → {name_change.new_first_name_given} {name_change.new_last_name}\n"
+        log += self._make_total(len(self.name_changes))
```
richardebeling, line 99: For readability, I'd put the count close towards the heading. If its hundreds of changes here, it's hard to associate the number.

evap/staff/importers/json.py
```diff
@@ -0,0 +1,325 @@
...
+
+        log += self._make_heading("New Courses")
+        for new_course in self.new_courses:
+            log += f"- {new_course}\n"
+        log += self._make_total(len(self.new_courses))
```
richardebeling, line 104: You could deduplicate the control flow / logic here, and simply iterate over a dict in the form of
```python3
{
   "New Courses": self.new_courses,
   "New Evaluations": self.new_evaluations,
   # ...
}
```

evap/staff/importers/json.py
```diff
@@ -0,0 +1,325 @@
...
+        self.course_type_cache[name] = course_type
+        return course_type
+
+    def _get_degree(self, name: str) -> Degree:
+        if name in self.degree_cache:
+            return self.degree_cache[name]
+
+        degree = Degree.objects.get_or_create(name_de=name, defaults={"name_en": name})[0]
+        self.degree_cache[name] = degree
+        return degree
```
richardebeling, line 154: These are duplicated with the user importer, right? Do we want to factor this out into a separate component?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,325 @@
...
+
+    def _import_students(self, data: list[ImportStudent]) -> None:
+        for entry in data:
+            email = clean_email(entry["email"])
+            user_profile, __, changes = update_or_create_with_changes(
+                UserProfile,
+                email=email,
+                defaults={"last_name": entry["name"], "first_name_given": entry["christianname"]},
+            )
+            user_profile: UserProfile
```
richardebeling, line 167: Shouldn't be necessary?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,325 @@
...
+            if changes:
+                change = NameChange(
+                    old_last_name=changes["last_name"][0] if changes.get("last_name") else user_profile.last_name,
+                    old_first_name_given=(
+                        changes["first_name_given"][0]
+                        if changes.get("first_name_given")
+                        else user_profile.first_name_given
+                    ),
+                    new_last_name=changes["last_name"][1] if changes.get("last_name") else user_profile.last_name,
```
richardebeling, line 176:
```suggestion
                    new_last_name=user_profile.last_name,
```
and so on?

(To me, it also makes sense to only create a `NameChange` instance if there was actually a name change, then we also don't need the `changes["last_name"][0] if changes.get("last_name") else user_profile.last_name` part above)

evap/staff/importers/json.py
```diff
@@ -0,0 +1,325 @@
...
+
+    def _import_lecturers(self, data: list[ImportLecturer]) -> None:
+        for entry in data:
+            email = clean_email(entry["email"])
+            user_profile, __ = UserProfile.objects.update_or_create(
+                email=email,
+                defaults={
+                    "last_name": entry["name"],
+                    "first_name_given": entry["christianname"],
+                    "title": entry["titlefront"],
```
richardebeling, line 195: I guess this is also a user profile change?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,325 @@
...
+        course_type = self._get_course_type(data["type"])
+        degrees = [self._get_degree(c["cprid"]) for c in data["courses"]]
+        responsibles = self._get_user_profiles(data["lecturers"])
+        course, created, changes = update_or_create_with_changes(
+            Course,
+            semester=self.semester,
+            cms_id=data["gguid"],
+            defaults={"name_de": data["title"], "name_en": data["title_en"], "type": course_type},
+        )
+        course: Course
```
richardebeling, line 211:
```suggestion
```

evap/staff/importers/json.py
```diff
@@ -0,0 +1,325 @@
...
+            evaluation_start_datetime = course_end.replace(hour=8, minute=0) - timedelta(
+                weeks=1, days=course_end.weekday()
+            )
+            # End date is on the sunday in the week the event ends
+            evaluation_end_date = (course_end + timedelta(days=6 - course_end.weekday())).date()
+
+            name_de, name_en = "", ""
+
+        # If events are graded for any degree, wait for grade upload before publishing
+        wait_for_grade_upload_before_publishing = any(filter(lambda grade: grade["scale"], data["courses"]))
```
richardebeling, line 247:
```suggestion
        wait_for_grade_upload_before_publishing = any(grade["scale"] for grade in data["courses"])
```

evap/staff/importers/json.py
```diff
@@ -0,0 +1,325 @@
...
+        if evaluation.state < Evaluation.State.APPROVED:
+            direct_changes = update_with_changes(evaluation, defaults)
+
+            participant_changes = set(evaluation.participants.all()) != set(participants)
+            evaluation.participants.set(participants)
+
+            lecturers_changes = False
+            for lecturer in data["lecturers"]:
+                __, lecturer_created, lecturer_changes = self._import_contribution(evaluation, lecturer)
+                if lecturer_changes or lecturer_created:
```
richardebeling, line 272: Can we maybe rename `lecturers_changes` and/or `lecturer_changes` to be more distinguishable?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,325 @@
...
+        contribution, created, changes = update_or_create_with_changes(
+            Contribution,
+            evaluation=evaluation,
+            contributor=user_profile,
+        )
+        return contribution, created, changes
+
+    def _import_events(self, data: list[ImportEvent]) -> None:
+        # Divide in two lists so corresponding courses are imported before their exams
+        normal_events = (event for event in data if not event["isexam"])
```
richardebeling, line 299: `non_exam_events`?

richardebeling (changes requested):

evap/staff/importers/json.py
```diff
@@ -81,47 +81,36 @@ class ImportStatistics:
     attempted_changes: list[Evaluation] = field(default_factory=list)
 
     @staticmethod
-    def _make_heading(heading: str) -> str:
-        heading += "\n" + "".join(["-" for i in heading]) + "\n"
-        return heading
+    def _make_heading(heading: str, separator: str = "-") -> str:
+        return "\n" + "".join([separator for i in heading]) + "\n"
```
richardebeling, line 85:
```suggestion
        return "\n" + separator * len(heading) + "\n"
```

evap/staff/importers/json.py
```diff
@@ -81,47 +81,36 @@ class ImportStatistics:
...
 
     @staticmethod
     def _make_total(total: int) -> str:
         return f"({total} in total)\n\n"
 
+    @staticmethod
+    def _make_stats(heading: str, new_objects: list):
```
richardebeling, line 92:
```suggestion
    def _make_stats(heading: str, new_objects: list) -> str:
```

evap/staff/importers/json.py
```diff
@@ -173,12 +161,8 @@ def _import_students(self, data: list[ImportStudent]) -> None:
...
                         else user_profile.first_name_given
                     ),
-                    new_last_name=changes["last_name"][1] if changes.get("last_name") else user_profile.last_name,
-                    new_first_name_given=(
-                        changes["first_name_given"][1]
-                        if changes.get("first_name_given")
-                        else user_profile.first_name_given
-                    ),
+                    new_last_name=user_profile.last_name,
+                    new_first_name_given=(user_profile.first_name_given),
```
richardebeling, line 165: Why a tuple?

evap/staff/importers/json.py
hansegucker: The mapper in the user importer expects objects to exist whether I want to create non-existing objects.

richardebeling (changes requested):

evap/staff/importers/json.py
```diff
@@ -0,0 +1,313 @@
...
+            return self.degree_cache[name]
+
+        degree = Degree.objects.get_or_create(name_de=name, defaults={"name_en": name})[0]
+        self.degree_cache[name] = degree
+        return degree
+
+    def _get_user_profiles(self, data: list[ImportRelated]) -> list[UserProfile]:
+        return [self.user_profile_map[related["gguid"]] for related in data]
+
+    def _create_name_change_from_changes(self, user_profile: UserProfile, changes: dict[str, tuple[any, any]]):
```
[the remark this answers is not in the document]
richardebeling: Ah, you're right, sorry. I just intended to use `str` whenever we have at attribute name (e.g. arguments we pass to `getattr`).

janno42 (changes requested):
not yet manually tested, because test data doesn't import (`course = self.course_map[event["relatedevents"]["gguid"]]     
KeyError: '0x6'`)

janno42:
If possible, we should remove the debug info text from the log email

evap/staff/importers/json.py
```diff
@@ -0,0 +1,313 @@
...
+    def _process_log(self) -> None:
+        log = self.statistics.get_log()
+        logger.info(log)
+
+    @transaction.atomic
+    def import_dict(self, data: ImportDict) -> None:
+        self._import_students(data["students"])
+        self._import_lecturers(data["lecturers"])
+        self._import_events(data["events"])
+        self._process_log()
```
richardebeling, line 310: just `logger.info(self.statistics.get_log())` (and get rid of `_process_log`)?

evap/staff/tests/test_json_importer.py
```diff
@@ -0,0 +1,295 @@
...
+        self.assertEqual(user_profile.first_name_given, self.students[0]["christianname"])
+
+        self.assertEqual(
+            importer.statistics.name_changes,
+            [
+                NameChange(
+                    old_last_name="Doe",
+                    old_first_name_given="Jane",
+                    new_last_name=self.students[0]["name"],
+                    new_first_name_given=self.students[0]["christianname"],
```
richardebeling, line 108: For staff user experience, it would probably make sense to also show the email address of the account.

(I'm guessing if the email is not contained in the `NameChange`, we don't currently log it in the output)

hansegucker:
> If possible, we should remove the debug info text from the log email

Done.

janno42:
Still missing

- [ ] use import names for `CourseType`s
- [ ] use import names for `Program`s

evap/staff/importers/json.py
```diff
@@ -0,0 +1,417 @@
...
+        return list(filter(lambda p: p.email not in settings.NON_RESPONSIBLE_USERS, user_profiles))
+
+    def _get_course_type(self, name: str) -> CourseType:
+        lookup = name.strip().lower()
+        if lookup in self.course_type_cache:
+            return self.course_type_cache[lookup]
+
+        course_type, __ = CourseType.objects.get_or_create(name_de=name, defaults={"name_en": name})
+        self.course_type_cache[name] = course_type
+        return course_type
```
richardebeling, line 187: This is always fully filled on `__init__`, so we should never not have an entry, right? (Same below for the non-covered line in `_get_program`)

```suggestion
        lookup = name.strip().lower()
        return self.course_type_cache[lookup]
```

evap/staff/importers/json.py
```diff
@@ -0,0 +1,417 @@
...
+            new_first_name_given=user_profile.first_name_given,
+            email=user_profile.email,
+        )
+        self.statistics.name_changes.append(change)
+
+    def _import_students(self, data: list[ImportStudent]) -> None:
+        for entry in data:
+            email = clean_email(entry["email"])
+            if not email:
+                self.statistics.warnings.append(
```
richardebeling, line 219: untested -- can we add a short test case for this?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,417 @@
...
+
+    # pylint: disable=too-many-locals
+    def _import_evaluation(self, course: Course, data: ImportEvent) -> Evaluation:
+        if "appointments" not in data:
+            self.statistics.warnings.append(
+                WarningMessage(obj=course.name, message="No dates defined, using default end date")
+            )
+            course_end = datetime.strptime(self.default_course_end, "%d.%m.%Y")
+        else:
+            last_appointment = sorted(data["appointments"], key=lambda x: x["end"])[-1]
```
richardebeling, line 295: can we just use `max`?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,417 @@
...
+        if created:
+            self.statistics.new_evaluations.append(evaluation)
+
+        return evaluation
+
+    def _import_contribution(
+        self, evaluation: Evaluation, data: ImportRelated
+    ) -> tuple[Contribution | None, bool, dict[str, tuple[any, any]]]:
+        if data["gguid"] not in self.user_profile_map:
+            return None, False, {}
```
richardebeling, line 378: Do we expect this to happen? If yes, in what case? Can we add a comment here and a test?

evap/staff/importers/json.py
```diff
@@ -284,6 +284,22 @@ def _import_course_programs(self, course: Course, data: ImportEvent) -> None:
...
         course.programs.set(programs)
 
+    def _import_course_from_unused_exam(self, data: ImportEvent) -> Course | None:
+        splitted_title = data["title"].split(":", 1)
+        if len(splitted_title) < 2:
+            return None
+        prefix = splitted_title[0].strip()
+
+        try:
+            course_type = CourseType.objects.get(import_names__contains=[prefix])
```
janno42, line 294: this should be `course_type = CourseType.objects.get(import_names__in=[prefix])` to ensure exact matches

evap/staff/importers/json.py
```diff
@@ -0,0 +1,447 @@
...
+        exam_events = (event for event in data if event["isexam"])
+
+        for event in non_exam_events:
+            course = self._import_course(event)
+
+            self._import_evaluation(course, event)
+
+        unused_events = []
+        for event in exam_events:
+            if not event["relatedevents"]:
```
janno42, line 420:
```suggestion
            if not "relatedevents" in event:
```

janno42 (changes requested):
- Currently all new evaluations are listed under "New Evaluations" and "Updated Evaluations" in the import log
- For all "unused events" the programs are currently not imported

evap/staff/importers/json.py
janno42: looks like array fields are weird - you can ignore this :)

evap/staff/importers/json.py
hansegucker: @janno42 Do we expect this to happen?

janno42: students and lecturers without email are skipped (see `_import_students` and `_import_lecturers`), so they are not in the list.

janno42:
@hansegucker tests are done. some mypy stuff is left, do you want to look into it?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,450 @@
...
+from evap.staff.tools import update_or_create_with_changes, update_with_changes
+
+logger = logging.getLogger("import")
+
+
+class ImportStudent(TypedDict):
+    gguid: str
+    email: str
+    name: str
+    christianname: str
```
niklasmohrin, line 23: I think I have seen enough for tonight :laughing:

evap/staff/importers/json.py
```diff
@@ -0,0 +1,450 @@
...
+        self.statistics = ImportStatistics()
+
+    def _choose_responsibles(self, user_profiles: list[UserProfile]) -> list[UserProfile]:
+        user_profiles_by_len: dict[int, list[UserProfile]] = {}
+        for user_profile in user_profiles:
+            user_profile_len = len(user_profile.title)
+            user_profiles_by_len.setdefault(user_profile_len, []).append(user_profile)
+        for key in sorted(user_profiles_by_len.keys(), reverse=True):
+            return user_profiles_by_len[key]
+        return []
```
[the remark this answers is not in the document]
niklasmohrin: Cursed version: `max(user_profiles_by_len.items(), default=("Niklas is cool", []))[1]` :^)

evap/staff/importers/json.py
```diff
@@ -0,0 +1,448 @@
...
+    gguid: str
+    email: str
+    name: str
+    christianname: str
+    titlefront: str
+
+
+class ImportCourse(TypedDict):
+    cprid: str
+    scale: str
```
niklasmohrin, line 37: Can you add a short comment explaining what these fields mean?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,448 @@
...
+
+        return self._import_course(data, course_type)
+
+    # pylint: disable=too-many-locals
+    def _import_evaluation(self, course: Course, data: ImportEvent) -> Evaluation:
+        if "appointments" not in data:
+            self.statistics.warnings.append(
+                WarningMessage(obj=course.name, message="No dates defined, using default end date")
+            )
+            course_end = datetime.strptime(self.default_course_end, "%d.%m.%Y")
```
niklasmohrin, line 309: You can leave this to the caller and take a date directly

evap/staff/importers/json.py
```diff
@@ -0,0 +1,448 @@
...
+
+
+class JSONImporter:
+    DATETIME_FORMAT = "%d.%m.%Y %H:%M:%S"
+
+    def __init__(self, semester: Semester, default_course_end: str) -> None:
+        self.semester = semester
+        self.default_course_end = default_course_end
+        self.user_profile_map: dict[str, UserProfile] = {}
+        self.course_type_cache: dict[str, CourseType] = {
```
niklasmohrin, line 156: I think we can improve the names here, how about `students_by_gguid` and `course_types_by_name` (and similar for the other variables)?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,448 @@
...
+            for program in Program.objects.all()
+            for import_name in program.import_names
+        }
+        self.course_map: dict[str, Course] = {}
+        self.statistics = ImportStatistics()
+
+    def _get_users_with_longest_title(self, user_profiles: list[UserProfile]) -> list[UserProfile]:
+        user_profiles_by_len = unordered_groupby((len(user.title), user) for user in user_profiles)
+        for key in sorted(user_profiles_by_len.keys(), reverse=True):
+            return user_profiles_by_len[key]
```
niklasmohrin, line 172: I think this is not what Richard meant :^)
```suggestion
```

evap/staff/importers/json.py
```diff
@@ -0,0 +1,448 @@
...
+        return user_profiles_by_len[max(user_profiles_by_len.keys())] if user_profiles_by_len else []
```
niklasmohrin, line 173: Actually, I would just replace the entire method with
```python
max_title_len = max((len(user.title) for user in user_profiles), default=0)
return [user for user in user_profiles if len(user.title) == max_title_len]
```

evap/staff/importers/json.py
```diff
@@ -0,0 +1,448 @@
...
+            return self.program_cache[lookup]
+
+        # It could happen that the importer needs a new program
+        program, __ = Program.objects.get_or_create(name_de=name, defaults={"name_en": name})
+        self.program_cache[name] = program
+        return program
+
+    def _get_user_profiles(self, data: list[ImportRelated]) -> list[UserProfile]:
+        return [
+            self.user_profile_map[related["gguid"]] for related in data if related["gguid"] in self.user_profile_map
```
niklasmohrin, line 200: Can it happen that such an id doesn't exist? And why do we ignore it if that happens?

niklasmohrin (changes requested):

evap/evaluation/models.py
```diff
@@ -333,6 +333,11 @@ class Course(LoggedModel):
     # grade publishers can set this to True, then the course will be handled as if final grades have already been uploaded
     gets_no_grade_documents = models.BooleanField(verbose_name=_("gets no grade documents"), default=False)
 
+    # unique reference for import from campus management system
+    cms_id = models.CharField(
+        verbose_name=_("campus management system id"), blank=True, null=True, unique=True, max_length=255
```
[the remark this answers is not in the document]
janno42: keeping it as is, because blank is needed for user interface, and blank + unique requires null. we have this in multiple places even after the cleanup in #1962

evap/staff/tools.py
```diff
@@ -412,3 +413,41 @@ def user_edit_link(user_id):
...
+
+    if created:
+        return obj, True, {}
+
+    changes = update_with_changes(obj, defaults)
+
+    return obj, False, changes
+
+
+def update_with_changes(obj: Model, defaults: dict[str, Any]) -> dict[str, tuple[Any, Any]]:
```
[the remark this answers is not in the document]
janno42: keeping the default: `defaults`

evap/staff/importers/json.py
janno42: are the current comments okay?

richardebeling (dismissed):

evap/staff/tests/test_json_importer.py
```diff
@@ -0,0 +1,510 @@
...
+                    new_first_name_given=self.lecturers[0]["christianname"],
+                    email=self.lecturers[0]["email"],
+                )
+            ],
+        )
+
+
+class TestImportEvents(TestCase):
+    def setUp(self):
+        self.semester = baker.make(Semester)
```
niklasmohrin:
```suggestion
    @classmethod
    def setUpTestData(cls):
        cls.semester = baker.make(Semester)
```

evap/staff/tests/test_json_importer.py
```diff
@@ -0,0 +1,510 @@
...
+    def test_management_command(self, mock_import_json):
+        output = StringIO()
+
+        with TemporaryDirectory() as temp_dir:
+            test_filename = os.path.join(temp_dir, "test.json")
+            with open(test_filename, "w", encoding="utf-8") as f:
+                f.write(EXAMPLE_JSON)
+            call_command("json_import", self.semester.id, test_filename, "01.01.2000", stdout=output)
+
+            mock_import_json.assert_called_once_with(EXAMPLE_JSON)
```
niklasmohrin, line 600: I would add this for coverage :)
```suggestion
            call_command("json_import", self.semester.id, test_filename, "01.01.2000", stdout=output)

            mock_import_json.assert_called_once_with(EXAMPLE_JSON)
            
            with self.assertRaises(CommandError):
                call_command("json_import", self.semester.id + 42, test_filename, "01.01.2000", stdout=output)

```

evap/staff/management/commands/json_import.py
```diff
@@ -0,0 +1,29 @@
...
+        parser.add_argument("semester", type=int)
+        parser.add_argument("file", type=str)
+        parser.add_argument("default_course_end", type=str)
+
+    def handle(self, *args, **options):
+        try:
+            semester = Semester.objects.get(pk=options["semester"])
+        except Semester.DoesNotExist:
+            self.stdout.write(self.style.ERROR("Semester does not exist."))
+            return
```
niklasmohrin: ... and my previous comment assumes that we change this as follows:
```suggestion
        except Semester.DoesNotExist as e:
            raise CommandError("Semester does not exist.") from e
```

Or is this explicitly designed to fail somewhat silently with exit code 0? In that case, there should be a test documenting that behavior

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+from django.conf import settings
+from django.core.mail import EmailMultiAlternatives
+from django.db import transaction
+from django.utils.timezone import now
+
+from evap.evaluation.models import Contribution, Course, CourseType, Evaluation, Program, Semester, UserProfile
+from evap.evaluation.tools import clean_email
+from evap.staff.tools import update_or_create_with_changes, update_with_changes
+
+logger = logging.getLogger("import")
```
niklasmohrin: The convention (and all other usages in EvaP today) do this:
```suggestion
logger = logging.getLogger(__name__)
```

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+                    )
+                )
+            else:
+                user_profile, __, changes = update_or_create_with_changes(
+                    UserProfile,
+                    email=email,
+                    defaults={
+                        "last_name": entry["name"],
+                        "first_name_given": entry["christianname"],
+                        "title": entry["titlefront"],
```
niklasmohrin, line 251: Just to be sure, students cannot have a title only lecturers?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+                    },
+                )
+                if changes:
+                    self._create_name_change_from_changes(user_profile, changes)
+
+                self.students_by_gguid[entry["gguid"]] = user_profile
```
niklasmohrin: As lecturers and students have separate importing treatment, I guess this dict should really be called something like `users_by_gguid`, right?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+
+    @transaction.atomic
+    def import_dict(self, data: ImportDict) -> None:
+        self._import_students(data["students"])
+        self._import_lecturers(data["lecturers"])
+        self._import_events(data["events"])
+        self.statistics.send_mail()
+
+    def import_json(self, data: str) -> None:
+        self.import_dict(json.loads(data))
```
niklasmohrin, line 501: Not a blocker for this PR, but it's a bit sad that the parsed object is just assumed to fit ImportDict, ideally we should validate that the data matches our expected schema.

We could look at pydantic (https://docs.pydantic.dev/dev/concepts/type_adapter/) or other solutions from here: https://stackoverflow.com/q/66665336/13679671

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+
+
+class ImportAppointment(TypedDict):
+    begin: str
+    end: str
+
+
+class ImportEvent(TypedDict):
+    gguid: str
+    lvnr: int
```
niklasmohrin: This field is not used for anything. I am fine with leaving it, but I would add a comment explaining what it means; deleting is also fine

Are the types otherwise complete here? Or are there more fields that we ignore? And do we want to document them or leave them be?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+            semester=self.semester,
+            cms_id=data["gguid"],
+            defaults={"name_de": data["title"], "name_en": data["title_en"], "type": course_type},
+        )
+        course.responsibles.set(responsibles)
+
+        if created:
+            self.statistics.new_courses.append(course)
+        elif changes:
+            self.statistics.updated_courses.append(course)
```
niklasmohrin, line 277: We might miss an update of the responsibles, because they are updated outside of `update_or_create_with_changes`

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+
+    def _import_course_programs(self, course: Course, data: ImportEvent) -> None:
+        programs = [
+            self._get_program(c["cprid"]) for c in data["courses"] if c["cprid"] not in settings.IGNORE_PROGRAMS
+        ]
+        course.programs.add(*programs)
+
+    def _import_course_from_unused_exam(self, data: ImportEvent) -> Course | None:
+        split_title = data["title"].split(":", 1)
+        if len(split_title) < 2:
```
niklasmohrin: I think this is slightly nicer as
```suggestion
        prefix, sep, actual_title = data["title"].partition(":")
        prefix = prefix.strip()
        actual_title = actual_title.strip()
        if not sep:
```
and then using actual_title below, but it's not super important

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+            return None
+        prefix = split_title[0].strip()
+
+        try:
+            course_type = CourseType.objects.get(import_names__contains=[prefix])
+        except CourseType.DoesNotExist:
+            return None
+
+        data["title"] = split_title[1].strip()
+        data["title_en"] = data["title_en"].split(":", 1)[1].strip() if ":" in data["title_en"] else data["title_en"]
```
```suggestion
        if ":" in data["title_en"]:
            data["title_en"] = data["title_en"].partition(":")[2].strip()
```

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+
+        return self._import_course(data, course_type)
+
+    # pylint: disable=too-many-locals
+    def _import_evaluation(self, course: Course, data: ImportEvent) -> Evaluation:
+        if "appointments" not in data:
+            self.statistics.warnings.append(
+                WarningMessage(obj=course.name, message="No dates defined, using default end date")
+            )
+            course_end = datetime.combine(self.default_course_end, time())
```
niklasmohrin: The `time()` here really threw me off, I thought that it was referring to `time.time` and thus get the current timestamp, but it is actually `datetime.time`, so it means a time of zero. Can we just add a class constant `self.MIDNIGHT` and make sure to not do `from datetime import time` but instead just use `datetime.time` in the one place we use it?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+            return None
+
+        data["title"] = split_title[1].strip()
+        data["title_en"] = data["title_en"].split(":", 1)[1].strip() if ":" in data["title_en"] else data["title_en"]
+
+        return self._import_course(data, course_type)
+
+    # pylint: disable=too-many-locals
+    def _import_evaluation(self, course: Course, data: ImportEvent) -> Evaluation:
+        if "appointments" not in data:
```
niklasmohrin: To avoid getting max of empty iterable:
```suggestion
        if "appointments" not in data or not data["appointments"]:
```

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+            self.statistics.warnings.append(
+                WarningMessage(obj=course.name, message="No dates defined, using default end date")
+            )
+            course_end = datetime.combine(self.default_course_end, time())
+        else:
+            last_appointment = max(data["appointments"], key=lambda x: x["end"])
+            course_end = datetime.strptime(last_appointment["end"], self.DATETIME_FORMAT)
```
niklasmohrin: Isn't this wrong? We are taking the maximum over the datetime _strings_, and with `DATETIME_FORMAT = "%d.%m.%Y %H:%M:%S"` this means that 30.01.24 would be after 01.12.24.

I think we need to parse all strings and then take the maximum in the datetime domain:
```suggestion
            course_end = max(datetime.strptime(app["end"], self.DATETIME_FORMAT) for app in data["appointments"])
```

And of course, we then need a test :)

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+
+        if data["isexam"]:
```
niklasmohrin, line 324: Note: this is one place for which I would really like to have schema validation, because if a string `"false"` sneaks in here we take the wrong branch. Should we maybe add an `assert isinstance(data["isexam"], bool)` here in the meantime?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+            # Set evaluation time frame of three days for exam evaluations:
+            evaluation_start_datetime = course_end.replace(hour=8, minute=0, second=0, microsecond=0) + timedelta(
+                days=1
+            )
+            evaluation_end_date = (course_end + timedelta(days=3)).date()
+
+            name_de = data["title"].split(" - ")[-1] if " - " in data["title"] else "Prüfung"
+            name_en = data["title_en"].split(" - ")[-1] if " - " in data["title_en"] else "Exam"
```
niklasmohrin, line 332: Not a blocker; these translations should be handled through our po file, not here in code

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+            # Start datetime is at 8:00 am on the monday in the week before the event ends
+            evaluation_start_datetime = course_end.replace(hour=8, minute=0, second=0, microsecond=0) - timedelta(
+                weeks=1, days=course_end.weekday()
+            )
+            # End date is on the sunday in the week the event ends
+            evaluation_end_date = (course_end + timedelta(days=6 - course_end.weekday())).date()
+
+            name_de, name_en = "", ""
+
+            weight = 9
```
niklasmohrin, line 356: In general, it's not super awesome that the knowledge about weights and timedeltas for exam evaluations are duplicated here; okay to fix in a follow up

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+        user_profile = self.students_by_gguid[data["gguid"]]
+
+        if user_profile.email in settings.NON_RESPONSIBLE_USERS:
+            return None, False, {}
+
+        contribution, created, changes = update_or_create_with_changes(
+            Contribution,
+            evaluation=evaluation,
+            contributor=user_profile,
+        )
```
niklasmohrin: Since `defaults` isn't set here, `changes` is empty in all return paths of this function, so we can just not return it at all - or is that not expected?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+    def _import_contribution(
+        self, evaluation: Evaluation, data: ImportRelated
+    ) -> tuple[Contribution | None, bool, dict[str, tuple[Any, Any]]]:
+        if data["gguid"] not in self.students_by_gguid:
+            return None, False, {}
+
+        user_profile = self.students_by_gguid[data["gguid"]]
+
+        if user_profile.email in settings.NON_RESPONSIBLE_USERS:
+            return None, False, {}
```
niklasmohrin: This line is not covered by tests, let's add a short one, otherwise the setting's usage is never checked

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+
+            self._import_evaluation(course, event)
+
+        unused_events = []
+        for event in exam_events:
+            if "relatedevents" not in event:
+                unused_events.append(event)
+                continue
+
+            exam_course = self.courses_by_gguid[event["relatedevents"][0]["gguid"]]
```
niklasmohrin: Do we assume that the list always contains exactly one entry? Let's assert

niklasmohrin: Also here: What is the meaning of `exam_course`? Is it a special kind of course only for the exam, no right?

And this course could have come from an import from the `non_exam_events`, or from an existing course in the database right? So we amend the previous import with more data from the exam event here.

Can we add some more commentary to this section of the code? Without having the spec from the issue in mind, I find it hard to follow

evap/staff/importers/json.py
```diff
@@ -0,0 +1,445 @@
...
+
+            self._import_evaluation(exam_course, event)
+        for event in unused_events:
+            course_from_unused_exam = self._import_course_from_unused_exam(event)
+            if not course_from_unused_exam:
+                self.statistics.warnings.append(
+                    WarningMessage(obj=event["title"], message="No related event or matching prefix found")
+                )
+                continue
+            event["isexam"] = False
```
niklasmohrin, line 457: Can you add a comment why we import in the non-exam way?

niklasmohrin (changes requested):
Looks good overall. Blocking because of one possibly critical bug (the datetime thing). Everything else is lower prio, but still easy enough to do that I would be more comfortable if changed.

I feel that the importer is not as robust as I would like, but should be good to go for a first release with the feedback applied. I tried to keep everything actionable, let me know what do you think :)

evap/staff/importers/json.py
niklasmohrin: yes, thanks

evap/staff/importers/json.py
niklasmohrin: Can you add this explanation as a comment?

evap/staff/importers/json.py
janno42: They can all stay as they are. This is the v1 of the importer, and there's already a list of ideas for how to upgrade it in the future. For now, it just needs to serve as a first version to test the most basic functionality in a production environment.

niklasmohrin: I am not suggesting to add complete typing today, I just want to know what the state of current typing is - is this everything, or is there more?

evap/staff/importers/json.py
hansegucker: I adapted the test data to cause this problem.

evap/staff/importers/json.py
niklasmohrin: @janno42 ?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,502 @@
...
+            self.statistics.warnings.append(
+                WarningMessage(
+                    obj=course.name, message="No 'courses' defined in import data, Programs can't be assigned"
+                )
+            )
+        else:
+            programs = [
+                self._get_program(c["cprid"]) for c in data["courses"] if c["cprid"] not in settings.IGNORE_PROGRAMS
+            ]
+            course.programs.add(*programs)
```
niklasmohrin, line 294: Is this something we want to track as changed?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,502 @@
...
+        evaluation, created = Evaluation.objects.get_or_create(
+            course=course,
+            cms_id=data["gguid"],
+            defaults=defaults,
+        )
+        if evaluation.state < Evaluation.State.APPROVED:
+            direct_changes = update_with_changes(evaluation, defaults)
+
+            participant_changes = set(evaluation.participants.all()) != set(participants)
+            evaluation.participants.set(participants)
```
niklasmohrin, line 380: Should we use our helper here? Or is it wasteful when we don't actually care about the old values?

evap/staff/importers/json.py
```diff
@@ -0,0 +1,502 @@
...
+            any_lecturers_changed = False
+            if "lecturers" not in data:
+                self.statistics.warnings.append(
+                    WarningMessage(obj=evaluation.full_name, message="No contributors defined")
+                )
+            else:
+                for lecturer in data["lecturers"]:
+                    __, lecturer_created = self._import_contribution(evaluation, lecturer)
+                    if lecturer_created:
+                        any_lecturers_changed = True
```
```suggestion
                    any_lecturers_changed |= True
```

niklasmohrin (approved):
Great work :ship:

PR #2190, opened 2024-05-14
Update a single comment :^)

niklasmohrin:
Found this during review. It is possible that the initial intention was to highlight that the questionnaires belong to the person and not the evaluation - in that case, I would opt for adding punctuation like

> A contributor, who is assigned to an evaluation, and their questionnaires.


richardebeling (approved):

Kakadus (approved):
Both with and without comma it's okay

PR #2194, opened 2024-05-20
fix dependabot composite action update

Kakadus:
After it apparently does not worked, I dug a bit deeper. Looking at a concrete example at [ansible-community/github-docs-build](https://www.github.com/ansible-community/github-docs-build/blob/c87c4e381fe8ef464e9d0318043ca1284af66b01/.github/dependabot.yml#L13) this should now finally be it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -32,7 +32,7 @@ updates:
     labels:
       - "[T] Dependencies"
   - package-ecosystem: "github-actions"
-    directory: "/actions/setup_nodejs"
+    directory: "/.github/setup_nodejs"
     schedule:
       interval: "weekly"
     groups:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -32,7 +32,7 @@ updates:
     labels:
       - "[T] Dependencies"
   - package-ecosystem: "github-actions"
-    directory: "/actions/setup_nodejs"
+    directory: "/.github/setup_nodejs"
```
richardebeling, line 35: Random question: What about `.github/workflows/tests.yml`? That also uses `actions/checkout@v4` and `actions/download-artifact@v4`, so it would also require updating, right?

richardebeling (approved):

.github/dependabot.yml
Kakadus: Those _are_ covered by the "/" directory above successfully (and were already updated by dependabot)

PR #2198, opened 2024-05-27
Fix person management none error

Kakadus:


evap/staff/views.py
```diff
@@ -1476,6 +1463,16 @@ def evaluation_person_management(request, evaluation_id):
             importer_log.forward_messages_to_django(request)
             return redirect("staff:semester_view", evaluation.course.semester.pk)
 
+        #  "test" in operation
```
richardebeling, line 1466: Maybe make it run-time checked?
```suggestion
       assert "test" in operation
```

evap/staff/views.py
```diff
@@ -1461,11 +1447,12 @@ def evaluation_person_management(request, evaluation_id):
                 delete_import_file(request.user.id, import_type)
             elif "copy" in operation:
                 copy_form.evaluation_selection_required = True
-                if copy_form.is_valid():
-                    import_evaluation = copy_form.cleaned_data["evaluation"]
-                    importer_log = import_persons_from_evaluation(
-                        import_type, evaluation, test_run=False, source_evaluation=import_evaluation
-                    )
+                if not copy_form.is_valid():
+                    raise SuspiciousOperation("Invalid copy form")
```
richardebeling, line 1451: What happens when I don't select an evaluation from the dropdown? (Possibly with JS disabled?)

richardebeling (changes requested):

evap/staff/tests/test_views.py
```diff
@@ -2389,6 +2389,23 @@ def test_copy_participants(self):
...
+            participants=self.evaluation2.participants.all(),
+        )
+        old_pk = old_evaluation.pk
+        old_evaluation.delete()
+
+        form = page.forms["participant-copy-form"]
+        form["pc-evaluation"] = str(self.evaluation2.pk)
+        params = dict(form.submit_fields() + [("operation", "copy-replace-participants")])
+        params["pc-evaluation"] = str(old_pk)
+        form.response.goto(form.action, method=form.method, params=params, status=400)
```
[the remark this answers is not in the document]
Kakadus: I can't do that, I then get `ValueError("Option '7' not found (from '', '5', '6')")` while setting the value :( 
I somewhat copied this code from `submit_with_modal` but don't know a better solution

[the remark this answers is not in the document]
niklasmohrin: The `submit` call checks that you only submit what is possible from the HTML - if there is no submit button with the given value, it will complain

For modals, I used this code because a modal with name and value attributes behaves like a submit button, but webtest does not know our modals of course

I havent read the diff here closely enough yet to judge whether the usage of this trick is appropriate here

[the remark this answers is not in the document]
richardebeling: I would have expected everything to work here since we're `self.app.get`ting the page before we delete the `old_evaluation`, so for webtest, the PK should still be a valid option. Will have to look into this.

evap/staff/views.py
Kakadus: We get into the same problem even with JS enabled

evap/staff/tests/test_views.py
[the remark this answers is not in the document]
Kakadus: you are right! I just get the page before creating the evaluation. I'll update this accordingly.

evap/staff/views.py
```diff
@@ -1461,11 +1449,24 @@ def evaluation_person_management(request, evaluation_id):
...
+                        "staff_evaluation_person_management.html",
+                        {
+                            "semester": evaluation.course.semester,
+                            "evaluation": evaluation,
+                            "participant_excel_form": participant_excel_form,
+                            "participant_copy_form": participant_copy_form,
+                            "contributor_excel_form": contributor_excel_form,
+                            "contributor_copy_form": contributor_copy_form,
+                            "importer_log": importer_log,
+                        },
```
Kakadus, line 1464: this is a bit duplicated but I don't know of any other way to redirect to the current page with the invalid copy form for the error. Any ideas?

evap/staff/forms.py
```diff
@@ -143,7 +143,7 @@ def __init__(self, *args, **kwargs):
         self.fields["evaluation"].choices = choices
 
     def clean(self):
-        if self.evaluation_selection_required and self.cleaned_data["evaluation"] is None:
+        if self.evaluation_selection_required and self.cleaned_data.get("evaluation") is None:
```
Kakadus, line 146: just stumbled on this bug: If there was no `evaluation` in `self.cleaned_data` we'd just get a `ValueError` instead of a `ValidationError`

evap/staff/views.py
```diff
@@ -1476,6 +1477,16 @@ def evaluation_person_management(request, evaluation_id):
...
+        if excel_form.is_valid():
+            excel_file = excel_form.cleaned_data["excel_file"]
+            file_content = excel_file.read()
+            importer_log = import_persons_from_file(import_type, evaluation, test_run=True, file_content=file_content)
+            if not importer_log.has_errors():
+                save_import_file(excel_file, request.user.id, import_type)
+
     participant_test_passed = import_file_exists(request.user.id, ImportType.PARTICIPANT)
     contributor_test_passed = import_file_exists(request.user.id, ImportType.CONTRIBUTOR)
     # casting warnings to a normal dict is necessary for the template to iterate over it.
```
niklasmohrin, line 1492: This comment does not refer to anything anymore, let's delete it :)

niklasmohrin (approved):

evap/staff/views.py
niklasmohrin: you could only write it once by defining
```python3
render_self = lambda: render(
    ...
)
```
but I don't know if that is an improvement

evap/staff/tests/test_views.py
```diff
@@ -2389,6 +2389,26 @@ def test_copy_participants(self):
...
+            course__semester=self.evaluation2.course.semester,
+            participants=self.evaluation2.participants.all(),
+        )
+        page = self.app.get(self.url, user=self.manager)
+        old_pk = old_evaluation.pk
+        old_evaluation.delete()
+
+        form = page.forms["participant-copy-form"]
+        form["pc-evaluation"] = old_pk
+        res = submit_with_modal(page, form, name="operation", value="copy-participants", status=200)
```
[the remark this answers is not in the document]
richardebeling: Afaik Django's handling is that a business logic failure causes a 200 and just renders the form again with an error message added for the user. (We do raise a `ValidationError` in the form validation, so what we observe here is just default django behavior).

I think this is fine.

richardebeling, line 2405: I think other code typically uses `page` or `response`, I don't think we've ever used `res` in the code base. For consistency, I'd prefer to use one of the existing variable names.

evap/staff/views.py
richardebeling: The standard django control flow would be to just fall through to the default `render` call at the end of the method in case any form fails to validate.

The control flow of the view was designed with the idea that a non-test operation cannot fail, so a fallthrough from the import operation would always end in the `redirect` that currently prevents us from applying the above logic here. Maybe, now that it is more complex, it should be split up into more components anyway, or maybe the redirect should be moved, idk.

We could accept duplicating the call and just try to deduplicate the kwargs. However, the `test_passed` args depend on the test run happening below.

Random thought: An immediately invoked function could use `return` to "jump" to the end while skipping the offending redirect -- I've never seen that done in python though, I think the syntax would require giving the function a name anyway?

evap/staff/tests/test_views.py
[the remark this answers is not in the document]
niklasmohrin: I thought the `ValidationError` is caught by `is_valid` on the form and we then decide ourselves to handle `is_valid() == False` by returning `render(...)`. Either way, if you say it's fine and Django-conformant, I have nothing to complain

evap/staff/views.py
niklasmohrin: Yeaaaaah, I think immediately invoked function is not really a thing I think and I also think that we would have to give a name

Green light from me for a bigger refactor :D

evap/staff/tests/test_views.py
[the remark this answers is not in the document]
richardebeling: True. Our views follow the approach that [django suggests in their docs](https://docs.djangoproject.com/en/5.0/topics/forms/#the-view), though. If we want to adopt 400 http codes on invalid forms, we should do that globally, not just here

evap/staff/views.py
Kakadus: I extracted a `import_or_copy_participants` returning the success of the operation. If successful, we can easily redirect to the semester page, otherwise just fallthrough to the common render() call. What do you think about it?

evap/staff/tests/test_views.py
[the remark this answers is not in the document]
Kakadus: I don't reeealy think this improves something... I could be that this does not integrate well with some other view...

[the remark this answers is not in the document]
niklasmohrin: Yup, like I said nothing wrong here if that is what Django suggests - just thought this was odd when reading the test

evap/staff/views.py
Kakadus: We could inline `copy_form = participant_copy_form if "participants" in operation else contributor_copy_form` or move it to the function as well. We wouldn't need the pylint ignore then, but I don't think this is better...

evap/staff/views.py
```diff
@@ -1386,6 +1386,36 @@ def evaluation_email(request, evaluation_id):
     )
 
 
+def import_or_copy_participants(request, operation, import_type, evaluation, copy_form) -> bool:
+    if "test" in operation:
+        return False
```
niklasmohrin, line 1391: This should not be possible given the calling code

evap/staff/views.py
```diff
@@ -1386,6 +1386,36 @@ def evaluation_email(request, evaluation_id):
...
+
+    if "replace" in operation:
+        deleted_person_count, deletion_message = helper_delete_users_from_evaluation(evaluation, operation)
+
+    if "import" in operation:
```
niklasmohrin, line 1396: I think we should just not pass `operation` and instead have `replace: bool` and `action: ImportAction` (or similar) where `ImportAction` is an enum with variants `Copy` and `Import`

evap/staff/views.py
```diff
@@ -1449,36 +1479,11 @@ def evaluation_person_management(request, evaluation_id):
...
-                    )
-
-            if "replace" in operation:
-                importer_log.add_success(
-                    format_html(deletion_message, deleted_person_count, evaluation.full_name),
-                    category=ImporterLogEntry.Category.RESULT,
-                )
-
-            importer_log.forward_messages_to_django(request)
+        elif import_or_copy_participants(request, operation, import_type, evaluation, copy_form):
```
niklasmohrin, line 1482: I would prefer if we would not have the call to `import_or_copy_participants` in the condition here - what do you think about making this an `else` and assigning `successfully_processed = import_or_copy_participants(...)` in there?

evap/staff/views.py
niklasmohrin: I think moving into the function is not the right move, I think leaving it as is or inlining to the call (maybe together with https://github.com/e-valuation/EvaP/pull/2198#discussion_r1641942557) both sounds fine to me :)

niklasmohrin:
(depending on how urgent fixing the error is, you might as well dismiss my comments for now :D)

evap/staff/views.py
Kakadus: yes. Should this become an assertion instead?

Kakadus: (resolved with other points)

deployment/localsettings.template.py
```diff
@@ -1,12 +1,12 @@
...
-    'default': {
-        'ENGINE': 'django.db.backends.postgresql',   # postgresql', 'mysql', 'sqlite3' or 'oracle'.
-        'NAME': 'evap',                         # Or path to database file if using sqlite3.
-        'USER': 'evap',                         # Not used with sqlite3.
-        'PASSWORD': 'evap',                     # Not used with sqlite3.
-        'HOST': '127.0.0.1',                    # Set to empty string for localhost. Not used with sqlite3.
-        'PORT': '',                             # Set to empty string for default. Not used with sqlite3.
-        'CONN_MAX_AGE': 600,
+    "default": {
+        "ENGINE": "django.db.backends.postgresql",  # postgresql', 'mysql', 'sqlite3' or 'oracle'.
```
niklasmohrin, line 3: side note: we should update this comment, as we only support postgres nowadays, right?

evap/staff/views.py
```diff
@@ -1,10 +1,11 @@
 import csv
+from enum import StrEnum
```
niklasmohrin, line 2: Starting in Python 3.11 :(

Any other enum is fine for us though, so no big problem

niklasmohrin (approved):
Cool :)

deployment/localsettings.template.py
Kakadus: oops, I did not intended to push (or even edit this in the first place...). I'll revert this here and move to a different PR

evap/staff/views.py
```diff
@@ -1386,22 +1388,67 @@ def evaluation_email(request, evaluation_id):
...
+
+    @classmethod
+    def from_operation(cls, operation: str) -> "ImportAction":
+        if "copy" in operation:
+            return cls.COPY
+        if "import" in operation:
+            return cls.IMPORT
+        if "test" in operation:
+            return cls.TEST
+        raise ValueError(f"Unknown operation: {operation}")
```
richardebeling, line 1404: `operation.startswith("copy-")` and similar would allow me not to worry about weird unintended substring matches

evap/staff/views.py
```diff
@@ -1448,37 +1496,15 @@ def evaluation_person_management(request, evaluation_id):
...
-            if "replace" in operation:
-                importer_log.add_success(
-                    format_html(deletion_message, deleted_person_count, evaluation.full_name),
-                    category=ImporterLogEntry.Category.RESULT,
-                )
-
-            importer_log.forward_messages_to_django(request)
-            return redirect("staff:semester_view", evaluation.course.semester.pk)
+            successfully_processed = import_or_copy_participants(
+                request, "replace" in operation, import_action, import_type, evaluation, copy_form
```
richardebeling, line 1501:
```suggestion
                request, "-replace-" in operation, import_action, import_type, evaluation, copy_form
```

richardebeling (approved):

Kakadus:
this is odd: pylint locally on 3.12 works, in vagrant fails. Looking at the code again, this should work! :angry:

evap/rewards/views.py
```diff
@@ -66,7 +66,7 @@ def redeem_reward_points(request):
                 error_string = _(
                     "It appears that your browser sent multiple redemption requests. You can see all successful redemptions below."
                 )
-            case _:
+            case _:  # abc
```
janno42, line 69:
```suggestion
            case _:
```

pyproject.toml
```diff
@@ -108,6 +108,10 @@ source = ["evap", "tools"]
 
 [tool.coverage.report]
 omit = ["*/migrations/*", "*__init__.py"]
+exclude_also = [
+    # mypy asserts these cases are not possible
+    "assert_never(.*)",
```
richardebeling, line 113: Looks like a regex to me -- the parentheses then aren't literal parentheses, but it looks like that's what you wanted?

evap/staff/views.py
```diff
@@ -1386,23 +1388,67 @@ def evaluation_email(request, evaluation_id):
...
     else:
-        assert "contributors" in operation
-        deleted_person_count = evaluation.contributions.exclude(contributor=None).count()
-        deletion_message = _("{} contributors were deleted from evaluation {}")
-        evaluation.contributions.exclude(contributor=None).delete()
+        assert_never(import_action)
 
-    return deleted_person_count, deletion_message
+    if replace:
+        importer_log.add_success(  # pylint: disable=possibly-used-before-assignment  # see https:www.//github.com/pylint-dev/pylint/issues/9780
```
richardebeling, line 1440: I don't understand the exclusion. If `replace == True` but `import_type` is any other value than `PARTICIPANT` or `CONTRIBUTOR`, the variable `deletion_message` will not be initialized here. I guess it's just incidental that we expect no other import type? But that's nothing pylint can see, we should just `assert` that then?

(The URL is invalid, misplaced `//`)

Kakadus: but the ci said: `evap/staff/views.py:1440:8: E0606: Possibly using variable 'importer_log' before assignment (possibly-used-before-assignment)`. However, the case you are describing seems to be a separate problem

richardebeling: Ok, I agree that pylint triggering for `importer_log` here is a false positive. I was just mentally checking the full statement :D

Kakadus:
```suggestion
        importer_log.add_success(  # pylint: disable=possibly-used-before-assignment  # see https://www.github.com/pylint-dev/pylint/issues/9780
```

Kakadus:
after pylint dependabot later this evening, this can finally be merged :tada:

pyproject.toml
```diff
@@ -108,6 +108,10 @@ source = ["evap", "tools"]
 
 [tool.coverage.report]
 omit = ["*/migrations/*", "*__init__.py"]
+exclude_also = [
+    # mypy asserts these cases are not possible
+    '\n.*assert_never(\(.*\))',
```
richardebeling, line 113: The intention of the leading `\n` is so that the whole `else`-block is excluded, right? I would add that as a comment (wasn't intuitive to me).

What is the purpose of the capturing group? Isn't clear to me from the coverage docs.

evap/staff/views.py
```diff
@@ -1386,23 +1388,67 @@ def evaluation_email(request, evaluation_id):
...
-        deletion_message = _("{} contributors were deleted from evaluation {}")
-        evaluation.contributions.exclude(contributor=None).delete()
+        assert_never(import_action)
 
-    return deleted_person_count, deletion_message
+    if replace:
+        importer_log.add_success(  # pylint: disable=possibly-used-before-assignment  # see https://www.github.com/pylint-dev/pylint/issues/9780
+            format_html(deletion_message, deleted_person_count, evaluation.full_name),
+            category=ImporterLogEntry.Category.RESULT,
+        )
```
richardebeling, line 1443: I'd expect the `pylint` comment to still mask issues with the possible usage of `deletion_message` etc. before usage.

Pylint has been updated in the meantime, with 3.2.6 I don't get the pylint error here, so I guess we should be fine without the `disable` comment.

evap/staff/views.py
```diff
@@ -1386,23 +1388,67 @@ def evaluation_email(request, evaluation_id):
...
+) -> bool:
+    if replace:
+        if import_type == ImportType.PARTICIPANT:
+            deleted_person_count = evaluation.participants.count()
+            deletion_message = _("{} participants were deleted from evaluation {}")
+            evaluation.participants.clear()
+        elif import_type == ImportType.CONTRIBUTOR:
+            deleted_person_count = evaluation.contributions.exclude(contributor=None).count()
+            deletion_message = _("{} contributors were deleted from evaluation {}")
+            evaluation.contributions.exclude(contributor=None).delete()
```
richardebeling, line 1423: Maybe add `else: assert_never(import_type)`?

richardebeling (changes requested):

evap/staff/views.py
Kakadus: yep :+1:

pyproject.toml
Kakadus: yes, it's https://coverage.readthedocs.io/en/latest/excluding.html#multi-line-exclusion-regexes

I tried to find a regex which excludes the whole else branch but my regex skills were to limited (maybe with some positive lookahead or so, but I did not really see the point there

evap/staff/views.py
Kakadus: done, but it needs a type: ignore instead, which is fixed on master though

niklasmohrin (approved):

pyproject.toml
richardebeling: I think the newlines do exactly what we want, just the capturing groups are not necessary. This should work:
```regex
\n.*assert_never\(.*\)
```
(To understand the intention of the "weird" newline, just add a comment, something along the lines of
```python3
# newline disables coverage for the whole block, see
# https://coverage.readthedocs.io/en/latest/excluding.html#multi-line-exclusion-regexes
```

richardebeling (approved):
Looks good to me once the extra parentheses (see https://github.com/e-valuation/EvaP/pull/2198#discussion_r1714190200) are fixed.

janno42 (approved):

Issue #2082, opened 2023-12-03
`sortable_form.js`: Convert to TypeScript, remove jQuery


richardebeling:
As a follow-up to #2036, we should
* convert `evap/static/js/sortable_form.js` to TypeScript (consistent file naming should then use a hyphen instead of an underscore)
* remove the dependency on jQuery since the [underlying library](https://sortablejs.github.io) doesn't depend on it. This might be a bit more work as all usages of `makeFormSortable` use jquery types iirc.
    Edit: We also use django-dynamic-formsets / `jquery.formset.js` in there. I don't see a simple non-jquery replacement, so it's probably just fine to keep using jquery for that.

richardebeling:
Marking as done for now as the main intention was `sortable_form.js`. Keeping jQuery around, for now, for django-dynamic-formsets/`jquery.formset.js`. Thanks @PFischbeck!

janno42:
note: Sisyphus.js also uses jQuery

PR #2199, opened 2024-05-27
Convert sortable-form and its usages to TypeScript

PFischbeck:
This helps (maybe resolves?) #2082. In particular, `sortable-form.js` is now in TypeScript, and the scripts using it also don't use jQuery anymore.
Note that `jquery.formset.js` still requires jQuery...

Possible future things:
- Update `sortablejs` dependency (remember to update the JS types and remove the "ignore" for dependabot then)
- Replace `jquery.formset.js` by some TS version that does not depend on jQuery

Please test the changed forms to make sure nothing broke :)


evap/static/ts/src/sortable-form.ts
```diff
@@ -0,0 +1,102 @@
...
+            row.querySelectorAll("select").forEach(el => {
+                el.querySelectorAll("option[selected]").forEach(option => {
+                    option.removeAttribute("selected");
+                });
+                el.querySelector("option")?.setAttribute("selected", "selected");
+            });
+
+            //Check the first item in every button group
+            row.querySelectorAll(".btn-group").forEach(group => {
+                group.querySelector("input")?.setAttribute("checked", "checked");
```
Kakadus, line 74: Could it be that we need to do this instead?
```suggestion
                group.querySelector("input")?.checked = true;
```
Ref: https://stackoverflow.com/a/27906003

evap/static/ts/src/sortable-form.ts
```diff
@@ -0,0 +1,105 @@
...
+            (row.querySelectorAll("input[id$=-order]") as NodeListOf<HTMLInputElement>).forEach(input => {
+                input.value = row.parentElement?.childElementCount.toString() ?? "";
+            });
+
+            // We have to empty the formset, otherwise sometimes old contents from
+            // invalid forms are copied (#644).
+            // Checkboxes with 'data-keep' need to stay checked.
+            row.querySelectorAll("input[type=checkbox]:not([data-keep]),input[type=radio]").forEach(el => {
+                el.removeAttribute("checked");
+            });
```
niklasmohrin, line 57:
```suggestion
            row.querySelectorAll<HTMLInputElement>("input[type=checkbox]:not([data-keep]),input[type=radio]").forEach(el => {
                el.checked = false;
            });
```

evap/static/ts/src/sortable-form.ts
```diff
@@ -0,0 +1,105 @@
...
+                input => {
+                    input.value = "";
+                },
+            );
+
+            row.querySelectorAll("select").forEach(el => {
+                el.querySelectorAll("option[selected]").forEach(option => {
+                    option.removeAttribute("selected");
+                });
+                el.querySelector("option")?.setAttribute("selected", "selected");
```
niklasmohrin, line 69: Similarly, this should select `HTMLOptionElement` and set `.selected = true;`

evap/static/ts/src/sortable-form.ts
```diff
@@ -0,0 +1,105 @@
...
+        new Sortable(tableBody, {
+            draggable: ".sortable",
+            handle: ".fa-up-down",
+            scrollSensitivity: 70,
+        });
+
+        document.querySelectorAll("form").forEach(form => {
+            form.addEventListener("submit", () => {
+                applyOrdering();
+                return true;
```
niklasmohrin, line 101: I think the return of `true` does nothing, only `return false` would be equivalent (but inferior) to `event.preventDefault()`

evap/static/ts/src/sortable-form.ts
```diff
@@ -0,0 +1,105 @@
...
+                el.remove();
+            });
+
+            rowAdded(row);
+        },
+        formTemplate: usesTemplate ? ".form-template" : null,
+    });
+
+    const tableBody = document.querySelector(`#${tableId} tbody`) as HTMLElement | null;
+    if (tableBody) {
```
niklasmohrin, line 91: I suppose we want to fail if the query fails? In that case, let's use `selectOrError` from our `utils.ts` file (it also supports the `selectOrError<HTMLElement>` usage like `querySelector` and `querySelectorAll`)

evap/static/ts/src/sortable-form.ts
```diff
@@ -0,0 +1,105 @@
+declare const Sortable: typeof import("sortablejs");
+
+class FormsetOptions {
+    prefix: string = "form";
+    deleteCssClass: string = "delete-row";
+    deleteText: string = "remove";
+    addText: string = "add another";
+    added: null | ((arg: JQuery<HTMLTableRowElement>) => void) = null;
+    formTemplate: string | null = null;
+}
```
niklasmohrin, line 10: How do you feel about the following?
```suggestion
interface FormsetOptions {
    prefix: string;
    deleteCssClass: string;
    deleteText: string;
    addText: string;
    added: (arg: JQuery<HTMLTableRowElement>) => void;
    formTemplate: string | null;
}
```
Using `interface`, we don't need to the default values; I also removed the `null |` for `added`, since we don't use it. What do you think?

niklasmohrin (approved):
Code looks all good to me, haven't verified on the website

The commit where you added `import { makeFormSortable } ...` instead of the script tag has a message starting with "Fix" - I am curious, was the previous version not working anymore?

Either way, I prefer the new version with the module import over the script tag. :+1:

PFischbeck:
> The commit where you added `import { makeFormSortable } ...` instead of the script tag has a message starting with "Fix" - I am curious, was the previous version not working anymore?

Correct! Once I added `import { selectOrError } from "./utils.js";`, it failed. To be honest, I can't remember exactly why, but this has something to do with modules and scripts, where I can't do imports within scripts. Switching to using `sortable-form` as a module import instead of a script fixed it.

niklasmohrin:
Ah right, it has to itself be a module to import from other modules. Alright!

janno42 (commented):
Ordering doesn't seem to work on any page. @richardebeling can you please test the frontend?

richardebeling:
On http://localhost:8000/staff/faq/, I also get
> Uncaught TypeError: row.querySelector(...) is null
>    rowChanged http://localhost:8000/staff/faq/:815
>    applyOrdering http://localhost:8000/static/js/sortable-form.js:5
>    applyOrdering http://localhost:8000/static/js/sortable-form.js:4
>    makeFormSortable http://localhost:8000/static/js/sortable-form.js:72

failing at
```
const nameDe = row.querySelector('input[id$=-title_de]').value;
```
while `row` is the header row (that doesn't contain any input fields).

PFischbeck:
That's weird, I can't reproduce this error locally, neither in Firefox nor Safari. I open http://localhost:8000/staff/faq/ and am able to drag and reorder the items.

richardebeling:
> That's weird, I can't reproduce this error locally, neither in Firefox nor Safari. I open http://localhost:8000/staff/faq/ and am able to drag and reorder the items.

Sorry, I should have added that: Drag+Drop works fine. The error happens when you click the submit button, after reloading the page the order is reset to the initial order.

richardebeling (approved):
Works for me, code looks good. Thanks!

package.json
```diff
@@ -4,7 +4,7 @@
         "@types/jest": "^29.5.12",
         "@types/jest-environment-puppeteer": "^5.0.3",
         "@types/jquery": "^3.5.16",
-        "@types/sortablejs": "^1.15.1",
+        "@types/sortablejs": "^1.3.32",
```
janno42, line 7: Should we document why?

janno42 (commented):
reordering the FAQ entries works for me but reordering the contributors on the evaluation edit page (e.g. `http://localhost:8000/staff/evaluation/1499/edit`) does not

package.json
richardebeling: I think its fine to have the version explicitly listed here, documentation is annoying because `.json` doesn't allow comments.

We should, however, use the "any patch version" marker (`~`), not the "any minor version" (`^`), if we want to keep the versions synced.

janno42: I meant at least writing somewhere in this PR why we have to use this older version

niklasmohrin: From the first comment, I thought that this would match the version of sortable that we have checked in in our repository

richardebeling: Yes, [our static file](https://github.com/e-valuation/EvaP/blob/main/evap/static/js/Sortable.min.js) is version 1.3.0-rc2. There is no release with patch-version 0 for the types package, but since this is for typechecking anyways, I think patch 32 is fine as well.

Note: I also explicitly accepted a higher minor version in another PR in the past -- I'm fine with both approaches. The matching version here is probably less surprising/confusing. We just have to make sure that dependabot doesn't come along and updates this again. I don't think we have an exclusion for this, yet.

evap/evaluation/templates/evap_evaluation_edit_js.html
```diff
@@ -1,32 +1,32 @@
...
-    <script src="{% static 'js/sortable_form.js' %}"></script>
-    <script type="text/javascript">
-        rowChanged = function(row) {
-            name = $(row.find("select[id$=-contributor]")).find(":selected").text();
-            nameChanged = name && name != "---------";
-            idChanged = row.find("input[id$=-id]").val();
+    <script type="module">
+        import { makeFormSortable } from "{% static 'js/sortable-form.js' %}";
+        const rowChanged = function(row) {
+            const name = row.querySelector("select[id$=-contributor]").querySelector(":selected").textContent;
```
Kakadus, line 7: Related to https://github.com/e-valuation/EvaP/pull/2199#pullrequestreview-2164156651

this fails for me, when submitting the form with
> SyntaxError: Element.querySelector: ':selected' is not a valid selector
```suggestion
            const name = row.querySelector("select[id$=-contributor]").querySelector("[selected]").textContent;
```
or just
```suggestion
            const name = row.querySelector("select[id$=-contributor] > [selected]").textContent;
```
should work instead

janno42:
@PFischbeck Do you want to continue working on this or should we take over?

PFischbeck:
@janno42 Feel free to take over. I thought I would have time to continue work on this, but I don't have it right now. Thanks!

Kakadus (approved):
did not found any bugs :+1:

evap/evaluation/templates/evap_evaluation_edit_js.html
```diff
@@ -1,32 +1,32 @@
...
+            const idChanged = row.querySelector("input[id$=-id]")?.value;
 
-            questionnaireChanged = false;
-            $(row.find("input[name$=-questionnaires]")).each(function () {
-                if (this.checked)
+            var questionnaireChanged = false;
+            row.querySelectorAll("input[name$=-questionnaires]")?.forEach(el => {
+                if (el.checked)
                     questionnaireChanged = true;
             })
```
richardebeling, line 15: Would
```suggestion
            const questionnaireChanged = row.querySelectorAll("input[name$=-questionnaires]:checked") !== null;
```
do the job?

richardebeling (approved):

evap/evaluation/templates/evap_evaluation_edit_js.html
```diff
@@ -1,32 +1,32 @@
...
-    <script type="text/javascript">
-        rowChanged = function(row) {
-            name = $(row.find("select[id$=-contributor]")).find(":selected").text();
-            nameChanged = name && name != "---------";
-            idChanged = row.find("input[id$=-id]").val();
+    <script type="module">
+        import { makeFormSortable } from "{% static 'js/sortable-form.js' %}";
+        const rowChanged = function(row) {
+            const name = row.querySelector("select[id$=-contributor] > [selected]")?.textContent;
+            const nameChanged = name && name != "---------";
```
Kakadus, line 8:
```suggestion
            const nameChanged = name && name !== "---------";
```

evap/evaluation/templates/evap_evaluation_edit_js.html
Kakadus: is `!== null` correct or do we need `.any()`? (possibly `?` before?)

richardebeling: Ah, yeah, I may have tested locally with `querySelector` instead of `querySelectorAll`

evap/static/ts/src/sortable-form.ts
```diff
@@ -0,0 +1,109 @@
...
+                el.querySelectorAll("option[selected]").forEach(option => {
+                    option.removeAttribute("selected");
+                });
+                const option = el.querySelector("option");
+                if (option) {
+                    option.selected = true;
+                }
+            });
+
+            //Check the first item in every button group
```
niklasmohrin, line 79: I would expect our formatter to add a space after `//` here :thinking: Will try to look at it later today

Kakadus (approved):
@niklasmohrin friendly ping :)

richardebeling (approved):

niklasmohrin (approved):

PR #2206, opened 2024-06-03
Fix incorrect evaluation operation error message

janno42:


niklasmohrin:
"cannot"?

evap/staff/views.py
```diff
@@ -327,8 +327,8 @@ def applicable_to(evaluation):
     @staticmethod
     def warning_for_inapplicables(amount):
         return ngettext(
-            "{} evaluation can not be reverted, because it already started. It was removed from the selection.",
-            "{} evaluations can not be reverted, because they already started. They were removed from the selection.",
+            "{} evaluation can not be reverted, because it was already approved. It was removed from the selection.",
+            "{} evaluations can not be reverted, because they were already approved. They were removed from the selection.",
```
niklasmohrin, line 331:
```suggestion
            "{} evaluation cannot be reverted, because it was already approved. It was removed from the selection.",
            "{} evaluations cannot be reverted, because they were already approved. They were removed from the selection.",
```

Kakadus (approved):

PR #2208, opened 2024-06-03
Allow list of values for graded column in importer

janno42:
Currently only one value can be defined in the settings. Multiple values are used on production, the importer should accept all of them.

### DIFF
--- a/evap/staff/importers/enrollment.py
+++ b/evap/staff/importers/enrollment.py
@@ -145,9 +145,9 @@ class IsGradedImportMapper:
     @classmethod
     def is_graded_from_import_string(cls, is_graded: str) -> bool:
         is_graded = is_graded.strip()
-        if is_graded == settings.IMPORTER_GRADED_YES:
+        if is_graded in settings.IMPORTER_GRADED_YES:
             return True
-        if is_graded == settings.IMPORTER_GRADED_NO:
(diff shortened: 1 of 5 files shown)


evap/staff/importers/enrollment.py
```diff
@@ -145,9 +145,9 @@ def __init__(self, *args, invalid_is_graded: str, **kwargs):
     @classmethod
     def is_graded_from_import_string(cls, is_graded: str) -> bool:
         is_graded = is_graded.strip()
-        if is_graded == settings.IMPORTER_GRADED_YES:
+        if is_graded in settings.IMPORTER_GRADED_YES:
             return True
-        if is_graded == settings.IMPORTER_GRADED_NO:
+        if is_graded in settings.IMPORTER_GRADED_NO:
```
richardebeling, line 150: I believe there should be a test for, probably testing with "yes" and "no" and some invalid value. Maybe use any of the non-first options in there (so we have that covered)?

richardebeling (approved):

richardebeling (approved):

Kakadus (approved):

niklasmohrin (approved):
Wow, this was an easy change, it almost seems as if good engineering pays of :thinking:

PR #2211, opened 2024-06-03
Improve navigation to textanswer review page

janno42:
use button instead of progress bar to navigate to textanswer review page

### DIFF
--- a/evap/staff/templates/staff_semester_view_evaluation.html
+++ b/evap/staff/templates/staff_semester_view_evaluation.html
@@ -129,7 +129,7 @@
                     {% include 'progress_bar.html' with done=evaluation.num_voters total=evaluation.num_participants icon='user' warning=True %}
                 </span>
             {% endif %}
-            {% if evaluation.num_textanswers > 0 and state != evaluation.State.PUBLISHED %}
+            {% if evaluation.num_textanswers > 0 and state < evaluation.State.PUBLISHED %}
                 <span data-bs-toggle="tooltip" data-bs-placement="left" title="{% translate 'Number of text answers' %}">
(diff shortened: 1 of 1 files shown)


richardebeling (approved):

Kakadus (approved):

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -129,11 +129,7 @@
                     {% include 'progress_bar.html' with done=evaluation.num_voters total=evaluation.num_participants icon='user' warning=True %}
                 </span>
             {% endif %}
-            {% if evaluation.num_textanswers > 0 and state != evaluation.State.PUBLISHED and not info_only %}
-                <a class="no-underline" href="{% url 'staff:evaluation_textanswers' evaluation.id %}" data-bs-toggle="tooltip" data-bs-placement="left" title="{% translate 'Review text answers' %}">
-                    {% include 'progress_bar.html' with done=evaluation.num_reviewed_textanswers total=evaluation.num_textanswers icon='comment' %}
-                </a>
-            {% elif info_only %}
```
niklasmohrin, line 136: Before, we would always display the progress bar with `info_only`, now we only display if the `num_textanswers` condition is true (i.e., reviewing is possible). Is that intended?

I suppose it makes sense, because the numbers are not correct anymore once reviewing is not possible anymore, right?

niklasmohrin (approved):

evap/staff/templates/staff_semester_view_evaluation.html
janno42: `info_only` is only used to show evaluations on the evaluation operation confirmation page. we don't need a progress bar there.
But I found another thing while thinking about it and changed the button's color when the text answers are already reviewed.

PR #2222, opened 2024-06-10
replace last jquery with plain browser api

### LINKED ISSUE #2217: Questionnaire reordering fails
When reordering questionnaires on the questionnaire index page, a `403` error is returned. This happens because the CSRF token is missing: `Forbidden (CSRF token missing.): /staff/questionnaire/update_indices`

Kakadus:
close #2217

### DIFF
--- a/evap/static/ts/src/datagrid.ts
+++ b/evap/static/ts/src/datagrid.ts
@@ -1,4 +1,4 @@
-import {CSRF_HEADERS} from "../../../static_collected/ts/src/csrf-utils";
+import { CSRF_HEADERS } from "./csrf-utils.js";
 
 declare const Sortable: typeof import("sortablejs");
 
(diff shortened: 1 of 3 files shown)


niklasmohrin:
Huh? Wouldn't the real fix be to replace the `$.post` call with a `fetch`? Or does this occur in more places? As far as I can tell, sortablejs does not use jquery itself and it should be all our doing

richardebeling (approved):

evap/static/ts/src/datagrid.ts
```diff
@@ -341,11 +343,19 @@ export class QuestionnaireGrid extends TableGrid {
             draggable: ".sortable",
             scrollSensitivity: 70,
             onUpdate: event => {
-                if (event.oldIndex && event.newIndex) {
+                if (event.oldIndex !== undefined && event.newIndex !== undefined) {
```
Kakadus, line 346: since migration to typescript: if indices were 0, this did not work

janno42 (approved):
✅ UI works

Thanks for fixing the bug with ordering the first item as well 🎆

richardebeling (approved):

Issue #2189, opened 2024-05-13
Assertion failure due to weak signal receivers having no guarantee of being garbage collected in time


richardebeling:
See discussion in https://github.com/e-valuation/EvaP/issues/1361.

Our two weak notification receivers (grep `weak=True`) were intended to only be active while the current request is being handled. However, we can not assume this, as garbage collection is not guaranteed. We've had at least two sporadic test failures in the last weeks due to this. To replicate, run the tests with a patched `manage.py` where you added
```python3
import gc
gc.disable()
```
to the top and run the tests with `--shuffle=3290553936` (any shuffle where the `user_edit` view is accessed at least once before the `evaluation_edit` view test that removes participants and triggers reward point granting.

I don't know yet how to properly implement the behavior we intended to have. If may suffice to unregister the signal receiver at the end of the view.

Edit: While we're at it, I think the `grant_reward_points_after_delete` function could use a `assert not grantings` in the `if reverse` in the `if granting` case -- we only expect `grantings` to be set once, otherwise, overwriting of the old value of `grantings` would be unexpected.

Kakadus:
same thing with seed `1116753744` at `test_remove_participants_proportional_reward_points` [here](https://github.com/e-valuation/EvaP/actions/runs/9555687802/job/26339427235?pr=2225)

Kakadus:
we discarded the approach in #2229 because it seemed implementation specific. Instead, we want a contextmanager which disconnects the signal directly.

PR #2229, opened 2024-06-17
manually del weakref functions to make sure they are disconnected

Kakadus:
fix https://github.com/e-valuation/EvaP/issues/2189

tested by reverting commit, adding `gc.disable()` and testing with `--shuffle=1116753744`

### DIFF
--- a/evap/staff/views.py
+++ b/evap/staff/views.py
@@ -1323,6 +1323,7 @@ def helper_evaluation_edit(request, evaluation):
         "editable": editable,
         "questionnaires_with_answers_per_contributor": questionnaires_with_answers_per_contributor,
     }
+    del notify_reward_points  # cleanup receiver
     return render(request, "staff_evaluation_form.html", template_data)
 
 
@@ -2228,7 +2229,7 @@ def user_edit(request, user_id):
         for message in form.remove_messages:
             messages.warning(request, message)
         return redirect("staff:user_index")
-
+    del notify_reward_points  # cleanup receiver
     return render(
         request,
         "staff_user_form.html",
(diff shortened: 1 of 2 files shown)


evap/staff/views.py
```diff
@@ -1323,6 +1323,7 @@ def notify_reward_points(grantings, **_kwargs):
         "editable": editable,
         "questionnaires_with_answers_per_contributor": questionnaires_with_answers_per_contributor,
     }
+    del notify_reward_points  # cleanup receiver
```
richardebeling, line 1326: Why does this solve the problem? From [the docs](https://docs.python.org/3/reference/simple_stmts.html#grammar-token-python-grammar-del_stmt):
> Deletion of a name removes the binding of that name from the local or global namespace, depending on whether the name occurs in a [global](https://docs.python.org/3/reference/simple_stmts.html#global) statement in the same code block.

To me, this sounds like `del x` and `x` going out of scope are the same thing. Both don't ensure that garbage collection actually runs. The tests pass, but that's probably just a CPython implementation detail?

There is [`Signal.disconnect`](https://docs.djangoproject.com/en/5.0/topics/signals/#disconnecting-signals). Ideally I'd want to use that through a context manager. I can't find any matching context manager in the django docs, we'd probably have to make our own, but that should be feasible.

Kakadus: my explaination: python uses both gc and ref counting. Maybe going out of scope does not mean directly that the refcount is decreased.
I tested this entirely without gc and test failures were avoidee

Kakadus: according to [this stackoverflow](https://stackoverflow.com/a/9063805) you are right: refcounting _is_ part of "gc" and is  even implementation dependent :+1:

PR #1931, opened 2023-04-24
Tooltips on disabled buttons in navbar switch

tanja04:
closes #1820

PR #2232, opened 2024-06-24
Customizable evaluation results filters

### LINKED ISSUE #1786: Customizable evaluation results filters
Currently, an evaluation's results page can be viewed as "Public" (shows results that are visible for everyone logged in), "Own" (shows all results that are visible for the current user), and "Export" (shows all results for general questions and the results for the current user).

This should be changed to two customizable filters which can be selected via switch button groups (two button groups placed below each other in the top right of the page):
- General results (only visible if current user can see text answers for general questions (reviewers, responsibles of this evaluation, contributors in this evaluation where `TextAnswerVisibility` is set to `GENERAL_TEXTANSWERS`, and delegates of such responsibles and contributors))
    - Show text answers
    - Hide text answers
- Contributor results (only visible if current user is reviewer, contributor in this evaluation, or delegate of such a contributor)
    - Show text answers
    - Hide text answers
    - Only my results (only visible if current user is contributor in this evaluation and for staff users who opened this view via the contributor's profile like in the current "Export" view)

jooooosef:
fix #1786


evap/results/templates/results_evaluation_detail.html
```diff
@@ -32,68 +32,81 @@
...
+                                    data-bs-placement="bottom"
+                                    title="{% translate 'Shows all general ratings available for you (without text answers)' %}"
+                                >
+                                    {% translate 'Ratings' %}
+                                </a>
+                            </div>
+                        </div>
+                    </div>
+                </div>
+                <div class="row mt-3">
```
janno42, line 68:
```suggestion
                <div class="row mt-1">
```

evap/results/tools.py
```diff
@@ -471,49 +471,52 @@ def textanswers_visible_to(contribution):
...
+                if textanswer.contribution.evaluation.contributions.filter(
+                    contributor__in=represented_users,
+                    textanswer_visibility=Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS,
+                ).exists():
+                    return True
+
+                # the people responsible for a course can see all general text answers for all its evaluations
+                if textanswer.contribution.is_general and any(
+                    user in represented_users
+                    for user in textanswer.contribution.evaluation.course.responsibles.all()  # includes self
```
janno42, line 504:
```suggestion
                    user in represented_users  # includes self
                    for user in textanswer.contribution.evaluation.course.responsibles.all()
```
did you mean it like this?

evap/results/views.py
```diff
@@ -167,23 +167,33 @@ def evaluation_detail(request, semester_id, evaluation_id):
     semester = get_object_or_404(Semester, id=semester_id)
     evaluation = get_object_or_404(semester.evaluations, id=evaluation_id, course__semester=semester)
 
-    view, view_as_user, represented_users, contributor_id = evaluation_detail_parse_get_parameters(request, evaluation)
+    # hier view_general added
```
janno42, line 170: remaining comment

evap/results/views.py
```diff
@@ -167,23 +167,33 @@ def evaluation_detail(request, semester_id, evaluation_id):
...
+        view_general_text,
+        view_contributor_results,
+        view_as_user,
+        represented_users,
+        contributor_id,
+    ) = evaluation_detail_parse_get_parameters(request, evaluation)
 
     evaluation_result = get_results(evaluation)
-    remove_textanswers_that_the_user_must_not_see(evaluation_result, view_as_user, represented_users, view)
+    # hier view_general added
```
janno42, line 180: remaining comment

evap/results/views.py
```diff
@@ -167,23 +167,33 @@ def evaluation_detail(request, semester_id, evaluation_id):
...
-        evaluation_result, view_as_user, view
+        evaluation_result, view_as_user, view_contributor_results
     )
 
     course_evaluations = get_evaluations_of_course(evaluation.course, request)
     course_evaluations.sort(key=lambda evaluation: evaluation.name)
 
     contributors_with_omitted_results = []
-    if view == "export":
+    if view_contributor_results == "personal":  # changed
```
janno42, line 196: remaining comment

evap/results/views.py
```diff
@@ -210,30 +220,40 @@ def evaluation_detail(request, semester_id, evaluation_id):
...
+        "view_general_text": view_general_text,
         "view_as_user": view_as_user,
         "contributors_with_omitted_results": contributors_with_omitted_results,
         "contributor_id": contributor_id,
     }
     return render(request, "results_evaluation_detail.html", template_data)
 
 
-def remove_textanswers_that_the_user_must_not_see(evaluation_result, user, represented_users, view):
+# hier view_general added
```
janno42, line 235: remaining comment

evap/results/views.py
```diff
@@ -210,30 +220,40 @@ def evaluation_detail(request, semester_id, evaluation_id):
...
+    evaluation_result, user, represented_users, view_general_text, view_contributor_results
+):
     for questionnaire_result in evaluation_result.questionnaire_results:
         for question_result in questionnaire_result.question_results:
             if isinstance(question_result, TextResult):
                 question_result.answers = [
                     answer
                     for answer in question_result.answers
-                    if can_textanswer_be_seen_by(user, represented_users, answer, view)
+                    # hier view_general added
```
janno42, line 245: remaining comment

evap/results/views.py
```diff
@@ -289,19 +309,24 @@ def remove_empty_questionnaire_and_contribution_results(evaluation_result):
...
     top_results = []
     bottom_results = []
     contributor_results = []
 
-    for contribution_result in evaluation_result.contribution_results:
+    for (
+        contribution_result
+    ) in (
+        evaluation_result.contribution_results
+    ):  # sort nach contributions_results, sodass jede Antwort ein Contributor hat, ist dieser None, gibts keinen
```
janno42, line 321: remaining comment

evap/results/views.py
```diff
@@ -376,38 +401,58 @@ def evaluation_detail_parse_get_parameters(request, evaluation):
     if not evaluation.can_results_page_be_seen_by(request.user):
         raise PermissionDenied
 
-    view = request.GET.get("view", "public" if request.user.is_reviewer else "full")
-    if view not in ["public", "full", "export"]:
-        view = "public"
+    view_general_text = request.GET.get(
+        "view_general_text",
+        "full" if request.user.is_reviewer or request.user.is_responsible_or_contributor_or_delegate else "ratings",
```
janno42, line 406: `request.user.is_responsible_or_contributor_or_delegate` is a general attribute of the account and not related to the current evaluation (you would need `evaluation.is_user_responsible_or_contributor_or_delegate(request.user)` for that). Could we just use `"full"` without conditions here because users who are not responsible, contributor, or delegate won't see the text answers and the choice buttons anyway?

@richardebeling asking for a second pair of eyes here

evap/results/views.py
```diff
@@ -376,38 +401,58 @@ def evaluation_detail_parse_get_parameters(request, evaluation):
...
+    )
+    if view_general_text not in ["full", "ratings"]:
+        if request.user.is_reviewer or request.user.is_responsible_or_contributor_or_delegate:
+            view_general_text = "full"
+        else:
+            view_general_text = "ratings"
```
janno42, line 412: same here

evap/results/views.py
```diff
@@ -376,38 +401,58 @@ def evaluation_detail_parse_get_parameters(request, evaluation):
...
+
+    view_contributor_results = request.GET.get(
+        "view_contributor_results",
+        "full" if request.user.is_reviewer or request.user.is_responsible_or_contributor_or_delegate else "ratings",
+    )
+    if view_contributor_results not in ["full", "ratings", "personal"]:
+        if request.user.is_reviewer or request.user.is_responsible_or_contributor_or_delegate:
+            view_contributor_results = "full"
+        else:
+            view_contributor_results = "ratings"
```
janno42, line 422: same here

janno42 (changes requested):
- looks good, most parts work as expected
- tests not reviewed yet

evap/results/views.py
richardebeling: Yes, sounds reasonable to me, and shouldn't have security implications.

evap/results/views.py
```diff
@@ -376,38 +399,51 @@ def evaluation_detail_parse_get_parameters(request, evaluation):
...
     represented_users = [view_as_user]
-    if view != "export":
-        represented_users += list(view_as_user.represented_users.all())
-    # redirect to non-public view if there is none because the results have not been published
-    if not evaluation.can_publish_rating_results and view == "public":
-        view = "full"
+    represented_users += list(view_as_user.represented_users.all())
 
-    return view, view_as_user, represented_users, contributor_id
+    return view_general_results, view_contributor_results, view_as_user, represented_users, contributor_id
```
richardebeling, line 429: Can we use strong typing (optimally statically checkable) for these? (-> have an enum representing the possible values for `view_general_results`, `view_contributor_results`, instead of using strings)

evap/results/tests/test_views.py
```diff
@@ -468,23 +468,38 @@ def test_heading_question_filtering(self):
...
+            self.url + "?view_contributor_results=ratings", user=self.manager
+        )
+        self.assertEqual(page_with_ratings_general_get_parameter.context["view_general_results"], "full")
+        self.assertEqual(page_with_ratings_general_get_parameter.context["view_contributor_results"], "ratings")
 
-        page_with_full_get_parameter = self.app.get(f"{self.url}?view=full", user=self.manager)
-        self.assertEqual(page_with_full_get_parameter.context["view"], "full")
+        # digga das doch qustsch unter default view test ihr atzen
+        # page_with_full_general_get_parameter = self.app.get(self.url + "?view_general_results=full", user=self.manager)
+        # self.assertEqual(page_with_full_general_get_parameter.context["view_general_results"], "full")
```
richardebeling, line 494: leftover comments

evap/results/tests/test_views.py
```diff
@@ -468,23 +468,38 @@ def test_heading_question_filtering(self):
...
 
-        page_with_export_get_parameter = self.app.get(f"{self.url}?view=export", user=self.manager)
-        self.assertEqual(page_with_export_get_parameter.context["view"], "export")
+        page_with_random_get_parameter = self.app.get(
+            self.url + "?view_general_results=josefwarhier&view_contributor_results=yannikwarhier", user=self.manager
+        )
+        self.assertEqual(page_with_random_get_parameter.context["view_general_results"], "full")
+        self.assertEqual(page_with_random_get_parameter.context["view_contributor_results"], "full")
+
+        # noch für den anderen param? auch von view_contributor_results
```
richardebeling, line 502: The existing checks suffice, you can remove the comment.

evap/results/tests/test_views.py
```diff
@@ -567,24 +582,6 @@ def test_invalid_contributor_id(self):
         self.app.get(self.url + "?contributor_id=asd", user=self.manager, status=400)
         self.app.get(self.url + "?contributor_id=1234", user=self.manager, status=404)
 
-    def test_evaluation_sorting(self):
```
richardebeling, line 570: Why was this removed?

evap/results/tests/test_views.py
```diff
@@ -715,7 +712,9 @@ def test_private_evaluation(self):
         self.app.get(url, user=student_external, status=200)
 
 
-class TestResultsTextanswerVisibilityForManager(WebTestStaffMode):
+class TestResultsTextanswerVisibilityForManager(
+    WebTestStaffMode
+):  # hier vielleicht nochmal schauen wegen diesem with root: ???
```
richardebeling, line 717: leftover comment

evap/results/tests/test_views.py
```diff
@@ -772,120 +771,212 @@ def test_textanswer_visibility_for_manager(self):
...
+class TestResultsTextanswerVisibilityForResponsible(WebTest):
+    fixtures = ["minimal_test_data_results"]
+
+    @classmethod
+    def setUpTestData(cls):
+        cache_results(Evaluation.objects.get(id=1))
+
+    def helper_static(self, page):
+
+        self.assertNotIn(".general_orig_deleted.", page)
```
richardebeling, line 830: Instead of defining one function per test case, I'd propose you simply store a list of all text answers. In your test, for one page, you then pass a list of expected visible text answers to a shared helper function. The helper asserts that the expected answers are in the page, and that all answers that are not expected are not in the page. That way, you should reduce the duplication of the `helper_static` method, the `for` loops, and the repeated logic of calling the "static" and the "dynamic" helper

richardebeling: (In the end, a test could probably look like:

```helper
my_test_helper("full", "full", [".some_textanswer", ".some_other_textanswer"])
my_test_helper("full", "personal", [".some_textanswer", ".some_other_textanswer"])
...
```
)

evap/results/tests/test_views.py
```diff
@@ -772,120 +771,212 @@ def test_textanswer_visibility_for_manager(self):
         self.assertNotIn(".responsible_contributor_additional_orig_hidden.", page)
 
 
-class TestResultsTextanswerVisibility(WebTest):
+# NOTE antworten für die zu testende person in eine liste packen, weil die je nach filter immer die gleichen assertions haben(?)
+# test
+class TestResultsTextanswerVisibilityForStudent(WebTest):
```
richardebeling, line 776: If possible, it would be nice (and I think it's feasible) to get back to one test class for the whole (django) view

richardebeling (changes requested):

evap/results/tests/test_views.py
ybrnr: that was an accident I added it again

evap/results/tests/test_views.py
```diff
@@ -732,216 +731,199 @@ def test_private_evaluation(self):
...
-        self.assertIn(".responsible_contributor_additional_orig_published.", page)
-        self.assertNotIn(".responsible_contributor_additional_orig_hidden.", page)
+    def helper_test_general(self, user, view_general_results, textanswers_in):
 
+        textanswers_not_in = list(set(self.general_textanswers) - set(textanswers_in))
+        for con_view in self.contributor_views:
+            page = self.app.get(
+                f"/results/semester/1/evaluation/1?view_general_results={view_general_results.value}&view_contributor_results={con_view.value}",
+                user=user,
+            )
```
janno42, line 774:
```suggestion
        for contributor_view in self.contributor_views:
            page = self.app.get(
                f"/results/semester/1/evaluation/1?view_general_results={view_general_results.value}&view_contributor_results={contributor_view.value}",
                user=user,
            )
```

evap/results/tests/test_views.py
```diff
@@ -732,216 +731,199 @@ def test_private_evaluation(self):
...
+        for gen_view in self.general_views:
+            page = self.app.get(
+                f"/results/semester/1/evaluation/1?view_contributor_results={view_contributor_results.value}&view_general_results={gen_view.value}",
+                user=user,
+            )
+
+            for answer in textanswers_in:
+                self.assertIn(answer, page)
+            for answer in textanswers_not_in:
+                self.assertNotIn(answer, page)
```
janno42, line 793:
```suggestion
        for general_view in self.general_views:
            page = self.app.get(
                f"/results/semester/1/evaluation/1?view_contributor_results={view_contributor_results.value}&view_general_results={general_view.value}",
                user=user,
            )

            for answer in textanswers_in:
                self.assertIn(answer, page)
            for answer in textanswers_not_in:
                self.assertNotIn(answer, page)
```

evap/results/templates/results_evaluation_detail.html
```diff
@@ -32,68 +32,74 @@
...
-                                        {% translate 'Shows results available for everyone logged in.' %}"
-                                        >
-                                        {% translate 'Public' %}
-                                    {% endif %}
-                            </a>
-                        {% endif %}
+                    <div class="row">
+                        <div class="col-auto">
+                            <div class="btn-switch btn-switch-light my-auto d-print-none">
+                                <div class="btn-switch-label">{% translate 'General results' %}</div>
```
janno42, line 39: when disabled, "Ratings" should be selected/active and its tooltip should be shown on hover.
a different tooltip should be shown for "All" when disabled: "You can't see text answers for general questions in this evaluation."

evap/results/templates/results_evaluation_detail.html
```diff
@@ -32,68 +32,74 @@
...
+                                    </a>
+                                </div>
+                            </div>
+                        </div>
+                    </div>
+                
+                <div class="row mt-1">
+                    <div class="col-auto">
+                        <div class="btn-switch btn-switch-light my-auto d-print-none">
+                            <div class="btn-switch-label">{% translate 'Contributor results' %}</div>
```
janno42, line 69: when disabled, "Ratings" should be selected/active and its tooltip should be shown on hover.
a different tooltip should be shown for the other buttons when disabled: "You can't see text answers for contributor questions because you aren't a contributor."

evap/results/templates/results_evaluation_detail.html
```diff
@@ -21,7 +21,7 @@
     {% if evaluation.state != evaluation.State.PUBLISHED %}
         <div class="alert alert-warning">{% translate 'This is a preview. The results have not been published yet.' %}</div>
     {% endif %}
-
+    <!--manager is automatically a reviewer (so this are all groups that can see the buttons)-->
```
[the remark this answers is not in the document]
Kakadus: maybe also fix the english
```suggestion
    {# manager is automatically a reviewer (so these are all groups that can see the buttons) #}
```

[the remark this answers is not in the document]
jooooosef: I no english.

evap/results/tests/test_views.py
```diff
@@ -732,216 +731,199 @@ def test_private_evaluation(self):
...
+
+    def test_manager(self):
+        with run_in_staff_mode(self):  # in staff mode, the manager can see everything
+            user = "manager@institution.example.com"
+
+            self.helper_test_general(
+                user,
```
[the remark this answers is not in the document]
ybrnr: do you mean we should initialize all users that get tested (e.g. manager, student, contributor...) in the setUpTestData() method?
So that we reference them by self.manager, self.student...

evap/results/tools.py
```diff
@@ -473,49 +493,54 @@ def textanswers_visible_to(contribution):
...
-        if contributor in represented_users:
-            return True
-        # users can see text answers from general contributions if one of their represented users has text answer
-        # visibility GENERAL_TEXTANSWERS for the evaluation
-        if (
-            textanswer.contribution.is_general
-            and textanswer.contribution.evaluation.contributions.filter(
+    if textanswer.contribution.is_general:
+        if view_general_results == ViewGeneralResults.RATINGS:
+            return False
```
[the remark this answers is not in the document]
jooooosef: Yes, we think so too. We removed it and tests are still working fine.

evap/results/tests/test_views.py
```diff
@@ -732,216 +731,199 @@ def test_private_evaluation(self):
...
+        self.helper_test_general(user, ViewGeneralResults.RATINGS, [])
+
+        self.helper_test_contributor(
+            user,
+            ViewContributorResults.FULL,
+            [
+                ".responsible_contributor_orig_published.",
+                ".responsible_contributor_changed_published.",
+                ".responsible_contributor_orig_private.",
+                ".responsible_contributor_additional_orig_published.",
```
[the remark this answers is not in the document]
jooooosef: For the 4 we introduced a new variable (contributor_textanswers_responsible_contributor_can_see) inside the method and use it.
We did this as well for contributor_textanswers_contributor_can_see, as this was used twice as well.
For the other 3 we introduced an instance variable (standard_general_textanswers), since this is used in multiple methods. 

Is this fine now?

evap/results/tests/test_views.py
```diff
@@ -732,216 +731,199 @@ def test_private_evaluation(self):
...
-        self.assertIn(".contributor_orig_private.", page)
-        self.assertIn(".responsible_contributor_orig_published.", page)
-        self.assertNotIn(".responsible_contributor_orig_hidden.", page)
-        self.assertNotIn(".responsible_contributor_orig_published_changed.", page)
-        self.assertIn(".responsible_contributor_changed_published.", page)
-        self.assertIn(".responsible_contributor_orig_private.", page)
-        self.assertNotIn(".responsible_contributor_orig_notreviewed.", page)
-        self.assertIn(".responsible_contributor_additional_orig_published.", page)
-        self.assertNotIn(".responsible_contributor_additional_orig_hidden.", page)
+    def helper_test_general(self, user, view_general_results, textanswers_in):
```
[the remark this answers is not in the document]
ybrnr: I think that is a good idea, I changed the variable names for both helper methods to `expected_visible_textanswers`

evap/results/views.py
```diff
@@ -377,38 +431,57 @@ def evaluation_detail_parse_get_parameters(request, evaluation):
...
+    if view_contributor_results == ViewContributorResults.PERSONAL and request.user.is_staff:
         view_as_user = contributor
     contributor_id = contributor.pk if contributor != request.user else None
 
     represented_users = [view_as_user]
-    if view != "export":
-        represented_users += list(view_as_user.represented_users.all())
-    # redirect to non-public view if there is none because the results have not been published
-    if not evaluation.can_publish_rating_results and view == "public":
-        view = "full"
```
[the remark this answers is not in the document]
jooooosef: Previously there was a view called "public" which now would be Genereral textanswers on ratings and contributor textanswers on ratings. But even in a preview, there are textanswers, so a user might want to hide these as well. So we dont want to force the user into a view which is hiding all textanswers.

evap/staff/templates/staff_user_form.html
```diff
@@ -101,7 +101,7 @@ <h5 class="card-title me-auto">{% translate 'Export evaluation results' %}</h5>
                                     {% for evaluation in semester_evaluations.list %}
                                         <li>
                                             {% if evaluation|can_results_page_be_seen_by:form.instance %}
-                                                <a href="{% url 'results:evaluation_detail' semester_evaluations.grouper.id evaluation.id %}?view=export&contributor_id={{ form.instance.id }}">
+                                                <a href="{% url 'results:evaluation_detail' semester_evaluations.grouper.id evaluation.id %}?view_contributor_results=personal&contributor_id={{ form.instance.id }}">
                                                     {{ evaluation.full_name }}
                                                 </a>
```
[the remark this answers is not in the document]
jooooosef: Seemed to be accidental. We reverted.

evap/results/views.py
```diff
@@ -199,6 +209,30 @@ def evaluation_detail(request, semester_id, evaluation_id):
...
+
+    if evaluation.is_user_contributor(view_as_user):
+        contributor_personal = True
+    for user in represented_users:
+        if user in evaluation.course.responsibles.all():
+            general_textanswers = True
+        if evaluation.is_user_contributor(user):
+            if Contribution.objects.filter(
+                contributor=user,
+                evaluation=evaluation,
```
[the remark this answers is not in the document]
jooooosef: I though the first "evaluation" is from the ORM (so a column in the database) and the second "evaluation" (after the =) is the defined at the beginning with `evaluation = get_object_or_404(semester.evaluations, id=evaluation_id, course__semester=semester)` and then in the if statement a couple lines earlier with `evaluation = get_evaluations_with_course_result_attributes(prefetched)[0]`. Is this how it works?

evap/results/templates/results_evaluation_detail.html
jooooosef: We added that ratings is selected when the button group is disabled.
Showing the tooltip is not this easily possible since the buttons are disabled and therefore dont accept any mouse events (inlcuding mouse hover), thus the title from the django template is not shown.

evap/results/tests/test_views.py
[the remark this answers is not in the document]
richardebeling: Yes, that was my intention. I just saw that they're part of the `minimal_test_data_results` fixture, when I wrote the comment I guess I thought that we create them in test setup anyways. Not that important, then, I think. Resolving.

evap/results/tests/test_views.py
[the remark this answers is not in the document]
richardebeling: Yup, looking good, thanks! :)

evap/results/views.py
```diff
@@ -199,6 +209,30 @@ def evaluation_detail(request, semester_id, evaluation_id):
 
     is_responsible_or_contributor_or_delegate = evaluation.is_user_responsible_or_contributor_or_delegate(view_as_user)
 
+    general_textanswers = False
```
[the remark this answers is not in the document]
richardebeling: Ok. I'm afraid that we're duplicating access control logic here, even if it is just for adding a html disabled attribute, it is very likely that we will foget to update this view if we ever want to change the access control logic in `can_textanswer_be_seen_by`. Optimally, both would reuse the same decision logic.

Maybe it helps if we make this more declarative and use more precise naming, although that doesn't fully resolve the underlying issue.

Would something like this work? (Helper is made up, would need to be added)
```
buttons_enabled = {
    "general_full": view_as_user.is_reviewer or represents_or_is_contributor_or_responsible(view_as_user, evaluation)
    "general_ratings": # ...
    "contributor_full": # ...
    "contributor_ratings": # ...
    "contributor_personal": # ...
}
```

Why is `contributor_personal = True` if `view_as_user` is a contributor, independently of what the contribution's `textanswer_visibility` is, but in the loop for the represented users, it is only set to true if the contribution has `TextAnswerVisibility.GENERAL_TEXTANSWERS`? It seems to me that both would need to be filtered for the contribution's text answer visibility.

Maybe, the helper we want here is `def has_textanswer_visibility(user: UserProfile, evaluation: Evaluation, visibility: TextAnswerVisibility)`, which would answer the question
> Is this user allowed to see text answers with this visibility classification (by being a reviewer, a contributor, a responsible, or by being a representative of a reviewer, contributor or responsible

(There might be better names) (Optimally, `can_textanswer_be_seen_by` would then also use that.)

@niklasmohrin any ideas on how to solve this nicely, here?

evap/results/templates/results_evaluation_detail.html
richardebeling: I think we found a way of having tooltips on disabled buttons in https://github.com/e-valuation/EvaP/pull/1931.

evap/results/views.py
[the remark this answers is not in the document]
richardebeling: You mean "which is showing all textanswers"? Hmm, ok.

@janno42 can you confirm that we no longer want to have this "redirect" here? I see that it made sense in the old system, I'm not sure how exactly we want to handle it in the new system.

evap/results/views.py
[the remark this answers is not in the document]
niklasmohrin: I agree that some refactoring is necessary. I think that the biggest problem is conflating "am I allowed to see this" and "do I want to see this" (both right now and before). The vocabulary developed along the way also seems a little off, both `ViewGeneralResults` and `ViewContributorResults` are possible "views" in the frontend, but this shouldn't play a role in something as backend as "can textanswer be seen by".

Once we split the filtering into "can see" and "want to see", we can think about what the model for "can see" should be. My proposed solution is the following:

We want to establish some access rights for the user before, and then check every answer against these access rights. A first draft could be:
```python
class AnswerAccess:
    def general(self) -> bool:
        return False

    def for_contributor(self, contributor: UserProfile) -> bool:
        return False


@dataclass
class ContributorAccess(AnswerAccess):
    user: UserProfile

    def for_contributor(self, contributor: UserProfile) -> bool:
        return self.user == contributor or contributor in self.user.represented_users.all()


class GeneralAccess(AnswerAccess):
    def general(self) -> bool:
        return True


class FullAccess(GeneralAccess):
    def for_contributor(self, contributor: UserProfile) -> bool:
        return True
```
The exact implementation of `AnswerAccess` (inheritance, or a class with flags, or something else) is not so important, only the idea that the access rights should be boiled down to some description before we actually look at the answers that exist. I will in fact suggest a modification of the above later on.

Then the view for the particular evaluation could be (in a wonderland where all these methods exist):
```python
textanswers = evaluation.all_textanswers()

textanswer_access = evaluation.textanswer_access_of(request.user)
textanswers = remove_inaccessible(textanswers, textanswer_access)

view_filters = parse_view_filters(request) # combination of ViewGeneralResults and ViewContributorResults
textanswers = filter_for_view(textanswers, view_filters)
```

The `parse_view_filters` method should return some `AnswerFilter` object which represents the choices made at the "view" buttons on the website. These filters have two dimensions: General results and contributor results, matching the button groups.

The values taken by these dimensions should have a subset relation, for example for general results: "none" <= "all", and for contributor results "none" <= "{niklas}" <= "{niklas, richard}" <= "all". We can also extend the relation on the individual dimension to whole answer filters so that a filter is a subset of another iff the subset relation holds in all dimensions (general and contributor).

Using this, we can generate two answer filters for each request:
1. The access filter, which only retains answers that the user is allowed to view. This could, for example, happen by having `AnswerAccess.to_filter` or so. Another possibility, which could be cleaner, would be to skip the whole `AnswerAccess` step altogether and just have `evaluation.access_filter_for(user)`. It kind of depends on how the details work out in the end.
2. The view filter, which is generated from the chosen options in the button groups.

Notably, both filters are constructed in the context of a particular evaluation, but do not refer to this evaluation afterwards. Your role in the evaluation should determine what kind of filter you get.

The UI should then display all buttons that are included by the access filter as enabled, and all others as disabled.

Conversely, the view should check that the view filter is a subset of the access filter. (After this check then don't need to apply the access filter to the answers anymore, because the view filter will take of it).

The logic could be nicely shared with `results.tools.textanswers_visible_to`, it would just be `{user for user in some_set_of_relevant_users if contribution.evaluation.textanswer_access_of(user).to_filter().retains(contribution.answer)}`.

Now, to be fair, this is more complicated to implement, but I think it's a nice abstraction that fits our requirements. What do you think @richardebeling ?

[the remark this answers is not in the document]
niklasmohrin: Another possible implementation (using `str` instead of `UserProfile`):
```python
from functools import total_ordering
from enum import IntEnum
from dataclasses import dataclass


class GeneralAnswerFilter(IntEnum):
    NONE = 0
    FULL = 1


@dataclass(order=True)
class PersonalAnswerFilter:
    @total_ordering
    @dataclass
    class Everyone:
        def __le__(self, other) -> bool:
            if isinstance(other, PersonalAnswerFilter.Everyone):
                return True
            if isinstance(other, set):
                return False
            return NotImplemented

    contributors: set[str] | Everyone


@dataclass
class AnswerFilter:
    general: GeneralAnswerFilter
    personal: PersonalAnswerFilter

    # Not using total_ordering, because the ordering defined here it not total

    def __eq__(self, other) -> bool:
        if not isinstance(other, AnswerFilter):
            return NotImplemented
        return self.general == other.general and self.personal == other.personal

    def __le__(self, other) -> bool:
        if not isinstance(other, AnswerFilter):
            return NotImplemented
        return self.general <= other.general and self.personal <= other.personal

    def __lt__(self, other) -> bool:
        return self <= other and self != other


assert GeneralAnswerFilter.NONE < GeneralAnswerFilter.FULL
assert (
    PersonalAnswerFilter(contributors=set())
    == PersonalAnswerFilter(contributors=set())
    < PersonalAnswerFilter(contributors={"niklas"})
    == PersonalAnswerFilter(contributors={"niklas"})
    < PersonalAnswerFilter(contributors={"niklas", "richard"})
    < PersonalAnswerFilter(contributors=PersonalAnswerFilter.Everyone())
    == PersonalAnswerFilter(contributors=PersonalAnswerFilter.Everyone())
)

none_everyone = AnswerFilter(GeneralAnswerFilter.NONE, PersonalAnswerFilter(PersonalAnswerFilter.Everyone()))
full_everyone = AnswerFilter(GeneralAnswerFilter.FULL, PersonalAnswerFilter(PersonalAnswerFilter.Everyone()))
full_noone = AnswerFilter(GeneralAnswerFilter.FULL, PersonalAnswerFilter(set()))

assert none_everyone < full_everyone
assert full_noone < full_everyone
assert not none_everyone <= full_noone
assert not none_everyone >= full_noone
```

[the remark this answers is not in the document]
richardebeling: @jooooosef @ybrnr given that this will be a non-trivial refactoring of access control, we'll accept the duplicated code here for now after it has been corrected (see my "Why is contributor_personal = True [...]" remark in https://github.com/e-valuation/EvaP/pull/2232#discussion_r1893912606). To improve this afterwards, please open an issue referencing this discussion. If you want to work on that, feel free to self-assign that issue.

evap/results/views.py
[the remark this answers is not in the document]
janno42: I'd say it doesn't really matter. The default can be the first option in the button groups and we don't have to redirect if results have not been published.

evap/results/views.py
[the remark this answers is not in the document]
Kakadus: basically, yes. 
>  the first "evaluation" is from the ORM (so a column in the database)

it's a parameter to the `filter` function which corresponds to the column.

With `Contribution.objects.filter(..., evaluation=evaluation, ...)` you will get exactly the same objects as using `evaluation.contributions.filter(...)`

evap/results/views.py
```diff
@@ -199,6 +209,30 @@ def evaluation_detail(request, semester_id, evaluation_id):
...
+        if evaluation.is_user_contributor(user):
+            if Contribution.objects.filter(
+                contributor=user,
+                evaluation=evaluation,
+                textanswer_visibility=Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS,
+            ).exists():
+                general_textanswers = True
+                contributor_textanswers = True
+            else:
+                contributor_textanswers = True
```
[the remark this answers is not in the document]
Kakadus:
```suggestion
            contributor_textanswers = True
```

evap/results/tools.py
```diff
@@ -473,49 +493,48 @@ def textanswers_visible_to(contribution):
...
-        ):
+            ).exists():
+                return True
+
+            # the people responsible for a course can see all general text answers for all its evaluations
+            if any(
+                user in represented_users  # includes self
+                for user in textanswer.contribution.evaluation.course.responsibles.all()
+            ):
+                return True
```
Kakadus, line 525: these checks could be combined, it readability doesn't suffer too much.

the last check could be performed in the database using `textanswer.contribution.evaluation.course.responsibles.filter(pk__in=...).exists()`

Kakadus (commented):
some last comments from me :)

evap/results/templates/results_evaluation_detail.html
jooooosef: nope, the buttons there are now enabled.
As of @janno42 idea, the tooltips from the individual buttons were removed and a "big tooltip" added on the label of the btn group.

evap/results/tests/test_views.py
jooooosef: Should be readded now. Please confirm.

evap/results/templates/results_evaluation_detail.html
jooooosef: should be implemented with `class="btn btn-sm btn-light {% if not general_textanswers %} disabled {% endif %} {% if view_general_results == ViewGeneralResults.RATINGS or not general_textanswers %} active{% endif %}"` no? @janno42

janno42 (dismissed):
This nice, want have.
(After changing these little things)

jooooosef:
I just realised that when I use the contributor and go in a contribution where I dont have general_textanswers rights but can edit change the contributor settings that the value for the view_general_results parameter is not set correctly. 
Example:
user: contributor@institution.example.com
contribution: Operating Systems I
click on ratings 

Why is that?

evap/results/views.py
```diff
@@ -199,6 +209,27 @@ def evaluation_detail(request, semester_id, evaluation_id):
...
+    general_textanswers = False
+    contributor_textanswers = False
+    contributor_personal = False
+
+    if view_as_user.is_reviewer:
+        general_textanswers = True
+        contributor_textanswers = True
+
+    if evaluation.is_user_contributor(view_as_user):
+        contributor_personal = True
```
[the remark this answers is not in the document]
niklasmohrin: and for the third one:
```python
    contributor_textanswers = view_as_user.is_reviewer or evaluation.user_is_contributor(view_as_user)
    contributor_personal = evaluation.is_user_contributor(view_as_user)

    user_represents_responsible = any(user in represented_users for user in evaluation.course.responsibles.all())
    user_represents_general_visibilty_contributor = any(
        evaluation.is_user_contributor(user)
        and evaluation.contributions.filter(
            contributor=user,
            textanswer_visibility=Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS,
        ).exists()
        for user in represented_users
    )

    general_textanswers = (
        view_as_user.is_reviewer or user_represents_responsible or user_represents_general_visibilty_contributor
    )
```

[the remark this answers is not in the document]
niklasmohrin: @richardebeling ?

niklasmohrin (changes requested):
Looks mostly good

evap/results/tools.py
```diff
@@ -473,49 +493,47 @@ def textanswers_visible_to(contribution):
...
-            user in represented_users for user in textanswer.contribution.evaluation.course.responsibles.all()
-        ):
+            ).exists():
+                return True
+
+            # the people responsible for a course can see all general text answers for all its evaluations
+            if textanswer.contribution.evaluation.course.responsibles.filter(
+                pk__in=(user.pk for user in represented_users)
+            ).exists():
+                return True
```
Kakadus, line 524: 1. invert the first condition and return `False` early to reduce indentation
2. maybe combine the checks 
```py
        return (
            user.is_reviewer
            or textanswer.contribution.evaluation.contributions.filter(
                contributor__in=represented_users,
                textanswer_visibility=Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS,
            ).exists()  #  represented user can see the textanswer
            or textanswer.contribution.evaluation.course.responsibles.filter(
                pk__in=(user.pk for user in represented_users)
            ).exists()  # responsible people for a course can see all general text answers for all its evaluations
        )
```
instead?

evap/results/views.py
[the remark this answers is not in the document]
jooooosef: @richardebeling ?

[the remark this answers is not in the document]
richardebeling: @jooooosef?

evap/results/tools.py
```diff
@@ -473,49 +493,47 @@ def textanswers_visible_to(contribution):
...
-            return False
-        if not textanswer.contribution.is_general and contributor != user:
-            return False
-    elif user.is_reviewer:
-        return True
-
-    if textanswer.is_private:
-        return contributor == user
-
     # NOTE: when changing this behavior, make sure all changes are also reflected in results.tools.textanswers_visible_to
```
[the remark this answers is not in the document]
jooooosef: Semantically we should not have to change anything in `textanswers_visible_to` since it calculates the users who should be able to see a textanswer. Since we only change the filters for a user (not changing that they should not see a textanswer) we should not have to do anything here. I think?

evap/results/tests/test_views.py
```diff
@@ -732,216 +747,183 @@ def test_private_evaluation(self):
...
-        self.assertIn(".contributor_orig_private.", page)
-        self.assertIn(".responsible_contributor_orig_published.", page)
-        self.assertNotIn(".responsible_contributor_orig_hidden.", page)
-        self.assertNotIn(".responsible_contributor_orig_published_changed.", page)
-        self.assertIn(".responsible_contributor_changed_published.", page)
-        self.assertIn(".responsible_contributor_orig_private.", page)
-        self.assertNotIn(".responsible_contributor_orig_notreviewed.", page)
-        self.assertIn(".responsible_contributor_additional_orig_published.", page)
-        self.assertNotIn(".responsible_contributor_additional_orig_hidden.", page)
+    def helper_test_general(self, user, view_general_results, expected_visible_textanswers):
```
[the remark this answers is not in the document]
ybrnr: i added the check_view_with function and replaced the helper functions for general and contributor. It makes the code a lot simpler. Thanks :)

evap/results/tests/test_views.py
```diff
@@ -732,216 +747,183 @@ def test_private_evaluation(self):
...
+                    ".responsible_contributor_orig_published.",
+                    ".responsible_contributor_changed_published.",
+                    ".responsible_contributor_orig_private.",
+                    ".responsible_contributor_additional_orig_published.",
+                ],
+            )
+            self.helper_test_contributor(user, ViewContributorResults.RATINGS, [])
+            self.helper_test_contributor(user, ViewContributorResults.PERSONAL, [])
+
+        user = "manager@institution.example.com"
```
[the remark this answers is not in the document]
ybrnr: I changed it to `user = self.manager`, since we have a class wide instance of a manager.

evap/results/views.py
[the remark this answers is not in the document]
jooooosef: I think I did that with 
`contributor_textanswers = view_as_user.is_reviewer or any(
        evaluation.is_user_contributor(user) for user in represented_users
    )`

evap/results/tools.py
jooooosef: I am unsure what you mean with "invert the first condition and return False early to reduce indentation". Which first condition? 

I did use your suggested code for the rest since it does the same thing but is more clean.

richardebeling: The [error message](https://app.codacy.com/gh/e-valuation/EvaP/pull-requests/2232/issues#issue-c241af76bc96cfff94ff32405cfa93ee) says that inline comments should start with "# " = "<hash><space>". Your comment has two spaces _after_ the hash, that triggers the error.

jooooosef: Ohhh thanks.

Kakadus (approved):

evap/results/templates/results_evaluation_detail.html
niklasmohrin, line 1: I wonder if we should try to make the columns align to make it look nicer, what do you think @janno42 ?

This is the current site:
image

niklasmohrin (approved):
Thanks, very cool!

evap/results/templates/results_evaluation_detail.html
janno42: We discussed that and both alignments have their disadvantages. I originally also thought right-aligned might look better, so yes, let's move them :)

niklasmohrin: Ah, but do you mean to keep the width of all elements the same, but move the top row to right? I thought that we would make the right border of "general results" and "contributor results" align, and then have the two "ratings and texts" buttons and the two "ratings" buttons be vertically aligned (with no button over "only mine").

In the end, I don't really care about which version we use

Kakadus: As I understood @niklasmohrin (and what I would think may look better), is that the colums align on top of each other, not that both are right-aligned.

evap/results/tests/test_views.py
```diff
@@ -732,216 +749,126 @@ def test_private_evaluation(self):
...
-    fixtures = ["minimal_test_data_results"]
+            for answer in expected_visible_textanswers:
+                self.assertIn(answer, page)
+            for answer in textanswers_not_in:
+                self.assertNotIn(answer, page)
+
+    def test_manager(self):
+        user = self.manager
+        self.check_with_view(user, [])
+        with run_in_staff_mode(self):  # in staff mode, the manager can see everything except deleted or changed
```
richardebeling, line 811: comment doesn't match assertion below, we assert that everything from ` self.standard_general_textanswers` is visible, which includes `self.standard_general_textanswers`.

I would, without checking more deeply, expect that noone can ever see deleted or the originals of changed answers? In that case, I would also try to generalize that out, so that the logic here doesn't have to list any answers at all, but literally just says: "As a manager, I can see all comments that could possibly be shown to any user".

evap/results/tests/test_views.py
```diff
@@ -732,216 +749,126 @@ def test_private_evaluation(self):
...
-        self.assertIn(".responsible_contributor_additional_orig_published.", page)
-        self.assertNotIn(".responsible_contributor_additional_orig_hidden.", page)
-
+    def check_with_view(
+        self,
+        user,
+        expected_visible_textanswers,
+        general=ViewGeneralResults,
+        contributor=ViewContributorResults,
+        all_textanswers=general_textanswers + contributor_textanswers,
```
richardebeling, line 794: Quick Ctrl+F says `all_textanswers` is never passed?

evap/results/views.py
```diff
@@ -199,6 +209,25 @@ def evaluation_detail(request, semester_id, evaluation_id):
...
+
+    user_represents_responsible = any(user in represented_users for user in evaluation.course.responsibles.all())
+    user_represents_general_visibilty_contributor = any(
+        evaluation.is_user_contributor(user)
+        and evaluation.contributions.filter(
+            contributor=user,
+            textanswer_visibility=Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS,
+        ).exists()
+        for user in represented_users
+    )
```
richardebeling, line 225: Semantically correct and nice code, _but_: We also need to worry about performance here. `is_user_contributor` hits the database once, the `.exists()` query hits the database once, all of this happens `n` times for `n` represented users. Can we maybe express this as a single queryset instead?

(same thing applies to the other queries above as well)

richardebeling (commented):
looking really good! I have a few more comments, but we can also address those in a follow-up PR if you prefer :)

evap/results/views.py
```diff
@@ -209,20 +209,20 @@ def evaluation_detail(request, semester_id, evaluation_id):
...
-            contributor=user,
-            textanswer_visibility=Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS,
-        ).exists()
-        for user in represented_users
-    )
+    user_represents_responsible = evaluation.course.responsibles.all().filter(
+        pk__in=(user.pk for user in represented_users)
+    ).exists()
+    
+    #any(user in represented_users for user in evaluation.course.responsibles.all())
```
richardebeling, line 222: commented code

evap/results/templates/results_evaluation_detail.html
```diff
@@ -32,68 +31,62 @@
...
+                    <div class="col-auto ms-auto">
+                        <div class="btn-switch btn-switch-light my-auto d-print-none">
+                            <div class="btn-switch-label">
+                                {% translate 'General results' %}
+                            </div>
+                            <div class="btn-switch btn-group">
+                                <a
+                                    href="{% url 'results:evaluation_detail' evaluation.course.semester.id evaluation.id %}?view_general_results=full&view_contributor_results={{ view_contributor_results.value }}"
+                                    role="button"
+                                    class="btn btn-sm btn-light {% if not general_textanswers %} disabled {% elif view_general_results == ViewGeneralResults.FULL %} active {% endif %}"
```
richardebeling, line 45: Typically we try to generate HTML output that doesn't contain superfluous spaces (which this currently would do). I'm not super obsessed about this, but if you touch the file again, maybe fix that (as you do below with `disabled active` and `active`

richardebeling (approved):
Thank you so much for pulling through with this. I much appreciate the new tests 👍

Kakadus (approved):

niklasmohrin (dismissed):

ybrnr (approved):

richardebeling (approved):
approved for after @janno42's comments are fixed

janno42 (approved):
<insert filter emoji here>

PR #2234, opened 2024-06-28
Add participant progress bar

### LINKED ISSUE #2233: Participation progress bar
For the active semester, a progress bar for the total participation value (how many voters submitted evaluations) should be displayed. It should show the percentage of total voters divided by the total number of participants in that semester. The bar should be divided into several steps, each of which is labeled.

- A settings variable defines the steps. Each step has a percentage value and a label (e.g., `0.4: "1 000 €"`, `0.8: "5 000 €"`).
- The highest step defines at which value the bar is filled (in the example above, a participation value of `0.8` would be represented by a completely filled bar.
- The progress bar can be turned off by setting the variable to an empty value.
- Two settings variables `EXCLUDED_COURSE_TYPE_IDS` and `EXCLUDED_EVALUATION_IDS` define which evaluations are not taken into account when calculating the participation value.
- Evaluations are only taken into account if they _do not meet any_ of these conditions:
    - They are excluded by their ID
    - They have an excluded CourseType
    - They are a `single_result`
    - They are in a state lower than `APPROVED`
    - They are `not rewarded`
    - They are `private`

The progress bar should be displayed at the top of the student index page.
An explanatory text that is defined in a settings variable (HTML formatting should be supported) should be displayed in a (initially collapsed) text area below the progress bar.
Also, the text "The last evaluation was submitted N minutes/hours/days ago" should be displayed next to it.

Kakadus:
close #2233

Open things:
- [x] collapsible text
- [x] last evaluation time (NEXT to text)
- [x] translations
- [x] refactor (esp. to avoid `else: ` case)
- [x] tests

### DIFF
--- a/deployment/localsettings.template.py
+++ b/deployment/localsettings.template.py
@@ -1,3 +1,5 @@
+from django.utils.safestring import mark_safe
+
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.postgresql',   # postgresql', 'mysql', 'sqlite3' or 'oracle'.
@@ -15,3 +17,18 @@ SECRET_KEY = "${SECRET_KEY}"  # nosec
 
 # Make apache work when DEBUG == False
 ALLOWED_HOSTS = ["localhost", "127.0.0.1"]
+
+### Evaluation progress rewards
+GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[float, str]] = [
+    (0, "0€"),
+    (0.25, "1.000€"),
+    (0.6, "3.000€"),
+    (0.7, "7.000€"),
+    (0.9, "10.000€"),
+]
+GLOBAL_EVALUATION_PROGRESS_EXCLUDED_COURSE_TYPE_IDS: list[int] = []
+GLOBAL_EVALUATION_PROGRESS_EXCLUDED_EVALUATION_IDS: list[int] = []
+GLOBAL_EVALUATION_PROGRESS_INFO_TEXT = {
+    "de": mark_safe("Deine Teilnahme am Evaluationsprojekt wird helfen. Evaluiere also <b>jetzt</b>!"),
+    "en": mark_safe("Your participation in the evaluation helps, so evaluate <b>now</b>!"),
+}
--- a/deployment/localsettings.template.py
+++ b/deployment/localsettings.template.py
@@ -1,3 +1,5 @@
+from fractions import Fraction
+
 from django.utils.safestring import mark_safe
 
 DATABASES = {
@@ -19,12 +21,12 @@ SECRET_KEY = "${SECRET_KEY}"  # nosec
 ALLOWED_HOSTS = ["localhost", "127.0.0.1"]
 
 ### Evaluation progress rewards
-GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[float, str]] = [
-    (0, "0€"),
-    (0.25, "1.000€"),
-    (0.6, "3.000€"),
-    (0.7, "7.000€"),
-    (0.9, "10.000€"),
+GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[Fraction, str]] = [
+    (Fraction("0"), "0€"),
+    (Fraction("0.25"), "1.000€"),
+    (Fraction("0.6"), "3.000€"),
+    (Fraction("0.7"), "7.000€"),
+    (Fraction("0.9"), "10.000€"),
 ]
 GLOBAL_EVALUATION_PROGRESS_EXCLUDED_COURSE_TYPE_IDS: list[int] = []
 GLOBAL_EVALUATION_PROGRESS_EXCLUDED_EVALUATION_IDS: list[int] = []
--- a/evap/evaluation/templatetags/evaluation_filters.py
+++ b/evap/evaluation/templatetags/evaluation_filters.py
@@ -112,9 +112,7 @@ def percentage_one_decimal(fraction, population):
 def percentage_zero_on_error(fraction, population):
     try:
         return f"{int(float(fraction) / float(population) * 100):.0f}%"
-    except ValueError:
-        return "0%"
-    except ZeroDivisionError:
+    except (ZeroDivisionError, ValueError):
         return "0%"
 
 
--- a/evap/evaluation/templatetags/evaluation_filters.py
+++ b/evap/evaluation/templatetags/evaluation_filters.py
@@ -109,6 +109,16 @@ def percentage_one_decimal(fraction, population):
 
 
 @register.filter
+def percentage_zero_on_error(fraction, population):
+    try:
+        return f"{int(float(fraction) / float(population) * 100):.0f}%"
+    except ValueError:
+        return "0%"
+    except ZeroDivisionError:
+        return "0%"
+
+
+@register.filter
 def to_colors(question_result: RatingResult | None):
     if question_result is None:
         # When displaying the course distribution, there are no associated voting choices.
--- a/evap/settings.py
+++ b/evap/settings.py
@@ -354,15 +354,14 @@ MEDIA_ROOT = os.path.join(BASE_DIR, "upload")
 
 ### Evaluation progress rewards
 GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[float, str]] = [
-    (0, "0€"),
-    (0.25, "1.000€"),
-    (0.6, "3.000€"),
-    (0.7, "7.000€"),
-    (0.9, "10.000€"),
+    # (required_voter_ratio between 0 and 1, reward_text)
 ]
-EXCLUDED_COURSE_TYPE_IDS: list[int] = []
-EXCLUDED_EVALUATION_IDS: list[int] = []
-PARTICIPANT_PROGRESS_TEXT = "Deine Teilnahme am Evaluationsprojekt wird helfen. Evaluiere also <b>jetzt</b>!"
+GLOBAL_EVALUATION_PROGRESS_EXCLUDED_COURSE_TYPE_IDS: list[int] = []
+GLOBAL_EVALUATION_PROGRESS_EXCLUDED_EVALUATION_IDS: list[int] = []
+GLOBAL_EVALUATION_PROGRESS_INFO_TEXT = {
+    "de": "",
+    "en": "",
+}
 
 ### Slogans
 SLOGANS_DE = [
--- a/evap/settings.py
+++ b/evap/settings.py
@@ -353,15 +353,10 @@ STATIC_ROOT = os.path.join(BASE_DIR, "static_collected")
 MEDIA_ROOT = os.path.join(BASE_DIR, "upload")
 
 ### Evaluation progress rewards
-GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[float, str]] = [
-    # (required_voter_ratio between 0 and 1, reward_text)
-]
+GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[float, str]] = []  # (required_voter_ratio between 0 and 1, reward_text)
 GLOBAL_EVALUATION_PROGRESS_EXCLUDED_COURSE_TYPE_IDS: list[int] = []
 GLOBAL_EVALUATION_PROGRESS_EXCLUDED_EVALUATION_IDS: list[int] = []
-GLOBAL_EVALUATION_PROGRESS_INFO_TEXT = {
-    "de": "",
-    "en": "",
-}
+GLOBAL_EVALUATION_PROGRESS_INFO_TEXT: dict[str, str] = {}  # language-tag ("de") to info text, possibly marked safe
 
 ### Slogans
 SLOGANS_DE = [
--- a/evap/settings.py
+++ b/evap/settings.py
@@ -352,6 +352,17 @@ STATIC_ROOT = os.path.join(BASE_DIR, "static_collected")
 # Absolute filesystem path to the directory that will hold user-uploaded files.
 MEDIA_ROOT = os.path.join(BASE_DIR, "upload")
 
+### Evaluation progress rewards
+GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[float, str]] = [
+    (0, "0€"),
+    (0.25, "1.000€"),
+    (0.6, "3.000€"),
+    (0.7, "7.000€"),
+    (0.9, "10.000€"),
+]
+EXCLUDED_COURSE_TYPE_IDS: list[int] = []
+EXCLUDED_EVALUATION_IDS: list[int] = []
+PARTICIPANT_PROGRESS_TEXT = "Deine Teilnahme am Evaluationsprojekt wird helfen. Evaluiere also <b>jetzt</b>!"
 
 ### Slogans
 SLOGANS_DE = [
--- a/evap/settings.py
+++ b/evap/settings.py
@@ -354,7 +354,9 @@ STATIC_ROOT = os.path.join(BASE_DIR, "static_collected")
 MEDIA_ROOT = os.path.join(BASE_DIR, "upload")
 
 ### Evaluation progress rewards
-GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[Fraction, str]] = []  # (required_voter_ratio between 0 and 1, reward_text)
+GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[Fraction, str]] = (
+    []
+)  # (required_voter_ratio between 0 and 1, reward_text)
 GLOBAL_EVALUATION_PROGRESS_EXCLUDED_COURSE_TYPE_IDS: list[int] = []
 GLOBAL_EVALUATION_PROGRESS_EXCLUDED_EVALUATION_IDS: list[int] = []
 GLOBAL_EVALUATION_PROGRESS_INFO_TEXT: dict[str, str] = {"de": "", "en": ""}
--- a/evap/settings.py
+++ b/evap/settings.py
@@ -356,7 +356,7 @@ MEDIA_ROOT = os.path.join(BASE_DIR, "upload")
 GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[float, str]] = []  # (required_voter_ratio between 0 and 1, reward_text)
 GLOBAL_EVALUATION_PROGRESS_EXCLUDED_COURSE_TYPE_IDS: list[int] = []
 GLOBAL_EVALUATION_PROGRESS_EXCLUDED_EVALUATION_IDS: list[int] = []
-GLOBAL_EVALUATION_PROGRESS_INFO_TEXT: dict[str, str] = {}  # language-tag ("de") to info text, possibly marked safe
+GLOBAL_EVALUATION_PROGRESS_INFO_TEXT: dict[str, str] = {"de": "", "en": ""}
 
 ### Slogans
 SLOGANS_DE = [
--- a/evap/settings.py
+++ b/evap/settings.py
@@ -11,6 +11,7 @@ https://docs.djangoproject.com/en/2.0/ref/settings/
 import logging
 import os
 import sys
+from fractions import Fraction
 from typing import Any
 
 from django.contrib.staticfiles.storage import ManifestStaticFilesStorage
@@ -353,7 +354,7 @@ STATIC_ROOT = os.path.join(BASE_DIR, "static_collected")
 MEDIA_ROOT = os.path.join(BASE_DIR, "upload")
 
 ### Evaluation progress rewards
-GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[float, str]] = []  # (required_voter_ratio between 0 and 1, reward_text)
+GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[Fraction, str]] = []  # (required_voter_ratio between 0 and 1, reward_text)
 GLOBAL_EVALUATION_PROGRESS_EXCLUDED_COURSE_TYPE_IDS: list[int] = []
 GLOBAL_EVALUATION_PROGRESS_EXCLUDED_EVALUATION_IDS: list[int] = []
 GLOBAL_EVALUATION_PROGRESS_INFO_TEXT: dict[str, str] = {"de": "", "en": ""}
--- a/evap/static/scss/_mixins.scss
+++ b/evap/static/scss/_mixins.scss
@@ -1,7 +1,7 @@
 @import "../bootstrap/scss/mixins";
 
 @mixin bar-shadow($color) {
-    text-shadow: 0 0 4px $color, 0 0 4px $color, 0 0 2px $color;
+    text-shadow: 0 0 2px $color, 0 0 2px $color, 0 0 2px $color, 0 0 2px $color, 0 0 2px $color, 0 0 2px $color;
 }
 
 @mixin no-user-select {
--- a/evap/static/scss/components/_progress.scss
+++ b/evap/static/scss/components/_progress.scss
@@ -59,8 +59,7 @@
     position: relative;
     display: grid;
     padding-top: .5rem;
-    padding-right: 1.5rem;
-    padding-left: 1.5rem;
+    padding-inline: 1.5rem;
 
     .progress-container {
         display: flex;
--- a/evap/static/scss/components/_progress.scss
+++ b/evap/static/scss/components/_progress.scss
@@ -54,3 +54,9 @@
 .multi-progress-bar .progress-container {
     padding-bottom: 3px;
 }
+
+.progress-step {
+    width: 1px;
+    height: 8px;
+    border: $dark-gray 1px solid;
+}
--- a/evap/static/scss/components/_progress.scss
+++ b/evap/static/scss/components/_progress.scss
@@ -55,8 +55,31 @@
     padding-bottom: 3px;
 }
 
-.progress-step {
-    width: 1px;
-    height: 8px;
-    border: $dark-gray 1px solid;
+.global-rewards-progress-bar {
+    position: relative;
+    display: grid;
+    padding-top: .5rem;
+    padding-right: 1.5rem;
+    padding-left: 1.5rem;
+
+    .progress-container {
+        display: flex;
+        height: 25px;
+        grid-column-start: 1;
+        grid-row-start: 1;
+    }
 }
+
+.progress-step {
+    transform: translateX(-50%);
+    width: fit-content;
+    grid-column-start: 1;
+    grid-row-start: 2;
+
+    .seperator {
+        width: 1px;
+        height: 8px;
+        border: $dark-gray 1px solid;
+    }
+}
+
--- a/evap/student/templates/student_global_reward.html
+++ b/evap/student/templates/student_global_reward.html
@@ -25,16 +25,14 @@
                         </span>
                     </span>
                 {% for reward in global_rewards.rewards_with_progress %}
-                    <div class="position-relative "
+                    <div class="position-relative"
                          style="left: {% widthratio reward.progress 1 100 %}%; transform: translateX(-50%); width: fit-content;grid-column-start: 1; grid-row-start: 2">
-                        <div style="width: fit-content">
-                            <div class="mx-auto progress-step"></div>
-                            <div class="small text-center">
-                                {% widthratio reward.vote_ratio 1 100 as percent %}
-                                {% blocktranslate %}{{ percent }}%{% endblocktranslate %}
-                            </div>
-                            <div class="text-lg-center fw-semibold">{{ reward.text }}</div>
+                        <div class="mx-auto progress-step"></div>
+                        <div class="small text-center">
+                            {% widthratio reward.vote_ratio 1 100 as percent %}
+                            {% blocktranslate %}{{ percent }}%{% endblocktranslate %}
                         </div>
+                        <div class="text-lg-center fw-semibold">{{ reward.text }}</div>
                     </div>
                 {% endfor %}
             </div>
--- a/evap/student/templates/student_global_reward.html
+++ b/evap/student/templates/student_global_reward.html
@@ -29,8 +29,7 @@
                     <div class="position-relative progress-step left-percent-{% widthratio reward.progress 1 100 %}">
                         <div class="mx-auto seperator"></div>
                         <div class="small text-center">
-                            {% widthratio reward.vote_ratio 1 100 as percent %}
-                            {% blocktranslate %}{{ percent }}%{% endblocktranslate %}
+                            {% widthratio reward.vote_ratio 1 100 %}%
                         </div>
                         <div class="text-lg-center fw-semibold">{{ reward.text }}</div>
                     </div>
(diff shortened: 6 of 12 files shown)


evap/settings.py
```diff
@@ -352,6 +352,17 @@ class ManifestStaticFilesStorageWithJsReplacement(ManifestStaticFilesStorage):
...
+### Evaluation progress rewards
+GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[float, str]] = [
+    (0, "0€"),
+    (0.25, "1.000€"),
+    (0.6, "3.000€"),
+    (0.7, "7.000€"),
+    (0.9, "10.000€"),
+]
+EXCLUDED_COURSE_TYPE_IDS: list[int] = []
+EXCLUDED_EVALUATION_IDS: list[int] = []
```
richardebeling, line 364: On its own, this name does not convey the context. Maybe it would make sense to nest related settings at some point. For now, I'd like to add some kind of context prefix to these. I guess it's not a big problem if they get verbose, so maybe simply `GLOBAL_EVALUATION_PROGRESS_EXCUDED_COURSE_TYPE_IDS` and `GLOBAL_EVALUATION_PROGRESS_EXCUDED_EVALUATION_IDS`?

evap/settings.py
```diff
@@ -352,6 +352,17 @@ class ManifestStaticFilesStorageWithJsReplacement(ManifestStaticFilesStorage):
...
+PARTICIPANT_PROGRESS_TEXT = "Deine Teilnahme am Evaluationsprojekt wird helfen. Evaluiere also <b>jetzt</b>!"
```
richardebeling, line 365: I'd want to mark this string as safe here, because here, the reader can reason about the contents. Once its passed through the view into the template, the line `{{ progress_description | safe }}` is hard for me to verify.

```suggestion
PARTICIPANT_PROGRESS_TEXT = mark_safe("Deine Teilnahme am Evaluationsprojekt wird helfen. Evaluiere also <b>jetzt</b>!")
```

(also: translation)

evap/student/views.py
```diff
@@ -105,12 +105,51 @@ def sorter(evaluation):
...
+        Semester.active_semester()
+        .evaluations.filter(is_single_result=False)
+        .exclude(state__lt=Evaluation.State.APPROVED)
+        .exclude(is_rewarded=False)
+        .exclude(id__in=settings.EXCLUDED_EVALUATION_IDS)
+        .exclude(course__type__id__in=settings.EXCLUDED_COURSE_TYPE_IDS)
+        .exclude(course__is_private=True)
+    ).aggregate(
+        total_votes=Sum("num_voters"),
+        participants=Sum("num_participants"),
```
richardebeling, line 118: "total_" vs no prefix is inconsistent
```suggestion
        total_votes=Sum("num_voters"),
        total_participants=Sum("num_participants"),
```

evap/student/views.py
```diff
@@ -105,12 +105,51 @@ def sorter(evaluation):
         )
 
     unfinished_evaluations.sort(key=sorter)
+    info = Evaluation.annotate_with_participant_and_voter_counts(
```
richardebeling, line 108: Can probably be unpacked by using `values_list` or something, so we immediatle have two meaningfully named local variables

evap/student/views.py
```diff
@@ -105,12 +105,51 @@ def sorter(evaluation):
...
+        .exclude(is_rewarded=False)
+        .exclude(id__in=settings.EXCLUDED_EVALUATION_IDS)
+        .exclude(course__type__id__in=settings.EXCLUDED_COURSE_TYPE_IDS)
+        .exclude(course__is_private=True)
+    ).aggregate(
+        total_votes=Sum("num_voters"),
+        participants=Sum("num_participants"),
+    )
+    if settings.GLOBAL_EVALUATION_PROGRESS_REWARDS:
+        max_reward = max(settings.GLOBAL_EVALUATION_PROGRESS_REWARDS)
```
richardebeling, line 121: Maybe
```suggestion
        max_reward_participation_quote, max_reward_value = max(settings.GLOBAL_EVALUATION_PROGRESS_REWARDS)
```

Or maybe we should use a dataclass / namedtuple thing in the settings, so we can have a `reward` instance that has `.participation_ratio` and `.value` or `.text` or something. The disadvantage of unpacking here is that the order of the definition matters and switching the order breaks things in weird ways.

evap/student/views.py
```diff
@@ -105,12 +105,51 @@ def sorter(evaluation):
...
         "semester_list": semester_list,
         "can_download_grades": request.user.can_download_grades,
         "unfinished_evaluations": unfinished_evaluations,
         "evaluation_end_warning_period": settings.EVALUATION_END_WARNING_PERIOD,
+        "num_total_votes": info["total_votes"],
+        "final_goal": final_goal,
+        "global_progress_rewards": positioned_labels,
+        "remaining_votes": next_goal - info["total_votes"],
+        "next_reward": next_reward,
+        "progress_description": settings.PARTICIPANT_PROGRESS_TEXT,
```
richardebeling, line 152: can we pack them all up in a dataclass / struct-y thing that is simply passed as an instance or `None` to the template? I don't like `info = defaultdict(lambda: 0)` above, just so we can have `info["total_votes"]` here.

evap/student/views.py
```diff
@@ -105,12 +105,51 @@ def sorter(evaluation):
...
+    if settings.GLOBAL_EVALUATION_PROGRESS_REWARDS:
+        max_reward = max(settings.GLOBAL_EVALUATION_PROGRESS_REWARDS)
+        final_goal = round(info["participants"] * max_reward[0])
+        next_goal = final_goal
+        next_reward = max_reward[1]
+        for percentage, reward in settings.GLOBAL_EVALUATION_PROGRESS_REWARDS:
+            candidate = info["participants"] * percentage
+            if info["total_votes"] < candidate < next_goal:
+                next_goal = round(candidate)
+                next_reward = reward
```
richardebeling, line 129: I think this would benefit from a bit more functional approach, something like
```python3
next_reward = min(
    reward for reward in settings.GLOBAL_EVALUATION_PROGRESS_REWARDS
    if current_participation_ratio < reward.required_participation_ratio
)
```

evap/student/templates/student_index.html
```diff
@@ -16,7 +16,40 @@
             </div>
         </div>
     {% endif %}
+{% if global_progress_rewards %}
+    <div class="card mb-3">
+        <div class="card-header">Nächstes Spendenziel: {{ next_reward }}</div>
```
richardebeling, line 21: blocktranslate all the things

evap/student/templates/student_index.html
```diff
@@ -16,7 +16,40 @@
...
+        <div class="card-body">
+            <div>
+                noch {{ remaining_votes }} Evaluierungen benötigt
+            </div>
 
+            <div class="position-relative d-grid py-2 px-4">
+            <span class="progress-container d-flex" style="height: 25px; grid-column-start: 1; grid-row-start: 1">
+                <span class="progress m-0 bg-info">
+                    <span class="progress-bar progress-bar-striped progress-bar-animated bg-success width-percent-{% widthratio num_total_votes final_goal 100 %}"></span>
```
richardebeling, line 30: @niklasmohrin `with-percent-x`, such a banger class :D

evap/student/templates/student_index.html
```diff
@@ -16,7 +16,40 @@
...
+                            <div class="text-center">|</div>
+                            <div class="text-lg-center fw-semibold">{{ tag }}</div>
+                            <div class="text-center">bei {% widthratio global_ratio 1 100 %}%</div>
+                        </div>
+                    </div>
+                {% endfor %}
+            </div>
+            {{ progress_description | safe }}. This should be collapsed by default.
+            <br>
+            The last evaluation was submitted N minutes/hours/days ago.
```
richardebeling, line 49: todos

evap/student/views.py
janno42: I'd just keep the simple version here. For me, adding a comment that the order is important is okay for now. In the long run, the settings will probably be replaced by proper database objects anyway.

evap/student/views.py
```diff
@@ -24,6 +28,79 @@
 SUCCESS_MAGIC_STRING = "vote submitted successfully"
 
 
+@dataclass
+class GlobalRewardWithProgress:
```
niklasmohrin, line 32: Maybe this class would be better placed under the `GlobalRewards` namespace, for example `GlobalRewards.Reward` or `GlobalRewards.Step`?

evap/student/tests/test_views.py
```diff
@@ -52,6 +53,88 @@ def test_num_queries_is_constant(self):
...
 
+    @override_settings(
+        GLOBAL_EVALUATION_PROGRESS_REWARDS=[(0.1, "a dog"), (0.5, "a quokka")],
+        GLOBAL_EVALUATION_PROGRESS_INFO_TEXT={"de": "info_text_str", "en": "info_text_str"},
+        GLOBAL_EVALUATION_PROGRESS_EXCLUDED_COURSE_TYPE_IDS=[1042],
+        GLOBAL_EVALUATION_PROGRESS_EXCLUDED_EVALUATION_IDS=[1043],
+    )
+    def test_global_reward_progress(self):
+        excluded_states = [state for state, __ in Evaluation.state.field.choices if state < Evaluation.State.APPROVED]
+        included_states = [state for state, __ in Evaluation.state.field.choices if state >= Evaluation.State.APPROVED]
```
niklasmohrin, line 64:
```suggestion
        excluded_states = [state for state in Evaluation.State if state < Evaluation.State.APPROVED]
        included_states = [state for state in Evaluation.State if state >= Evaluation.State.APPROVED]
```

evap/student/tests/test_views.py
```diff
@@ -52,6 +53,88 @@ def test_num_queries_is_constant(self):
...
+        baker.make(Evaluation, course__is_private=True, **baker_kwargs)
+        baker.make(Evaluation, id=1043, **baker_kwargs)
+        baker.make(Evaluation, course__type__id=1042, **baker_kwargs)
+        baker.make(Evaluation, _quantity=len(excluded_states), state=iter(excluded_states), **baker_base_kwargs)
+
+        # included
+        included_evaluations = [
+            *baker.make(Evaluation, _quantity=len(included_states), state=iter(included_states), **baker_base_kwargs),
+            baker.make(Evaluation, _voter_count=123, _participant_count=456, **baker_kwargs),
+        ]
```
niklasmohrin, line 82: Not sure how you feel about this, but we could also use `functools.partial` to get the "override particular attributes" behavior:
```suggestion
        make_evaluation = partial(
            baker.make,
            Evaluation,
            course__semester=self.semester,
            participants=users,
            voters=users[:10],
            state=Evaluation.State.APPROVED,
        )

        # excluded
        make_evaluation(is_rewarded=False)
        make_evaluation(is_single_result=True)
        make_evaluation(course__is_private=True)
        make_evaluation(id=1043)
        make_evaluation(course__type__id=1042)
        make_evaluation(_quantity=len(excluded_states), state=iter(excluded_states))

        # included
        included_evaluations = [
            *make_evaluation(_quantity=len(included_states), state=iter(included_states)),
            make_evaluation(_voter_count=123, _participant_count=456),
        ]
```

evap/student/templates/student_index.html
```diff
@@ -16,7 +16,56 @@
             </div>
         </div>
     {% endif %}
-
+    {% if global_rewards %}
```
niklasmohrin, line 19: Should we move this into its own file?

evap/student/templates/student_index.html
```diff
@@ -16,7 +16,56 @@
...
+                {% if global_rewards.next_reward_text %}
+                    {% blocktranslate with next_reward=global_rewards.next_reward_text %}Next reward: {{ next_reward }}{% endblocktranslate %}
+                {% else %}
+                    {% blocktranslate %}All rewards achieved{% endblocktranslate %}
+                {% endif %}
+            </div>
+            <div class="card-body">
+                <div>
+                    {% if global_rewards.next_reward_remaining_votes %}
+                        {% blocktranslate with remaining_votes=global_rewards.next_reward_remaining_votes %}{{ remaining_votes }} more evaluations required.{% endblocktranslate %}
```
niklasmohrin, line 31: Maybe "... until the next reward"?

evap/student/templates/student_index.html
```diff
@@ -16,7 +16,56 @@
...
+                                width-percent-{% widthratio global_rewards.bar_width_votes global_rewards.max_reward_votes 100 %}"></span>
+                            <span class="progress-bartext">
+                                {{ global_rewards.current_votes }}/{{ global_rewards.max_reward_votes }}
+                            </span>
+                        </span>
+                    </span>
+                    {% for reward in global_rewards.rewards_with_progress %}
+                        <div class="position-relative "
+                             style="left: {% widthratio reward.progress 1 100 %}%; transform: translateX(-50%); width: fit-content;grid-column-start: 1; grid-row-start: 2">
+                            <div style="width: fit-content">
```
niklasmohrin, line 52: I think this wrapper div can be removed without any change

evap/student/templates/student_index.html
```diff
@@ -16,7 +16,56 @@
...
+                    {% if global_rewards.next_reward_remaining_votes %}
+                        {% blocktranslate with remaining_votes=global_rewards.next_reward_remaining_votes %}{{ remaining_votes }} more evaluations required.{% endblocktranslate %}
+                    {% endif %}
+                    {% if global_rewards.last_vote_datetime %}
+                        {% blocktranslate trimmed with time_interval=global_rewards.last_vote_datetime|timesince %}
+                            The last evaluation was submitted {{time_interval}} ago.
+                        {% endblocktranslate %}
+                    {% endif %}
+                </div>
+                <div class="position-relative d-grid py-2 px-4">
```
niklasmohrin, line 39: I suppose the usage of grid makes sense to position everything relative to this container. Could we put the inline styles into one class `global-rewards-progress-bar` though? I would also go for generating `left-percent-xx` classes like we have `width-percent-xx` already

evap/student/templates/student_index.html
```diff
@@ -16,7 +16,56 @@
...
+                    </span>
+                    {% for reward in global_rewards.rewards_with_progress %}
+                        <div class="position-relative "
+                             style="left: {% widthratio reward.progress 1 100 %}%; transform: translateX(-50%); width: fit-content;grid-column-start: 1; grid-row-start: 2">
+                            <div style="width: fit-content">
+                                <div class="text-center">|</div>
+                                <div class="text-lg-center fw-semibold">{{ reward.text }}</div>
+                                <div class="text-center">
+                                    {% widthratio reward.vote_ratio 1 100 as percent %}
+                                    {% blocktranslate %}at {{percent}}%{% endblocktranslate %}
```
niklasmohrin, line 57:
```suggestion
                                    {% blocktranslate %}at {{ percent }}%{% endblocktranslate %}
```

evap/student/templates/student_index.html
```diff
@@ -16,7 +16,56 @@
...
+                        <span class="progress m-0 bg-info">
+                            <span class="progress-bar progress-bar-striped progress-bar-animated bg-success
+                                width-percent-{% widthratio global_rewards.bar_width_votes global_rewards.max_reward_votes 100 %}"></span>
+                            <span class="progress-bartext">
+                                {{ global_rewards.current_votes }}/{{ global_rewards.max_reward_votes }}
+                            </span>
+                        </span>
+                    </span>
+                    {% for reward in global_rewards.rewards_with_progress %}
+                        <div class="position-relative "
```
niklasmohrin, line 50: On a phone, the texts could very well overlap, not sure what an easy fix for this would be though - maybe the text could become smaller?

niklasmohrin (commented):
Looks good to me, mostly code style comments

Why does is the bar full when 90% are reached? Wouldn't it be nicer if the students could still strive for >90% (or the epic gaming move of going 100%), even if all rewards are reached?

deployment/localsettings.template.py
```diff
@@ -15,3 +17,18 @@
...
 # Make apache work when DEBUG == False
 ALLOWED_HOSTS = ["localhost", "127.0.0.1"]
+
+### Evaluation progress rewards
+GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[float, str]] = [
+    (0, "0€"),
+    (0.25, "1.000€"),
+    (0.6, "3.000€"),
+    (0.7, "7.000€"),
+    (0.9, "10.000€"),
```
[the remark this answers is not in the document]
richardebeling: I thought about using `Decimal` instead of float here. Since we always round/ceil, any decimal part inconsistency doesn't really matter, so I decided against it. I think the only thing we could run into is numers so big that they can not be represented by floats anymore, but I doubt we'll ever get close to 10^15 votes/participations. Am I missing something?

I'm still open to using `Decimal` or `Fraction`, just thought it would be less canonical with no advantage.

deployment/localsettings.template.py
```diff
@@ -15,3 +17,18 @@
...
+]
+GLOBAL_EVALUATION_PROGRESS_EXCLUDED_COURSE_TYPE_IDS: list[int] = []
+GLOBAL_EVALUATION_PROGRESS_EXCLUDED_EVALUATION_IDS: list[int] = []
+GLOBAL_EVALUATION_PROGRESS_INFO_TEXT = {
```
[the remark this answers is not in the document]
richardebeling: Thought about it, but didn't do it for 1. override_settings and 2. consistency with other "grouped" settings.

I am open to thinking about this as a more generic refactoring. We could likely build a customized `override_settings` that allows sub-setting modification? Didn't consider this in scope for this PR, though.

deployment/localsettings.template.py
```diff
@@ -15,3 +17,18 @@
 
 # Make apache work when DEBUG == False
 ALLOWED_HOSTS = ["localhost", "127.0.0.1"]
+
+### Evaluation progress rewards
+GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[float, str]] = [
+    (0, "0€"),
+    (0.25, "1.000€"),
```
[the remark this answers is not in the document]
richardebeling: We can have all the anomalies: Both counts might "magically" go up or down if Evaluations are added/removed from the pool of relevant evaluations.

With regard to real world use cases (e.g. adding missing participants, dropping wrongly assigned participants, adding a missing evaluation), I think it makes sense to just immediately reflect these changes. A frozen value would just be wrong. Yes, moving across reward milestones is possible, staff users will have to watch out for that.

evap/settings.py
```diff
@@ -352,6 +352,11 @@ class ManifestStaticFilesStorageWithJsReplacement(ManifestStaticFilesStorage):
 # Absolute filesystem path to the directory that will hold user-uploaded files.
 MEDIA_ROOT = os.path.join(BASE_DIR, "upload")
 
+### Evaluation progress rewards
+GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[float, str]] = []  # (required_voter_ratio between 0 and 1, reward_text)
```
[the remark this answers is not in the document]
richardebeling: Code only accesses it as pairs, we don't need key-lookup, so we don't need the hashing part of the dictionary. Fair, a dictionary cannot contain the same key twice, and we don't expect two rewards with the same ratio.

Both, list and dict, are fine with me. Preferences on your side?

evap/student/tests/test_views.py
richardebeling: Ah, nice, this is what I wanted to do, I just didn't consider `partial` :D

deployment/localsettings.template.py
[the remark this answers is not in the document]
niklasmohrin: I am only worrying about cases where we jump from one integer to the other, so for ceiling, cases where the float multiplication ends up at 500 vs 500.00001. I am not super sure whether any grave problems could actually occur, I think I would slightly prefer fractions just for the peace of mind though

evap/settings.py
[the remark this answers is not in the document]
niklasmohrin: Multiplicity is not too much of a modelling concern for me, my only real criticism is that the list carries an ordering that is never used - dict feels more natural to me in that regard

Im the end both are totally fine

deployment/localsettings.template.py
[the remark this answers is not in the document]
richardebeling: Fair. Problem example, for a possible test case:
```python3
>>> print(0.07 * 100, math.ceil(0.07 * 100))
7.000000000000001 8
```

deployment/localsettings.template.py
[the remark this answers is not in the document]
richardebeling: Created #2236

evap/settings.py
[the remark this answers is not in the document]
richardebeling: Sticking with list for now, simply because it is "the" default container. Yes it has ordering, but a dict has as well. We could switch to an unordered_set (don't know if python `set` keeps insertion order like `dict`?)

evap/student/views.py
richardebeling: I'm generally open for restructuring (and the static factory method makes sense to me), but I don't think `.Reward` or `.Step` are better names than `WithProgress`.

evap/student/views.py
```diff
@@ -24,6 +28,79 @@
...
+    max_reward_votes: int
+    bar_width_votes: int
+    next_reward_remaining_votes: int
+    next_reward_text: str | None
+    last_vote_datetime: datetime.datetime
+    rewards_with_progress: list[GlobalRewardWithProgress]
+    info_text: str
+
+
+def get_global_rewards() -> GlobalRewards | None:
```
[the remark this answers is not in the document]
Kakadus: for is not allowed as a function name. Also, I don't really see the benefit of moving it to a separate `staticmethod`. Maybe moving all the code that uses settings together already helps?

evap/student/views.py
Kakadus: I called it `GlobalRewards.RewardProgress` which seemed to be a decent compromise to me

deployment/localsettings.template.py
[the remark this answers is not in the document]
Kakadus: tested that this cannot happen with `max_reward_votes` and `next_reward_remaining_votes` are rounded correctly. The template filter is okay, because it uses round and `round(float("0.XX")*100) == XX` for all XX < 100.

evap/student/templates/student_index.html
Kakadus: should be obsolete

deployment/localsettings.template.py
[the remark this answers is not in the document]
Kakadus: with the current design, this is not easily tested anymore; the only position where `max_reward_votes` is used is indirectly for `width-percent-XX`, where we would have a rounding error (`voter_count=4`, `participant_count=100`, `Fraction("0.07")`)

evap/student/views.py
richardebeling: Sounds good to me, resolving for v1

evap/student/templates/student_index.html
richardebeling: Both inlined and in an extra file are fine for me here

evap/student/views.py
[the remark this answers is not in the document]
richardebeling: @niklasmohrin is this (mostly) resolved with the current state? I think the methods / code structure makes sense. the "creaming snake case" problem is still relevant with `settings.X`, but we will have to use those identifiers somewhere, so I don't see what we can improve there.

evap/student/templates/student_index.html
richardebeling: UI/UX discussion with @janno42 resulted in: We are fine with that for v1. We likely want to have a second, more fancy version of the UI anyway. Resolving for now.

evap/student/templates/student_index.html
richardebeling: Never mind, was moved out in [7946709](https://github.com/e-valuation/EvaP/pull/2234/commits/79467098d0aa5e7774eb4a612aeacca831bf66e1). Resolving.

evap/student/tests/test_views.py
```diff
@@ -52,6 +55,95 @@ def test_num_queries_is_constant(self):
...
+        # more voters than required for last reward
+        baker.make(
+            Evaluation,
+            course__semester=semester,
+            _voter_count=89,
+            _participant_count=97,
+            state=Evaluation.State.EVALUATED,
+        )
+        page = self.app.get(self.url, user=self.user)
+        self.assertIn("89 submitted evaluations (91%)", page)  # 91% is intentionally rounded down
```
richardebeling, line 134: @janno42 the current code uses a rounded value here. Hypothetical szenario: 89 voters, 97 participants, so participation at 89/97 ~= 0.9175. The bar currently shows "92", it has the width for 92%, and it would thus look as if a 92%-goal was hit, while it technically was not yet reached. I'd want to be very careful about making a possibly wrong promise here.

I guess the production team has to decide what should happen here. If we're fine with the rounding, this test line can be adapted to pass. Otherwise, we have to fix the rounding in the template.

richardebeling (approved):

evap/student/tests/test_views.py
janno42: It would be good to always round down. I would expect to need 90 voters to get to 92% here.
We could use a custom template filter? I don't like `widthratio` anyway, both the name and the parameters feel weird for our use case.

evap/student/views.py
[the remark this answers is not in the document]
niklasmohrin: For me, these are the two steps that happen here:
```python
@staticmethod
def for_evaluations(
    evaluations: QuerySet[Evaluation],
    rewards: Collection[tuple[Fraction, str]],
    info_text: str,
) -> "GlobalRewards":
    vote_count, participation_count = (
        Evaluation.annotate_with_participant_and_voter_counts(evaluations)
        .aggregate(Sum("num_voters", default=0), Sum("num_participants", default=0))
        .values()
    )

    max_reward_vote_ratio, __ = max(rewards)
    max_reward_votes = math.ceil(max_reward_vote_ratio * participation_count)

    rewards_with_progress = [
        GlobalRewards.RewardProgress(progress=vote_ratio / max_reward_vote_ratio, vote_ratio=vote_ratio, text=text)
        for vote_ratio, text in rewards
    ]

    last_vote_datetime = VoteTimestamp.objects.filter(evaluation__in=evaluations).aggregate(Max("timestamp"))[
        "timestamp__max"
    ]

    return GlobalRewards(
        vote_count=vote_count,
        participation_count=participation_count,
        max_reward_votes=max_reward_votes,
        bar_width_votes=min(vote_count, max_reward_votes),
        last_vote_datetime=last_vote_datetime,
        rewards_with_progress=rewards_with_progress,
        info_text=info_text,
    )

@staticmethod
def from_settings() -> "GlobalRewards | None":
    if not settings.GLOBAL_EVALUATION_PROGRESS_REWARDS:
        return None

    if not Semester.active_semester():
        return None

    evaluations = (
        Semester.active_semester()
        .evaluations.filter(is_single_result=False)
        .exclude(state__lt=Evaluation.State.APPROVED)
        .exclude(is_rewarded=False)
        .exclude(id__in=settings.GLOBAL_EVALUATION_PROGRESS_EXCLUDED_EVALUATION_IDS)
        .exclude(course__type__id__in=settings.GLOBAL_EVALUATION_PROGRESS_EXCLUDED_COURSE_TYPE_IDS)
        .exclude(course__is_private=True)
    )

    return GlobalRewards.for_evaluations(
        evaluations,
        settings.GLOBAL_EVALUATION_PROGRESS_REWARDS,
        settings.GLOBAL_EVALUATION_PROGRESS_INFO_TEXT[get_language()],
    )
```

I prefer this version, but in the end, either version is fine, you can decide which one you want to take

evap/student/tests/test_views.py
```diff
@@ -52,6 +54,95 @@ def test_num_queries_is_constant(self):
         with self.assertNumQueries(FuzzyInt(0, 100)):
             self.app.get(self.url, user=self.user)
 
+    @override_settings(
+        GLOBAL_EVALUATION_PROGRESS_REWARDS=[(0.1, "a dog"), (0.5, "a quokka")],
```
niklasmohrin, line 58:
```suggestion
        GLOBAL_EVALUATION_PROGRESS_REWARDS=[(Fraction(1, 10), "a dog"), (Fraction(5, 10), "a quokka")],
```

evap/static/scss/components/_progress.scss
```diff
@@ -54,3 +54,32 @@
 .multi-progress-bar .progress-container {
     padding-bottom: 3px;
 }
+
+.global-rewards-progress-bar {
+    position: relative;
+    display: grid;
+    padding-top: .5rem;
+    padding-right: 1.5rem;
+    padding-left: 1.5rem;
```
niklasmohrin, line 63:
```suggestion
    padding-inline: 1.5rem;
```

evap/static/scss/_mixins.scss
```diff
@@ -1,7 +1,7 @@
 @import "../bootstrap/scss/mixins";
 
 @mixin bar-shadow($color) {
-    text-shadow: 0 0 4px $color, 0 0 4px $color, 0 0 2px $color;
+    text-shadow: 0 0 2px $color, 0 0 2px $color, 0 0 2px $color, 0 0 2px $color, 0 0 2px $color, 0 0 2px $color;
```
niklasmohrin, line 4: Fine with me, I wonder whether we could achieve the same by manipulating the color directly and sticking with one shadow - no idea though :shrug:

evap/student/templates/student_global_reward.html
```diff
@@ -0,0 +1,53 @@
...
+                            {% endblocktranslate %}
+                        </span>
+                    </span>
+                </span>
+                {% for reward in global_rewards.rewards_with_progress %}
+                    <div class="position-relative progress-step left-percent-{% widthratio reward.progress 1 100 %}">
+                        <div class="mx-auto seperator"></div>
+                        <div class="small text-center">
+                            {% widthratio reward.vote_ratio 1 100 as percent %}
+                            {% blocktranslate %}{{ percent }}%{% endblocktranslate %}
```
niklasmohrin, line 33: This shouldn't need to be translated, right?
```suggestion
                            {% widthratio reward.vote_ratio 1 100 %}%
```

niklasmohrin (approved):

evap/static/scss/_mixins.scss
janno42: the problem is that there is not really a high-contrast color on either of the two blues, so we need a shadow to create the necessary contrast, and a smaller number of shadows will not increase the contrast enough.

evap/student/views.py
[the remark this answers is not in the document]
janno42: we'll keep the current version because i want to ship it today :shipit:

janno42 (approved):

evap/evaluation/templatetags/evaluation_filters.py
```diff
@@ -108,6 +108,16 @@ def percentage_one_decimal(fraction, population):
...
 
 
+@register.filter
+def percentage_zero_on_error(fraction, population):
+    try:
+        return f"{int(float(fraction) / float(population) * 100):.0f}%"
+    except ValueError:
+        return "0%"
+    except ZeroDivisionError:
+        return "0%"
```
Kakadus, line 118: we could combine the except clauses
```suggestion
    except (ZeroDivisionError, ValueError):
        return "0%"
```

Kakadus (commented):
:+1:

PR #2237, opened 2024-07-01
Fix infobox collapse

Kakadus:
fixes a bug, when the info box is closed during opening of the infobox. Now, we hard abort the opening process and further clear the "opening" timeout to prevent the class to be removed unexpectedly.

### DIFF
--- a/evap/static/ts/src/infobox.ts
+++ b/evap/static/ts/src/infobox.ts
@@ -17,6 +17,7 @@ export class InfoboxLogic {
         // close the infobox and save state
         this.closeButton.addEventListener("click", event => {
             this.infobox.classList.add("closing");
+            this.infobox.classList.remove("opening");
             setTimeout(() => {
                 this.infobox.classList.replace("closing", "closed");
             }, OPEN_CLOSE_TIMEOUT);
--- a/evap/static/ts/src/infobox.ts
+++ b/evap/static/ts/src/infobox.ts
@@ -6,6 +6,7 @@ export class InfoboxLogic {
     private readonly infobox: HTMLDivElement;
     private readonly closeButton: HTMLButtonElement;
     private readonly storageKey: string;
+    private timeout?: number;
 
     constructor(infobox_id: string) {
         this.infobox = selectOrError("#infobox-" + infobox_id);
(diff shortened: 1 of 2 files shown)


richardebeling (approved):

evap/static/ts/tsconfig.json
```diff
@@ -4,7 +4,7 @@
         "target": "es2019",
         "strict": true,
         "esModuleInterop": true,
-        "moduleResolution": "node"
+        "moduleResolution": "node",
```
niklasmohrin, line 7: I think trailing commas are disallowed in standard json - lets revert this line?

evap/static/ts/src/infobox.ts
```diff
@@ -6,6 +6,7 @@ export class InfoboxLogic {
     private readonly infobox: HTMLDivElement;
     private readonly closeButton: HTMLButtonElement;
     private readonly storageKey: string;
+    private timeout?: NodeJS.Timeout;
```
niklasmohrin, line 9: This seems odd, we are not targetting node, so the inferred type should be a browser timeout, not a nodejs one. Do we have typescript misconfigured somewhere? 🤔

evap/static/ts/src/infobox.ts
```diff
@@ -17,6 +18,8 @@ export class InfoboxLogic {
         // close the infobox and save state
         this.closeButton.addEventListener("click", event => {
             this.infobox.classList.add("closing");
+            this.infobox.classList.remove("opening");
+            clearTimeout(this.timeout);
```
niklasmohrin, line 22: Should we reset `this.timeout` afterwards? Also in the case the timeout callback is triggered normally? Or do we not care?

niklasmohrin (commented):
Makes sense, havent tested, thoughts:

evap/static/ts/src/infobox.ts
Kakadus: clearTimeout ignores invalid timeout parameters, so  I think it should be fine

evap/static/ts/src/infobox.ts
Kakadus: okay: by default, tsc imports all `@types/*` into the global namespace. As we have `@types/node` because of jest, this works here.

If I excluded the jest types (and thereby node's types) this compiles with `number` but compilation of the ts tests fail, see https://github.com/e-valuation/EvaP/pull/2237/commits/e501844e1ecb1b4280a9d91684e052c31b38a6fa

niklasmohrin: Ah, okay. According to [this](https://stackoverflow.com/a/51040768), using `window.setTimeout` (or the returntype trick they use) could work - if not, we can come back to this later

Kakadus: now it works! Adding it to `tsconfig.compile.json` should work!

janno42 (approved):
✅ UI checked

PR #2242, opened 2024-07-08
only remount if REPO_FOLDER does not contain any items

fekoch:
This allows users to rerun `vagrant provision` without a failure, if the folder is already mounted

### DIFF
--- a/deployment/provision_vagrant_vm.sh
+++ b/deployment/provision_vagrant_vm.sh
@@ -44,7 +44,8 @@ cp /etc/skel/.bashrc /home/$USER/
 OWNER=$(stat -c %u "$MOUNTPOINT/evap")
 apt-get -q install -y bindfs
 mkdir -p "$REPO_FOLDER"
-bindfs --map="$OWNER/1042:@$OWNER/@1042" "$MOUNTPOINT" "$REPO_FOLDER" || exit 1
+# remount if REPO_FOLDER does not contain any files (meaning it is not mounted)
+[[ -z $(ls -A "$REPO_FOLDER") ]] && bindfs --map="$OWNER/1042:@$OWNER/@1042" "$MOUNTPOINT" "$REPO_FOLDER" || exit 1
 echo "[[ -z \$(ls -A '$REPO_FOLDER') ]] && sudo bindfs --map='$OWNER/1042:@$OWNER/@1042' '$MOUNTPOINT' '$REPO_FOLDER'  #  remount iff folder empty" >> /home/$USER/.bashrc
 
 # allow ssh login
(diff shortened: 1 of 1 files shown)


Kakadus (approved):
thanks!

richardebeling:
My mental model was that bindfs will print an error (something like "mountpoint not empty") and exit immediately if the destination directory contains any files. This PR then just suppresses the error, but doesn't fix any actual failure (because there is none), right?

I'm wondering if it makes sense to keep the error. `vagrant provision` is noisy anyway. If stuff breaks because the mountpoint is not empty with the wrong files (some leftover files in there, wrong source dir mounted, idk), the error message might actually help finding that problem.

fekoch:
The issue is that the mounting is run as part of provisioning the VM. I believe it sometimes makes sense to run `vagrant provision` to ensure all software is up to date. In that case it is expected that the folder contains files because it is already mounted and should not be remounted.

Another idea is that we could, if already mounted, unmount the folder before mounting it.

richardebeling:
> The issue is that the mounting is run as part of provisioning the VM. I believe it sometimes makes sense to run `vagrant provision` to ensure all software is up to date. In that case it is expected that the folder contains files because it is already mounted and should not be remounted.

What I was trying to say is: On main, there is no problem blocking the remaining provision operations, right? The bindfs call will print an error and exit, but since the folder is already mounted, all following commands work fine.

I might be missing something here, but from my current understanding, the change here only suppesses an error message output to the console, but doesn't change any other part of the behavior. I'm thinking that we'd rather want to keep the error message, so that if something is unexpected with the mounting point, we will be hinted towards that.

fekoch:
It would change the behaviour. Currently, if one runs `vagrant provision` and the folder is already mounted, the provision script will not move past the mounting-attempt. It exits there. 

I was thinking that it would be useful to continue with the provision script (eg. to make sure all dependencies are up to date, all migrations are run, ...) even if the folder is already mounted.

deployment/provision_vagrant_vm.sh
```diff
@@ -44,7 +44,8 @@ cp /etc/skel/.bashrc /home/$USER/
 OWNER=$(stat -c %u "$MOUNTPOINT/evap")
 apt-get -q install -y bindfs
 mkdir -p "$REPO_FOLDER"
-bindfs --map="$OWNER/1042:@$OWNER/@1042" "$MOUNTPOINT" "$REPO_FOLDER" || exit 1
+# remount if REPO_FOLDER does not contain any files (meaning it is not mounted)
+[[ -z $(ls -A "$REPO_FOLDER") ]] && bindfs --map="$OWNER/1042:@$OWNER/@1042" "$MOUNTPOINT" "$REPO_FOLDER" || exit 1
```
Kakadus, line 48: change to if-else clause with a logging message in the else clause

fekoch: @Kakadus done

richardebeling (approved):

Kakadus (approved):

PR #2245, opened 2024-07-15
Fix has_answers returning False once any RatingResult has no answers

janno42:


richardebeling (commented):
The change is fine but I'd say we should really have a test for this

niklasmohrin (approved):

janno42 (commented):
👍

PR #2249, opened 2024-07-16
Make natural key of userprofile unique

### LINKED ISSUE #2247: Loading of backup fails due to nullable `natural_key`: the `USERNAME_FIELD`
Loading a larger backup which contains multiple users with their email `None` fails with a unique constraint failure.

This should be reproduced by expanding the testdata accordingly and fixed by replacing the default natural key with `(self.pk,settings.USERNAME_FIELD)`

Kakadus:
close #2247

~size of `test_data.json` increases a _bit_ because every relation to userprofile now takes two lines...~
size of `test_data.json` decreases a _bit_ because every relation to userprofile now takes two less lines...


Kakadus:
next try: we use the pk as natural key for UserProfile. ModelBackend has to be disabled because it still passes the email (USERNAME_FIELD) to `get_by_natural_key`, which causes a value error.

Kakadus:
whyever the ci now breaks...

evap/evaluation/models.py
```diff
@@ -1643,6 +1643,11 @@ def create_superuser(self, *, email, password=None, first_name=None, last_name=N
         user.groups.add(Group.objects.get(name="Manager"))
         return user
 
+    def get_by_natural_key(
+        self, pk
+    ):  # pylint: disable=arguments-renamed  # natural_key is defined like this below; if django does pass `email`, then retrieval fails.
+        return self.get(pk=pk)
```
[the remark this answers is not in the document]
richardebeling: fine with me if we don't expect to ever need it

Kakadus:
mentioning https://code.djangoproject.com/ticket/35729

Kakadus:
okay.. Current solution: Add an unique, non-editable UUIDField, and make it the natural key

richardebeling:
> okay.. Current solution: Add an unique, non-editable UUIDField, and make it the natural key

I think this is worse than `del AbstractBaseUser.natural_key`. We're not a library but an "end user" of django, so I'd be very confident that we're the only ones using that class. Manipulating it in hacky ways to work around bugs seems fine to me.

The additional database fields adds data clutter and migration clutter and is harder to reverse once django fixes their stuff.

evap/evaluation/models.py
```diff
@@ -1645,7 +1646,70 @@ def create_superuser(self, *, email, password=None, first_name=None, last_name=N
...
-class UserProfile(AbstractBaseUser, PermissionsMixin):
+class EvapBaseUser(models.Model):
+    """This is strongly related to the django.contrib.auth.base_user.AbstractBaseUser model, but does not define natural_key."""
+
+    password = models.CharField(_("password"), max_length=128)
+    last_login = models.DateTimeField(_("last login"), blank=True, null=True)
+
+    # Stores the raw password if set_password() is called so that it can
+    # be passed to password_changed() after the model is saved.
+    _password = None
```
richardebeling, line 1657: Do we need this? I don't think we care about `password_changed`.

evap/evaluation/models.py
```diff
@@ -1645,7 +1646,70 @@ def create_superuser(self, *, email, password=None, first_name=None, last_name=N
...
+    def get_username(self):
+        """Return the username for this User."""
+        return getattr(self, self.USERNAME_FIELD)
+
+    @property
+    def is_anonymous(self):
+        """
+        Always return False. This is a way of comparing User objects to
+        anonymous users.
+        """
```
richardebeling, line 1671: I don't think our code uses this -- why do we need it?

(I wouldn't copy the docstring from the django class)

evap/evaluation/models.py
```diff
@@ -1645,7 +1646,70 @@ def create_superuser(self, *, email, password=None, first_name=None, last_name=N
...
+            self.set_password(raw_password)
+            # Password hash upgrades shouldn't be considered password changes.
+            self._password = None
+            self.save(update_fields=["password"])
+
+        return check_password(raw_password, self.password, setter)
+
+    def set_unusable_password(self):
+        # Set a value that will never be a valid hash
+        self.password = make_password(None)
```
richardebeling, line 1702: Can we just make this the default value (factory) for the field? There's only one usage which would go away.

evap/evaluation/models.py
```diff
@@ -1645,7 +1646,70 @@ def create_superuser(self, *, email, password=None, first_name=None, last_name=N
...
+    # Stores the raw password if set_password() is called so that it can
+    # be passed to password_changed() after the model is saved.
+    _password = None
+
+    class Meta:
+        abstract = True
+
+    def get_username(self):
+        """Return the username for this User."""
+        return getattr(self, self.USERNAME_FIELD)
```
richardebeling, line 1664: Why needed?

Kakadus: required by django-webtest, see https://redirect.github.com/django-webtest/django-webtest/issues/134.

evap/evaluation/models.py
Kakadus: needed by `django.contrib.auth.checks`

richardebeling (commented):
Code looks good to me, haven't verified the changes to test_data.json yet.

richardebeling (approved):
Thanks!

evap/evaluation/models_logging.py
```diff
@@ -19,6 +21,17 @@
...
 
+CREATE_LOGENTRIES = True
+
+
+@contextmanager
+def disable_logentries() -> Iterator[None]:
+    global CREATE_LOGENTRIES  # noqa: PLW0603
+    old_mode = CREATE_LOGENTRIES
+    CREATE_LOGENTRIES = False
+    yield
```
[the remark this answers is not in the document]
Kakadus: you are right! I did not expected this...

evap/evaluation/models.py
```diff
@@ -1668,7 +1670,48 @@ def create_superuser(self, *, email, password=None, first_name=None, last_name=N
...
 
 
-class UserProfile(AbstractBaseUser, PermissionsMixin):
+class EvapBaseUser(models.Model):
+    """This is strongly related to the django.contrib.auth.base_user.AbstractBaseUser model, but does not define natural_key."""
+
+    USERNAME_FIELD: str
+
+    # set unusable password by default: We don't want password resets.
+    password = models.CharField(_("password"), max_length=128, default=partial(make_password, None))
```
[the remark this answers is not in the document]
richardebeling: We don't need a `password_changed` event in our code, and we also do not use [password validation](https://docs.djangoproject.com/en/5.1/topics/auth/passwords/#module-django.contrib.auth.password_validation).

We could add an assertion somewhere that would trigger if anyone ever tried to configure password validation, something like
```python3
assert AUTH_PASSWORD_VALIDATORS == [], "Password validation configured, but evap will not apply it"
```
that crashes at application startup.

evap/evaluation/migrations/0147_unusable_password_default.py
```diff
@@ -0,0 +1,24 @@
...
+    dependencies = [
+        ("evaluation", "0146_grade_reminder_template"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="userprofile",
+            name="password",
+            field=models.CharField(
+                default=functools.partial(django.contrib.auth.hashers.make_password, *(None,)),
```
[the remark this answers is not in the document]
Kakadus: this is necessary so that django does not create a migration.... I get this otherwise:
`You have 1 unapplied migration(s). Your project may not work properly until you apply the migrations for app(s): evaluation.`

evap/evaluation/models.py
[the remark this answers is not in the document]
Kakadus: just to mention here: the deprecation is because of the `*args` parameter (https://github.com/django/django/commit/3915d4c70d0d7673abe675525b58117a5099afd3), not because of the `password_changed` call.

However, that assertion seems still valuable to have :+1:

niklasmohrin (approved):

PR #2250, opened 2024-07-16
Fix user delete form submit

### LINKED ISSUE #2248: user delete form not working
The html standard does not support nested forms. The user form is nested though. Firefox thus just removes the user delete form and submitting the modal submits the parent user edit form.

Kakadus:
close #2248

don't nest the forms, fix layout and redirect to student index when user has been deleted


evap/staff/views.py
```diff
@@ -2252,7 +2252,7 @@ def user_delete(request):
         raise SuspiciousOperation("Deleting user not allowed")
     user.delete()
     messages.success(request, _("Successfully deleted user."))
-    return HttpResponse()  # 200 OK
+    return HttpResponseRedirect(reverse_lazy("staff:user_index"))
```
niklasmohrin, line 2255:
```suggestion
    return redirect("staff:user_index")
```

evap/staff/templates/staff_user_form.html
```diff
@@ -113,35 +113,33 @@ <h5 class="card-title me-auto">{% translate 'Export evaluation results' %}</h5>
...
         {% endif %}
-        <div class="card card-submit-area text-center mb-3">
-            <div class="card-body">
-                <button type="submit" class="btn btn-primary">{% translate 'Save user' %}</button>
-                {% if form.instance and form.instance.can_be_deleted_by_manager %}
-
-
-                    <form reload-on-success method="POST" action="{% url 'staff:user_delete' %}">
-                        {% csrf_token %}
+    </form>
```
niklasmohrin, line 116: I thought that instead of making the big form smaller, we would instead move the deletion-form outside of it like so:

```patch
diff --git a/evap/staff/templates/staff_user_form.html b/evap/staff/templates/staff_user_form.html
index 99d05715..87623889 100644
--- a/evap/staff/templates/staff_user_form.html
+++ b/evap/staff/templates/staff_user_form.html
@@ -56,6 +56,9 @@
         </div>
     {% endif %}
 
+    <form id="user-deletion-form" class="d-none" method="POST" action="{% url 'staff:user_delete' %}">
+        {% csrf_token %}
+    </form>
 
     <form id="user-form" method="POST" class="form-horizontal multiselect-form">
         {% csrf_token %}
@@ -117,12 +120,7 @@
             <div class="card-body">
                 <button type="submit" class="btn btn-primary">{% translate 'Save user' %}</button>
                 {% if form.instance and form.instance.can_be_deleted_by_manager %}
-
-
-                    <form reload-on-success method="POST" action="{% url 'staff:user_delete' %}">
-                        {% csrf_token %}
-
-                        <confirmation-modal type="submit" name="user_id" value="{{ form.instance.id }}" confirm-button-class="btn-danger">
+                        <confirmation-modal type="submit" form="user-deletion-form" name="user_id" value="{{ form.instance.id }}" confirm-button-class="btn-danger">
                             <span slot="title">{% translate 'Delete user' %}</span>
                             <span slot="action-text">{% translate 'Delete user' %}</span>
                             <span slot="question">
@@ -134,7 +132,6 @@
 
                             <button slot="show-button" type="button" class="btn btn-danger">{% translate 'Delete user' %}</button>
                         </confirmation-modal>
-                    </form>
                 {% else %}
                     <span tabindex="0" data-bs-toggle="tooltip" title="{% blocktranslate %}This user contributes to an evaluation, participates in an evaluation whose participations haven't been archived yet or has special rights and as such cannot be deleted.{% endblocktranslate %}">
                     <button type="button" disabled class="btn btn-danger">
```

(plus fixing the indentation, which I omitted here)

For me, this is the more natural structure - the other form is big anyways and that is fair, most of the page is about this very form. What do you think?

Kakadus: you are definitely right. I did not had the idea to move the deletion form. Instead I tried to recycle the user form, which did not work. This works though :+1:

niklasmohrin (approved):

janno42 (approved):
✔ UI tested

richardebeling (approved):
Code looks good

PR #2253, opened 2024-07-29
Abort confirmation modal flow if attached form is invalid

niklasmohrin:
Otherwise, the modal would just close without anything happening.

This first came up at https://github.com/e-valuation/EvaP/pull/2050#discussion_r1661445961.

One problem that could occur is that there is an invalid input present, but outside of the modal, so it is hard to tell what is wrong. Optimally, the modal confirm button could then display a message like "there are form errors outside of this modal".

Since we currently only have one occurrence of `setCustomValidity`, there should be no problems with this new behavior of the modal for the existing modals.


niklasmohrin:
Actually, there is an easier way, which is already used for semester deletion: Overwrite the `submit-group` slot and specify a new submit button - this button is then attached to the outer form, not the form that triggers the submit handler of the `dialogForm` of the modal. I wonder if that approach is universally better or whether we should convert that to this new pattern instead

richardebeling (approved):
Code is fine with me.

> Overwrite the submit-group slot and specify a new submit button - this button is then attached to the outer form, not the form that triggers the submit handler of the dialogForm of the modal. I wonder if that approach is universally better or whether we should change the semester deletion modal to this new pattern instead

I have no clue. Both approaches sound reasonable to me, I think you can judge this better.

PR #2261, opened 2024-08-05
Do calculation of redemptions by user in database

hansegucker:


evap/rewards/models.py
```diff
@@ -34,13 +33,9 @@ def can_delete(self):
         return not self.reward_point_redemptions.exists()
 
     def redemptions_by_user(self):
```
richardebeling, line 35: Maybe also rename to `users_with_redeemed_points` or similar?

richardebeling (approved):

janno42 (approved):

evap/rewards/models.py
niklasmohrin: + Also where it is used :)

evap/rewards/exporters.py
```diff
@@ -17,5 +17,5 @@ def export_impl(self, redemptions_by_user):  # pylint: disable=arguments-differ
             "bold",
         )
 
-        for user, value in redemptions_by_user.items():
-            self.write_row([user.last_name, user.first_name, user.email, value])
+        for user in redemptions_by_user:
```
richardebeling, line 20: also rename to `users_with_redeemed_points`?

niklasmohrin (approved):

PR #2262, opened 2024-08-05
Dropped Courses

### LINKED ISSUE #991: Dropped courses

fekoch:
fixes #991

(@janno42) As discussed, we want:

- [x] A check-box "Allow drop-out?" when creating new evaluations
- [x] ~~A setting "DROPOUT_QUESTIONNAIRE_ID" to set which questionnaire should be shown when using the drop-out function.~~
- [x] The "I dropped this course" button should be shown only on the student index and only for evaluations that "allow drop-out"
- [x] Show Dropout Questionnaire answers in Result UI
  + (only shown to people who can see general text answers)
- [x] ~~Dropout questionnaire Textanswers should not be shown in the textanswer review~~


evap/evaluation/models.py
```diff
@@ -438,6 +441,8 @@ class State(models.IntegerChoices):
     # Disable to prevent editors from changing evaluation data
     allow_editors_to_edit = models.BooleanField(verbose_name=_("allow editors to edit"), default=True)
 
+    allow_drop_out = models.BooleanField(verbose_name=_("allow editors to drop out"), default=True)
```
richardebeling, line 444: I guess

```suggestion
    allow_drop_out = models.BooleanField(verbose_name=_("allow students to drop out"), default=True)
```

evap/static/font-awesome/scss/_mixins.scss
```diff
@@ -70,4 +70,3 @@
     content: unquote("\"#{ $fa-var }\"");
   }
 }
-
```
richardebeling, line 73: In one of the earlier files, you added a final newline. This now removes it (I think?)

evap/static/font-awesome/scss/fontawesome.scss
```diff
@@ -1,7 +1,7 @@
 /*!
- * Font Awesome Free 6.1.1 by @fontawesome - https://fontawesome.com
+ * Font Awesome Free 6.6.0 by @fontawesome - https://fontawesome.com
```
richardebeling, line 2: Feel free to move the font awesome update to a separate PR

evap/student/templates/student_index_evaluate_or_drop.html
```diff
@@ -0,0 +1,16 @@
...
+        <span slot="question">{% translate 'If you mark this course as dropped, you can still give feedback, but ...' %}</span>
+        <!-- TODO@Felix: explanatory text -->
+        <a class="btn btn-danger ms-2" slot="submit-group"
+           href="{% url 'student:drop' evaluation.id %}">{% translate 'I dropped this course' %}</a>
+        <button slot="show-button" class="btn btn-sm btn-secondary" title="{% translate 'I dropped this course' %}"
+                data-bs-toggle="tooltip" data-bs-placement="top">
+            <span class="fas fa-user-xmark"></span>
+        </button>
+    </confirmation-modal>
+{% endif %}
```
richardebeling, line 16: Missing final newline

evap/evaluation/models.py
```diff
@@ -222,6 +236,18 @@ def __lt__(self, other):
...
+    @transaction.atomic()
+    def set_active_dropout(self):
+        # TODO@Felix: test this method
+        if self.type != Questionnaire.Type.DROPOUT:
+            raise ValueError("Can only set DROPOUT-type Questionnaires as active dropout questionnaire.")
+
+        Questionnaire.objects.active_dropout_questionnaire().update(is_active_dropout_questionnaire=False)
+        self.is_active_dropout_questionnaire = True
+        self.save()
+        # TODO@Felix: why does this not work [wip]
```
richardebeling, line 248: I think the (unique=True, blank=True, null=True) tri-state approach forces us to have at most one instance with value `True` and all other instance with value `None`/`NULL`. The update above sets it to `False`, which would give it a unique validation failure. Guess that's one issue here

evap/evaluation/models.py
```diff
@@ -162,6 +163,12 @@ def general_questionnaires(self):
     def contributor_questionnaires(self):
         return super().get_queryset().filter(type=Questionnaire.Type.CONTRIBUTOR)
 
+    def dropout_questionnaires(self):
```
richardebeling, line 166: If you're touching these and it's no trouble, it would probably be nice if you could also add the return type annotation (`-> QuerySet[Questionnaire]` or whatever it should be)

(same for the views below that return `HttpResponse`. It's a bit annoying, but for backwards compatibility, no annotation always means "Any", even if the method always returns some specific type. I wonder if [mypy issue 4409](https://www.github.com/python/mypy/issues/4409) will one day be implemented.)

evap/evaluation/models.py
```diff
@@ -191,6 +199,12 @@ class Type(models.IntegerChoices):
 
     order = models.IntegerField(verbose_name=_("ordering index"), default=0)
 
+    # (unique=True, blank=True, null=True) allows having multiple non-active but only one active questionnaire
+    # TODO@Felix: hidden True?
+    is_active_dropout_questionnaire = models.BooleanField(
+        default=None, unique=True, blank=True, null=True, verbose_name=_("questionnaire is selected as active ")
```
richardebeling, line 205:
```suggestion
        default=None, unique=True, blank=True, null=True, verbose_name=_("questionnaire is selected as active")
```

evap/evaluation/models.py
```diff
@@ -222,6 +236,18 @@ def __lt__(self, other):
     def __gt__(self, other):
         return (self.type, self.order, self.pk) > (other.type, other.order, other.pk)
 
+    @transaction.atomic()
+    def set_active_dropout(self):
+        # TODO@Felix: test this method
+        if self.type != Questionnaire.Type.DROPOUT:
+            raise ValueError("Can only set DROPOUT-type Questionnaires as active dropout questionnaire.")
```
richardebeling, line 243: probably fair to just use an assertion -- or do we expect this to ever happen for a correct program execution?

evap/staff/templates/staff_questionnaire_index_list.html
```diff
@@ -32,7 +34,13 @@
                                     {% blocktranslate count questionnaire.contributions.count as count %}used {{ count }} time{% plural %}used {{ count }} times{% endblocktranslate %}
                                 </td>
                                 <td>
-                                    {% if type != 'contributor' %}
+                                    {% if type == 'dropout' %}
+                                        <!-- TODO@Felix: Selection via AJAX-->
+                                        <a href="{% url 'staff:questionnaire_set_active_dropout' questionnaire.id %}"
+                                           class="btn btn-sm btn-light {% if questionnaire.is_active_dropout_questionnaire %} active{% endif %}">
```
richardebeling, line 40: We commonly do
```suggestion
                                           class="btn btn-sm btn-light{% if questionnaire.is_active_dropout_questionnaire %} active{% endif %}">
```
so that the resuling whitespace after template processing is canonical

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -59,7 +59,11 @@ <h4 class="{% if not forloop.first %} mt-3 {% elif contributor %} me-auto {% end
...
                                 {% for choice, color in field|zip:field.field.widget.attrs.choices.colors %}
-                                    <input id="{{ choice.id_for_label }}" class="tab-selectable num-selectable btn-check" name="{{ choice.data.name }}" type="radio" value="{{ choice.data.value }}" autocomplete="off"{% if field.value == choice.data.value %} checked{% endif %}{% if preview %} disabled{% endif %}/>
+                                    <!-- TODO@Janno: It seems to me that the "checked" comparison could not work the way it was before.
+                                                     Maybe somewhen between now and 2014 someone changed either field or choice.data.value?
+                                                     I believe now it does the thing it should do
+                                                     -->
+                                    <input id="{{ choice.id_for_label }}" class="tab-selectable num-selectable btn-check" name="{{ choice.data.name }}" type="radio" value="{{ choice.data.value }}" autocomplete="off"{% if field.field.value.0 == choice.data.value %} checked="checked"{% endif %}{% if preview %} disabled{% endif %}/>
```
fekoch, line 66: @janno42

evap/student/templates/student_index_evaluate_or_drop.html
```diff
@@ -0,0 +1,24 @@
...
+        <span slot="title">{% translate 'Did you drop this course?' %}</span>
+        <span slot="question">
+            {% blocktranslate %}
+                If you mark this course as dropped, you can still give feedback like normal, but your answers will only
+                be visible to contributors.
+            {% endblocktranslate %}
+            <br>
+            {% blocktranslate %}
+                You will not be able to vote for this course afterwards.
+            {% endblocktranslate %}
```
fekoch, line 15: @janno42 is this sensible? Or do you need some translation context or whatnot?

janno42, line 15: I think this would be fine

```suggestion
            {% blocktranslate %}
                If you mark this course as dropped, you will still be able to provide feedback as usual. However, you can only submit the questionnaire once.
            {% endblocktranslate %}
```

fekoch:
@janno42 New proposal: You can add drop-out-questionnaires for each evaluation, just like adding general questionnaires.

image

Simply by having at least one drop-out-questionnaire, the `is_dropout_allowed`-property is set on an evaluation and students can use the drop-out button in the UI.

A) Does this seem like a good idea?

and

B) Do we still need to have the whole "active-drop-out-questionnaire" logic if we implement it like that?

fekoch:
@janno42: is this still relevant?

> Dropout questionnaire Textanswers should not be shown in the textanswer review

evap/student/templates/student_index_evaluate_or_drop.html
```diff
@@ -0,0 +1,20 @@
+<a class="btn btn-sm btn-primary btn-row-hover" href="{% url 'student:vote' evaluation.id %}">
+    {% translate 'Evaluate' %}
+</a>
+{% if evaluation.is_dropout_allowed %}
+    <confirmation-modal>
+        <span slot="title">{% translate 'Did you drop this course?' %}</span>
+        <span slot="question">
+            {% blocktranslate %}
+                If you mark this course as dropped, you will still be able to provide feedback as usual. However, you
+                can only submit the questionnaire once.
```
janno42, line 10:
```suggestion
                If you mark this course as dropped, you will still be able to provide feedback as usual. However, you
                will not be able to submit the questionnaire another time.
```

evap/student/templates/student_vote.html
```diff
@@ -55,6 +55,20 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
 
+        {% if show_dropout_questionnaire %}
+            <div class="card card-outline-warning mb-3">
+                <div class="card-header">
+                    <span class="fas fa-triangle-exclamation"></span> {% translate 'Dropout Questionnaire' %}
+                </div>
+                <div class="card-body tab-row">
+                    {% blocktranslate %}
+                        You are marking this course as dropped. Your answers will only be shown to contributors and will
+                        not be counted in this course's final rating.
```
janno42, line 66:
```suggestion
                        You are marking this course as dropped. Your feedback on questions you can answer is still very welcome.
```

janno42 (commented):
- The rating questions' vote buttons "no answer" look partly selected even after selecting other answers
- Please add the dropout questionnaire field to the contributor `EvaluationForm`
- Please add the dropout questionnaires to the questionnaire bulk assignment

evap/staff/forms.py
```diff
@@ -387,21 +395,30 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields["course"].queryset = Course.objects.filter(semester=semester)
 
+        # TODO: Wieso hier nicht Visibility.HIDDEN excluden
```
fekoch, line 398: @fekoch change to Visibility Hidden

evap/evaluation/templates/evaluation_form_general_questionnaires.html
```diff
@@ -1,8 +1,10 @@
...
 
+<!-- TODO: stimmt dieser template Name ? -->
+
 <fieldset>
     {% for field in evaluation_form %}
-        {% if field == evaluation_form.general_questionnaires %}
+        {% if field == evaluation_form.general_questionnaires or field == evaluation_form.dropout_questionnaires %}
             <div class="mb-3 d-flex">
                 {% include 'bootstrap_form_field_label.html' with field=field class='col-md-3 pe-4' %}
                 <div class="col-md-7{% if field.errors %} is-invalid{% endif %}">
```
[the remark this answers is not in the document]
janno42: do you mean `evaluation_form_general_questionnaires` although it now also includes the dropout questionnaires? that's okay for me.

janno42 (changes requested):

evap/evaluation/templates/evaluation_form_general_questionnaires.html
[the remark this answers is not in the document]
janno42: we decided renaming to `evaluation_form_fields` would be better because it renders all form fields, not just questionnaires.

evap/evaluation/templates/evaluation_form_fields.html
```diff
@@ -1,8 +1,10 @@
 {% load evaluation_filters %}
 
+<!-- TODO: stimmt dieser template Name ? -->
+
```
janno42, line 4: left over todo

janno42 (changes requested):
If the "Dropout Questionnaires" list in the contributor evaluation form is empty, the label should not be shown

evap/contributor/forms.py
```diff
@@ -73,6 +89,19 @@ def clean(self):
             self.add_error("vote_start_datetime", "")
             self.add_error("vote_end_date", _("The first day of evaluation must be before the last one."))
 
+        # Ensure locked questionnaires cannot be deselected and only unlocked ones can be added
+        selected_questionnaires = self.instance.general_contribution.questionnaires.filter(is_locked=True).distinct()
+        selected_questionnaires |= self.cleaned_data.get("general_questionnaires").filter(is_locked=False)
+        selected_questionnaires |= self.cleaned_data.get("dropout_questionnaires").filter(is_locked=False)
+
+        self.cleaned_data.update(
```
niklasmohrin, line 97: I find it a bit odd that we update the form values instead of throwing a `ValidationError` when things don't add up. In person, we discussed that we can leave it as is for this PR because it is the same behavior as before, but I think we should revisit this afterwards. What do you think @richardebeling @Kakadus ?

evap/evaluation/tests/test_views.py
```diff
@@ -284,3 +285,100 @@ def test_reset_to_new(self) -> None:
...
+                    )
+                else:
+                    self.assertEqual(
+                        fields[0].value,
+                        str(NO_ANSWER),
+                        "Answers to Questions in the general contribution should be set to NO_ANSWER",
+                    )
+
+    def test_choosing_dropout_sets_to_no_answer(self):
+        response = self.app.get(url=reverse("student:drop", args=[self.evaluation.id]), user=self.user, status=200)
```
niklasmohrin, line 332: I suppose this test class would be better placed in the `student` app

niklasmohrin (changes requested):
I haven't looked at results and staff yet, but here is what I have so far :)

janno42 (changes requested):
- The dropout button on the student index page should be placed inline before the "Evaluate" button.
- For the questions in the dropout questionnaire(s), no answers should be preselected.

evap/development/fixtures/test_data.json
```diff
@@ -140450,8 +140637,8 @@
   "pk": 452,
   "fields": {
     "is_superuser": true,
-    "password": "pbkdf2_sha256$600000$8dcuxbY2OgmCEEj3nlUQWG$qk5ic3Er9FrkCd3VUjC1DEfxd76LPVKrzaMEWGSaxyA=",
-    "last_login": "2023-09-18T21:02:20.716",
+    "password": "pbkdf2_sha256$870000$FI1vp6JGLvZULamq2a9pIL$TLLlUbFnxARDpJFRmuL+gVR9QtFc9fxww/K7pM1PixM=",
+    "last_login": "2025-02-17T18:41:52.756",
```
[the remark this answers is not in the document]
fekoch: reverted :+1:

evap/student/views.py
```diff
@@ -255,31 +303,38 @@ def render_vote_page(request, evaluation, preview, for_rendering_in_modal=False)
...
+        return render_vote_page(request, evaluation, preview=False, dropout=dropout)
 
     # all forms are valid, begin vote operation
     with transaction.atomic():
         # add user to evaluation.voters
         # not using evaluation.voters.add(request.user) since that fails silently when done twice.
         evaluation.voters.through.objects.create(userprofile_id=request.user.pk, evaluation_id=evaluation.pk)
 
+        if dropout:
+            evaluation.dropout_count += 1
```
niklasmohrin, line 327: As discussed in person: The evaluation is fetched before the `transaction.atomic` starts, so if two people drop out at roughly the same time, we loose one increase here

evap/results/templates/results_evaluation_detail.html
```diff
@@ -227,6 +227,23 @@ <h3>{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</h3>
         </div>
     {% endif %}
 
+    {% if evaluation.is_dropout_allowed or evaluation.dropout_count > 0 or dropout_questionnaire_results %}
+    <div class="card card-outline-primary mb-3">
```
niklasmohrin, line 231: Please indent

evap/results/tests/test_views.py
```diff
@@ -509,6 +509,30 @@ def test_evaluation_sorting(self):
...
+
+        cache_results(self.evaluation)
+
+        self.evaluation.general_contribution.questionnaires.add(questionnaire)
+        response = self.app.get(self.url, user=self.manager, status=200)
+
+        self.assertContains(response, "Dropout")
+        self.assertContains(response, "42")
+        self.assertContains(response, "15")
+        self.assertContains(response, "SomeQuestionText")
```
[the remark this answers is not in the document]
fekoch: Is it more clear like that?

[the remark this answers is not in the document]
niklasmohrin: I thought more that we would change the "42" to something like `<span class="the-class-for-our-feature>42</span>` or so, otherwise the assertion might pass because we have a 42 somewhere else, for example in an evaluation id or similar

evap/contributor/tests/test_views.py
```diff
@@ -272,6 +272,24 @@ def test_display_request_buttons(self):
         self.assertEqual(page.body.decode().count("Request changes"), 0)
         self.assertEqual(page.body.decode().count("Request creation of new account"), 2)
 
+    def test_questionnaire_input_hidden_if_options_empty(self):
```
niklasmohrin, line 275: If I understand correctly, the test currently does not show that responsibles can see their options for dropout questionnaires. If we change the code to only show previously selected questionnaires, it still passes. However, we want to guarantee that responsibles can themselves select dropout questionnaires.

I think you should have one dropout questionnaire that is only visible to managers, but selected, and one that is not selected, but visible to everyone.

evap/contributor/views.py
```diff
@@ -241,7 +243,7 @@ def evaluation_preview(request, evaluation_id):
     ):
         raise PermissionDenied
 
-    return render_vote_page(request, evaluation, preview=True)
+    return render_vote_page(request, evaluation, preview=True, dropout=False)
```
niklasmohrin, line 246: This means that we cannot preview the "dropout experience". I think this is fine for this PR, but we should probably later add a way for responsibles to preview what it looks like to drop out @janno42 ?

evap/development/fixtures/test_data.json
niklasmohrin, line 1: Can we also add the dropout questionnaire as the questionnaire of at least one evaluation in the test data? Otherwise it is tedious to always add it if someone wants to work on it

evap/evaluation/migrations/0148_evaluation_dropout_count_alter_questionnaire_type.py
```diff
@@ -0,0 +1,32 @@
...
+        ("evaluation", "0147_unusable_password_default"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="evaluation",
+            name="dropout_count",
+            field=models.IntegerField(default=0, verbose_name="dropout count"),
+        ),
+        migrations.AlterField(
```
niklasmohrin, line 18: @richardebeling What happens when reversing this migration? Do the dropout questionnaires just stay around, or are they deleted? And which behavior do we want :thinking:

evap/evaluation/models.py
```diff
@@ -157,18 +157,25 @@ def evaluations(self):
         return Evaluation.objects.filter(course__semester=self)
 
 
-class QuestionnaireManager(Manager):
-    def general_questionnaires(self):
-        return super().get_queryset().exclude(type=Questionnaire.Type.CONTRIBUTOR)
+class QuestionnaireManager(Manager["Questionnaire"]):
+    def general_questionnaires(self) -> QuerySet["Questionnaire"]:
+        return super().get_queryset().exclude(type__in=[Questionnaire.Type.CONTRIBUTOR, Questionnaire.Type.DROPOUT])
```
niklasmohrin, line 162:
```suggestion
        return super().get_queryset().filter(type__in=[Questionnaire.Type.TOP, Questionnaire.Type.BOTTOM])
```

evap/student/tools.py
```diff
@@ -1,8 +1,26 @@
...
+def parse_answer_field_id(formfield_id: str) -> tuple[int, int, int, bool]:
+    """Parses the contribution-, questionnaire- and question-id from a formfield string,
+    as well as if it is the field for the corresponding additional text answer"""
+
+    parts = formfield_id.split("_")
+    assert parts[0] == "question"
+
+    if len(parts) == 5 and parts[4] == "ta":
+        return *map(int, parts[1:4]), True  # type: ignore[return-value]
+
```
niklasmohrin, line 25:
```suggestion
    assert len(parts) == 4
```

evap/student/views.py
```diff
@@ -185,7 +194,40 @@ def sorter(evaluation):
...
+            for question in questionnaire.rating_questions
+        }
+
+    return QuestionnaireVotingForm(
+        request.POST or None, contribution=contribution, questionnaire=questionnaire, initial=initial
+    )
+
+
+def create_voting_forms(
+    request, contribution: Contribution, questionnaires: QuerySet[Questionnaire], preselect_no_answer: bool
```
niklasmohrin, line 214:
```suggestion
    request, contribution: Contribution, questionnaires: Iterable[Questionnaire], preselect_no_answer: bool
```

evap/student/views.py
```diff
@@ -185,7 +194,40 @@ def sorter(evaluation):
...
 
 
-def get_vote_page_form_groups(request, evaluation, preview):
+def create_voting_form(
+    request, contribution: Contribution, questionnaire: Questionnaire, preselect_no_answer: bool
+) -> QuestionnaireVotingForm:
+    initial = None
+
+    if preselect_no_answer:
+        initial = {
```
niklasmohrin, line 203: With recent ruff, I think this would suggest to use `dict.fromkeys`

evap/student/views.py
```diff
@@ -196,15 +238,19 @@ def get_vote_page_form_groups(request, evaluation, preview):
...
 
 
-def render_vote_page(request, evaluation, preview, for_rendering_in_modal=False):
-    form_groups = get_vote_page_form_groups(request, evaluation, preview)
+def render_vote_page(
+    request: HttpRequest,
+    evaluation: Evaluation,
+    preview: bool,
+    dropout: bool,
+    for_rendering_in_modal: bool = False,
```
niklasmohrin, line 251: As they are three bools now, let's make the keyword only (and also the other ones taking multiple preview and dropout bools):
```suggestion
    *,
    preview: bool,
    dropout: bool,
    for_rendering_in_modal: bool = False,
```

evap/student/views.py
```diff
@@ -255,21 +308,30 @@ def render_vote_page(request, evaluation, preview, for_rendering_in_modal=False)
 
 
 @participant_required
-def vote(request, evaluation_id):  # noqa: PLR0912
+def vote(request: HttpRequest, evaluation_id: int, dropout=False) -> HttpResponse:  # noqa: PLR0912
     # pylint: disable=too-many-nested-blocks
     evaluation = get_object_or_404(Evaluation, id=evaluation_id)
+
+    if dropout and not evaluation.is_dropout_allowed:
+        raise SuspiciousOperation("Drop out not allowed")
```
niklasmohrin, line 316:
```suggestion
        raise SuspiciousOperation("Dropping out is not allowed")
```

evap/student/views.py
```diff
@@ -185,7 +194,40 @@ def sorter(evaluation):
...
+            questionnaire,
+            preselect_no_answer=(preselect_no_answer and not questionnaire.is_dropout),
+            # dropout questionnaires should not be preselected
+        )
+        for questionnaire in questionnaires
+    ]
+
+
+def get_vote_page_form_groups(
+    request, evaluation: Evaluation, preview: bool, preselect_no_answer: bool
```
niklasmohrin, line 229: same here with keyword only: someone could pass a bool for dropout when they wanted to pass one for preview

evap/student/views.py
```diff
@@ -310,15 +371,15 @@ def vote(request, evaluation_id):  # noqa: PLR0912
...
         TextAnswer.objects.filter(contribution__evaluation=evaluation).update(id=F("id"))
 
+        # TODO@Felix: do this in on_commit? (or via select for update?)
         if not evaluation.can_publish_text_results:
             # enable text result publishing if first user confirmed that publishing is okay or second user voted
             if (
                 request.POST.get("text_results_publish_confirmation_top") == "on"
                 or request.POST.get("text_results_publish_confirmation_bottom") == "on"
-                or evaluation.voters.count() >= 2
+                or voters.filter(pk=evaluation.pk).count() >= 2
```
niklasmohrin, line 380: Does this new query still include the new voter, or is it constrained to the old set of voters? I would think that it is a fresh query, which then means that we can still use `evaluation.voters` right?

Currently, I am a bit confused why the `voters` from before is used, so a change or an explanatory comment would be good :)

evap/student/views.py
```diff
@@ -310,15 +371,15 @@ def vote(request, evaluation_id):  # noqa: PLR0912
         RatingAnswerCounter.objects.filter(contribution__evaluation=evaluation).update(id=F("id"))
         TextAnswer.objects.filter(contribution__evaluation=evaluation).update(id=F("id"))
 
+        # TODO@Felix: do this in on_commit? (or via select for update?)
```
niklasmohrin, line 374: I don't think `on_commit` is needed, the change of `can_publish_text_results` would be rolled back if the transaction fails, right?

niklasmohrin (commented):
Very nice work! I checked some more files off as viewed, but in some I have suggestions :)

evap/evaluation/migrations/0148_evaluation_dropout_count_alter_questionnaire_type.py
richardebeling: `migrations.AddField` above would be reverted automatically by removing the field.

The `migrations.AlterField` step here would probably go through. The choices are, afaik, not enforced by database constrained, just by Django model validation. This migration would not update existing instances in the database, so we'd still have instances in the database with `type=5`. Trying to save any such instance  would then throw a validation error in django later (and there's a chance we hit application logic errors / asserts about non-existing questionnaire types).

Probably the only sensible thing to do here is to delete all questionnaires of type 5. @fekoch there is a way to specify a custom "reverse" action for a migration step, can you add the explicit reverse step here to delete all data that, logically, shouldn't exist anymore? (Let us know if it's too much hassle, we don't really expect to ever hit it. Logically, it should be one delete query with cascading)

evap/contributor/views.py
janno42: I don't think responsibles need to see this. In practice I expect all dropout questionnaires to be locked so that only staff users are in control of selecting them.

evap/student/tools.py
```diff
@@ -1,8 +1,26 @@
...
+    """Parses the contribution-, questionnaire- and question-id from a formfield string,
+    as well as if it is the field for the corresponding additional text answer"""
+
+    parts = formfield_id.split("_")
+    assert parts[0] == "question"
+
+    if len(parts) == 5 and parts[4] == "ta":
+        return *map(int, parts[1:4]), True  # type: ignore[return-value]
+    assert len(parts) == 4
+    return tuple(list(map(int, parts[1:4])) + [False])  # type: ignore[return-value]
```
Kakadus, line 26: this can also be:

```suggestion
    return *map(int, parts[1:4]), False  # type: ignore[return-value]
```

evap/contributor/forms.py
janno42: The UI already prevents selecting/deselecting locked questionnaires. But the final check can't be on client site.

janno42 (changes requested):
- when no answers in the dropout questionnaire have been selected, the error message "Please make sure to vote for all rating questions ..." is not shown

janno42 (dismissed):
🏋️ Dropping approved

evap/student/tests/test_views.py
```diff
@@ -559,3 +561,109 @@ def test_xmin_of_all_answers_is_updated(self):
...
+            self.evaluation.general_contribution.questionnaires.filter(type=Questionnaire.Type.DROPOUT).exists()
+        )
+
+    def test_view_not_shown_if_dropout_not_allowed(self):
+        normal_questionnaires = self.evaluation.general_contribution.questionnaires.exclude(
+            type=Questionnaire.Type.DROPOUT
+        ).all()
+        self.evaluation.general_contribution.questionnaires.set(normal_questionnaires)
+
+        _ = self.app.get(url=reverse("student:drop", args=[self.evaluation.id]), user=self.user, status=400)
```
richardebeling, line 648: `test_view_not_shown` makes me expect that the test asserts that something isn't on the page, but there is no assertion here, just the 400 status. Can we maybe change the name?

Maybe also remove the `_ =`, we don't do that anywhere else in the codebase

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -59,8 +59,8 @@ <h4 class="{% if not forloop.first %} mt-3 {% elif contributor %} me-auto {% end
...
                                 <div class="vote-inputs tab-row {{ field.field.widget.attrs.choices.css_class }} btn-group" data-bs-toggle="buttons">
                                 {% for choice, color in field|zip:field.field.widget.attrs.choices.colors %}
-                                    <input id="{{ choice.id_for_label }}" class="tab-selectable num-selectable btn-check" name="{{ choice.data.name }}" type="radio" value="{{ choice.data.value }}" autocomplete="off"{% if field.value == choice.data.value %} checked{% endif %}{% if preview %} disabled{% endif %}/>
-                                    <label for="{{ choice.id_for_label }}" class="btn btn-sm btn-light vote-btn vote-btn-{{ color }} d-flex{% if field.value == choice.data.value %} active{% endif %}{% if field.errors %} choice-error{% endif %}">
+                                    <input id="{{ choice.id_for_label }}" class="tab-selectable num-selectable btn-check" name="{{ choice.data.name }}" type="radio" value="{{ choice.data.value }}" autocomplete="off"{% if field.value == choice.data.value|stringformat:"d" %} checked="checked"{% endif %}{% if preview %} disabled{% endif %}/>
```
richardebeling, line 62: can we add / modify a test to ensure these are properly filled when reloading the page after submitting form values that are not accepted by the server?

evap/student/templates/student_index_evaluate_or_drop.html
```diff
@@ -0,0 +1,24 @@
+<div class="d-inline-flex">
+    {% if evaluation.is_dropout_allowed %}
+        <form method="get" action="{% url 'student:drop' evaluation.id %}" class="me-2">
+            <confirmation-modal type="submit" confirm-button-class="btn-danger">
+                <span slot="title">{% translate 'Did you drop this course?' %}</span>
+                <span slot="question">
+                    {% blocktranslate %}
+                        If you mark this course as dropped, you will still be able to provide feedback as usual. However, you
+                        will not be able to submit the questionnaire another time.
```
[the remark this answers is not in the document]
janno42: New suggestion: `If you mark this course as dropped, you will still be able to provide feedback right now. Once you submitted the questionnaire, you can't submit it another time.`

evap/evaluation/models.py
```diff
@@ -828,6 +849,8 @@ def publish(self):
         self._participant_count = self.num_participants
 
         if not self.can_publish_text_results:
+            if self.voters.count() > 1:
+                logger.exception("If there is more than one voter, can_publish_text_results should be true.")
```
richardebeling, line 853: If we ever hit this, it would probably be nice if `self.pk` and probably the name (in case there isn't a pk yet) were included in the log message. Maybe you can just dump `self` in there.

evap/development/fixtures/test_data.json
```diff
@@ -153074,6 +153422,131 @@
...
   }
 },
+{
+  "model": "evaluation.userprofile",
+  "pk": 954,
+  "fields": {
+    "is_superuser": false,
+    "password": "!dOl57Qs91TuNY39aO66wTs5TBRnuBhd3cWbk257M",
+    "last_login": null,
+    "email": null,
```
richardebeling, line 153432: Looks like these are the users you added as voters/participants to evaluation 1673 above.

The problem with these, if generated by model-bakery, is that they will have unrealistically empty fields for email, name, etc. We should have enough users in the test data already.

Can you remove these instances here and instead just pick some of the existing users for participants/voters above for 1673? (Should be solvable in a few minutes by just editing this json file, let us know if you get stuck)

hansegucker (approved):
Nice 🥳

evap/student/tests/test_views.py
```diff
@@ -559,3 +561,136 @@ def test_xmin_of_all_answers_is_updated(self):
...
+
+        form = self.app.get(url=reverse("student:drop", args=[self.evaluation.id]), user=self.user, status=200).forms[
+            "student-vote-form"
+        ]
+        field_id = answer_field_id(
+            self.evaluation.general_contribution, self.dropout_questionnaire, self.dropout_question
+        )
+        form[field_id] = NO_ANSWER  # dropout question must be answered
+        form.submit()
+        self.evaluation = Evaluation.objects.get(pk=self.evaluation.pk)
```
niklasmohrin, line 664:
```suggestion
        self.evaluation.refresh_from_db()
```
(similarly elsewhere)

evap/student/templates/student_vote.html
```diff
@@ -55,6 +55,23 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
+                            You are marking this course as dropped. Your feedback on questions you can answer is still
+                            very welcome.
+                        {% endblocktranslate %}
+                    </p>
+                    {% include 'student_vote_questionnaire_group.html' with questionnaire_group=evaluation_form_group_dropout textanswers_visible_to=general_contribution_textanswers_visible_to preview=preview %}
+                </div>
+            </div>
+        {% endif %}
+
         {% if small_evaluation_size_warning and not preview %}
```
niklasmohrin: Should we not move the dropout questionnaires below this warning? @janno42

evap/student/views.py
```diff
@@ -256,31 +311,37 @@ def render_vote_page(request, evaluation, preview, for_rendering_in_modal=False)
 
 
 @participant_required
-def vote(request, evaluation_id):  # noqa: PLR0912
+def vote(request: HttpRequest, evaluation_id: int, dropout=False) -> HttpResponse:  # noqa: PLR0912
```
niklasmohrin:
```suggestion
def vote(request: HttpRequest, evaluation_id: int, dropout: bool = False) -> HttpResponse:  # noqa: PLR0912
```
otherwise this will still be inferred as Any

niklasmohrin (approved):
Awesome, thanks :)

evap/student/templates/student_vote.html
janno42: Yes, the warning should be moved to the top.

evap/student/tests/test_views.py
fekoch: This results in an `AttributeError: Direct state modification is not allowed`.

janno42 (approved):

evap/student/templates/student_vote.html
```diff
@@ -84,6 +84,23 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
         {% endif %}
 
+        {% if show_dropout_questionnaire %}
+            <div class="card card-outline-warning mb-3">
+                <div class="card-header">
+                    <span class="fas fa-triangle-exclamation"></span> {% translate 'Dropout Questionnaire' %}
+                </div>
+                <div class="card-body">
+                    <p>
+                        {% blocktranslate %}
```
richardebeling:
```suggestion
                        {% blocktranslate trimmed %}
```

evap/student/templates/student_vote.html
```diff
@@ -74,7 +74,7 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
                         </p>
                         <div class="form-check mt-3">
                             <input class="form-check-input tab-selectable" id="text_results_publish_confirmation_top" type="checkbox"
-                                name="text_results_publish_confirmation_top" />
+                                   name="text_results_publish_confirmation_top" />
```
richardebeling: Generally, please try to avoid reformatting existing code that you're not otherwise modifying

```suggestion
                                name="text_results_publish_confirmation_top" />
```

richardebeling (dismissed):

evap/student/tests/test_views.py
niklasmohrin: right

richardebeling (approved):

evap/evaluation/migrations/0149_course_cms_id_evaluation_cms_id.py
niklasmohrin, line 1: huh? you can't just reorder our existing migrations :D

evap/student/templates/student_vote.html
```diff
@@ -84,6 +84,23 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
             </div>
         {% endif %}
 
+        {% if show_dropout_questionnaire %}
```
niklasmohrin: Can't we just do
```suggestion
        {% if evaluation_form_group_dropout %}
```
and remove this variable entirely? Or is there a chance that dropout is enabled, students want to drop out, but we don't assign any dropout questionnaires? No right?

evap/student/tests/test_views.py
```diff
@@ -559,3 +561,136 @@ def test_xmin_of_all_answers_is_updated(self):
...
+        dropout_question_id = answer_field_id(
+            self.evaluation.general_contribution, self.dropout_questionnaire, self.dropout_question
+        )
+
+        self.assertEqual(form[normal_question_id].value, str(NO_ANSWER))
+        self.assertIsNone(form[dropout_question_id].value)
+
+        form[normal_question_id] = -1
+
+        form = form.submit().forms["student-vote-form"]
```
niklasmohrin: just to be sure, this form is returned with errors so we should have the following, right?
```suggestion
        form = form.submit().forms["student-vote-form"]
        self.assertEqual(self.evaluation.dropout_count, 0)
```

niklasmohrin (changes requested):
Last round of edits, everything else looks great :^)

niklasmohrin (approved):
Thanks, great work :)

richardebeling:
🎉🎉🎉

PR #2264, opened 2024-08-12
Reward point redemption events

### LINKED ISSUE #1825: Reward point redemption: Consider using django form or formset
On the reward point redemption page, we have a list of redemption events, and for each of them, the user can select an amount of points to redeem. Upon submission, we manually analyze the POST arguments and do correctness verification the view.

I feel like this logic (building the list of events, validating chosen points per event, storing to database) could be perfectly encapsulated in a form, or maybe a formset.

### LINKED ISSUE #2258: Add configuration option in redemption events allowing redemption of reward points only in certain steps

hansegucker:
Close #2258
Close #1825


evap/rewards/templates/rewards_index.html
```diff
@@ -33,13 +36,14 @@
                                 </tr>
                             </thead>
                             <tbody>
-                                {% for event in events %}
+                                    {% for form, event in forms %}
```
janno42, line 39: indentation error

evap/rewards/forms.py
```diff
@@ -1,14 +1,95 @@
...
+
+        self.fields["points"].validators.append(MaxValueValidator(self.initial["total_points_available"]))
+        self.fields["points"].widget.attrs["max"] = self.initial["total_points_available"]
+
+        self.fields["points"].validators.append(StepValueValidator(self.initial["event"].step))
+        self.fields["points"].widget.attrs["step"] = self.initial["event"].step
+
+        self.fields["points"].help_text = _("multiples of {}").format(self.initial["event"].step)
+
+    def clean(self):
```
janno42, line 41: as discussed, this can probably be removed because both error cases can't occur here

evap/rewards/views.py
```diff
@@ -13,76 +13,54 @@
...
-    return 200
+from evap.rewards.tools import grant_eligible_reward_points_for_semester, redeemed_points_of_user, reward_points_of_user
 
 
 @reward_user_required
 def index(request):
     status = 200
+
+    total_points_available = reward_points_of_user(request.user)
+    events = RewardPointRedemptionEvent.objects.filter(redeem_end_date__gte=datetime.now()).order_by("date")
```
janno42, line 36:
```suggestion
    events = RewardPointRedemptionEvent.objects.filter(redeem_end_date__gte=datetime.now().date).order_by("date")
```

evap/rewards/forms.py
```diff
@@ -1,14 +1,95 @@
...
+        if points % event.step != 0:
+            raise ValidationError(
+                {"points": _("You can only redeem multiples of {} for this event.").format(event.step)}
+            )
+
+        if event.redeem_end_date < date.today():
+            raise ValidationError({"points": _("Sorry, the deadline for this event expired already.")})
+
+
+class BaseRedeemRewardPointsFormSet(forms.BaseFormSet):
```
janno42, line 57:
```suggestion
class BaseRewardPointRedemptionFormSet(forms.BaseFormSet):
```

janno42 (changes requested):

janno42 (approved):
✔️ Meets requirements
✔️ UI functionality checked

evap/rewards/views.py
```diff
@@ -13,76 +13,54 @@
...
     status = 200
+
+    total_points_available = reward_points_of_user(request.user)
+    events = RewardPointRedemptionEvent.objects.filter(redeem_end_date__gte=datetime.now().date()).order_by("date")
+
+    formset = RewardPointRedemptionFormSet(
+        request.POST or None,
+        initial=[{"event": e, "points": 0, "total_points_available": total_points_available} for e in events],
+        user=request.user,
+    )
```
richardebeling, line 42: problematic: `total_points_available` is computed outside of the transaction/locked critical section, formset `clean` is computed without database locking, and `save` might run arbitrarily later -- with the possibility that the same points have been redeemed in between.

For this to be safe, we need to ensure that during the update path, all three steps (calculate limit, clean, save) are stored in the transaction while the rows are locked.

richardebeling (changes requested):

evap/rewards/tests/test_views.py
```diff
@@ -50,81 +50,69 @@ def setUpTestData(cls):
...
 
     def test_redeem_too_many_points(self):
         response = self.app.get(self.url, user=self.student)
         form = response.forms["reward-redemption-form"]
-        form.set(f"points-{self.event1.pk}", 3)
-        form.set(f"points-{self.event2.pk}", 3)
-        response = form.submit(status=400)
-        self.assertContains(response, "have enough reward points.", status_code=400)
+        form.set("form-0-points", 3)
+        form.set("form-1-points", 3)
```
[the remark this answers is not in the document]
hansegucker: There is always a corresponding form field like `form-0-pk` that does the mapping.

richardebeling (approved):
thanks!

Kakadus (approved):

evap/rewards/views.py
```diff
@@ -13,76 +13,58 @@
...
+
+    events = RewardPointRedemptionEvent.objects.filter(redeem_end_date__gte=datetime.now().date()).order_by("date")
+
+    # pylint: disable=unexpected-keyword-arg
+    formset = RewardPointRedemptionFormSet(
+        request.POST or None,
+        initial=[{"event": e, "points": 0} for e in events],
+        user=request.user,
+    )
+    with formset.lock():
```
niklasmohrin, line 43: This seems a bit fragile, I would much rather have something where `formset` is created inside a `with lock_for_reward_changes(request.user)` and never escapes it, but I see that this conflicts with the Django form workflow. Thoughts?

I now wonder whether `.lock()` is the best name here, because the formset is not being locked, but rather we touch some rows in the database to include them in the transaction that we have made atomic. From my current understanding, `formset.atomic()` would better fit this. What do you think?

niklasmohrin (commented):
Very nice! I have some final thoughts that I want to clear up before I approve :)

evap/rewards/forms.py
```diff
@@ -1,14 +1,108 @@
...
+    event = forms.ModelChoiceField(queryset=RewardPointRedemptionEvent.objects.all(), widget=forms.HiddenInput())
+    points = forms.IntegerField(min_value=0, label="")
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if not self.initial:
+            return
+
+        self.fields["points"].validators.append(MaxValueValidator(self.initial["total_points_available"]))
+        self.fields["points"].widget.attrs["max"] = self.initial["total_points_available"]
```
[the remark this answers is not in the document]
richardebeling: Also bothered me. iirc, we need the validator to get the correct drop-down choices in the frontend.

I wouldn't want to extensively test django validation. We can check that it is configured correctly (max value via the validator, min value via the form vield), but two simple assertions should be enough for that, I wouldn't start checking cross product stuff.

We can add an assertion to "our" validation logic that none of the values are negative.

evap/rewards/views.py
richardebeling: > where formset is created inside a with lock_for_reward_changes(request.user) and never escapes it, but I see that this conflicts with the Django form workflow.

We can have the context manager be useable as a class thing instead of a instance thing, have it notify its locking globally. I don't see why this wouldn't work. Don't know if its more elegant, though.

No opinion on the name. `lock`, `assert`, `critical_section`, whatever, all fine with me. I think "context within which we validate correctness of requested redemptions and apply them" is too long, so I don't see a name that wouldn't require people to look up what exactly it does when modifying the code.

evap/rewards/forms.py
```diff
@@ -1,14 +1,108 @@
...
+        self.user = user
+        super().__init__(*args, **kwargs)
+        self.locked = False
+
+    def get_form_kwargs(self, index):
+        kwargs = super().get_form_kwargs(index)
+        if not self.initial:
+            return kwargs
+        kwargs["initial"] = self.initial[index]
+        kwargs["initial"]["total_points_available"] = reward_points_of_user(self.user)
```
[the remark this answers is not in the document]
richardebeling: Doesn't need to be inside the critical section. This only ends up in the `MaxValueValidator` in `RewardPointRedemptionForm`, which is only there for frontend/usability. For (security) correctness, only the `total_points_available = reward_points_of_user(self.user)` inside `BaseRewardPointRedemptionFormSet.clean` is relevant.

evap/rewards/forms.py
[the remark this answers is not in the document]
niklasmohrin: I agree, but I mean it's just three cases, I think that it is fair to do (I think that all three are interesting enough). In particular the case that the sum is bound, but the individual values are not, seems important to me.

evap/rewards/forms.py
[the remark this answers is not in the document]
niklasmohrin: But just to be sure, these validators also run later when handling the POST data right? Just with maybe outdated bounds

evap/rewards/views.py
niklasmohrin: > We can have the context manager be useable as a class thing instead of a instance thing, have it notify its locking globally.

I think I would prefer that, it would probably look okay with the form escaping the with-block:
```python
with lock_for_reward_points_upgrade(request.user):
    formset = ...
    if ...:
        formset.save()
        return ...

return render(..., formset)
```

niklasmohrin: @hansegucker pointed out that this would make the assertions in `clean` and `save` more awkward - I didn't know that these were a requirement, it's fair though.

I now thought that maybe a good way to do it would be something like
```python
class TheFormset:
    @classmethod
    @contextmanager
    @transaction.atomic
    def make_in_transaction(cls, *args, **kwargs):
        list(...) # lock the stuff
        instance = cls(*args, **kwargs)
        instance.locked = True
        try:
            yield instance
        finally:
            instance.locked = False
```

evap/rewards/forms.py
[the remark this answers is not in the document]
richardebeling: It's just a django form field validator -- yes, it better be validated when calling `is_valid`.

The reward points being outdated could theoretically be an issue if a user was awarded points in the meantime. The `MaxValueValidator` allows providing a callable instead of a value. Maybe that allows shifting the computation to a time inside the locked phase -- but for just rendering the page, we never lock any rows anyway...

evap/rewards/forms.py
richardebeling, line 36: @hansegucker where does "max" come from, and why do we need a separate `MaxValueValidator`? In the [django docs on IntegerField](https://docs.djangoproject.com/en/5.1/ref/forms/fields/#django.forms.IntegerField), it sounds to me as if the field automatically does min- and max value validation if the values are set up, but it says that this should be done as `min_value` and `max_value`. We use `min_value` above, why not just `max_value` here?

Same for the StepValidator below, the IntegerField is supposed to have support for `step_size`.

richardebeling (changes requested):

evap/rewards/views.py
richardebeling: > @hansegucker pointed out that this would make the assertions in clean and save more awkward - I didn't know that these were a requirement, it's fair though.

Why? I'd have thought we just set and unset a global/class variable, instead of an instance variable.

> I now thought that maybe a good way to do it would be something like

Also fine with your approach

evap/rewards/forms.py
```diff
@@ -1,14 +1,108 @@
...
+            self.locked = False
+
+    def clean(self):
+        assert self.locked
+
+        if any(self.errors):
+            return
+
+        total_points_available = reward_points_of_user(self.user)
+        total_points_redeemed = sum(form.cleaned_data["points"] for form in self.forms)
```
richardebeling, line 83: To make sure the comment from the other thread isn't lost: I propose to add
```suggestion
        total_points_redeemed = sum(form.cleaned_data["points"] for form in self.forms)
        assert all(form.cleaned_data["points"] > 0 for form in self.forms)
```
here

hansegucker: We allow points to be zero if the user doesn't want to redeem points for an event. For values below 0, we validate the IntegerField with min_value=0. So, this additional check seems to be superfluous in my opinion.

evap/rewards/forms.py
hansegucker: We have to differentiate on form fields and form widgets. max is a form widget attribute, max_value is a form field attribute. At this time, the form is already initialized and changing max_value doesn't have any effect anymore.

evap/rewards/forms.py
richardebeling: > We allow points to be zero if the user doesn't want to redeem points for an event.

Fair, I'm proposing `form.cleaned_data["points"] >= 0` then

> For values below 0, we validate the IntegerField with min_value=0. So, this additional check seems to be superfluous in my opinion.

It is relevant for security that >= 0 is validated somewhere. The contract of this snippet is "must have been validated before". Adding an assert would be a run-time checked codification of this contract. I don't regard this as validation, we'd still get an internal server error if it ever doesn't hold. In security-relevant questions, I strongly think we should do these kinds of contract-checks.

evap/rewards/forms.py
richardebeling: Can we just do
```python3
self.fields['year'] = forms.IntegerField(
    min_value=0,
    max_value=self.initial["total_points_available"],
    step_size=self.initial["event"].step,
    label=""
)
```
and remove `points = forms.IntegerField(min_value=0, label="")` above? I think this would simplify things

richardebeling (approved):

niklasmohrin (approved):
Thanks!

richardebeling (approved):

PR #2267, opened 2024-08-12
Pin django to minor version

richardebeling:
Fixup for #2084, where dependabot (without asking) changed our pinning from "any patch" to "any minor version".

Should fix current CI failures.

The [official PyPI package](https://pypi.org/project/Django/) is named `Django` with a capital D, I also fixed the file to reflect that.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 django-extensions==3.2.3
 django-fsm==2.8.2
-django~=5.0
+Django~=5.0.0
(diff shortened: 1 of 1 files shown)


niklasmohrin (approved):
Not sure if you want to mention the issue in the commit message, if I recall correctly, this always results in a lot of pings when people update their forks and therefore push commits mentioning the issue again - not sure if that is still a problem though

PR #2276, opened 2024-09-05
Change icon-yellow to icon-blue

janno42:
The filter buttons could get a complete overhaul in the future, this is a quick fix upon individual manager request.


richardebeling (approved):
I′m blue da ba dee da ba di

niklasmohrin (approved):
I would have thought that we would be able to query for a theming preference from the browser (like with light / dark theme), but it seems there is no such feature. I found this discussion which gives some insight: https://discourse.wicg.io/t/user-set-media-types-e-g-to-detect-color-blind-users/946/

Anyways, good change!

PR #2281, opened 2024-09-16
Update to bootstrap 5.3.3

richardebeling:
fixes the deprecation warning we're currently getting with sass

The "current" version of [@types/bootstrap](https://www.npmjs.com/package/@types/bootstrap) is 5.2.10, so cannot (yet) put "^5.3.0" there.


janno42 (changes requested):
the only problems i noticed are these:
- hovering over hover rows doesn't change the background color
- active navbar items (when opening dropdowns) have black text color

janno42 (approved):

PR #2285, opened 2024-09-23
Add eslint

Kakadus:
Enables just recommended configs from eslint, strict typecheck and stylistic configs from tseslint.

This will not run in ci because it would just fail. Intended for local checks and Codacy: Codacy should pick this up and only check the diff

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -58,10 +58,17 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/setup_evap
+        with:
+          npm-ci: 'true'
+
       - name: Run ruff
         run: ruff check .
       - name: Run pylint
         run: pylint evap tools
+      - name: Run ESLint
+        run: |
+          cd evap/static/ts
+          npx eslint --quiet
 
   formatter:
     runs-on: ubuntu-22.04
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -68,7 +68,7 @@ jobs:
       - name: Run ESLint
         run: |
           cd evap/static/ts
-          npx eslint --quiet
+          npx eslint
 
   formatter:
     runs-on: ubuntu-22.04
--- a/evap/static/ts/eslint.config.js
+++ b/evap/static/ts/eslint.config.js
@@ -49,7 +49,6 @@ export default tseslint.config(
                 },
             ],
             "@typescript-eslint/no-confusing-void-expression": "warn",
-            "@typescript-eslint/no-floating-promises": "warn",
             "@typescript-eslint/no-unsafe-return": "warn",
             "@typescript-eslint/no-redundant-type-constituents": "warn",
         },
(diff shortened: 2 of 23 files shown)


richardebeling (changes requested):
as discussed: If we provide a config, it should run and pass in CI

evap/static/ts/src/quick-review-slider.ts
```diff
@@ -447,7 +451,9 @@ export class QuickReviewSlider {
             // because the element has zero opacity until we add `.active`.
             nextActiveElement.classList.remove("to-left", "to-right");
             nextActiveElement.classList.add(`to-${direction}`);
-            requestAnimationFrame(() => nextActiveElement.classList.add("active"));
+            requestAnimationFrame(() => {
+                nextActiveElement.classList.add("active");
+            });
```
richardebeling, line 456: Personally, I'd be fine with allowing single statements without an explicit block, but I don't care too much.

richardebeling (approved):

evap/static/ts/src/quick-review-slider.ts
Kakadus: we can allow the "ArrowShorthand", I think this makes definitely sense :+1:

Kakadus: I've dropped the commit in question and further enabled a rule that forces using the arrow shorthand "as-needed" (enforces no braces where they can be omitted). Might be too restrictive though :man_shrugging:

richardebeling (approved):

evap/static/ts/eslint.config.js
```diff
@@ -0,0 +1,54 @@
...
+export default tseslint.config(
+    eslint.configs.recommended,
+    ...tseslint.configs.strictTypeChecked,
+    ...tseslint.configs.stylisticTypeChecked,
+    {
+        ignores: ["rendered/", "eslint.config.js"],
+    },
+    {
+        languageOptions: {
+            ecmaVersion: 5,
```
niklasmohrin, line 15: Don't we use features from a newer version already? Does anything break when using a more recent version?

niklasmohrin (approved):
Very cool!

Kakadus:
(just rebased without code changes)

evap/static/ts/eslint.config.js
Kakadus: actually, the default is "latest" and I don't think there is a reason why I set this in the first place. I can also set it to 2019 as that is our es target specified in `tsconfig.json`?

.github/workflows/tests.yml
```diff
@@ -58,10 +58,17 @@ jobs:
...
+          npm-ci: 'true'
+
       - name: Run ruff
         run: ruff check .
       - name: Run pylint
         run: pylint evap tools
+      - name: Run ESLint
+        run: |
+          cd evap/static/ts
+          npx eslint --quiet
```
niklasmohrin, line 71: I think we can show warnings in CI - if the job still passes, the message to the contributors should be clear

niklasmohrin (approved):
Cool!

evap/static/ts/eslint.config.js
```diff
@@ -0,0 +1,54 @@
...
+    {
+        languageOptions: {
+            ecmaVersion: 2019,
+            parserOptions: {
+                project: ["tsconfig.json", "tsconfig.compile.json"],
+            },
+            sourceType: "module",
+        },
+    },
+    // ignore @typescript-eslint/no-misused-promises
```
Kakadus, line 22:
```suggestion
```

PR #2309, opened 2024-10-14
Reset translation in test tearDown

richardebeling:
Fixes order-dependent tests (as caused failure in https://github.com/e-valuation/EvaP/actions/runs/11331145512/job/31510516990)

### DIFF
--- a/evap/contributor/tests/test_forms.py
+++ b/evap/contributor/tests/test_forms.py
@@ -1,10 +1,9 @@
 from django.forms.models import inlineformset_factory
-from django.test import TestCase
 from model_bakery import baker
 
 from evap.contributor.forms import EditorContributionForm, EvaluationForm
 from evap.evaluation.models import Contribution, Evaluation, Program, Questionnaire, UserProfile
-from evap.evaluation.tests.tools import WebTest, get_form_data_from_instance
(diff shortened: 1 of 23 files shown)


Kakadus (approved):

niklasmohrin:
[The second failure](https://github.com/e-valuation/EvaP/pull/2274#issuecomment-2412508545) with seed 7524442357 still breaks on my machine, now even with two tests failing. Maybe we can use https://github.com/asottile/detect-test-pollution ?

richardebeling:
> [The second failure](https://github.com/e-valuation/EvaP/pull/2274#issuecomment-2412508545) with seed 7524442357 still breaks on my machine, now even with two tests failing.

The problem was that `django.test.TestCase.tearDown` doesn't call `super.tearDown()`, so with inheritance order `class TestCase(django.test.TestCase, MyTearDownMixin)`, the `tearDown` method of the mixin is never called. Fixed it by swapping the order. `./manage.py test --keepdb --shuffle 7524442357` now works for me.

niklasmohrin (approved):

niklasmohrin:
Another seed fixed by this PR 1773734012 :)

PR #2314, opened 2024-10-14
Add `packages.poetry`

niklasmohrin:
Use for example as `nix run .#poetry -- lock`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -26,7 +26,7 @@
         in
         rec {
           evap = pkgs.callPackage ./nix/shell.nix {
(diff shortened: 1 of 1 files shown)


Kakadus (approved):
Somewhy I had to fetch rust and build cryptography on my own, butt then it worked

niklasmohrin:
Yeah I also noticed that, but it's because the Python packages are not cached for the Python version we use here. If we were to always use the latest stable Python from nix packages, we would just get the final poetry package from cache

richardebeling (approved):

PR #2317, opened 2024-10-21
Add python wrapper for IDEs

Kakadus:
This might come in handy for IDEs insisting on a single python executable path for setting the python interpreter. So far, tested with JetBrains' PyCharm on Windows and Linux.

For PyCharm, one can [select](https://www.jetbrains.com/help/pycharm/configuring-python-interpreter.html) this as an existing interpreter to get full python hightlighing, introspection and documentations. Just tested with PyCharm 2024.2.3 (#JBC-242.23339.19 on wsl 2.3.24.0). For VSCode, this might come in handy as well.
For PyCharm, one can set the "terminal shell path" (search for it) to `nix develop` to instantly start in a development shell. I expect there to be a similar config for VSCode.
Maybe this is worth a comment in the wiki? However, this just works around missing IDE features.

Adapted https://intellij-support.jetbrains.com/hc/en-us/community/posts/360008227939/comments/9528610214802 for `nix develop` and inlined the `PYTHONCMD` to prevent unexpected parameter unpacking.

### DIFF
--- a/nix/nix-python
+++ b/nix/nix-python
@@ -1,3 +1,3 @@
 #!/bin/sh
-SCRIPTPATH="$( cd -- "$(dirname "$0")"/.. >/dev/null 2>&1 || exit 1 ; pwd -P )"
+SCRIPTPATH="$(realpath "$(dirname "$0")"/.. )"
 exec nix develop "$SCRIPTPATH" --command python "$@"
--- a/nix/nix-python
+++ b/nix/nix-python
@@ -1,0 +1,3 @@
+#!/bin/sh
+SCRIPTPATH="$( cd -- "$(dirname "$0")"/.. >/dev/null 2>&1 || exit 1 ; pwd -P )"
+exec nix develop "$SCRIPTPATH" --command python "$@"
--- a/nix/nix-python
+++ b/nix/nix-python
@@ -1,3 +1,4 @@
 #!/bin/sh
-SCRIPTPATH="$(realpath "$(dirname "$0")"/.. )"
-exec nix develop "$SCRIPTPATH" --command python "$@"
+SCRIPTPATH=$(dirname "$0")
+FLAKEPATH=$(realpath "$SCRIPTPATH/..")
+exec nix develop "$FLAKEPATH" --command python "$@"


niklasmohrin:
I would see this more in the wiki; additionally, what's up with with the cd-and-pwd approach, wouldn't `realpath` work?

Kakadus:
`realpath` looks good AND works :scream:

Kakadus:
Just a sidenote: We have three people here (except from me) which is using this (or a similar) workaround for their IDEs. What does you @richardebeling, @janno42 think? Even with this file, manual configuration is still needed for the IDEs.

richardebeling:
> Just a sidenote: We have three people here (except from me) which is using this (or a similar) workaround for their IDEs. What does you @richardebeling, @janno42 think? Even with this file, manual configuration is still needed for the IDEs.

I'm fine with merging a small helper script if it helps setting up PyCharm or VSCode. In case you were asking for that: I don't think I've ever had my language server configured for venv/VM access, so I'm likely not impacted.

nix/nix-python
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+SCRIPTPATH="$(realpath "$(dirname "$0")"/.. )"
```
niklasmohrin, line 2: I would think that we need to escape some of the `"` here (or use `'` instead). Does something like
```suggestion
SCRIPTPATH="$(dirname \"$0\")"
FLAKEPATH="$(realpath \"$SCRIPTPATH/..\")"
```
work, or am I messing up bash here?

Kakadus: I think it works because its inside a subshell `$(...)` (shellcheck+my sh likes it).
+1 for having two lines so that we don't have to think about quoting

Kakadus: To quote the good old POSIX standard, as to why the initial code does not need escaping the `"`: 

> The input characters within the quoted string that are also enclosed between "$(" and the matching ')' shall not be affected by the double-quotes, but rather shall define that command whose output replaces the "$(...)" when the word is expanded

(https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html, 2.2.3, paragraph 3)

richardebeling: > To quote the good old POSIX standard

power move, good one! :D

niklasmohrin: What @Kakadus didn't mention is the two of us spending 10 minutes on understanding the wording of the standard :D

richardebeling (approved):

niklasmohrin (approved):

PR #2283, opened 2024-09-16
Add result filter slider

Kakadus:
close https://github.com/e-valuation/EvaP/issues/1975

PR #2318, opened 2024-10-21
Fixup result filter slider 

Kakadus:
Fixup to address comment by @niklasmohrin in #2283 

edit: https://github.com/e-valuation/EvaP/pull/2283#discussion_r1799952398 (for @janno42)

### DIFF
--- a/evap/static/ts/src/slider.ts
+++ b/evap/static/ts/src/slider.ts
@@ -16,7 +16,7 @@ export class RangeSlider {
     private readonly minLabel: HTMLSpanElement;
     private readonly rangeLabel: HTMLSpanElement;
     private allowed: Range = { low: 0, high: 0 };
-    private selection: Range = { low: 0, high: 0 };
+    private _selection: Range = { low: 0, high: 0 };
 
     private debounceTimeout?: number;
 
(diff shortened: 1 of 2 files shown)


richardebeling (approved):

janno42 (approved):

evap/static/ts/src/slider.ts
```diff
@@ -15,10 +15,8 @@ export class RangeSlider {
     private readonly maxLabel: HTMLSpanElement;
     private readonly minLabel: HTMLSpanElement;
     private readonly rangeLabel: HTMLSpanElement;
-    private readonly min = 0;
-    private max = 0;
-    private low = 0;
-    private high = 0;
+    private allowed: Range = { low: 0, high: 0 };
+    private _selection: Range = { low: 0, high: 0 };
```
niklasmohrin, line 19: I now wonder whether `value` would be a better name, to be consistent with other inputs :thinking:

niklasmohrin (approved):

evap/static/ts/src/slider.ts
```diff
@@ -31,27 +29,26 @@ export class RangeSlider {
         this.rangeLabel = selectOrError<HTMLSpanElement>(".range-values", this.rangeSlider);
 
         const setRangeFromSlider = (): void => {
```
niklasmohrin, line 31:
```suggestion
        const setValueFromSlider = (): void => {
```

evap/static/ts/src/slider.ts
```diff
@@ -65,25 +62,25 @@ export class RangeSlider {
...
             this.reset();
         }
     }
 
     public reset(): void {
-        this.range = { low: this.min, high: this.max };
+        this.value = { low: this.allowed.low, high: this.allowed.high };
     }
 
     private updateLimits(): void {
```
niklasmohrin, line 78: Is this more fitting?
```suggestion
    private updateNestedElements(): void {
```
It might be misleading because we don't update the `value` attribute of the nested elements

niklasmohrin (commented):
Some final remarks, feel free to disregard though

evap/static/ts/src/slider.ts
Kakadus: I just noticed, that I used it only once, so if the naming is misleading, it might be better to just inline this function into `includeValues`?

niklasmohrin: hm, I think it's better with `updateNestedElements`

niklasmohrin (approved):

PR #2321, opened 2024-10-28
MacOS nix fixes

niklasmohrin:
Mac users reported issues that boiled down to `psycopg[binary]` not having a matching wheel. I thought that using the "pure Python installation" (just `psycopg`) would work, but this breaks even on Linux as the libpq dynamic library cannot be found (even when adding postgres to the nix shell packages or the build inputs of the Python packages). There is also `psycopg[c]` which compiles the extension at installation time. This seems to work and the compilation time is not too long either (on a Surface from this decade it took ~1min to compile the extension, same in a Mac VM).

Should we just default to using `psycopg[c]` on Linux too, or stick to using the binary installation on Linux?

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -23,6 +23,7 @@
       devShells = forAllSystems (system:
         let
           pkgs = pkgsFor.${system};
+          extras = if pkgs.stdenv.isDarwin then [ "psycopg-c" ] else [ "psycopg-binary" ];
         in
         rec {
           evap = pkgs.callPackage ./nix/shell.nix {
@@ -30,6 +31,7 @@
             poetry2nix = inputs.poetry2nix.lib.mkPoetry2Nix { inherit pkgs; };
             pyproject = ./pyproject.toml;
             poetrylock = ./poetry.lock;
+            inherit extras;
           };
           evap-dev = evap.override { poetry-groups = [ "dev" ]; };
           default = evap-dev;
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,9 @@ django-fsm = "~2.8.2"
 Django = "~5.1.0"
 mozilla-django-oidc = "~4.0.1"
 openpyxl = "~3.1.5"
-psycopg = { version = "~3.2.3", extras = ["binary"] }
+psycopg = "~3.2.3"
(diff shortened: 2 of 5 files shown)


Kakadus:
just a sidenote: running `pip install "psycopg[binary]"` on macos works, the same is when just installing poetry, so this seems to be a poetry2nix problem.

IMHO, using `psycopg[c]` is fine on all platforms.

flake.nix
Kakadus, line 1: just want to note that `./manage.py test` still fails with an import error for psycopg.

richardebeling (approved):
Compiling is fine for me as a workaround, but in general, I don't want people to compile stuff, I want to use internet-provided builds to not waste time and energy.

> running pip install "psycopg[binary]" on macos works, the same is when just installing poetry, so this seems to be a poetry2nix problem.

Is there a chance we could get this fixed in poetry2nix, does it make sense to open an issue somewhere (@niklasmohrin)?

niklasmohrin:
If I understood correctly it's either a mismatch between what nix and python (and different Python libraries) use as a name for arm macs, or the wheels are just missing. That being said, I recently read that nixpkgs updated and refactored the apple sdks,  so I suppose I will look into this again before moving forward here

niklasmohrin:
@ybrnr Can you do one last check of the following?
```
nix run --refresh github:niklasmohrin/evap/psycopg-non-binary#services-full
nix develop --refresh github:niklasmohrin/evap/psycopg-non-binary
```

niklasmohrin:
If this works, I think we will just keep it as is, because we should probably move away from poetry2nix anyways, as the maintainer stepped down a couple of days ago :/

pyproject.toml
```diff
@@ -12,7 +12,9 @@ django-fsm = "~2.8.2"
 Django = "~5.1.0"
 mozilla-django-oidc = "~4.0.1"
 openpyxl = "~3.1.5"
-psycopg = { version = "~3.2.3", extras = ["binary"] }
+psycopg = { version = "~3.2.3", extras = [] }
```
Kakadus, line 15:
```suggestion
psycopg = "~3.2.3"
```

Kakadus (approved):
also works with macos on x64

PR #2328, opened 2024-11-04
Publish EvaP to PyPI

### LINKED ISSUE #2303: "Official" production setup / Nix usage / Migration Plan
Follow-Up to #2274: Do we also want production instances to be set-up using nix? It might simplify stuff. We may be able to test on a dev server.

* 2274 removed the requirements text files, so venv management might currently need to be done with poetry (although poetry can, according to the internet, produce a requirements file)
* We should make sure that the `update_production.sh` and `load_production_backup.sh` scripts match the envisioned setup.
    * Production update currently doesn't necessarily work, due to commented/missing pip package update after switching to new version
* We might need a plan for migration from a native installation to a nix installation
* We maybe should document why we trust the software we're using.

niklasmohrin:
Closes #2303 

Since we don't want to use nix in production (for example because security updates might take a long time to propagate through nixpkgs / through our github release cycle), `pip install .` or `pip install evap` seems to be the most canonical way. This allows us to install security patches via `pip install --upgrade evap`.

If we really need to override a dependency version beyond the allowed range in `pyproject.toml`, we can install EvaP from sources outside of PyPI. The script to build the release artifacts is available as `nix run .#build-dist`. It is possible to run this on a development machine and copy the `.whl` file over to the server.

The release wheel includes compiled TypeScript, SCSS and translation files, so there is no need to install NodeJS or gettext on the production server.

The new release workflow is:
1. Add translations and commit them.
2. Bump version in `pyproject.toml`, run `nix develop .#impure --command uv lock`, commit it, tag it, push it.
3. After the CI is done on main, dispatch a release workflow run to publish the wheel to PyPI.
4. Create Github Release.

The versioning scheme we agreed on is year-month-patch, for example "2024.11.0". I suppose we should take care that we don't necessarily push breaking changes like updating the Python version in a new patch. Following the Python Packaging User Guide, we don't include leading zeros in release versions like "2025.4.0".

Since upgrading to the next stable revision of EvaP is handled by pip, there is no need for the `release` branch anymore.

See
- https://packaging.python.org/en/latest/tutorials/packaging-projects/
- https://docs.pypi.org/trusted-publishers/using-a-publisher/
- https://packaging.python.org/en/latest/specifications/version-specifiers/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "evap"
 description = "EvaP"
-version = "0.0.0"
+version = "2025.04.0"
 readme = "README.md"
 requires-python = "~=3.11.0"
 dependencies = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "evap"
 description = "EvaP"
-version = "2025.04.0rc1"
+version = "2025.4.0rc1"
 readme = "README.md"
 requires-python = "~=3.11.0"
 dependencies = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "evap"
 description = "EvaP"
-version = "2025.04.0"
+version = "2025.04.0rc1"
 readme = "README.md"
 requires-python = "~=3.11.0"
 dependencies = [
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,26 @@
+name: EvaP Release
+
+on:
+  workflow_dispatch:
+
+jobs:
+  pypi-publish:
+    name: upload release to PyPI
+    runs-on: ubuntu-latest
+    environment: release
+    permissions:
+      id-token: write
+    steps:
+      - run: |
+          RUN_ID=$(gh run --repo ${{ github.repository }} list --commit ${{ github.sha }} --status success --workflow "EvaP Test Suite" --json databaseId --jq '.[].databaseId')
+          if [ -z "$RUN_ID" ]; then
+            echo "No run found"
+            exit 1
+          fi
+          gh run --repo ${{ github.repository }} download "$RUN_ID" --name wheel --dir dist
+        env:
+          GH_TOKEN: ${{ github.token }}
+      - run: tar tvf dist/*.tar.gz
+      - run: unzip -l dist/*.whl
+      - name: Publish package distributions to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - main
   pull_request:
+  workflow_dispatch:
 
 jobs:
   test:
--- a/uv.lock
+++ b/uv.lock
@@ -329,7 +329,7 @@ wheels = [
 
 [[package]]
 name = "evap"
-version = "2025.4.0"
+version = "2025.4.0rc1"
 source = { editable = "." }
 dependencies = [
     { name = "django" },
--- a/uv.lock
+++ b/uv.lock
@@ -329,7 +329,7 @@ wheels = [
 
 [[package]]
 name = "evap"
-version = "0.0.0"
+version = "2025.4.0"
 source = { editable = "." }
 dependencies = [
     { name = "django" },


niklasmohrin:
See: https://test.pypi.org/project/evap/

niklasmohrin:
@richardebeling Thoughts? Furthermore, if you create a PyPI account, I can add you as a co-owner of the package. (I also requested the creation of an e-valuation organization on PyPI, not sure if it will be nicer / needed, we'll see once it's processed)

pyproject.toml
```diff
@@ -1,7 +1,8 @@
 [tool.poetry]
-package-mode = false
 name = "evap"
 description = "EvaP"
+authors = [ "EvaP Developers" ]
```
niklasmohrin, line 4: Should we add the organization members here? Right now, PyPI displays "EvaP Developers" under unverified details, I suppose we could also just add the names which will hopefully be deduplicated then

niklasmohrin: Looks like it won't be deduplicated, I suppose no one really cares about this, see https://test.pypi.org/project/evap/2024.10.0/ or https://pypi.org/project/black/

niklasmohrin:
One thing we discussed in person was that we now don't really need a full checkout of the EvaP repository on production. In fact, this is a chance to move most of the (apache-)deployment files out of the main repository and into a separate repository in the e-valuation organization. @janno42 agreed that this would be a nice (while not _really_ needed) change. The only drawback seems to be that testing the backup process in the EvaP repository will be more complicated. Currently, I believe that we would want to pin a commit of the deployment repository, for example through flake.lock, and run the tests here. I believe that we could add the backup process test logic through nix in the deployment repository and then import this test in the EvaP repository to use the updated source of EvaP. In particular, this would allow us to share the test definition and run the "easily" run the same tests in both repos. I am not too settled on this approach though.

Thoughts @richardebeling ?

pyproject.toml
richardebeling: I must be blind -- what section exactly would contain duplicated names? I see 5 unique user names on the black project page under "Maintainers"

I'd be fine with just putting "EvaP Developers", which will always remain correct.

Do we even need this in the pyproject.toml? I would expect that for username mapping, you need to configure this on PyPI separately anyway?

niklasmohrin: "Ambv" is the same person as the unverified "author" (=> big projects don't care about about unverified data) and also on the evap release my name appears twice (=> it doesn't get deduplicated)

richardebeling:
> @richardebeling Thoughts? Furthermore, if you create a PyPI account, I can add you as a co-owner of the package. (I also requested the creation of an e-valuation organization on PyPI, not sure if it will be nicer / needed, we'll see once it's processed)

While certainly unexpected, I think it's a funny and efficient solution, and I don't have any reasons against doing this.

> One thing we discussed in person was that we now don't really need a full checkout of the EvaP repository on production. In fact, this is a chance to move most of the (apache-)deployment files out of the main repository and into a separate repository in the e-valuation organization. janno42 agreed that this would be a nice (while not really needed) change. The only drawback seems to be that **testing the backup process in the EvaP repository will be more complicated.**

Why? "Backup" essentially is "Dump database via `dumpdata`", isn't it? This would be separate from apache and all the other deployment aspects. I can see that the "Upgrade" part of the script wouldn't be testable -- we can compromise on that, I think. It's important that backup and especially restore is tested.

---

Re (separate repo, version pinning): Adds complexity, but if there is a need for this or it helps in any way, I guess we won't die 🤷‍♂️

A proper release procedure with some github magic action could also be used to deliver compiled TS and SCSS files automatically, something that I dislike a bit about our current approach (necessity to have npm somewhere)

pyproject.toml
richardebeling: Ah, I see -- I was looking for duplication inside one list. Yeah, "evap developers" 👍

evap/settings.py
```diff
@@ -349,17 +350,17 @@ class ManifestStaticFilesStorageWithJsReplacement(ManifestStaticFilesStorage):
 
 # Additional locations of static files
 STATICFILES_DIRS = [
-    os.path.join(BASE_DIR, "static"),
+    MODULE / "static",
 ]
 
 # Absolute path to the directory static files should be collected to.
-STATIC_ROOT = os.path.join(BASE_DIR, "static_collected")
+STATIC_ROOT = DATADIR / "static_collected"
```
niklasmohrin, line 357: Since we are already setting the variable here, it might be counterintuitive if someones does
```python
from evap.settings import *

DATADIR = Path("...")
```
Do we care (now)?

niklasmohrin:
I updated the PR description to fit the current state of things. I tested the currently published release with the following suggested setup in a podman container (`podman run --rm -p 8888:80 -it ubuntu`):
```bash
```
apt update && apt upgrade -y && apt install -y sudo redis postgresql apache2 python3 python3-venv libapache2-mod-wsgi-py3

service redis-server start
service postgresql start
service apache2 start

sudo -u postgres psql
create user evap password 'evap' createdb;
create database evap owner evap;
exit

useradd -m evap
sudo -i -u evap
mkdir evap && cd evap
python3 -m venv venv
. venv/bin/activate
pip install evap[psycopg-binary]
# set up productionsettings.py and wsgi.py with template
python -m evap collectstatic
python -m evap migrate
# collectmessages not needed, already included in release

exit # venv
exit # sudo -i
chown -R evap:www-data /home/evap

cd /etc/apache2
# set up sites-available/100-evap.conf with template, replace /evap with /home/evap/evap
a2enmod headers
a2enmod wsgi
a2dissite 000-default
a2ensite 100-evap
service apache2 restart
```
Afterwards, evap should be available at the given port. To log in, one should also run `python -m evap reload_testdata` after `collectstatic`.

**Comment by niklasmohrin** [comment]
We made https://github.com/e-valuation/evap-deployment as a place to move the deployment files, because production will not have a checkout of the EvaP repository anymore

File: evap/development/management/commands/dump_testdata.py [hunk 3]
```diff
@@ -12,7 +12,7 @@ class Command(BaseCommand):
     requires_migrations_checks = True
 
     def handle(self, *args, **options):
-        outfile_name = os.path.join(settings.BASE_DIR, "development", "fixtures", "test_data.json")
```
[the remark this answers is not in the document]
richardebeling: I think that's not a problem.

evap/settings.py
richardebeling: We can prefix with `EVAP_` if you anticipate problems, but best practice is to not start-import, and you'll always have the chance to use another name locally for an imported symbol, so for me this is fine as it currently is.

flake.nix
```diff
@@ -90,6 +90,24 @@
...
+                python-dev = self.devShells.${system}.evap-dev.passthru.venv;
+                python-build = self.packages.${system}.python3.withPackages (ps: [ ps.build ]);
+              in
+              ''
+                set -x
+                npm ci
+                ${python-dev}/bin/python ./manage.py compilemessages
+                ${python-dev}/bin/python ./manage.py scss --production
+                ${python-dev}/bin/python ./manage.py ts compile --fresh
+                ${python-build}/bin/python -m build
```
richardebeling, line 108: We now use nix, npm, uv, hatch/hatchling, and build. Granted, nix is just for dev environments. But still, do we _need_ them all? uv promises that it can create distributions? If there's some specific thing we need missing it would be nice to have this documented. It would be nice in general to have a very brief explanation of the job of every tool used here.

richardebeling (commented):
Code changes look fine to me, my biggest concern is complicated release process adding dependencies/complexity that add maintenance and security burden.

One future bug I expect to happen sooner or later is forgetting to add files to the release.

flake.nix
niklasmohrin: I believe uv also has some ideas on how to build and publish python projects, but I wanted to keep that separate and follow the general guidelines from the python package authority: https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#packaging-your-project

This allows us to remove uv if we don't want it anymore for locking dependencies. Hatch (and even more so `build`) seems like the blessed option right now and in general rather stable. So while we don't need it, I think it is the better option right now

niklasmohrin:
> One future bug I expect to happen sooner or later is forgetting to add files to the release.

Yeah, that could be a problem. Maybe once we have #1854, we can let the browser fail on unhandled non-200 responses for any secondary requests like javascript files etc and run it on an instance from the wheel.

niklasmohrin:
I now decided that I will try to split the changes here into multiple PRs, with the one here probably being the final one in the end

pyproject.toml
niklasmohrin: As we have dropped poetry in the meantime, this is not a problem anymore

evap/settings.py
niklasmohrin: I didn't mean that it would be a problem of someone mistakenly overriding our settings, but intentionally trying to do so, like in my proposed settings here: https://github.com/e-valuation/evap-deployment/blob/7f30a50bd2dbbfe157a85fb33e6ed937d6cb3266/productionsettings.template.py

I think it would be nice if it would work this way, but it doesn't because the settings like `MEDIA_ROOT` etc. are already set using the old value of `DATADIR`. Anyways, I suppose it's fair to not care right now

richardebeling: Ah, yeah. Yeah I'd say you just can't override `DATADIR`, you have to overide the derived values - not a big deal.

niklasmohrin:
@janno42 @richardebeling Please review my updated PR description to ensure we are on the same page.

Kakadus:
> 3. After the CI is done on main, dispatch a release workflow run to publish the wheel to PyPI.
> 4. Create Github Release.

we could merge these steps using the `release` trigger: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release

niklasmohrin:
Yeah, I had something like this earlier, but I thought that it would be easier for now to do these steps separately. It's important that the PyPI release CI runs after the main CI is completely finished, otherwise the artifact is not available (and I wanted to reuse the wheel from the main CI run). There isn't really a good way to have a dependency between different workflow files, so I figured that a manual dispatch would be best, in particular because it can be re-run if it was triggered to early.

Notably, it is recommended to have the release in a separate workflow, because then the normal CI doesn't have access to the release environment, which is used to authenticate to PyPI.

We could automatically create a release as part of the PyPI release workflow, but again, I wanted to be a bit cautious with this automation for now.

richardebeling:
Updated description sounds good to me :+1:

richardebeling (approved):

janno42 (approved):
:shipit:

PR #2329, opened 2024-11-04
Delete participations of long inactive users

### LINKED ISSUE #2176: Delete participations of long inactive users
When performing a bulk update of users, the participations of long inactive users should be deleted.
Long inactive users are users who are inactive (or `can_be_marked_inactive_by_manager`) after the bulk update, and where the last `vote_end_date` of the user's participations is more than a certain number of months ago. This number should be configurable in the settings as `PARTICIPATION_DELETION_AFTER_INACTIVE_MONTHS`, with 18 as default.

The bulk update test message should list all users whose participations will be deleted, along with their participation counts.

Redflashx12:
Fixes #2176.

Additionally added two test in class RemoveUserDueToInactivity. During test run of "Bulk update users" in the web with evap/staff/fixtures/test_user_bulk_update_file.txt, Lyndsey Lattimore should get removed from 1 participation(s) due to inactivity..


evap/settings.py
```diff
@@ -36,6 +36,7 @@
 VOTER_COUNT_NEEDED_FOR_PUBLISHING_RATING_RESULTS = 2
 VOTER_PERCENTAGE_NEEDED_FOR_PUBLISHING_AVERAGE_GRADE = 0.2
 SMALL_COURSE_SIZE = 5  # up to which number of participants the evaluation gets additional warnings about anonymity
+PARTICIPATION_DELETION_AFTER_INACTIVE_MONTHS = 18 * 30  # unit in days, i.e. 18 months = 18 * 30
```
richardebeling, line 39: `PARTICIPATION_DELETION_AFTER_INACTIVE_MONTHS` heavily implies that the unit is "months" -- if we expect any other unit in code, we will shoot ourselved in the foot with that sooner or later.

Was there a specific reason to use "days" here? If not, I'd just use months. I don't think we will need sub-month precision. Using the "30 days per month" approximation is probably valid enough in the code that needs to create timedeltas in days.

evap/staff/tests/test_tools.py
```diff
@@ -9,6 +10,7 @@
 from model_bakery import baker
 from openpyxl import load_workbook
 
+from evap import settings
```
richardebeling, line 13: Django has this "weirdness" that you [shouldn't import your settings module directly](https://docs.djangoproject.com/en/5.1/topics/settings/#using-settings-in-python-code), and instead always just use `from django.conf import settings`. It should be a drop-in replacement here.

```suggestion
from django.conf import settings
```

evap/staff/tests/test_tools.py
```diff
@@ -219,6 +222,52 @@ def test_do_nothing_if_test_run(self):
         self.assertEqual(len(messages), 4)
 
 
+class RemoveUserDueToInactivity(TestCase):
+    def test_remove_user_due_to_inactivity(self):
+        inactive_user = baker.make(UserProfile, last_name="Inactive User")
```
richardebeling, line 227: We try to keep the tests as minimal/simple as possible for the required/targeted assertiveness. I think the last_name here is not relevant to the test, so we could remove it. The library we use, [model_bakery](https://model-bakery.readthedocs.io/en/latest/), should fill in any required fields with random values.

(This is also the case below for last_name in the other tests, and possibly also for some other field values)

evap/staff/tests/test_tools.py
```diff
@@ -219,6 +222,52 @@ def test_do_nothing_if_test_run(self):
...
+            vote_end_date=datetime.today() - timedelta(days=settings.PARTICIPATION_DELETION_AFTER_INACTIVE_MONTHS + 1),
+            participants=[inactive_user],
+        )
+
+        semester = inactive_evaluation.course.semester
+        semester.archive()
+
+        messages = remove_inactive_participations(inactive_user)
+        self.assertTrue(inactive_user.can_be_marked_inactive_by_manager)
+        self.assertEqual(len(messages), 1)
```
richardebeling, line 242: I think we should additionally assert what the content of the message is, i.e.
```python3
self.assertEqual(messages, [f"{user.full_name} will be removed from 1 participation(s) due to inactivity"])
```
to make the test more assertive. Same for the other test that asserts that len == 1 below.

evap/staff/tests/test_tools.py
```diff
@@ -219,6 +222,52 @@ def test_do_nothing_if_test_run(self):
...
+
+    def test_do_not_remove_user_due_to_inactivity(self):
+        active_user = baker.make(UserProfile, last_name="Active User")
+
+        baker.make(Evaluation, state=Evaluation.State.PUBLISHED, participants=[active_user])
```
richardebeling, line 247: This is interesting. I don't think there is an actual guarantee here about what will be the `vote_start_datetime` and `vote_end_date` of this evaluation. Thus, this test might (randomly) fail, if model_bakery decides to give us dates way out in the past.

Since the test actually relies on the values, I think we will have to set them.

evap/staff/tests/test_tools.py
```diff
@@ -219,6 +222,52 @@ def test_do_nothing_if_test_run(self):
...
+    def test_do_nothing_if_test_run(self):
+        inactive_user = baker.make(UserProfile, last_name="Test Run User")
+        inactive_evaluation = baker.make(
+            Evaluation,
+            state=Evaluation.State.PUBLISHED,
+            vote_start_datetime=datetime.today()
+            - timedelta(days=settings.PARTICIPATION_DELETION_AFTER_INACTIVE_MONTHS + 1),
+            vote_end_date=datetime.today() - timedelta(days=settings.PARTICIPATION_DELETION_AFTER_INACTIVE_MONTHS + 1),
+            participants=[inactive_user],
+        )
```
richardebeling, line 262: Since all three tests need a user and an evaluation, it would make sense to use the [setup logic that the test framework provides](https://docs.djangoproject.com/en/5.1/topics/testing/tools/#django.test.TestCase.setUpTestData) here. If you add a method
```python3
def setUpTestData(cls):
    cls.user = baker.make(UserProfile)
    six_months_ago = datetime.today() - datetime.timedelta(days=6*30)
    cls.evaluation = baker.make(
        Evaluation,
        state=Evaluation.State.PUBLISHED,
        vote_start_datetime=six_months_ago,
        vote_end_date=six_months_ago.date() + timedelta(days=10)
        participants=[cls.user]
    )
```
you already have `self.user` and `self.evaluation` available in all test methods. You can then
* either change and save the evaluation to make it be before/after the "cutoff" date, or
* vary the settings value of `PARTICIPATION_DELETION_AFTER_INACTIVE_MONTHS ` to make that evaluation be before or after the cutoff. Django provides [`override_settings`](https://docs.djangoproject.com/en/5.1/topics/testing/tools/#django.test.override_settings) to do that. Check out other usages on the codebase (or feel free to just ask someone in person).

evap/staff/tools.py
```diff
@@ -195,6 +195,9 @@ def bulk_update_users(request, user_file_content, test_run):  # noqa: PLR0912
                 user, deletable_users + users_to_mark_inactive, test_run
             ):
                 messages.warning(request, message)
+        for user in users_to_mark_inactive:
+            for message in remove_inactive_participations(user, test_run):
+                messages.warning(request, message)
```
richardebeling, line 200: Ah, this is a bit tricky: The bulk update is supposed to process hundreds or thousands of users. This code now breaks it down to "a thousand times, process one user", which is fine in isolation, but the logic below for "process one user" (in `remove_inactive_participations`) happens to make very latency-heavy database queries (`user.evaluations_participating_in.filter` boils down to a sql join query).

We might still be fine here for now if we just add a mitigation later (i.e., rewrite the code to do one big query instead of many small ones). I'm not sure about current performance and performance requirements.

@janno42 @niklasmohrin @Kakadus thoughts? Might be fine for now. User Bulk Update isn't that often -- but you know me :D We can create a follow up if you're fine with it.

evap/staff/tools.py
```diff
@@ -375,6 +379,30 @@ def remove_user_from_represented_and_ccing_users(user, ignored_users=None, test_
...
 
+def remove_inactive_participations(user, test_run=False):
+    remove_messages = []
+    evaluations = []
+    if user.is_active and user.can_be_marked_inactive_by_manager:
+        evaluations = user.evaluations_participating_in.filter(
+            state=Evaluation.State.PUBLISHED,
+            vote_end_date__lt=datetime.today() - timedelta(days=settings.PARTICIPATION_DELETION_AFTER_INACTIVE_MONTHS),
+        )
+    if len(evaluations) > 0:
```
richardebeling, line 390: In Python, programmers often try to go after this [notion of "Pythonicity"](https://stackoverflow.com/questions/25011078/what-does-pythonic-mean). Often, this boils down to ["why waste time say lot word when few word do trick"](https://www.youtube.com/watch?v=VvPaEsuz-tY). Here specifically, all well-behaved container-like structured in python allow to test for emptiness by simply "using as" boolean, i.e.:

```suggestion
    if evaluations:
```

evap/staff/tools.py
```diff
@@ -375,6 +379,30 @@ def remove_user_from_represented_and_ccing_users(user, ignored_users=None, test_
...
+                _("{} will be removed from {} participation(s) due to inactivity.").format(
+                    user.full_name, len(evaluations)
+                )
+            )
+        else:
+            for evaluation in evaluations:
+                evaluation.participants.remove(user)
+            remove_messages.append(
+                _("Removed {} from {} participation(s) due to inactivity.").format(user.full_name, len(evaluations))
+            )
```
richardebeling, line 402: I don't think this is the correct logic for what we envisioned in #2176. We want:
> A user who hasn't participated in evaluations in the last X months is "long_inactive". Participations of "long_inactive" users are removed when performing bulk update

whereas this code currently does
> Given a user x, if x has any participations in evaluations that ended more than X months ago, remove them as a participant from those evaluations.

If I have been studying for 5 years and have been participating in evaluations regularly, I shouldn't be classified as "long_inactive" since my last evaluation is not that long ago. But I will still have participated in evaluations that have ended a long time ago.

---

There should be django queryset magic that allows to, given a set of users, annotate to each of them when exactly they last participated in an evaluation (proxied by the evaluation end date of the evaluations they participated in). Based on that distinction, we can then remove all participations of the user (or none). It's probably easier to explain this in person, so I recommend asking one of the maintainers/organizers at the next meetup :)

richardebeling: Note that the test suite passed. This can be considered as an indication that the tests aren't assertive enough to distuingish correct from incorrect behavior, and should be expanded to include more edge-caseish situations (e.g. the one I outlined above).

evap/staff/tools.py
```diff
@@ -375,6 +379,30 @@ def remove_user_from_represented_and_ccing_users(user, ignored_users=None, test_
     return remove_messages
 
 
+def remove_inactive_participations(user, test_run=False):
+    remove_messages = []
+    evaluations = []
+    if user.is_active and user.can_be_marked_inactive_by_manager:
```
richardebeling, line 385: For the check as described in the issue::
> Long inactive users are users who are inactive (or can_be_marked_inactive_by_manager) and [...]

I would expect that we need to have a condition like
```python3
if not user.is_active or user.can_be_marked_inactive_by_manager:
```
here, or am I missing something?

richardebeling (changes requested):
Good work so far! Sorry, quite a few comments added up, this could be a bit overwhelming. We want to address everything so we don't get stuck in endless review loops. We're always happy to help out and answer any questions, so make sure to ask if anything is unclear.

evap/settings.py
niklasmohrin: Is there any reason to not specify a timedelta directly here in the settings?

evap/staff/tools.py
niklasmohrin: The same problem already exists in the loop right before right? In that case, I am fine with (maybr even prefer) leaving it as is for now

evap/settings.py
richardebeling: We can do that. I was hesitant to suggest it since we haven't done that for any setting yet. I was thinking of migrating to `datetime` and `fraction` instances in a separate PR. We have used `Fraction`s for one setting and it didn't cause any problems, so introducing `datetime` here would probably be fine.

Redflashx12: > Was there a specific reason to use "days" here? If not, I'd just use months. I don't think we will need sub-month precision. Using the "30 days per month" approximation is probably valid enough in the code that needs to create timedeltas in days.

The biggest unit in timedelta is weeks, there is no option for months. But I thought days make more sense, i.e. 30 days are more accurate than 4 weeks. There are parameters for days, seconds, microseconds, milliseconds, minutes, hours and weeks.

But I can also just use the timedelta here, which makes it probably also less confusing.

niklasmohrin: I think that using timedelta directly is better than the current version; I think I prefer it over having "30 * days" hardcoded - managers can still choose to use this calculation, but could also put something like `days = 2 * 365`. @richardebeling ?

evap/staff/tools.py
niklasmohrin: https://docs.djangoproject.com/en/5.1/topics/db/aggregation/#joins-and-aggregates

Redflashx12:
> Good work so far! Sorry, quite a few comments added up, this could be a bit overwhelming. We want to address everything so we don't get stuck in endless review loops. We're always happy to help out and answer any questions, so make sure to ask if anything is unclear.

Thanks for your time in reviewing this PR, especially for all the fixes and explanations. Much appreciated 👍.

evap/staff/tools.py
Redflashx12: Is this how it's supposed to be now?

evap/staff/tests/test_tools.py
Redflashx12: Tried to implement it with override_settings, please have a look and give me some feedback if you see any improvements to be made =)

evap/staff/tools.py
```diff
@@ -375,6 +381,25 @@ def remove_user_from_represented_and_ccing_users(user, ignored_users=None, test_
...
+    ):
+        return []
+
+    evaluation_count = user.evaluations_participating_in.count()
+    if test_run:
+        return [
+            _("{} will be removed from {} participation(s) due to inactivity.").format(user.full_name, evaluation_count)
+        ]
+    user.evaluations_participating_in.clear()
+    return [_("Removed {} from {} participation(s) due to inactivity.").format(user.full_name, evaluation_count)]
```
janno42, line 400:
```suggestion
    return [
        ngettext(
            "{} participation of {} was removed due to inactivity.",
            "{} participations of {} were removed due to inactivity.",
            evaluation_count,
        ).format(evaluation_count, user.full_name)
    ]
```

janno42 (changes requested):

niklasmohrin:
@Redflashx12 Do you want to keep working on this PR (this or next year), or should we take over?

Redflashx12:
I'll be active in EvaP in January again, still busy with moving in rn.
Sorry 😐

niklasmohrin:
Great, good luck with that! No need to worry or hurry, we just want to track the status of the PR :)

evap/staff/tests/test_tools.py
```diff
@@ -217,6 +221,73 @@ def test_do_nothing_if_test_run(self):
...
+        self.assertTrue(self.user.can_be_marked_inactive_by_manager)
+        self.assertEqual(messages, [f"Removed {self.user.full_name} from 1 participation(s) due to inactivity."])
+
+        messages = remove_inactive_participations(self.user)
+
+        self.assertEqual(messages, [])
+
+    @patch("evap.evaluation.models.UserProfile.is_active", True)
+    @patch("evap.evaluation.models.UserProfile.can_be_marked_inactive_by_manager", True)
+    def test_do_not_remove_user_due_to_inactivity_with_recently_archived_evaluation(self):
```
richardebeling, line 254: This relies on the default setting value, but it would be nice if the test passed even if some developers locally changed the setting. Thus, we should also

```suggestion
    @override_settings(PARTICIPATION_DELETION_AFTER_INACTIVE_TIME=timedelta(360))
    def test_do_not_remove_user_due_to_inactivity_with_recently_archived_evaluation(self):
```

evap/staff/tests/test_tools.py
```diff
@@ -217,6 +221,73 @@ def test_do_nothing_if_test_run(self):
...
+        self.assertTrue(self.user.evaluations_participating_in.exists())
+
+        messages = remove_inactive_participations(self.user)
+
+        self.assertTrue(self.user.evaluations_participating_in.exists())
+        self.assertEqual(messages, [])
+
+    @patch("evap.evaluation.models.UserProfile.is_active", True)
+    @patch("evap.evaluation.models.UserProfile.can_be_marked_inactive_by_manager", False)
+    def test_do_not_remove_user_due_to_inactivity_with_active_evaluation(self):
```
richardebeling, line 264: Needs an override on PARTICIPATION_DELETION_AFTER_INACTIVE_TIME to ensure we actually don't do anything because of `can_be_marked_inactive_by_manager`, and not because the time simply hasn't passed.

evap/staff/tests/test_tools.py
```diff
@@ -217,6 +221,73 @@ def test_do_nothing_if_test_run(self):
...
+        self.assertEqual(
+            messages, [f"{self.user.full_name} will be removed from 1 participation(s) due to inactivity."]
+        )
+
+        messages = remove_inactive_participations(self.user, test_run=True)
+
+        self.assertEqual(
+            messages, [f"{self.user.full_name} will be removed from 1 participation(s) due to inactivity."]
+        )
+
```
richardebeling, line 289: This is just the same assertion again, duplicated?

We should instead assert that the participations of the user haven't changed.

evap/staff/tools.py
```diff
@@ -132,12 +132,14 @@ def bulk_update_users(request, user_file_content, test_run):  # noqa: PLR0912
...
 
     emails_of_non_obsolete_users = set(imported_emails) | {user.email for user, _ in users_to_be_updated}
-    deletable_users, users_to_mark_inactive = [], []
+    deletable_users, users_to_mark_inactive, inactive_users = [], [], []
     for user in UserProfile.objects.exclude(email__in=emails_of_non_obsolete_users):
         if user.can_be_deleted_by_manager:
             deletable_users.append(user)
         elif user.is_active and user.can_be_marked_inactive_by_manager:
             users_to_mark_inactive.append(user)
+        elif not user.is_active:
```
richardebeling, line 141: The elif here seems suspicious to me. An inactive user that has `can_be_marked_inactive_by_manager == False` wouldn't end up in `inactive_users`, and thus we wouldn't call `remove_inactive_participations`. I don't know whether such a user can exist, but I'd prefer to not have to think about this.

I think the loop here was meaningful and concise before, and just handled user deletion (or deactivation, if deletion isn't possible). I'd view removal of participations as orthogonal, and try to keep these as separate as possible, i.e., first just handle all deletion/marking inactive, and then handle participation archival on the remaining users. What do you think about that?

richardebeling (changes requested):

evap/staff/tests/test_tools.py
Redflashx12: I wanted to check that the behaviour was consistent and that nothing had changed. The function implicitly checked the count, but I can just check it myself, which makes it independent and probably much cleaner overall. Thanks.

evap/staff/tools.py
Redflashx12: I added that line because the requirement in the original issue #2176 mentioned that "long inactive users are users who are inactive (or can_be_marked_inactive_by_manager)", but I just checked the if condition and set a breakpoint and during the bulk update test nothing was ever added to that list.

evap/staff/tools.py
```diff
@@ -375,6 +380,35 @@ def remove_user_from_represented_and_ccing_users(user, ignored_users=None, test_
     return remove_messages
 
 
+def remove_inactive_participations(user: UserProfile, test_run=False) -> list[StrOrPromise]:
```
niklasmohrin, line 383: I was a bit confused on first read, because it's not the participations that are inactive, but rather it's the user who could be. How do you feel about renaming to something like `remove_participations_if_inactive`?

evap/staff/tools.py
```diff
@@ -195,6 +196,9 @@ def bulk_update_users(request, user_file_content, test_run):  # noqa: PLR0912
                 user, deletable_users + users_to_mark_inactive, test_run
             ):
                 messages.warning(request, message)
+        for user in users_to_mark_inactive:
+            for message in remove_inactive_participations(user, test_run):
+                messages.warning(request, message)
```
niklasmohrin, line 201: This line is not covered by tests, because we only test the implementation of `remove_inactive_participations` - should we modify `TestUserBuldUpdateView` to include this? @richardebeling @Kakadus 

It would boil down to setting up a situation where a user is marked as inactive and checking that the message appears.

evap/staff/tests/test_tools.py
```diff
@@ -217,6 +220,70 @@ def test_do_nothing_if_test_run(self):
...
+    @patch("evap.evaluation.models.UserProfile.is_active", True)
+    @patch("evap.evaluation.models.UserProfile.can_be_marked_inactive_by_manager", True)
+    @override_settings(PARTICIPATION_DELETION_AFTER_INACTIVE_TIME=timedelta(360))
+    def test_do_not_remove_user_due_to_inactivity_with_recently_archived_evaluation(self):
+        self.assertTrue(self.user.evaluations_participating_in.exists())
+
+        messages = remove_inactive_participations(self.user)
+
+        self.assertTrue(self.user.evaluations_participating_in.exists())
```
niklasmohrin, line 258: We can be more explicit here (and below):
```suggestion
        self.assertEqual(list(self.user.evaluations_participating_in.all()), [self.evaluation])
```

niklasmohrin (commented):
Looks mostly good to me, I only have minor comments:

evap/staff/tools.py
Redflashx12: I'm not sure whether you want me to implement this and if so, where can I find the `TestUserBuildUpdateView`? I didn't find it when searching through the project files.

evap/staff/tests/test_tools.py
Redflashx12: Done.

evap/staff/tools.py
Redflashx12: Sure, changed it.

evap/staff/tools.py
niklasmohrin: Typo, it's supposed to be _bulk_ update (`TestUserBulkUpdateView`). Since @richardebeling seems to agree, I think we would want a short test in that class so that some of the warnings you added are shown

evap/staff/tests/test_tools.py
```diff
@@ -217,6 +220,70 @@ def test_do_nothing_if_test_run(self):
...
+    @patch("evap.evaluation.models.UserProfile.is_active", True)
+    @patch("evap.evaluation.models.UserProfile.can_be_marked_inactive_by_manager", False)
+    @override_settings(PARTICIPATION_DELETION_AFTER_INACTIVE_TIME=timedelta(360))
+    def test_do_not_remove_user_due_to_inactivity_with_active_evaluation(self):
+        self.assertEqual(list(self.user.evaluations_participating_in.all()), [self.evaluation])
+
+        messages = remove_participations_if_inactive(self.user)
+
+        self.assertEqual(list(self.user.evaluations_participating_in.all()), [self.evaluation])
+        self.assertEqual(messages, [])
```
Kakadus, line 270:
```suggestion
        self.assertQuerySetEqual(self.user.evaluations_participating_in.all(), [self.evaluation])
        self.assertEqual(messages, [])
```
and also above?

janno42 (approved):
In the future, it would probably be nice to have the accounts listed in one message like those to be marked inactive, but I think it's fine for now.

Kakadus (changes requested):
requesting the missing test from above

evap/staff/tools.py
Kakadus: should be resolved now with my latest commit

richardebeling (approved):

niklasmohrin (commented):
Looks good, only two small suggestions:

niklasmohrin (approved):

PR #2338, opened 2024-11-18
format before linting

Kakadus:
ruff and black have some overlaps, such as detecting/fixing trailing whitespaces. So, it is more user friendly to format before linting, so that these kinds of trivial linting errors can be fixed beforehand.


niklasmohrin (approved):
Can you change the PR title to indicate that this is about the pre-commit command?

PR #2344, opened 2024-11-25
Fix create_user and create_superuser

### LINKED ISSUE #2204: Creating Superuser fails on a fresh install
Using docker provider. Start up a fresh docker instance via Vagrant. Try the create superuser command.
It asks for login email and password. Enter credentials. It then crashes due to the postgresql having a constraint that first name of a user must not be null.

richardebeling:
fixed #2204

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -1660,18 +1660,18 @@ class Infotext(models.Model):
 
 
 class UserProfileManager(BaseUserManager):
-    def create_user(self, *, email, password=None, first_name=None, last_name=None):
-        user = self.model(email=self.normalize_email(email), first_name_given=first_name, last_name=last_name)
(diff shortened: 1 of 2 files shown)


niklasmohrin (approved):

Kakadus (approved):
I'm wondering if it's worth to have a test that `createsuperuser` and such works, in case we modify our UserProfile model yet again

janno42 (approved):

richardebeling:
> I'm wondering if it's worth to have a test that `createsuperuser` and such works, in case we modify our UserProfile model yet again

@niklasmohrin said the same. I don't really care. Do you want me to add one?

niklasmohrin:
For me it depends on how much effort it is. If it's 20 lines, I would just add it and stop thinking about it. If it's more, I am also fine with not-caring and will, likewise, stop thinking about it

niklasmohrin (approved):

PR #2349, opened 2024-12-02
Dispatch a `click` event on an off-DOM `<a>` element when clicking on `.table-seamless-links` rows

### LINKED ISSUE #2292: Open evaluations in new tabs
Control-clicking on evaluation rows on student index and contributor index should open the pages in new tabs (see results page for an example).
Currently, clicking reloads the page.

niklasmohrin:
Closes #2292

This brings the interaction closer to clicking on an actual `a` tag and enables Ctrl-Click again. Having an `a` in there is still not feasible, because we cannot have any elements between `tr` and `td`.

### DIFF
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -183,7 +183,7 @@
             document.querySelectorAll(".table-seamless-links [data-url]").forEach(row => {
                 row.addEventListener("click", event => {
                     // We navigate via the <a> tag to enable browser behavior for Ctrl-Click. We cannot insert an actual
-                    // <a> tag, because there cannot be any non-table tags between the for example <tbody> and <tr>.
+                    // <a> tag, because there cannot be any non-table tags between for example <tbody> and <tr>.
                     const anchor = document.createElement("a");
                     anchor.href = row.dataset.url;
                     // As `event` is bound to `row`, we need to clone it.
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -182,10 +182,12 @@
             // activate clickable hover tables
             document.querySelectorAll(".table-seamless-links [data-url]").forEach(row => {
                 row.addEventListener("click", event => {
-                    // We navigate via the <a> tag to enable browser behavior for Ctrl-Click.
+                    // We navigate via the <a> tag to enable browser behavior for Ctrl-Click. We cannot insert an actual
+                    // <a> tag, because there cannot be any non-table tags between the for example <tbody> and <tr>.
                     const anchor = document.createElement("a");
                     anchor.href = row.dataset.url;
-                    anchor.dispatchEvent(new PointerEvent("click", event));
+                    // As `event` is bound to `row`, we need to clone it.
+                    anchor.dispatchEvent(new event.constructor(event.type, event));
                 });
             });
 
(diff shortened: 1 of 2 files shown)


evap/evaluation/templates/base.html
```diff
@@ -181,8 +181,11 @@
 
             // activate clickable hover tables
             document.querySelectorAll(".table-seamless-links [data-url]").forEach(row => {
-                row.addEventListener("click", () => {
-                    window.location.assign(row.dataset.url);
+                row.addEventListener("click", event => {
+                    // We navigate via the <a> tag to enable browser behavior for Ctrl-Click.
```
richardebeling, line 185: maybe add a short comment why we do this so we don't go back to creating `a` tags. Something like
> HTML disallows `a` as a direct child of `tr`.

evap/evaluation/templates/base.html
```diff
@@ -181,8 +181,11 @@
...
+                    const anchor = document.createElement("a");
+                    anchor.href = row.dataset.url;
+                    anchor.dispatchEvent(new PointerEvent("click", event));
```
richardebeling, line 188: What's going on here, why don't just forward `event`? I understand PointerEvent may be preferrable to MouseEvent, but I would have hoped that we wouldn't need to care in this context. We're just handling `click` and forwarding to `click`.

richardebeling (approved):

evap/evaluation/templates/base.html
niklasmohrin: For `anchor.dispatchEvent(event)` I get `Uncaught DOMException: An attempt was made to use an object that is not, or is no longer, usable`. I makes some sense, because `event` is tied to `row` via `event.target` and has state about event-bubbling and whatnot.

I then picked PointerEvent based on https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event#event_type. If we are not confident with that (for example because of the Firefox disclaimer on the mdn page), I am also happy with `new event.constructor(event.type, event)` (from https://stackoverflow.com/a/20541207/13679671)

richardebeling: Ah, makes sense. Weird, the MDN site I checked earlier listed "MouseEvent" as the event class for "click".

I'm fine with any approach, with a slight tendency towards
```javascript
// `event` has target-dependent state (`target`, bubbling, ...) -> create a new one
anchor.dispatchEvent(new event.constructor(event.type, event));
```

janno42 (commented):
This solves the main problem, but it's not very nice. After control-clicking, there's an ugly selection box around part of the row, and because we don't have an anchor element, there's no link preview in the browser.
Maybe we should use a grid like on the results page instead of tables?

niklasmohrin:
With this version, the outline is fixed, but I think we can't show a link preview with this approach at all. I think replacing the table with grid is the end solution, but I don't know when someone will do this and it will definitely take longer than the fix here

evap/evaluation/templates/base.html
```diff
@@ -181,8 +181,13 @@
 
             // activate clickable hover tables
             document.querySelectorAll(".table-seamless-links [data-url]").forEach(row => {
-                row.addEventListener("click", () => {
-                    window.location.assign(row.dataset.url);
+                row.addEventListener("click", event => {
+                    // We navigate via the <a> tag to enable browser behavior for Ctrl-Click. We cannot insert an actual
+                    // <a> tag, because there cannot be any non-table tags between the for example <tbody> and <tr>.
```
janno42, line 186:
```suggestion
                    // <a> tag, because there cannot be any non-table tags between for example <tbody> and <tr>.
```

janno42 (approved):

Kakadus (approved):

PR #2353, opened 2024-12-09
Silence `createsuperuser` during tests

Kakadus:
We did not spot that in #2344.


niklasmohrin (approved):

richardebeling:
ah, sorry. This has happened a few times now, it might make sense to just fail CI if anything is printed to stdout during tests.

PR #2355, opened 2024-12-16
Exit staff mode after permission revocation

### LINKED ISSUE #2205: Clear session of user on permission change
When removing manager or reviewer permissions from a user, that user's session should be cleared.
Currently, logging in with a user who was a reviewer and previously used Staff mode fails with an AssertionError:

```
File "/opt/evap/evap/staff/staff_mode.py", line 56, in update_staff_mode
    assert request.user.has_staff_permission
```

ybrnr:
fix #2205

### DIFF
--- a/evap/staff/staff_mode.py
+++ b/evap/staff/staff_mode.py
@@ -30,6 +30,7 @@ def staff_mode_middleware(get_response):
                     messages.info(request, _("Your staff mode timed out."))
 
         if is_in_staff_mode(request):
+            assert request.user.has_staff_permission
             request.user.is_participant = False
             request.user.is_student = False
             request.user.is_editor = False
@@ -52,10 +53,9 @@ def is_in_staff_mode(request):
 
 
 def update_staff_mode(request):
-    if not request.user.has_staff_permission: 
+    if not request.user.has_staff_permission:
         exit_staff_mode(request)
         return
-    assert request.user.has_staff_permission
 
     request.session["staff_mode_start_time"] = time.time()
     request.session.modified = True
--- a/evap/staff/staff_mode.py
+++ b/evap/staff/staff_mode.py
@@ -15,7 +15,6 @@ def staff_mode_middleware(get_response):
     """
 
     def middleware(request):
-        print("jetzt hier")
         if is_in_staff_mode(request):
             current_time = time.time()
             if current_time <= request.session.get("staff_mode_start_time", 0) + STAFF_MODE_TIMEOUT:
--- a/evap/staff/staff_mode.py
+++ b/evap/staff/staff_mode.py
@@ -15,6 +15,7 @@ def staff_mode_middleware(get_response):
     """
 
     def middleware(request):
+        print("jetzt hier")
         if is_in_staff_mode(request):
             current_time = time.time()
             if current_time <= request.session.get("staff_mode_start_time", 0) + STAFF_MODE_TIMEOUT:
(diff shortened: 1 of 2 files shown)


janno42 (approved):
Nice 👍 

Please check the formatting

evap/staff/staff_mode.py
```diff
@@ -52,6 +52,9 @@ def is_in_staff_mode(request):
 
 
 def update_staff_mode(request):
+    if not request.user.has_staff_permission: 
+        exit_staff_mode(request)
+        return
     assert request.user.has_staff_permission
```
Kakadus, line 58: I think this assertion is no longer needed

Kakadus (commented):
naturally, a test would be nice : )

Thanks!

evap/staff/staff_mode.py
richardebeling: I'd move it up to line 33, where I think it would still add value.

The call graphs we can get now are a bit wonky. We can have
`enter_staff_mode` -> `update_staff_mode` -> finds out that `not request.user.has_staff_permission` -> `exit_staff_mode` and returns without exception, which is probably never what the initial caller intended?

It seems to me that `update_staff_mode` was about "update the timestamp in the session to now()" in the past, now it has changed semantics. Could use a little clean up in general, I think.

richardebeling (changes requested):
Blocking for missing test, since this is authorization logic and we discovered a code path in the issue that we considered not to happen naturally before. Feel free to assign the PR to me if you want me to write the test.

evap/staff/staff_mode.py
ybrnr: I moved the assertion to line 33

evap/staff/staff_mode.py
```diff
@@ -52,7 +53,9 @@ def is_in_staff_mode(request):
 
 
 def update_staff_mode(request):
-    assert request.user.has_staff_permission
+    if not request.user.has_staff_permission:
+        exit_staff_mode(request)
+        return
```
niklasmohrin, line 58: @richardebeling @Kakadus Should we log something here?

niklasmohrin (approved):
Very cool, thanks!

evap/staff/staff_mode.py
niklasmohrin: I agree that, while it's not super urgent, the names could be improved. In general, `enter_staff_mode` seems unneeded since #1587. The current `update_staff_mode` is more of a `try_entering_or_refreshing_staff_mode`. I don't have a strong opinion, but having a "try to enter, but we will check whether you are allowed" as the only way to enter staff mode seems fair to me. Maybe the `enter_staff_mode` view should check that entering worked?

Kakadus (approved):

niklasmohrin:
@ybrnr How about "Exit staff mode after permission revocation"?

evap/staff/staff_mode.py
richardebeling: I'm fine with both adding and not adding logging.

niklasmohrin: If you both don't care, I would prefer if we added a small info-level log like "user {} tried to update staff mode session without permission" @ybrnr

niklasmohrin (changes requested):
Blocking for https://github.com/e-valuation/EvaP/pull/2355#discussion_r1931197134

niklasmohrin (approved):
Nothing to see here

richardebeling (approved):

richardebeling:
Thanks! :)

PR #2358, opened 2024-12-21
Grant reward points on evaluation deletion

### LINKED ISSUE #2279: Reward points not granted on evaluation deletion
`grant_reward_points_after_delete` is not triggered when an evaluation is deleted, so currently reward points are not granted in this case.

Reward points should also be granted on evaluation deletion.

richardebeling:
Fixes #2279

### DIFF
--- a/evap/rewards/tools.py
+++ b/evap/rewards/tools.py
@@ -2,6 +2,7 @@ import logging
 
 from django.conf import settings
 from django.contrib import messages
+from django.http import HttpRequest
 from django.db import models
 from django.db.models import Sum
 from django.dispatch import receiver
@@ -15,11 +16,11 @@ from evap.rewards.models import RewardPointGranting, RewardPointRedemption, Seme
 logger = logging.getLogger(__name__)
 
 
-def can_reward_points_be_used_by(user):
+def can_reward_points_be_used_by(user: UserProfile) -> bool:
     return not user.is_external and user.is_participant
 
 
-def reward_points_of_user(user):
+def reward_points_of_user(user: UserProfile) -> int:
     count = 0
     for granting in user.reward_point_grantings.all():
         count += granting.value
@@ -29,15 +30,19 @@ def reward_points_of_user(user):
     return count
 
 
-def redeemed_points_of_user(user):
+def redeemed_points_of_user(user: UserProfile) -> int:
     return RewardPointRedemption.objects.filter(user_profile=user).aggregate(Sum("value"))["value__sum"] or 0
 
 
-def is_semester_activated(semester):
+def is_semester_activated(semester: Semester) -> bool:
     return SemesterActivation.objects.filter(semester=semester, is_active=True).exists()
 
 
-def grant_reward_points_if_eligible(user, semester):
+def deactivate_semester(semester: Semester) -> None:
+    SemesterActivation.objects.filter(semester=semester).update(is_active=False)
+
+
+def grant_reward_points_if_eligible(user: UserProfile, semester: Semester) -> tuple[RewardPointGranting | None, bool]:
     if not can_reward_points_be_used_by(user):
         return None, False
     if not is_semester_activated(semester):
@@ -62,7 +67,7 @@ def grant_reward_points_if_eligible(user, semester):
     return None, False
 
 
-def grant_eligible_reward_points_for_semester(request, semester):
+def grant_eligible_reward_points_for_semester(request: HttpRequest, semester: Semester) -> None:
     users = UserProfile.objects.filter(evaluations_voted_for__course__semester=semester)
     reward_point_sum = 0
     for user in users:
@@ -82,7 +87,7 @@ def grant_eligible_reward_points_for_semester(request, semester):
 
 
 @receiver(Evaluation.evaluation_evaluated)
-def grant_reward_points_after_evaluate(request, semester, **_kwargs):
+def grant_reward_points_after_evaluate(request: HttpRequest, semester: Semester, **_kwargs) -> None:
     granting, completed_evaluation = grant_reward_points_if_eligible(request.user, semester)
     if granting:
         message = ngettext(
@@ -105,7 +110,7 @@ def grant_reward_points_after_evaluate(request, semester, **_kwargs):
 
 
 @receiver(models.signals.m2m_changed, sender=Evaluation.participants.through)
-def grant_reward_points_on_participation_change(instance, action, reverse, pk_set, **_kwargs):
+def grant_reward_points_on_participation_change(instance, action: str, reverse: bool, pk_set, **_kwargs) -> None:
     # if users do not need to evaluate anymore, they may have earned reward points
     if action == "post_remove":
         grantings = []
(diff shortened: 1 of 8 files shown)


evap/rewards/tools.py
```diff
@@ -123,4 +124,27 @@ def grant_reward_points_after_delete(instance, action, reverse, pk_set, **_kwarg
...
+
+@receiver(models.signals.pre_delete, sender=Evaluation)
+def grant_reward_points_on_evaluation_delete(instance, **_kwargs):
+    if not inside_transaction():
+        # This will always be true in a django TestCase, so our tests can't meaningfully catch calls that are not
+        # wrapped in a transaction. Requiring a transaction is a precaution so that an (unlikely) failing .delete()
+        # execution afterwards doesn't leave us in half-deleted state. Chances are, if deletion fails, staff users
+        # will still want to delete the instance.
+        # Currently, only staff:evaluation_delete and staff:semester_delete call .delete()
+        logger.exception("Called while not inside transaction")
```
richardebeling, line 138: @niklasmohrin this is all a bit annoying. Ideally, I'd like to
1. start a transaction
2. store a list of participants
3. delete the evaluation, including all m2m-participants
4. create RewardPointGrantings (with transaction rollback on failure)

While the signal model allows me to run code before and after deletion, I don't see a nice way to pass the list of participants to it without having global state

I think this is fine for now. If we ever run into a "wanted to delete, failed, now need cleanup"-state, we will likely need manual cleanup anyway.

I'll look into whether we can ditch the signals and use a custom manager instead, which is what the [Django docs](https://docs.djangoproject.com/en/5.1/ref/signals/) propose to do instead:
> Signals can make your code harder to maintain. Consider implementing a helper method on a [custom manager](https://docs.djangoproject.com/en/5.1/topics/db/managers/#custom-managers), to both update your models and perform additional logic, or else [overriding model methods](https://docs.djangoproject.com/en/5.1/topics/db/models/#overriding-model-methods) before using model signals.

evap/staff/views.py
```diff
@@ -671,6 +674,8 @@ def semester_make_active(request):
 @require_POST
 @manager_required
 def semester_delete(request):
+    # TODO(rebeling): Do we expect reward point granting here? Do we want to notify? I don't see why it couldn't happen
```
richardebeling, line 677: @janno42 I didn't find anything in the code that prohibits deleting semesters with active rewards. I think it's possible that we also get a reward point granting here? Do we want that? (If yes, I guess we want a matching `messages.info`)

evap/staff/views.py
```diff
@@ -1418,9 +1423,15 @@ def helper_single_result_edit(request, evaluation):
...
 @require_POST
 @manager_required
+@transaction.atomic
 def evaluation_delete(request):
     evaluation = get_object_from_dict_pk_entry_or_logged_40x(Evaluation, request.POST, "evaluation_id")
 
+    # See comment in helper_evaluation_edit
+    @receiver(RewardPointGranting.granted_by_evaluation_deletion, weak=True)
+    def notify_reward_points(grantings, **_kwargs):
+        logger.info(f"Deletion of evaluation {evaluation} has created {len(grantings)} reward point grantings")
```
richardebeling, line 1433: @janno42 evaluation deletion currently just removes the row, it doesn't reload the page, so a `messages.info` call will only be shown on the next page (re)load. Do we want this to be shown to users, and if yes, what should it look like?

evap/rewards/tools.py
niklasmohrin, line 138: In the docs they say we should only use `logger.exception` in an exception handler, so I guess it should be `logger.error`?

evap/staff/views.py
janno42: Semesters can only be deleted when they have no evaluations or all evaluations have been archived. So there can't be any evaluations which should be counted towards reward points, right?

evap/staff/views.py
janno42: It's okay to only put it in the log file. We don't have to show it to managers.

evap/staff/views.py
richardebeling: I think `archive` on an evaluation just sets `_participant_count` and `_voter_count`, it doesn't remove participants/voters. Then, we could still have an evaluation where, after archiving, someone is participant, but not voter. Removing them from participants would trigger reward point redemption, I think.

Maybe I'm missing something, but I think we just want to disable rewards for a semester before deleting?

evap/rewards/tools.py
richardebeling: I decided that I don't want to have the hassle here, I think the current signal approach is fine for now. Resolving. Open to anyone migrating this to a custom manager, but I don't want to do it here.

evap/staff/views.py
Kakadus: Yes. I'd say `archive` or `semester_delete` should deactivate or check the `SemesterActivation`

evap/rewards/tools.py
```diff
@@ -123,4 +128,27 @@ def grant_reward_points_after_delete(instance, action, reverse, pk_set, **_kwarg
...
+        # will still want to delete the instance.
+        # Currently, only staff:evaluation_delete and staff:semester_delete call .delete()
+        logger.error("Called while not inside transaction")
+
+    grantings = []
+
+    participants = list(instance.participants.all())
+    instance.participants.clear()
+    for user in participants:
+        granting, __ = grant_reward_points_if_eligible(user, instance.course.semester)
```
richardebeling, line 149: note to self: TODO: Add tests that the reward points are actually granted (evaluation deletion, user profile edit)

evap/staff/views.py
```diff
@@ -1418,9 +1421,18 @@ def helper_single_result_edit(request, evaluation):
...
 @require_POST
 @manager_required
+@transaction.atomic
 def evaluation_delete(request):
     evaluation = get_object_from_dict_pk_entry_or_logged_40x(Evaluation, request.POST, "evaluation_id")
 
+    # See comment in helper_evaluation_edit
+    @receiver(RewardPointGranting.granted_by_evaluation_deletion, weak=True)
+    def notify_reward_points(grantings, **_kwargs):
+        logger.info(
```
niklasmohrin, line 1434: The other `notify_reward_points` methods use `messages.info` instead, is there any reason to use one over the other here?

evap/staff/tests/test_views.py
```diff
@@ -899,6 +901,28 @@ def test_success_with_data(self):
 
         self.assertFalse(Semester.objects.filter(pk=self.instance.pk).exists())
 
+    @override_settings(REWARD_POINTS=[(1, 3)])
+    def test_does_not_grant_redemption_points(self):
+        student = baker.make(UserProfile, email="student@institution.example.com")
+        evaluation = baker.make(Evaluation, participants=[student], state=Evaluation.State.PUBLISHED, course__semester=self.instance)
+        SemesterActivation.objects.update_or_create(semester=self.instance, defaults={"is_active": True})
```
niklasmohrin, line 910: Why update _or_ create? Shouldn't we be in control of whether or not one exists already?

niklasmohrin (approved):
Very nice!

janno42 (approved):

evap/staff/views.py
richardebeling: This is an ajax post handler, so the page of the user will not reload (and thus not show messages) when the deletion goes through. If we called messages.info, the message would show on the next page that the user manually goes to, which I'd say is probably surprising/unexpected.

Discussed here before: https://github.com/e-valuation/EvaP/pull/2358#discussion_r1894641174

evap/staff/tests/test_views.py
richardebeling: My thought here was that the test doesn't care whether `Semester.create` / `save` / the login in `setUpTestData` will create a semester activation, and if it does, what its value is. All we want is the semester to be active. This line generically gets us to that state, independently of what state we were in before.

Kakadus (approved):

PR #2364, opened 2025-01-06
remove get_by_natural_key implementation on UserProfileManager

Kakadus:
leftover from #2249.

You can see an effect, if using the "natural key notation" `[null]`" for a UserProfile fk:

WITHOUT `remove get_by_natural_key implementation on UserProfileManager`, loading the fixture below results in 
```py
django.core.serializers.base.DeserializationError: Problem installing fixture '/home/jonas/development/evap/breakage.json': get() returned more than one UserProfile -- it returned 2!: (evaluation.evaluation:pk=356000) field_value was '[None]'
```

while WITH the commit, the error is (correctly) 
```py
django.core.serializers.base.DeserializationError: Problem installing fixture '/home/jonas/development/evap/breakage.json': ['“[None]” value must be an integer.']: (evaluation.evaluation:pk=356000) field_value was '[None]'
```

json fixture to load:
```json
[{"model": "evaluation.evaluation","pk": 356000,"fields": {"state": 80,"course": 141,"name_de": "","name_en": "","weight": 1,"is_single_result": false,"is_rewarded": true,"is_midterm_evaluation": false,"can_publish_text_results": true,"_participant_count": 62,"_voter_count": 35,"vote_start_datetime": "2022-02-02T00:00:00","vote_end_date": "2022-02-12","allow_editors_to_edit": true,"wait_for_grade_upload_before_publishing": true,"participants": [[null]],"voters": []}}]
```

I cannot think of any test for this, but browsing the django code revealed no critical changes.

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -1688,7 +1688,9 @@ class Infotext(models.Model):
         return not (self.title or self.content)
 
 
-class UserProfileManager(BaseUserManager):
+class UserProfileManager(models.Manager["UserProfile"]):
+    normalize_email = BaseUserManager.normalize_email
+
     def create_user(self, *, email, password=None, first_name_given=None, last_name=None):
         user = self.model(email=self.normalize_email(email), first_name_given=first_name_given, last_name=last_name)
         validate_password(password, user=user)
--- a/evap/development/fixtures/test_data.json
+++ b/evap/development/fixtures/test_data.json
@@ -152737,7 +152737,7 @@
     "is_superuser": false,
     "password": "pbkdf2_sha256$12000$lRV0h1V7qLSK$pKzYyUhT7l65JxoIv7Du3btUHYzfGCZ81JFtbF9ymOY=",
     "last_login": "2014-03-24T10:25:43.302",
-    "email": "dot.carlisle@external.example.com",
+    "email": null,
     "title": "",
     "first_name_given": "Dot",
     "first_name_chosen": "",


evap/evaluation/models.py
Kakadus, line 1: I do want to note here that `get_by_natural_keys` is required by `django.contrib.auth.handers.modwsgi`, but we don't use it.

richardebeling (approved):

Kakadus:
Dö döööm. `create_super_user` tries to check if there is a `get_by_natural_key` method by checking if the email field is unique. Ours _is_ unique but also `null=True`

Kakadus:
Here is the code in question:

```py
    def _validate_username(self, username, verbose_field_name, database):
        """Validate username. If invalid, return a string error message."""
        if self.username_is_unique:
            try:
                self.UserModel._default_manager.db_manager(database).get_by_natural_key(
                    username
                )
            except self.UserModel.DoesNotExist:
                pass
            else: ...
```

which crashes when we do not have `get_by_natural_key` implemented. 

```py
    @cached_property
    def username_is_unique(self):
        if self.username_field.unique:
            return True
        return any(
            len(unique_constraint.fields) == 1
            and unique_constraint.fields[0] == self.username_field.name
            for unique_constraint in self.UserModel._meta.total_unique_constraints
        )
```
https://github.com/django/django/blob/main/django/contrib/auth/management/commands/createsuperuser.py#L291-L309

@niklasmohrin and I concluded that this is a django bug, I will open another ticket later.

Kakadus:
Let's wait and see for https://code.djangoproject.com/ticket/36225#ticket

richardebeling:
This is a bit stalled since it breaks `create_super_user` and we're not aware of any Django-compatible way to solve this. From my understanding, backup/restore should work with what we currently have (@Kakadus please correct me if this is wrong, then we need to fix it).

PR #2366, opened 2025-01-13
Average grades in export

### LINKED ISSUE #2001: Add averages to results exports

fidoriel:
closes #2001


evap/results/exporters.py
```diff
@@ -271,58 +285,118 @@ def write_overall_results(
...
+        avg_values = []
+        count_avg = 0
+        avg_approval = []
+        count_approval = 0
+
+        for __, results in evaluations_with_results:
+            if (
+                results.get(questionnaire_id) is None
+            ):  # ignore all results without the questionaire for average calculation
+                continue
```
richardebeling, line 334: From the code, it is very clear to me that this snippet skips results without a questionnaire. I'd much more appreciate the comment telling me 1. why this can happen in the first place (aren't all results always mapped to some question which needs to be part of some questionaire?) and 2. why we want to skip them

richardebeling (changes requested):

evap/results/exporters.py
```diff
@@ -271,58 +285,118 @@ def write_overall_results(
...
+            avg, percent_approval = cls._calculate_display_result(questionnaire_id, question, results)
```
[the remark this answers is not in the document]
janno42: It's not time critical but shouldn't take longer than ~10 seconds.

evap/results/exporters.py
```diff
@@ -271,58 +285,118 @@ def write_overall_results(
...
+            return None, None
+
+        avg = sum(values) / count_sum
+        if question.is_yes_no_question:
+            percent_approval = approval_count / count_sum if count_sum > 0 else 0
+            return avg, percent_approval
+        return avg, None
+
+    @classmethod
+    def _calculate_display_result_average(
```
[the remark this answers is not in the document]
fidoriel: One is using the other function. They look very similar but do very different things.

evap/results/exporters.py
```diff
@@ -249,17 +261,19 @@ def write_overall_results(
...
             # Borders only if there is a course grade below. Offset by one column
             self.write_empty_row_with_styles(
-                ["default"] + ["border_left_right" if gt1 else "default" for gt1 in count_gt_1]
+                ["default", "default"] + ["border_left_right" if gt1 else "default" for gt1 in count_gt_1]
             )
 
             self.write_cell(_("Evaluation weight"), "bold")
-            weight_percentages = (
+            self.write_cell("")
+            weight_percentages = tuple(
```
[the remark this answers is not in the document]
fidoriel: We need to write an empty cell for the average line

evap/results/exporters.py
```diff
@@ -112,24 +112,29 @@ def filter_text_and_heading_questions(questions: Iterable[Question]) -> list[Que
...
+
+        evaluations_filter = Q()
+        if semesters:
+            evaluations_filter &= Q(course__semester__in=semesters)
+        if evaluation_states:
+            evaluations_filter &= Q(state__in=evaluation_states)
+        if program_ids:
+            evaluations_filter &= Q(course__programs__in=program_ids)
+        if course_type_ids:
+            evaluations_filter &= Q(course__type__in=course_type_ids)
```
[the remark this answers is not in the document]
fidoriel: But then empty iterables are not covered, and the code breaks

evap/results/exporters.py
```diff
@@ -271,58 +285,118 @@ def write_overall_results(
             self.next_row()
 
             # Same reasoning as above.
-            self.write_empty_row_with_styles(["default"] + ["border_top" if gt1 else "default" for gt1 in count_gt_1])
+            self.write_empty_row_with_styles(
+                ["default", "default"] + ["border_top" if gt1 else "default" for gt1 in count_gt_1]
+            )
+
+    @classmethod
+    def _calculate_display_result(
```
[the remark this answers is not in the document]
fidoriel: think `_calculate_display_result` is accurate what it does. Maybe `_calculate_evaluation_ratio` and `_calculate_evaluation_average_ratio`?

evap/results/exporters.py
```diff
@@ -271,58 +285,118 @@ def write_overall_results(
...
+                all_evaluations_with_results, questionnaire.id, question
+            )
+
+            if question_average is not None:
+                if question.is_yes_no_question:
+                    self.write_cell(f"{question_approval_count:.0%}", self.grade_to_style(question_average))
+                else:
+                    self.write_cell(question_average, self.grade_to_style(question_average))
+            else:
+                self.write_cell("", "border_left_right")
```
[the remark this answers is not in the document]
fidoriel: Think its fine

evap/results/exporters.py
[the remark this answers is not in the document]
fidoriel: Tuple was something with typing.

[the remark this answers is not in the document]
niklasmohrin: The expected type below is `Iterable`, so `tuple` shouldn't be necessary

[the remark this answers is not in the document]
niklasmohrin: Maybe we should introduce a style "missing_average" or so to make these `write_cell` calls more clear? It could be identical to another style, but just to make it more readable

evap/results/exporters.py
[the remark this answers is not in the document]
niklasmohrin: The second one is an average of averages (why?). I also don't like the current names, it don't know what a "display result average" is, let's emphasize what the sample space of this average is

niklasmohrin (commented):
Seems mostly correct, but I think we should make an effort to keep the structure of these methods clear

evap/results/exporters.py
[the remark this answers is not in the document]
richardebeling: My thought here was that with addition being commutative and division being distributive, the "average of averages" should just be like computing the 1-level-average correctly.

I now see that the average-of-averages approach has different weights (-> it doesn't matter that on evaluation1, 50 people voted, while it was 200 people on evaluation2), which probably makes sense for the number we show in the exporter, so I'm fine with keeping the two methods. Agree with @niklasmohrin on the naming, I could see ` _get_average_of_average_grade_and_approval_ratio` working, don't know if we have something better.

evap/results/exporters.py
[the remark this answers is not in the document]
fidoriel: we are relying on this

see old code
```
evaluations_filter = Q(
    course__semester__in=semesters,
    state__in=evaluation_states,
    course__programs__in=program_ids,
    course__type__in=course_type_ids,
)
```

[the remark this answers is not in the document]
niklasmohrin: As discussed in person, I don't think this defeats Richard's any my argument

evap/results/exporters.py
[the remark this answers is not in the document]
fidoriel: renamed

evap/results/tests/test_exporters.py
```diff
@@ -590,3 +590,120 @@ def test_text_answer_export(self):
...
+            ],
+            True,
+            True,
+        )
+        binary_content.seek(0)
+        workbook = xlrd.open_workbook(file_contents=binary_content.read())
+
+        self.assertEqual(
+            float(workbook.sheets()[0].row_values(5)[1]),
+            (float(workbook.sheets()[0].row_values(5)[2]) + float(workbook.sheets()[0].row_values(5)[3])) / 2,
```
richardebeling, line 647: From a quick glance, it looks like the setup above creates the same exact average result for both evaluations and same count of answers inside of them. This doesn't test whether we compute the average-of-averages with correct weighting. Can we change them to have different answer counts?

Coincidentally, float-equality here made me wonder about correctness. It should work if the two values here are identical, the sum is then 2 * the original value, which is divisible by 2 to perfectly recreate the original value. However, I'd like the test to test a somewhat more sophisticated case. We'll likely need to use `assertAlmostEqual` then.

evap/results/tests/test_exporters.py
```diff
@@ -590,3 +590,120 @@ def test_text_answer_export(self):
...
+                )
+            ],
+            True,
+            True,
+        )
+        binary_content.seek(0)
+        workbook = xlrd.open_workbook(file_contents=binary_content.read())
+
+        self.assertEqual(float(workbook.sheets()[0].row_values(8)[1]), float(workbook.sheets()[0].row_values(8)[3]))
+        self.assertEqual("", workbook.sheets()[0].row_values(8)[2])
```
richardebeling, line 707: I may be tired, but I don't understand what we're asserting here. What is the intention here? Can we maybe improve the test name or add a comment?

evap/results/exporters.py
```diff
@@ -198,6 +203,8 @@ def write_headings_and_evaluation_info(
         else:
             self.write_cell(export_name, "headline")
 
+        self.write_cell("Average for this Question", "evaluation")
```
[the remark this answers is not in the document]
janno42: This is still not marked for translation, is it? I would also prefer to have something like "Average for this question over all evaluations in all published semesters" and please make it italic to distinguish it from the course names.

janno42 (changes requested):

evap/results/exporters.py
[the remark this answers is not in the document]
fidoriel: has been translated, look at the diff will tag you.

evap/results/exporters.py
```diff
@@ -198,6 +203,8 @@ def write_headings_and_evaluation_info(
         else:
             self.write_cell(export_name, "headline")
 
+        self.write_cell(_("Average for this Question"), "evaluation")
```
fidoriel, line 206: @richardebeling @janno42 here

evap/results/tests/test_exporters.py
richardebeling: @fidoriel This currently compares a value the system produces against another value that the system produces, which leaves the chance that both might be wrong, but the test still passes.

For a bit more assertiveness, I would appreciate if we could change the assertion here to the exact string that should be shown in the output, optimally in the string-domain (without to-number conversion)

evap/staff/views.py
```diff
@@ -1523,7 +1523,9 @@ def evaluation_person_management(request, evaluation_id: int):
             raise SuspiciousOperation("Invalid POST operation")
 
         import_action = ImportAction.from_operation(operation)
-        import_type: Final = ImportType.PARTICIPANT if "participants" in operation else ImportType.CONTRIBUTOR
+        import_type: Literal[ImportType.PARTICIPANT, ImportType.CONTRIBUTOR] = (
+            ImportType.PARTICIPANT if "participants" in operation else ImportType.CONTRIBUTOR
+        )
```
Kakadus, line 1528: revert

evap/results/tests/test_exporters.py
fidoriel: we asserting here that the heading question is also empty for averages

evap/results/tests/test_exporters.py
```diff
@@ -590,3 +590,68 @@ def test_text_answer_export(self):
...
+        self.assertAlmostEqual(workbook.sheets()[0].row_values(5)[1], 1.5833333333333335)
+
+        self.assertEqual(
+            workbook.sheets()[0].row_values(0)[1],
+            "Average for this question over all evaluations in all published semesters",
+        )
+
+        # test ragged questionaire
+        self.assertEqual(float(workbook.sheets()[0].row_values(8)[1]), float(workbook.sheets()[0].row_values(8)[3]))
+        self.assertEqual("", workbook.sheets()[0].row_values(8)[2])
```
richardebeling, line 657:
```suggestion
        # average for second questionnaire must be the value from evaluation2 (since evaluation1 doesn't have the questionnaire)
        self.assertEqual(float(workbook.sheets()[0].row_values(8)[1]), float(workbook.sheets()[0].row_values(8)[3]))
        # average field next to the questionnaire title must be empty
        self.assertEqual("", workbook.sheets()[0].row_values(8)[2])
```

richardebeling (dismissed):

evap/results/exporters.py
```diff
@@ -271,58 +288,118 @@ def write_overall_results(
...
+            if avg is not None:
+                avg_value_sum += avg
+                count_avg += 1
+            if average_approval_ratio is not None:
+                avg_approval_sum += average_approval_ratio
+                count_approval += 1
+
+        return avg_value_sum / count_avg if count_avg else None, (
+            avg_approval_sum / count_approval if count_approval else None
+        )
```
janno42, line 348: maybe like this?
```suggestion
        return (
            avg_value_sum / count_avg if count_avg else None,
            avg_approval_sum / count_approval if count_approval else None
        )
```

evap/results/tests/test_exporters.py
janno42: isn't `row_values(8)[2]` the second evaluation (which doesn't use this questionnaire, so the cell must be empty)?

evap/results/exporters.py
Kakadus: or just
```suggestion
        avg_value = avg_value_sum / count_avg if count_avg else None
        avg_approval = avg_approval_sum / count_approval if count_approval else None
        return avg_value, avg_approval
```

evap/results/exporters.py
```diff
@@ -271,58 +288,118 @@ def write_overall_results(
...
+    ) -> tuple[float | None, float | None]:
+        avg_value_sum = 0.0
+        count_avg = 0
+        avg_approval_sum = 0.0
+        count_approval = 0
+
+        for __, results in evaluations_with_results:
+            if (
+                results.get(questionnaire_id) is None
+            ):  # we iterate over all distinct questionaires from all evaluations but some evaluations do not include a specific questionaire
```
Kakadus, line 336:
```suggestion
            ):  # we iterate over all distinct questionnaires from all evaluations but some evaluations do not include a specific questionnaire
```

janno42 (dismissed):
📊

evap/evaluation/tools.py
```diff
@@ -228,6 +228,7 @@ def content(self, value):
 class ExcelExporter(ABC):
     styles = {
         "default": xlwt.Style.default_style,
+        "missing_average": xlwt.Style.default_style,
```
niklasmohrin, line 298: both styles should be defined in the same place, so this should be moved to the other class

evap/results/exporters.py
```diff
@@ -112,24 +115,29 @@ def filter_text_and_heading_questions(questions: Iterable[Question]) -> list[Que
...
-        course_type_ids: Iterable[int],
-        contributor: UserProfile | None,
-        include_not_enough_voters: bool,
+    def filter_evaluations(  # noqa: PLR0912
+        semesters: Iterable[Semester] | None = None,
+        evaluation_states: Iterable[Evaluation.State] | None = None,
+        program_ids: Iterable[int] | None = None,
+        course_type_ids: Iterable[int] | None = None,
+        contributor: UserProfile | None = None,
+        include_not_enough_voters: bool = False,
```
niklasmohrin, line 126: I don't think that any of these should have defaults, there are two callers and both of them know exactly what they want to filter for.

I was wondering if we could get away without None by passing in `Semester.objects.all()` which results in something like `WHERE "evaluation_course"."semester_id" IN (SELECT "evaluation_semester"."id" FROM "evaluation_semester")` - I guess the database should be able to see that this check can be avoided @richardebeling ? But having None is also fine for now, I just think that the caller should be explicit about this

For contributor and include_not_enough_voters, I am not even sure if the current version is correct - wouldn't we want to use answers across all contributors for the average? Currently, we only use answers to the general questionnaire. And I guess include_not_enough_voters should be true for the average (@janno42 ?)

evap/results/exporters.py
```diff
@@ -335,6 +412,8 @@ def export_impl(
         # We want to throw early here, since workbook.save() will throw an IndexError otherwise.
         assert len(selection_list) > 0
 
+        all_evaluations_with_results, _, _ = self.filter_evaluations(evaluation_states=[Evaluation.State.PUBLISHED])
+
         for sheet_counter, (program_ids, course_type_ids) in enumerate(selection_list, 1):
```
niklasmohrin, line 419: Just to be sure, we want to have the same average in all sheets and not the average for the programs and course types of _that_ sheet for every sheet? @janno42

evap/results/exporters.py
```diff
@@ -271,58 +288,118 @@ def write_overall_results(
...
 
             # Same reasoning as above.
-            self.write_empty_row_with_styles(["default"] + ["border_top" if gt1 else "default" for gt1 in count_gt_1])
+            self.write_empty_row_with_styles(
+                ["default", "default"] + ["border_top" if gt1 else "default" for gt1 in count_gt_1]
+            )
+
+    @classmethod
+    def _get_average_grade_and_approval_ratio(
+        cls, questionnaire_id: int, question: Question, results: OrderedDict[int, list[QuestionResult]]
```
niklasmohrin, line 299: This uses `questionnaire_id` and `results` only once to do `results[questionnaire_id]`, so we should just do that in the caller

evap/results/exporters.py
```diff
@@ -271,58 +288,118 @@ def write_overall_results(
...
+    ) -> tuple[float | None, float | None]:
+        value_sum = 0.0
+        count_sum = 0
+        approval_count = 0
+
+        for grade_result in results[questionnaire_id]:
+            if grade_result.question.id != question.id or not RatingResult.has_answers(grade_result):
+                continue
+
+            value_sum += grade_result.average * grade_result.count_sum
```
niklasmohrin, line 309: I think it would be fair to just ask the AnsweredRatingResult for the sum of grades instead of having it divided and then multiplied again; we can thus extract the summing out of its average method and make it available

evap/results/exporters.py
```diff
@@ -271,58 +288,118 @@ def write_overall_results(
...
+            count_sum += grade_result.count_sum
+            if grade_result.question.is_yes_no_question:
+                approval_count += grade_result.approval_count
+
+        if not value_sum:
+            return None, None
+
+        avg = value_sum / count_sum
```
niklasmohrin, line 317: I think that the code would be a lot easier to understand if we would give more precise names to the different averages; this here could be `grade_average` and the "value" could be a "grade" as well

evap/results/exporters.py
```diff
@@ -271,58 +288,118 @@ def write_overall_results(
...
+            if grade_result.question.id != question.id or not RatingResult.has_answers(grade_result):
+                continue
+
+            value_sum += grade_result.average * grade_result.count_sum
+            count_sum += grade_result.count_sum
+            if grade_result.question.is_yes_no_question:
+                approval_count += grade_result.approval_count
+
+        if not value_sum:
+            return None, None
```
niklasmohrin: What is the reason for this? I would expect to see
```suggestion
        if not count_sum:
            return None, None
```
Please change it or add a comment explaining why value_sum is correct :)

evap/results/exporters.py
```diff
@@ -271,58 +288,118 @@ def write_overall_results(
...
+        count_sum = 0
+        approval_count = 0
+
+        for grade_result in results[questionnaire_id]:
+            if grade_result.question.id != question.id or not RatingResult.has_answers(grade_result):
+                continue
+
+            value_sum += grade_result.average * grade_result.count_sum
+            count_sum += grade_result.count_sum
+            if grade_result.question.is_yes_no_question:
```
niklasmohrin:
```suggestion
            if question.is_yes_no_question:
```

evap/results/exporters.py
```diff
@@ -271,58 +288,118 @@ def write_overall_results(
...
+                results.get(questionnaire_id) is None
+            ):  # we iterate over all distinct questionnaires from all evaluations but some evaluations do not include a specific questionnaire
+                continue
+            avg, average_approval_ratio = cls._get_average_grade_and_approval_ratio(questionnaire_id, question, results)
+            if avg is not None:
+                avg_value_sum += avg
+                count_avg += 1
+            if average_approval_ratio is not None:
+                avg_approval_sum += average_approval_ratio
+                count_approval += 1
```
niklasmohrin, line 346: Since we are doing this in three places now, I think we could make this easier by introducing a small helper class `Averager` which would be used like
```python
if avg is not None:
    grade_averager.record_value(avg)
if average_approval_ratio is not None:
    approval_ratio_averager.record_value(approval_ratio)

# ...

return grade_averager.current_average(), approval_ratio_averager.current_average()
```

The same class could be used in _get_average_grade_and_approval_ratio with another method `record_summed_values(value_sum=..., value_count=...)` (or `record_value(..., count=...)` if we call it for each grade separately) to also eliminate the counting business over there

evap/results/exporters.py
```diff
@@ -198,6 +206,8 @@ def write_headings_and_evaluation_info(
         else:
             self.write_cell(export_name, "headline")
 
+        self.write_cell(_("Average for this question over all evaluations in all published semesters"), "average")
```
niklasmohrin: Isn't it more like this? @janno42 
```suggestion
        self.write_cell(_("Average result for this question over all published evaluations in all semesters"), "average")
```

evap/results/exporters.py
```diff
@@ -208,17 +218,19 @@ def write_headings_and_evaluation_info(
 
         self.next_row()
         self.write_cell(_("Programs"), "bold")
+        self.write_cell("", "program")
```
niklasmohrin: Please add such a comment to all relevant lines (please go through the entire PR and check all added lines like this)
```suggestion
        self.write_cell("", "program")  # empty cell in grade-average column
```

evap/results/exporters.py
```diff
@@ -208,17 +218,19 @@ def write_headings_and_evaluation_info(
...
         for evaluation, __ in evaluations_with_results:
             self.write_cell("\n".join([d.name for d in evaluation.course.programs.all()]), "program")
 
         self.next_row()
         self.write_cell(_("Course Type"), "bold")
+        self.write_cell("", "program")
```
niklasmohrin: The style here does not match the one in the next cell
```suggestion
        self.write_cell("", "border_left_right")  # empty cell in grade-average column
```

evap/results/exporters.py
```diff
@@ -271,58 +288,118 @@ def write_overall_results(
...
         evaluations_with_results: list[tuple[Evaluation, OrderedDict[int, list[QuestionResult]]]],
         contributor: UserProfile | None,
+        all_evaluations_with_results: list[tuple[Evaluation, OrderedDict[int, list[QuestionResult]]]],
     ) -> None:
         if contributor and questionnaire.type == Questionnaire.Type.CONTRIBUTOR:
             self.write_cell(f"{questionnaire.public_name} ({contributor.full_name})", "bold")
         else:
             self.write_cell(questionnaire.public_name, "bold")
 
+        self.write_cell("", "border_left_right")
```
niklasmohrin, line 364: this can just be merged with the next line (where the comment is also not correct anymore)

evap/results/exporters.py
```diff
@@ -271,58 +288,118 @@ def write_overall_results(
...
+
         # first cell of row is printed above
         self.write_empty_row_with_styles(["border_left_right"] * len(evaluations_with_results))
 
         for question in self.filter_text_and_heading_questions(questionnaire.questions.all()):
             self.write_cell(question.text, "italic" if question.is_heading_question else "default")
 
+            average_grade, approval_ratio = self._get_average_of_average_grade_and_approval_ratio(
```
niklasmohrin, line 372: together with `avg` and `average_approval_ratio` below, it's not really clear what the difference between those four variables is; I think we can should use a clearer name

evap/results/tests/test_exporters.py
```diff
@@ -589,3 +589,69 @@ def test_text_answer_export(self):
         self.assertEqual(sheet.row_values(3)[0], questions[0].text)
         self.assertEqual(sheet.row_values(5)[0], questions[1].text)
         self.assertEqual(sheet.row_values(6)[0], questions[2].text)
+
+    def test_total_average(self):
```
niklasmohrin, line 610: this does not cover several important details:
1. evaluations from previous semesters are included
2. evaluations from other programs and course types are included
3. only published evaluations are included
  a. even if the unpublished evaluations are part of the export
4. which contributions are used
5. we are averaging over the averages of several results
6. are results with not enough voters used

So this test should be extended / more tests should be added. I think it is very easy to mess up some of these details in the future and when that happens, these tests should catch it

niklasmohrin (changes requested):
The output looks correct, but I think we need to spend some more time on the code and testing, otherwise it will become hard to maintain the exporter

evap/results/exporters.py
janno42: I'm confused. Aren't there average values for contributor questions in the current version of this PR? I thought so.

But yes, `include_not_enough_voters` should be `True` for the averages.

evap/results/exporters.py
janno42: Now that you mention it... I think I wouldn't add this filter if we'd just have different course types, because their questions often differ. But for programs it might actually be important to have different averages. And then, when we're already at it, we can do the separation for both.
So yes, please calculate the averages for evaluations within the programs and course types of that sheet only.

evap/results/exporters.py
niklasmohrin: What I take from https://github.com/fidoriel/EvaP/blob/8a736d0713823104b69910edfbf23cfe49ab1ce4/evap/results/exporters.py#L415 is that even when `contributor` is not None we use the default of None for `filter_evaluations`

janno42:
replaced by #2708

PR #2367, opened 2025-01-13
Main language for evaluations

hansegucker:


evap/contributor/tests/test_views.py
```diff
@@ -292,6 +293,25 @@ def test_display_request_buttons(self):
...
+            form["main_language"] = "x"
+
+        for lang, __ in settings.LANGUAGES:
+            form["main_language"] = lang
+            form.submit(name="operation", value="save")
+            self.assertEqual(Evaluation.objects.get().main_language, lang)
+
+        with self.assertRaises(ValueError):
+            form["main_language"] = "some_other_wrong_value"
+            form.submit(name="operation", value="save")
```
richardebeling, line 313: Like the "Invalid value doesn't work"-block in 303, I think this will already fail on setting the form field as django webtest throws when we try to set a value that is not in the choices list. I think testing that server-side validation works is a good idea here, not sure if it can be done with WebTest form framework though. I think in the past we always manually crafted POST requests to achieve this.

evap/evaluation/migrations/0148_evaluation_main_language.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 5.1.3 on 2024-11-25 20:19
+
+from django.db import migrations, models
+
+
+def _migrate(apps, schema_editor):
+    Evaluation = apps.get_model("evaluation", "Evaluation")
+    for evaluation in Evaluation.objects.filter(state__gte=40):
+        evaluation.main_language = "de"
```
richardebeling, line 9: Why? I would have expected "undecided" (which is the default value -> no explicit migration code needed?)

(If we keep this, can we use some named variable instead of the magic 40?)

evap/evaluation/models.py
```diff
@@ -759,15 +771,20 @@ def can_publish_rating_results(self):
...
+    )
     def editor_approve(self):
         pass
 
     @transition(
         field=state,
         source=[State.NEW, State.PREPARED, State.EDITOR_APPROVED],
         target=State.APPROVED,
-        conditions=[lambda self: self.general_contribution_has_questionnaires],
+        conditions=[lambda self: self.general_contribution_has_questionnaires and self.has_valid_main_language],
```
richardebeling, line 787: The issue says that staff users should be able to set "undecided". I think this change here would block that workflow?

evap/staff/tests/test_views.py
```diff
@@ -1920,6 +1920,38 @@ def test_evaluation_create(self):
         form.submit()
         self.assertEqual(Evaluation.objects.get().name_de, "lfo9e7bmxp1xi")
 
+    def _set_valid_form(self, form):
```
richardebeling, line 1923: factoring out the helper is nice -- but we should also use it in `test_evaluation_create` then :D

evap/staff/tests/test_views.py
```diff
@@ -2349,6 +2381,19 @@ def test_state_change_log_translated(self, trans):
...
+        form = response.forms["evaluation-form"]
+
+        form["main_language"] = "x"
+        response = form.submit("operation", value="approve")
+        self.assertContains(response, "You have to set a main language to approve this evaluation.")
+
+        self.assertNotEqual(Evaluation.objects.first().state, self.evaluation.State.APPROVED)
+
+        form["main_language"] = "en"
+        form.submit("operation", value="approve")
```
richardebeling, line 2395: We probably have to add an assertion that the response now doesn't contain the above error anymore (our views return 200 even if there is a business logic error / form validation failed)

evap/staff/views.py
```diff
@@ -1330,17 +1330,19 @@ def notify_reward_points(grantings, **_kwargs):
...
     )
 
-    evaluation_form = EvaluationForm(request.POST or None, instance=evaluation, semester=evaluation.course.semester)
+    operation = request.POST.get("operation")
+
+    if request.method == "POST" and operation not in ("save", "approve"):
+        raise SuspiciousOperation("Invalid POST operation")
+
+    evaluation_form = EvaluationForm(
+        request.POST or None, instance=evaluation, semester=evaluation.course.semester, operation=operation
```
richardebeling, line 1339: random consideration: A narrower, more precise interface would be passing `require_main_language: bool` instead of passing an operation string that implicitly allows to infer that bool

evap/student/templates/student_vote.html
```diff
@@ -51,7 +51,26 @@
...
+        <div class="btn-switch btn-switch-light float-end">
+            <div class="btn-switch-label my-auto">{% trans "Questionnaire language" %}</div>
+            <div class="btn-switch btn-group">
+                {% for language_code, language_name in languages %}
+                    {% csrf_token %}
+                    <a data-bs-toggle="tooltip"
+                       data-bs-placement="bottom"
+                       title="{{ language_name }}"
+                       role="button"
+                       href="{% url 'student:vote' evaluation.pk %}?language={{ language_code }}"
```
richardebeling, line 66: This should probably be a POST-form, not a GET-anchor. (Currently, this wouldn't need the CSRF token)

richardebeling: Ah, I assumed that this would save the language so that if a user came back later to the same evaluation, they would still go back to the language they last selected.

If we keep the language stateless, the anchor is fine here, but then we don't need the csrf_token above.

evap/evaluation/migrations/0148_evaluation_main_language.py
```diff
@@ -0,0 +1,30 @@
...
+    dependencies = [
+        ("evaluation", "0147_unusable_password_default"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="evaluation",
+            name="main_language",
+            field=models.CharField(
+                choices=[("en", "English"), ("de", "Deutsch"), ("x", "undecided")],
```
Kakadus, line 23: I want to note that now a new migration is needed when the user changes the languages in the settings. Is it okay that this is still a setting then?

If not, then this can probably become a `models.TextChoices`

niklasmohrin:
As far as I understand, this waits for #2377, so should we mark this one here as draft for now?

evap/student/templates/student_vote.html
janno42: we should save the language in localstorrage but not on the server

evap/evaluation/migrations/0148_evaluation_main_language.py
hansegucker: We want each published evaluation to have a decided main language. We check this during state transitions. So, we have to set a reasonable default for all evaluations which changing the main language is no longer possible.

richardebeling: Ok, sounds reasonable. Could we still replace `40` here with some named constant?

evap/evaluation/migrations/0148_evaluation_main_language.py
richardebeling: Yes, but I would do that separately from this PR

janno42 (changes requested):
Works almost as expected 👍 

One thing though: If the language is still "undecided", contributors will see the language field in the evaluation form filled in with the first option ("English"). Instead, they should see that no language has been chosen yet, and should deliberately choose a language.

hansegucker:
> One thing though: If the language is still "undecided", contributors will see the language field in the evaluation form filled in with the first option ("English"). Instead, they should see that no language has been chosen yet, and should deliberately choose a language.

Done.

evap/staff/tests/test_views.py
hansegucker: Which error do you mean?

richardebeling: > You have to set a main language to approve this evaluation.")

from line 2437

evap/static/ts/src/sisyphus.LICENSE
niklasmohrin, line 1: delete

evap/evaluation/migrations/0148_evaluation_main_language.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 5.1.3 on 2024-11-25 20:19
+
+from django.db import migrations, models
+
+
+def _migrate(apps, _schema_editor):
+    _approved_status = 40
+    Evaluation = apps.get_model("evaluation", "Evaluation")
+    Evaluation.objects.filter(state__gte=_approved_status).update(main_language="de")
```
niklasmohrin, line 9: Should the default really be German here? @janno42 ?

evap/evaluation/models.py
```diff
@@ -624,6 +632,10 @@ def is_in_evaluation_period(self):
     def general_contribution_has_questionnaires(self):
         return self.general_contribution and self.general_contribution.questionnaires.count() > 0
 
+    @property
+    def has_valid_main_language(self):
```
niklasmohrin, line 636: How do you feel about this, so that we don't add any more vocabulary?
```suggestion
    def has_decided_main_language(self):
```

evap/evaluation/models.py
```diff
@@ -759,15 +771,20 @@ def can_publish_rating_results(self):
     def ready_for_editors(self):
         pass
 
-    @transition(field=state, source=State.PREPARED, target=State.EDITOR_APPROVED)
+    @transition(
+        field=state,
+        source=State.PREPARED,
+        target=State.EDITOR_APPROVED,
+        conditions=[lambda self: self.has_valid_main_language],
```
niklasmohrin, line 778: Should we enforce this with a constraint? I think it would be a sensible thing to do, but it would also mean changing a lot of tests

evap/student/views.py
```diff
@@ -22,6 +23,7 @@
     get_evaluations_with_course_result_attributes,
     textanswers_visible_to,
 )
+from evap.settings import LANGUAGES
```
niklasmohrin, line 26: please use django.conf.settings

evap/staff/tests/test_views.py
```diff
@@ -1958,6 +1963,19 @@ def test_evaluation_create(self):
...
+    def test_evaluation_create_main_language(self):
+        response = self.app.get(self.url_for_semester, user=self.manager, status=200)
+        form = response.forms["evaluation-form"]
+        self._set_valid_form(form)
+
+        with self.assertRaises(ValueError):
+            form["main_language"] = "some_wrong_value"
+            form.submit()
+
+        form["main_language"] = "x"
```
niklasmohrin, line 1975: Can we make the "x" a constant in Evaluation, something like UNDECIDED_LANGUAGE?

niklasmohrin (commented):
haven't taken a closer look at the complicated parts yet, but here are some easy things I spotted already

evap/evaluation/models.py
richardebeling: Constraint would be good, but let's pull this out from this PR, it's fine to merge without the constraint I think.

I think we have some other fields as well that need to be set up in a specific way in order to get "sensible" instances. In the long run, it would be nice to have those filled automatically to simplify tests. I could see us switching to [recipes](https://model-bakery.readthedocs.io/en/latest/recipes.html) or maybe adding some custom `model_bakery_default` members to the fields (but we would need to [patch](https://model-bakery.readthedocs.io/en/latest/how_bakery_behaves.html#customizing-baker) model bakery to respect them). Not a trivial thing, I don't want to mix it up in this PR.

evap/evaluation/migrations/0148_evaluation_main_language.py
janno42: Please change it to English

evap/student/templates/student_vote.html
```diff
@@ -200,6 +221,18 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
                 bottomCard: document.querySelector("#bottom_text_results_publish_confirmation_card"),
             };
 
+            // Ensure that selected questionnaire language is saved and loaded
+            const languageStorageKey = "student-vote-language-{{ evaluation.id }}-{{ request.user.id }}";
+            let params = new URLSearchParams(document.location.search);
+            let currentlySelectedLanguage = params.get("language") || "";
+            let savedLanguage = localStorage.getItem(languageStorageKey);
```
richardebeling, line 228: I think these three can all be `const`

richardebeling (commented):
Looking pretty good, thanks :)

janno42 (approved):
🗣️ Nice!

janno42 (approved):

evap/contributor/tests/test_views.py
```diff
@@ -272,6 +274,26 @@ def test_display_request_buttons(self):
...
+        self.assertContains(page, "A decision on the main language has to be made.")
+
+        for lang, __ in settings.LANGUAGES:
+            form["main_language"] = lang
+            form.submit(name="operation", value="save")
+            self.assertEqual(Evaluation.objects.get().main_language, lang)
+
+        with self.assertRaises(ValueError):
+            form["main_language"] = "some_other_wrong_value"
+            form.submit(name="operation", value="save")
```
niklasmohrin, line 295: The raise is already on the assignment before
```suggestion
```

This raises (ha) the question whether we want to have a test that actually carries out the request; I think it's fine as is, the only pro argument for me would be coverage

evap/development/fixtures/test_data.json
niklasmohrin, line 1: Can we have some of them be German?

evap/evaluation/migrations/0148_evaluation_main_language.py
```diff
@@ -0,0 +1,29 @@
...
+            model_name="evaluation",
+            name="main_language",
+            field=models.CharField(
+                choices=[("en", "English"), ("de", "Deutsch"), ("x", "undecided")],
+                default="x",
+                max_length=2,
+                verbose_name="main language",
+            ),
+        ),
+        migrations.RunPython(_migrate),
```
niklasmohrin, line 28: To make this migration reversible:
```suggestion
        migrations.RunPython(_migrate, reverse_code=migrations.RunPython.noop),
```

evap/evaluation/models.py
```diff
@@ -425,6 +427,14 @@ class State(models.IntegerChoices):
     name_en = models.CharField(max_length=1024, verbose_name=_("name (english)"), blank=True)
     name = translate(en="name_en", de="name_de")
 
+    # questionaire is shown in this language per default
```
niklasmohrin, line 430:
```suggestion
    # questionnaires are shown in this language by default
```

evap/evaluation/models.py
niklasmohrin, line 1: @richardebeling The state transition `skip_review_single_result` doesn't have `self.has_decided_main_language` as a condition; I suppose this is fine because we are never showing questionnaires for such evaluations. We will have to clean those up when we introduce a database constraint but it's no problem, because we need a migration then anyways. No action needed now IMO

evap/staff/tests/test_views.py
```diff
@@ -1958,6 +1963,19 @@ def test_evaluation_create(self):
...
         self.assertEqual(Evaluation.objects.get().name_de, "lfo9e7bmxp1xi")
 
+    def test_evaluation_create_main_language(self):
+        response = self.app.get(self.url_for_semester, user=self.manager, status=200)
+        form = response.forms["evaluation-form"]
+        self._set_valid_form(form)
+
+        with self.assertRaises(ValueError):
+            form["main_language"] = "some_wrong_value"
+            form.submit()
```
niklasmohrin, line 1973: Same here, the second line is never reached
```suggestion
        with self.assertRaises(ValueError):
            form["main_language"] = "some_wrong_value"
```

evap/staff/forms.py
```diff
@@ -385,6 +387,7 @@ class Meta:
 
     def __init__(self, *args, **kwargs):
         semester = kwargs.pop("semester", None)
+        self.operation = kwargs.pop("operation", "save")
```
niklasmohrin, line 390: I think this is clearer as a `operation: Literal["approve", "save"] = "save"` argument, outside of `kwargs` - in fact, should there be a default? Because this is effectively a toggle for one of the checks, it's a bit odd to have it off-by-default

evap/staff/tests/test_views.py
```diff
@@ -2396,6 +2415,19 @@ def test_state_change_log_translated(self, trans):
...
+        form = response.forms["evaluation-form"]
+
+        form["main_language"] = "x"
+        response = form.submit("operation", value="approve")
+        self.assertContains(response, "You have to set a main language to approve this evaluation.")
+
+        self.assertNotEqual(Evaluation.objects.first().state, self.evaluation.State.APPROVED)
+
+        form["main_language"] = "en"
+        response = form.submit("operation", value="approve", status=302)
```
niklasmohrin, line 2429: The function just ends here without touch `response` again - can you add one more assert?

evap/student/views.py
```diff
@@ -204,7 +205,9 @@ def get_vote_page_form_groups(request, evaluation, preview):
 
 
 def render_vote_page(request, evaluation, preview, for_rendering_in_modal=False):
-    form_groups = get_vote_page_form_groups(request, evaluation, preview)
+    language = request.GET.get("language", evaluation.main_language)
+    with override(language):
```
niklasmohrin, line 209: Nit: All our current uses are `translation.override(...)`, which I also find clearer

evap/student/templates/student_vote.html
```diff
@@ -51,7 +51,26 @@
             </small>
         </div>
     {% endif %}
-    <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</h3>
+    <h3 class="mb-3">
+        {{ evaluation.full_name }} ({{ evaluation.course.semester.name }})
+
+        <div class="btn-switch btn-switch-light float-end">
+            <div class="btn-switch-label my-auto">{% trans "Questionnaire language" %}</div>
```
niklasmohrin, line 58:
```suggestion
            <div class="btn-switch-label my-auto">{% translate "Questionnaire language" %}</div>
```

evap/student/templates/student_vote.html
niklasmohrin: csrf token is still there

evap/student/templates/student_vote.html
```diff
@@ -51,7 +51,26 @@
...
+            <div class="btn-switch btn-group">
+                {% for language_code, language_name in languages %}
+                    {% csrf_token %}
+                    <a data-bs-toggle="tooltip"
+                       data-bs-placement="bottom"
+                       title="{{ language_name }}"
+                       role="button"
+                       href="{% url 'student:vote' evaluation.pk %}?language={{ language_code }}"
+                       class="btn btn-sm btn-light{% if evaluation_language == language_code %} active{% endif %}">
+                        <span id="span-set-language-{{ language_code }}">{{ language_code|upper }}</span>
```
niklasmohrin, line 68: This id is duplicated with the navbar

evap/student/templates/student_vote.html
```diff
@@ -200,6 +221,18 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
                 bottomCard: document.querySelector("#bottom_text_results_publish_confirmation_card"),
             };
 
+            // Ensure that selected questionnaire language is saved and loaded
+            const languageStorageKey = "student-vote-language-{{ evaluation.id }}-{{ request.user.id }}";
+            const params = new URLSearchParams(document.location.search);
+            const currentlySelectedLanguage = params.get("language") || "";
+            const savedLanguage = localStorage.getItem(languageStorageKey);
+            if (currentlySelectedLanguage) {
+                localStorage.setItem(languageStorageKey, currentlySelectedLanguage);
```
niklasmohrin, line 230: These entries don't expire automatically, should we do something about it? @richardebeling

evap/student/templates/student_vote.html
```diff
@@ -200,6 +221,18 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
                 bottomCard: document.querySelector("#bottom_text_results_publish_confirmation_card"),
             };
 
+            // Ensure that selected questionnaire language is saved and loaded
```
niklasmohrin, line 224: Maybe not for this PR, but it would probably be worthwhile to test this with selenium

niklasmohrin (changes requested):
Looks mostly good :+1:

Some details that still need figuring out:

evap/student/templates/student_vote.html
niklasmohrin: Yes, also no blocker for me. Not sure how to best approach this, but either way, let's do something about it in a separate issue

evap/student/templates/student_vote.html
niklasmohrin: Yes, I have considered this :D Yeah Richard, why don't we have this?

evap/student/templates/student_vote.html
hansegucker: Yeah, probably a good follow-up issue.

evap/contributor/tests/test_views.py
richardebeling: still an issue with the current code :D

evap/staff/tests/test_views.py
hansegucker: There is notting to assert as it's a redirect.

evap/staff/forms.py
hansegucker: That doesn't work due to the forms structure

evap/staff/forms.py
```diff
@@ -459,6 +462,12 @@ def clean_weight(self):
             self.add_error("weight", _("At least one evaluation of the course must have a weight greater than 0."))
         return weight
 
+    def clean_main_language(self):
+        main_language = self.cleaned_data.get("main_language")
+        if self.requires_decided_main_language and main_language == "x":
```
richardebeling: Do we want to put the string-constant `"x"` into some constant variable (so every occurrence is easy to find and we can change it easily if we want)? I think it would be a good idea.

richardebeling (approved):

richardebeling (dismissed):

evap/staff/forms.py
```diff
@@ -459,6 +462,12 @@ def clean_weight(self):
             self.add_error("weight", _("At least one evaluation of the course must have a weight greater than 0."))
         return weight
 
+    def clean_main_language(self):
+        main_language = self.cleaned_data.get("main_language")
+        if self.requires_decided_main_language and main_language == Evaluation.UNDECIDED_MAIN_LANGUAGE:
+            self.add_error("main_language", _("You have to set a main language to approve this evaluation."))
```
niklasmohrin: This form does not know about approving evaluations, so we should make this message less specific. If we _really_ want to mention this approving here, we should change the `requires_decided_main_language` to something like `decided_main_language_reason: StrOrPromise | None` and pass in this string from the outside

evap/student/templates/student_vote.html
```diff
@@ -51,7 +51,25 @@
...
+            <div class="btn-switch-label my-auto">{% translate "Questionnaire language" %}</div>
+            <div class="btn-switch btn-group">
+                {% for language_code, language_name in languages %}
+                    <a data-bs-toggle="tooltip"
+                       data-bs-placement="bottom"
+                       title="{{ language_name }}"
+                       role="button"
+                       href="{% url 'student:vote' evaluation.pk %}?language={{ language_code }}"
+                       class="btn btn-sm btn-light{% if evaluation_language == language_code %} active{% endif %}">
+                        <span id="span-set-questionaire-language-{{ language_code }}">{{ language_code|upper }}</span>
```
niklasmohrin: The id here seems to be unused. For the language picker in the nav bar, we use such ids to trigger a spinning icon until the page reloads. The code to add this icon seems to be missing

niklasmohrin (changes requested):
Very cool :)

evap/student/templates/student_vote.html
```diff
@@ -200,6 +220,18 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
                 bottomCard: document.querySelector("#bottom_text_results_publish_confirmation_card"),
             };
 
+            // Ensure that selected questionnaire language is saved and loaded
```
janno42, line 243: when a language is stored in localstorage, the page always double-loads, even if the correct language is already shown.

janno42 (changes requested):

richardebeling (dismissed):
current code lgtm once https://github.com/e-valuation/EvaP/pull/2367/files#r2190829870 is fixed

evap/student/templates/student_vote.html
```diff
@@ -60,11 +60,12 @@ <h3 class="mb-3">
                 {% for language_code, language_name in languages %}
                     <a data-bs-toggle="tooltip"
                        data-bs-placement="bottom"
+                       data-set-spinner-icon="span-set-evaluation-language-{{ language_code }}"
```
janno42: The button gets a bit smaller when the spinner is shown - so the language selection jumps a bit. The button should keep its size.

janno42 (dismissed):
🌍

evap/student/templates/student_vote.html
```diff
@@ -90,77 +109,79 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
-                            {% translate 'Show my text answers to the contributors even if I remain the only person who takes part in this evaluation.' %}
-                        </label>
+            {% endif %}
+            {% if small_evaluation_size_warning and not preview and evaluation.num_voters == 0 %}
+                <div class="callout callout-warning mb-3" id="bottom_text_results_publish_confirmation_card">
+                    <div class="callout-header">
+                        <span class="fas fa-triangle-exclamation"></span> {% translate 'Small number of participants' %}
+                    </div>
+                    <div class="callout-content tab-row">
+                        {% blocktranslate %}You're the first person taking part in this evaluation. If you want your text answers to be shown to the contributors even if you remain the only person taking part, please check the box below. Otherwise your text answers will be deleted if no other person takes part in the evaluation.{% endblocktranslate %}
```
niklasmohrin: This warning is inside the `{{ language ... }}` block, while the one at the top of the site is not. This means that the same warning can be displayed in two different languages.

In general, it is a bit odd that the between the language toggle for the evaluation and the evaluation form there is still a warning and the "Thank you for taking part ..." message which are displayed in the site-wide language. Should we just translate those too? @janno42

niklasmohrin (changes requested):
Everything else looks good :)

evap/student/templates/student_vote.html
janno42: We specifically decided that only the questionnaire section should be affected, the rest of the page stays in the selected page language. The fix in [7ecc418](https://github.com/e-valuation/EvaP/pull/2367/commits/7ecc418200e20c542a1feda64786d8db92698a1a) is fine.

evap/static/scss/components/_buttons.scss
```diff
@@ -136,6 +136,9 @@ a:not([href]):not(.disabled) {
 }
 
 .btn-switch-evaluation-language {
+    @include media-breakpoint-down(sm) {
+        width: 100%;
+    }
```
janno42: I think you can remove this part again, it's okay to keep the original width.
It looks a bit weird in a wide sm just before the breakpoint and we don't have it from md onwards anyway and it still looks fine.
I would prefer having it right-aligned though.

sm:
image
md:
image

hansegucker: Done

richardebeling (dismissed):

evap/contributor/forms.py
```diff
@@ -112,6 +113,12 @@ def clean_vote_end_date(self):
 
         return vote_end_date
 
+    def clean_main_language(self):
+        main_language = self.cleaned_data.get("main_language")
+        if main_language == Evaluation.UNDECIDED_MAIN_LANGUAGE:
+            self.add_error("main_language", _("A decision on the main language has to be made."))
```
niklasmohrin: Don't know if we care, but the staff and contributor forms have different error strings for this check

evap/evaluation/migrations/0151_evaluation_main_language.py
niklasmohrin, line 1: we need to make sure that this doesn't conflict with other PRs adding migrations - there are a lot of such PRs at the moment and I think we wanted to merge #2484 first

evap/student/views.py
```diff
@@ -252,7 +253,9 @@ def render_vote_page(
     dropout: bool,
     for_rendering_in_modal: bool = False,
 ) -> HttpResponse:
-    form_groups = get_vote_page_form_groups(request, evaluation, preview=preview, preselect_no_answer=dropout)
+    language = request.GET.get("language", evaluation.main_language)
+    with translation.override(language):
+        form_groups = get_vote_page_form_groups(request, evaluation, preview=preview, preselect_no_answer=dropout)
```
niklasmohrin, line 257: So one thing that could potentially go wrong here is
```python
>>> with translation.override("de"):
...     s = translation.gettext("evaluation")
...
>>> s
'Evaluierung'
>>> with translation.override("de"):
...     s = translation.gettext_lazy("evaluation")
...
>>> s
'evaluation'
```
It doesn't seem to be a problem today, but might bite us in the future. Do we think we should do something about it, or just leave it be for now? @richardebeling @hansegucker 

If we ever implement screenshot tests, we could screenshot a page with a main language other than the site language. For now, I guess we can add a test that calls `get_vote_page_form_groups` with a mocked `gettext_lazy` and ensures it isn't called?

niklasmohrin: Ah, or is it actually not a problem because we are still rendering things inside the `{% language ... %}` environment?

evap/student/templates/student_vote.html
```diff
@@ -217,6 +240,19 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
                 bottomCard: document.querySelector("#bottom_text_results_publish_confirmation_card"),
             };
 
+            // Ensure that selected questionnaire language is saved and loaded
+            const languageStorageKey = "student-vote-language-{{ evaluation.id }}-{{ request.user.id }}";
+            const params = new URLSearchParams(document.location.search);
+            const currentlySelectedLanguage = "{{ evaluation_language }}";
```
niklasmohrin, line 246: This is currently using the auto-escaping which is for HTML, not javascript. I couldn't come up with an example that breaks because of this, but I think the correct way to do this is
```suggestion
            const currentlySelectedLanguage = "{{ evaluation_language|escapejs }}";
```
The ids above are not user input and we know that they don't contain any dangerous characters, so HTML escaping should be fine. The Django docs say that it's preferred to use data attributes in the DOM instead of interpolating scripts like this. I don't have a strong opinion on which is better, eventually all this code should be Typescript anyways.

By the way, why isn't this in `student-vote.ts`? :sweat_smile: It's fine by me, we don't need to drag this PR out longer because of JS/TS

evap/contributor/forms.py
janno42: Please use the string from the staff form for both

evap/student/templates/student_vote.html
niklasmohrin: The comment was closed and PR merged without this included? :( @janno42 @hansegucker

janno42: You said that you don't have a strong opinion 😅 
Moving the code to student-vote.ts is a nice new issue that I can write in the next days.

niklasmohrin: I meant between escapejs and using data attributes 😅 We should do one of those two, not what we are currently doing

TS as a new issue sounds good

janno42: Fixed: I added `|escapejs` for now.

PR #2374, opened 2025-01-20
Split `settings.BASE_DIR` into `MODULE` and `DATADIR`

niklasmohrin:
This is in preparation for #2328

The idea is that we don't want to place any files at runtime into `MODULE`, as that might be `venv/lib/python3.10/site-packages/evap/`, or a path into the Nix store or something similar.

Best reviewed by commit.


niklasmohrin:
@janno42 This could require small changes to the settings on production

evap/development/fixtures/upload/grades/1577/test_results_03.txt
Kakadus, line 1: why do we even have those in the first place?

Kakadus (commented):
looks good in general

evap/evaluation/tests/test_misc.py
```diff
@@ -29,7 +29,7 @@ def test_sample_semester_file(self):
         original_user_count = UserProfile.objects.count()
 
         form = page.forms["semester-import-form"]
-        form["excel_file"] = (os.path.join(settings.BASE_DIR, "static", "sample.xlsx"),)
+        form["excel_file"] = (os.path.join(settings.MODULE, "static", "sample.xlsx"),)
```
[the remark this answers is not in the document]
niklasmohrin: Yeah, I was thinking about that too, but I figured this would also be sensible to do in another PR

evap/development/fixtures/upload/grades/1577/test_results_03.txt
niklasmohrin: These are the associated files for the `GradeDocument` instances from `test_data.json`

Kakadus (approved):

evap/evaluation/tests/test_misc.py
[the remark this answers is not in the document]
Kakadus: :shrug:

.gitignore
```diff
@@ -26,10 +26,8 @@ evap/localsettings.py
 evap/static/css/evap.css
 evap/static/css/evap.css.map
 evap/static/ts/rendered
-evap/static_collected
 evap/media
 evap/database.sqlite3
-evap/upload
```
richardebeling, line 32: To me, `evap/media` also looks like something that should have its root in `DATADIR` (at least if it is about uploading e.g. images, which is what I'd currently expect)?

evap/development/management/commands/reload_testdata.py
```diff
@@ -27,4 +32,9 @@ def handle(self, *args, **options):
 
         logged_call_command(self.stdout, "refresh_results_cache")
 
+        upload_dir = settings.DATADIR / "upload"
+        if upload_dir.exists():
+            shutil.rmtree(upload_dir)
```
richardebeling, line 37: Maybe add a hint that this will happen to `help` and the warning print text above (so that noone can claim we didn't tell them)

richardebeling (approved):
Friendship ended with `os.path.join`. This is so nice, we should have done this years ago.

.gitignore
niklasmohrin: I don't think it's used at all, is it?

evap/development/management/commands/reload_testdata.py
niklasmohrin: What do you think about the wording I pushed?

richardebeling: sounds good, thanks

.gitignore
richardebeling: quite possible :D

richardebeling (approved):

PR #2377, opened 2025-01-27
Refactor Sisyphus to ts and drop unnecessary functionality

### LINKED ISSUE #2371: Fork sisyphus and maintain
While implementing #1575, I discovered that sisyphus is broken and isn't maintained since seven years. As the source code is only 550 lines long and we could save some more lines by dropping unnecessary stuff, I would suggest to take the source code, drop the jQuery dependency and maintain it as .ts file in our source.

hansegucker:
Close #2371


evap/static/ts/src/sisyphus.ts
```diff
@@ -0,0 +1,309 @@
...
+
+                if (field instanceof HTMLInputElement && fieldType === "checkbox") {
+                    if (field.name.indexOf("[") !== -1) {
+                        if (multiCheckboxCache[field.name]) {
+                            return;
+                        }
+                        let tempValue: string[] = [];
+                        $("[name='" + field.name + "']:checked").each(function () {
+                            tempValue.push(field.value);
+                        });
```
richardebeling, line 148: Re https://github.com/e-valuation/EvaP/pull/2377#pullrequestreview-2590630107: This is the only jquery left in here, right? Should be translatable to `querySelectorAll` -- would be really nice to have that

hansegucker:
> In addition to the ESLint errors, there is also some jQuery left, is there anything holding us back from porting it to DOM apis?

Fixed.

evap/static/ts/src/sisyphus.ts
```diff
@@ -0,0 +1,312 @@
...
+    /**
+     * Bind release form fields data from local storage on submit/reset form
+     */
+    bindReleaseData() {
+        for (const target of this.targets) {
+            const releaseHandler = () => {
+                this.releaseData(target);
+            };
+            target.addEventListener("submit", releaseHandler);
+            target.addEventListener("reset", releaseHandler);
```
niklasmohrin, line 281: ... but even without `autoRelease`, we might want to have this event handler. We don't have any `type="reset"` inputs in EvaP, so it would be fine to ignore it. What do you think?

evap/static/ts/src/sisyphus.ts
```diff
@@ -0,0 +1,312 @@
...
+    locationBased: boolean;
+    timeout: number;
+    autoRelease: boolean;
+    onSave: (arg0: Sisyphus) => void;
+    onBeforeRestore: (arg0: Sisyphus) => boolean | undefined;
+    onRestore: (arg0: Sisyphus) => void;
+    onRelease: (arg0: Sisyphus) => void;
+}
+
+export class Sisyphus {
```
niklasmohrin, line 22: I think we could rename this to something like `FormPersister`, and then we are also at a point where I think it's fine to just credit the original work in the top comment of the file, but not include so much licensing information.

evap/static/ts/src/sisyphus.ts
```diff
@@ -0,0 +1,312 @@
+/**
+ * Plugin developed to save html forms data to LocalStorage to restore them after browser crashes, tabs closings
+ * and other disasters.
+ *
+ * @author Alexander Kaupanin <kaupanin@gmail.com>
+ * @author Jonathan Weth <dev@jonathanweth.de>
+ */
+
+interface SisyphusOptions {
```
niklasmohrin, line 9: A lot of the options are not needed, like `timeout` or `onBeforeRestore`. I think we can remove them

evap/static/ts/src/sisyphus.ts
```diff
@@ -0,0 +1,312 @@
...
+     */
+    restoreFieldsData(field: Element, resque: string) {
+        if (field.getAttribute("name") === undefined) {
+            return false;
+        }
+        if (
+            field instanceof HTMLInputElement &&
+            field.type === "checkbox" &&
+            resque !== "false" &&
+            !field.name.includes("[")
```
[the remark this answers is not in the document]
hansegucker: That's for some special kind of checkbox.

evap/static/ts/src/sisyphus.ts
```diff
@@ -0,0 +1,312 @@
...
+            field instanceof HTMLInputElement &&
+            field.type === "checkbox" &&
+            resque === "false" &&
+            !field.name.includes("[")
+        ) {
+            field.checked = false;
+        } else if (field instanceof HTMLInputElement && field.type === "radio") {
+            if (field.value === resque) {
+                field.checked = true;
+            }
```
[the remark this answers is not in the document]
hansegucker: ?

evap/static/ts/src/sisyphus.ts
```diff
@@ -0,0 +1,312 @@
...
+        } else if (field instanceof HTMLInputElement && !field.name.includes("[")) {
+            field.value = resque;
+        } else {
+            // @ts-expect-error Definitely an input field with a value, but not known by type
```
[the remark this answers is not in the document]
hansegucker: But there are some more possible input fields like selects and I don't want to add exceptions for all of them.

evap/static/ts/src/sisyphus.ts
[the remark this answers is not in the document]
niklasmohrin: But what kind?

evap/static/ts/src/sisyphus.ts
richardebeling: While I agree on an ethical level, I don't think this can be taken as granted from a legal standpoint. Copyright interpretation in various countries across the globe (including Germany) is very strict in what already constitutes a "substantial portion" in the sense of the MIT license. I can see judges in Germany deciding that our .ts file here is still derivative work.

I'd rather not have any legal battle about this and would keep the MIT license note. It doesn't hurt us.

@janno42 thoughts? (For comparison, original JS file is [here](https://github.com/simsalabim/sisyphus/blob/master/sisyphus.js))

niklasmohrin: What about something like
```typescript
/**
 * `FormPersister` class to save user inputs to local storage and restore them.
 * Based on https://github.com/simsalabim/sisyphus (MIT license).
 * A copy of the MIT license can be found in the repository root.
 *
 * Copyright (c) 2011-2018 Alexander Kaupanin https://github.com/simsalabim
 * Copyright (c) 2025 EvaP Developers https://github.com/e-valuation
 */
```
as a file header?

richardebeling: I'd err on the side of caution. We allowed ourselves to be a bit sneaky with the current license linking in our LICENSE.md as we only use release-bundles of all these dependencies _as they are published_, we assume that a published release bundle of some software is a form that is intended for (re)distribution.

Here we have derivative work where this argument doesn't hold. I'd vote to be cautious and just adhere to the requirements of the original MIT license strictly, which means that we quote [lines 1-12 of the original license](https://github.com/simsalabim/sisyphus/blob/master/MIT-LICENSE#L1-L12) verbatim. We can put additional stuff before or after that, for example
> Derived from https://github.com/simsalabim/sisyphus, originally distributed under the following license:
> ---------------------------------------------- begin original license
> [original license text]
> ---------------------------------------------- end original license
> New portions and modifications are licensed under the conditions in LICENSE.md

If we want to be shorter, I wouldn't reference our LICENSE.md as it doesn't just contain the clean MIT license text, but additional stuff at the top, and the Copyright line is different, which could be considered misleading for this file here.

Maybe it would be fine to say

> Derived from https://github.com/simsalabim/sisyphus with original copyright
> Copyright (c) 2011-2013 Alexander Kaupanin https://github.com/simsalabim
> licensed under MIT license at https://github.com/simsalabim/sisyphus/blob/1878a0f/MIT-LICENSE

In general, I'd like to prevent adding any new "Copyright" lines with years/names/whatever

evap/static/ts/src/auto-form-saver.ts
```diff
@@ -0,0 +1,198 @@
...
+NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+---------------------------------------------- end original license
+
+New portions and modifications are licensed under the conditions in LICENSE.md
+ */
+
+interface AutoFormSaverOptions {
```
niklasmohrin, line 30: Out of curiosity: Did we consider using `type` instead of `interface`, and did we have any reason for one or the other? I would think that with `Partial` in the constructor, `type` would be the correct construct, but not sure

evap/static/ts/src/auto-form-saver.ts
```diff
@@ -0,0 +1,198 @@
...
+            this.options.onRestore(this);
+        }
+    }
+
+    restoreFieldsData(field: Element, storedValue: string) {
+        if (field.getAttribute("name") === undefined) {
+            return false;
+        }
+        if (field instanceof HTMLInputElement && field.type === "checkbox") {
+            field.checked = storedValue !== "false";
```
niklasmohrin, line 157: Now that I see it, should we go for `storedValue === "true"` instead? What do we save to localStorage if we have `input type=checkbox checked=1`?

evap/static/ts/src/auto-form-saver.ts
```diff
@@ -0,0 +1,198 @@
...
+            }
+        }
+
+        if (restored) {
+            this.options.onRestore(this);
+        }
+    }
+
+    restoreFieldsData(field: Element, storedValue: string) {
+        if (field.getAttribute("name") === undefined) {
```
niklasmohrin, line 153: From https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute#return_value I would think that this can never happen. Could it be that `hasAttribute` is what we want here?

evap/static/ts/src/auto-form-saver.ts
```diff
@@ -0,0 +1,198 @@
+/**
+Derived from https://github.com/simsalabim/sisyphus, originally distributed under the following license:
```
niklasmohrin, line 2: We still mention sisyphus in our `LICENSE.md` file, I suppose we should remove it then? @richardebeling

niklasmohrin (approved):
Very cool, thanks!

richardebeling (approved):

janno42 (approved):
Bye Sisyphos 👋

PR #2382, opened 2025-02-03
Add step to rewards.rewardpointredemptionevent in testdata

jooooosef:
in #2264 the `step` field was added to `rewards.rewardpointredemptionevent` but not reflected in the `test_data.json`.

It is now added :)

### DIFF
--- a/evap/development/fixtures/test_data.json
+++ b/evap/development/fixtures/test_data.json
@@ -153476,7 +153476,8 @@
   "fields": {
     "name": "Big party",
     "date": "2099-12-31",
-    "redeem_end_date": "2099-12-30"
+    "redeem_end_date": "2099-12-30",
+    "step": 1
   }
 },
 {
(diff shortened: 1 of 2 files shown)


evap/development/fixtures/test_data.json
```diff
@@ -153476,7 +153476,8 @@
   "fields": {
     "name": "Big party",
     "date": "2099-12-31",
-    "redeem_end_date": "2099-12-30"
+    "redeem_end_date": "2099-12-30",
+    "step": 1
```
janno42, line 153480: This should be something other than 1 to have more useful testdata

richardebeling (approved):

janno42 (approved):

evap/development/fixtures/test_data.json
niklasmohrin: Let's add a comment to `RewardPointRedemptionEvent.step` saying something like

> Note that we allow this value to change throughout the lifetime of the event.

PR #2384, opened 2025-02-03
Remove `magic-nix-cache`

niklasmohrin:
See https://determinate.systems/posts/magic-nix-cache-free-tier-eol/

### DIFF
--- a/.github/setup_evap/action.yml
+++ b/.github/setup_evap/action.yml
@@ -19,7 +19,6 @@ runs:
   using: "composite"
   steps:
     - uses: DeterminateSystems/nix-installer-action@main
-    - uses: DeterminateSystems/magic-nix-cache-action@main
     - uses: nicknovitski/nix-develop@v1
       with:
         arguments: "${{ inputs.shell }}"


.github/setup_evap/action.yml
```diff
@@ -19,7 +19,6 @@ runs:
   using: "composite"
   steps:
     - uses: DeterminateSystems/nix-installer-action@main
-    - uses: DeterminateSystems/magic-nix-cache-action@main
```
richardebeling, line 22: That was the thing that moves build times from "10 minutes" to "a few seconds", right?

What do we replace it with? Do we want to ask FlakeHub for a sponsored open source account? GitHub's caching should also work, but I guess the tldr of the blogpost is that there's no infrastructure to use it for nix?

niklasmohrin: In theory, yes. However, we didn't really notice a speed up today and this PR also finished in 2, 3 minutes. If we need it, I would first try to contact them for an open source account

richardebeling: 🤷 alright

richardebeling (approved):

Kakadus (approved):

PR #2385, opened 2025-02-03
Test mismatching testdata

jooooosef:
in #2382 we added a field to some data in test_data.json after the migration that added this field was added because it was forgetten in the original pr.
To prevent that from happening in the future I added this test.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,6 +39,23 @@ jobs:
       - name: Run tests
         run: python manage.py test --shuffle
 
+  test_testdata:
+    name: Test if testdata is up to date
+
+    runs-on: ubuntu-22.04
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/setup_evap
+        with:
+          start-db: true
+      - name: Create new test_data.json
+        run: |
+          python manage.py reload_testdata --noinput
+          python manage.py dump_testdata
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -39,6 +39,21 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/setup_evap
+        with:
+          start-db: true
+
+      - run: python manage.py reload_testdata --noinput
+      - run: python manage.py dump_testdata
+      - run: git diff --exit-code
```
Kakadus, line 55: Please group the steps together into one (or two), and name them appropriately :)

evap/development/management/commands/dump_testdata.py
Kakadus, line 1: what was the reason for this again? (looks good - it's just that this is not used currently)

Kakadus (commented):
Nice!

evap/development/management/commands/dump_testdata.py
jooooosef: no I thought I might need that for the test but I didnt. But its more functionality so I think its fine.

niklasmohrin (approved):

Kakadus (approved):

Kakadus:
@jooooosef could you rebase this?

niklasmohrin (approved):

PR #2387, opened 2025-02-10
Move `manage.py` content to `evap.__main__`

niklasmohrin:
This is in preparation for #2328, where we want to include the `manage.py` code, but cannot include the `manage.py` file because it is outside the `evap` module. Using `__main__` allows running management commands with `python -m evap` if evap is installed into the environment. Running `manage.py` still works, as it uses the builtin `runpy` module to simulate running `python -m evap`.

### DIFF
--- a/evap/__main__.py
+++ b/evap/__main__.py
@@ -8,6 +8,7 @@ from django.core.management import execute_from_command_line
 
 
 def main():
+    assert not settings.configured
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "evap.settings")
     settings.DATADIR.mkdir(exist_ok=True)
     execute_from_command_line(sys.argv)
--- a/evap/__main__.py
+++ b/evap/__main__.py
@@ -1,0 +1,17 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+
+from django.conf import settings
+from django.core.management import execute_from_command_line
+
+
+def main():
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "evap.settings")
+    settings.DATADIR.mkdir(exist_ok=True)
+    execute_from_command_line(sys.argv)
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/manage.py
+++ b/manage.py
@@ -1,13 +1,5 @@
 #!/usr/bin/env python3
 
-import os
-import sys
+from evap.__main__ import main
 
-if __name__ == "__main__":
-    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "evap.settings")
-
-    from django.conf import settings
-    from django.core.management import execute_from_command_line
-
-    settings.DATADIR.mkdir(exist_ok=True)
-    execute_from_command_line(sys.argv)
+main()
--- a/manage.py
+++ b/manage.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
 
-from evap.__main__ import main
+import runpy
 
-main()
+runpy.run_module("evap", run_name="__main__")


evap/__main__.py
```diff
@@ -0,0 +1,17 @@
...
+
+import os
+import sys
+
+from django.conf import settings
+from django.core.management import execute_from_command_line
+
+
+def main():
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "evap.settings")
```
richardebeling, line 11: I think the environment variable must be set before we import `settings`.

manage.py
```diff
@@ -1,13 +1,5 @@
...
 
-if __name__ == "__main__":
-    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "evap.settings")
-
-    from django.conf import settings
-    from django.core.management import execute_from_command_line
-
-    settings.DATADIR.mkdir(exist_ok=True)
-    execute_from_command_line(sys.argv)
+main()
```
richardebeling, line 5: Not that anyone would ever import this file, but is there a specific reason to remove the `if __name__ == "__main__"` guard?

niklasmohrin: No reason except brevity

niklasmohrin: I can add it back if you want

evap/__main__.py
niklasmohrin: I believe that it's configured on first access to an attribute:
```python
def main():
    assert not settings.configured
    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "evap.settings")
    settings.DATADIR.mkdir(exist_ok=True)
    assert not settings.configured
    execute_from_command_line(sys.argv)
    assert not settings.configured
```
with `./manage.py help` fails on the second assertion. We could add the first assertion, or we could also move the import back in here. I think that it's fine at the top, since we can't guarantee that the settings haven't been configured from just this file anyways

richardebeling (approved):

manage.py
niklasmohrin: `runpy` sure is ... something :D It fixes the missing coverage, but it adds some complexity that we might not want. @richardebeling ?

Kakadus (approved):
I think both runpy or not is fine

manage.py
richardebeling: Looks like the right thing to do to me. I'm not sure I get your points about missing coverage (you mean test coverage? Would that have been a problem?) and complexity (import runpy and runpy.run_module look canonical to me, I wouldn't care what they do under the hood -- do I need to care?), unless I'm missing something here I'm very fine with the current code

niklasmohrin: I was wondering whether `runpy` magic could break something like `./manage.py test --pdb`, but it seems to work just fine. My points are
- without runpy, the line in `__main__.py` that calls `main` is not covered, but with runpy, it is (although the branch where `__name__` is _not_ `__main__` is uncovered now :D)
- I also don't care what runpy does and don't want to care, I was just afraid that we would have to care some day. It also makes the `manage.py` file less approachable to first semester students, but I guess they have to accept our arcane magic anyways once they hit `execute_from_command_line`

I also think runpy is fun, so we can keep it if the above don't concern you :)

niklasmohrin:
I noticed that the usage text is sometimes messed up, for example `python -m evap test --help` prints `usage: __main__.py test [snip]`. Normally, [Django wants to cover this up](https://github.com/django/django/blob/4a3ad9eebbc16ce80b348644b557c84ecc741be7/django/core/management/__init__.py#L205-L206), but it [doesn't seem to work with subcommands](https://code.djangoproject.com/ticket/32177#comment:8). Either way, the "fixed" variant would be `usage: python -m django test [snip]`, which would also be wrong. If we were to really care about this, we would have to reimplement `execute_from_command_line` ourselves. I, personally, don't care

manage.py
richardebeling: Alright, makes sense. I'm fine with both approaches.

Issue #2333, opened 2024-11-18
Collapse participant list in staff evaluation form


janno42:
The staff evaluation form lists, among other things, all participants in that evaluation. This list can get quite long, and if users want to edit other parts of the form, this results in a lot of unnecessary scrolling.

Therefore, the participants field should be collapsed as soon as it contains at least one participant, and the number of participants should be displayed in the heading of this collapse element, so that the number is visible even in the collapsed state. The number should update on field changes.

PR #2388, opened 2025-02-10
Collapsable box for participants in evaluation form

ybrnr:
implements visual changes suggested by @janno42 in #2333 
for non staff users the list is always shown by default. For staff users the list is collapsed by default except when there are no participants


janno42:
From the issue: "The number should update on field changes."
Can you please add some JS for that? :)

Have a look at https://tom-select.js.org/examples/events/ for the events of the component.

evap/evaluation/templates/evaluation_form_general_questionnaires.html
```diff
@@ -10,6 +10,25 @@
                     {% include 'questionnaires_widget.html' with field=field questionnaires_with_answers=questionnaires_with_answers_per_contributor|get:None%}
                 </div>
             </div>
+        {% elif field == evaluation_form.participants %}
+        <div class="mb-3 d-md-flex">
+            {% include 'bootstrap_form_field_label.html' with field=field class='col-md-3 pe-4' %}
+            <div class="col-md-7 card collapsible overflow-visible">
+                <div class="card-header d-flex">
+                    <a class="collapse-toggle {% if collapse_participants and evaluation.participants.count > 0 %}collapsed{% endif %}"
```
janno42, line 18:
```suggestion
                    <a class="collapse-toggle{% if collapse_participants and evaluation.participants.count > 0 %} collapsed{% endif %}"
```

evap/evaluation/templates/evaluation_form_general_questionnaires.html
```diff
@@ -10,6 +10,25 @@
                     {% include 'questionnaires_widget.html' with field=field questionnaires_with_answers=questionnaires_with_answers_per_contributor|get:None%}
                 </div>
             </div>
+        {% elif field == evaluation_form.participants %}
+        <div class="mb-3 d-md-flex">
```
janno42, line 14: please indent

evap/evaluation/templates/evaluation_form_general_questionnaires.html
```diff
@@ -10,6 +10,25 @@
...
+            {% include 'bootstrap_form_field_label.html' with field=field class='col-md-3 pe-4' %}
+            <div class="col-md-7 card collapsible overflow-visible">
+                <div class="card-header d-flex">
+                    <a class="collapse-toggle {% if collapse_participants and evaluation.participants.count > 0 %}collapsed{% endif %}"
+                        data-bs-toggle="collapse" href="#contributor-{{ contribution_result.contributor.id }}"
+                        aria-controls="contributor-{{ contribution_result.contributor.id }}">
+                    </a>
+                    <span class="badge rounded-pill badge-btn ms-auto" id="participant_counter"> {{ evaluation.participants.count }} </span>
```
janno42, line 22:
```suggestion
                    <span class="badge rounded-pill badge-btn ms-auto" id="participant_counter">{{ evaluation.participants.count }}</span>
```

evap/evaluation/templates/evaluation_form_general_questionnaires.html
```diff
@@ -10,6 +10,25 @@
...
+                </div>
+                <div class="collapse {% if not collapse_participants or evaluation.participants.count == 0 %}show{% endif %}" id="contributor-{{ contribution_result.contributor.id }}">
```
janno42, line 24:
```suggestion
                <div class="collapse{% if not collapse_participants or evaluation.participants.count == 0 %} show{% endif %}" id="contributor-{{ contribution_result.contributor.id }}">
```

evap/evaluation/templates/evap_evaluation_edit_js.html
```diff
@@ -71,6 +71,20 @@
             });
         }
     }
+
```
janno42, line 74:
```suggestion

    // Update participant count displayed in form
```

evap/evaluation/templates/evap_evaluation_edit_js.html
```diff
@@ -71,6 +71,20 @@
...
+    });
+    participant_list.on("item_remove", function(){
+        count_participants();
+    });
+
+    function count_participants(){
+        participant_counter = document.getElementById("participant_counter");
+        participant_counter.innerHTML = participant_list.items.length;
+    }
+
```
janno42, line 87:
```suggestion
```

evap/evaluation/templates/evap_evaluation_edit_js.html
```diff
@@ -71,6 +71,20 @@
             });
         }
     }
+
+    let participant_list = document.getElementById("id_participants").tomselect;
+    participant_list.on("item_add", function() {
+        count_participants();
+    });
+    participant_list.on("item_remove", function(){
```
janno42, line 79:
```suggestion
    participant_list.on("item_remove", function() {
```

janno42 (changes requested):

evap/evaluation/templates/evap_evaluation_edit_js.html
```diff
@@ -71,6 +71,20 @@
...
+    participant_list.on("item_add", function() {
+        count_participants();
+    });
+    participant_list.on("item_remove", function() {
+        count_participants();
+    });
+
+    function count_participants(){
+        participant_counter = document.getElementById("participant_counter");
+        participant_counter.innerHTML = participant_list.items.length;
```
niklasmohrin, line 86: I think we can remove some specifics from this: Let's make it so that the participant_counter does not need to have a `id` anymore and instead it has, say, `data-track-tomselect-count="#id_participants"`.

Additionally, we should use `innerText` instead of `innerHTML`

evap/evaluation/templates/evaluation_form_general_questionnaires.html
```diff
@@ -10,6 +10,24 @@
...
+                    <div class="card-header d-flex">
+                        <a class="collapse-toggle{% if collapse_participants and evaluation.participants.count > 0 %} collapsed{% endif %}"
+                            data-bs-toggle="collapse" href="#contributor-{{ contribution_result.contributor.id }}"
+                            aria-controls="contributor-{{ contribution_result.contributor.id }}">
+                        </a>
+                        <span class="badge rounded-pill badge-btn ms-auto" id="participant_counter">{{ evaluation.participants.count }}</span>
+                    </div>
+                    <div class="collapse{% if not collapse_participants or evaluation.participants.count == 0 %} show{% endif %}" id="contributor-{{ contribution_result.contributor.id }}">
+                        <div class="card-body">
+                            {% include 'bootstrap_form_field_widget.html' with field=field %}
```
niklasmohrin, line 26: What is this about contributors here?

niklasmohrin (changes requested):
Can we add a test?

evap/evaluation/templates/base.html
```diff
@@ -213,6 +213,18 @@
...
+                trackedElement.on("item_add", function() {
+                    update_count(trackerElement, trackedElement);
+                });
+                trackedElement.on("item_remove", function() {
+                    update_count(trackerElement, trackedElement);
+                });
+                function update_count(trackerElement, trackedElement) {
+                    trackerElement.innerText = trackedElement.items.length;
+                }
+            });
```
Kakadus, line 227: I think we like it more if this is written in a typescript function and executed like `setupForms`

janno42:
it would be nice if the whole card header would toggle the collapsible

janno42 (approved):
🔢 
Great!

evap/evaluation/templates/evaluation_form_fields.html
```diff
@@ -12,6 +12,20 @@
                     </div>
                 </div>
             {% endif %}
+        {% elif field == evaluation_form.participants %}
+        <div class="mb-3 d-md-flex">
```
niklasmohrin: Please indent :)

evap/evaluation/templates/evaluation_form_fields.html
```diff
@@ -12,6 +12,20 @@
...
+            {% include 'bootstrap_form_field_label.html' with field=field class='col-md-3 pe-4' %}
+            <div class="col-md-7 card collapsible overflow-visible">
+                <a class="card-header d-flex collapse-toggle{% if collapse_participants and evaluation.participants.count > 0 %} collapsed{% endif %}" data-bs-toggle="collapse" href="#participant-box">
+                    <span class="badge rounded-pill badge-btn ms-auto" data-track-tomselect-count="id_participants">{{ evaluation.participants.count }}</span>
```
niklasmohrin: I would remove the initial value here and instead populate it by calling `update_count` once when setting up the tracking. If for some reason javascript doesn't work, then we are not displaying the outdated value here. Additionally, the count inserted here could be different to what is selected, see https://github.com/e-valuation/EvaP/pull/2388#discussion_r2249982175

evap/static/ts/src/tom-select-count.ts
```diff
@@ -0,0 +1,15 @@
...
+            update_count(trackerElement, trackedElement);
+        });
+        trackedElement.on("item_remove", function () {
+            update_count(trackerElement, trackedElement);
+        });
+        function update_count(trackerElement: any, trackedElement: any) {
+            trackerElement.innerText = trackedElement.items.length;
+        }
+    });
+};
```
niklasmohrin, line 14: Multiple things: We can get of rid of some `any` annotations by using our `assert` helper. For tomselect itself, we are out of luck, because there is no just-types package available - I don't know if we could use the npm package for types but keep our vendored file. We could also consider just always pulling tomselect from npm. However, this is not relevant for this PR, I think using any for this is fine as it will throw an error just like assertions would, and the live test checks for that.

I would thus suggest the following:
```suggestion
import { assert, selectOrError } from "./utils.js";

export const setupTomSelectCount = () => {
    document.querySelectorAll("[data-track-tomselect-count]").forEach(trackerElement => {
        assert(trackerElement instanceof HTMLElement);
        const trackedTomselect = (selectOrError(trackerElement.dataset.trackTomselectCount!) as any).tomselect;
        const update_count = () => trackerElement.innerText = trackedTomselect.items.length;

        update_count()
        trackedTomselect.on("item_add", update_count);
        trackedTomselect.on("item_remove", update_count);
    });
};
```

Note that by using selectOrError, we should use a CSS selector like with querySelector instead of an id, so in the template we need to use `#id_participants`

evap/staff/tests/test_live.py
```diff
@@ -71,3 +71,50 @@ def test_submit_changes_form_data(self):
...
+            course=baker.make(Course, programs=[baker.make(Program)], responsibles=[responsible]),
+            participants=participants,
+            vote_start_datetime=datetime(2099, 1, 1, 0, 0),
+            vote_end_date=date(2099, 12, 31),
+        )
+
+        with self.enter_staff_mode():
+            self.selenium.get(self.live_server_url + reverse("staff:evaluation_edit", args=[evaluation.id]))
+
+        card_header = self.selenium.find_element(By.CSS_SELECTOR, ".card .card-header")
```
niklasmohrin: To ensure that this doesn't fail if we add more cards:
```suggestion
        card_header = self.selenium.find_element(By.CSS_SELECTOR, ".card:has(#id_participants) .card-header")
```

evap/staff/tests/test_live.py
```diff
@@ -71,3 +71,50 @@ def test_submit_changes_form_data(self):
...
+        assert "collapsed" in card_header.get_attribute("class")  # type: ignore
+        card_header.click()
+        assert "collapsed" not in card_header.get_attribute("class")  # type: ignore
+
+        counter = card_header.find_element(By.CSS_SELECTOR, ".rounded-pill")
+        assert counter.get_attribute("innerText") == "20"
```
niklasmohrin: For this one they have a helper:
```suggestion
        assert counter.text == "20"
```

evap/staff/tests/test_live.py
```diff
@@ -71,3 +71,50 @@ def test_submit_changes_form_data(self):
         self.assertEqual(contribution1.order, 0)
         self.assertEqual(contribution1.role, Contribution.Role.EDITOR)
         self.assertEqual(contribution1.textanswer_visibility, Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS)
+
+
+# ruff: noqa: PGH003
```
```suggestion
```

Instead, annotate the type ignores with `ignore[operator]`

evap/staff/tests/test_live.py
niklasmohrin, line 1: Please use `self.assertXXX` instead of plain assert

evap/staff/tests/test_live.py
```diff
@@ -71,3 +71,50 @@ def test_submit_changes_form_data(self):
...
+class ParticipantCollapseTests(LiveServerTest):
```
niklasmohrin, line 117: Can you enhance the test to add and remove participants and check that the count is updated accordingly?

evap/evaluation/templates/evaluation_form_fields.html
niklasmohrin, line 1: @janno42 One thing that happens (also on main) in Firefox for me is that the list of participants is somehow cached: Removing a participant and reloading the page still has the participant missing, although this is not true in the database. Curiously, this only happens when reloading; when navigating elsewhere and back to the evaluation, the correct list is shown. Force-reloading also shows the correct list.

Do we care?

richardebeling: Sounds like browser form value caching messing with us again (see https://github.com/e-valuation/EvaP/pull/1815/commits/ae3d44dfa72e8886728838b71e38d7ef3c348089#diff-8d3c2186fe706215b9c76b5141c5fd1b5b8c2790274dbc3f734381f399f6e39eR2-R6). I could see how adding `autocomplete=off` fixes this. Interestingly, I don't find any past discussion about the autocomplete issue on github

evap/staff/tests/test_live.py
ybrnr: i added one for adding participants. Should I also add one for removing a participant?

evap/staff/tests/test_live.py
```diff
@@ -71,3 +71,62 @@ def test_submit_changes_form_data(self):
...
+        with self.enter_staff_mode():
+            self.selenium.get(self.live_server_url + reverse("staff:evaluation_edit", args=[evaluation.id]))
+
+        card_header = self.selenium.find_element(By.CSS_SELECTOR, ".card:has(#id_participants) .card-header")
+        self.assertFalse("collapsed" in classes_of_element(card_header))
+        card_header.click()
+        self.assertTrue("collapsed" in classes_of_element(card_header))
+
+        counter = card_header.find_element(By.CSS_SELECTOR, ".rounded-pill")
+        self.assertTrue(counter.text == "0")
```
niklasmohrin: Sorry, what I meant was not to just use `self.assertTrue` and `self.assertFalse`, but to make use of the more specific `self.assertEqual`, `self.assertIn`, etc. methods. The reason for this is that _if_ these assertions fail the command line output will contain what the expected values were (i.e. what `counter.text` really was here). Right now, we would just get something like "expected value to be true, but was false" and we would have to re-run the test with a debugger to inspect the actual value

evap/static/ts/src/tom-select-count.ts
```diff
@@ -0,0 +1,20 @@
+import { assert, selectOrError } from "./utils.js";
+
+export const setupTomSelectCount = () => {
+    document.querySelectorAll("[data-track-tomselect-count]").forEach(trackerElement => {
+        assert(trackerElement instanceof HTMLElement);
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+        const trackedElement = (selectOrError(trackerElement.dataset.trackTomselectCount!) as any).tomselect;
```
niklasmohrin: We can just fix the lint instead of disabling it:
```suggestion
        const trackedElement = selectOrError<any>(trackerElement.dataset.trackTomselectCount!).tomselect;
```

evap/static/ts/src/tom-select-count.ts
```diff
@@ -0,0 +1,20 @@
...
+
+        update_count(trackerElement, trackedElement);
+        trackedElement.on("item_add", function () {
+            update_count(trackerElement, trackedElement);
+        });
+        trackedElement.on("item_remove", function () {
+            update_count(trackerElement, trackedElement);
+        });
+        function update_count(trackerElement: any, trackedElement: any) {
```
niklasmohrin: Please get rid of these `any` annotations by just capturing the variables in a function instead of passing them here; see my previous comment for a possible solution

evap/staff/tests/test_live.py
```diff
@@ -71,3 +72,66 @@ def test_submit_changes_form_data(self):
         self.assertEqual(contribution1.order, 0)
         self.assertEqual(contribution1.role, Contribution.Role.EDITOR)
         self.assertEqual(contribution1.textanswer_visibility, Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS)
+
+
+class ParticipantCollapseTests(LiveServerTest):
```
niklasmohrin, line 117: Very nice!

evap/evaluation/templates/evaluation_form_fields.html
```diff
@@ -12,6 +12,20 @@
                     </div>
                 </div>
             {% endif %}
+        {% elif field == evaluation_form.participants %}
+            <div class="mb-3 d-md-flex">
+                {% include 'bootstrap_form_field_label.html' with field=field class='col-md-3 pe-4' %}
+                <div class="col-md-7 card collapsible overflow-visible">
+                    <a class="card-header d-flex collapse-toggle{% if state >= evaluation.State.EDITOR_APPROVED and evaluation.participants.count > 0 %} collapsed{% endif %}" data-bs-toggle="collapse" href="#participant-box">
+                        <span class="badge rounded-pill badge-btn ms-auto" data-track-tomselect-count="#id_participants"></span>
```
janno42, line 20: Please make this not being underlined when hovered

janno42 (changes requested):

evap/evaluation/templates/evaluation_form_fields.html
ybrnr: this is fixed now

ybrnr:
> oh and did you fix [#2388 (comment)](https://github.com/e-valuation/EvaP/pull/2388#discussion_r2249982175)? I couldn't find any changes related to that

yes :)

evap/staff/forms.py
```diff
@@ -424,6 +424,7 @@ def __init__(self, *args, requires_decided_main_language=False, **kwargs):
         if self.instance.pk is not None:
             queryset = (queryset | self.instance.participants.all()).distinct()
         self.fields["participants"].queryset = queryset
+        self.fields["participants"].widget.attrs["autocomplete"] = "off"
```
niklasmohrin, line 428: nit: if you want, it would still be nice to have a test that fails if this line is missing - maybe we can add a couple more lines to one of the tests?

niklasmohrin (approved):
Very nice!

janno42 (approved):
🔽🔼  Thanks!

janno42:
Bonus points if you want: You could remove the padding and border from the user select field so that it fills the entire card body. If you don't want to do that, I'll add it to my list for a larger redesign of the form.

evap/staff/forms.py
ybrnr: i added some lines. Is this what you imagined or should I do something different?

janno42 (approved):
🎉

PR #2397, opened 2025-02-24
Remove compiled files in clean-setup and don't collect static files in init

hansegucker:
We don't want to call `collectstatic` in startup as it's only necessary for production and not for development. Additionally, this MR also deletes generated JS and CSS files on clean-setup. clean-setup now also runs as `nix run .#clean-setup`.


niklasmohrin (approved):
Nice :+1:

richardebeling (approved):

PR #2399, opened 2025-02-24
Semesters not ordered correctly in evaluation selection on person management page 

### LINKED ISSUE #2361: Semesters not ordered correctly in evaluation selection on person management page
The semesters in the evaluation selection dropdown on the evaluation person management page (when copying participants or contributors) are not sorted correctly. We sort them in our code, but TomSelect seems to apply its own alphabetical order.
Instead, the form element should keep the original semester order.

jooooosef:
fix #2361.
This could also apply to any other tomselects where we sort the results into semester groups and sort these semester groups.
This can (now) be done adding the HTML `tomselect-no-sort` attribute to the `select` element.

### DIFF
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -164,7 +164,7 @@
                         plugins: {},
                         hideSelected: false,
                     };
-                    if (element.hasAttribute("tomselect-no-sort") && element.getAttribute("tomselect-no-sort") == 'true') {
+                    if (element.hasAttribute("tomselect-no-sort")) {
                         baseOptions.sortField = [{field:'$order'},{field:'$score'}];
                     }
                     if(element.hasAttribute("data-tomselect-fullwidth")) {
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -164,6 +164,9 @@
                         plugins: {},
                         hideSelected: false,
                     };
+                    if (element.hasAttribute("tomselect-no-sort") && element.getAttribute("tomselect-no-sort") == 'true') {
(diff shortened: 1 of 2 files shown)


evap/evaluation/templates/base.html
```diff
@@ -164,6 +164,9 @@
                         plugins: {},
                         hideSelected: false,
                     };
+                    if (element.hasAttribute("tomselect-no-sort") && element.getAttribute("tomselect-no-sort") == 'true') {
+                        baseOptions.sortField = [{field:'$order'},{field:'$score'}];
+                    }
```
richardebeling, line 169: I'd say, for developer sanity, we should follow HTML's [conventions of boolean attributes here](https://html.spec.whatwg.org/#boolean-attribute): If the attribute exists, it is treated as "true". If you want to not have it, you remove the attribute.

The standard explicitly says
> The values "true" and "false" are not allowed on boolean attributes. To represent a false value, the attribute has to be omitted altogether.

jooooosef: Alright, did that now. I need to explicity set the attribute to an empty string with `self.fields["evaluation"].widget.attrs["tomselect-no-sort"] = ""` or I would try to access the `tomselect-no-sort` which does not exist at this point.
Is this acceptabe?

richardebeling: Sounds good to me 👍

Kakadus (approved):

niklasmohrin:
Is there a sensible way for us to test this?

richardebeling (approved):

janno42 (approved):
🔢

PR #2404, opened 2025-03-03
Correct redeeming users count in RewardPointRedemptionEvents overview

### LINKED ISSUE #2386: RewardPointRedemptionEvents overview does not show number of redeeming users
d2fed72de8 removed `redemptions_by_user` which is still used in the template `rewards_reward_point_redemption_event_list.html`.
This causes all numbers of redeeming users per event to be displayed as 0.

Instead, the number of redeeming users (which is not necessarily the number of redemptions) should be shown.

Kakadus:
close #2386.

also adds a test that checks the corresponding number of redeeming users are present in the html.

### DIFF
--- a/evap/rewards/tests/test_views.py
+++ b/evap/rewards/tests/test_views.py
@@ -120,22 +120,25 @@ class TestEventsView(WebTestStaffModeWith200Check):
 
     @classmethod
     def setUpTestData(cls) -> None:
-        cls.test_users = [make_manager()]
+        cls.manager = make_manager()
+        cls.test_users = [cls.manager]
 
         cls.redemption_event1 = baker.make(RewardPointRedemptionEvent, redeem_end_date=date.today() + timedelta(days=1))
         baker.make(RewardPointRedemptionEvent, redeem_end_date=date.today() + timedelta(days=1))
 
     def test_redemption_event_list(self) -> None:
+        quantity = 2
         baker.make(
             RewardPointRedemption,
             event=self.redemption_event1,
             value=1,
-            _quantity=2,
+            _quantity=quantity,
         )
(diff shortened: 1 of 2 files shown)


richardebeling (approved):

evap/rewards/tests/test_views.py
```diff
@@ -119,11 +119,27 @@ class TestEventsView(WebTestStaffModeWith200Check):
     url = reverse("rewards:reward_point_redemption_events")
 
     @classmethod
-    def setUpTestData(cls):
+    def setUpTestData(cls) -> None:
         cls.test_users = [make_manager()]
```
niklasmohrin, line 123: Can we do this so that we can use `self.manager` in the test?
```suggestion
        cls.manager = make_manager()
        cls.test_users = [cls.manager]
```

evap/rewards/tests/test_views.py
```diff
@@ -119,11 +119,27 @@ class TestEventsView(WebTestStaffModeWith200Check):
...
+        baker.make(
+            RewardPointRedemption,
+            event=self.redemption_event1,
+            value=1,
+            _quantity=2,
+        )
+        response = self.app.get(self.url, user=self.test_users[0])
+
+        self.assertContains(response, self.redemption_event1.name)
+        self.assertInHTML('<td><span class="fas fa-user"></span>2</td>', response.text)
```
niklasmohrin, line 138: Not sure if it's worth it, but optimally we would make this check a bit stricter and ensure that the "2" appears in the row of the correct event. What do you think?

niklasmohrin (approved):
Nice :)

janno42 (approved):

niklasmohrin (approved):

PR #2405, opened 2025-03-10
Suppress python callstacks on lint/ts/scss command failure

richardebeling:
just forward the exit code of the executed program instead

### DIFF
--- a/evap/evaluation/management/commands/tools.py
+++ b/evap/evaluation/management/commands/tools.py
@@ -50,7 +50,10 @@ def logged_call_command(stdout, *args, **kwargs):
     call_command(*args, **kwargs)
 
 
-def subprocess_run_or_exit(command: list) -> None:
+def subprocess_run_or_exit(command: list, stdout=None) -> None:
+    if stdout is not None:
+        stdout.write(f"Executing {' '.join(str(el) for el in command)}")
+
     exit_code = subprocess.run(command, check=False).returncode
     if exit_code != 0:
         sys.exit(exit_code)
--- a/evap/evaluation/management/commands/tools.py
+++ b/evap/evaluation/management/commands/tools.py
@@ -1,4 +1,5 @@
 import logging
+import subprocess
 import sys
 
 from django.conf import settings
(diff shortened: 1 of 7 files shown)


evap/evaluation/management/commands/tools.py
```diff
@@ -47,3 +48,10 @@ def logged_call_command(stdout, *args, **kwargs):
     """Log execution of management command with all args."""
     stdout.write("Executing python manage.py " + " ".join(list(args) + [f"{a}={b}" for a, b in kwargs.items()]))
     call_command(*args, **kwargs)
+
+
+def subprocess_run_or_exit(command: list) -> None:
+    exit_code = subprocess.run(command, check=False).returncode
+    if exit_code != 0:
+        sys.exit(exit_code)
```
Kakadus, line 56: maybe a short 
```py
    def test_failing_command_exits(self) -> None:
        with self.assertRaises(SystemExit):
            subprocess_run_or_exit(["false"])
```
but may be too much

Kakadus (dismissed):

niklasmohrin (dismissed):
Nice! I wonder if we should (by default?) print a short "Running <the command>", otherwise it might be difficult to find where a printed error came from

evap/evaluation/management/commands/tools.py
```diff
@@ -47,3 +48,13 @@ def logged_call_command(stdout, *args, **kwargs):
     """Log execution of management command with all args."""
     stdout.write("Executing python manage.py " + " ".join(list(args) + [f"{a}={b}" for a, b in kwargs.items()]))
     call_command(*args, **kwargs)
+
+
+def subprocess_run_or_exit(command: list, stdout=None) -> None:
+    if stdout is not None:
+        stdout.write(f"Executing {' '.join(str(el) for el in command)}")
+
+    exit_code = subprocess.run(command, check=False).returncode
```
niklasmohrin, line 57: Should we
```suggestion
    exit_code = subprocess.run(command, stdout=stdout, check=False).returncode
```
or does that not make sense?

niklasmohrin (commented):
Nice

PR #2407, opened 2025-03-10
Add `evap-dev-with-lsp` shell

niklasmohrin:


Kakadus (approved):

richardebeling (approved):

richardebeling:
Marking as draft: intention was to provide access into python dependencies inside the nix shell (for vscode / vim / other lsp setups, e.g. to help with imports), but for our setups, "go to definition" jumping to a django definition doesn't work yet

niklasmohrin:
I played around with it a bit and I think the most robust version would be to get the Python LSP packages in the same way we get other development dependencies - through uv and PyPI.

For reference, the previous attempt used, roughly speaking, `nixpkgs.python312.pkgs.python-lsp-server`. This is set up to use `nixpkgs.python312` as the Python interpreter - one that (correctly so!) doesn't know about our virtualenv. I then tried to use the Python from the venv, but running pylsp with that also didn't automatically pick up our venv.

So (optionally) including the LSP in the venv seems to be the most intended way. If we really don't want to do this, we could also try to override the Python used by nixpkgs' pylsp, but I suspect that this will cause rebuilds of pylsp and it's dependencies (which compile some Rust crates).

I guess locking pylsp is fine, as we are already locking things like black. Let's hope that no weird dependency resolution things come up, and if they do, we can purge pylsp again.

richardebeling (approved):
lgtm, tested "go to definition" on django code, "organize imports" (ruff) and django import autocompletion, all work. Thanks! :)

PR #2408, opened 2025-03-10
Add preparation reminder text view

### LINKED ISSUE #2137: Preparation reminder text view
When preparing the evaluations for a semester, managers want to remind about evaluations that have not yet been confirmed by editors. EvaP currently has a "Preparation reminder" view. Managers can send individual emails from there, but sometimes all unconfirmed evaluations need to be sent as a complete list.

A new view mode should be added to the preparation reminder view, that lists all unconfirmed evaluations per responsible person. The style should be similar to the "Grade publish reminder".

The view mode should be toggled by a `btn-switch` in the top right corner, changing between "Interactive" and "Text":
image

Kakadus:
close #2137 

Example: 
image

### DIFF
--- a/evap/staff/templates/staff_semester_preparation_reminder.html
+++ b/evap/staff/templates/staff_semester_preparation_reminder.html
@@ -33,7 +33,7 @@
         <div class="btn-switch btn-switch-light ms-2 d-print-none h-">
             <div class="btn-switch-label">{% translate 'View' %}</div>
             <div class="btn-switch btn-group">
-                <a href="{% url 'staff:semester_preparation_reminder' semester_id=semester.id %}" role="button" class="btn btn-sm btn-light {% if interactive %} active{% endif %}">
+                <a href="{% url 'staff:semester_preparation_reminder' semester_id=semester.id %}" role="button" class="btn btn-sm btn-light{% if interactive %} active{% endif %}">
                     {% translate 'Interactive' %}
                 </a>
                 <a href="{% url 'staff:semester_preparation_reminder' semester_id=semester.id %}?mode=text" role="button" class="btn btn-sm btn-light{% if not interactive %} active{% endif %}">
--- a/evap/staff/templates/staff_semester_preparation_reminder.html
+++ b/evap/staff/templates/staff_semester_preparation_reminder.html
@@ -9,11 +9,12 @@
 
 {% block content %}
     {{ block.super }}
-
-    {% if responsible_list %}
-        <div class="d-flex mb-3">
-            <div class="me-auto">
-
+    <h3>
+        {% translate 'Preparation reminder' %} ({{ semester.name }})
+    </h3>
+    <div class="d-flex mb-3">
+        <div class="me-auto">
+            {% if responsible_list and interactive %}
                 <form reload-on-success method="POST" action="{% url 'staff:semester_preparation_reminder' semester_id=semester.id %}">
                     {% csrf_token %}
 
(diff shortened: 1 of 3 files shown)


richardebeling (approved):

evap/staff/templates/staff_semester_preparation_reminder.html
```diff
@@ -27,35 +28,47 @@
                         <button slot="show-button" type="button" id="remindAllButton" class="btn btn-sm btn-light">{% translate 'Remind all' %}</button>
                     </confirmation-modal>
                 </form>
+            {% endif %}
+        </div>
+        <div class="btn-switch btn-switch-light ms-2 d-print-none h-">
+            <div class="btn-switch-label">{% translate 'View' %}</div>
+            <div class="btn-switch btn-group">
+                <a href="{% url 'staff:semester_preparation_reminder' semester_id=semester.id %}" role="button" class="btn btn-sm btn-light {% if interactive %} active{% endif %}">
```
janno42, line 36:
```suggestion
                <a href="{% url 'staff:semester_preparation_reminder' semester_id=semester.id %}" role="button" class="btn btn-sm btn-light{% if interactive %} active{% endif %}">
```

janno42 (approved):

evap/staff/views.py
```diff
@@ -919,7 +919,11 @@ def semester_questionnaire_assign(request, semester_id):
 
 
 @manager_required
-def semester_preparation_reminder(request, semester_id):
+def semester_preparation_reminder(request: HttpRequest, semester_id: int) -> HttpResponse:
+    mode = request.GET.get("mode", "interactive")
+    if request.method == "POST" and mode == "text" or mode not in ["interactive", "text"]:
```
niklasmohrin, line 924: Aren't extraneous GET parameters ignored by default? I think it would be better if we would move the `move = ` line down just before `template_data =`, it's not really important for the rest of the view

evap/staff/tests/test_views.py
```diff
@@ -938,41 +938,36 @@ class TestSemesterPreparationReminderView(WebTestStaffModeWith200Check):
...
-    def test_remind_all(self, email_template_mock):
-        user = baker.make(UserProfile)
-        evaluation = baker.make(
-            Evaluation,
-            course=baker.make(Course, semester=self.semester, responsibles=[user]),
-            state=Evaluation.State.PREPARED,
-        )
+        response = self.app.get(self.url, params={"mode": "text"}, user=self.manager)
+        self.assertContains(response, self.user.full_name)
+        self.assertContains(response, self.evaluation.full_name)
```
niklasmohrin, line 967: If we change the default mode to "text", this test still passes - the assertions are the same for both views. Should we add something to check that the HTML is actually different, for example checking for the presence of the "send reminder" buttons?

evap/staff/templates/staff_semester_preparation_reminder.html
```diff
@@ -27,35 +28,47 @@
...
+                    </div>
                 </div>
-            </div>
-            <div class="card-body">
-                <table class="table table-striped">
-                    <thead>
+                <div class="card-body">
+                    <table class="table table-striped">
+                        <thead>
                         <tr>
```
niklasmohrin, line 65: Indentation

evap/staff/templates/staff_semester_preparation_reminder.html
```diff
@@ -27,35 +28,47 @@
...
                             <th class="width-percent-57">{% translate 'Name' %}</th>
                             <th class="width-percent-18">{% translate 'Start of evaluation' %}</th>
                             <th class="width-percent-25">{% translate 'Last modified by' %}</th>
-                       </tr>
-                    </thead>
-                    <tbody>
+                        </tr>
+                        </thead>
```
niklasmohrin, line 70: And here

evap/staff/templates/staff_semester_preparation_reminder.html
```diff
@@ -27,35 +28,47 @@
...
+                        <tbody>
                         {% for evaluation in evaluations %}
```
niklasmohrin, line 72: Indentation

niklasmohrin (changes requested):

evap/staff/tests/test_views.py
```diff
@@ -938,41 +938,37 @@ class TestSemesterPreparationReminderView(WebTestStaffModeWith200Check):
...
```
niklasmohrin, line 968: Currently, this would still pass if we would ignore the `mode=text` parameter, we need something like
```suggestion
        self.assertContains(response, self.evaluation.full_name)
        self.assertNotContains(response, "Send reminder")
```

niklasmohrin (approved):

PR #2410, opened 2025-03-17
Fix TomSelect autocompleting prefixes to existing options on tag inputs

richardebeling:
By default, TomSelect prefers previous options if the currently entered text is a prefix/substring of any of those, instead of adding a new option (note that "Add awe..." is not selected by default)

We have hidden the dropdown menu for tag inputs by attaching it to some div that is not in the DOM, so from the perspective of the user, the input just "disappears" when pressing enter (as the prefix-expanded already existing element is added and then deduplicated), or a magically expanded version is added, unless we tell tomselect to prefer adding a new option

Setting this option fixes that (although still hacky, but it works, so 🤷🏼‍♂️)

### DIFF
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -115,8 +115,12 @@
             const tagTomSelectOptions = {
                 create: true,
                 createOnBlur: true,
-                dropdownParent: document.createElement("div"),  // To make it invisible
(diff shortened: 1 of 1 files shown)


niklasmohrin (approved):
I don't get why it's hacky, doesn't it make perfect sense that we want to add tags by default?

janno42 (approved):
how beautiful

PR #2412, opened 2025-03-23
Log adding participants via User Profile edit page

### LINKED ISSUE #2381: Log adding participants via User Profile edit page
Adding a participant to an evaluation via the user profile edit page does not create a log entry for the evaluation.
Similar to adding a participant via the evaluation's form, this should also be logged.

jooooosef:
fix #2381


jooooosef:
I first had an implementation to not hardcode "participants" field in the forms.py:
```python
if added_evaluations or removed_evaluations:
	example_evaluation = next(iter(added_evaluations | removed_evaluations), None)
	if example_evaluation and isinstance(example_evaluation, LoggedModel):
		for field in type(example_evaluation)._meta.many_to_many:
			related_model = getattr(field.remote_field, "model", None)
			if isinstance(self.instance, related_model):
				field_name = field.name
				break
```
But this would find _any_ M2M field in the Evaluation Model, that has a UserProfile Model.
If there is a way to find the participants field dynamically (in case it changes that would be great), please enlighten me :)

niklasmohrin:
I don't know if you discussed this already with the others, but I would think that we want to change this behavior in general. From what I can tell, our current code just doesn't handle the case where these log entries would be generated: https://github.com/e-valuation/EvaP/blob/65863d4f8bec5b0a7034372f70c49a4984a15ab4/evap/evaluation/models_logging.py#L344-L347

From https://docs.djangoproject.com/en/5.1/ref/signals/#m2m-changed I would expect that this branch is taken when an edit like in our issue here is performed

jooooosef:
yes @niklasmohrin that was my first idea as well but see [commit where i removed general implementation](https://github.com/e-valuation/EvaP/pull/2412/commits/b86ba8911f683e6f81fbc021f82ef2bdf3b4d3d4).

As explained in the commit message can pk_set be None so I would have to add a pk_set is None check. But this would defeat the purpose of a general implementation since some changes would not be logged.
I discussed this with @richardebeling. Please correct if I'm wrong or misunderstood anything.

niklasmohrin:
Ah, right. Still, if the action is pre_clear, then we should be able to find which objects are related. My intuition is that Django sets `pk_set` to `None` to avoid including a possibly very large set of pks - however, we are fine with this cost, because we don't have such big tables. The only problem is finding the correct field name, but it should be doable

jooooosef:
After some reading, came up with [this](https://github.com/e-valuation/EvaP/pull/2412/commits/752e95c8151d65a61e603f126af1b07d52846dd4) implementation.

This is pretty similar to the previous one (see [here](https://github.com/e-valuation/EvaP/pull/2412/commits/ec1c2e95d9bea6c722c185fcf60ffde1f96d25a9)).

The sender is always the M2M "table" which gets changed (so the through model). Here this would be `<class 'evap.evaluation.models.Evaluation_participants'> `.  In the reverse case the `model` is the the LoggedModel, so here `<class 'evap.evaluation.models.Evaluation'>`. We can therefore use the `model` and `sender` to get the `field_name` similar as to how its on the forward case.
With 
```python
field = model._meta.get_field(field_name)
related_name = field.remote_field.get_accessor_name()
```
We can the the defined `related_name` from that m2m field, and use that in case the `pk_set` is None.
According to the [django docs](https://docs.djangoproject.com/en/5.0/ref/signals/#m2m-changed) `pk_set` is None when `pre_clear` is send. With `pre_add` and `pre_remove` the `pk_set` contains the added/removed primary keys. 
So if `pk_set` is not None, we can use `related_instances = model.objects.filter(pk__in=pk_set)`. If it is None, we can use `related_instances = getattr(instance, related_name).all()` which in our example would translate to something like `user.evaluations_participating_in.all()'. We then iterate over every `LoggedModel` in `related_instances` and create appropiate log entries. 

When `pre_clear` is sent, we only use the remove-action, since for example here not the participants of an eval get cleared, but the evals a user is participating in, so the user gets removed from those evals.

jooooosef:
Should I put an similar version of the tests into `test_models_logging.py` then? (without using the form)

niklasmohrin:
Impressive work on code and explanation so far, very well done!

Just to loop you in, I think that this is probably the right solution, but didn't want to give an approval yet as this is a very critical code area and I want to briefly discuss with @richardebeling and you on Monday to ensure we are all on the same page about the changes and consequences

richardebeling (commented):
looking mostly good, thanks

evap/evaluation/models_logging.py
```diff
@@ -341,21 +341,48 @@ def unlogged_fields(self):
...
+        if pk_set:
+            related_instances = model.objects.filter(pk__in=pk_set)
+        else:
+            # When action is pre_clear, pk_set is None, so we need to get the related instances from the instance itself
+            related_instances = getattr(instance, related_name).all()
+
+        for related_instance in related_instances:
+            if isinstance(related_instance, LoggedModel):
+                if field_name in related_instance.unlogged_fields:
+                    continue
```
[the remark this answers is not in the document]
jooooosef: no.

Since unlogged_fields is a propety and class propeties via the decorators `@property` and  `@classmethod` were depracated, we would have to use a method, which would also need to be callable class-side (`@classmethod`). We would then need to change all implementations of logged_fields including: 
```python
  @property
  def unlogged_fields(self):
      return super().unlogged_fields + ["evaluation"] + (["contributor"] if self.is_general else [])
```
from the Contribution model. Since this requries `self.is_general` and is therefore dependent on the instance, this is not possible.
We therefore need to keep this unfortunatly.

evap/evaluation/models_logging.py
```diff
@@ -341,28 +341,51 @@ def unlogged_fields(self):
...
+            if field_name in related_instance.unlogged_fields:
+                continue
+
+            if action == "pre_remove":
+                related_instance.log_m2m_change(field_name, FieldActionType.M2M_REMOVE, [instance.pk])
+            elif action == "pre_add":
+                related_instance.log_m2m_change(field_name, FieldActionType.M2M_ADD, [instance.pk])
+            elif action == "pre_clear":
+                # Since we are not clearing the LoggedModdel instance, we need to log the removal of the related instances
+                related_instance.log_m2m_change(field_name, FieldActionType.M2M_REMOVE, [instance.pk])
```
Kakadus, line 380: can we add a check here, that these cases are complete? If `action` would be typed like `Literal["pre_remove", ......]`, we could add a `assert_never(action)` that checks exhaustiveness both statically (does we check all cases we expect) and dynamically (an update does not introduce new action types).

Kakadus: (also applies further below)

evap/evaluation/models_logging.py
```diff
@@ -341,28 +341,51 @@ def unlogged_fields(self):
...
+        (field.name for field in model_class._meta.many_to_many if getattr(model_class, field.name).through == sender),
+        None,
+    )
+
+
 @receiver(m2m_changed)
-def _m2m_changed(sender, instance, action, reverse, model, pk_set, **kwargs):  # pylint: disable=unused-argument
-    if reverse:
+def _m2m_changed(sender, instance, action, reverse, model, pk_set, **kwargs):  # noqa: PLR0912
+    field_name = _get_m2m_field_name(model if reverse else type(instance), sender)
```
richardebeling, line 353: This is the only caller now, I don't think the function is a nice abstraction, my vote is to inline this again.

evap/evaluation/tests/test_models_logging.py
```diff
@@ -147,3 +147,16 @@ def test_simultaneous_add_and_remove(self):
...
+
+        # Temporarily modify unlogged_fields to include 'participants'
+        original_unlogged_fields = self.evaluation.unlogged_fields
+        self.evaluation.__class__.unlogged_fields = property(lambda self: original_unlogged_fields + ["participants"])
+
+        self.evaluation.participants.add(participant)
+        self.assertFalse(any("participants" in entry.data for entry in self.evaluation.related_logentries()))
+
+        # Restore original unlogged_fields or other test might have unexpected behavior
+        self.evaluation.__class__.unlogged_fields = property(lambda self: original_unlogged_fields)
```
richardebeling, line 162: If the assertion fails (or anything else goes wrong), we never arrive at the code that undoes our hot patching.

Can you use [`unittest.mock.patch`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) (or something similar) instead, with the context-manager approach, so that we have a guarantee that the modification is undone?

evap/evaluation/tests/test_models_logging.py
```diff
@@ -147,3 +147,16 @@ def test_simultaneous_add_and_remove(self):
             self.evaluation.related_logentries().order_by("id").last().data,
             {"participants": {"add": [participant2.id], "remove": [participant1.id]}},
         )
+
+    def test_logging_respects_unlogged_fields(self):
+        participant = baker.make(UserProfile)
```
richardebeling, line 152: There are two separate things here, the first is that our model logging code handles `unlogged_fields` correctly, and the second is that it handles m2m changes via the reverse relation. It would be nice to have the four test cases covered:
* "simple" field, not unlogged (this test already exists)
* "simple" field, unlogged: It would be nice if you could add this
* reverse m2m field, not unlogged: It would be nice if you could add this
* reverse m2m field, unlogged: This is what this tests does

evap/evaluation/models_logging.py
jooooosef: @niklasmohrin and @richardebeling say its okay without.
There are also the "post_*" options, which we dont handle. 

That also prob is not gonna change in the future.

evap/evaluation/tests/test_models_logging.py
jooooosef: This test is testing the "simple" field, unlogged.

I now also added the other tests.

evap/evaluation/models_logging.py
```diff
@@ -342,27 +342,47 @@ def unlogged_fields(self):
...
-            if getattr(type(instance), field.name).through == sender
-        ),
+        (field.name for field in model_class._meta.many_to_many if getattr(model_class, field.name).through == sender),
         None,
     )
+    if field_name is None:
+        return
 
-    if field_name in instance.unlogged_fields:
+    if not (issubclass(model, LoggedModel) if reverse else isinstance(instance, LoggedModel)):
```
richardebeling, line 354:
```suggestion
    if not issubclass(model_class, LoggedModel):
```

evap/evaluation/tests/test_models_logging.py
```diff
@@ -147,3 +148,44 @@ def test_simultaneous_add_and_remove(self):
...
+        participant.evaluations_participating_in.remove(self.evaluation)
+        self.assertEqual(
+            self.evaluation.related_logentries().order_by("id").last().data,
+            {"participants": {"remove": [participant.id]}},
+        )
+
+        participant.evaluations_participating_in.add(self.evaluation)
+        participant.evaluations_participating_in.clear()
+        self.assertEqual(
+            self.evaluation.related_logentries().order_by("id").last().data,
```
richardebeling, line 170: This could, by accident, just be the remove entry we observed above in L162. We could be a bit more assertive here by also asserting that there was at least one additional log entry

evap/evaluation/tests/test_models_logging.py
```diff
@@ -147,3 +148,44 @@ def test_simultaneous_add_and_remove(self):
...
+    def test_logging_m2m_reverse_changes(self):
+        participant = baker.make(UserProfile)
+
+        participant.evaluations_participating_in.add(self.evaluation)
+        self.assertEqual(
+            self.evaluation.related_logentries().order_by("id").last().data,
+            {"participants": {"add": [participant.id]}},
+        )
+
+        participant.evaluations_participating_in.remove(self.evaluation)
```
richardebeling, line 161: Please also test adding/removing more than one element at a time (and possibly also adding zero elements, since that might break something in the inversion logic we have)

evap/evaluation/tests/test_models_logging.py
```diff
@@ -147,3 +148,44 @@ def test_simultaneous_add_and_remove(self):
...
+        participant.evaluations_participating_in.clear()
+        self.assertEqual(
+            self.evaluation.related_logentries().order_by("id").last().data,
+            {"participants": {"remove": [participant.id]}},
+        )
+
+        result_no_field_name = _m2m_changed(None, None, None, True, Evaluation, None)
+        self.assertEqual(result_no_field_name, None)
+
+    def test_logging_respects_unlogged_fields(self):
```
richardebeling, line 177:
```suggestion
    def test_m2m_logging_respects_unlogged_fields(self):
```

evap/evaluation/tests/test_models_logging.py
```diff
@@ -147,3 +148,44 @@ def test_simultaneous_add_and_remove(self):
...
+        participant = baker.make(UserProfile)
+        original_unlogged_fields = self.evaluation.unlogged_fields
+
+        # Temporarily modify unlogged_fields to include 'participants'
+        with patch.object(
+            self.evaluation.__class__,
+            "unlogged_fields",
+            new=property(lambda self: original_unlogged_fields + ["participants"]),
+        ):
```
richardebeling, line 186: Does this not work? If so, why?
```suggestion
        with patch.object(Evaluation, "unlogged_fields", new=["participants"]):
```

evap/evaluation/tests/test_models_logging.py
jooooosef: I added an assertEqual after adding the evaluation to definitly check that it was added again.

richardebeling: I don't see how your last change addresses the issue I outlined above. The test now asserts that the last log entry after adding an additional participant compares equal to the last log entry before that operation, but before that, we already added a participant as the last operation, so in theory, we could still be comparing the same or two different log entries, the assertion would be true for both cases.

evap/evaluation/tests/test_models_logging.py
richardebeling: Adding more than one element in a single call is still not covered, is it?

jooooosef:
```python
participant.evaluations_participating_in.add(self.evaluation, evaluation2)
        self.assertEqual(
            self.evaluation.related_logentries().order_by("id").last().data,
            {"participants": {"add": [participant.id]}},
        )
        self.assertEqual(
            evaluation2.related_logentries().order_by("id").last().data,
            {"participants": {"add": [participant.id]}},
        )
```

This is adding more than on element in one call?

evap/evaluation/tests/test_models_logging.py
jooooosef:
```python
participant.evaluations_participating_in.remove(self.evaluation, evaluation2)
self.assertEqual(
    self.evaluation.related_logentries().order_by("id").last().data,
    {"participants": {"remove": [participant.id]}},
)
self.assertEqual(
    evaluation2.related_logentries().order_by("id").last().data,
    {"participants": {"remove": [participant.id]}},
)

# Check if evaluation was added again, so the remove is entry (for the clear) is not just read twice.
self.assertEqual(
    self.evaluation.related_logentries().order_by("id").last().data,
    {"participants": {"add": [participant.id]}},
)
self.assertEqual(
    self.evaluation.related_logentries().order_by("id").last().data,
    {"participants": {"remove": [participant.id]}},
)
```

This first removes the evaluations and checks that the last log entry is a remove action (the removal of more than one element).
Then it is added again, and checked if the last log entry is now an add action.
Then the clear action is performed and the assert checks if that last log entry is a remove.
So we now have:
remove --> check if remove action in log
add --> check if add action in log
clear --> check if remove action in log

This should make sure the clear action works properly, otherwise one of the asserts should fail, I think.

richardebeling (dismissed):
thanks :)

evap/evaluation/tests/test_models_logging.py
```diff
@@ -147,3 +148,73 @@ def test_simultaneous_add_and_remove(self):
...
+            self.evaluation.related_logentries().order_by("id").last().data,
+            {"participants": {"add": [participant.id]}},
+        )
+
+        participant.evaluations_participating_in.remove(self.evaluation)
+        self.assertEqual(
+            self.evaluation.related_logentries().order_by("id").last().data,
+            {"participants": {"remove": [participant.id]}},
+        )
+
```
niklasmohrin: This is already tested below
```suggestion
```

evap/evaluation/models_logging.py
```diff
@@ -342,27 +342,47 @@ def unlogged_fields(self):
...
+            # When action is pre_clear, pk_set is None, so we need to get the related instances from the instance itself
+            field = model._meta.get_field(field_name)
+            related_name = field.remote_field.get_accessor_name()
+            related_instances = getattr(instance, related_name).all()
+
+        for related_instance in related_instances:
+            if field_name in related_instance.unlogged_fields:
+                continue
+
+            if action == "pre_remove":
```
niklasmohrin: We previously discussed that this loop might be slow; to avoid doing this loop if action is none of these three strings, we should do the following before the `if reverse`:
```python
match action:
    case "pre_remove":
        action_type = FieldActionType.M2M_REMOVE
    case ... # the other two
    case _:
        return
```
Then, we should not use `action` again and only use `action_type`.

evap/evaluation/models_logging.py
```diff
@@ -342,27 +342,52 @@ def unlogged_fields(self):
...
         return
 
-    if action == "pre_remove":
-        instance.log_m2m_change(field_name, FieldActionType.M2M_REMOVE, list(pk_set))
-    elif action == "pre_add":
-        instance.log_m2m_change(field_name, FieldActionType.M2M_ADD, list(pk_set))
-    elif action == "pre_clear":
-        instance.log_m2m_change(field_name, FieldActionType.M2M_CLEAR, [])
+    if reverse:
+        action_type = None
```
niklasmohrin:
```suggestion
```

janno42:
> Thanks!
> 
> @janno42 My gut feeling is like last time, if we want to release a bugfix release I would hold this PR back just in case. 2025.7 can be the time of reverse logging and dropped courses. But do with it what you want

This is technically also a bugfix because the related issue is a bug :)

PR #2413, opened 2025-03-24
Explicitly disconnect signals when no longer needed

### LINKED ISSUE #2189: Assertion failure due to weak signal receivers having no guarantee of being garbage collected in time

richardebeling:
Fixes #2189.

### DIFF
--- a/evap/evaluation/tools.py
+++ b/evap/evaluation/tools.py
@@ -2,6 +2,7 @@ import datetime
 import typing
 from abc import ABC, abstractmethod
 from collections.abc import Iterable, Mapping
+from contextlib import contextmanager
 from typing import TYPE_CHECKING, Any, TypeVar
 from urllib.parse import quote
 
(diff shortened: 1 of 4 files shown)


evap/evaluation/tools.py
```diff
@@ -33,6 +34,15 @@
 CV = TypeVar("CV", bound=CellValue)
 
 
+@contextmanager
+def temporary_receiver(signal, func, **kwargs):
+    # semi-copy of https://github.com/django/django/blob/3266f2516c27dd25abebe8e8f7b8778650ab4f18/django/dispatch/dispatcher.py#L472
+    # with sane contextmanager behavior (receivers are unlinked on exit). We don't support lists of signals (for now).
+    signal.connect(receiver=func, **kwargs)
+    yield
+    signal.disconnect(receiver=func, **kwargs)
```
niklasmohrin, line 43: Shouldn't we do the following?
```suggestion
    try:
        yield
    finally:
        signal.disconnect(receiver=func, **kwargs)
```

niklasmohrin (changes requested):

evap/evaluation/tools.py
richardebeling: absolutely correct

niklasmohrin (approved):
Epic!

Kakadus (dismissed):

niklasmohrin:
Interesting that the lines for `operation == "approve"` are not covered :thinking:

niklasmohrin (approved):

PR #2414, opened 2025-03-24
Fix various TypeScript / ESLint warnings

niklasmohrin:
The remaining warnings are:
- Various "unsafe" operations in `datagrid.ts`. I am going to leave these as is for now, as they are somewhat contained and I don't know what the best approach is.
- `typescript-eslint/no-misused-promises`, which indicates that we passed an `async` function where a sync one was expected. Our two usages are fine though.

Importantly for the second point, calling `event.preventDefault()` in an async event handler is fine if done before the first `await`, [quote MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#description):

> Top-level code, up to and including the first await expression (if there is one), is run synchronously. In this way, an async function without an await expression will run synchronously.

Still, we might want to research whether this pattern can be enforced at compile time.

### DIFF
--- a/evap/grades/templates/grades_course_view.html
+++ b/evap/grades/templates/grades_course_view.html
@@ -1,5 +1,7 @@
 {% extends 'grades_course_base.html' %}
 
+{% load static %}
+
 {% block content %}
     {{ block.super }}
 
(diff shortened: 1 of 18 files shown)


evap/static/ts/data/evap.log
richardebeling, line 1: sneaky

richardebeling (approved):

Kakadus (approved):
looks good so far :+1:

richardebeling (approved):

PR #2419, opened 2025-04-03
fixup renovate configuration

Kakadus:
Some stuff was still missing:
- npm patch & sass
- django-fsm
- sortablejs types
- python

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -6,6 +6,7 @@
     enabled: true,
   },
   separateMajorMinor: false,
+  ignoreDeps: ["@types/sortablejs"], // we use a static version
   packageRules: [
     {
       dependencyNames: ["django-fsm"],
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -8,6 +8,11 @@
   separateMajorMinor: false,
   packageRules: [
     {
+      dependencyNames: ["django-fsm"],
+      groupSlug: "python",
+      allowedVersions: "!/^3.0.0$/", // 3.0.0 adds deprecation warnings without functional changes
+    },
+    {
       groupName: "python dependencies",
       groupSlug: "python",
       matchManagers: ["pep621"],
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -23,9 +23,9 @@
       matchManagers: ["github-actions"],
     },
     {
-      groupName: "ubuntu",
+      groupName: "ubuntu & python upgrade",
       groupSlug: "ubuntu",
-      matchDepNames: ["ubuntu"],
+      matchDepNames: ["ubuntu", "python"],
       dependencyDashboardApproval: true, // manual update when deployment is updated
     },
     {
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -19,6 +19,17 @@
       matchManagers: ["pep621"],
     },
     {
+      matchManagers: ["npm"],
+      matchUpdateTypes: ["patch"],
+      enabled: false,
+    },
+    {
+      matchManagers: ["npm"],
+      matchUpdateTypes: ["patch", "minor"],
+      matchDepNames: ["sass"],  // causes deprecation warnings in bootstrap
+      enabled: false
+    },
+    {
       groupName: "node dependencies",
       groupSlug: "node",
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -23,9 +40,9 @@
       matchManagers: ["github-actions"],
     },
     {
-      groupName: "ubuntu",
+      groupName: "ubuntu & python upgrade",
       groupSlug: "ubuntu",
-      matchDepNames: ["ubuntu"],
+      matchDepNames: ["ubuntu", "python"],
       dependencyDashboardApproval: true, // manual update when deployment is updated
     },
```
niklasmohrin, line 47: Dose this make it so that we get a suggestion for the current Python version of Ubuntu, or the latest Python version? I would expect the latter. I think then we should just drop Python updates, because the latest Python version might still be newer, and then this change blocks the Ubuntu update we actually want

niklasmohrin:
Can renovate do some kind of config-check? I personally don't know what config keys are valid, I can only judge that the diff seems to do the right thing, but cannot really test

.github/renovate.json5
Kakadus: The latter.
You are right, although I'd say that the python version will be updated in a new ubuntu version almost certainly, so we have to edit the version anyway. I thought that there may be a non-zero chance that the updated python version corresponds to the ubuntu python version and this won't increase the workload.
If we don't change the python in the ubuntu update, we may forget updating the supported python version in the toml file, making it unusable on that ubuntu version. The error caused by the update may remind us of that.
I also see ignoring python or even disabling both. Renovate won't be able to create a functional PR directly. That's why we initially picked manual PR creation for ubuntu.

Kakadus:
> Can renovate do some kind of config-check? I personally don't know what config keys are valid, I can only judge that the diff seems to do the right thing, but cannot really test

As I googled, renovate picks it up, if it is at branch `renovate/reconfigure`. I look if I can change it

.github/renovate.json5
richardebeling: My vote is to handle python the same as ubuntu: Make it show up in the Dashboard-Issue, but don't auto-open PRs for it, we handle it manually (and likely do not use Renovate-PR for this, but do everything manually)

If I understand correctly, this is exactly what the change here does, then 👍 from me on the current version here.

.github/renovate.json5
```diff
@@ -6,12 +6,29 @@
     enabled: true,
   },
   separateMajorMinor: false,
+  ignoreDeps: ["@types/sortablejs"], // we use a static version
   packageRules: [
```
richardebeling, line 10: Django updates are more involved than just clicking "accept" on a PR. Often, we have to wait for support of the new version from other dependencies. During that time, having a PR open doesn't really help I think. If it's grouped together with the other python dependencies, that makes it harder to single it out as "cannot yet upgrade". I think it would make sense to put it in a separate group, and maybe also have it only trigger manually?

richardebeling (approved):

.github/renovate.json5
niklasmohrin: Doesn't the current change group the ubuntu and Python updates together in one PR?

I would be fine with Richard's approach, which I understand as having one PR for ubuntu and one for Python, but both manually triggered

niklasmohrin:
Ough, running renovate-config-validator locally says it needs a github token. Why??? Okay, whatever then

.github/renovate.json5
richardebeling: Yes, this change here groups them. So far we've kept Ubuntu and Python in sync, so this makes sense to me, but I'd also be ok to have two separate checkboxes. My understanding is that the Renovate-created PR is mostly useless for these anyway, so personally, I wouldn't click the checkbox, but create the PR manually.

niklasmohrin: Okay, either way is fine

niklasmohrin (approved):
Splitting off Django makes sense, but we don't need to make the change right now, this way we can already move forward with the node dependencies

niklasmohrin:
nevermind, you can run the validator locally without a token. In fact, it is very easy: `nix shell nixpkgs#renovate --command renovate-config-validator`

PR #2427, opened 2025-04-08
Update django and django-webtest

Kakadus:
This fixes the leftover failures from #2418


richardebeling (approved):
None of the changes in 5.2 look breaking for us, tests pass without deprecation warnings (regarding django stuff)

Interesting things in the release notes:
* [`reverse`](https://docs.djangoproject.com/en/dev/ref/urlresolvers/#django.urls.reverse) now allows passing a `query` and `fragment` string, so we don't need to manually concatenate things to URLs anymore (will open separate PR)
* [`manage.py shell` now supports automatic imports](https://docs.djangoproject.com/en/dev/howto/custom-shell/#customizing-shell-auto-imports), making `shell_plus` superfluous. ~Maybe we can drop django-extensions (I don't think I'm using any of its functionality)~ Ah, we're still using `reset_db` sometimes, Django doesn't have anything native for that. Crazy.

niklasmohrin (approved):

PR #2428, opened 2025-04-11
Use django 5.2 features: `reverse` with query/fragment, auto-imports on `manage.py shell`

richardebeling:
depends on #2427.

We still need django-extensions for `reset_db`. Maybe we can get by without it, I think we could then drop the dependency. There is [one usage in `reload_testdata`](https://github.com/e-valuation/EvaP/blob/main/evap/development/management/commands/reload_testdata.py#L21) and [one usage in the load_production_backup script](https://github.com/e-valuation/evap-deployment/blob/770d35fd5cb64fb6f380a0aa87ab74455d03f461/load_production_backup.sh#L47) in the deployment repo.

### DIFF
--- a/evap/evaluation/management/commands/shell.py
+++ b/evap/evaluation/management/commands/shell.py
@@ -1,0 +1,28 @@
+from django.core.management.commands import shell
+
+
+class Command(shell.Command):
+    def get_auto_imports(self):
+        return super().get_auto_imports() + [
+            "django.conf.settings",
+            "django.contrib.auth.get_user_model",
+            "django.core.cache.cache",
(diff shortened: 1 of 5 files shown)


niklasmohrin (approved):
After merging, we should also update the mention of `shell_plus` in the wiki

For `reset_db`, we could also roll our own implementation, but I don't know if we want to. On the one hand, it's really only `DROP DATABASE X` and `CREATE DATABASE X ...` (and we can remove a lot of the features of `django_extensions`' implementation. On the other hand, maybe we don't want to care about details with database auth and what not

PR #2434, opened 2025-04-28
Automatically update email address on subdomain change

### LINKED ISSUE #1992: Change email address on login when subdomain changes
When a user's status group changes, their email address may also change. For example, emails are `name@student.institution.example.com` for students and change to `name@institution.example.com` when they become PhD students or staff.
If the names are unique across these domains, the change can be made automatically at login when using OIDC.

A setting should be added that specifies a list of tuples of email addresses that should be automatically changed in this way.
When a user logs in via OIDC and a new account would be created because their email address doesn't exist yet, this list should be checked first. If the domain is in the list, and an account exists for the other address, then the email address of that account should be changed to the new address, and the login should proceed as normal.
A notification about the automatic merge should be sent to the `CONTACT_EMAIL` address.

niklasmohrin:
Closes #1992

### DIFF
--- a/evap/evaluation/auth.py
+++ b/evap/evaluation/auth.py
@@ -1,16 +1,20 @@
 import inspect
+import logging
 from collections.abc import Callable, Iterable
 from functools import wraps
 
 from django.contrib.auth.backends import ModelBackend
 from django.conf import settings
 from django.core.exceptions import PermissionDenied
+from django.core.mail import EmailMessage
 from django.utils.decorators import method_decorator
 from mozilla_django_oidc.auth import OIDCAuthenticationBackend
 
 from evap.evaluation.models import UserProfile
 from evap.evaluation.tools import clean_email, openid_login_is_active, password_login_is_active
 from evap.rewards.tools import can_reward_points_be_used_by
+
+logger = logging.getLogger(__name__)
 
 
 class RequestAuthUserBackend(ModelBackend):
@@ -175,6 +179,18 @@ class OpenIDAuthenticationBackend(OIDCAuthenticationBackend):
             user.save()
         new_email = claims.get("email", "")
         if user.email != new_email:
+            notification = EmailMessage(
+                subject="[EvaP] User email changed automatically",
+                body=f"The email address of the user with id {user.pk} was automatically changed from {user.email} to {new_email}.",
+                to=[settings.CONTACT_EMAIL],
+            )
+
             user.email = new_email
             user.save()
+
+            try:
+                notification.send()
+            except Exception:
+                logger.exception("Failed to send notification:\n%s\n", notification.message())
+                raise
         return user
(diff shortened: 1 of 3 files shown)


hansegucker (approved):

richardebeling (approved):

evap/evaluation/tests/test_auth.py
```diff
@@ -144,6 +144,59 @@ def test_oidc_login(self):
...
 
+    @override_settings(ACTIVATE_OPEN_ID_LOGIN=True)
+    @patch.multiple(
+        auth.OIDCAuthenticationBackend,
+        get_userinfo=DEFAULT,
+        get_token=DEFAULT,
+        store_tokens=DEFAULT,
+        verify_token=DEFAULT,
+    )
+    def test_oidc_email_transitions(self, get_userinfo, verify_token, **_other_mocks):
```
Kakadus, line 155: I was wondering what happens if there already exists both an `institution.example.com` and `student.institution.example.com` login. This works currently, because we `yield email` before `settings.OIDC_EMAIL_TRANSITIONS.get(domain)`. In fact if the yield order was changed and you'd try to log in with such a user using the `institution.example.com` email, you get an IntegretyError as the accounts are not merged. Applies for example to `vincenzo.boston@institution.example.com`.

Maybe we should add a test that the "correct" account is used when both accounts exist? In a follow up ticket (or even here?) we could directly merge the two user accounts

evap/evaluation/auth.py
```diff
@@ -133,16 +135,28 @@ def reward_user_required(user):
 
 # see https://mozilla-django-oidc.readthedocs.io/en/stable/
 class OpenIDAuthenticationBackend(OIDCAuthenticationBackend):
+    @staticmethod
+    def possible_existing_account_emails(email: str) -> Iterable[str]:
+        yield email
+        name, domain = email.split("@", maxsplit=1)
```
Kakadus, line 141: TIL: email addresses can contain multiple `@` characters, if they are "quoted"

> Space and special characters "(),:;<>@[\] are allowed with restrictions (they are only allowed inside a quoted string...)

In fact, django uses `rsplit("@", maxsplit=1)` to get the domain part: https://github.com/django/django/blob/0f5dd0dff3049189a3fe71a62670b746543335d5/django/core/validators.py#L250

```suggestion
        name, domain = email.rsplit("@", maxsplit=1)
```

niklasmohrin: Yeah we also thought about that on Monday, but I thought that a) we do this wrong in other places in the codebase and b) this is only wrong if the settings contain a key with an @ in it, which would be wrong according to spec. I wanted to follow up on this separately

janno42 (approved):
Nice

richardebeling (approved):

PR #2435, opened 2025-04-28
Add program merge view

### LINKED ISSUE #2431: Merge Programs
It should be possible to merge `Programs` like currently already possible for `CourseType`s.
The UI should also look like the existing one.

richardebeling:
fixes #2431

Tests todo, but I think the changes so far are somewhat reviewable

### DIFF
--- a/evap/staff/templates/staff_program_merge.html
+++ b/evap/staff/templates/staff_program_merge.html
@@ -1,0 +1,39 @@
+{% extends 'staff_base.html' %}
+
+{% load evaluation_filters %}
+
+{% block breadcrumb %}
+    {{ block.super }}
+    <li class="breadcrumb-item"><a href="{% url 'staff:program_index' %}">{% translate 'Programs' %}</a></li>
+    <li class="breadcrumb-item">{% translate 'Merge programs' %}</li>
+{% endblock %}
+
+{% block content %}
+    {{ block.super }}
+    <h3>{% translate 'Merge programs' %}</h3>
+
+    <div class="card mb-3">
+        <div class="card-body">
+            <p>
+                {% blocktranslate %}The following courses currently have the program <strong>{{ other_instance }}</strong> and will
+                get the program <strong>{{ main_instance }}</strong> once you merge the programs. The program <strong>
+                {{ other_instance }}</strong> will be deleted after that.{% endblocktranslate %}
+            </p>
+
+            <ul>
+                {% for course in courses_with_other_program %}
+                    <li>{{ course }} ({{ course.semester }})</li>
+                {% endfor %}
+            </ul>
+        </div>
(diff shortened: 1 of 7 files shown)


janno42 (dismissed):
🔀 Merge the merging!

niklasmohrin:
Do we need to update logs?

janno42 (dismissed):

evap/staff/views.py
```diff
@@ -2120,6 +2121,52 @@ class ProgramIndexView(SuccessMessageMixin, SaveValidFormMixin, FormsetView):
...
+
+    return render(request, "staff_program_merge_selection.html", {"form": form})
+
+
+@manager_required
+def program_merge(request, main_id, other_id):
+    main_instance = get_object_or_404(Program, id=main_id)
+    other_instance = get_object_or_404(Program, id=other_id)
+
+    if request.method == "POST":
```
niklasmohrin, line 2141: Why are we not using the form here? Although the other view prevents me from navigating to, for example, /staff/programs/1/merge/1, I can still POST to it and thereby delete the master program. Although this is only managers, and only by modifying the URL, this seems like a bug to me (the same happens with the course type merge)

evap/staff/templates/staff_program_merge.html
```diff
@@ -0,0 +1,39 @@
...
+{% block content %}
+    {{ block.super }}
+    <h3>{% translate 'Merge programs' %}</h3>
+
+    <div class="card mb-3">
+        <div class="card-body">
+            <p>
+                {% blocktranslate %}The following courses currently have the program <strong>{{ other_instance }}</strong> and will
+                get the program <strong>{{ main_instance }}</strong> once you merge the programs. The program <strong>
+                {{ other_instance }}</strong> will be deleted after that.{% endblocktranslate %}
```
niklasmohrin, line 20: I _think_ this will show up more nicely in the translation file when using `blocktranslate trimmed` like with "Do you really want to delete the semester ..."

niklasmohrin (changes requested):
Logging sounds good; what do you think about the following?

evap/staff/views.py
```diff
@@ -2120,6 +2121,52 @@ class ProgramIndexView(SuccessMessageMixin, SaveValidFormMixin, FormsetView):
...
+        with transaction.atomic():
+            main_instance.import_names += other_instance.import_names
```
niklasmohrin, line 2143: This could lead to one name occurring multiple times in the list - we should introduce a database constraint eventually anyways

evap/staff/views.py
richardebeling: We agreed on adding an assert here that `main_instance != other_instance`

PR #2440, opened 2025-05-07
Replace `do_not_call_in_templates` property with `enum.nonmember`

niklasmohrin:
In #2325, I promised to do this


richardebeling (approved):
noice

PR #2446, opened 2025-05-20
Fix artifact glob for .mo files

niklasmohrin:
I am going to follow up with a PR to ensure that all artifacts specified in pyproject.toml are actually used.


richardebeling (approved):
oops

PR #2447, opened 2025-05-20
Add `tools/check_dist.py` to ensure all specified artifacts are used

niklasmohrin:
This should prevent situations like #2446. I also enabled source map generation for `./manage.py scss --production`.

Note that I used the `pathspec` library instead of the builtin `fnmatch` module, because that is what hatch uses: https://github.com/pypa/hatch/blob/4ebce0e1fe8bf0fcdef587a704c207a063d72575/backend/src/hatchling/builders/config.py#L185


niklasmohrin:
Ah, this just found another packaging bug: It turns out we don't generate CSS source maps in `./manage.py scss --production`: https://github.com/e-valuation/EvaP/blob/774db04ddbe7130be02684345d21b6a1339df5df/evap/evaluation/management/commands/scss.py#L34

This has been the behavior from #1529 onwards, but I don't really see why. @richardebeling @janno42 any idea?

pyproject.toml
richardebeling, line 1: Re missing css source maps (https://github.com/e-valuation/EvaP/pull/2447#issuecomment-2895354820):

I think we thought we wouldn't need them. I guess source maps would be extra files that wouldn't be accessed/loaded by a webbrowser in normal usage, but they would be loaded if you open the dev tools and look at the css? In that case, I think we're fine if remove the `--no-source-map`  argument and ship them.

richardebeling (dismissed):

nix/check-dist.py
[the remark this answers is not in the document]
niklasmohrin: I ended up adding some tests

tools/check_dist.py
```diff
@@ -0,0 +1,40 @@
...
+            if all(pattern.match_file(name) is None for name in included_files):
+                print(f"{wheel_path}: No file matches artifact: {artifact}")
+                status |= 1
+    return status
+
+
+def main(argv):
+    if len(argv) < 3:
+        print(f"USAGE: {argv[0]} <pyproject.toml path> <.whl path>...")
+        return 1
```
hansegucker, line 31: I don't like this direct usage of `argv`. I would vote for using https://docs.python.org/3/library/argparse.html.

hansegucker (changes requested):

hansegucker (approved):

richardebeling (approved):

PR #2451, opened 2025-05-26
Remove unused deconstructor

jooooosef:
After talking to @niklasmohrin and @janno42 this deconstructor would only be used if a instance of those classes would be serialized (see [Django Docs](https://docs.djangoproject.com/en/3.2/topics/migrations/#:~:text=Adding%20a%20deconstruct()%20method%C2%B6)). 

Since this is a form field in the student voting questionnaire, this will never happen and can therefore be safely removed.

(it was btw introduced in [this pr](https://github.com/e-valuation/EvaP/pull/1564))

### DIFF
--- a/evap/student/forms.py
+++ b/evap/student/forms.py
@@ -23,14 +23,6 @@ class TextAnswerField(forms.CharField):
         kwargs["widget"] = forms.Textarea(attrs={"related_answer_field_id": self.related_answer_field_id})
         super().__init__(*args, **kwargs)
 
-    def deconstruct(self):
-        name, path, args, kwargs = super().deconstruct()
-        del kwargs["required"]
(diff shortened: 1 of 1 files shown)


niklasmohrin (approved):

richardebeling (approved):
Alternative PR description (took me a while while reading the [docs on "Field deconstruction"](https://docs.djangoproject.com/en/5.2/howto/custom-model-fields/#field-deconstruction) to realize this)
> Both `HeadingField` and `RatingAnswerField` are `forms.Field`s, not model fields. [Field deconstruction](https://docs.djangoproject.com/en/5.2/howto/custom-model-fields/#field-deconstruction) is only relevant for model fields.

PR #2455, opened 2025-06-02
Make legal notice text configurable in settings

niklasmohrin:
On production, we should use something like

    from pathlib import Path
    from django.utils.safestring import mark_safe

    LEGAL_NOTICE_TEXT = mark_safe(Path("legal_notice_text.html").read_text())

I also made the view class based and made sure that the no_login_required decorator supports class based views.


evap/settings.py
```diff
@@ -166,6 +166,7 @@ class ManifestStaticFilesStorageWithJsReplacement(ManifestStaticFilesStorage):
 
 CONTACT_EMAIL = "webmaster@localhost"
 ALLOW_ANONYMOUS_FEEDBACK_MESSAGES = True
+LEGAL_NOTICE_TEXT = "Objection! (this is a default setting that the administrators should change, please contact them)"
```
richardebeling, line 169: Should probably be mentioned in https://github.com/e-valuation/evap-deployment/blob/main/productionsettings.template.py then

richardebeling (approved):

PR #2458, opened 2025-06-16
Add animation on evaluation deletion

### LINKED ISSUE #2457: Missing animation on evaluation deletion
After confirming the deletion of an evaluation on the staff semester page, the evaluation is actually deleted. However, the table is not updated.
As with deleting courses, the table row should be removed with an animation.

styrix560:
Closes #2457


janno42 (approved):
Thank you for your first contribution! 🎆 
Looking forward to more 😁

PR #2460, opened 2025-06-17
Take `INSTITUTION_EMAIL_REPLACEMENTS` into account when checking email transitions

niklasmohrin:
The current version wrongly reports that a user's email has changed, even though we changed it ourselves in `UserProfile.save`. This is because the user is queried with the email address cleaned with `clean_email`, and saved with the cleaned email, but we compare the stored email with the uncleaned one.

I now changed it so that the email is cleaned right after we fetch the claims from the OIDC provider. This ensures that we only deal with cleaned emails everywhere else.

### DIFF
--- a/evap/evaluation/auth.py
+++ b/evap/evaluation/auth.py
@@ -142,13 +142,17 @@ class OpenIDAuthenticationBackend(OIDCAuthenticationBackend):
         if previous_domain := settings.OIDC_EMAIL_TRANSITIONS.get(domain):
             yield f"{name}@{previous_domain}"
 
+    def get_userinfo(self, *args, **kwargs):
+        claims = super().get_userinfo(*args, **kwargs)
+        if "email" in claims:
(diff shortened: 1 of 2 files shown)


evap/evaluation/auth.py
niklasmohrin, line 1: I wonder whether we should now remove the email cleaning in `UserProfile.save`. I guess it's still useful for importers and stuff? But really they also need to clean the emails beforehand anyways, otherwise they might also miss existing users

niklasmohrin (commented):
Still needs a test

evap/evaluation/auth.py
janno42: What about manual changes? I'd say `UserProfile.save` is a canonical place to handle the cleaning.

janno42 (dismissed):

evap/evaluation/auth.py
richardebeling: I also think keeping it in `save` is reasonable

PR #2464, opened 2025-06-23
Validate data format in JSON importer

### LINKED ISSUE #2459: Validate JSON in JSON enrollment importer
Not a blocker for this PR, but it's a bit sad that the parsed object is just assumed to fit ImportDict, ideally we should validate that the data matches our expected schema.

We could look at pydantic (https://docs.pydantic.dev/dev/concepts/type_adapter/) or other solutions from here: https://stackoverflow.com/q/66665336/13679671

_Originally posted by @niklasmohrin in https://github.com/e-valuation/EvaP/pull/2187#discussion_r2127418513_

hansegucker:
Close #2459


evap/staff/tests/test_json_importer.py
```diff
@@ -587,6 +589,17 @@ def test_importer_log_email_sent(self):
         self.assertEqual(mail.outbox[0].subject, "[EvaP] JSON importer log")
         self.assertEqual(mail.outbox[0].recipients(), [manager.email])
 
+    def test_importer_wrong_data(self):
+        with self.assertRaises(ValidationError):
+            wrong_data = EXAMPLE_DATA.copy()
+            wrong_data["unexpected_attribute"] = 0
```
niklasmohrin, line 595: can we also test extra attributes nested more deeply?

pyproject.toml
```diff
@@ -11,6 +11,7 @@ dependencies = [
     "mozilla-django-oidc~=4.0.1",
     "openpyxl~=3.1.5",
     "psycopg~=3.2.3",
+    "pydantic>=2.11.7",
```
niklasmohrin:
```suggestion
    "pydantic~=2.11.7",
```

janno42 (dismissed):

richardebeling (dismissed):

evap/staff/importers/json.py
```diff
@@ -491,7 +497,9 @@ def _import_events(self, data: list[ImportEvent]) -> None:
                 main_evaluation.save()
 
     @transaction.atomic
-    def import_dict(self, data: ImportDict) -> None:
+    def import_dict(self, data: dict) -> None:
+        import_dict_adapter.validate_python(data, strict=True)
```
niklasmohrin: If you do
```suggestion
        validated_data = import_dict_adapter.validate_python(data, strict=True)
```
and use that instead of `data` below, the types passed into the function will match. Otherwise, we are still passing them as `Any`. This is kind of a whack mypy setting we have, I guess we should investigate whether we are ready to turn it off

richardebeling (approved):

PR #2465, opened 2025-06-23
Introduce new bootstrap_form styling option

### LINKED ISSUE #2463: Option to remove unnecessary blank space in bootstrap_form
Currently all forms via `bootstrap_form.html` have very limited options to change the styling / layout of the fields.
This leads to issues like #2124 where unnecessary whitespace leads to bad UI.

There should be an option for a form to display the label left-aligned and use the remaining space for the field itself.

jooooosef:
fix #2463 

I couldn't find any other modals with only one field, except the delegate preparation modal.


richardebeling (approved):

niklasmohrin:
Good old whitepsace

PR #2466, opened 2025-06-23
Don't wait for grades if all participants are external

### LINKED ISSUE #2332: Don't wait for grades if all participants are external
Currently EvaP holds back evaluation results until the grading process is finished (grade document has been uploaded or it has been confirmed that grading is finished but no grade document will be uploaded for the course), if `wait_for_grade_upload_before_publishing` is `True`.
In evaluations where all participants are external, we don't want to wait for the process to be finished (this can happen in D-School courses for example). So in this case, the results can be published as soon as the TextAnswer reviews are finished.

(`grading_process_is_finished` should always be `True` for Evaluations where all participants are external.)

jooooosef:
fix #2332

### DIFF
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -985,11 +985,19 @@ class Evaluation(LoggedModel):
         return RatingAnswerCounter.objects.filter(contribution__evaluation=self)
 
     @property
+    def all_participants_are_external(self):
+        for participant in self.participants.all():
+            if not participant.is_external:
+                return False
+        return True
+
+    @property
     def grading_process_is_finished(self):
         return (
             not self.wait_for_grade_upload_before_publishing
             or self.course.gets_no_grade_documents
             or self.course.final_grade_documents.exists()
+            or self.all_participants_are_external
         )
 
     @classmethod
--- a/evap/evaluation/tests/test_models.py
+++ b/evap/evaluation/tests/test_models.py
@@ -102,11 +102,14 @@ class TestEvaluations(WebTest):
         evaluation = Evaluation.objects.get(pk=evaluation.pk)
         self.assertEqual(evaluation.state, Evaluation.State.EVALUATED)
 
+    @override_settings(INSTITUTION_EMAIL_DOMAINS=["institution.example.com"])
     def test_in_evaluation_to_reviewed(self):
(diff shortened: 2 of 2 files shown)


evap/evaluation/models.py
```diff
@@ -984,12 +984,20 @@ def textanswer_review_state(self):
...
+                return False
+        return True
+
     @property
     def grading_process_is_finished(self):
         return (
             not self.wait_for_grade_upload_before_publishing
             or self.course.gets_no_grade_documents
             or self.course.final_grade_documents.exists()
+            or self.all_participants_are_external
```
richardebeling, line 998: This might look a bit strange in code. Can you link the issue (just by referencing #2332 in the code in a comment)?

evap/evaluation/tests/test_models.py
```diff
@@ -140,6 +147,31 @@ def test_in_evaluation_to_published(self):
...
+    @override_settings(INSTITUTION_EMAIL_DOMAINS=["institution.example.com"])
+    def test_in_evaluation_to_published_with_external_participants(self):
+        # participant is external, since no institution email domain is set
+        participant = baker.make(UserProfile, email="foo@example.com")
+        evaluation = baker.make(
+            Evaluation,
+            state=Evaluation.State.IN_EVALUATION,
+            participants=[participant],
+            vote_start_datetime=datetime.now() - timedelta(days=2),
+            vote_end_date=date.today() - timedelta(days=1),
```
richardebeling: For the test to pass between 0:00 and 3:00 (am), we need to use days=2 here (since the actual end is at 3am on the morning of the day after the "end" day, with 3 being configurable in the settings)

evap/evaluation/models.py
```diff
@@ -984,12 +984,20 @@ def textanswer_review_state(self):
     def ratinganswer_counters(self):
         return RatingAnswerCounter.objects.filter(contribution__evaluation=self)
 
+    @property
+    def all_participants_are_external(self):
+        for participant in self.participants.all():
+            if not participant.is_external:
+                return False
+        return True
```
```suggestion
        return all(participant.is_external for participant in self.participants.all())
```
or at this point we might just inline the function?

richardebeling (approved):

PR #2469, opened 2025-06-30
Add animation on evaluation deletion

### LINKED ISSUE #2302: 'can't give feedback' button does not clear textanswers
To skip a contributor and select "no answer" for all its questions, the "I can't give feedback about this contributor" button can be used. If the user changes a radio input value, this button gets activated again.

After clicking this button, textanswers for that contributor are still submitted, which might be unexpected.
When the contributor only has textanswers, the button is useless and cannot be activated again after being clicked once.

Instead, all textanswers for that contributor should be cleared as well - also textanswers attached to rating questions. When the user edits a textanswer, this button should be activated again.

styrix560:
fixes #2302


evap/static/ts/src/confirmation-modal.ts
```diff
@@ -68,6 +68,10 @@ export class ConfirmationModal extends HTMLElement {
         }
     };
 
+    show = () => {
+        this.dialog.show();
```
niklasmohrin: This should behave like when clicking the button, so
```suggestion
        this.dialog.showModal();
```
Additionally, the call to `showModal` above should now use `this.show()`

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -13,8 +13,18 @@ <h4 class="{% if not forloop.first %} mt-3 {% elif contributor %} me-auto {% end
...
                 </button>
+                <confirmation-modal class="mark-no-answer-modal" confirm-button-class="btn-danger" id="mark-no-answer-modal">
+                    <span slot="title">{% translate 'Bewertungen für Person zurücknehmen' %}</span>
+                    <span slot="action-text">{% translate 'Antworten zurücknehmen' %}</span>
+                    <span slot="question">
+                            {% blocktranslate trimmed %}
+                                Möchtest du alle zu dieser Person abgegebenen Antworten zu rücknehmen?
+                            {% endblocktranslate %}
+                    </span>
+                    <span slot="show-button" class="visually-hidden"></span>
```
niklasmohrin:
```suggestion
                    <span slot="show-button" class="d-none"></span>
```

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -87,4 +97,4 @@ <h4 class="{% if not forloop.first %} mt-3 {% elif contributor %} me-auto {% end
             {% endif %}
         {% endfor %}
     </div>
-{% endfor %}
+{% endfor %}
```
niklasmohrin: please make sure to always have a trailing newline in every file

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -13,8 +13,18 @@ <h4 class="{% if not forloop.first %} mt-3 {% elif contributor %} me-auto {% end
                 >
                     {% blocktranslate %}I can't give feedback about this contributor{% endblocktranslate %}
                 </button>
+                <confirmation-modal class="mark-no-answer-modal" confirm-button-class="btn-danger" id="mark-no-answer-modal">
+                    <span slot="title">{% translate 'Bewertungen für Person zurücknehmen' %}</span>
```
niklasmohrin: The translations in the source code should be in English, the German translation will then be done before the release by janno

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -13,8 +13,18 @@ <h4 class="{% if not forloop.first %} mt-3 {% elif contributor %} me-auto {% end
                 >
                     {% blocktranslate %}I can't give feedback about this contributor{% endblocktranslate %}
                 </button>
+                <confirmation-modal class="mark-no-answer-modal" confirm-button-class="btn-danger" id="mark-no-answer-modal">
```
niklasmohrin: The id is not used - and wrong, because it is duplicated when there is more than one contributor

evap/student/templates/student_vote.html
```diff
@@ -312,34 +312,73 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
             document.querySelectorAll("[data-mark-no-answers-for]").forEach(button => {
                 const contributorId = button.dataset.markNoAnswersFor;
                 const voteArea = document.getElementById(`vote-area-${contributorId}`);
+                const markNoAnswerModal = voteArea.querySelector(".mark-no-answer-modal");
+                const textareaAnswerFields = voteArea.querySelectorAll(".vote-inputs textarea");
                 const collapseToggle = voteArea.closest(".collapsible").querySelector(".collapse-toggle");
 
-                button.addEventListener("click", () => {
+                const onButtonClicked = () => {
```
niklasmohrin: This name is not really accurate anymore, how about
```suggestion
                const clearContributorAnswers = () => {
```

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -13,15 +13,15 @@ <h4 class="{% if not forloop.first %} mt-3 {% elif contributor %} me-auto {% end
                 >
                     {% blocktranslate %}I can't give feedback about this contributor{% endblocktranslate %}
                 </button>
-                <confirmation-modal class="mark-no-answer-modal" confirm-button-class="btn-danger" id="mark-no-answer-modal">
-                    <span slot="title">{% translate 'Bewertungen für Person zurücknehmen' %}</span>
-                    <span slot="action-text">{% translate 'Antworten zurücknehmen' %}</span>
+                <confirmation-modal class="mark-no-answer-modal" confirm-button-class="btn-danger">
+                    <span slot="title">{% translate 'Take back answer for this person' %}</span>
```
janno42: `answers`

janno42 (changes requested):
Nice! Please collapse all textboxes from the additional textanswers that have been cleared. When a user opens a collapsed contributor again, these textboxes should be collapsed, as they were initially.

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -88,3 +98,4 @@ <h4 class="{% if not forloop.first %} mt-3 {% elif contributor %} me-auto {% end
         {% endfor %}
     </div>
 {% endfor %}
+
```
richardebeling:
```suggestion
```

evap/student/templates/student_vote_questionnaire_group.html
richardebeling: Should we use more descriptive/precise wording than "Take back"? I can see how someone could understand "Take back" to mean: "Don't include with my submission, but keep the data, so I could later decide to include it again if I want to". We may want to use "reset" or "delete" or any other established terms for destructive operations.

evap/student/templates/student_vote.html
```diff
@@ -312,34 +312,73 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
+                        for (const radioButton of radioButtons) {
+                            if (radioButton.checked) {
+                                return true;
+                            }
+                        }
+                        return false;
+                    }
+
+                    if (shouldShowModal()) {
+                        markNoAnswerModal.show();
```
niklasmohrin: @richardebeling

> @niklasmohrin can you take a look at the modal logic? I think it matches [what we did for the reset evaluation button](https://github.com/e-valuation/EvaP/blob/30f7bba46bbc0642472fbb9c97cdea55e16cf0be/evap/staff/templates/staff_evaluation_operation.html#L87).

Ohh, so we _are_ already doing this somewhere else. I if remember correctly, Felix and I concluded that we ended up not needing this, but I guess we were wrong. I then suggested the logic for this PR to Nils. However, it is still interesting to consider the approach from the evaluation-operation form.

First, we would have to introduce a form. I think that it _could_ be a nice control flow abstraction to have a single form where you submit with for example `<button type=submit form=contributor-mark-no-answer-form name=contributor-id value={{ contributor.id}}>` and the form has a submit handler that does all of this. The modals also submit to this form and again have a data attribute marking them as "trusted to delete the answers". However, the modal reopening when submitting with a normal button would need to somehow keep track of which button was used to submit the form. Alternatively, each contributor could have their own modal or even their own form. So this makes things already a bit complicated.

Then, there is also the question of who handles the button activation. It seems that this would be done in the submit handler of the form, but that would kind of defeat the purpose of introducing the form as the overarching controller, because it needs to reach down to the specific buttons after all.

So I think that overall this current logic makes the most sense here. I would be happy to talk about a refactor towards forms in a separate PR though, I still believe that there could be a particularly nice implementation in that idea. For now, I suggest that we keep this PR as is with the additional change that @styrix560 should change the other occurrence of `.dialog.showModal()` to the new `.show()` function

evap/student/templates/student_vote_questionnaire_group.html
janno42: I suggest using `Delete my answers about this person` then.

evap/student/templates/student_vote.html
richardebeling: Sounds good, let's leave it as is here, I'm fine with the current code.

styrix560: #2479

evap/student/templates/student_vote.html
```diff
@@ -116,16 +116,15 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
                         <div class="card collapsible{% if not forloop.last %} mb-3{% endif %}">
                             <div class="card-header d-flex tab-row">
                                 <div class="me-auto">
-                                    <button class="collapse-toggle{% if errors_exist and not contributor_has_errors %} collapsed tab-selectable{% endif %} bg-transparent" data-bs-toggle="collapse"
-                                        data-bs-target="#vote-area-{{ contributor.id }}" aria-expanded="false" aria-controls="vote-area-{{ contributor.id }}" tabindex="-1" type="button">
+                                    <button class="collapse-toggle{% if errors_exist and not contributor_has_errors %} collapsed tab-selectable{% endif %} bg-transparent" data-bs-toggle="collapse" data-bs-target="#vote-area-{{ contributor.id }}" aria-expanded="false" aria-controls="vote-area-{{ contributor.id }}" tabindex="-1" type="button">
```
richardebeling: It looks like there are quire some formatting changes here that are unrelated to your intended changes. Please take a look again and revert anything that isn't intentionally changed here

richardebeling: student_vote.html still has [formatting changes in line 119](https://github.com/e-valuation/EvaP/pull/2469/files#diff-8428cbf7196d5ea3af0a2d1ba803a42f924156e06765c285dd5c56e16d2c6f81L119-R120).

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -13,8 +13,16 @@ <h4 class="{% if not forloop.first %} mt-3 {% elif contributor %} me-auto {% end
                 >
                     {% blocktranslate %}I can't give feedback about this contributor{% endblocktranslate %}
                 </button>
+                <confirmation-modal class="mark-no-answer-modal" confirm-button-class="btn-danger">
+                    <span slot="title">{% translate 'Delete answers about this person' %}</span>
+                    <span slot="action-text">{% translate 'Delete answers' %}</span>
+                    <span slot="question">
+                            {% blocktranslate trimmed %}Do you want to delete your answers about this person?{% endblocktranslate %}
```
richardebeling:
```suggestion
                        {% blocktranslate trimmed %}Do you want to delete your answers about this person?{% endblocktranslate %}
```

evap/student/templates/student_vote.html
```diff
@@ -309,42 +309,97 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
-                    // Disable this button, until user changes a value
-                    button.classList.remove("tab-selectable");
-                    button.disabled = true;
+                    collapseEmptyTextanswerFields();
+                }
+                const collapseEmptyTextanswerFields = () => {
+                    textareaAnswerFields.forEach(textareaAnswerField => {
+                        if (textareaAnswerField.value.trim().length > 0) return;
+                        const parent = textareaAnswerField.closest('.row-question');
+                        console.log(parent);
```
richardebeling: console.log()

evap/student/templates/student_vote.html
```diff
@@ -309,42 +309,97 @@ <h3 class="mb-3">{{ evaluation.full_name }} ({{ evaluation.course.semester.name
...
+                            }
+                        }
+
+                        const radioButtons = voteArea.querySelectorAll(".vote-inputs [type=radio]:not([value='6'])");
+                        for (const radioButton of radioButtons) {
+                            if (radioButton.checked) {
+                                return true;
+                            }
+                        }
+                        return false;
```
niklasmohrin: I think we can do something like this:
```suggestion
                        return voteArea.querySelector(".vote-inputs [type=radio]:not([value='6']):checked") !== null;
```

evap/static/ts/src/confirmation-modal.ts
```diff
@@ -68,6 +68,10 @@ export class ConfirmationModal extends HTMLElement {
         }
     };
 
+    show = () => {
```
niklasmohrin: There is another occurrence of showModal in staff_evaluation_operation that should be changed to use this new .show function, as mentioned here: https://github.com/e-valuation/EvaP/pull/2469#discussion_r2203501978

niklasmohrin, line 71: There is another occurrence of showModal in staff_evaluation_operation that should be changed to use this new .show function, as mentioned here: https://github.com/e-valuation/EvaP/pull/2469#discussion_r2203501978

evap/student/templates/student_vote.html
niklasmohrin, line 1: This has become quite intricate, I think we should have frontend tests for the behavior for
1. when does the modal show
2. when do things collapse

Testing the tabbing is out of scope I think, because there is more to test outside of your changes.

Writing these tests can be quite hairy, please talk to us if you need help. Your tests should go into this file, which already contains similar test cases: https://github.com/e-valuation/EvaP/blob/4eac3c6c78de205c9f6b023b201024816ec1a026/evap/student/tests/test_live.py

niklasmohrin (commented):
Nice work, there seem to be a handful of moving parts, but complexity is still manageable in my opinion.

Please make sure to actively review the code in the Files Changed tab of github to ensure everything is clean before requesting reviews, right now there is for example console.log, unrelated formatting, or unnecessarily moved lines like at the `// Disable this button, until user changes a value` line. Otherwise we reviewers might overlook the important bits and the PR could take longer to merge overall.

I think this PR is looking very promising! I hope we can manage to capture the changes in tests so that we reduce the risk that someone messes this up on accident at a later time - it's super hard to always remember all the small UI flows when reviewing, so having tests as a safety net is a great help.

niklasmohrin (changes requested):
not changed since last time

evap/student/templates/student_vote.html
styrix560: we cannot, because the checked in my code is not a css property, but instead exists on the javascript object

evap/student/templates/student_vote.html
styrix560: Just as a list of todos for me:
When does the modal show (at least one of):
- Any of the text fields have non-whitespace content
- Any of the radio buttons have a value that is not "No answer"

A textbox collapses if and only if it is empty and one of the buttons "I can't give feedback..." (+ modal) or "Collapse contributor" is pressed.

niklasmohrin:
Can you briefly explain what caused the modal showing up on hover?

evap/student/templates/student_vote.html
niklasmohrin: I think they refer to the same thing: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:checked

evap/student/tests/test_live.py
```diff
@@ -121,3 +122,55 @@ def test_skip_contributor_clears_warning(self) -> None:
...
         self.assertEqual(len(self.selenium.find_elements(By.CSS_SELECTOR, f"#vote-area-{id_} .choice-error")), 0)
+
+    def test_skip_contributor_modal_appears(self) -> None:
+        def assert_modal_visible_and_close():
+            modal = self.selenium.find_elements(By.CSS_SELECTOR, "confirmation-modal.mark-no-answer-modal")
+            self.assertEqual(len(modal), 2)
+            cancel_button = modal[0].shadow_root.find_elements(By.CSS_SELECTOR, "section.button-area > button")
+            self.assertEqual(len(cancel_button), 1)
+            cancel_button[0].click()
+            time.sleep(1)
```
niklasmohrin: Are the sleep calls really needed? I thought that selenium waits until all events triggered by the click are over?

Either way, in this case here we can also press Esc instead of clicking on the button. Then the modal will close instantly, and this is standard browser behavior, so selenium should definitely know what is happening

evap/student/tests/test_live.py
```diff
@@ -121,3 +122,55 @@ def test_skip_contributor_clears_warning(self) -> None:
 
         id_ = button.get_attribute("data-mark-no-answers-for")
         self.assertEqual(len(self.selenium.find_elements(By.CSS_SELECTOR, f"#vote-area-{id_} .choice-error")), 0)
+
+    def test_skip_contributor_modal_appears(self) -> None:
+        def assert_modal_visible_and_close():
+            modal = self.selenium.find_elements(By.CSS_SELECTOR, "confirmation-modal.mark-no-answer-modal")
+            self.assertEqual(len(modal), 2)
+            cancel_button = modal[0].shadow_root.find_elements(By.CSS_SELECTOR, "section.button-area > button")
```
niklasmohrin: why the `[0]`? I think what we really want to do is something along the lines of `confirmation-modal.mark-no-answer-modal:has(dialog:open)`

evap/student/tests/test_live.py
```diff
@@ -121,3 +122,55 @@ def test_skip_contributor_clears_warning(self) -> None:
...
+        self.selenium.get(self.url)
+
+        button = self.wait.until(presence_of_element_located((By.CSS_SELECTOR, "[data-mark-no-answers-for]")))
+        id_ = button.get_attribute("data-mark-no-answers-for")
+        vote_area = self.selenium.find_element(By.ID, f"vote-area-{id_}")
+        radio_buttons = vote_area.find_elements(By.CSS_SELECTOR, "input[value='1'] + label.vote-btn")
+        self.assertEqual(len(radio_buttons), 1)
+
+        # Waiting for element_to_be_clickable does not work unfortunately
+        time.sleep(1)
```
niklasmohrin: but does something not work then? the only thing we need is that the event handlers from the start are set, right? so why do we need to wait again here?

evap/student/tests/test_live.py
```diff
@@ -121,3 +122,55 @@ def test_skip_contributor_clears_warning(self) -> None:
...
+        textarea.send_keys(Keys.BACKSPACE)
+        textarea = self.wait.until(
+            expected_conditions.element_to_be_clickable(
+                (By.CSS_SELECTOR, f"textarea[name='{textareas[1].get_attribute('name')}']")
+            )
+        )
+        textarea.click()
+        textarea.send_keys("b")
+        button.click()
+        assert_modal_visible_and_close()
```
niklasmohrin: I don't understand, we replace 'a' with 'b' and click again and then the modal shows again; what is the test? Isn't the interesting bit that it does _not_ show if we edit textareas somewhere else?

niklasmohrin (commented):
I didn't thoroughly check the implementation now, except that I saw some console log and an empty line being added. I will come back to that later. In the meantime, please perform a quick self-review to sort out these small things.

For the tests, I have some questions.

(I also unresolved a previous discussion, please take a look)

evap/student/tests/test_live.py
```diff
@@ -121,3 +119,44 @@ def test_skip_contributor_clears_warning(self) -> None:
...
+            self.assertEqual(len(modals), 0)
+
+        self.selenium.get(self.url)
+
+        button = self.wait.until(presence_of_element_located((By.CSS_SELECTOR, "[data-mark-no-answers-for]")))
+        id_ = button.get_attribute("data-mark-no-answers-for")
+        vote_area = self.selenium.find_element(By.ID, f"vote-area-{id_}")
+
+        textareas = vote_area.find_elements(By.CSS_SELECTOR, "textarea")
+        textarea = vote_area.find_element(By.CSS_SELECTOR, f"textarea[name='{textareas[0].get_attribute('name')}']")
```
fekoch:
```suggestion
        textarea = vote_area.find_elements(By.CSS_SELECTOR, "textarea")[0]
```

evap/student/tests/test_live.py
```diff
@@ -121,3 +119,44 @@ def test_skip_contributor_clears_warning(self) -> None:
 
         id_ = button.get_attribute("data-mark-no-answers-for")
         self.assertEqual(len(self.selenium.find_elements(By.CSS_SELECTOR, f"#vote-area-{id_} .choice-error")), 0)
+
+    def test_skip_contributor_modal_appears(self) -> None:
```
fekoch: You should also test, if confirming the deletion really deletes the textanswers

evap/student/tests/test_live.py
```diff
@@ -121,3 +119,44 @@ def test_skip_contributor_clears_warning(self) -> None:
...
+        self.selenium.get(self.url)
+
+        button = self.wait.until(presence_of_element_located((By.CSS_SELECTOR, "[data-mark-no-answers-for]")))
+        id_ = button.get_attribute("data-mark-no-answers-for")
+        vote_area = self.selenium.find_element(By.ID, f"vote-area-{id_}")
+
+        textareas = vote_area.find_elements(By.CSS_SELECTOR, "textarea")
+        textarea = vote_area.find_element(By.CSS_SELECTOR, f"textarea[name='{textareas[0].get_attribute('name')}']")
+        radio_buttons = vote_area.find_elements(By.CSS_SELECTOR, "input[value='1'] + label.vote-btn")
+        self.assertEqual(len(radio_buttons), 1)
```
fekoch: This is irrelevant.

evap/student/tests/test_live.py
```diff
@@ -121,3 +119,44 @@ def test_skip_contributor_clears_warning(self) -> None:
...
+
+        open_textanswer = vote_area.find_element(By.CSS_SELECTOR, "button.btn-textanswer")
```
fekoch, line 159: this seems like it does not serve a purpose anymore?

janno42 (approved):
✅ UI checked
✅ Fulfills requirements

evap/student/templates/student_vote.html
```diff
@@ -366,42 +366,88 @@ <h3 class="mb-3 d-flex justify-content-between flex-column-mobile">
...
+                            button.click();
+                        }
+                    });
+                };
+                const clearContributorAnswers = () => {
                     voteArea.querySelectorAll(".vote-inputs [type=radio][value='6']").forEach(radioInput => {
                         radioInput.checked = true;
                         clearChoiceError(radioInput);
                     });
+                    textareaAnswerFields.forEach((textarea) => {
```
richardebeling: Let's pull the `voteArea.querySelectorAll(".vote-inputs textarea")` call into this function here and copy it into `collapseEmptyTextanswerFields`, so that if anything modified the page and added or removed text area inputs, this function still works correctly, and to be consistent with the vote-inputs.

evap/student/templates/student_vote.html
```diff
@@ -366,42 +366,88 @@ <h3 class="mb-3 d-flex justify-content-between flex-column-mobile">
...
+                        textarea.value = "";
+                        textarea.dispatchEvent(new Event("input"));
```
richardebeling, line 396: Is this required? If yes, can we add a short comment why?

evap/student/templates/student_vote.html
```diff
@@ -366,42 +366,88 @@ <h3 class="mb-3 d-flex justify-content-between flex-column-mobile">
...
+                }
+                markNoAnswerButton.addEventListener("click", () => {
+                    const shouldShowModal = () => {
+                        for (const textarea of textareaAnswerFields) {
+                            if (textarea.value.trim() !== "") {
+                                return true;
+                            }
+                        }
+                        return false;
+                    };
```
richardebeling:
```suggestion
                    const shouldShowModal = textAnswerFields.values().some(textarea => textarea.value.trim() !== "");
```

evap/student/templates/student_vote.html
```diff
@@ -366,42 +366,88 @@ <h3 class="mb-3 d-flex justify-content-between flex-column-mobile">
...
-                            collapseToggle.classList.remove("tab-selectable");
-                            button.classList.add("tab-selectable");
-                            button.disabled = false;
+                            enableMarkNoAnswerButton();
                         });
                     });
+                textareaAnswerFields.forEach((textarea) => {
+                    textarea.addEventListener("input", () => {
+                        enableMarkNoAnswerButton();
+                    });
```
richardebeling: I think you can remove one layer of call indirection here and above and use
```suggestion
                    textarea.addEventListener("input", enableMarkNoAnswerButton);
```

evap/student/tests/test_live.py
```diff
@@ -121,3 +122,47 @@ def test_skip_contributor_clears_warning(self) -> None:
...
+        vote_area = self.selenium.find_element(By.ID, f"vote-area-{id_}")
+
+        radio_button = vote_area.find_element(By.CSS_SELECTOR, "input[value='1'] + label.vote-btn")
+
+        open_textanswer = vote_area.find_element(By.CSS_SELECTOR, "button.btn-textanswer")
+        open_textanswer.click()
+
+        radio_button.click()
+        button.click()
+        self.assertEqual(len(self.get_open_modals()), 0)
```
richardebeling, line 147: We should also assert that the text area has been collapsed again

evap/student/tests/test_live.py
```diff
@@ -121,3 +122,47 @@ def test_skip_contributor_clears_warning(self) -> None:
...
+        open_textanswer.click()
+
+        textarea.click()
+        textarea.send_keys("a")
+        button.click()
+        modals = self.get_open_modals()
+        self.assertEqual(len(modals), 1)
+        modal = modals[0]
+        confirm_button = modal.shadow_root.find_element(By.CSS_SELECTOR, "button[data-event-type='confirm']")
+        confirm_button.click()
```
richardebeling, line 172: Since we had a bug with dismissed modals still execution their actions in the past: Please make this test  go through the modal two times. The first time, close it again without accepting. Assert that the text is not changed. Then open again, accept, and assert that the text was changed.

evap/student/templates/student_vote.html
styrix560: Dispatching the input event is required, because otherwise the button to open the text remains black (because it still thinks, that the textarea contains text). Should I also describe this in a comment?

evap/student/templates/student_vote.html
styrix560: We duplicate this call 4 different times now. Let me know, if that's fine for you.

evap/student/tests/test_live.py
```diff
@@ -121,3 +123,54 @@ def test_skip_contributor_clears_warning(self) -> None:
...
+        open_textanswer = vote_area.find_element(By.CSS_SELECTOR, "button.btn-textanswer")
+        open_textanswer.click()
+
+        collapsible = vote_area.find_element(By.CSS_SELECTOR, "div.collapse.show")
+
+        radio_button.click()
+        button.click()
+        self.assertEqual(len(self.get_open_modals()), 0)
+        classes = collapsible.get_attribute("class") or ""
+        self.assertNotIn("show", classes.split())
```
fekoch:
```suggestion
        self.assertNotIn("show", classes)
```

fekoch: or better `self.assertFalse(collapsible.is_displayed())`

evap/student/templates/student_vote.html
richardebeling: Works for me. If you want to get rid of the duplication, move the code out into a function so that the users looks something like
```typescript
getTextAreaAnswerFields().forEach((textarea) => { /*...*/ }
```
(but the list is still computed on demand instead of on page load)

evap/student/templates/student_vote_questionnaire_group.html
```diff
@@ -13,8 +13,16 @@ <h4 class="{% if not forloop.first %} mt-3 {% elif contributor %} me-auto {% end
                 >
                     {% blocktranslate %}I can't give feedback about this contributor{% endblocktranslate %}
                 </button>
+                <confirmation-modal class="mark-no-answer-modal" confirm-button-class="btn-danger">
+                    <span slot="title">{% translate 'Delete answers about this person' %}</span>
+                    <span slot="action-text">{% translate 'Delete answers' %}</span>
+                    <span slot="question">
+                        {% blocktranslate trimmed %}Do you want to delete your answers about this person?{% endblocktranslate %}
```
niklasmohrin: Since there is no whitespace in there, we don't need `trimmed`:
```suggestion
                        {% blocktranslate %}Do you want to delete your answers about this person?{% endblocktranslate %}
```

niklasmohrin: Do we want to mention something like "including already written textanswers" @janno42 ?

niklasmohrin (approved):
Very nice, thank you! Some closing remarks:

evap/student/templates/student_vote_questionnaire_group.html
janno42: Let's use `Do you want to delete your answers about this person? This includes written text answers.`

PR #2470, opened 2025-06-30
Rename degrees to programs in logs

### LINKED ISSUE #2443: Rename degree to program in logs
In #2282 we renamed `Degree` to `Program` but we forgot to replace `degrees` with `programs` in the `LogEntry`s.

Similar to https://github.com/e-valuation/EvaP/blob/main/evap/evaluation/migrations/0127_fix_logentry_types.py we should fix the original migration (0145) and add a new additional migration calling the operation.

ronjakrg:
Attempts to fix #2443


evap/evaluation/migrations/0149_rename_degrees_to_programs.py
niklasmohrin, line 1: The current test failure is because the `dependencies` value of the migration is empty. It should just be the name of the previous migration (0148_...) - then our CI should stop complaining

evap/evaluation/templates/progress_bar.html
niklasmohrin, line 1: These files are from your other pull request and should not be included in this one. You can fix this by running:
```bash
git rebase --interactive main
```
This will open your editor which will contain two lines with your two commits. In the first line, change "pick" to "drop" and then save and exit the file. Afterwards, `git log` should show that there is only one commit after main. To update this branch, you have to use `git push --force`, because you will then delete the old commits here

... and if something doesn't work, we are happy to take a look why on Monday

ronjakrg: Thanks! No idea how this happened

evap/evaluation/migrations/0149_rename_degrees_to_programs.py
```diff
@@ -0,0 +1,19 @@
+import importlib
+from django.db import migrations
+
+original_migration_module = importlib.import_module("evap.evaluation.migrations.0145_rename_degree_to_program")
+
+
+class Migration(migrations.Migration):
+    """
+    We forgot to migrate log entries in 0145, so this migrations fixes this.
```
hansegucker:
```suggestion
    We forgot to migrate log entries in 0145, so this migration fixes this.
```

hansegucker (dismissed):
Only a small typo ;-)

evap/evaluation/migrations/0145_rename_degree_to_program.py
```diff
@@ -1,6 +1,22 @@
 from django.db import migrations, models
 
 
+def logentries_degrees_to_programs(apps, _schema_editor):
+    LogEntry = apps.get_model("evaluation", "LogEntry")
+    for entry in LogEntry.objects.filter(content_type__app_label="evaluation", content_type__model="course"):
+        if "degrees" in entry.data:
+            entry.data["programs"] = entry.data.pop("degrees")
```
niklasmohrin, line 9: @richardebeling Should we `assert "programs" not in entry.data` before?

richardebeling: I don't know if there are cases where having the migration being idempotent (= noop on second invocation) is beneficial. Naively, my intuition about Django migrations is that this shouldn't happen, but sometimes Django is full of surprises. No strong opinion on the `assert` here, we can add it if you want, I wouldn't have added it on my own.

niklasmohrin: Okay so Richard and I briefly discussed whether we actually want to be more defensive here for the case that a log entry would exist that has both `data["degrees"]` and `data["programs"]` set, as the current code would just lose the "programs" entry. This could hypothetically happen if a log entry object would be reused and thus be in use before and after we installed the update that changed the name from degree to program.

However, we don't think that this happens in practice and thus doesn't warrant the extra code. Still, let's add the assertion so that we abort the migration if such a weird log entry exists.

Before merging, maybe @janno42 could run the following code on a production shell to check that @richardebeling and I are not missing anything?

```python
for entry in LogEntry.objects.filter(content_type__app_label="evaluation", content_type__model="course"):
    if "degrees" in entry.data and "programs" in entry.data:
        print("Cowabunga, mother")
```

janno42: We don't have both on production, I don't think that we need the assertion

niklasmohrin: Thanks :)

I would still prefer to have the assertion, it should never fail, and if it does then it means that it saved users from data loss and that we have to fix the migration

Issue #2046, opened 2023-10-23
Test frontend "looks"


niklasmohrin:
If we want to pay more attention to how EvaP looks on mobile, we should come up a way to automatically check that our cool CSS keeps working in the future.

One possible approach would be screenshot tests, that is, automatically taking screenshots of rendered pages and comparing them against a previous screenshot (that is in the repository). Every PR that changes the CSS would probably also include an update of the screenshot (which reviewers could compare and see that it still looks okay). The downside is of course that PRs get bloated with updates to the screenshots and that they take up space in the git repo forever, which could lead to slow clones. We should investigate how other projects do this, one other project someone presented me had a separate branch for these screenshots and they were reviewed after the fact, but this doesn't seem optimal to me.

niklasmohrin:
Good idea, it would be cool if the "do the screenshot" thing would be a script that I could also run locally to use with git bisect for example. 

Another thought: we could probably cache the screenshot of each commit for efficiency and for being able to inspect if from the GitHub actions panel. If this works like I am currently thinking it does, it seems like a nice workflow to me :D

karyon:
Regarding the repo size concern, there is git-lfs, which generally just works, except that github [apparently ](https://github.com/orgs/community/discussions/15267)doesn't show diffs for images in git lfs. Which is quite a letdown.

hansegucker:
With the current progress of implementing live tests with Selenium, we are able to programmatically create screenshots. Together with a tool like Argos CI or Percy, visual comparisons should be possible.

PR #2471, opened 2025-06-30
Don't send reminders shortly after evaluation start

### LINKED ISSUE #2400: Don't remind shortly after start
If you have set up evaluation reminders to be sent, e.g.,  3 and 1 days before an evaluation ends, and the evaluation runs for 4 days starting in the evening, a start notification will be sent, followed by a first reminder in the morning.
Instead, the reminder should be skipped if the evaluation started only yesterday.

If the reminder was triggered by another evaluation, the evaluation that just started should still be listed in the reminder overview, this does not change.

gitjannes:
fixes #2400

- with updated send_reminders tests since their timing would have the described issue

### DIFF
--- a/evap/evaluation/tests/test_commands.py
+++ b/evap/evaluation/tests/test_commands.py
@@ -365,7 +365,7 @@ class TestSendRemindersCommand(TestCase):
         self.assertEqual(len(mail.outbox), 0)
 
     def test_dont_remind_evaluation_started_yesterday(self):
-    # May fail if run across a day boundary, candidate for freezegun
+        # May fail if run across a day boundary, candidate for freezegun
         user = baker.make(UserProfile)
         course = baker.make(Course)
         recent_evaluation = baker.make(


evap/evaluation/management/commands/send_reminders.py
```diff
@@ -52,7 +52,10 @@ def send_student_reminders():
         for evaluation in Evaluation.objects.filter(
             state=Evaluation.State.IN_EVALUATION, vote_end_date__in=check_dates
         ):
-            recipients.update(evaluation.due_participants)
+            if (
+                datetime.date.today() - datetime.timedelta(days=1) != evaluation.vote_start_datetime.date()
```
niklasmohrin: Two things:
1. I think we should not use a "the start is yesterday" logic, but instead a "this was created less than x days ago". The "yesterday" logic only makes sense because days are discrete, but I think it's more natural to use a less-than check for anything time related
2. We should be able to do this filtering in the database by modifying the arguments to `Evaluation.objects.filter` above. It should be something like `vote_start_datetime__date__ge=datetime.data.today() - datetime.timedelta(days=1)`. Possibly, we need Django database functions, which can be a bit wild when first using them, so please reach out if you need help :)

evap/evaluation/tests/test_commands.py
```diff
@@ -364,6 +364,31 @@ def test_dont_remind_already_voted(self):
         self.assertEqual(mock.call_count, 0)
         self.assertEqual(len(mail.outbox), 0)
 
+    def test_remind_only_once_evaluation_started_yesterday(self):
```
niklasmohrin: I think this test can be slightly nicer if structured like this:
```python
user = ...
recent_evaluation = baker.make(Evaluation, ...)
management.call_command(...)
assertNotCalled()

old_evaluation = baker.make(Evalaution, ...)
management.call_command(...)
assertCalled(..., due_evaluations=[recent_evaluation, old_evaluation])
```

With the current test, it is a bit hard to see that we are actually also testing that user1 does not receive an email, whereas with the above it is clear. Additionally, we also want to check that when an email is sent, it also includes recent evaluations

niklasmohrin (changes requested):

evap/evaluation/tests/test_commands.py
```diff
@@ -364,6 +364,36 @@ def test_dont_remind_already_voted(self):
...
+        recent_evaluation = baker.make(
+            Evaluation,
+            state=Evaluation.State.IN_EVALUATION,
+            vote_start_datetime=datetime.now() - timedelta(days=1),
+            vote_end_date=date.today() + timedelta(days=2),
+            participants=[user],
+        )
+
+        with patch("evap.evaluation.models.EmailTemplate.send_reminder_to_user") as mock:
+            management.call_command("send_reminders", stdout=StringIO())
```
richardebeling, line 383: This has the funny side effect that if I run this test at 23:59:59, and my clock happens to jump to 00:00:00 of the next day after the computation of `datetime.now()`, but before we check the time in `send_reminders`, this test will fail

I think we can make this unrealistically unlikely by using `vote_start_datetime=datetime.now() - timedelta(hours=22),` (not 23 so it also works during time zone changes) instead. Would still break when actively provoked, e.g. if the test is stopped in a debugger, but that's probably good enough for us for now.

evap/evaluation/management/commands/send_reminders.py
```diff
@@ -50,7 +50,10 @@ def send_student_reminders():
 
         recipients = set()
         for evaluation in Evaluation.objects.filter(
-            state=Evaluation.State.IN_EVALUATION, vote_end_date__in=check_dates
+            state=Evaluation.State.IN_EVALUATION,
+            vote_end_date__in=check_dates,
+            vote_start_datetime__date__lt=datetime.date.today()
+            - datetime.timedelta(days=1),  # only want evaluation which started before yesterday, see Issue#2400
```
richardebeling: I think this is a case of _very_ unfortunate auto formatting: The comment at the end forces a line break, and our formatter wants to keep some code on the same line as the comment, so it breaks before the `-` (and doesn't indent, because we're in a function call argument list)

I read this multiple times wrong, already, while just reviewing this PR. What do you think about this?
```suggestion
            # only want evaluation which started before yesterday, see Issue#2400
            vote_start_datetime__date__lt=datetime.date.today() - datetime.timedelta(days=1),  
```

evap/evaluation/tests/test_commands.py
niklasmohrin: @richardebeling Do you remember if we ever considered using something like https://github.com/spulec/freezegun ? Not necessarily needed for this PR, but maybe in a follow up?

evap/evaluation/management/commands/send_reminders.py
```diff
@@ -50,7 +50,10 @@ def send_student_reminders():
 
         recipients = set()
         for evaluation in Evaluation.objects.filter(
-            state=Evaluation.State.IN_EVALUATION, vote_end_date__in=check_dates
+            state=Evaluation.State.IN_EVALUATION,
+            vote_end_date__in=check_dates,
+            vote_start_datetime__date__lt=datetime.date.today()
```
niklasmohrin: In addition to what Richard said regarding time shenanigans, can you make sure that we call `date.today()` only once in the function? Otherwise we might get different dates back when sending reminders at 23:59:59. (we are not always super great at paying attention to this in existing code, but can't hurt to do it right)

evap/evaluation/tests/test_commands.py
janno42: I think we haven't because it wasn't necessary so far. I'd say having tests fail when running over midnight or daylight saving changes isn't a real trouble.

janno42 (approved):

evap/evaluation/tests/test_commands.py
richardebeling: For https://github.com/e-valuation/EvaP/issues/2046, we're also currently considering adding freezegun.

@gitjannes can you add a short comment here to the test, something like
> May fail if run across a day boundary, candidate for freezegun

or similar? Intention is to unblock the PR here while leaving some pointer in code so that we find this place again in the future

niklasmohrin (approved):
Except for [the comment suggestion](https://github.com/e-valuation/EvaP/pull/2471#discussion_r2487564259), everything looks good! Thanks

evap/evaluation/tests/test_commands.py
```diff
@@ -364,11 +364,17 @@ def test_dont_remind_already_voted(self):
         self.assertEqual(mock.call_count, 0)
         self.assertEqual(len(mail.outbox), 0)
 
-    def test_dont_remind_evaluation_started_yesterday(self):
+    def test_dont_remind_evaluation_started_yesterday(
+        self,
+    ):  # May fail if run across a day boundary, candidate for freezegun
```
niklasmohrin:
```suggestion
    def test_dont_remind_evaluation_started_yesterday(self):
    # May fail if run across a day boundary, candidate for freezegun
```

niklasmohrin (approved):

Issue #2352, opened 2024-12-09
Use CSS grids for all major tables


niklasmohrin:
Following #1818 and https://github.com/e-valuation/EvaP/pull/2349#issuecomment-2528789034, we want to use CSS grid instead of HTML `table`s for all big tables with complex navigation features. In #1889 we went with bootstrap grid in the end, but we probably want to try a more lightweight system with just CSS grid - something like the results page.

PR #2474, opened 2025-07-07
define non searchable datagrid columns 

### LINKED ISSUE #2461: Define searchable columns in datagrids
Currently, the search in a datagrid is performed on all columns, which can lead to unexpected results. For example, on the staff semester page, a search for "exam" shows all rows because they have a "Create _exam_ evaluation" button.
It should be possible to define which columns of a datagrid the search should be performed on.

jooooosef:
fix #2461 

TODO:
- [ ] add attribute to appropiate grids

### DIFF
--- a/evap/static/ts/src/datagrid.ts
+++ b/evap/static/ts/src/datagrid.ts
@@ -47,14 +47,14 @@ abstract class DataGrid {
             const column = header.dataset.col!;
             this.sortableHeaders.set(column, header);
         });
-        
+
         const headers = [...head.querySelectorAll<HTMLElement>("th")];
-        headers.forEach((header, index) => {            
+        headers.forEach((header, index) => {
             if (!header.hasAttribute("data-not-searchable")) {
                 this.searchableColumnIndices.push(index);
             }
         });
-        
+
         this.container = container;
         this.searchInput = searchInput;
         this.state = this.restoreStateFromStorage();
--- a/evap/static/ts/src/datagrid.ts
+++ b/evap/static/ts/src/datagrid.ts
@@ -31,7 +31,7 @@ interface DataGridParameters extends BaseParameters {
 abstract class DataGrid {
     private readonly storageKey: string;
     protected sortableHeaders: Map<string, HTMLElement>;
-    protected searchableColumns: string[];
+    protected searchableColumnIndices: number[];
     protected container: HTMLElement;
     private searchInput: HTMLInputElement;
     protected rows: Row[] = [];
@@ -41,16 +41,20 @@ abstract class DataGrid {
     protected constructor({ storageKey, head, container, searchInput }: DataGridParameters) {
         this.storageKey = storageKey;
         this.sortableHeaders = new Map();
-        this.searchableColumns = [];
-        head.querySelectorAll<HTMLElement>("[data-col]").forEach(header => {
+        this.searchableColumnIndices = [];
+
+        head.querySelectorAll<HTMLElement>(".col-order").forEach(header => {
             const column = header.dataset.col!;
-            if (header.classList.contains("col-order")) {
             this.sortableHeaders.set(column, header);
-            }
-            if (header.dataset.searchable !== "false") {
-                this.searchableColumns.push(column);
+        });
+        
+        const headers = [...head.querySelectorAll<HTMLElement>("th")];
+        headers.forEach((header, index) => {            
+            if (!header.hasAttribute("data-not-searchable")) {
+                this.searchableColumnIndices.push(index);
             }
         });
+        
         this.container = container;
         this.searchInput = searchInput;
         this.state = this.restoreStateFromStorage();
@@ -260,7 +264,8 @@ export class TableGrid extends DataGrid {
     }
 
     protected findSearchableCells(row: HTMLElement): HTMLElement[] {
-        return this.searchableColumns.map(column => row.querySelector(`[data-col=${column}]`)!);
+        const children = [...row.children] as HTMLElement[];
+        return this.searchableColumnIndices.map(index => children[index]).filter(cell => cell !== undefined);
     }
 
     protected fetchRowFilterValues(row: HTMLElement): Map<string, string[]> {
--- a/evap/static/ts/src/datagrid.ts
+++ b/evap/static/ts/src/datagrid.ts
@@ -31,6 +31,7 @@ interface DataGridParameters extends BaseParameters {
 abstract class DataGrid {
     private readonly storageKey: string;
     protected sortableHeaders: Map<string, HTMLElement>;
+    protected searchableColumns: string[];
     protected container: HTMLElement;
     private searchInput: HTMLInputElement;
     protected rows: Row[] = [];
@@ -40,9 +41,15 @@ abstract class DataGrid {
     protected constructor({ storageKey, head, container, searchInput }: DataGridParameters) {
         this.storageKey = storageKey;
         this.sortableHeaders = new Map();
-        head.querySelectorAll<HTMLElement>(".col-order").forEach(header => {
+        this.searchableColumns = [];
+        head.querySelectorAll<HTMLElement>("[data-col]").forEach(header => {
             const column = header.dataset.col!;
+            if (header.classList.contains("col-order")) {
             this.sortableHeaders.set(column, header);
+            }
+            if (header.dataset.searchable !== "false") {
+                this.searchableColumns.push(column);
+            }
         });
         this.container = container;
         this.searchInput = searchInput;
@@ -253,7 +260,7 @@ export class TableGrid extends DataGrid {
     }
 
     protected findSearchableCells(row: HTMLElement): HTMLElement[] {
-        return [...row.children] as HTMLElement[];
+        return this.searchableColumns.map(column => row.querySelector(`[data-col=${column}]`)!);
     }
 
     protected fetchRowFilterValues(row: HTMLElement): Map<string, string[]> {
--- a/evap/staff/templates/staff_semester_view.html
+++ b/evap/staff/templates/staff_semester_view.html
@@ -415,8 +415,8 @@
                                     <th class="col-order" data-col="state-results"></th>
                                     <th class="col-order" data-col="name">{% translate 'Name' %}</th>
                                     <th class="col-order" data-col="period">{% translate 'Evaluation period' %}</th>
-                                    <th></th>
-                                    <th></th>
+                                    <th data-searchable="false"></th>
+                                    <th data-searchable="false"></th>
                                 </tr>
                             </thead>
                             <tbody>
@@ -517,7 +517,7 @@
                                 <th class="col-order" data-col="name">{% translate 'Course' %}</th>
                                 <th class="col-order" data-col="responsible">{% translate 'Responsible' %}</th>
                                 <th class="col-order" data-col="evaluation-count">{% translate '#Evaluations' %}</th>
-                                <th></th>
+                                <th data-col="actions" data-searchable="false"></th>
                             </tr>
                         </thead>
                         <tbody>
--- a/evap/staff/templates/staff_semester_view.html
+++ b/evap/staff/templates/staff_semester_view.html
@@ -415,8 +415,8 @@
                                     <th class="col-order" data-col="state-results"></th>
                                     <th class="col-order" data-col="name">{% translate 'Name' %}</th>
                                     <th class="col-order" data-col="period">{% translate 'Evaluation period' %}</th>
-                                    <th data-searchable="false"></th>
-                                    <th data-searchable="false"></th>
+                                    <th data-not-searchable></th>
+                                    <th data-not-searchable></th>
                                 </tr>
                             </thead>
                             <tbody>
@@ -517,7 +517,7 @@
                                 <th class="col-order" data-col="name">{% translate 'Course' %}</th>
                                 <th class="col-order" data-col="responsible">{% translate 'Responsible' %}</th>
                                 <th class="col-order" data-col="evaluation-count">{% translate '#Evaluations' %}</th>
-                                <th data-col="actions" data-searchable="false"></th>
+                                <th data-not-searchable></th>
                             </tr>
                         </thead>
                         <tbody>


evap/static/ts/src/datagrid.ts
```diff
@@ -40,10 +41,20 @@ abstract class DataGrid {
...
+
         head.querySelectorAll<HTMLElement>(".col-order").forEach(header => {
             const column = header.dataset.col!;
             this.sortableHeaders.set(column, header);
         });
+
+        const headers = [...head.querySelectorAll<HTMLElement>("th")];
+        headers.forEach((header, index) => {
+            if (!header.hasAttribute("data-not-searchable")) {
+                this.searchableColumnIndices.push(index);
```
niklasmohrin: I think it would be better if we would mark the columns that we want to be searched instead of marking the ones that should not be searched. What do you think?

evap/static/ts/src/datagrid.ts
```diff
@@ -40,10 +41,20 @@ abstract class DataGrid {
     protected constructor({ storageKey, head, container, searchInput }: DataGridParameters) {
         this.storageKey = storageKey;
         this.sortableHeaders = new Map();
+        this.searchableColumnIndices = [];
```
niklasmohrin: This variable is declared in the abstract DataGrid, but only used in the TableGrid subclass - this means that this variable is unused in the ResultsGrid. I think I am fine with this only working with TableGrid for now, because the results page does not seem to need this (@janno42 ?). However, then we should move this to the TableGrid constructor

evap/static/ts/src/datagrid.ts
```diff
@@ -253,7 +264,8 @@ export class TableGrid extends DataGrid {
     }
 
     protected findSearchableCells(row: HTMLElement): HTMLElement[] {
-        return [...row.children] as HTMLElement[];
+        const children = [...row.children] as HTMLElement[];
+        return this.searchableColumnIndices.map(index => children[index]).filter(cell => cell !== undefined);
```
niklasmohrin: To remove any type casts, you can do
```suggestion
        return this.searchableColumnIndices.map(index => {
            const child = row.children[index];
            assert(child instanceof HTMLElement);
            return child;
        });
```
after importing assert from our utils.

evap/static/ts/src/datagrid.ts
niklasmohrin, line 1: Can you try to write some tests for this with our frontend test setup? I think this should work without much trouble - if not, we should make our testing easier to use

niklasmohrin (commented):
Implementation generally makes sense and seems to accomplish this with quite little changes - nice work!

evap/static/ts/src/datagrid.ts
jooooosef: This would mean changing every single table, so I thought specifically declaring which one should not be searched (also for future grids) would be easer to implement (via `data-not-searchable`). 
I think in most use cases we want most of the columns to be searchable?

evap/static/ts/src/datagrid.ts
jooooosef: Are we talking about the Jest frontend tests? Like in `static/ts/tests`?

evap/static/ts/src/datagrid.ts
richardebeling: I think the opt-out solution we have currently makes sense for us, "included" is a sane default

evap/static/ts/src/datagrid.ts
richardebeling: I think Niklas is referring to the tests in [evap/staff/tests/test_live.py](https://github.com/e-valuation/EvaP/blob/main/evap/staff/tests/test_live.py). As Niklas mentioned above, this shouldn't be too much work/code, but the infrastructure is fairly new -- talk to us if anything doesn't work as expected.

evap/static/ts/src/datagrid.ts
niklasmohrin: okay sure

evap/staff/tests/test_live.py
```diff
@@ -71,3 +75,45 @@ def test_submit_changes_form_data(self):
...
+        search_input.clear()
+        search_input.send_keys("exam")
+
+        self.wait.until(invisibility_of_element_located((By.XPATH, "//td[contains(text(),'course name')]")))
+
+        evaluation_table = self.selenium.find_element(By.ID, "evaluation-table")
+        tds = evaluation_table.find_elements(By.TAG_NAME, "td")
+        self.assertFalse(any("course name" in td.text for td in tds))
+
+    #     question1 = baker.make(Question, type=Question.Type.TEXT)
```
richardebeling: commented code

evap/staff/tests/test_live.py
```diff
@@ -71,3 +75,32 @@ def test_submit_changes_form_data(self):
         self.assertEqual(contribution1.order, 0)
         self.assertEqual(contribution1.role, Contribution.Role.EDITOR)
         self.assertEqual(contribution1.textanswer_visibility, Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS)
+
+    def test_staff_semester_view_columns_not_searchable(self):
+        """Regression test for #2461"""
+
+        semester = baker.make("Semester")
```
```suggestion
        semester = baker.make(Semester)
```

evap/staff/tests/test_live.py
```diff
@@ -71,3 +75,32 @@ def test_submit_changes_form_data(self):
...
+        course = baker.make(Course, semester=semester, name_en="course name")
+        baker.make(Evaluation, course=course, name_en="evaluation name")
+
+        with self.enter_staff_mode():
+            self.selenium.get(self.live_server_url + reverse("staff:semester_view", args=[semester.pk]))
```
niklasmohrin: not for this PR: we should come up with a way to not always write `self.live_server_url + `. @hansegucker do you know if there is an easy way?

evap/staff/tests/test_live.py
```diff
@@ -71,3 +75,32 @@ def test_submit_changes_form_data(self):
...
+        self.assertTrue(any("course name" in td.text for td in tds))
+
+        search_input.clear()
+        search_input.send_keys("exam")
+
+        self.wait.until(invisibility_of_element_located((By.XPATH, "//td[contains(text(),'course name')]")))
+
+        evaluation_table = self.selenium.find_element(By.ID, "evaluation-table")
+        tds = evaluation_table.find_elements(By.TAG_NAME, "td")
+        self.assertFalse(any("course name" in td.text for td in tds))
```
niklasmohrin: Can you explain why we do the iteration over all td elements in addition to the waiting for invisibility? Is there a case where the latter could still fail or is where it is better?

Additionally, why to we not wait for visibility before the `assertTrue` before?

evap/static/ts/src/datagrid.ts
```diff
@@ -40,10 +41,12 @@ abstract class DataGrid {
     protected constructor({ storageKey, head, container, searchInput }: DataGridParameters) {
         this.storageKey = storageKey;
         this.sortableHeaders = new Map();
+
```
niklasmohrin: please revert these unrelated formatting changes

evap/static/ts/src/datagrid.ts
```diff
@@ -240,6 +244,15 @@ export class TableGrid extends DataGrid {
...
         this.resetSearch = resetSearch;
+        this.searchableColumnIndices = [];
+
+        const thead = table.querySelector("thead")!;
+        const headers = [...thead.querySelectorAll<HTMLElement>("th")];
+        headers.forEach((header, index) => {
+            if (!header.hasAttribute("data-not-searchable")) {
+                this.searchableColumnIndices.push(index);
+            }
+        });
```
niklasmohrin, line 252: You can slightly simplify this and remove the type cast:
```suggestion

        this.searchableColumnIndices = [];
        const thead = selectOrError("thead", table);
        thead.querySelectorAll("th").forEach((header, index) => {
            if (!header.hasAttribute("data-not-searchable")) {
                this.searchableColumnIndices.push(index);
            }
        });
```
I was playing around with https://developer.mozilla.org/en-US/docs/Web/API/HTMLTableCellElement/cellIndex and related properties, but I think in light of #2352 it's not worth it to spend time on this. Assuming we return spec compliant HTML (which I am 95% is the case for tables) the enumerating here should always be equivalent anyways

evap/grades/templates/grades_semester_view.html
niklasmohrin, line 1: (not related to this file)

What about the user list? If I enter for example "contributor", I get all users with the badge and not just the user `contributor@institution.example.com` user. Is this expected @janno42 ?

niklasmohrin, line 1: questionnaire table is also missing, I can still search for "delete" and get matches based on the modal buttons

evap/staff/tests/test_live.py
jooooosef: see #2488

evap/staff/tests/test_live.py
jooooosef:
```python
evaluation_table = self.wait.until(visibility_of_element_located((By.ID, "evaluation-table")))
tds = evaluation_table.find_elements(By.TAG_NAME, "td")
self.assertTrue(any("course name" in td.text for td in tds))
```

We're waiting for the visibility before the `assertTrue`, right? Or did you mean `//td[contains(text(),'course name')]`?
(Read this in an Italian accent): Then the assert would be kind of unnecessary, no? 🤌
But the same applies to the `assertFalse` here, so yeah, it's kind of unnecessary — but I couldn’t really come up with a way to reliably wait for the table to refresh without explicitly checking if the course is gone.

Should I just remove it then?
if a test takes too long it times out and fails, so I think just testing visibility and invisiblity should be fine

niklasmohrin: Yes I meant visibility of the cell. I am just wondering why there are seemingly two different checks for the same thing. Waiting is fine, we need to do it. I don't know if the other check is needed, that's what I am asking :D

If the waiting suffices and the other check is not needed then throw it out; if there is any sensible benefit to also doing the second check, please add a comment and make sure that both the before and after check use both visibility and the assert(any(...)).

jooooosef: 🫡

evap/grades/templates/grades_semester_view.html
janno42: Yes, that's expected. The badges should be part of the search.

jooooosef:
On another note:
I think we (me) should add a comment somewhere to explain that is "not searchable" with the attribute since its not intuitive.
The problem is that stuff that is considered "textContent" from JS is part of the searchable text.
So if this is not the case (e.g. the confirmiation model is not directly defined in the grid) the text is also not searchable (e.g. "delete" needs to be in the text of the confirmation model).
idk if this comment is useful

evap/staff/tests/test_live.py
```diff
@@ -71,3 +84,28 @@ def test_submit_changes_form_data(self):
...
+            self.selenium.get(self.live_server_url + reverse("staff:semester_view", args=[semester.pk]))
+
+        search_input = self.wait.until(
+            visibility_of_element_located((By.CSS_SELECTOR, "input[type='search'][name='search-evaluation']"))
+        )
+        search_input.clear()
+        search_input.send_keys("course name")
+
+        self.wait.until(visibility_of_element_located((By.ID, "evaluation-table")))
+        self.wait.until(visibility_of_element_located((By.XPATH, "//td//a[contains(text(),'course name')]")))
```
richardebeling: The test is really cool!

Can we maybe add an assertion to this first block that one of the visible table cells contains the word "exam", so that if we change the wording in the future, the test fails (instead of just always passing because the invisibility assertion below trivially always holds)?

richardebeling (approved):
Good job, code looks generally good to me. I have one potential improvement of the test, good to merge otherwise from my side

evap/static/ts/src/datagrid.ts
janno42: The results page does not need it (for now).

evap/staff/tests/test_live.py
```diff
@@ -19,7 +32,6 @@ def test_submit_changes_form_data(self):
             course=baker.make(Course, programs=[baker.make(Program)], responsibles=[responsible]),
             vote_start_datetime=datetime(2099, 1, 1, 0, 0),
             vote_end_date=date(2099, 12, 31),
-            main_language="en",
```
niklasmohrin, line 21: please leave this unchanged :)

evap/static/ts/src/datagrid.ts
```diff
@@ -44,6 +45,7 @@ abstract class DataGrid {
             const column = header.dataset.col!;
             this.sortableHeaders.set(column, header);
         });
+
```
niklasmohrin:
```suggestion
```

evap/static/ts/src/datagrid.ts
```diff
@@ -240,6 +243,14 @@ export class TableGrid extends DataGrid {
             ...options,
         });
         this.resetSearch = resetSearch;
+        this.searchableColumnIndices = [];
+
+        const thead = selectOrError("thead", table);
```
niklasmohrin: we already select this above, please change the select above to use a variable

evap/staff/tests/test_live.py
```diff
@@ -72,3 +84,33 @@ def test_submit_changes_form_data(self):
         self.assertEqual(contribution1.order, 0)
         self.assertEqual(contribution1.role, Contribution.Role.EDITOR)
         self.assertEqual(contribution1.textanswer_visibility, Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS)
+
+    def test_staff_semester_view_columns_not_searchable(self):
```
niklasmohrin, line 88: very good!

evap/staff/tests/test_live.py
richardebeling: Whoops, your moved-out issue was fixed faster than your original PR was merged. Guess this means you now have to change your usages :P

https://github.com/e-valuation/EvaP/pull/2512/files

```python3
self.selenium.get(self.reverse("staff:semester_view", args=[semester.pk]))
```

evap/staff/tests/test_live.py
```diff
@@ -72,3 +84,33 @@ def test_submit_changes_form_data(self):
...
+        search_input.clear()
+        search_input.send_keys("course name")
+
+        self.wait.until(visibility_of_element_located((By.ID, "evaluation-table")))
+        self.wait.until(visibility_of_element_located((By.XPATH, "//td//a[contains(text(),'course name')]")))
+        self.wait.until(
+            visibility_of_element_located(
+                (By.XPATH, "//button[@slot='show-button' and @aria-label='Create exam evaluation']")
+            )
+        )
```
richardebeling: Do we need all three, or would one of them suffice?

evap/staff/tests/test_live.py
```diff
@@ -72,3 +84,33 @@ def test_submit_changes_form_data(self):
...
+
+        search_input.clear()
+        search_input.send_keys("exam")
+
+        self.wait.until(visibility_of_element_located((By.ID, "evaluation-table")))
```
richardebeling: what does this do? (given that we waited for the evaluation-table to be visible already before the `send_keys`. Does it disappear in between? I wouldn't expect it to)

evap/staff/tests/test_live.py
jooooosef: uiuiuiui

evap/staff/tests/test_live.py
jooooosef: One is fine as far as I know. 
This was when I was writing the test, I had my problems properly "finding" the elements :)

evap/staff/tests/test_live.py
jooooosef: When searching for exam, this line should not be shown, since there is no "exam" in the title. This directly tests the original issue being fixed.

EDIT: sorry, yes the waiting for "evaluation-table" is unnessary, same reason as before (see above).

PR #2480, opened 2025-07-21
Require just one default name for exam evaluation

### LINKED ISSUE #2475: Creating second exam evaluation fails when non-default name was used before
If an exam evaluation exists where either `name_de` is not "Klausur" or `name_en` is not "Exam" (e.g., `name_de` might be "Prüfung"), then `has_exam_evaluation` for the main evaluation is `False`. Thus, the button to create an exam evaluation is displayed on the staff semester page, but creating it fails due to a unique constraint.

Instead, `has_exam_evaluation` should return `True` also if the default exam name exists in only one of the languages.

ronjakrg:
Fixes #2475

### DIFF
--- a/evap/staff/tests/test_views.py
+++ b/evap/staff/tests/test_views.py
@@ -2079,6 +2079,19 @@ class TestEvaluationExamCreation(WebTestStaffMode):
         with assert_no_database_modifications():
             self.app.post(self.url, user=self.manager, status=400, params=self.params)
 
+    def test_exam_evaluation_for_already_existing_exam_evaluation_without_default_en_name(self):
+        baker.make(Evaluation, course=self.course, name_en="Test", name_de="Klausur")
+        self.assertTrue(self.evaluation.has_exam_evaluation)
+        with assert_no_database_modifications():
+            self.app.post(self.url, user=self.manager, status=400, params=self.params)
+        baker.make(Evaluation, course=self.course, name_en="Exam", name_de="Prüfung")
+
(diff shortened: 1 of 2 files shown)


evap/staff/tests/test_views.py
```diff
@@ -2079,6 +2079,19 @@ def test_exam_evaluation_for_already_existing_exam_evaluation(self):
         with assert_no_database_modifications():
             self.app.post(self.url, user=self.manager, status=400, params=self.params)
 
+    def test_exam_evaluation_for_already_existing_exam_evaluation_without_default_en_name(self):
+        baker.make(Evaluation, course=self.course, name_en="Test", name_de="Klausur")
+        self.assertTrue(self.evaluation.has_exam_evaluation)
+        with assert_no_database_modifications():
+            self.app.post(self.url, user=self.manager, status=400, params=self.params)
+        baker.make(Evaluation, course=self.course, name_en="Exam", name_de="Prüfung")
```
niklasmohrin: This line is left over here
```suggestion
```

evap/staff/tests/test_views.py
```diff
@@ -2079,6 +2079,19 @@ def test_exam_evaluation_for_already_existing_exam_evaluation(self):
...
+
+    def test_exam_evaluation_for_already_existing_exam_evaluation_without_default_de_name_(self):
```
niklasmohrin:
```suggestion
    def test_exam_evaluation_for_already_existing_exam_evaluation_without_default_de_name(self):
```

richardebeling (approved):

janno42 (approved):
🪪

Issue #2185, opened 2024-05-13
Database validation for model field choices


richardebeling:
See discussion in https://github.com/e-valuation/EvaP/pull/1776. Django doesn't actually validate on the database level that only choice values are stored. This lead to us not noticing an outdated test data dump in #1776.

Optimally, we'd like to have database constraints for all model fields with choices, and we would like to not repeat the choices. See the discussion at #1776 for a suggested approach using a decorator that seems reasonable to me at first glance.

PR #2481, opened 2025-07-21
Database constraints for model fields with choices

richardebeling:
fixes https://github.com/e-valuation/EvaP/issues/2185

### DIFF
--- a/evap/evaluation/tools.py
+++ b/evap/evaluation/tools.py
@@ -24,8 +24,8 @@ from django.views.generic import FormView
 from evap.tools import date_to_datetime
 
 if TYPE_CHECKING:
-    from django_stubs_ext import StrOrPromise  # use proper definition with mypy
     from django.db.models.fields import _ChoicesList
+    from django_stubs_ext import StrOrPromise  # use proper definition with mypy
 else:
     try:
         from django_stubs_ext import StrOrPromise  # if installed, use proper definition for typeguard


evap/evaluation/models_logging.py
```diff
@@ -103,6 +103,7 @@ class LogEntry(models.Model):
     request_id = models.CharField(max_length=36, blank=True)
     data = models.JSONField(default=dict, encoder=LogJSONEncoder)
 
+    @inject_choices_constraint("logentry", locals())
```
niklasmohrin: can't we use [`__qualname__`](https://peps.python.org/pep-3155/) from the locals instead of passing the name here?

evap/evaluation/tools.py
```diff
@@ -35,6 +37,54 @@
 CV = TypeVar("CV", bound=CellValue)
 
 
+def choice_database_values_from_django_choices_spec(django_choices_spec) -> list:
+    assert isinstance(django_choices_spec, list)
+    assert all(isinstance(element, tuple) for element in django_choices_spec)
```
richardebeling: type annotations vs assertions is up for discussion here, no strong opinion on my side, the django requirements are not so easy to use with static typing, but annotating as `list[tuple]` could probably do the trick. It may be or become wrong, but then our runtime typing check should hopefully complain.

evap/evaluation/migrations/0151_choices_constraints.py
```diff
@@ -0,0 +1,91 @@
...
+                condition=models.Q(
+                    (
+                        "action_type__in",
+                        [
+                            evap.evaluation.models_logging.InstanceActionType["CREATE"],
+                            evap.evaluation.models_logging.InstanceActionType["CHANGE"],
+                            evap.evaluation.models_logging.InstanceActionType["DELETE"],
+                        ],
+                    )
+                ),
```
niklasmohrin: The following does not make `makemigrations` complain and seems better to me. Any opinions?
```suggestion
                condition=models.Q(("action_type__in", ["create", "change", "delete"])),
```

evap/evaluation/tools.py
```diff
@@ -35,6 +37,54 @@
...
+
+
+def inject_choices_constraint(model_locals):
+    model_name = model_locals["__qualname__"]
+    assert re.match("[a-zA-Z]+", model_name), "deduced model name doesn't look like a pure class name"
+
+    def decorator(meta_cls):
+        """
+        This decorator is meant to decorate Meta classes within Django Model classes
+        It injects a constraints for each model field that has choices, enforcing that only
```
niklasmohrin:
```suggestion
        It injects a constraint for each model field that has choices, enforcing that only
```

evap/evaluation/tools.py
niklasmohrin: I would add the annotations as much as possible. [Django-stubs has `_ChoicesList`](https://github.com/typeddjango/django-stubs/blob/ba4a1fa0f0581a1c945ffecd1f50c1c26b93b361/django-stubs/db/models/fields/__init__.pyi#L139), should we use that? We will then actually still have to use the assertion to check that we are in the correct variant of the union, but I think that's a good thing

For the rest, I think the best we can do is
```python
def inject_choices_constraint(model_locals: Mapping[str, Any]) -> Callable[[type], type]:
    # ...
    def decorator(meta_cls: type) -> type:
```

There is [TypedModelMeta in django-stubs-ext](https://github.com/typeddjango/django-stubs/blob/ba4a1fa0f0581a1c945ffecd1f50c1c26b93b361/ext/django_stubs_ext/db/models/__init__.py#L12), which might also work. Then all `type`s become `type[TypedModelMeta]`

niklasmohrin: Did ChoicesList and TypedModelMeta not work / what came of it?

niklasmohrin (dismissed):
Wow looks great!

PR #2484, opened 2025-07-27
Remove single results

janno42:
We didn't use single results for several years now. So let's remove lots of special handling!

This migrates all single results into "normal" evaluations with an overall rating questions holding the existing answers and removes all single results logic.


richardebeling (dismissed):

evap/evaluation/migrations/0150_migrate_single_results_to_normal_evaluations.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('evaluation', '0149_evaluation_dropout_count_alter_questionnaire_type'),
+    ]
+
+    operations = [
+        migrations.RunPython(migrate_single_results)
```
niklasmohrin: To make the migration reversible:
```suggestion
        migrations.RunPython(migrate_single_results, reverse_code=migrations.RunPython.noop)
```

evap/evaluation/migrations/0150_migrate_single_results_to_normal_evaluations.py
```diff
@@ -0,0 +1,58 @@
...
+    Questionnaire = apps.get_model("evaluation", "Questionnaire")
+    RatingAnswerCounter = apps.get_model("evaluation", "RatingAnswerCounter")
+    TextAnswer = apps.get_model("evaluation", "TextAnswer")
+
+    # Don't create a questionnaire when running the migration on a database without single results
+    # E.g. when running on an empty database before loading test data
+    if not Evaluation.objects.filter(is_single_result=True).exists():
+        return
+
+    single_result_questionnaire = Questionnaire.objects.get(name_en='Single result')
```
niklasmohrin: Can we add the following too?
```suggestion
    single_result_questionnaire = Questionnaire.objects.get(name_en='Single result')
    assert single_result_questionnaire.questions.count() == 1
    assert single_result_questionnaire.question.type == 2
```

evap/evaluation/migrations/0150_migrate_single_results_to_normal_evaluations.py
```diff
@@ -0,0 +1,58 @@
...
+
+    overall_rating_questionnaire, created = Questionnaire.objects.get_or_create(name_en="Total grade", type=30)  # BOTTOM questionnaire
+    if created:
+        Question.objects.create(
+            questionnaire = overall_rating_questionnaire,
+            type=2,
```
niklasmohrin: nit: can you also add a comment for which type this is like you did with BOTTOM?

evap/evaluation/migrations/0150_migrate_single_results_to_normal_evaluations.py
niklasmohrin, line 1: needs to be rebased before merging

evap/evaluation/migrations/0150_migrate_single_results_to_normal_evaluations.py
janno42: ah right. at first i thought it shouldn't be reversible because the single results can't be converted back but we don't need to have the reversion to restore data, that's right.

evap/evaluation/migrations/0151_migrate_single_results_to_normal_evaluations.py
```diff
@@ -0,0 +1,60 @@
...
+        return
+
+    single_result_questionnaire = Questionnaire.objects.get(name_en="Single result")
+    assert single_result_questionnaire.questions.count() == 1
+    assert single_result_questionnaire.questions.first().type == 2  # GRADE question
+
+    overall_rating_questionnaire, created = Questionnaire.objects.get_or_create(
+        name_en="Total grade",
+        type=30,  # BOTTOM questionnaire
+    )
```
niklasmohrin: We probably also want this, right?
```suggestion
    overall_rating_questionnaire, created = Questionnaire.objects.get_or_create(
        name_en="Total grade",
        type=30,  # BOTTOM questionnaire
        defaults={"name_de": "Gesamtnote"},
    )
```

niklasmohrin (approved):
No more hot single results in this area :sunglasses:

PR #2486, opened 2025-07-28
Stop excluding migrations from formatters

niklasmohrin:
Just saw that #2484 got red CI from codacy because of formatting. I don't want to think about such things. [The original formatting PR](https://github.com/e-valuation/EvaP/pull/1585) did not have much motivation why they were excluded except to avoid the churn, but I don't think it's a problem today because nobody is touching these files, and I would much rather get rid of the exception.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -6,3 +6,5 @@
 02ea57d2c33eb2801414982662acd8582e361d10
 # enforce pylint's `use-dict-literal`
 0d3fed856961fd453555637ada50080901160e9a
+# run formatters on migration files
+2cf3e4308ff4165e6c4c5983cb76d1bfcc685651


hansegucker (dismissed):

richardebeling (approved):

evap/evaluation/migrations/0077_course_is_single_result.py
```diff
@@ -6,27 +6,26 @@
 
 
 def populate_is_single_result(apps, _schema_editor):
-    Course = apps.get_model('evaluation', 'Course')
+    Course = apps.get_model("evaluation", "Course")
     for course in Course.objects.all():
-        if course.contributions.filter(responsible=True, questionnaires__name_en=evap_models.Questionnaire.SINGLE_RESULT_QUESTIONNAIRE_NAME).exists():
+        if course.contributions.filter(
+            responsible=True, questionnaires__name_en=evap_models.Questionnaire.SINGLE_RESULT_QUESTIONNAIRE_NAME
```
niklasmohrin, line 12: huh, funnily enough this didn't break in #2484 :thinking:

evap/evaluation/migrations/0117_logentry.py
```diff
@@ -1,40 +1,79 @@
 # Generated by Django 3.0.7 on 2020-07-28 12:06
 
+import django.contrib.postgres.fields.jsonb
+import django.db.models.deletion
 from django.conf import settings
 from django.db import migrations, models
-import django.db.models.deletion
-import django.contrib.postgres.fields.jsonb
+
 import evap.evaluation.models_logging
```
niklasmohrin, line 8: gotta fix that

evap/evaluation/migrations/0124_add_html_content.py
```diff
@@ -1,30 +1,33 @@
...
+            new_name="plain_content",
         ),
         migrations.AddField(
-            model_name='emailtemplate',
-            name='html_content',
-            field=models.TextField(default='', validators=[evap.evaluation.models.validate_template], verbose_name='HTML'),
+            model_name="emailtemplate",
+            name="html_content",
+            field=models.TextField(
+                default="", validators=[evap.evaluation.models.validate_template], verbose_name="HTML"
```
niklasmohrin, line 29: not sure about this one

evap/evaluation/migrations/0133_add_infotext_model.py
```diff
@@ -1,4 +1,5 @@
 from django.db import migrations, models
+
 from evap.evaluation.models import NotHalfEmptyConstraint
```
niklasmohrin, line 3: hm

evap/grades/migrations/0003_add_upload_path_and_change_last_modified_user_related_name.py
```diff
@@ -1,24 +1,31 @@
...
     ]
 
     operations = [
         migrations.AlterField(
-            model_name='gradedocument',
-            name='file',
-            field=models.FileField(verbose_name='File', upload_to=evap.grades.models.helper_upload_path),
+            model_name="gradedocument",
+            name="file",
+            field=models.FileField(verbose_name="File", upload_to=evap.grades.models.helper_upload_path),
```
niklasmohrin, line 18: :eyes:

niklasmohrin (commented):
i gotta sit down and check some of these possible bugs, it probably makes sense to ban importing evap in migrations with ruff and enforce that any import must have an ignore?

evap/evaluation/migrations/0077_course_is_single_result.py
janno42: why should it?

niklasmohrin: Because the old migration is not runnable anymore with the new code - Questionnaire.SINGLE_RESULT_QUESTIONNAIRE_NAME is not an attribute anymore. To reproduce, checkout #2484 and add a `baker.make(Course)` before the `for course` loop in the 0077 migration, then the error is raised.

So it didn't break because our test data contains no courses during the initial migration.

PR #2490, opened 2025-07-30
UI changes

janno42:
A number of UI changes - see commit messages


richardebeling (dismissed):

evap/evaluation/templates/base.html
```diff
@@ -47,12 +47,12 @@
...
 
-        {% if user.is_authenticated %}
+        {% if user.is_authenticated and not plain_page %}
             {% include 'notebook.html' %}
         {% endif %}
 
-        <div class="p-3 mb-4" id="evapContent">
-            <div class="container">
+        <div{% if not plain_page %} class="p-3 mb-4"{% endif %} id="evapContent">
+            <div class="{% if plain_page %}container-fluid d-flex max-content-height{% else %}container{% endif %}">
```
niklasmohrin, line 55: I am not a huuuge fan of having these conditional styles via Django and I think it would be nicer if we would have something like `extra_body_classes` instead of `plain_page` - then we could handle these styling things with CSS.

niklasmohrin (dismissed):

evap/evaluation/templates/base.html
niklasmohrin: So the evaluation form view would set something like `"extra_body_classes": "no-footer no-notebook evap-content-max"` or so

niklasmohrin: Actually, instead of specifying this in the view, it should be done in a the template using a template block (or, if it's possible `extends base.html with... `). That makes more sense

richardebeling: I'd say, as long as we have exactly these two options (full page, plain page), we don't have to build further modularization logic for now

evap/staff/templates/staff_evaluation_form.html
```diff
@@ -4,63 +4,67 @@
...
-            {% endif %}
-            {% if state == evaluation.State.IN_EVALUATION %}
-                <div class="alert alert-warning">{% translate 'You are editing an evaluation, which is already running. Please note that only the participants who did not evaluate yet will see your changes.' %}</div>
-            {% endif %}
-            {% if state == evaluation.State.EVALUATED or state == evaluation.State.REVIEWED %}
-                <div class="alert alert-warning">{% translate 'You are editing an evaluation, for which the evaluation already finished. Please be careful to not destroy any results.' %}</div>
-            {% endif %}
-            {% if state == evaluation.State.NEW or state == evaluation.State.PREPARED %}
-                <div class="alert alert-warning">{% translate 'You are editing an evaluation which has not been approved by the editor yet.' %}</div>
+    <div class="row flex-grow-1 overflow-hidden">
```
niklasmohrin, line 7: by the way how do you feel about adding `mt-2` here? otherwise it feels very tight

evap/evaluation/templates/base.html
```diff
@@ -65,7 +65,9 @@
             </div>
         </div>
 
-        {% include 'footer.html' %}
+        {% if not plain_page %}
+            {% include 'footer.html' %}
+        {% endif %}
```
niklasmohrin, line 70: Is there a reason not to keep the footer, except that it makes the CSS more annoying? Because I think then we should try to figure out how to make it work instead

evap/evaluation/templates/base.html
niklasmohrin: hm, I don't know, I just think one view alone does not warrant such specific changes in base.html for it. I might be wrong here, but I don't see that plain_page will become useful for more than this view as max-content-height is really specific to the two column independent scroll and doesn't add anything to normal layouts (again, correct me if I'm wrong). I think I wouldn't let it pass if a Bachelor's student proposed this without more reasons why we shouldn't try harder

If you two think it's fine I don't want to block the PR, after all the site works as intended

evap/evaluation/templates/base.html
janno42: I wanted to remove everything unnecessary to create as much space as possible for the form.

evap/evaluation/templates/base.html
janno42: In favor of refactoring once we add more options or use it in multiple different ways. However, if we did that, we would definitely have to change it. I like the idea of keeping everything in the respective template.

evap/staff/templates/staff_evaluation_form.html
janno42: Yes, it's much tighter than other pages, but at the end that's the overall idea.
I'd say the padding on the heading is sufficient :)

PR #2496, opened 2025-08-04
change branding of HPI-Schools

funnyanni-pdm:
decided to keep the explicit naming of d-school and Engine, because people won't get it otherwise

### DIFF
--- a/evap/evaluation/templates/index.html
+++ b/evap/evaluation/templates/index.html
@@ -43,7 +43,7 @@
                     {% csrf_token %}
                     <input type="hidden" name="submit_type" value="new_key" />
                     <h4 class="card-title text-dark">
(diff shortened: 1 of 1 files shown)


evap/evaluation/templates/index.html
```diff
@@ -43,7 +43,7 @@ <h4 class="card-title text-primary">
                     {% csrf_token %}
                     <input type="hidden" name="submit_type" value="new_key" />
                     <h4 class="card-title text-dark">
-                        {% translate 'D-School, E-School and externals' %}
+                        {% translate 'Engine, d-school and externals' %}
```
janno42:
```suggestion
                        {% translate 'Login for external users' %}
```

PR #2497, opened 2025-08-04
Fix Text Answer review redirect

### LINKED ISSUE #1696: Text answer review redirects
When editing a text answer and saving, the page always redirects to the first unreviewed text answer in quick view, regardless of the previous view and which text answer was edited.
Instead, the review page should be shown in the view used before and when in quick view, the text answer just edited should be shown.

_Originally posted by @niklasmohrin in https://github.com/e-valuation/EvaP/issues/1689#issuecomment-1006927837 and @fidoriel in https://github.com/e-valuation/EvaP/issues/1689#issuecomment-1006935810_

geromequa:
closes #1696

### DIFF
--- a/evap/staff/templates/staff_evaluation_textanswers_quick.html
+++ b/evap/staff/templates/staff_evaluation_textanswers_quick.html
@@ -88,6 +88,7 @@
                                 <div
                                     class="slider-item card-body"
                                     data-layer="2"
+                                    id="textanswer-{{ answer.id }}"
                                     data-id="{{ answer.id }}"
                                     {% if contributor %} data-contribution="yes"{% endif %}
                                     {% if answer.is_reviewed %}
--- a/evap/staff/views.py
+++ b/evap/staff/views.py
@@ -1832,7 +1832,9 @@ def evaluation_textanswer_edit(request, textanswer_id):
     if form.is_valid():
         form.save()
         # jump to edited answer
-        url = reverse("staff:evaluation_textanswers", args=[evaluation.pk], fragment=str(textanswer.id))
+        url = reverse(
+            "staff:evaluation_textanswers", args=[evaluation.pk], fragment=("textanswer-" + str(textanswer.id))
+        )
         return HttpResponseRedirect(url)
 
     template_data = {


richardebeling (dismissed):

niklasmohrin (changes requested):
This doesn't work yet. Firstly, when coming from the full view and editing an answer, I still get redirected to the quick view afterwards. Secondly, the question selection in the quick view also still shows the first unreviewed question, not the one from the fragment. Is it just broken on my computer?

Here is what I think we need to do: In `quick-review-slider.ts` we are calling `this.startOver(StartOverWhere.Undecided);`. I believe that we should move this out of `QuickReviewSlider.attach` and into the calling code (`staff_evaluation_textanswers_quick.html`) where we can then check if the page URL has a fragment for a specific textanswer - in that case we `startOver` with that. Otherwise, we do what we did before.

For the redirect into the correct view, I think we have to include the view we want to return to in the request to `evaluation_textanswer_edit`. We have this pattern already in the index view after logging in (check the end of the `index` function in evap/evaluation/views.py). Although here we don't need the full url, so maybe `?next-view=full` or so would make for shorter / nicer urls.

geromequa:
The redirect to the proper view (quick, full etc.) should work now. The view is passed to the edit view via url parameter inside `staff_evaluation_textanswers_section.html` and then used to reverse to the correct view afterwards. Because the quick view does not render `staff_evaluation_textanswers_section.html` inside its own template, the redirect to the quick view is based on it being the fallback case (see views.py, line 1831).

evap/staff/templates/staff_evaluation_textanswers_quick.html
```diff
@@ -88,6 +88,7 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
                                 <div
                                     class="slider-item card-body"
                                     data-layer="2"
+                                    id="textanswer-{{ answer.id }}"
```
niklasmohrin, line 91: What about this question from my last review?

> Secondly, the question selection in the quick view also still shows the first unreviewed question, not the one from the fragment. Is it just broken on my computer?

I still don't think it works as intended from the issue:

> [...] the review page should be shown in the view used before **and when in quick view, the text answer just edited should be shown.**

Although the id is being set here, we are not doing anything with it in the frontend. Note that it seems the answers are shown in a somewhat random order as they are returned from the database, so the feature could work sometimes on accident. I outlined what I think needs to happen in my previous review :)

evap/staff/views.py
```diff
@@ -1828,11 +1828,16 @@ def evaluation_textanswer_edit(request, textanswer_id):
     assert_textanswer_review_permissions(evaluation)
 
     form = TextAnswerForm(request.POST or None, instance=textanswer)
-
+    view = request.GET.get("view", "quick")
```
niklasmohrin: I think it would be better if we did not duplicate the knowledge of what the default view is here; instead, we can just omit adding a `view=...` query string in the redirect here when we were not called with one ourselves. Additionally, I think it would slightly nicer if this would be called "next-view" instead of just "view", because we are not actually doing anything with it here but only after we are done.

niklasmohrin (changes requested):

janno42:
@geromequa Do you want to continue working on this or should we hand it over to someone else?

evap/static/ts/src/quick-review-slider.ts
```diff
@@ -377,15 +377,18 @@ export class QuickReviewSlider {
...
         );
         this.answerSlides = Array.from(decided).concat(Array.from(undecided));
 
-        const startOverOnUndecided = where === StartOverWhere.Undecided && undecided.length > 0;
-        const startIndex = startOverOnUndecided ? decided.length : 0;
+        let startIndex;
+        if (where === StartOverWhere.All || where === StartOverWhere.Undecided) {
+            const startOverOnUndecided = where === StartOverWhere.Undecided && undecided.length > 0;
+            startIndex = startOverOnUndecided ? decided.length : 0;
+        } else startIndex = this.answerSlides.findIndex(element => element.id === `textanswer-${where}`);
```
niklasmohrin: We should add an assertion that this did in fact find an index, otherwise we will get a cryptic error later

evap/staff/templates/staff_evaluation_textanswers_quick.html
```diff
@@ -200,6 +201,12 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
...
             "{% url 'staff:evaluation_textanswers_skip' %}",
         );
-        slider.attach();
+        const params = new URLSearchParams(document.location.search);
+        const textAnswerId = params.get("textanswer-id")
+        if (!textAnswerId) {
+            slider.attach()
+        } else {
+            slider.attach(textAnswerId)
+        }
```
niklasmohrin: Shouldn't we be able to just use `slider.attach(params.get(...))`? I guess we should change the type annotation to allow passing `null` then. Hm, I don't really care

evap/staff/templates/staff_evaluation_textanswers_section.html
```diff
@@ -26,7 +26,7 @@
                             </div>
                             <div>
                                 {% if user.is_manager %}
-                                    <a class="btn btn-sm btn-outline-secondary" href="{% url 'staff:evaluation_textanswer_edit' answer.id %}"><span class="fas fa-pencil"></a>
+                                    <a class="btn btn-sm btn-outline-secondary" href="{% url 'staff:evaluation_textanswer_edit' answer.id %}?view={{ view }}"><span class="fas fa-pencil"></a>
```
niklasmohrin: Please rename this to `next-view`, because it is not changing the view in the linked page

evap/staff/views.py
```diff
@@ -1768,11 +1828,15 @@ def evaluation_textanswer_edit(request, textanswer_id):
...
-
+    view = request.GET.get("view")
     if form.is_valid():
         form.save()
         # jump to edited answer
-        url = reverse("staff:evaluation_textanswers", args=[evaluation.pk], fragment=str(textanswer.id))
+        url = reverse(
+            "staff:evaluation_textanswers",
+            args=[evaluation.pk],
+            query={"view": view, "textanswer-id": textanswer.id} if view else {"textanswer-id": textanswer.id},
```
niklasmohrin: To avoid duplication of the textanswer-id bit here we could reformulate it to
```suggestion
            query={"textanswer-id": textanswer.id} | view_query_params,
```
where we define `view_query_params` accordingly above

edit: does not really apply anymore after my other comment https://github.com/e-valuation/EvaP/pull/2497/files#r2486491384

niklasmohrin: @janno42 There is also the possibility that the previous textanswer is not available in the view we are returning to, for example because we were in the Undecided view, clicked publish on an answer and then edited it. We _could_ make it so that we would return to the All view in that case, but it seems a bit excessive to me. So if we want it, I think we should do it in a separate PR. What do you think?

evap/staff/views.py
niklasmohrin, line 1837: We should keep the fragment around because it is actually used in the full view to scroll down to the edited answer (there is no code for this, just normal browser behavior). In fact, I feel that even in the quick view, the fragment would be the canonical place to put this "initial textanswer id" information. Can you change the code to use the fragment instead of a query parameter for both views?

evap/staff/views.py
janno42: Should be fine, we don't need to handle that case and can always stay in the current view.

evap/staff/templates/staff_evaluation_textanswers_quick.html
```diff
@@ -200,6 +201,7 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
             document.querySelector("#shared-full-textanswer-flag-form"),
             "{% url 'staff:evaluation_textanswers_skip' %}",
         );
-        slider.attach();
+        const hash = document.location.hash
+        slider.attach(hash.startsWith("#textanswer-") ? hash.substring(12) : null)
```
niklasmohrin: can you move it into a variable so that I don't have to be afraid that the string is truly 12 characters long?

niklasmohrin (changes requested):
Can you add (frontend) tests? Implementation looks good 👍

evap/staff/templates/staff_evaluation_textanswers_quick.html
```diff
@@ -200,6 +201,8 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
             document.querySelector("#shared-full-textanswer-flag-form"),
             "{% url 'staff:evaluation_textanswers_skip' %}",
         );
-        slider.attach();
+        const hash = document.location.hash
+        const hashPrefix = "#textanswer-"
+        slider.attach(hash.startsWith(hashPrefix) ? hash.substring(hashPrefix.length) : null)
```
niklasmohrin: strictly speaking, this should be
```suggestion
        slider.attach(hash.startsWith(hashPrefix) ? hash.substring(hashPrefix.length) : undefined)
```
to adhere to the typescript types

evap/staff/tests/test_live.py
```diff
@@ -174,3 +175,65 @@ def test_collapse_without_editor_approved(self) -> None:
...
+            course=baker.make(Course, programs=[baker.make(Program)], responsibles=[responsible]),
+            vote_start_datetime=datetime(2099, 1, 1, 0, 0),
+            vote_end_date=date(2099, 12, 31),
+            state=Evaluation.State.EVALUATED,
+            can_publish_text_results=True,
+        )
+
+        question1 = baker.make(
+            Question,
+        )
```
niklasmohrin: if you remove the trailing comma, our formatter would also yield
```suggestion
        question1 = baker.make(Question)
```

evap/staff/tests/test_live.py
```diff
@@ -174,3 +175,65 @@ def test_collapse_without_editor_approved(self) -> None:
...
+            TextAnswer,
+            question=question1,
+            contribution=contribution1,
+            answer="this is a dummy answer",
+            original_answer=None,
+            review_decision=TextAnswer.ReviewDecision.UNDECIDED,
+        )
+
+        with self.enter_staff_mode():
+            self.selenium.get(self.live_server_url + reverse("staff:evaluation_textanswer_edit", args=[textanswer1.pk]))
```
```suggestion
            self.selenium.get(self.reverse("staff:evaluation_textanswer_edit", args=[textanswer1.pk]))
```

niklasmohrin: more importantly, we are starting too far in here already; the test should start in the quick edit view, navigate to the question, click edit, edit it, save, and then we should assert that we are back on the quick view and looking at this question again

evap/staff/tests/test_live.py
```diff
@@ -174,3 +175,65 @@ def test_collapse_without_editor_approved(self) -> None:
...
+        textanswer_field.send_keys("edited answer")
+
+        with self.enter_staff_mode():
+            submit_btn.click()
+
+        answer = self.selenium.find_elements(
+            By.XPATH, "//div[@class='slider-item card-body active' and @data-layer='2']"
+        )
+
+        self.assertEqual(str(answer[0].get_attribute("id")).split("-", 1)[1], str(textanswer1.pk))
```
niklasmohrin: the `[0]` also seems like it could break arbitrarily depending on the order in which the answers were returned from the database; I think we can just check that the edited text is visible and the one from the other questions is invisible, take a look at existing tests that contain `self.wait.until(visibility_of_element ...)`

evap/staff/tests/test_live.py
```diff
@@ -174,3 +177,93 @@ def test_collapse_without_editor_approved(self) -> None:
...
+
+        while True:
+            try:
+                WebDriverWait(self.selenium, 1).until(
+                    visibility_of_element_located((By.ID, f"textanswer-{str(textanswer1.pk)}"))
+                )
+                break
+            except TimeoutException:
+                with self.enter_staff_mode():
+                    next_textanswer_btn.click()
```
niklasmohrin: We are not navigating here, so the following suffices:
```suggestion
                next_textanswer_btn.click()
```

evap/staff/tests/test_live.py
```diff
@@ -174,3 +177,93 @@ def test_collapse_without_editor_approved(self) -> None:
...
+
+        with self.enter_staff_mode():
+            submit_btn.click()
+
+        WebDriverWait(self.selenium, 1).until(
+            visibility_of_element_located((By.XPATH, "//div[contains(text(), 'edited answer')]"))
+        )
+        WebDriverWait(self.selenium, 1).until(
+            invisibility_of_element_located((By.XPATH, "//div[contains(text(), 'this is a dummy answer')]"))
+        )
```
niklasmohrin: I think here we don't need to diverge from our defaults
```suggestion
        self.wait.until(
            visibility_of_element_located((By.XPATH, "//div[contains(text(), 'edited answer')]"))
        )
        self.wait.until(
            invisibility_of_element_located((By.XPATH, "//div[contains(text(), 'this is a dummy answer')]"))
        )
```

evap/staff/tests/test_live.py
```diff
@@ -174,3 +177,93 @@ def test_collapse_without_editor_approved(self) -> None:
...
+            self.selenium.get(
+                self.reverse("staff:evaluation_textanswers", query={"view": "quick"}, args=[evaluation.pk])
+            )
+
+        next_textanswer_btn = self.selenium.find_element(By.XPATH, "//span[@data-slide='right']")
+        edit_btn = self.selenium.find_element(By.ID, "textanswer-edit-btn")
+
+        while True:
+            try:
+                WebDriverWait(self.selenium, 1).until(
```
niklasmohrin, line 244: For reference, I am a bit afraid that this is going to become flaky at some point, but we couldn't find a better way right now. So this is fine with me :)

evap/staff/tests/test_live.py
niklasmohrin: (not for this issue) we should have something like WebTestStaffMode for live server tests, we don't really care about staff mode in here

evap/staff/templates/staff_evaluation_textanswers_quick.html
```diff
@@ -88,6 +88,7 @@ <h5 class="modal-title" id="hotkeys-modal-title">Hotkeys</h5>
                                 <div
                                     class="slider-item card-body"
                                     data-layer="2"
+                                    id="textanswer-{{ answer.id }}"
                                     data-id="{{ answer.id }}"
```
richardebeling, line 92: `id`  and `data-id` both contain the same info --  do we need both? I feel like `data-id` should suffice for everything we're doing here

niklasmohrin: one is for scrolling to the right place, the other is for ergonomic access in javascript with `element.dataset.id` without string manipulation; I would keep it

janno42 (approved):

PR #2499, opened 2025-08-04
Structure settings in modules

niklasmohrin:
Since we are shipping EvaP via PyPI, using the evap.localsettings module doesn't really cut it anymore. We want to ship the majority of settings as default, but overwrite some of them in the deployment. Furthermore, we may want to have default config values that depend on values overwritten later, and we may want to mix in settings that add to previously defined values. Clearly, we need to implement the nix module system for our settings.

On production (with the current naming, which I want to still change later) it would be used like this:
```python
from evap.settings import default, dev, local_services, open_id, test
from evap.settings_resolver import resolve_settings


class LocalSettings:
    DEBUG = False
    SECRET_KEY = "xxx"  # nosec
    # ... some more required fields


globals().update(resolve_settings([default, open_id, LocalSettings]))
```
Alternatively, one can also make two Python files, one containing what I put in the `LocalSettings` class namespace here, and the other just containing `globals().update(...)` which then also includes the other file.


evap/development/cisettings.py
```diff
@@ -0,0 +1,10 @@
+from evap.settings import default, dev, local_services, open_id, test  # noqa: TID251
+from evap.settings_resolver import resolve_settings
+
+
+class CiSettings:
+    DEBUG = True
+    SECRET_KEY = "evap-github-actions-secret-key"  # nosec
```
richardebeling, line 7: This now doesn't have any type information, I could add any arbitrary key/value pairs here and nothing would complain, and my LSP wouldn't be able to help me autocomplete here, or am I missing some black magic voodoo?

(I'm fine with it, just want to double check, since we talked about it a bit)

evap/settings/local_services.py
```diff
@@ -0,0 +1,42 @@
+# pylint: disable=unused-argument,invalid-name
```
richardebeling, line 1: can we override linter settings inside the `settings` folder so that we don't have to ignore these in each file?

evap/settings/open_id.py
```diff
@@ -0,0 +1,25 @@
...
+ACTIVATE_OPEN_ID_LOGIN = required()
+
+# replace 'example.com', OIDC_RP_CLIENT_ID and OIDC_RP_CLIENT_SECRET with real values in localsettings when activating
+OIDC_RENEW_ID_TOKEN_EXPIRY_SECONDS = 60 * 60 * 24 * 7  # one week
+OIDC_RP_SIGN_ALGO = "RS256"
+OIDC_USERNAME_ALGO = ""
+OIDC_RP_SCOPES = "openid email profile"
+
+OIDC_RP_CLIENT_ID = "evap"
+OIDC_RP_CLIENT_SECRET = "evap-secret"  # nosec
```
richardebeling, line 12: Next level: Can we make these `required()` iff `ACTIVE_OPEN_ID_LOGIN` is true? (If not, that's fine with me, just wondering, seems not too hard from my simple mind)

evap/settings/schema.pyi
richardebeling, line 1: Can this be automatically derived from the .python file with some command-line-foo? If yes, should we put a comment here on how to do this?

I fear that if this is unchecked, it would go out-of-sync rather quickly. The advantage we gain is auto-completion if you have an LSP (which the average non-dev wouldn't have), right? In that case, I think I would vote to just go without type hints / type checking

evap/evaluation/tests/test_tools.py
```diff
@@ -242,3 +245,81 @@ def test_inside_transaction(self):
...
             self.assertTrue(inside_transaction())
+
+
+class TestResolveSettings(TestCase):
+    def test_not_set(self):
+        self.assertEqual(resolve_settings([Namespace(FOO=42), Namespace(FOO=not_set())]), {})
+        self.assertEqual(resolve_settings([Namespace(FOO=not_set()), Namespace(FOO=42)]), {"FOO": 42})
+
+    def test_key_names(self):
+        self.assertEqual(resolve_settings([Namespace(foo=1, _FOO=2, FoO=3)]), {})
```
richardebeling, line 258: These look a bit unintuitive at first glance. Can we make intention explicit with some short comments maybe?

* Why does having one namespace with a value set, and another namespace with a value not set, derive it to be unset? (-> What's a real-world use-case where we need this?)
* What are the three values in `test_key_names` ignored? (-> Why can't I have a setting with a different naming scheme?)

evap/evaluation/tests/test_tools.py
```diff
@@ -242,3 +245,81 @@ def test_inside_transaction(self):
...
+                [
+                    Namespace(
+                        FOO=1,
+                        BAR=derived(final={"FOO"})(lambda prev, final: final.FOO + 10),
+                    ),
+                    Namespace(
+                        FOO=derived(prev={"FOO"})(lambda prev, final: prev.FOO + 100),
+                        BAR=derived(prev={"BAR"})(lambda prev, final: prev.BAR + 1000),
+                    ),
+                ]
```
richardebeling, line 272: unhinged :D (but I see how a generic solver allows this)

What was the most complex real-world value dependency chain we have? I'm a bit worried that we end up with code that does not allow a human looking at the code to statically infer what the final value of a setting will be.

evap/settings/open_id.py
```diff
@@ -0,0 +1,25 @@
+from evap.settings_resolver import required
+
+ACTIVATE_OPEN_ID_LOGIN = required()
+
+# replace 'example.com', OIDC_RP_CLIENT_ID and OIDC_RP_CLIENT_SECRET with real values in localsettings when activating
```
richardebeling, line 5: I feel like the ones you would need to change should be part of localsettings.py, I think they don't really make sense in here. Any argument I'm missing for keeping them here?

evap/settings_resolver.py
```diff
@@ -0,0 +1,116 @@
...
+    prev: set[str]
+    final: set[str]
+
+
+def derived(*, prev: set[str] | None = None, final: set[str] | None = None) -> Callable[[Callable], Derived]:
+    return partial(Derived, prev=prev or set(), final=final or set())
+
+
+class Required(Enum):
+    REQUIRED = auto()
```
richardebeling, line 32: out of curiosity, does this have an advantage over a global `REQUIRED = object()`?

evap/settings_resolver.py
```diff
@@ -0,0 +1,116 @@
...
+
+class NotSet(Enum):
+    NOT_SET = auto()
+
+
+def not_set() -> NotSet:
+    return NotSet.NOT_SET
+
+
+class SettingResolver(Generic[T]):
```
richardebeling, line 50: Maybe add a short-ish doc comment that explains terminology and the general idea? I.e., what are `layers`, `prev`, `final`, `dep`s, and that we do topological sorting -> derive values in non-conflict order -> ??? -> profit

I think the code here is nice and makes sense, but a short introductory comment for potential new reads would help a lot to shape the mental image

richardebeling (commented):
Surprisingly unpainful to read, given the mental complexity that I felt our discussions about this required. Nice, I like it!

I'll have to play around with it a bit locally when the first round of discussions is through, particularly it's super hard on github to check whether stuff was forgotten when moving code around.

evap/development/cisettings.py
niklasmohrin: yeah, here there is no type info and I think there is no plain-Python way to do it easily, but I am not sure (I would guess one would have to write a short mypy plugin)

with the schema.pyi I got the django-stubs mypy plugin to add back type information to django.conf.settings, but it is not used for any setting modules; I haven't checked if typeguard would catch any mistyped setting at runtime, but I guess that it should do

evap/settings/local_services.py
niklasmohrin: sounds good, I'll check

evap/settings/open_id.py
niklasmohrin: yup, sounds good

evap/settings/schema.pyi
niklasmohrin: The problem is that without this some typing inside evap fails; for going out of sync, I think that typeguard should check this (or if not I should make it do so)

About non-devs: Why would we care about them here? The schema is only used for local development

Finally, about auto-generating: yeah, maybe, I think it should be doable to generate it based on the final types of for example the proposed local settings module and then compare it in CI. I am not sure if I like that more though, I'll think about it a bit. Best case would be to have the annotations in the setting modules directly, but I also didn't manage to get it to work (I stopped before committing coding crimes)

evap/evaluation/tests/test_tools.py
niklasmohrin: > Why does having one namespace with a value set, and another namespace with a value not set, derive it to be unset? (-> What's a real-world use-case where we need this?)

The whole not_set business is because we used to have some `if ...: X = 42` without else branches; the whole "overwrite setting with not_set()" thing is just what I felt is the natural generalization in the context of these layers. So this test is more of a "is the code doing what I think it is doing" check instead of actually documenting a needed feature

> What are the three values in `test_key_names` ignored? (-> Why can't I have a setting with a different naming scheme?)

The use case here is that there are more names flying around in the module like `not_set` or stuff like `Path` from pathlib, and those should not participate in the setting resolution. All settings we actually use are screaming snake case, so that was the natural choice, and the test should document this intention. Do you think we should do something else?

evap/evaluation/tests/test_tools.py
niklasmohrin: We don't really have complex prev and final cycles, and most chains are like 2 edges long or so. Usually we use prev to have a `def XXX: if ...: return new_value; else return prev.XXX`, and final for stuff like datadir.

evap/settings_resolver.py
niklasmohrin: yes, the reason for this is to have a separate type for the type checking in the resolver. This is also the reason it is generic, so that T is separate from Required etc., when in reality we only ever invoke it with `T = Any` which swallows the Required type anyways

richardebeling: ah, fancy

evap/evaluation/tests/test_tools.py
richardebeling: Ok, sounds good

evap/settings/schema.pyi
richardebeling: > About non-devs: Why would we care about them here? The schema is only used for local development

My reasoning was that the average dev doesn't really modify settings all that much, and if they do and put something of the wrong type, they likely will notice immediately (why would they put the setting there if they're not using it for developing). So I think having type checking on the settings is nice, but not super necessary. The failure scenario for a sysadmin is worse: The website would be broken, but the sysadmin would only notice once users hit the bug and the admin gets crash reports.

If our maintenance approach is "fix it whenever typeguard complains" then I'm fine with it 👍

evap/evaluation/tests/test_tools.py
richardebeling: Regarding `not_set`: Sounds good to me. Do we want to add a short comment, something like this?
> No known use case, but this is how we expect it should behave

Regarding the key names: Makes sense, maybe also a short comment?
> All setting names happen to be screaming upper case, we use that to filter out non-setting globals

Both not too pressing, I just didn't immediately get the intention when reading the tests here

evap/settings/local_services.py
niklasmohrin: i think only with https://pypi.org/project/pylint-per-file-ignores/ ; I think it's not worth the dependency

nix/services.nix
```diff
@@ -58,14 +58,15 @@
...
                 exit 0
             fi
             set -x
-            cp evap/development/localsettings.template.py evap/localsettings.py
-            sed -i -e "s/\$SECRET_KEY/$(head /dev/urandom | LC_ALL=C tr -dc A-Za-z0-9 | head -c 32)/" evap/localsettings.py
+            cp evap/development/localsettings.template.py localsettings.py
+            ln -s evap/settings/schema.pyi localsettings.pyi
+            sed -i -e "s/\$SECRET_KEY/$(head /dev/urandom | LC_ALL=C tr -dc A-Za-z0-9 | head -c 32)/" localsettings.py
             git submodule update --init
             ./manage.py compilemessages --locale de
```
fekoch, line 71:
```suggestion
           git submodule update --init
           export DJANGO_SETTINGS_MODULE=localsettings
           ./manage.py compilemessages --locale de
```

niklasmohrin:
Unsure on how to proceed, closing for now

PR #2501, opened 2025-08-11
Separate manual changes

### LINKED ISSUE #2138: Separate imported participants from manual changes
Once we use the import from #2131 on production, we need to separate imported participation data from manual changes. The import will update data regularly and should not overwrite manual changes made by editors or managers.

It should be possible to
- manually override that specific accounts are participants in an evaluation regardless of what data is imported
- manually override that specific accounts are *not* participants in an evaluation regardless of what data is imported
- revoke these decisions (so that the imported data will be used)

... and all this without a complicated UI with many participation fields.

A possible solution would be to add three fields `participants_imported`, `participants_manually_added` and `participants_manually_removed`.
`participants` will then be the result of `participants_imported + participants_manually_added - participants_manually_removed`.
The frontend would always show the computed participants list and do all the necessary additions and deletions to the respective lists when saving changes.

hansegucker:
Close #2138

### DIFF
--- a/evap/evaluation/migrations/0151_participantimportoverride.py
+++ b/evap/evaluation/migrations/0151_participantimportoverride.py
@@ -44,4 +44,8 @@ class Migration(migrations.Migration):
                 "verbose_name_plural": "participant import overrides",
             },
         ),
+        migrations.AddConstraint(
+            model_name="participantimportoverride",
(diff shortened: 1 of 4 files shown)


niklasmohrin:
We closed #2138, so also closing here

PR #2502, opened 2025-08-11
fix: incorrect reward points in export

### LINKED ISSUE #2478: Reward point export returns incorrect data
The reward point export returns incorrect data, because it combines multiple aggregations, [which should not be done](https://docs.djangoproject.com/en/5.2/topics/db/aggregation/#combining-multiple-aggregations).

Example:
Currently used code:
```
>>> us = UserProfile.objects.annotate(
...     points=Sum("reward_point_grantings__value", default=0) - Sum("reward_point_redemptions__value", default=0)
... )
>>> us.get(email="<email>").points
-177
```

Without combined aggregation:
```
>>> u = UserProfile.objects.get(email="<email>")
>>> u.reward_point_grantings.all().aggregate(Sum("value"))
{'value__sum': 22}
>>> u.reward_point_redemptions.all().aggregate(Sum("value"))
{'value__sum': 13}
```
The second implementation returns correct values, the difference here would be the expected `22 - 13 = 9` instead of `-177`

velramiir:
This closes #2478.

This pull request fixes the calculation of reward points in the reward points export.

* In `evap/rewards/views.py`: Replaced the two problematic `Sum` aggregations within an `annotated` clause with two subqueries. They are surrounded with `Coalesce` to handle users with no grantings or redemptions respectively. The subqueries first filter for grantings or redemptions that are related to the respective user in the outer query. If there are no matching rows, then the annotation has no effect, so the default value in the `Sum` clause is not sufficient to ensure 0 is returned instead of `None`.

### DIFF
--- a/evap/rewards/views.py
+++ b/evap/rewards/views.py
@@ -5,7 +5,6 @@ from django.contrib import messages
 from django.contrib.messages.views import SuccessMessageMixin
 from django.core.exceptions import BadRequest, SuspiciousOperation
 from django.db.models import OuterRef, Subquery, Sum
-from django.db.models.functions import Coalesce
 from django.http import HttpResponse
 from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse_lazy
@@ -156,22 +155,17 @@ def reward_points_export(request):
 
     writer = csv.writer(response, delimiter=";", lineterminator="\n")
     writer.writerow([_("Email address"), _("Number of points")])
-    sub_reward_point_grantings = (
-        RewardPointGranting.objects.filter(user_profile=OuterRef("pk"))
-        .values("user_profile")
-        .annotate(total=Sum("value", default=0))
-        .values("total")
-    )
-    sub_reward_point_redemptions = (
-        RewardPointRedemption.objects.filter(user_profile=OuterRef("pk"))
-        .values("user_profile")
-        .annotate(total=Sum("value", default=0))
-        .values("total")
-    )
+
+    def aggregated_sum(relation: str) -> Subquery:
+        return Subquery(
+            UserProfile.objects.filter(pk=OuterRef("pk"))
+            .annotate(aggregated=Sum(f"{relation}__value", default=0))
+            .values("aggregated")
+        )
+
     profiles_with_points = (
         UserProfile.objects.annotate(
-            points=Coalesce(Subquery(sub_reward_point_grantings), 0)
-            - Coalesce(Subquery(sub_reward_point_redemptions), 0)
+            points=(aggregated_sum("reward_point_grantings") - aggregated_sum("reward_point_redemptions"))
         )
         .filter(points__gt=0)
         .order_by("-points")


velramiir:
This still needs tests.

evap/rewards/views.py
```diff
@@ -155,9 +156,22 @@ def reward_points_export(request):
 
     writer = csv.writer(response, delimiter=";", lineterminator="\n")
     writer.writerow([_("Email address"), _("Number of points")])
+    sub_reward_point_grantings = (
+        RewardPointGranting.objects.filter(user_profile=OuterRef("pk"))
+        .values("user_profile")
```
richardebeling: Do we need `.values("user_profile")` in both these subqueries? I think this has no effect here (and confused me a bit) -- or I'm missing something.

evap/rewards/views.py
```diff
@@ -155,9 +156,22 @@ def reward_points_export(request):
...
+        .annotate(total=Sum("value", default=0))
+        .values("total")[:1]
```
richardebeling: with `total` being a sum, it should only have one row, so the `[:1]` limit should not be necessary. I would remove it in that case

evap/rewards/views.py
```diff
@@ -155,9 +156,22 @@ def reward_points_export(request):
...
+    sub_reward_point_redemptions = (
+        RewardPointRedemption.objects.filter(user_profile=OuterRef("pk"))
+        .values("user_profile")
+        .annotate(total=Sum("value", default=0))
+        .values("total")[:1]
+    )
     profiles_with_points = (
         UserProfile.objects.annotate(
-            points=Sum("reward_point_grantings__value", default=0) - Sum("reward_point_redemptions__value", default=0)
+            points=Coalesce(Subquery(sub_reward_point_grantings), 0)
```
richardebeling: We use `Sum("value", default=0)` above, which to me looks like it would always return a number, never NULL. Is the `Coalesce` required here? If not, I'd simplify

richardebeling (commented):
Thanks! Looks like some commits slipped in there that shouldn't be there, possibly because there was some force-push sketchiness on main. You'll likely need to rebase again and remove those commits, we can help if there are any problems with that

evap/rewards/views.py
velramiir: See PR description on this. I'm not 100% sure the reason I gave is correct. However, it does not work without coalesce

evap/rewards/views.py
velramiir: This part is needed for grouping.

evap/rewards/views.py
richardebeling: I had a wrong mental model here, we will have a a `total` column with a value for each original row, all returning the same value. "aggregate" return a single aggregate value, but we can't use that with subqueries

evap/rewards/views.py
richardebeling: You're right. Since the subquery queries on RewardPointGrantings/Redemptions, we will not have rows for users that don't have any -- that means that the `default=0` in the `Sum` has no effect (and can be removed), and we need the `Coalesce`.

evap/rewards/views.py
richardebeling: my proposal:
```python3
    # Use of subquery is forced to make sure that join doesn't mess with sums, see #2478
    UserProfile.objects.annotate(
        points=(
            Subquery(
                UserProfile.objects.filter(pk=OuterRef("pk"))
                .annotate(granted=Sum("reward_point_grantings__value", default=0))
                .values("granted")
            )
            - Subquery(
                UserProfile.objects.filter(pk=OuterRef("pk"))
                .annotate(redeemed=Sum("reward_point_redemptions__value", default=0))
                .values("redeemed")
            )
        )
    )
```

richardebeling: Maybe
```python3
    def aggregated_sum(relation: str) -> Subquery:
        return Subquery(
            UserProfile.objects.filter(pk=OuterRef("pk"))
            .annotate(aggregated=Sum(f"{relation}__value", default=0))
            .values("aggregated")
        )

    UserProfile.objects.annotate(
        points=(aggregated_sum("reward_point_grantings") - aggregated_sum("reward_point_redemptions"))
    )
```

janno42 (approved):

niklasmohrin (approved):
Awesome!

PR #2511, opened 2025-09-22
Switch tracking of views with no-login-required to a global WeakSet

### LINKED ISSUE #2505: Server Error on permission checks
When an anonymous user (not logged in) requests a view that requires permission (e.g., `grades.views.SemesterView`), an `AttributeError` occurs because the `AnonymousUser` doesn't have the respective attribute (e.g., `is_grade_publisher`).

Instead, the permission decorator functions should always return `False` for anonymous users.

richardebeling:
fixes #2505


Kakadus (approved):

evap/middleware.py
```diff
@@ -13,29 +23,30 @@ def __call__(self, request):
...
             return None
 
-        if "no_login_required" in view_func.__dict__ and view_func.no_login_required:
-            return None
-
-        if hasattr(view_func, "view_class") and view_func.view_class.__name__ in [
-            "OIDCAuthenticationRequestView",
-            "OIDCAuthenticationCallbackView",
-        ]:
+        view = getattr(view_func, "view_class", view_func)
```
niklasmohrin, line 41: This breaks catalog.js which we declared as `no_login_required(JavaScriptCatalog.as_view())`

niklasmohrin: So we should fix the catalogview usage and probably add a check to the decorator to ensure that it is not used on functions that have view_class set

PR #2512, opened 2025-09-22
Add reverse helper to LiveServerTests

### LINKED ISSUE #2488: Simplify URL construction in Selenium tests
In normal tests, we can issue requests with `self.app.get(reverse(...))`. However, in live server tests it looks like this:

```python
self.selenium.get(self.live_server_url + reverse("staff:semester_view", args=[semester.pk]))
```

We should come up with a way to not always write `self.live_server_url + `.

_Originally posted by @niklasmohrin in https://github.com/e-valuation/EvaP/pull/2474#discussion_r2237031887_

richardebeling:
fixes #2488

### DIFF
--- a/evap/contributor/tests/test_live.py
+++ b/evap/contributor/tests/test_live.py
@@ -1,6 +1,5 @@
 from datetime import date, datetime
 
-from django.urls import reverse
 from model_bakery import baker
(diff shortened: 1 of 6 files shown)


Kakadus (approved):
small but effective :+1:

PR #2522, opened 2025-10-06
Allow extra keys in JSON importer pydantic validation

hansegucker:


niklasmohrin:
Now that I think about it, we should have a test that has an extra attribute 😅

evap/staff/importers/json.py
richardebeling, line 1: For my understanding: The use case here is that the imported json contains additional data that we don't care about, right?

niklasmohrin: also: when we request new data to be included in the json, the CAS team will not sync their update where the json schema changes with our release schedule

PR #2527, opened 2025-10-13
Add support for ignoring specific users during JSON import

### LINKED ISSUE #2523: List of users to ignore on import
Some fictional test users are registered for courses in the campus management system. These would be imported into EvaP by the JSON import, which would distort the actual participation numbers.

Similar to the `IGNORE_PROGRAMS` setting, a new `IGNORE_USERS` setting should be introduced. Any users listed there via email should be completely ignored during import. They should not become participants or contributors.

Belissimo-T:
fixes #2523

### DIFF
--- a/evap/staff/tests/test_json_importer.py
+++ b/evap/staff/tests/test_json_importer.py
@@ -150,7 +150,7 @@ EXAMPLE_DATA_SPECIAL_CASES: ImportDict = {
             "title_en": "",
             "type": "Vorlesung",
             "isexam": False,
-            "lecturers": [{"gguid": "0x4"}, {"gguid": "0x5"}],
+            "lecturers": [{"gguid": "0x4"}, {"gguid": "0x5"}, {"gguid": "0x12"}],
             "students": [],
             "appointments": [{"begin": "29.07.2024 10:15:00", "end": "29.07.2024 11:45:00"}],
         },
@@ -529,7 +529,7 @@ class TestImportEvents(TestCase):
         self.assertEqual(evaluation_life.weight, 1)
 
     def test_import_ignore_non_responsible_users(self):
-        with override_settings(NON_RESPONSIBLE_USERS=["4@example.com"]):
+        with override_settings(NON_RESPONSIBLE_USERS=["4@example.com", "ignored.lecturer2@example.com"]):
             self._import(EXAMPLE_DATA_SPECIAL_CASES)
             evaluation = Evaluation.objects.get(cms_id="0x9")
             self.assertEqual(set(evaluation.course.responsibles.values_list("email", flat=True)), {"5@example.com"})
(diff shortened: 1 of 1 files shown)


evap/staff/importers/json.py
```diff
@@ -182,7 +182,7 @@ def _get_users_with_longest_title(self, user_profiles: list[UserProfile]) -> lis
         return [user for user in user_profiles if len(user.title) == max_title_len]
 
     def _remove_non_responsible_users(self, user_profiles: list[UserProfile]) -> list[UserProfile]:
-        return list(filter(lambda p: p.email not in settings.NON_RESPONSIBLE_USERS, user_profiles))
+        return [user for user in user_profiles if user.email not in settings.NON_RESPONSIBLE_USERS]
```
hansegucker, line 184: Good take!

evap/settings.py
```diff
@@ -113,12 +113,15 @@
...
 EXAM_QUESTIONNAIRE_IDS: list[int] = []
 
-# Emails of users that shouldn't be imported as responsibles during JSON import
+# Emails of users that shouldn't be imported as responsibles and contributors during JSON import
 NON_RESPONSIBLE_USERS: set[str] = set()
 
 # Study programs that shouldn't be imported during JSON import
 IGNORE_PROGRAMS: set[str] = set()
 
+# Users that shouldn't be imported as participants or contributors during JSON import
```
niklasmohrin: Can you clarify what strings should be put into the set? For the other setting, it explicitly says "emails", I would like that here as well

evap/staff/importers/json.py
```diff
@@ -222,6 +222,9 @@ def _create_name_change_from_changes(self, user_profile: UserProfile, changes: d
 
     def _import_students(self, data: list[ImportStudent]) -> None:
         for entry in data:
+            if entry["email"] in settings.IGNORE_USERS:
+                continue
+
             email = clean_email(entry["email"])
```
niklasmohrin: This is an interesting detail: the filtering is based on the email we get from the json _before_ cleaning. I think we should not do this, in particular because with the `NON_RESPONSIBLE_USERS` setting we _do_ use the cleaned email because we check the email of a created UserProfile.

Either way, we should add a test for this subtle detail! If you are up for it, we should also test this for the other setting, but you don't have to do it.

@janno42 any additional thoughts?

janno42: Good catch. Yes, we should use the cleaned email here.

evap/staff/tests/test_json_importer.py
```diff
@@ -503,7 +521,13 @@ def test_import_courses_special_cases(self):
         self.assertEqual(evaluation_life.weight, 1)
 
     def test_import_ignore_non_responsible_users(self):
-        with override_settings(NON_RESPONSIBLE_USERS=["4@example.com"]):
+        with override_settings(
+            NON_RESPONSIBLE_USERS=[
+                "4@example.com",
+                # to test that this filters cleaned emails
```
niklasmohrin: This is approaching this from the wrong angle though, right? It is never meaningful to have unclean emails in the setting. What we should test however is that if a user comes in with an unclean email, but their cleaned email is in the setting, then they still get filtered.

niklasmohrin (approved):
Thanks

PR #2530, opened 2025-10-20
search reset for result page

### LINKED ISSUE #2525: Clear text filter button on results index page
A 'Clear filter' button similar to the one on, e.g., the questionnaire list should be added to the text search field on the results index page. Unlike the existing 'Clear all filters' button, this button would only clear the text field and reset this part of the filter.

HannesClausnitzer:
fixes #2525

### DIFF
--- a/evap/static/scss/_components.scss
+++ b/evap/static/scss/_components.scss
@@ -3,7 +3,6 @@
 @import "components/tables";
 @import "components/forms";
 @import "components/buttons";
-@import "components/input";
 @import "components/transitions";
 @import "components/dropdown";
 @import "components/button-group";


evap/results/templates/textanswer_list.html
richardebeling, line 17: Fyi: You continued working here from your branch from https://github.com/e-valuation/EvaP/pull/2526, so the commit that added the `d-print-none` here is also included in this PR.

Should not cause any trouble, but if you want to get some experience with git interactive rebase, attempting to clean this up would be a nice exercise :D (make sure to make a backup of the branch, first)

richardebeling (approved):
Code looks good to me, @janno42 will review UI changes

richardebeling (approved):

evap/results/templates/results_index.html
```diff
@@ -21,8 +21,14 @@ <h5 class="card-title">
...
                     <label>
-                        <input type="search" name="search" class="form-control mb-3" placeholder="{% translate 'Search...' %}" />
+                        <div class="input-group">
+                            <input type="search" name="search" class="form-control" placeholder="{% translate 'Search...' %}" />
+                            <button class="btn btn-light text-secondary" type="button" data-reset="search" data-bs-toggle="tooltip" data-bs-placement="top" title="{% translate 'Clear search filter' %}">
+                                <span class="fas fa-delete-left"></span>
+                            </button>
+                        </div>
                     </label>
+
```
niklasmohrin: This added line is not needed :)

```suggestion
```

evap/static/ts/src/datagrid.ts
```diff
@@ -465,6 +455,13 @@ export class ResultGrid extends DataGrid {
             this.reflectFilterStateOnInputs();
         });
 
+        this.resetSearch.addEventListener("click", () => {
+            this.state.search = "";
+            this.filterRows();
+            this.renderToDOM();
+            this.reflectFilterStateOnInputs();
+        });
```
niklasmohrin: This (and the rest of the code) is now duplicated with TableGrid. I think instead of doing that, we should generalize the functionality from TableGrid to work on any DataGrid so that it is also available in the ResultGrid.

The thing is, not all data grids on evap have such a reset button. So we must make the `resetSearch` parameter optional when adding it to DataGrid.

Does that make sense? What do you think?

(cc @fekoch @ronjakrg because I think you work on something similar)

niklasmohrin (changes requested):

evap/static/ts/src/datagrid.ts
fekoch: The same applies for the filter buttons (and search in general?)

But yeah, I think it would be best, if we have all that functionality available in DataGrid.

evap/static/scss/evap.scss
```diff
@@ -22,6 +22,10 @@ body {
     &.mb-footer {
         margin-bottom: 60px; // keep space for footer
     }
+
+    ::-webkit-search-cancel-button {
```
niklasmohrin: hm, right now this style is applied to the entire html body; I would expect that we just do this in the places where we provide our own reset button.

@HannesClausnitzer Is there a good reason to do it globally?

evap/static/ts/src/datagrid.ts
```diff
@@ -465,6 +455,13 @@ export class ResultGrid extends DataGrid {
             this.reflectFilterStateOnInputs();
         });
 
+        this.resetSearch?.addEventListener("click", () => {
```
niklasmohrin: This is still duplicated with TableGrid: Both TableGrid and ResultGrid have the same code for such reset buttons. Instead, we should move the implementation from TableGrid up into the DataGrid base class. Then, we can just use it with ResultGrid, because ResultGrid also inherits from DataGrid.

evap/static/ts/src/datagrid.ts
```diff
@@ -396,6 +392,7 @@ interface ResultGridParameters extends DataGridParameters {
     sortColumnSelect: HTMLSelectElement;
     sortOrderCheckboxes: HTMLInputElement[];
     resetFilter: HTMLButtonElement;
+    resetSearch: HTMLButtonElement;
```
niklasmohrin: This _narrows_ the ResultGridParameters type so that `resetSearch` is required. I think we can leave it optional
```suggestion
```

niklasmohrin (dismissed):
nice :+1:

evap/static/scss/components/_input.scss
janno42, line 1: Please move this into `adjustments.scss`, we probably don't need an individual file for this one change.

richardebeling (dismissed):
Code looks good to me, thanks! :)

jooooosef (approved):
code looks good to me :)

PR #2531, opened 2025-10-20
removed unnecessary log info for general contributions

### LINKED ISSUE #2298: Unnecessary log info for general contribution
When creating an evaluation, a general contribution for this evaluation is created. This is logged as
```
 The Contribution "General Contribution" was created.

    Role: Contributor
    Label:
    Text answer visibility: Own
```

The values (role, label, and text answer visibility) are unnecessary technical details because they have no effect and cannot be changed. The log entry should omit this information, and simply indicate that the general contribution was created.

cakeillionair:
log entry is not removed entirely
fixes #2298

### DIFF
--- a/evap/evaluation/models_logging.py
+++ b/evap/evaluation/models_logging.py
@@ -258,7 +258,10 @@ class LoggedModel(models.Model):
             self._logentry = self._create_log_entry(*args, **kwargs)
 
     def _update_log(self, changes, action_type: InstanceActionType, store_in_db=True):
-        if (not changes and action_type != InstanceActionType.CREATE) or not CREATE_LOGENTRIES:
+        if not CREATE_LOGENTRIES:
+            return
+        if action_type == InstanceActionType.CHANGE and not changes:
+            # All changes are on unlogged fields
             return
 
         self._attach_log_entry_if_not_exists(action_type)
--- a/evap/staff/tests/test_views.py
+++ b/evap/staff/tests/test_views.py
@@ -2395,8 +2395,7 @@ class TestEvaluationEditView(WebTestStaffMode):
         self.assertEqual(Evaluation.objects.first().state, self.evaluation.State.APPROVED)
 
     def test_general_contribution_log_entry(self):
-        evaluation = baker.make(Evaluation)
-        page = self.app.get(reverse("staff:evaluation_edit", args=[evaluation.pk]), user=self.manager)
+        page = self.app.get(self.url, user=self.manager)
         self.assertContains(
             page, '<p class="mt-3">The Contribution "General Contribution" was created.</p><ul></ul>', html=True
         )


evap/evaluation/models.py
```diff
@@ -1146,7 +1146,15 @@ class Meta:
...
     def unlogged_fields(self):
-        return super().unlogged_fields + ["evaluation"] + (["contributor"] if self.is_general else [])
+        return (
+            super().unlogged_fields
+            + ["evaluation"]
+            + (["contributor"] if self.is_general else [])
+            + (["role"] if self.is_general else [])
+            + (["role"] if self.is_general else [])
+            + (["label"] if self.is_general else [])
+            + (["textanswer_visibility"] if self.is_general else [])
```
niklasmohrin: you can make this shorter using something like this:
```suggestion
            + (["contributor", "role", "label", "textanswer_visibility"] if self.is_general else [])
```

evap/evaluation/models_logging.py
```diff
@@ -258,7 +258,7 @@ def _attach_log_entry_if_not_exists(self, *args, **kwargs):
             self._logentry = self._create_log_entry(*args, **kwargs)
 
     def _update_log(self, changes, action_type: InstanceActionType, store_in_db=True):
-        if not changes or not CREATE_LOGENTRIES:
+        if (not changes and action_type != InstanceActionType.CREATE) or not CREATE_LOGENTRIES:
```
niklasmohrin: I guess if we always log creation, we should also log deletion? @richardebeling What do you think?

If so, I think I would restructure the condition like this:
```suggestion
        if not CREATE_LOGENTRIES:
            return
        if action_type == InstanceActionType.CHANGE and not changes:
            # All changes are on unlogged fields
            return
```
@cakeillionair How do you feel about this rephrasing?

evap/staff/tests/test_views.py
```diff
@@ -2394,6 +2394,13 @@ def test_evaluation_confirm_without_main_language(self):
 
         self.assertEqual(Evaluation.objects.first().state, self.evaluation.State.APPROVED)
 
+    def test_general_contribution_log_entry(self):
+        evaluation = baker.make(Evaluation)
+        page = self.app.get(reverse("staff:evaluation_edit", args=[evaluation.pk]), user=self.manager)
```
niklasmohrin: In this test class, there is already an evaluation (and corresponding url) set up in `setUpTestData` which we can use here like so:
```suggestion
        page = self.app.get(self.url, user=self.manager)
```
Or did you create a separate evaluation on purpose and I am not understanding something?

cakeillionair: I did not realize I could just use that

evap/evaluation/models_logging.py
cakeillionair: Are the only actions with changes InstanceActionType.CREATE and InstanceActionType.CHANGE?

cakeillionair: I just checked and yes they are. I think this is an improvement.

cakeillionair:
> Hm, it seems that the git diff is the same after the last push - are your intended changes included here?

not sure what h

janno42 (approved):
Thank you for your first PR! 🎉

PR #2532, opened 2025-10-20
Fix questionnaire warnings for dropout questions

### LINKED ISSUE #2521: Don't show participant warning in dropout questionnaires
The text "Only a few participants answered these questions" should not be shown in the dropout questionnaire card on an evaluation's results page. Also, result bars from the dropout questionnaire should not be half-transparent and should not contain the warning text in the tooltip.

rsahwe:
Fixes #2521

### DIFF
--- a/evap/results/views.py
+++ b/evap/results/views.py
@@ -407,6 +407,8 @@ def add_warnings(evaluation, evaluation_result):
         )
 
     for questionnaire_result in evaluation_result.questionnaire_results:
+        if questionnaire_result.questionnaire.is_dropout:
+            continue
         rating_results = [
             question_result
             for question_result in questionnaire_result.question_results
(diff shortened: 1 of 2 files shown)


evap/results/views.py
```diff
@@ -415,13 +415,15 @@ def add_warnings(evaluation, evaluation_result):
         max_answers = max((rating_result.count_sum for rating_result in rating_results), default=0)
         questionnaire_result.warning = (
             0 < max_answers < questionnaire_warning_thresholds[questionnaire_result.questionnaire]
+            and not questionnaire_result.questionnaire.is_dropout
```
niklasmohrin: Good idea adding the check here instead of in the template, that certainly makes the code simpler!

I think it would be slightly easier to understand if we would insert a snippet like
```python
if questionnaire_result.questionnaire.is_dropout:
    continue
```
above the code here instead of pushing it into the conditions. If it would be only the `questionnaire_result.warning`, then I would also do it like this, but since we also want to not set `rating_result.warning` below, having the guard clause is simpler. What do you think?

rsahwe: I didn't do that because then the warning would be unset instead of false (assuming that that is a problem)

niklasmohrin: Ah, that's a fair intuition. I feel like we still shouldn't compromise on readability right here and then instead set `warning = False` in the constructor of the `*Result` classes.

For reference: It also wouldn't be a problem for the template, because in Django templates when some attribute is missing it just evaluates to something falsy. I agree that this is not so nice though

niklasmohrin (approved):

PR #2533, opened 2025-10-20
fixed wrong redirect when changing language in dropout

### LINKED ISSUE #2520: Questionnaire language change switches away from dropout questionnaire
After clicking the button to mark a course as dropped, the evaluation with dropout questionnaire opens. Changing the questionnaire language then reloads the page and the normal evaluation without the dropout questionnaire is shown.
Instead, the dropout questionnaire should still be shown, only the language should change.

ZoqkMaze:
fixes #2520

### DIFF
--- a/evap/student/tests/test_views.py
+++ b/evap/student/tests/test_views.py
@@ -713,8 +713,5 @@ class TestDropoutView(WebTest):
         page = self.app.get(url=url_dropout, user=self.user, status=200)
         html_site = page.body.decode()
 
-        # do not redirect to vote page
         self.assertNotIn(url_vote, html_site)
-
-        # redirect to vote page
         self.assertIn(url_dropout, html_site)


evap/student/templates/student_vote.html
```diff
@@ -60,12 +60,16 @@ <h3 class="mb-3">
...
-                           class="btn btn-sm btn-light{% if evaluation_language == language_code %} active{% endif %}">
+                            data-bs-placement="bottom"
+                            data-set-spinner-icon="span-set-evaluation-language-{{ language_code }}"
+                            title="{{ language_name }}"
+                            role="button"
+                            {% if evaluation_form_group_dropout %}
+                                href="{% url 'student:drop' evaluation.pk %}?language={{ language_code }}"
+                            {% else %}
+                                href="{% url 'student:vote' evaluation.pk %}?language={{ language_code }}"
+                            {% endif %}
```
niklasmohrin: Hm, I would think that there is a way to get whatever the current URL is that we are on using Django templates. Can you check if there is something like that? Then we wouldn't need the if-else and could just always use the current URL

niklasmohrin (changes requested):
Can you add a short test for this behavior in `evap/student/tests/test_views.py`? (We can help you with doing so, of course)

evap/student/templates/student_vote.html
ZoqkMaze: I changed the code to use the django template instead of if-else. A test for the behavior will follow

evap/student/templates/student_vote.html
```diff
@@ -60,12 +60,12 @@ <h3 class="mb-3">
                 <div class="btn-switch btn-group">
                     {% for language_code, language_name in languages %}
                         <a data-bs-toggle="tooltip"
-                           data-bs-placement="bottom"
-                           data-set-spinner-icon="span-set-evaluation-language-{{ language_code }}"
-                           title="{{ language_name }}"
-                           role="button"
-                           href="{% url 'student:vote' evaluation.pk %}?language={{ language_code }}"
-                           class="btn btn-sm btn-light{% if evaluation_language == language_code %} active{% endif %}">
+                            data-bs-placement="bottom"
```
richardebeling: You added an extra space of indentation here to each line, please keep the old indentation scheme (where each attribute is aligned to the attribute on the first line)

evap/student/templates/student_vote.html
```diff
@@ -60,12 +60,12 @@ <h3 class="mb-3">
...
+                            data-set-spinner-icon="span-set-evaluation-language-{{ language_code }}"
+                            title="{{ language_name }}"
+                            role="button"
+                            href="{{request.path}}?language={{ language_code }}"
```
richardebeling: For consistency, please use spaces inside the `{{ }}` pairs:

```suggestion
                            href="{{ request.path }}?language={{ language_code }}"
```

richardebeling (commented):
Code generally looks good to me, waiting for test

evap/student/tests/test_views.py
```diff
@@ -706,3 +706,17 @@ def test_choice_fields_correctly_populated_by_server(self):
...
+
+    def test_change_language_in_dropout(self):
+        url_dropout = reverse("student:drop", args=[self.evaluation.id])
+        url_vote = reverse("student:vote", args=[self.evaluation.id])
+        page = self.app.get(url=url_dropout, user=self.user, status=200)
+        html_site = page.body.decode()
+
+        # do not redirect to vote page
+        self.assertNotIn(f"{url_vote}?language=en", html_site)
+        self.assertNotIn(f"{url_vote}?language=de", html_site)
```
richardebeling: Would this assertion also pass? If yes, I'd prefer to use that (so that this assertion does not _just_ pass because the query part of the URL changed)

```suggestion
        # do not redirect to vote page
        self.assertNotIn(url_vote, html_site)
```

If this doesn't work, disregard

evap/student/templates/student_vote.html
```diff
@@ -64,8 +64,9 @@ <h3 class="mb-3">
                            data-set-spinner-icon="span-set-evaluation-language-{{ language_code }}"
                            title="{{ language_name }}"
                            role="button"
-                           href="{% url 'student:vote' evaluation.pk %}?language={{ language_code }}"
-                           class="btn btn-sm btn-light{% if evaluation_language == language_code %} active{% endif %}">
+                           href="{{ request.path }}?language={{ language_code }}"
+                           class="btn btn-sm btn-light{% if evaluation_language == language_code %} active{% endif %}"
+                           id="chg-lang">
```
richardebeling: The indentation is fixed now, but it looks like you accidentally added a new `id="chg-lang">` attribute that looks like it's not needed to me?


```suggestion
                           class="btn btn-sm btn-light{% if evaluation_language == language_code %} active{% endif %}">
```

evap/student/tests/test_views.py
```diff
@@ -706,3 +706,15 @@ def test_choice_fields_correctly_populated_by_server(self):
...
+        url_dropout = reverse("student:drop", args=[self.evaluation.id])
+        url_vote = reverse("student:vote", args=[self.evaluation.id])
+        page = self.app.get(url=url_dropout, user=self.user, status=200)
+        html_site = page.body.decode()
+
+        # do not redirect to vote page
+        self.assertNotIn(url_vote, html_site)
+
+        # redirect to vote page
+        self.assertIn(url_dropout, html_site)
```
niklasmohrin:
```suggestion
        self.assertNotIn(url_vote, html_site)
        self.assertIn(url_dropout, html_site)
```

PR #2534, opened 2025-10-20
CSS Grid for staff_semester_view & grades_semester_view

### LINKED ISSUE #2352: Use CSS grids for all major tables

fekoch:
fix #2352 

done:
 
 - staff_semester_view

in progress:

 - grades_semester_view

### DIFF
--- a/evap/evaluation/tests/tools.py
+++ b/evap/evaluation/tests/tools.py
@@ -280,6 +280,7 @@ class LiveServerTest(SeleniumTestCase):
     browser = "firefox"
     selenium: WebDriver
     headless = True
+    viewport = "1920x4096"
     window_size = (1920, 4096)  # large height to workaround scrolling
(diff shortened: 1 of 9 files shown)


visual-regresion-tracking-evap/docker-compose.yml
niklasmohrin, line 1: this seems very excessive, I would hope that there is an easier way to set up something similar; either way, let's discuss this in #2046

richardebeling: Also, for the sake of being able to review and merge stuff, please separate the "frontend looks testing" and the "CSS grid" portions here into at least two separate PRs

niklasmohrin:
@fekoch This PR is pretty outdated now, would it makes sense to close it until vrt works, or do you want to continue on this in parallel?

fekoch:
> @fekoch This PR is pretty outdated now, would it makes sense to close it until vrt works, or do you want to continue on this in parallel?

Yeah, closing it for now.

PR #2540, opened 2025-10-27
fixed wrong word limiter for answer warnings

### LINKED ISSUE #2516: Regard word limits in text answer warnings
When text answer warnings are displayed, word limits should be regarded - only full words should match. Whitespace and string boundaries are word limits.
E.g., when the trigger word is "talk", the warning should not be shown if someone writes "talking".

AlexanderSchicktanz:
fix #2516

### DIFF
--- a/evap/static/ts/src/text-answer-warnings.ts
+++ b/evap/static/ts/src/text-answer-warnings.ts
@@ -73,4 +73,3 @@ export const testable = {
     doesTextContainTriggerString,
     matchesTriggerString,
 };
-
--- a/evap/static/ts/src/text-answer-warnings.ts
+++ b/evap/static/ts/src/text-answer-warnings.ts
@@ -7,13 +7,12 @@ function isTextMeaningless(text: string): boolean {
 }
 
 function containsSubArray(arr: string[], sub: string[]): boolean {
-    for (let i = 0; i < arr.length; i++) {
-        if (i + sub.length > arr.length) return false;
+    for (let i = 0; i <= arr.length - sub.length; i++) {
         let j;
         for (j = 0; j < sub.length; j++) {
-            if (arr[i + j] !== sub[j]) return false;
+            if (arr[i + j] !== sub[j]) break;
         }
-        return j == sub.length;
+        if (j == sub.length) return true;
     }
     return false;
 }
--- a/evap/static/ts/src/text-answer-warnings.ts
+++ b/evap/static/ts/src/text-answer-warnings.ts
@@ -19,7 +19,6 @@ function containsSubArray(arr: string[], sub: string[]): boolean {
 }
 
 function matchesTriggerString(text: string, triggerString: string): boolean {
-    console.log("Works");
     const words = text.split(" ");
     const triggerWords = triggerString.split(" ");
     return containsSubArray(words, triggerWords);
@@ -74,3 +73,4 @@ export const testable = {
     doesTextContainTriggerString,
     matchesTriggerString,
 };
+
--- a/evap/static/ts/src/text-answer-warnings.ts
+++ b/evap/static/ts/src/text-answer-warnings.ts
@@ -6,13 +6,13 @@ function isTextMeaningless(text: string): boolean {
     return text.length > 0 && ["", "ka", "na", "none", "keine", "keines", "keiner"].includes(text.replace(/\W/g, ""));
 }
 
-function containsSubArray<Type>(arr: Type[], sub: Type[]): boolean {
+function containsPhrase(arr: string[], sub: string[]): boolean {
     for (let i = 0; i <= arr.length - sub.length; i++) {
         let j;
         for (j = 0; j < sub.length; j++) {
             if (j == sub.length - 1 && arr[i + j].length > sub[j].length) {
                 if (!arr[i + j].startsWith(sub[j])) break;
-                if (null !== arr[i + j].substring(sub[j].length).match(RegExp("^[!?.]*$", ""))) return true;
+                if (null !== RegExp("^[!?.]*$", "").exec(arr[i + j].substring(sub[j].length))) return true;
             }
             if (arr[i + j] !== sub[j]) break;
         }
@@ -24,7 +24,7 @@ function containsSubArray<Type>(arr: Type[], sub: Type[]): boolean {
 function matchesTriggerString(text: string, triggerString: string): boolean {
     const words = text.split(" ");
     const triggerWords = triggerString.split(" ");
-    return containsSubArray(words, triggerWords);
+    return containsPhrase(words, triggerWords);
 }
 
 function doesTextContainTriggerString(text: string, triggerStrings: string[]): boolean {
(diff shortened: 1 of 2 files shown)


evap/static/ts/src/text-answer-warnings.ts
```diff
@@ -6,8 +6,27 @@ function isTextMeaningless(text: string): boolean {
...
+        for(j = 0; j < sub.length; j++){
+            if(arr[i+j]!==sub[j]) return false;
+        }
+        return (j==sub.length);
+    }
+    return false;
+}
+
+function matchesTriggerString(text: string, triggerString: string): boolean{
+    console.log("Works");
```
hansegucker: Please drop debug log.

evap/static/ts/tests/unit/text-answer-warnings.ts
```diff
@@ -34,3 +34,12 @@ test.each([
...
 });
+
+test.each([
+  ["", "s.o.", false],
+  ["s.o.", "s.o.", true],
+  ["s.o.s", "s.o.", false]
+])('matchesTriggerString(%p, %p) should return %p', (a, b, expected) => {
+  const result = testable.matchesTriggerString(a, b);
+  expect(result).toBe(expected);
+});
```
hansegucker: Please add new line.

hansegucker (dismissed):

evap/static/ts/src/text-answer-warnings.ts
```diff
@@ -6,8 +6,26 @@ function isTextMeaningless(text: string): boolean {
     return text.length > 0 && ["", "ka", "na", "none", "keine", "keines", "keiner"].includes(text.replace(/\W/g, ""));
 }
 
+function containsSubArray(arr: string[], sub: string[]): boolean {
```
janno42: when the text contains another word before a trigger word, it doesn't correctly trigger. for example, when "weird" is a trigger word and the text answer, "weird" correctly triggers, but "it's weird" doesn't.

janno42 (changes requested):

evap/static/ts/src/text-answer-warnings.ts
AlexanderSchicktanz: It should now work properly and also has a dedicated test.

evap/static/ts/tests/unit/text-answer-warnings.ts
niklasmohrin, line 1: can you add test cases where the "trigger word" contains multiple words? because your function actually allows this :)

evap/static/ts/src/text-answer-warnings.ts
```diff
@@ -6,8 +6,25 @@ function isTextMeaningless(text: string): boolean {
...
+    for (let i = 0; i <= arr.length - sub.length; i++) {
+        let j;
+        for (j = 0; j < sub.length; j++) {
+            if (arr[i + j] !== sub[j]) break;
+        }
+        if (j == sub.length) return true;
```
niklasmohrin: I think this could be slightly nicer if you create another helper function `arrayEquals(a: string[]. b: string[]): bool` :)

Bonus points: Can you remove the `string` type and make a generic function instead? Take a look at our `utils.ts` file for examples, or ask if you need some input :)

janno42 (dismissed):
✅ functionality tested

evap/static/ts/src/text-answer-warnings.ts
```diff
@@ -6,8 +6,25 @@ function isTextMeaningless(text: string): boolean {
...
+    }
+    return false;
+}
+
+function matchesTriggerString(text: string, triggerString: string): boolean {
+    const words = text.split(" ");
+    const triggerWords = triggerString.split(" ");
```
niklasmohrin, line 26: Splitting by a single space character means that we don't match if users accidentally put two spaces. We should split by any amount of whitespace (like what happens in Python when using `str.split` without passing any argument)

evap/static/ts/src/text-answer-warnings.ts
```diff
@@ -6,8 +6,25 @@ function isTextMeaningless(text: string): boolean {
...
+    return containsSubArray(words, triggerWords);
```
niklasmohrin: This probably means that if users write something like "see above." at the end of a sentence we will not show a warning because the dot is included in the last word.

Doing this completely correct seems tough, unless we at some point change the UI so that staff users can enter arbitrary regexes as trigger words. What do you think we should aim for in this PR @janno42 ?

evap/static/ts/src/text-answer-warnings.ts
niklasmohrin: I still think that this change would make a nice improvement to the current implementation

niklasmohrin (changes requested):

evap/static/ts/src/text-answer-warnings.ts
richardebeling: We could first preprocess (before splitting) by regex-replacing all non-word characters with a delimiter, then splitting on the delimiter. Then, punctuation, arbitrarily weird (repeated) whitespace and other non-printable characters would not be an issue.

AlexanderSchicktanz: It now matches strings including dots, question- and exclamation-marks by trying to match the start of the string and then checking, if the rest is made up of only the aforementioned characters.
This way, trigger words including these delimiters (i.e. "s.o.") can also be matched.

evap/static/ts/src/text-answer-warnings.ts
```diff
@@ -6,8 +6,29 @@ function isTextMeaningless(text: string): boolean {
     return text.length > 0 && ["", "ka", "na", "none", "keine", "keines", "keiner"].includes(text.replace(/\W/g, ""));
 }
 
+function containsPhrase(arr: string[], sub: string[]): boolean {
```
niklasmohrin, line 9: This function is very difficult to read, it does too many things at once. We should probably have something like
```typescript
function matchesTriggerString(text: string, triggerString: string): boolean {
    const words = extractWords(text);
    const triggerWords = triggerString.split(" ");
    return isSubArray(triggerWords, words);
}
```

Then the isSubArray should also not be this complicated, it should use another function areArraysEqual, as I previously suggested

evap/static/ts/src/text-answer-warnings.ts
niklasmohrin: but there is still the problem that if there is more than one space character we have an empty string as one of the words

evap/static/ts/src/text-answer-warnings.ts
niklasmohrin: note that this also means that all of our custom "how do we handle weird text things" is separated from the entire "do the arrays match" logic

evap/static/ts/tests/unit/text-answer-warnings.ts
AlexanderSchicktanz: Some commit after you posted this added multiple words, so now it´s implemented.

janno42:
Hey @AlexanderSchicktanz, do you want to continue on this or should we assign someone else?

PR #2541, opened 2025-10-27
Add plainpage tag to evaluation creation and copy.

### LINKED ISSUE #2536: Use plain page for evaluation creation and copying
When editing an evaluation, the full width of the browser is used to provide sufficient space for the form and the log to appear next to each other (by using `"plain_page": True`).

However, when creating a new evaluation or copying an existing one, the form is too narrow because space is allotted for a log that does not yet exist, and the `plain_page` parameter is missing.

For these use cases, the form should be displayed as it is when editing, using the full browser width and leaving the log space empty.

LukasMansour:
Fixes #2536


janno42 (approved):
Works perfectly :)
Thanks!

PR #2542, opened 2025-10-27
Max value for participant slider on results page

### LINKED ISSUE #2485: Max value for participant slider on results page
On the results page, the participant filter slider currently allows values from 0 to the maximum number of participants across all published evaluations. However, this is not useful when some evaluations have more than 1,000 participants. The main use case for the slider is to find courses of different sizes, where "large" could mean more than 50 people and "very large" could mean 100+.

We should add a setting called `PARTICIPANT_FILTER_MAX_VALUE` to the settings (default value: 150) that defines the largest possible number that can be selected in the slider. This value should not be affected by the maximum number of participants across the published evaluations.
If this value is selected, it should be interpreted as an open end to the filter. This should be indicated by writing "150+" in the label instead of just "150".
For instance, choosing "100-149" would display all results with a minimum of 100 and a maximum of 149 participants. Selecting "100-150+" would show all results with at least 100 participants and would have no upper limit.

jooooosef:
implement #2485

### DIFF
--- a/evap/results/templates/results_index.html
+++ b/evap/results/templates/results_index.html
@@ -145,7 +145,7 @@
         import { ResultGrid } from "{% static 'js/datagrid.js' %}";
         import { RangeSlider } from "{% static 'js/slider.js' %}";
 
-        const participantFilterMaxValue = {{ participant_filter_max_value }};
+        const participantFilterMaxValue = Number("{{ results_participant_filter_max_value|escapejs }}");
 
         new ResultGrid({
             storageKey: "results-data-grid",
--- a/evap/results/templates/results_index.html
+++ b/evap/results/templates/results_index.html
@@ -145,6 +145,8 @@
         import { ResultGrid } from "{% static 'js/datagrid.js' %}";
         import { RangeSlider } from "{% static 'js/slider.js' %}";
 
+        const participantFilterMaxValue = {{ participant_filter_max_value }};
+
         new ResultGrid({
             storageKey: "results-data-grid",
             head: document.querySelector(".grid-header"),
@@ -166,7 +168,7 @@
             ]),
             filterSliders: new Map([
                 ["participants", {
-                    slider: new RangeSlider("participant-filter"),
+                    slider: new RangeSlider("participant-filter", participantFilterMaxValue),
                     selector: "[data-col=voters]",
                 }],
             ]),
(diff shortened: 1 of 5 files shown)


evap/settings.py
```diff
@@ -45,6 +45,9 @@
 RESULTS_WARNING_COUNT = 4
 RESULTS_WARNING_PERCENTAGE = 0.5
 
+# when this value is selected, the filter acts as an open end (no upper limit)
+PARTICIPANT_FILTER_MAX_VALUE = 150
```
niklasmohrin: We have to put at least something like "results" in the name, otherwise this setting sounds more important than it is; the comment could also use some more context

evap/static/ts/src/datagrid.ts
```diff
@@ -147,16 +147,20 @@ abstract class DataGrid {
...
-                    .get(name)
-                    ?.map(rawValue => parseFloat(rawValue))
-                    .some(rowValue => rowValue >= bound.low && rowValue <= bound.high),
-            );
+            const isDisplayedByRangeFilters = this.checkRangeFilters(row);
             row.isDisplayed = isDisplayedBySearch && isDisplayedByFilters && isDisplayedByRangeFilters;
         }
     }
 
+    protected checkRangeFilters(row: Row): boolean {
```
niklasmohrin: Check for what?

evap/static/ts/src/slider.ts
```diff
@@ -17,16 +17,18 @@ export class RangeSlider {
     private readonly rangeLabel: HTMLSpanElement;
     private allowed: Range = { low: 0, high: 0 };
     private _value: Range = { low: 0, high: 0 };
+    private configurableMaxValue?: number;
```
niklasmohrin: I don't think we need to put "configurable" in there. It is not configurable after we passed the value initially. And the possibility of its absence is already clear from the type.

Further, why is this a number at all? Shouldn't this number always equal `allowed.high` anyways? Then a boolean would suffice. Alternatively, can we avoid the new code in `checkRangeFilters` by setting `allowed.high` to infinity at the right time?

niklasmohrin (changes requested):

evap/results/templates/results_index.html
```diff
@@ -139,6 +139,8 @@ <h5 class="card-title mb-lg-0">
         import { ResultGrid } from "{% static 'js/datagrid.js' %}";
         import { RangeSlider } from "{% static 'js/slider.js' %}";
 
+        const participantFilterMaxValue = {{ participant_filter_max_value }};
```
niklasmohrin: There should be something like `escapejs` here, I am not sure right now which method is the right one, but there is definitely one :D

PR #2543, opened 2025-10-27
Notebook layouting compatibility with high DPI / widescreen displays #2098

### LINKED ISSUE #2098: Notebook layouting compatibility with high DPI / widescreen displays
The current notebook breakpoints used to determine whether the layout should be top/bottom or left/right doesn't consider the content width limit that the page otherwise has. On High-DPI or widescreen monitors, this can easily lead to cases where the top/bottom layout is used although the left/right layout would be more appropriate. 

This is what currently happens when you zoom out of the page far enough:
image

I think we might want to
* Limit the width of the notebook in top/bottom mode to the same width as the content of the page
* Use the limited content width for calculating whether the media breakpoint is hit or not (instead of the viewport width, which we currently use)

gitjannes:
fixes #2098 
- only has approximation to match width of notebook to evap content
- custom width breakpoint chosen by expert counseling


evap/static/scss/components/_notebook.scss
```diff
@@ -43,42 +43,44 @@
...
             min-height: 43.5vh; // approximation for size needed in smallest window (50vh - (spacer + leeway))
         }
+
+
+    @media screen and (min-aspect-ratio: $notebook-break), (min-width: 1400px) {//1400px chosen by expert judgement
+    .notebook-container {
+        width: 25vw;
+        padding-left: $spacer;
+        padding-bottom: $spacer;
+        padding-top: $spacer;
```
richardebeling: All three of these should be inherited from `padding: $spacer;` from `.notebook-container`, we should only need the `padding-right: 0;` line (or am I misremembering CSS)?

evap/static/scss/components/_notebook.scss
```diff
@@ -43,42 +43,44 @@
...
+        padding-right: 0;
+        z-index: none;
```
richardebeling: Is `none` a valid value for `z-index`? It's not listed on [the MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/z-index#values)...

evap/static/scss/components/_notebook.scss
```diff
@@ -43,42 +43,44 @@
...
     }
 }
 
-@media screen and (max-aspect-ratio: $notebook-break) {
+
     .notebook-margin {
         margin-top: 50vh;
     }
+
+@media screen and (min-aspect-ratio: $notebook-break), (min-width: 1400px) {
```
richardebeling: We use the 1400px in at least two places. Can you put it in a variable (next to and with similar naming as `notebook-break`)?

evap/static/scss/components/_notebook.scss
```diff
@@ -43,42 +43,44 @@
...
+        z-index: none;
     }
-}
 
-@media screen and (min-aspect-ratio: $notebook-break) {
-    .notebook-margin {
-        max-width: 75vw;
-        margin-left: 25vw;
+        .notebook-card {
+            height: calc((100vh - #{$navbar-brand-height}) - (#{$footer-height} + ((#{$spacer} * 1.5) * 5))); // subtract 3 spacer with padding
```
richardebeling: Comment does not match computation

I can visualize 100vh - navbar-height - footer-height, and I would understand subtracting 2 spacer heights, but the factors 1.5 and 5 are a bit unclear to me here

evap/static/tests/test_live.py
```diff
@@ -0,0 +1,49 @@
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support.expected_conditions import visibility_of_element_located
+
+from evap.evaluation.tests.tools import LiveServerTest
+
+
+class NoteBookPlacementLiveTest(LiveServerTest):
```
richardebeling:
```suggestion
class NotebookPlacementLiveTest(LiveServerTest):
```
(we consistently treat it as one word)

richardebeling: @niklasmohrin what's your opinion on these tests? So far, most of our frontend tests test/assert things that are stable, consistent user-visible truths ("There should be a button "Submit", click it, the page should go to XYZ, a success message with text "ABC" should be visible somewhere on the page"). Asserting stuff like "At 1400px browser window width, the element should have a width of 25vw" seems a bit too much like a meaningless/brittle implementation detail to me. It does not assert in a meaningful way that the page layout is sane, and it likely will fail exactly when people intentionally change the CSS.

I currently don't think this adds enough much value that it justifies the test runtime (how long does each of these tests take, 500ms?)

niklasmohrin: Yes, I agree, we are just replicating what is written in the CSS here. I guess we can only have a meaningful test here once we have #2046

richardebeling: Okay, then sorry @gitjannes for possibly wasted work here, but we do not want to merge these tests, please remove these from the PR

evap/static/scss/components/_notebook.scss
gitjannes: Tried the equation both with 2 spacer heights and the documented 3 spacer heights, it looked wrong. My "solution" now is to change the comment to an educated guess of the purpose.

evap/static/scss/components/_notebook.scss
```diff
@@ -43,42 +43,43 @@
...
 
-    @media screen and (min-aspect-ratio: $notebook-break) {
-        .notebook-container {
-            width: 25vw;
-            padding-left: $spacer;
-            padding-bottom: $spacer;
-            padding-top: $spacer;
-        }
+    display: flex;
+    justify-content: center;
```
richardebeling: I feel like we should put direct attributes at the top or bottom, not somewhere in the middle between nested selector blocks (or whatever they are called).

Why do we need these now, if we were happy without them before? This is required to center the top-bottom notebook, right?

@niklasmohrin @janno42 any guidance for "member" ordering here?

evap/static/scss/components/_notebook.scss
```diff
@@ -43,42 +43,43 @@
...
+    .notebook-container {
+        width: 80vw; //approximation to be similar to evap-Content width
```
richardebeling, line 49: Can we use the exact same value as we use with evap-content? I feel like there should be a way to factor that out into a variable and reuse it here.

evap/static/scss/components/_notebook.scss
```diff
@@ -43,42 +43,43 @@
...
+        z-index: $zindex-fixed;
```
richardebeling: We only change sizes, if I understand correctly. Why is a change to z-index necessary?

evap/static/scss/components/_notebook.scss
```diff
@@ -43,42 +43,43 @@
...
-    }
 
-    @media screen and (max-aspect-ratio: $notebook-break) {
-        .notebook-container {
-            width: 100vw;
-            z-index: $zindex-fixed;
-            padding: $spacer;
-        }
+
+    @media screen and (min-aspect-ratio: $notebook-break), (min-width: $notebook-break-width) {
```
richardebeling, line 60: You inverted the rule logic here. Before, we had the "non-selective" filters match the left-right layout, and then had a more selective ruleset (via media query) that changed it into top-down. Why is it necessary to invert this, what do we gain from that?

evap/static/scss/components/_notebook.scss
richardebeling: I just now saw that this line was already in the CSS and you just moved it -- GitHub just happens to show this as an added line first. It would be nice to come up with a good explanation of why this makes sense, or finding a better value if it doesn't, but it's not required from you here.

evap/static/scss/components/_notebook.scss
niklasmohrin: intuitively I would also say that nested blocks should come only after all declarations for the parent, but I will take whatever makes the most natural diff for this PR; currently it looks like things were moved around and I would appreciate if stuff would stay mostly where it is :D

janno42: I would expect to find these attributes at the top

gitjannes: > Why do we need these now, if we were happy without them before? This is required to center the top-bottom notebook, right?

This centering is necessary to emulate the look of the previous not width-limited top-bottom notebook, since it used to fill out the screen and was centered implicitly.

richardebeling: > necessary to emulate the look of the previous not width-limited top-bottom notebook

okay, but then it should be fine to have them inside the corresponding css media selector, right? I feel like structuring the CSS here like that could possibly help having a good overview of both modes -- but following an approach of getting a minimal diff is probably also fine, but if you need top-level CSS styles, please put them at the top.

evap/static/scss/components/_notebook.scss
gitjannes: //approximation for similar gap to the footer as evap-Content
is my current understanding of the formula

evap/static/scss/components/_notebook.scss
gitjannes: My understanding is that there were two checks for both cases previously, and having both checks while introducing the fixed upper limit for the width introduced some weird behaviour when at the border of the checks. This is why I removed one of the checks and  made the top-bottom view the default in the code, since the theoretical first instances of the screen at 1x1 px would the top-bottom view.

evap/static/scss/components/_notebook.scss
gitjannes: The exact value would be the same as the one set by the "container" class, adding this to the notebook would lead to the need to suppress a lot of the other properties introduced by that class for the left-right notebook view.

richardebeling: I was thinking of putting the "80vh" value into a variable that is then used by `container` and `notebook-container`, just like we have done with `$notebook-break` etc. No need to add extra CSS classes then, we just want to move out the shared magic number into a shared variable

PR #2547, opened 2025-11-03
Introduce exam types

### LINKED ISSUE #2500: JSON enrollment data importer (v2)
*Edits:*
*- 2025-08-15: Added "language" which defines the main language of an evaluation*
*- 2025-10-20: lecturers don't have "callingname"*
*- 2025-11-30: exam evaluations should keep their imported name*

----

_building on #2131_

-- The EXAM UPDATE --

- Skip certain course types
  - [x] `CourseType` should get a new boolean attribute `skip_on_automated_import` (default: False)
  - [x] This attribute must be available on the `staff/course_types` page.
  - [x] When importing enrollment data via the JSON importer, all courses with a type having this set to `True` should be skipped. A warning should be shown in the log for each course skipped in this way.
- Use `callingname`
  - [x] Instead of the attribute `christianname`, use `callingname` to populate `first_name_given` for students; for lecturers `christianname` is used.
- Exam types
  - [x] A new model `ExamType` should be added, similar to `CourseType`, having the same attributes. The exam type is linked to an `Evaluation` via a new attribute `exam_type` and can be null (default).
  - [x] `Course.has_exam_evaluation` should be redefined: Any evaluation with an exam type other than null should now be considered an exam evaluation.
  - [x] When importing an event with `"isexam": true`, use the event's `type` attribute to add the respective exam type to the evaluation.
  - [x] The names of the exam evaluation should be set to the part of the event's name after the last ` - `. If there is none, the names should be set to the respective exam types' names (de and en).
  - [x] In the staff and grades views, the exam type should be shown in the list of labels ("course_badges.html"), instead of the currently shown course type.
  - [x] When bulk assigning questionnaires, exam evaluations should be handled separately from non-exam evaluations. The logic for non-exam evaluations doesn't change. For exam evaluations, the assignment should now be based on the exam type instead of the course type.
- Rewards
  - [x] Set `is_rewarded` to `False` when importing exam evaluations. 
- Edit rights for contributors
  - [x] All contributors added during the import should have the role `EDITOR`.
- Use `language`
  - [x] The attribute `language` of an event should be used to initialize the `main_language` of the respective evaluations of the course ("Deutsch" -> German; "Englisch" -> English; any other value (including emtpy) -> undecided)

hansegucker:
Close #2500

### DIFF
--- a/evap/staff/tests/test_json_importer.py
+++ b/evap/staff/tests/test_json_importer.py
@@ -797,6 +797,7 @@ class TestImportEvents(TestCase):
         self.assertEqual(importer._disambiguate_name("Klausur", ["Klausur (-1)"]), "Klausur")
         # doesn't fail on arbitary strings
         self.assertEqual(importer._disambiguate_name("Klausur", ["Klausur (mündlich)"]), "Klausur")
+        self.assertEqual(importer._disambiguate_name("Klausur", ["Klausur 2"]), "Klausur")
 
     def test_exam_type_existing(self):
         exam_type = ExamType.objects.create(name_en="Presentation", name_de="Präsentation")


evap/staff/importers/json.py
```diff
@@ -179,6 +195,16 @@ def __init__(self, semester: Semester, default_course_end: date) -> None:
         self.courses_by_gguid: dict[str, Course] = {}
         self.statistics = ImportStatistics()
 
+    def _disambiguate_name(self, wanted_name: str, current_names: list[str]):
+        if not current_names:
+            return wanted_name
+        matches = [re.match(rf"^{wanted_name}(?:\s+\((\d+)\))?$", name) for name in current_names]
```
richardebeling: optional group (`?:`) and optional count operator `?` -- only one should be required?

evap/staff/importers/json.py
```diff
@@ -201,6 +227,16 @@ def _get_course_type(self, name: str) -> CourseType:
...
         return course_type
 
+    def _get_exam_type(self, name: str) -> ExamType:
+        lookup = name.strip().lower()
+        if lookup in self.exam_type_cache:
+            return self.exam_type_cache[lookup]
+
+        # It could happen that the importer needs a new exam type
+        exam_type, __ = ExamType.objects.get_or_create(name_de=name, defaults={"name_en": name})
+        self.exam_type_cache[name] = exam_type
```
richardebeling:
```suggestion
        self.exam_type_cache[lookup] = exam_type
```

evap/staff/importers/json.py
```diff
@@ -346,15 +382,27 @@ def _import_evaluation(self, course: Course, data: ImportEvent) -> Evaluation:
...
-            name_de = data["title"].split(" - ")[-1] if " - " in data["title"] else "Prüfung"
-            name_en = data["title_en"].split(" - ")[-1] if " - " in data["title_en"] else "Exam"
+            exam_type = self._get_exam_type(data["type"])
+            if not evaluation:
+                name_de = self._disambiguate_name(
+                    exam_type.name_de, list(course.evaluations.values_list("name_de", flat=True))
+                )
+                name_en = self._disambiguate_name(
+                    exam_type.name_en, list(course.evaluations.values_list("name_en", flat=True))
+                )
```
richardebeling, line 444: This degrades to O(n^2) operations. Maybe fine for now, I think we have superlinear runtime in other parts here as well?

evap/staff/templates/staff_exam_type_index.html
```diff
@@ -0,0 +1,89 @@
...
+            <div class="card-body table-responsive">
+                <table id="exam_types_table" class="table table-vertically-aligned">
+                    <thead>
+                        <tr>
+                            <th class="movable"></th>
+                            <th class="width-percent-30">{% translate 'Name (German)' %}</th>
+                            <th class="width-percent-30">{% translate 'Name (English)' %}</th>
+                            <th class="width-percent-30">{% translate 'Import names' %}</th>
+                            <th class="width-percent-30">{% translate 'Skip on automated import' %}</th>
+                            <th class="width-percent-10 text-end">{% translate 'Actions' %}</th>
```
richardebeling: Table is now 130% wide

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -182,10 +182,17 @@
                         {% endblocktranslate %}
                     </span>
                     <div slot="extra-inputs">
-                        <label>
-                            {% translate 'Exam Date:' %}
-                            <input type="date" name="exam_date" min="{{ evaluation.earliest_possible_exam_date }}" class="form-control" required form="exam_creation_form_{{ evaluation.id }}"/>
-                        </label>
+                        {% include 'bootstrap_form_field.html' with field=evaluation.create_exam_evaluation_form.base_evaluation %}
+
+                        <!-- Exam Date rendered manually due to bug with Bootstrap datepicker -->
```
richardebeling: Can we use a django template comment (`{# #}` syntax)?

evap/staff/tests/test_json_importer.py
```diff
@@ -699,3 +703,37 @@ def test_management_command(self, mock_import_json):
...
                 call_command("json_import", self.semester.id + 42, test_filename, "01.01.2000", stdout=output)
+
+    def test_disambiguate_name(self):
+        importer = JSONImporter(self.semester, date(2000, 1, 1))
+
+        self.assertEqual(importer._disambiguate_name("Klausur", []), "Klausur")
+        self.assertEqual(importer._disambiguate_name("Klausur", [""]), "Klausur")
+        self.assertEqual(importer._disambiguate_name("Klausur", ["Klausur"]), "Klausur (2)")
+        self.assertEqual(importer._disambiguate_name("Klausur", ["Klausur", "Klausur (2)"]), "Klausur (3)")
+        self.assertEqual(importer._disambiguate_name("Klausur", ["Klausur", "Klausur (3)"]), "Klausur (4)")
```
richardebeling, line 794: interesting edge case: What if `Klausur (0)` or `Klausur (-1)` already exist? (I think we're fine with the behavior, but it would be nice to have it codified here

evap/staff/views.py
```diff
@@ -2179,6 +2178,21 @@ def course_type_merge(request, main_type_id, other_type_id):
...
+    model = ExamType
+    formset_class = modelformset_factory(
+        ExamType,
+        form=ExamTypeForm,
+        formset=ModelWithImportNamesFormset,
+        can_delete=True,
+        extra=1,
+    )
+    template_name = "staff_exam_type_index.html"
+    success_url = reverse_lazy("staff:exam_type_index")
```
richardebeling, line 2198: Why reverse_lazy? (If there is a reason, maybe add a comment?)

richardebeling (changes requested):

evap/staff/views.py
niklasmohrin: Isn't this always used for class based views because the urls module imports the views ? In that case I wouldn't comment

richardebeling: Ah, that makes sense, thanks

evap/staff/importers/json.py
hansegucker: Subtle difference: `?:` is not an optional group, but a non-capturing group.

evap/staff/templates/staff_exam_type_index.html
hansegucker: Solved in other PR

evap/staff/tests/test_views.py
```diff
@@ -949,18 +950,29 @@ def setUpTestData(cls):
...
         )
+        cls.exam_evaluations = [
+            baker.make(
+                Evaluation,
+                course=reg_evaluation.course,
+                exam_type=exam_type,
+                name_de=exam_type.name_de,
+                name_en=exam_type.name_en,
+            )
+            for reg_evaluation, exam_type in zip(cls.evaluations, cls.exam_types, strict=True)
```
richardebeling: Maybe `main_evaluation` instead of `reg_evaluation` (which I assume is short for "regular")?

evap/staff/urls.py
```diff
@@ -70,6 +70,8 @@
     path("course_types/merge", views.course_type_merge_selection, name="course_type_merge_selection"),
     path("course_types/<int:main_type_id>/merge/<int:other_type_id>", views.course_type_merge, name="course_type_merge"),
 
+    path("exam_types/", views.ExamTypeIndexView.as_view(), name="exam_type_index"),
```
niklasmohrin, line 73: for consistency with the others: 
```suggestion
    path("exam_type/", views.ExamTypeIndexView.as_view(), name="exam_type_index"),
```

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -182,10 +182,17 @@
                         {% endblocktranslate %}
                     </span>
                     <div slot="extra-inputs">
-                        <label>
-                            {% translate 'Exam Date:' %}
-                            <input type="date" name="exam_date" min="{{ evaluation.earliest_possible_exam_date }}" class="form-control" required form="exam_creation_form_{{ evaluation.id }}"/>
-                        </label>
+                        {% include 'bootstrap_form_field.html' with field=evaluation.create_exam_evaluation_form.base_evaluation %}
+
+                        {# Exam Date rendered manually due to bug with Bootstrap datepicker #}
```
richardebeling: which bug exactly, can we leave some trace that allows to reconstruct in the future whether this is still needed?

evap/development/fixtures/test_data.json
niklasmohrin, line 1: can we have some non-null exam types in the test data so that it is easy to test related functionality on the site?

evap/evaluation/migrations/0158_examtype.py
```diff
@@ -0,0 +1,61 @@
...
+import django.contrib.postgres.fields
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+def _migrate(apps, _schema_editor):
+    Evaluation = apps.get_model("evaluation", "Evaluation")
+    ExamType = apps.get_model("evaluation", "ExamType")
+    exam_type = ExamType.objects.create(name_de="Klausur", name_en="Exam")
+    Evaluation.objects.filter(models.Q(name_de="Klausur") | models.Q(name_en="Exam")).update(exam_type=exam_type)
```
richardebeling, line 12: Please do this for a few evaluations in `test_data.json`. Currently, every evaluation has exam type null.

evap/evaluation/models.py
```diff
@@ -319,6 +319,31 @@ def can_be_deleted_by_manager(self):
...
+    order = models.IntegerField(verbose_name=_("exam type order"), default=-1)
+
+    class Meta:
+        ordering = ["order"]
+
+    def __str__(self):
+        return self.name
+
+    def can_be_deleted_by_manager(self):
+        if not self.pk:
```
niklasmohrin: hm, can this break if the id is AutoField starting at 0?

answer doesn't matter, let's use `is None` so that we don't have to find out

evap/evaluation/models.py
```diff
@@ -319,6 +319,31 @@ def can_be_deleted_by_manager(self):
         return not self.courses.all().exists()
 
 
+class ExamType(models.Model):
+    """Model for the exam type of evaluation, e.g. a written exam"""
```
niklasmohrin, line 323:
```suggestion
    """Different kinds of exams, for example written or oral."""
```

evap/evaluation/models.py
```diff
@@ -319,6 +319,31 @@ def can_be_deleted_by_manager(self):
...
+
+    order = models.IntegerField(verbose_name=_("exam type order"), default=-1)
+
+    class Meta:
+        ordering = ["order"]
+
+    def __str__(self):
+        return self.name
+
+    def can_be_deleted_by_manager(self):
```
```suggestion
    def can_be_deleted_by_manager(self) -> bool:
```

evap/evaluation/templates/evaluation_badges.html
```diff
@@ -4,7 +4,11 @@
     {% endfor %}
 {% endif %}
 {% if not mode == 'subentry' %}
-    <span class="badge bg-secondary badge-course-type">{{ evaluation.course.type }}</span>
+    {% if evaluation.exam_type %}
+        <span class="badge bg-secondary badge-course-type">{{ evaluation.exam_type }}</span>
```
niklasmohrin:
```suggestion
        <span class="badge bg-secondary badge-exam-type">{{ evaluation.exam_type }}</span>
```
and then we should maybe make the two classes look slightly different? @janno42 you probably have a good intuition for color here

edit: ah actually this isn't even styled separately 🤔 what it _is_ used for is in the result grid, which is now probably wrong with the change here

evap/evaluation/models.py
```diff
@@ -501,21 +530,28 @@ class State(models.IntegerChoices):
...
     def has_exam_evaluation(self):
-        return self.course.evaluations.filter(Q(name_de="Klausur") | Q(name_en="Exam")).exists()
+        return self.course.evaluations.filter(exam_type__isnull=False).exists()
 
     @property
     def earliest_possible_exam_date(self):
         return self.vote_start_datetime.date() + timedelta(days=1)
 
+    @property
+    def create_exam_evaluation_form(self):
```
niklasmohrin: type hint

evap/staff/urls.py
hansegucker: I don't think that this will increase consistency with `programs` or `course_types`.

evap/evaluation/migrations/0158_examtype.py
hansegucker: There are no exams in the evaluations yet. @janno42 wanted to update the test data afterwards.

evap/staff/importers/json.py
janno42:
```suggestion
                    message=f"Event has an unknown language ("{data['language']}"), main language was set to undecided",
```

evap/evaluation/migrations/0158_examtype.py
janno42: For now it would still be nice to have a few in the testdata already :)

evap/evaluation/templates/evaluation_badges.html
janno42:
```suggestion
        <span class="badge bg-light badge-course-type">{{ evaluation.exam_type }}</span>
```

janno42: For styling

janno42: The current implementation seems okay for now - exam evaluations usually don't exist on their own. If they are a subentry along another evaluation, the subentries don't have badges and the course entry correctly shows the course type and no exam type.

evap/staff/templates/staff_exam_type_index.html
janno42: incorrect rebase?

evap/staff/templates/staff_semester_questionnaire_assign_form.html
```diff
@@ -27,6 +27,16 @@ <h4 class="card-title">{% translate 'General' %}</h4>
                                 </fieldset>
                             </div>
                         </div>
+                    <div class="card">
```
janno42: wrong indentation

evap/staff/templates/staff_semester_questionnaire_assign_form.html
```diff
@@ -13,11 +13,11 @@ <h3>{% translate 'Assign questionnaires' %}</h3>
         {% csrf_token %}
         <div class="card mb-3">
             <div class="card-body">
-                <p>{% translate 'Select the questionnaires to be used for each course type. This will only change evaluations in preparation. Their questionnaires will be overwritten by this operation. Selecting nothing will not change the currently used questionnaires.' %}</p>
+                <p>{% translate 'Select the questionnaires to be used for each course/exam type. This will only change evaluations in preparation. Their questionnaires will be overwritten by this operation. Selecting nothing will not change the currently used questionnaires.' %}</p>
                 {% include 'bootstrap_form_errors.html' with errors=form.non_field_errors %}
                 <div class="row">
                     <div class="col-md-6">
-                        <div class="card">
+                        <div class="card mb-2">
```
janno42: i would prefer having `mt-2` on the second card instead

evap/staff/templates/staff_semester_questionnaire_assign_form.html
```diff
@@ -27,6 +27,16 @@ <h4 class="card-title">{% translate 'General' %}</h4>
                                 </fieldset>
                             </div>
                         </div>
+                    <div class="card">
+                            <div class="card-body">
+                                <h4 class="card-title">{% translate 'Exam' %}</h4>
+                                <fieldset>
+                                    {% for field in exam_fields %}
+                                        {% include 'bootstrap_form_field.html' with field=field wide=True%}
```
janno42:
```suggestion
                                        {% include 'bootstrap_form_field.html' with field=field wide=True %}
```

evap/staff/templates/staff_semester_view_evaluation.html
janno42: seems to be unresolved @hansegucker ?

evap/staff/tests/test_json_importer.py
```diff
@@ -774,6 +778,43 @@ def test_management_command(self, mock_import_json):
...
+    def test_exam_type_existing(self):
+        exam_type = ExamType.objects.create(name_en="Presentation", name_de="Präsentation")
+        data = deepcopy(EXAMPLE_DATA)
+        data["events"][1]["type"] = "Präsentation"
+
+        self._import(data)
+
+        exam_evaluation = Evaluation.objects.get(cms_id=data["events"][1]["gguid"])
+
+        assert exam_evaluation.name_de == exam_type.name_de
```
janno42: why not `self.assertEqual`?

evap/staff/urls.py
niklasmohrin: okay fair

evap/staff/templates/staff_semester_view_evaluation.html
hansegucker: It just won't show up – I am not sure how to describe this further.

evap/evaluation/migrations/0158_examtype.py
hansegucker: Ok, then I just create same random exams and dump the test data afterwards?

evap/evaluation/templates/evaluation_badges.html
niklasmohrin: But do we actually want to treat exam types like course types here? I don't see that doing it like this yields any benefit over just putting badge-exam-type when both are not used anyways. With the current code, it seems like a bug waiting to happen

Maybe I am not getting it, but if there is a deeper reason, we should also add a comment to explain it. Otherwise, this will be hard to track down again later

janno42: I'm not sure I'm getting it either 😅
Normal evaluations should have course type badges, exam evaluations should have exam type badges. They should differ in both styling and class names that can be used wherever else.

janno42:
Here I'd probably squash everything together. It's really just one (larger) batch of changes all belonging together. None of it really makes sense on its own, does it?

evap/evaluation/templates/evaluation_badges.html
niklasmohrin: Ah, okay, I thought you meant that we should leave it like the reviewed version where both badges have the course type class.

niklasmohrin: Ah, okay, I thought you meant that we should leave it like the reviewed version where both badges have the course type class.

evap/staff/importers/json.py
```diff
@@ -364,15 +399,31 @@ def _import_evaluation(self, course: Course, data: ImportEvent) -> Evaluation:
...
+                name_de = self._clean_whitespaces(
+                    self._disambiguate_name(
+                        exam_type.name_de, list(course.evaluations.values_list("name_de", flat=True))
+                    )
+                )
+                name_en = self._clean_whitespaces(
+                    self._disambiguate_name(
+                        exam_type.name_en, list(course.evaluations.values_list("name_en", flat=True))
+                    )
+                )
```
janno42, line 444: Testing lead to another small change we should include (I updated the issue accordingly):
- The names of the exam evaluation should be set to the part of the event's name after the last ` - ` (like before). If there is none, the names should be set to the respective exam types' names (de and en).

evap/evaluation/models.py
```diff
@@ -319,6 +322,31 @@ def can_be_deleted_by_manager(self):
...
+    )
+    skip_on_automated_import = models.BooleanField(verbose_name=_("skip on automated import"), default=False)
+
+    order = models.IntegerField(verbose_name=_("exam type order"), default=-1)
+
+    class Meta:
+        ordering = ["order"]
+
+    def __str__(self):
+        return self.name
```
niklasmohrin, line 339: Where is this used? I think it is probably better to use `.name` directly there

evap/evaluation/models.py
```diff
@@ -501,21 +533,28 @@ class State(models.IntegerChoices):
...
     def has_exam_evaluation(self):
-        return self.course.evaluations.filter(Q(name_de="Klausur") | Q(name_en="Exam")).exists()
+        return self.course.evaluations.filter(exam_type__isnull=False).exists()
 
     @property
     def earliest_possible_exam_date(self):
         return self.vote_start_datetime.date() + timedelta(days=1)
 
+    @property
+    def create_exam_evaluation_form(self) -> "ExamEvaluationForm":
```
niklasmohrin: Let's not pollute the evaluation model with the staff form logic here; what do you think about making a template tag in the staff app instead?

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -182,10 +182,17 @@
                         {% endblocktranslate %}
                     </span>
                     <div slot="extra-inputs">
-                        <label>
-                            {% translate 'Exam Date:' %}
-                            <input type="date" name="exam_date" min="{{ evaluation.earliest_possible_exam_date }}" class="form-control" required form="exam_creation_form_{{ evaluation.id }}"/>
-                        </label>
+                        {% include 'bootstrap_form_field.html' with field=evaluation.create_exam_evaluation_form.base_evaluation %}
```
niklasmohrin: Currently we call `create_exam_evaluation_form` twice; can we use one `{% with ...` which calls this?

evap/staff/forms.py
```diff
@@ -517,6 +545,34 @@ def __init__(self, data=None, instance=None):
...
+            self.fields["base_evaluation"].initial = evaluation
+
+    def clean(self):
+        cleaned_data = super().clean()
+        if cleaned_data["base_evaluation"].has_exam_evaluation:
+            raise ValidationError(_("An exam evaluation already exists for this course."))
+        if cleaned_data["exam_date"] < cleaned_data["base_evaluation"].earliest_possible_exam_date:
+            raise ValidationError(
+                _("The end date of the main evaluation would be before its start date. No exam evaluation was created.")
+            )
```
niklasmohrin, line 574:
```suggestion
            )
        return cleaned_data
```

evap/staff/views.py
```diff
@@ -902,11 +905,16 @@ def semester_questionnaire_assign(request, semester_id):
...
+    form = QuestionnairesAssignForm(request.POST or None, course_types=course_types, exam_types=exam_types)
 
     if form.is_valid():
         for evaluation in evaluations:
-            general_questionnaires = list(form.cleaned_data[f"general-{evaluation.course.type.id}"])
+            general_questionnaires = (
+                list(form.cleaned_data[f"general-{evaluation.course.type.id}"])
+                if evaluation.exam_type is None
+                else list(form.cleaned_data[f"exam-{evaluation.exam_type.id}"])
+            )
```
niklasmohrin, line 917: Not clear from the issue, but should contributor questionnaires still be assigned in the exam evaluations? My gut feeling would be _no_. The form fields for the contributors are still accessed through the course type, which also seems odd. @janno42 ?

evap/staff/templates/staff_exam_type_index.html
```diff
@@ -0,0 +1,92 @@
...
+
+{% block breadcrumb %}
+    {{ block.super }}
+    <li class="breadcrumb-item">{% translate 'Exam types' %}</li>
+{% endblock %}
+
+{% block content %}
+    {{ block.super }}
+
+    <form id="exam-type-form" method="POST" class="form-horizontal tomselectform">
```
niklasmohrin, line 13:
```suggestion
    <form method="POST" class="form-horizontal tomselectform">
```

evap/staff/templates/staff_exam_type_index.html
```diff
@@ -0,0 +1,92 @@
...
+        {% csrf_token %}
+        {{ formset.management_form }}
+
+        <div class="card mb-3">
+            <div class="card-body table-responsive">
+                <table id="exam_types_table" class="table table-vertically-aligned">
+                    <thead>
+                        <tr>
```
niklasmohrin, line 21: widths are not adding up to 100 because last column doesn't have a th

evap/staff/importers/json.py
```diff
@@ -189,6 +204,16 @@ def _clean_whitespaces(text: str) -> str:
         # Use regex for cleaning to also include non-ASCII whitespaces like non-breaking whitespaces
         return re.sub(r"\s+", " ", text.strip())
 
+    def _disambiguate_name(self, wanted_name: str, current_names: list[str]):
```
```suggestion
    def _disambiguate_name(self, wanted_name: str, current_names: list[str]) -> str:
```

evap/staff/importers/json.py
```diff
@@ -189,6 +204,16 @@ def _clean_whitespaces(text: str) -> str:
...
+        if not current_names:
+            return wanted_name
+        matches = [re.match(rf"^{wanted_name}(?:\s+\((\d+)\))?$", name) for name in current_names]
```
niklasmohrin: evap developers when CMS data contains the exam type `[test]\s+adminswereherelolol`: 

Image

niklasmohrin: (so I would prefer string manipulation with startswith, removeprefix, lstrip)

evap/staff/importers/json.py
```diff
@@ -212,6 +237,16 @@ def _get_course_type(self, name: str) -> CourseType:
...
         return course_type
 
+    def _get_exam_type(self, name: str) -> ExamType:
+        lookup = name.strip().lower()
+        if lookup in self.exam_type_cache:
+            return self.exam_type_cache[lookup]
+
+        # It could happen that the importer needs a new exam type
+        exam_type, __ = ExamType.objects.get_or_create(name_de=name, defaults={"name_en": name})
+        self.exam_type_cache[lookup] = exam_type
```
niklasmohrin, line 267: side question: am I missing something, or is it that the cache will not be populated with this lookup the next time unless we also add it as an import name?

evap/staff/importers/json.py
```diff
@@ -410,27 +461,32 @@ def _import_evaluation(self, course: Course, data: ImportEvent) -> Evaluation:
...
-            defaults=defaults,
-        )
+
+        created = not evaluation
+        if not evaluation:
+            evaluation = Evaluation.objects.create(
+                course=course,
+                cms_id=data["gguid"],
+                name_de=name_de,
+                name_en=name_en,
```
niklasmohrin, line 505: Why are these not in `defaults`? Do we not want to log in some case? Please add a comment :)

evap/staff/tests/test_json_importer.py
```diff
@@ -552,11 +556,11 @@ def test_import_courses_special_cases(self):
 
         # use second part of title after dash
```
niklasmohrin: comment out of date?

evap/staff/tests/test_json_importer.py
```diff
@@ -671,29 +675,29 @@ def test_import_courses_evaluation_approved(self):
...
-        evaluation.name_en = "Test"
+        evaluation.is_rewarded = False
         evaluation.save()
 
         importer = self._import()
 
         evaluation = Evaluation.objects.get(pk=evaluation.pk)
 
-        self.assertEqual(evaluation.name_en, "Test")
+        self.assertFalse(evaluation.is_rewarded)
```
niklasmohrin, line 706: side question: could we even use assertNoDatabaseModifications?

evap/staff/tests/test_json_importer.py
```diff
@@ -774,6 +778,43 @@ def test_management_command(self, mock_import_json):
...
+        importer = JSONImporter(self.semester, date(2000, 1, 1))
+
+        self.assertEqual(importer._disambiguate_name("Klausur", []), "Klausur")
+        self.assertEqual(importer._disambiguate_name("Klausur", [""]), "Klausur")
+        self.assertEqual(importer._disambiguate_name("Klausur", ["Klausur"]), "Klausur (2)")
+        self.assertEqual(importer._disambiguate_name("Klausur", ["Klausur", "Klausur (2)"]), "Klausur (3)")
+        self.assertEqual(importer._disambiguate_name("Klausur", ["Klausur", "Klausur (3)"]), "Klausur (4)")
+        self.assertEqual(importer._disambiguate_name("Klausur", ["Klausur (0)"]), "Klausur (1)")
+        # doesn't match on negative numbers
+        self.assertEqual(importer._disambiguate_name("Klausur", ["Klausur (-1)"]), "Klausur")
```
niklasmohrin, line 797: can you also add one with `Klausur (mündlich)` to ensure the `int(...)` call doesn't raise out of the function?

evap/staff/views.py
janno42: Oh, that's right. Contributors in exam evaluations should not get questionnaires assigned via the bulk assignment.

evap/evaluation/models.py
hansegucker: To implement `__str__` is best practice: https://docs.djangoproject.com/en/5.2/ref/models/instances/#django.db.models.Model.__str__

evap/staff/tests/test_json_importer.py
hansegucker: No, there are still some valid effects.

richardebeling: Which ones? If you run the test under the assert, it should be pretty obvious what exactly was changed -- my naive assumption here would be that nothing changes in this import?

evap/evaluation/models.py
niklasmohrin: okay sure

evap/staff/importers/json.py
richardebeling: (there's also https://docs.python.org/3/library/re.html#re.escape, which should safely and consistently solve this, if string manipulation is much more verbose)

evap/staff/templates/staff_exam_type_index.html
```diff
@@ -0,0 +1,93 @@
...
+                                    {% endfor %}
+                                    {% include 'bootstrap_form_field_widget.html' with field=form.name_de %}
+                                </td>
+                                <td>
+                                    {% include 'bootstrap_form_field_widget.html' with field=form.name_en %}
+                                </td>
+                                <td>
+                                    {% include 'bootstrap_form_field_widget.html' with field=form.import_names %}
+                                </td>
+                                 <td>
```
richardebeling: indentation

evap/staff/tests/test_json_importer.py
hansegucker: A log entry is created.

hansegucker:
```
AssertionError: Unexpected modifying query found: INSERT INTO "evaluation_logentry" ("content_type_id", "content_object_id", "attached_to_object_type_id", "attached_to_object_id", "datetime", "user_id", "action_type", "request_id", "data") VALUES (17, 115, 17, 115, '2025-12-01 20:44:15.381554'::timestamp, NULL, 'change', '', '{"programs": {"add": []}}'::jsonb) RETURNING "evaluation_logentry"."id"
```

evap/evaluation/templates/evaluation_badges.html
```diff
@@ -4,7 +4,11 @@
     {% endfor %}
 {% endif %}
 {% if not mode == 'subentry' %}
-    <span class="badge bg-secondary badge-course-type">{{ evaluation.course.type }}</span>
+    {% if evaluation.exam_type %}
+        <span class="badge bg-light badge-course-type">{{ evaluation.exam_type }}</span>
```
janno42:
```suggestion
        <span class="badge bg-light badge-exam-type">{{ evaluation.exam_type }}</span>
```

evap/staff/tests/test_json_importer.py
richardebeling: But we generally only create log entry when some other change happens, so there must be some actual underlying change that happens, right?

hansegucker: Resolved in favor of #2576

evap/staff/importers/json.py
```diff
@@ -189,6 +202,33 @@ def _clean_whitespaces(text: str) -> str:
...
+            return 1
+        if not (name.startswith("(") and name.endswith(")")):
+            return None
+        name = name.removeprefix("(").removesuffix(")")
+        try:
+            number = int(name)
+            if number < 0:
+                return None
+            return number
+        except ValueError:
```
niklasmohrin, line 219: nit: you can move this right after the first `number` assignment and have everything else afterwards

evap/staff/importers/json.py
```diff
@@ -212,6 +252,19 @@ def _get_course_type(self, name: str) -> CourseType:
...
+        name = self._clean_whitespaces(name)
+        if name in self.exam_type_cache:
+            return self.exam_type_cache[name]
+
+        # It could happen that the importer needs a new exam type
+        exam_type, __ = ExamType.objects.get_or_create(name_de=name, defaults={"name_en": name, "import_names": [name]})
+        if name not in exam_type.import_names:
+            exam_type.import_names.append(name)
+            exam_type.save()
+        self.exam_type_cache[name] = exam_type
```
niklasmohrin: bug: should use `.lower` like in normal cache initialization; same thing in the lookup at the top of the function; let's use the `lookup` pattern like before

evap/staff/importers/json.py
```diff
@@ -364,15 +417,29 @@ def _import_evaluation(self, course: Course, data: ImportEvent) -> Evaluation:
...
             evaluation_start_datetime = course_end.replace(hour=8, minute=0, second=0, microsecond=0) + timedelta(
                 days=1
             )
             evaluation_end_date = (course_end + settings.EXAM_EVALUATION_DEFAULT_DURATION).date()
 
-            name_de = data["title"].split(" - ")[-1] if " - " in data["title"] else "Prüfung"
-            name_en = data["title_en"].split(" - ")[-1] if " - " in data["title_en"] else "Exam"
+            exam_type = self._get_exam_type(data["type"])
+            if not evaluation:
+                name_de = data["title"].split(" - ")[-1] if " - " in data["title"] else exam_type.name_de
```
niklasmohrin, line 437: do we ever want to `split` more than once here? otherwise I would see this as
```suggestion
                name_de = data["title"].partition(" - ")[2] or exam_type.name_de
```

evap/staff/importers/json.py
```diff
@@ -364,15 +417,29 @@ def _import_evaluation(self, course: Course, data: ImportEvent) -> Evaluation:
...
+                name_en = data["title_en"].split(" - ")[-1] if " - " in data["title_en"] else exam_type.name_en
+                name_de = self._clean_whitespaces(
```
niklasmohrin: why is it `clean(disambiguate(name))` and not `disambiguate(clean(name))`? Don't we expect the existing names to be cleaned?

evap/staff/forms.py
```diff
@@ -517,6 +545,35 @@ def __init__(self, data=None, instance=None):
...
+        required=True,
+        localize=True,
+    )
+    exam_type = forms.ModelChoiceField(ExamType.objects.all(), required=True, label=_("Exam Type"))
+
+    def __init__(self, *args, evaluation=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        if evaluation:
+            for field in ["base_evaluation", "exam_date", "exam_type"]:
+                self.fields[field].widget.attrs["form"] = f"exam_creation_form_{evaluation.id}"
```
niklasmohrin: if you are doing one final round of changes anyways, can you also include the refactor we discussed with `form_id` being an optional parameter of ExamEvaluationForm? :D

evap/staff/importers/json.py
```diff
@@ -189,6 +202,33 @@ def _clean_whitespaces(text: str) -> str:
...
+            number = int(name)
+            if number < 0:
+                return None
+            return number
+        except ValueError:
+            return None
+
+    def _disambiguate_name(self, wanted_name: str, current_names: list[str]) -> str:
+        if not current_names:
+            return wanted_name
```
niklasmohrin: not needed because the `not real_matches` branch below catches this
```suggestion
```

niklasmohrin: this also allows you to weaken the type of current_names to Iterable and remove the conversion to `list` in the caller - if you want

evap/staff/importers/json.py
niklasmohrin: note that this also uses the exam type name if for some reason the name ends after ` - `, however it might still emit ` ` as the name; we can also add one `.strip` if we want to handle this, but it's not really important for merging

evap/staff/importers/json.py
niklasmohrin: (at this point, we should consider adding a helper class for the three caches because we will keep messing this up whenever we change something; stuff for a next issue though)

evap/staff/importers/json.py
niklasmohrin: @richardebeling or do you forbid the `str or str`? :D

richardebeling: No that's fine for me

evap/staff/importers/json.py
niklasmohrin: so on second look, the getter here is technically correct because the key scheme for the exam type cache is just different than the course type and program caches and doesn't use lowercase names. however, there is no rationale for why the two would differ in the code, so if there really is a good reason, we must at least add a comment. it seems to me like all caches should behave the same though and we should just return to the `.lower` convention for exam types too

evap/staff/importers/json.py
```diff
@@ -212,6 +252,20 @@ def _get_course_type(self, name: str) -> CourseType:
         self.course_type_cache[lookup] = course_type
         return course_type
 
+    def _get_exam_type(self, name: str) -> ExamType:
+        name = self._clean_whitespaces(name)
+        lookup = name.lower()
+        if name in self.exam_type_cache:
+            return self.exam_type_cache[name]
```
richardebeling: These two should use `lookup`, not `name`, right?

evap/staff/forms.py
```diff
@@ -517,6 +545,35 @@ def __init__(self, data=None, instance=None):
...
+    exam_type = forms.ModelChoiceField(ExamType.objects.all(), required=True, label=_("Exam Type"))
+
+    def __init__(self, *args, evaluation=None, form_id=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        if form_id:
+            for field in ["base_evaluation", "exam_date", "exam_type"]:
+                self.fields[field].widget.attrs["form"] = form_id
+            self.fields["exam_date"].widget.attrs["min"] = evaluation.earliest_possible_exam_date
+            self.fields["exam_type"].initial = ExamType.objects.order_by("order").first()
+            self.fields["base_evaluation"].initial = evaluation
```
niklasmohrin, line 564: Why do we only set this if there is a form id?
```suggestion
        self.fields["exam_date"].widget.attrs["min"] = evaluation.earliest_possible_exam_date
        self.fields["exam_type"].initial = ExamType.objects.order_by("order").first()
        self.fields["base_evaluation"].initial = evaluation
```

niklasmohrin, line 564: `if evaluation` is missing

evap/staff/forms.py
```diff
@@ -517,6 +545,35 @@ def __init__(self, data=None, instance=None):
...
+    exam_date = forms.DateField(
+        label=_("Exam Date"),
+        required=True,
+        localize=True,
+    )
+    exam_type = forms.ModelChoiceField(ExamType.objects.all(), required=True, label=_("Exam Type"))
+
+    def __init__(self, *args, evaluation=None, form_id=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        if form_id:
```
niklasmohrin: also, both for evaluation and form id prefer `is not None` :)

evap/staff/importers/json.py
```diff
@@ -212,6 +253,20 @@ def _get_course_type(self, name: str) -> CourseType:
         self.course_type_cache[lookup] = course_type
         return course_type
 
+    def _get_exam_type(self, name: str) -> ExamType:
+        name = self._clean_whitespaces(name)
+        lookup = name.lower()
+        if name in self.exam_type_cache:
```
```suggestion
        if lookup in self.exam_type_cache:
```

evap/staff/importers/json.py
```diff
@@ -212,6 +253,20 @@ def _get_course_type(self, name: str) -> CourseType:
...
+            return self.exam_type_cache[lookup]
+
+        # It could happen that the importer needs a new exam type
+        exam_type, __ = ExamType.objects.get_or_create(name_de=name, defaults={"name_en": name, "import_names": [name]})
+        if lookup not in exam_type.import_names:
+            exam_type.import_names.append(name)
```
```suggestion
        if name not in exam_type.import_names:
            exam_type.import_names.append(name)
```

evap/staff/importers/json.py
```diff
@@ -176,6 +187,11 @@ def __init__(self, semester: Semester, default_course_end: date) -> None:
             for course_type in CourseType.objects.all()
             for import_name in course_type.import_names
         }
+        self.exam_type_cache: dict[str, ExamType] = {
+            import_name.lower(): exam_type
```
richardebeling:
```suggestion
            import_name.strip().lower(): exam_type
```

evap/staff/forms.py
```diff
@@ -517,6 +545,36 @@ def __init__(self, data=None, instance=None):
...
+        required=True,
+        localize=True,
+    )
+    exam_type = forms.ModelChoiceField(ExamType.objects.all(), required=True, label=_("Exam Type"))
+
+    def __init__(self, *args, evaluation=None, form_id=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        if form_id is not None:
+            for field in self.fields.keys():
+                self.fields[field].widget.attrs["form"] = form_id
```
richardebeling: could we iterate over values, since we only use the key for a lookup in the dict again?

evap/staff/importers/json.py
```diff
@@ -189,6 +205,31 @@ def _clean_whitespaces(text: str) -> str:
...
         return re.sub(r"\s+", " ", text.strip())
 
+    def _extract_number_in_name(self, name: str, wanted_name: str) -> int | None:
+        if not name.startswith(wanted_name):
+            return None
+        name = name.removeprefix(wanted_name).strip()
+        if not name:
+            return 1
+        if not (name.startswith("(") and name.endswith(")")):
+            return None
```
niklasmohrin, line 215: branch not covered

evap/staff/forms.py
```diff
@@ -215,6 +216,31 @@ def save(self, *args, **kwargs):
         return course_type
 
 
+class ExamTypeForm(forms.ModelForm):
```
niklasmohrin, line 219: this form is not covered, we should add a short test class for the corresponding view (also the can_be_deleted_by_manager case)

evap/staff/importers/json.py
niklasmohrin: so we noticed that there sometimes _are_ multiple dashes, and we do want to only get the last part; in the future, we should add a test for that

PR #2549, opened 2025-11-03
Introduce settings for exam evaluation weights and duration

### LINKED ISSUE #2456: Follow-up from JSON import: Do not use hard-coded weights
In general, it's not super awesome that the knowledge about weights and timedeltas for exam evaluations are duplicated here; okay to fix in a follow up

_Originally posted by @niklasmohrin in https://github.com/e-valuation/EvaP/pull/2187#discussion_r2127464083_

ZoqkMaze:
fixes #2456

### DIFF
--- a/evap/staff/tests/test_json_importer.py
+++ b/evap/staff/tests/test_json_importer.py
@@ -319,6 +319,7 @@ class TestImportEvents(TestCase):
         importer.import_json(data)
         return importer
 
+    @override_settings(EXAM_EVALUATION_DEFAULT_DURATION=timedelta(days=3))
     def test_import_courses(self):
         importer = self._import()
 
@@ -366,7 +367,7 @@ class TestImportEvents(TestCase):
         self.assertEqual(exam_evaluation.name_en, "Exam")
         # [{"begin": "29.07.2024 10:15", "end": "29.07.2024 11:45"}]
         self.assertEqual(exam_evaluation.vote_start_datetime, datetime(2024, 7, 30, 8, 0))
-        self.assertEqual(exam_evaluation.vote_end_date, date(2024, 7, 29) + settings.EXAM_EVALUATION_DEFAULT_DURATION)
+        self.assertEqual(exam_evaluation.vote_end_date, date(2024, 8, 1))
         self.assertEqual(
             set(exam_evaluation.participants.values_list("email", flat=True)),
             {"1@example.com", "2@example.com"},


niklasmohrin:
@ZoqkMaze can you modify a test to check that these settings are used?

ZoqkMaze:
I included the timedelta for the exam. The test will follow

evap/settings.py
```diff
@@ -56,6 +56,10 @@
 GENERAL_GRADE_QUESTIONS_WEIGHT = 1
 # the average grade of all general non-grade questions is weighted this much for the evaluation's average grade
 GENERAL_NON_GRADE_QUESTIONS_WEIGHT = 1
+# default weight for exam evaluations
+EXAM_EVALUATION_WEIGHT = 1
+# default weight for evaluations with exam
+NORMAL_EVALUATION_WEIGHT = 9
```
niklasmohrin:
```suggestion
# default weight for exam evaluations
EXAM_EVALUATION_DEFAULT_WEIGHT = 1
# default weight for evaluations with exam
MAIN_EVALUATION_DEFAULT_WEIGHT = 9
```

evap/settings.py
```diff
@@ -110,18 +114,18 @@
 # Amount of hours in which participant will be warned
 EVALUATION_END_WARNING_PERIOD = 5
 
+# default timedelta for exam evaluation vote_end_date after exam date
```
niklasmohrin, line 117: soooo, it's the duration right? can we factor that into the name?

evap/staff/tests/test_json_importer.py
```diff
@@ -391,7 +366,7 @@ def test_import_courses(self):
         self.assertEqual(exam_evaluation.name_en, "Exam")
         # [{"begin": "29.07.2024 10:15", "end": "29.07.2024 11:45"}]
         self.assertEqual(exam_evaluation.vote_start_datetime, datetime(2024, 7, 30, 8, 0))
-        self.assertEqual(exam_evaluation.vote_end_date, date(2024, 8, 1))
+        self.assertEqual(exam_evaluation.vote_end_date, date(2024, 7, 29) + settings.EXAM_EVALUATION_TIMEDELTA)
```
niklasmohrin: this is just repeating the actual implementation; we should leave the computed date in and instead fix the setting value using `override_settings`

evap/staff/importers/json.py
```diff
@@ -361,10 +355,11 @@ def _import_evaluation(self, course: Course, data: ImportEvent) -> Evaluation:
             )
             # End date is on the sunday in the week the event ends
             evaluation_end_date = (course_end + timedelta(days=6 - course_end.weekday())).date()
+            # Ende unterschiedlich zu Evaluation class: hier auch nicht abhängig von exam TODO
```
```suggestion
```

evap/evaluation/models.py
```diff
@@ -508,17 +508,17 @@ def earliest_possible_exam_date(self):
...
         exam_evaluation = Evaluation(
             course=self.course,
             name_de="Klausur",
             name_en="Exam",
-            weight=1,
+            weight=settings.EXAM_EVALUATION_WEIGHT,
             is_rewarded=False,
             vote_start_datetime=datetime.combine(exam_date + timedelta(days=1), time(8, 0)),
-            vote_end_date=exam_date + timedelta(days=3),
+            vote_end_date=exam_date + settings.EXAM_EVALUATION_TIMEDELTA,
```
niklasmohrin: actually, maybe this is a case where having the full timedelta as a setting does not make sense and we should instead expose the number of days as a setting; any timedelta that is not a full number of days will be rounded down anyways. @richardebeling ?

evap/staff/tests/test_json_importer.py
niklasmohrin: this also partly applies to the other tests, but it doesn't stick out so much there

evap/settings.py
ZoqkMaze: maybe EXAM_EVALUATION_DEFAULT_DURATION?

evap/evaluation/models.py
richardebeling: Both options are fine with me. If we have some similar setting already, I would re-use what we have there.

We could add startup validation of setting values if we want to keep it a timedelta but only want to allow full days. I wouldn't force that into this PR here.

niklasmohrin: okay, then we can stick to timedelta

niklasmohrin (approved):

PR #2556, opened 2025-11-10
Fix error when previewing evaluations with undecided main language

### LINKED ISSUE #2555: Evaluation preview fails with unselected main language
When trying to preview an evaluation which does not have a selected main language yet, the server throws a KeyError. Instead, we should render the page using the default language (English).

Naturally, a test should be added.

Jakobus400:
Fixes #2555

Preview of evaluation form failed with undecided language. Fixed by checking in render_vote_page()

### DIFF
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -128,6 +128,7 @@ class EvaluationForm(forms.ModelForm):
         main_language = self.cleaned_data.get("main_language")
         if main_language == Evaluation.UNDECIDED_MAIN_LANGUAGE:
             self.add_error("main_language", _("You have to set a main language for this evaluation."))
+
         return main_language
 
     def save(self, *args, **kw):
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -128,7 +128,6 @@ class EvaluationForm(forms.ModelForm):
         main_language = self.cleaned_data.get("main_language")
         if main_language == Evaluation.UNDECIDED_MAIN_LANGUAGE:
             self.add_error("main_language", _("You have to set a main language for this evaluation."))
-
         return main_language
 
     def save(self, *args, **kw):
--- a/evap/staff/tests/test_views.py
+++ b/evap/staff/tests/test_views.py
@@ -2509,15 +2509,14 @@ class TestEvaluationPreviewView(WebTestStaffModeWith200Check):
         cls.url2 = reverse("staff:evaluation_preview", args=[cls.evaluation_un.pk])
         cls.evaluation_un.general_contribution.questionnaires.set([baker.make(Questionnaire)])
 
-
     def test_without_questionnaires_assigned(self):
         # regression test for #1747
         self.evaluation.general_contribution.questionnaires.set([])
         self.app.get(self.url, user=self.manager, status=200)
 
     def test_lang_undecided(self):
-        response = self.app.get(self.url2, user=self.manager, status=200)
-        
+        self.app.get(self.url2, user=self.manager, status=200)
+
 
 class TestEvaluationImportPersonsView(WebTestStaffMode):
     @classmethod
--- a/evap/staff/tests/test_views.py
+++ b/evap/staff/tests/test_views.py
@@ -2510,11 +2510,11 @@ class TestEvaluationPreviewView(WebTestStaffModeWith200Check):
         self.evaluation.general_contribution.questionnaires.set([])
         self.app.get(self.url, user=self.manager, status=200)
 
-    def test_lang_undecided(self):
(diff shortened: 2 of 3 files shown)


evap/contributor/forms.py
```diff
@@ -117,6 +117,7 @@ def clean_main_language(self):
         main_language = self.cleaned_data.get("main_language")
         if main_language == Evaluation.UNDECIDED_MAIN_LANGUAGE:
             self.add_error("main_language", _("You have to set a main language for this evaluation."))
+
```
niklasmohrin: please revert this unrelated change :)

evap/student/views.py
```diff
@@ -253,6 +253,8 @@ def render_vote_page(
     for_rendering_in_modal: bool = False,
 ) -> HttpResponse:
     language = request.GET.get("language", evaluation.main_language)
+    if language == "x":
+        language = "en"
```
niklasmohrin: two thoughts:
1. we do not expect that `request.GET["language"]` would ever be "x", we only expect it for `evaluation.main_language`; can you change it so that only this case is checked?
2. please don't use "x" directly, instead use the `Evaluation.UNDECIDED_MAIN_LANGUAGE` constant

niklasmohrin (commented):
we also still need a test that checks that accessing the page works :)

evap/student/views.py
```diff
@@ -252,6 +252,8 @@ def render_vote_page(
     dropout: bool,
     for_rendering_in_modal: bool = False,
 ) -> HttpResponse:
+    if evaluation.main_language == evaluation.UNDECIDED_MAIN_LANGUAGE:
+        evaluation.main_language = "en"
```
niklasmohrin: Hm, this is not quite right and makes me afraid that we display (or even save!) this wrongly somewhere later - a caller using `render_vote_page` likely does not expect that the evaluation gets changed in here.

What do you think about going a bit more defensive and defining an extra variable `fallback_language` which is `evaluation.main_language` or `"en"` if the main language is not decided. Then we use this variable in `request.GET.get("language", fallback_language)`. How do you feel about this change?

niklasmohrin (changes requested):
code looks good, but still needs tests

evap/staff/tests/test_views.py
```diff
@@ -2465,11 +2465,18 @@ def setUpTestData(cls):
...
+        cls.evaluation_un.general_contribution.questionnaires.set([baker.make(Questionnaire)])
+
     def test_without_questionnaires_assigned(self):
         # regression test for #1747
         self.evaluation.general_contribution.questionnaires.set([])
         self.app.get(self.url, user=self.manager, status=200)
 
+    def test_lang_undecided(self):
+        self.app.get(self.url2, user=self.manager, status=200)
+
```
richardebeling: We generally try to only use `setUpTestData` for a generic setup that would likely be reused across multiple test methods. For edge-case tests like this one, I think the likelihood that another test will want to reuse a "evaluation with undecided main language" is pretty low, so I would just move the test setup into the one test method using it, to keep the scope smaller


```suggestion
    def test_without_questionnaires_assigned(self):
        # regression test for #1747
        self.evaluation.general_contribution.questionnaires.set([])
        self.app.get(self.url, user=self.manager, status=200)

    def test_lang_undecided(self):
        evaluation = baker.make(Evaluation, main_language=Evaluation.UNDECIDED_MAIN_LANGUAGE)
        self.app.get(reverse("staff:evaluation_preview", args=[evaluation.pk]), user=self.manager, status=200)

```

richardebeling (approved):
thanks

evap/student/views.py
```diff
@@ -252,7 +252,11 @@ def render_vote_page(
     dropout: bool,
     for_rendering_in_modal: bool = False,
 ) -> HttpResponse:
-    language = request.GET.get("language", evaluation.main_language)
+    fallback_language = (
+        evaluation.main_language if evaluation.main_language != evaluation.UNDECIDED_MAIN_LANGUAGE else "en"
```
niklasmohrin:
```suggestion
        evaluation.main_language if evaluation.main_language != Evaluation.UNDECIDED_MAIN_LANGUAGE else "en"
```

niklasmohrin (approved):
I agree with Richard. Other than that, I have one small suggestion. Thanks!

niklasmohrin:
Ah, our linter now complains that the function has too many local variables. I think for this PR it is fine to disable this lint in this function. You can put a `pylint: disable=too-many-local-variables` comment in the function (there are other examples of this in the code base and you can run `./manage.py lint pylint` to check whether it worked)

evap/staff/tests/test_views.py
richardebeling: (the current version still sets these instances as attributes on the test class instance -- please only make it local variables)

richardebeling (approved):

PR #2558, opened 2025-11-17
Fixed alignment of messages on plain_page htmls #2553

### LINKED ISSUE #2553: Message placement in plain_page layout
Messages are not placed correctly on pages when `plain_page` is used:
Image

The message should appear at the top of the content, rather than on the side. Ideally, the base template should allow such pages to specify the placement of a new block "messages" on their own, so that, in this case, all messages could be placed in the log area on the right.

The example shown in the screenshot can be replicated by creating a new course and saving it using the "Save and create evaluation" button.

HannesClausnitzer:
fixes #2553

### DIFF
--- a/evap/evaluation/templates/messages.html
+++ b/evap/evaluation/templates/messages.html
@@ -1,12 +1,8 @@
 {% load evaluation_filters %}
 
-{% if messages %}
-	{% if styles is not None %}<div class={{ styles }}>{% endif %}
-		{% for message in messages %}
-			<div class="alert alert-{{ message.tags|message_class }} alert-dismissible">
-				{{ message }}
-				<button type="button" class="btn-close" data-bs-dismiss="alert"></button>
-			</div>
-		{% endfor %}
-	{% if styles is not None %}</div>{% endif %}
-{% endif %}
+{% for message in messages %}
+	<div class="alert alert-{{ message.tags|message_class }} alert-dismissible">
+		{{ message }}
+		<button type="button" class="btn-close" data-bs-dismiss="alert"></button>
+	</div>
+{% endfor %}
--- a/evap/staff/templates/staff_evaluation_form.html
+++ b/evap/staff/templates/staff_evaluation_form.html
@@ -6,7 +6,11 @@
     {{ block.super }}
     <div class="row flex-grow-1 overflow-hidden">
         <div class="col-12 col-xl-8 mh-100 overflow-auto">
-            {% include 'messages.html' with styles="pt-3" %}
+            {% if messages %}
+	            <div class="pt-3">
+                    {% include 'messages.html' %}
+                </div>
+            {% endif %}
             <h3 class="pt-2">
                 {% if evaluation_form.instance.id %}
                     {{ evaluation.full_name }} ({{ evaluation.course.semester.name }})


evap/evaluation/templates/base.html
```diff
@@ -53,12 +53,9 @@
...
         <div{% if not plain_page %} class="p-3 mb-4"{% endif %} id="evapContent">
             <div class="{% if plain_page %}container-fluid d-flex max-content-height{% else %}container{% endif %}">
-                {% for message in messages %}
-                    <div class="alert alert-{{ message.tags|message_class }} alert-dismissible">
-                        {{ message }}
-                        <button type="button" class="btn-close" data-bs-dismiss="alert"></button>
-                    </div>
-                {% endfor %}
+                {% if not plain_page %}
+                    {% include 'messages.html' %}
```
niklasmohrin, line 57: @richardebeling @janno42 How will we ensure that we don't forget to include this file if we ever introduce a new plain_page template? Do we just hope that we will remember?

evap/evaluation/templates/messages.html
```diff
@@ -0,0 +1,12 @@
+{% load evaluation_filters %}
+
+{% if messages %}
+	<div{% if plain_page %} class="pt-3"{% endif %}>
```
niklasmohrin: Can we keep plain_page out of files other than base.html? Maybe we can have a `wrapper-classes` argument and use it like this in staff_evaluation_form: `include "messages.html" with wrapper-class="pt-3"`. Or is that too complicated? @richardebeling

janno42: we should add the `pt-3` in a new div around the messages in staff_evaluation_form instead. then this file doesn't need to handle it at all.

evap/evaluation/templates/base.html
janno42: yes, i think it's fair to say that plain_pages have to handle it themselves and we have to remember it

evap/evaluation/templates/messages.html
niklasmohrin: but the padding shouldn't be there is there are no messages to display, or does it not matter?

HannesClausnitzer: Yes, I think the padding should only be visible if there are messages. Could you clarify how I can avoid using plain_page? I thought it's necessary because the padding should also only be displayed on plain_page htmls.

evap/evaluation/templates/base.html
richardebeling: A page "forgetting" to show messages should not be an issue, Django will keep any pending messages queued until we render a template that iterates `messages`, see https://docs.djangoproject.com/en/5.2/ref/contrib/messages/#displaying-messages.

I think it makes sense to not show messages, the intention of a plain page is to not contain fluff, messages are fluff.

evap/evaluation/templates/messages.html
richardebeling: The proposal @niklasmohrin made sounds good to me: When including `messages.html` from another template, we can pass some "context" to this template. This context can include what Niklas called a "wrapper-class" above, the template can then use that  in the div here as
```html
<div class="{{ wrapper_class }}">
```

See https://docs.djangoproject.com/en/5.2/ref/templates/builtins/#include for more details

evap/evaluation/templates/messages.html
```diff
@@ -0,0 +1,12 @@
+{% load evaluation_filters %}
+
+{% if messages %}
+	{% if styles is not None %}<div class={{ styles }}>{% endif %}
```
niklasmohrin: First: let's put quotes around the `styles` variable so that it doesn't break if the styles string contains spaces

Second: Why do we not create the div if styles is None? This seems counterintuitive

evap/evaluation/templates/messages.html
janno42: what's the reason why you prefer this over staff_evaluation_form conditionally putting a div with padding around the messages block if there are messages?

niklasmohrin: that would also work for me

niklasmohrin (approved):

PR #2562, opened 2025-11-17
add language selection in questionnaire preview

### LINKED ISSUE #2552: Language selection in questionnaire preview
Currently, the staff questionnaire preview page always shows the questionnaire in the language of the page.
Similar to the voting page, there should be language selection buttons at the top that change only the language of the questionnaire, not the entire page.
The language initially displayed when the page opens should always be the page's language.

ZoqkMaze:
fix #2552

### DIFF
--- a/evap/staff/templates/staff_questionnaire_view.html
+++ b/evap/staff/templates/staff_questionnaire_view.html
@@ -8,7 +8,7 @@
 {% block content %}
     {{ block.super }}
     <h3 class="mb-3">
-        <div class=" mt-3 mt-lg-0 d-flex justify-content-end">
+        <div class="mt-3 mt-lg-0 d-flex justify-content-end">
             <div class="btn-switch btn-switch-light btn-switch-evaluation-language">
                 <div class="btn-switch-label my-auto">
                     {% translate "Questionnaire language" %}


ZoqkMaze:
Currently, the questionnaire language tooltip reads"This only changes the language of the questionnaire and not the whole page. Changing this will not delete your existing answers." Should this be changed?

evap/staff/templates/staff_questionnaire_view.html
```diff
@@ -7,9 +7,35 @@
...
+                            class="btn btn-sm btn-light{% if evaluation_language == language_code %} active{% endif %}">
+                            <span id="span-set-evaluation-language-{{ language_code }}">{{ language_code|upper }}</span>
+                        </a>
+                    {% endfor %}
+                </div>
+            </div>
+        </div>
+    </h3>
+    {% language evaluation_language %}
     <div class="card mb-3">
```
niklasmohrin: can you indent the contents of the `language` block? optimally, do this in a separate commit so that we can easily see this in the review :)

evap/staff/tests/test_views.py
```diff
@@ -3315,6 +3316,17 @@ def setUpTestData(cls):
             allows_additional_textanswers=False,
         )
 
+    def test_preview_change_language(self):
+        user = self.test_users[0]
+        user.language = "de"
+        user.save()
+        page = self.app.get(url=self.url, user=user, status=200)
+        self.assertIn(self.question.text_de, page)
```
niklasmohrin, line 3324: can you add the corresponding `assertNotIn` checks with the text in the other language?

evap/staff/views.py
```diff
@@ -1827,13 +1828,24 @@ def questionnaire_index(request):
 
 @manager_required
 def questionnaire_view(request, questionnaire_id):
+    language = request.GET.get("language", request.user.language)
     questionnaire = get_object_or_404(Questionnaire, id=questionnaire_id)
 
     # build forms
     contribution = Contribution(contributor=request.user)
-    form = QuestionnaireVotingForm(request.POST or None, contribution=contribution, questionnaire=questionnaire)
+    with translation.override(language):
```
niklasmohrin, line 1836: Why do we need this? I suspect that we have some non-lazy translation somewhere and thus run into errors here otherwise, but did you actually observe this?

evap/staff/templates/staff_questionnaire_view.html
```diff
@@ -7,9 +7,35 @@
 
 {% block content %}
     {{ block.super }}
+    <h3 class="mb-3">
+        <div class=" mt-3 mt-lg-0 d-flex justify-content-end">
+            <div class="btn-switch btn-switch-light btn-switch-evaluation-language">
+                <div class="btn-switch-label my-auto">
+                    {% translate "Questionnaire language" %}
+                    <span class="fas fa-circle-info" data-bs-toggle="tooltip" title="{% translate 'This only changes the language of the questionnaire and not the whole page. Changing this will not delete your existing answers.' %}"></span>
```
janno42:
```suggestion
```

we don't need the info icon and tooltip in the preview

evap/staff/views.py
janno42: yes, the questions in the voting form are otherwise not translated, only the voting buttons

niklasmohrin: right, I was looking in the wrong place

evap/staff/templates/staff_questionnaire_view.html
```diff
@@ -7,9 +7,34 @@
 
 {% block content %}
     {{ block.super }}
-    <div class="card mb-3">
-        <div class="card-body">
-            {% include 'student_vote_questionnaire_group.html' with questionnaire_group=forms preview=True %}
+    <h3 class="mb-3">
+        <div class=" mt-3 mt-lg-0 d-flex justify-content-end">
```
janno42:
```suggestion
        <div class="mt-3 mt-lg-0 d-flex justify-content-end">
```

PR #2566, opened 2025-11-24
Frontend regression testing

fekoch:
This PR adds a subclass of `LiveServerTest`: `VisualRegressionTestCase`. This allows to write test cases like `StaffSemesterViewRegressionTest`, which are standard live server tests where you can use `self.trigger_screenshot(<identifier>)`.
The screenshot is then compared to a known good screenshot and the testcase fails, if there is a large enough difference (as specified by us) between the two.

The CI needs to provide three environment variables for the test to establish a connection to testing api:

 - `VRT_APIURL`
 - `VRT_APIKEY`
 - `VRT_PROJECT`


evap/evaluation/tests/tools.py
```diff
@@ -7,13 +9,15 @@
 
 import django.test
 import django_webtest
+import requests
```
niklasmohrin, line 12: we also need to declare this dependency

evap/grades/tests/test_live.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class GradesViewTest(VisualRegressionTestCase):
+
+    @freeze_time("2025-10-27")
+    @override_settings(SLOGANS_EN=["Einigermaßen verlässlich aussehende Pixeltestung"])
+    def test_grades_semester_view(self):
+        baker.seed(31902)
+
+        semester = baker.make(Semester)
```
niklasmohrin: should we do freeze_time and baker.seed in the base class set up already? feels like something we would have to always do anyways

evap/evaluation/tests/tools.py
```diff
@@ -341,6 +346,91 @@ def setUpClass(cls) -> None:
...
+
+    @property
+    def viewport(self):
+        return f"{self.window_size[0]}x{self.window_size[1]}"
+
+    def trigger_screenshot(self, name: str):
+        full_name = self.__class__.__name__ + "_" + name
+
+        config = {
+            "apiUrl": os.environ.get("VRT_APIURL"),
```
niklasmohrin: intuitively this should happen in `VisualRegressionTestCase.__init__`; any reason against it?

evap/evaluation/tests/tools.py
```diff
@@ -341,6 +346,91 @@ def setUpClass(cls) -> None:
...
+            "ciBuildId": os.environ.get("VRT_CIBUILDID"),
+            "branchName": os.environ.get("VRT_BRANCHNAME"),
+            "apiKey": os.environ.get("VRT_APIKEY"),
+            "projectId": os.environ.get("VRT_PROJECT"),
+        }
+
+        headers = {
+            "apiKey": config.get("apiKey"),
```
niklasmohrin: and all of these config.get calls should not use `.get` - these attributes should be set at this point

evap/evaluation/tests/tools.py
```diff
@@ -341,6 +346,91 @@ def setUpClass(cls) -> None:
...
+            "Content-Type": "application/json",
+        }
+
+        data = {
+            "project": config.get("projectId"),
+            "branchName": config.get("branchName"),
+            "ciBuildId": config.get("ciBuildId"),
+        }
+
+        registration_response = requests.post(
```
niklasmohrin, line 419: can we also do this in the initialization? (not `__init__` maybe, but setUp or so?)

evap/evaluation/tests/tools.py
```diff
@@ -341,6 +346,91 @@ def setUpClass(cls) -> None:
...
+
+        review_url = test_response.json().get("url", "<url-not-found>")
+        switcher = {
+            "new": f"No Baseline! Review manually: {review_url}",
+            "unresolved": f"Difference found: {review_url}",
+        }
+
+        error_message = switcher.get(test_response.json().get("status"))
+
+        _ = requests.patch(
```
niklasmohrin: what are we doing here? does it reset the state to what was the baseline before the test? if so, this should be more robust with try-finally or similar

evap/grades/tests/test_live.py
```diff
@@ -0,0 +1,35 @@
...
+            courses = baker.make(Course, semester=semester, _quantity=30)
+            _ = [
+                baker.make(
+                    Evaluation,
+                    course=course,
+                    wait_for_grade_upload_before_publishing=True,
+                    state=Evaluation.State.IN_EVALUATION,
+                )
+                for course in courses
+            ]
```
richardebeling: You can remove all the `_ =` assignments here and in the other files -- ignored return values can simply be ignored.

For the baker.make call here, you can pass `course=iter(courses)` and `_quantity=len(courses)` instead of looping N times

evap/evaluation/tests/tools.py
```diff
@@ -341,6 +346,92 @@ def setUpClass(cls) -> None:
...
 
+@tag("vrt")
+class VisualRegressionTestCase(LiveServerTest):
+    window_size = (1920, 1080)
+    _http_timeout_seconds = 3
+
+    def setUp(self) -> None:
+        with connection.cursor() as cursor:
+            # reset userprofile id sequence to keep test reproducible
+            cursor.execute("ALTER SEQUENCE evaluation_userprofile_id_seq restart")
```
richardebeling: How does the userprofile ID affect screen shots (and how does that differ from sequence numbers of other models)? I would have expected that this isn't visible anywhere on the page

evap/staff/tests/test_live.py
```diff
@@ -174,3 +183,66 @@ def test_collapse_without_editor_approved(self) -> None:
...
+            ),
+            vote_start_datetime=datetime(2099, 1, 1, 0, 0),
+            vote_end_date=date(2099, 12, 31),
+            main_language="en",
+        )
+        _ = baker.make(
+            Evaluation,
+            course=baker.make(
+                Course, semester=evaluation.course.semester, programs=[baker.make(Program)], responsibles=[responsible]
+            ),
```
richardebeling: you can pass `course__semester=evaluation.course.semester` and `course__responsibles=[responsible]` instead of nesting baker.make calls.

(repeated above/below)

evap/evaluation/tests/tools.py
richardebeling: Out of curiosity, roughly how painful would it be to do this without `requests`? If it's just 10-ish more lines of python with `urllib`, I'm tempted to use that instead

niklasmohrin: On the other hand, we have requests in our uv.lock either way through mozilla-django-oidc; I don't have a strong opinion towards either direction

richardebeling: Ah if we already install it anyway, fine with me

(although mozilla-django-oidc is a candidate for removal as soon as possible :D)

evap/evaluation/tests/tools.py
fekoch: The default implementation of `UserProfile.full_name` returns `<unnamed> (User 2)` (where 2 is the id).

But thinking about this again, I guess it would be nicer to just set the name to a sample string.

evap/evaluation/tests/tools.py
fekoch: The python standard library documentation explicitly recommends using the Requests package: https://docs.python.org/3/library/http.client.html

Considering that, I think it is fair to explicitly depend on it.

evap/evaluation/tests/tools.py
richardebeling: Yeah let's make baker just fill the name then :+1:

niklasmohrin:
Closing for now, can be reopened once work on this continues

PR #2572, opened 2025-12-01
Add `--headed` flag for live tests

### LINKED ISSUE #2567: Add `--no-headless` flag for live tests
When developing and consequently debugging live frontend tests, we often want to view what the browser is doing. This can be done by _not_ running it in headless mode, which we currently always do. To show the browser, I currently advise contributors to change the `headless = True` line in the definition of `LiveServerTest`. Instead, I would like to advise them to run the test with `--no-headless`. Someone should add this option, so that I don't have to lie.

@richardebeling I guess the easiest option would be to provide our own test command that inherits from the default one. Alternatively, we could use an env var for this issue instead of a flag. What do you think?

Belissimo-T:
fix #2567

### DIFF
--- a/evap/evaluation/tests/tools.py
+++ b/evap/evaluation/tests/tools.py
@@ -58,7 +58,7 @@ class EvapTestRunner(DiscoverRunner):
         parser.add_argument(
             "--headed",
             help="Run the tests in non-headless mode, which makes the browser window visible. Useful for debugging.",
-            action="store_true"
+            action="store_true",
         )
 
     def setup_test_environment(self, **kwargs):


evap/evaluation/tests/tools.py
```diff
@@ -43,11 +43,25 @@
...
+        self.__headed = headed
+
         if not self.tags and not self.exclude_tags:
             self.exclude_tags = {"selenium"}
 
+    @classmethod
+    def add_arguments(cls, parser):
+        super().add_arguments(parser)
+
+        parser.add_argument("--headed", help="Do not run selenium tests in headless mode.", action="store_true")
```
richardebeling: The description is technically correct but has little information amount, because it just says what we do **not** do. Maybe this would be a bit more clear/intuitive?
> Run the tests in non-headless mode, which makes the browser window visible. Useful for debugging.

richardebeling (approved):
Looks good, thanks.

PR #2575, opened 2025-12-01
use ruff for formatting

### LINKED ISSUE #2565: Use ruff for formatting Python code instead of black
We currently already use `ruff` as a part of our linting process, so the dependency is already included. ruff is a drop-in replacement for black and is much faster than black. Currently, there would be some minor changes (dropping empty lines in migrations, some smaller white space changes).

suttrop:
closes #2565

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,7 +76,6 @@ artifacts = [
 
 [tool.ruff]
 line-length = 120
-include = ["evap/**/*.py", "tools/**/*.py"]
 
 [tool.ruff.lint]
 # Those are interesting, but not yet addressed: PYI,Q,SIM,PTH,TRY,RUF


niklasmohrin:
for future reference, python-lsp-ruff supports fixing the I lint group on format if set up during lsp initialization, see their readme here: https://github.com/python-lsp/python-lsp-ruff

evap/evaluation/management/commands/format.py
```diff
@@ -10,24 +10,24 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument(
-            "formatter", nargs="?", choices=["black", "isort", "prettier", "python"], help="Specify a formatter to run."
+            "formatter", nargs="?", choices=["ruff", "isort", "prettier", "python"], help="Specify a formatter to run."
```
richardebeling: `isort` was used because the program was called `isort`. If we make ruff sort the imports, I think we can drop the "isort" subcommands, and can make the invocation of "ruff check --select I --fix" part of our `run_ruff` method below

uv.lock
richardebeling, line 1: isort also has a [[package]] section in this file that should now be removed if rerunning `uv lock`

richardebeling: @niklasmohrin pointed out it's a transitive dependency of pylint, so it's expected that it's still in there. Resolving.

.github/workflows/tests.yml
```diff
@@ -190,9 +190,7 @@ jobs:
           npm-ci: true
 
       - name: Check code formatting
-        run: black --check .
-      - name: Check imports formatting
-        run: isort . --check --diff
+        run: ruff format --check .
```
niklasmohrin, line 193: we should also check the isort lints here

niklasmohrin: (from conversation: this is already checked in the lint step because we have the I group selected by default)

niklasmohrin (approved):
Great, thanks!

PR #2579, opened 2025-12-08
Fix dropout questionnaire blocking vote submission

### LINKED ISSUE #2578: Dropout questionnaire blocks evaluation submission
When submitting an evaluation with a dropout questionnaire (via default voting, not using the dropout option), the submission fails silently (the vote page reloads without errors).

student.views.vote builds the forms without considering whether `dropout` is `True`, so on submission the dropout forms are invalid because the post request provides no data for them.
The dropout forms should not be part of the vote page form groups when `dropout` is `False`.

We also need a test making sure that submitting an evaluation with dropout questionnaire doesn't fail 😅

geromequa:
Closes #2578.

### DIFF
--- a/evap/student/views.py
+++ b/evap/student/views.py
@@ -244,9 +244,7 @@ def get_vote_page_form_groups(
         questionnaires = contribution.questionnaires.all()
         if not questionnaires.exists():
             continue
-        form_groups[contribution] = create_voting_forms(
-            request, contribution, questionnaires, dropout=dropout
-        )
+        form_groups[contribution] = create_voting_forms(request, contribution, questionnaires, dropout=dropout)
 
     return form_groups
 
@@ -261,9 +259,7 @@ def render_vote_page(
 ) -> HttpResponse:
     language = request.GET.get("language", evaluation.main_language)
     with translation.override(language):
-        form_groups = get_vote_page_form_groups(
-            request, evaluation, preview=preview, dropout=dropout
-        )
+        form_groups = get_vote_page_form_groups(request, evaluation, preview=preview, dropout=dropout)
 
     assert preview or not all(form.is_valid() for form_group in form_groups.values() for form in form_group)
 
@@ -333,9 +329,7 @@ def vote(request: HttpRequest, evaluation_id: int, dropout: bool = False) -> Htt
     if not evaluation.can_be_voted_for_by(request.user):
         raise PermissionDenied
 
-    form_groups = get_vote_page_form_groups(
-        request, evaluation, preview=False, dropout=dropout
-    )
+    form_groups = get_vote_page_form_groups(request, evaluation, preview=False, dropout=dropout)
     if not all(form.is_valid() for form_group in form_groups.values() for form in form_group):
         return render_vote_page(request, evaluation, preview=False, dropout=dropout)
 


evap/student/views.py
```diff
@@ -212,7 +212,11 @@ def create_voting_form(
 
 
 def create_voting_forms(
-    request, contribution: Contribution, questionnaires: Iterable[Questionnaire], preselect_no_answer: bool
+    request,
+    contribution: Contribution,
+    questionnaires: Iterable[Questionnaire],
+    preselect_no_answer: bool,
+    dropout: bool,
```
richardebeling: I think `preselect_no_answer` is exclusively used for dropout questionnaires, so passing both this and `dropout` doesn't make much sense

evap/student/tests/test_views.py
```diff
@@ -203,6 +204,19 @@ def setUpTestData(cls):
...
+        )
+        cls.dropout_text_question = baker.make(
+            Question, questionnaire=cls.dropout_questionnaire, order=1, type=QuestionType.TEXT
+        )
+        cls.dropout_likert_question = baker.make(
+            Question, questionnaire=cls.dropout_questionnaire, order=2, type=QuestionType.POSITIVE_LIKERT
+        )
+        cls.droptout_grade_question = baker.make(
+            Question, questionnaire=cls.dropout_questionnaire, order=3, type=QuestionType.GRADE
+        )
```
richardebeling: If we only use this in one test, let's move the setup into that test case. We try to only keep things in the shared setup that are required for more than one test

evap/student/views.py
```diff
@@ -239,7 +244,9 @@ def get_vote_page_form_groups(
         questionnaires = contribution.questionnaires.all()
         if not questionnaires.exists():
             continue
-        form_groups[contribution] = create_voting_forms(request, contribution, questionnaires, preselect_no_answer)
+        form_groups[contribution] = create_voting_forms(
+            request, contribution, questionnaires, preselect_no_answer, dropout
```
richardebeling: Can we use keyword-arguments style (as with the other two calls you touched in here, i.e. `request=request, ...` so that it becomes impossible for us mess up the argument order here?

evap/student/views.py
niklasmohrin, line 219: nit: since we now have two bools, can we do
```suggestion
    questionnaires: Iterable[Questionnaire],
    *,
    preselect_no_answer: bool,
    dropout: bool,
```
and update the usage below?

evap/student/tests/test_views.py
```diff
@@ -346,6 +360,15 @@ def test_incomplete_contributors_vote_form(self):
...
         self.assertEqual(form[field_id(self.contributor_text_question)].value, "some more text")
 
+    def test_answer_with_dropout_questionnaire(self):
+        self.evaluation.general_contribution.questionnaires.add(self.dropout_questionnaire)
+
+        page = self.app.get(self.url, user=self.voting_user1, status=200)
+        form = page.forms["student-vote-form"]
+        self.fill_form(form)
+        response = form.submit()
+        self.assertEqual(SUCCESS_MAGIC_STRING, response.body.decode())
```
niklasmohrin: @richardebeling How do you feel about the following?

```suggestion
    def test_answer_with_dropout_questionnaire(self):
        self.evaluation.general_contribution.questionnaires.add(self.dropout_questionnaire)
        self.test_answer()
```

evap/student/tests/test_views.py
```diff
@@ -346,6 +360,15 @@ def test_incomplete_contributors_vote_form(self):
         field_id = partial(answer_field_id, self.contribution2, self.contributor_questionnaire)
         self.assertEqual(form[field_id(self.contributor_text_question)].value, "some more text")
 
+    def test_answer_with_dropout_questionnaire(self):
```
niklasmohrin: can you add a comment that this is a regression test for issue xxx?

evap/student/tests/test_views.py
richardebeling: lgtm, unless there is some very weird side effects going on then, I don't know if calling a `test_` method directly is supported/intended, or if e.g. transaction stuff is injected into the function definitions by unittest.

If it works without issue, fine with me, otherwise we need to pull out the shared logic into some helper function

evap/student/views.py
geromequa: What does this do?

richardebeling: https://peps.python.org/pep-3102/ makes the following arguments keyword-only, so that they cannot be passed as positional arguments

evap/student/tests/test_views.py
niklasmohrin: i see, helper function seems even better for our sanity

PR #2188, opened 2024-05-13
Downgrade django-fsm to 2.8.2 to fix deprecation warning

richardebeling:
Version 3.0.0 [didn't have functional changes](https://github.com/viewflow/django-fsm/compare/2.8.2...master) except for the annoying deprecation warning

PR #2581, opened 2025-12-15
Migrate to django-fsm-2

niklasmohrin:
See #2160, #2188, and https://github.com/e-valuation/EvaP/pull/2571#issuecomment-3656367991


Kakadus (approved):
django commons looks legit

PR #2584, opened 2025-12-15
Don't log empty m2m changes (fix #2576)

### LINKED ISSUE #2576: Don't log empty m2m changes
Currently, also empty m2m changes are logged, e. g. when calling `course.programs.add(*[])`. This should bef

Example:

```
INSERT INTO "evaluation_logentry" ("content_type_id", "content_object_id", "attached_to_object_type_id", "attached_to_object_id", "datetime", "user_id", "action_type", "request_id", "data") VALUES (17, 115, 17, 115, '2025-12-01 20:44:15.381554'::timestamp, NULL, 'change', '', '{"programs": {"add": []}}'::jsonb) RETURNING "evaluation_logentry"."id"
```

LukasMansour:
See #2576

### DIFF
--- a/evap/evaluation/models_logging.py
+++ b/evap/evaluation/models_logging.py
@@ -368,10 +368,9 @@ def _m2m_changed(sender, instance, action, reverse, model, pk_set, **kwargs):  #
         case _:
             return
 
-    if action_type == (FieldActionType.M2M_ADD or FieldActionType.M2M_REMOVE):
-        if not pk_set:
-            # we don't need to log empty removals or additions, but we do log empty clears, that never enter this if.
-            return
+    if action_type in (FieldActionType.M2M_ADD, FieldActionType.M2M_REMOVE) and not pk_set:
+        # we don't need to log empty removals or additions, but we do log empty clears, that never enter this if.
+        return
 
     if reverse:
         # Since we are not clearing the LoggedModdel instance, we need to log the removal of the related instances
--- a/evap/evaluation/tests/test_models_logging.py
+++ b/evap/evaluation/tests/test_models_logging.py
@@ -4,9 +4,9 @@ from unittest.mock import patch
 from django.utils.formats import localize
 from model_bakery import baker
 
-from evap.evaluation.models import Contribution, Course, Evaluation, Questionnaire, UserProfile
+from evap.evaluation.models import Contribution, Course, Evaluation, Program, Questionnaire, UserProfile
 from evap.evaluation.models_logging import FieldAction, InstanceActionType, _m2m_changed
-from evap.evaluation.tests.tools import TestCase
+from evap.evaluation.tests.tools import TestCase, assert_no_database_modifications
 
 
 class TestLoggedModel(TestCase):
@@ -205,3 +205,12 @@ class TestLoggedModel(TestCase):
             participant.evaluations_participating_in.clear()
             participant.evaluations_participating_in.add(self.evaluation)
             self.assertFalse(any("participants" in entry.data for entry in self.evaluation.related_logentries()))
+
+    def test_no_empty_m2m_logs(self):
+        course = baker.make(Course)
+        program = baker.make(Program)
+        course.programs.add(program)
+        course = Course.objects.get(pk=course.pk)
+
+        with assert_no_database_modifications():
+            course.programs.add(program)
--- a/evap/staff/tests/test_json_importer.py
+++ b/evap/staff/tests/test_json_importer.py
@@ -711,8 +711,6 @@ class TestImportEvents(TestCase):
         evaluation.course.name_en = "Change"
         evaluation.course.save()
 
-        # There is still an update occuring here, but empty m2m changes are no longer an issue.
-        # with assert_no_database_modifications():
         importer = self._import()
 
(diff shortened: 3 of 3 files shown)


evap/evaluation/tests/tools.py
```diff
@@ -294,6 +296,45 @@ def assert_no_database_modifications(*args, **kwargs):
...
+            if query["sql"].startswith('INSERT INTO "evaluation_logentry"') and has_bad_logging_data(
+                extract_jsonb(query["sql"])
+            ):
+                raise AssertionError("Unexpected modifying query found: " + query["sql"])
+
+
+# Helper method that extracts the json literal from an SQL Values clause.
+# Assumes the JSON appears between the last single quotes.
+# Added to solve issue: #2576
+def extract_jsonb(sql: str):
```
richardebeling: unused? I would hope we don't need to write a broken-in-edge-cases extractor here, [there must be a better way](https://www.reddit.com/r/Python/comments/n54f9s/raymond_hettinger_hitting_the_table_there_must_be/)

evap/staff/tests/test_importers.py
```diff
@@ -341,7 +341,8 @@ def create_existing_course(self) -> tuple[Course, Evaluation]:
         return existing_course, existing_course_evaluation
 
     def test_valid_file_import(self):
-        importer_log = import_enrollments(self.default_excel_content, self.semester, None, None, test_run=True)
+        with assert_no_database_useless_logging():
+            importer_log = import_enrollments(self.default_excel_content, self.semester, None, None, test_run=True)
```
richardebeling: My understanding was that this doesn't do any actual database modifications, but it just "accidentally" added the empty log entries. If this PR fixes the empty log entries, we should be good to use `assert_no_database_modifications` (and hopefully we don't need `assert_no_database_useless_logging` at all, then)

evap/evaluation/tests/tools.py
```diff
@@ -294,6 +296,45 @@ def assert_no_database_modifications(*args, **kwargs):
                 raise AssertionError("Unexpected modifying query found: " + query["sql"])
 
 
+@contextmanager
+def assert_no_database_useless_logging(*args, **kwargs):
```
richardebeling: I think with proper testing setup this shouldn't be necessary. We should:
* have a unit test testing the case from the issue, ` course.programs.add(*[])`. This should pass under `assert_no_database_modifications`
* modify the tests for the importer to use `assert_no_database_modifications` (which we would have liked to do before, but couldn't, due to the logging entries)

evap/evaluation/models_logging.py
```diff
@@ -358,18 +358,25 @@ def _m2m_changed(sender, instance, action, reverse, model, pk_set, **kwargs):  #
...
-                action_type = FieldActionType.M2M_REMOVE
-            case _:
-                return
+    match action:
+        case "pre_remove":
+            action_type = FieldActionType.M2M_REMOVE
+        case "pre_add":
+            action_type = FieldActionType.M2M_ADD
+        case "pre_clear":
+            action_type = FieldActionType.M2M_CLEAR
```
richardebeling, line 367: Can you explain why the changes here are necessary, and how this is now supposed to work for the reverse-clear case? My understanding is that if we do a reverse-clear (i.e., `A.b_instances.clear()`), we want to treat it as N forward removes (i.e., as if `B.a_instances.remove()` was called on each B that was in `A.b_instances`). This, in my head, should still work perfectly fine with a generic "If adding or removing an empty set, bail out early" logic below -- and it should even require it: If `A.b_instances` was already empty, we don't want to trigger logging, but right now, your early-out wouldn't catch this case.

evap/evaluation/models_logging.py
```diff
@@ -358,18 +358,26 @@ def _m2m_changed(sender, instance, action, reverse, model, pk_set, **kwargs):  #
...
+        case _:
+            return
+
+    if action_type == (FieldActionType.M2M_ADD or FieldActionType.M2M_REMOVE):
```
niklasmohrin: This does not do what we intend; it needs to be something like
```suggestion
    if action_type in (FieldActionType.M2M_ADD, FieldActionType.M2M_REMOVE):
```

evap/evaluation/models_logging.py
```diff
@@ -358,18 +358,26 @@ def _m2m_changed(sender, instance, action, reverse, model, pk_set, **kwargs):  #
...
+        if not pk_set:
```
niklasmohrin: (I also think we can merge the two nested ifs into one)

evap/evaluation/tests/test_models_logging.py
```diff
@@ -205,3 +205,12 @@ def test_m2m_logging_respects_unlogged_fields(self):
             participant.evaluations_participating_in.clear()
             participant.evaluations_participating_in.add(self.evaluation)
             self.assertFalse(any("participants" in entry.data for entry in self.evaluation.related_logentries()))
+
+    def test_no_empty_m2m_logs(self):
+        evaluation = baker.make(Evaluation)
```
niklasmohrin: what do we need the evaluation for? shouldn't making a course be enough?

richardebeling (approved):

PR #2587, opened 2025-12-15
Questionnaires: add archived state, tabs in management page

### LINKED ISSUE #2545: Questionnaire management
Currently, questionnaires can have one of three visibilities (`HIDDEN`: is not shown in the questionnaire selection in evaluation forms and can be hidden via a switch button on the staff questionnaire page, `MANAGERS`: only managers see them in the questionnaire selection, `EDITORS`: managers and editors see them in the questionnaire selection).
On the staff questionnaire page, all questionnaire types (`DROPOUT`, `TOP`, `CONTRIBUTOR`, `BOTTOM`) are shown below each other in one card per type. Navigating a large number of questionnaires (close to 500 on production) is quite tedious.

The following changes should be made to improve the questionnaire page:
- The four types of questionnaires should be shown in individual tabs (https://getbootstrap.com/docs/5.3/components/navs-tabs/#tabs). In each tab, only questionnaires of one type are shown.
- An additional visibility `ARCHIVED` should be added (icon: `box-archive`)
  - Archived questionnaires should be shown last within the questionnaires of their type.
  - The visibility switch "Hidden questionnaires" on top of the page should be changed to "Show questionnaires" with three options: `All` (all questionnaires are shown), `Non-archived` (shows all but archived), `Visible` (shows only `MANAGERS` and `EDITORS`)
  - The current action "Hide and create new version" should become "Archive and create new version" and the old questionnaire should be archived instead of hidden when clicking on the button.

ZoqkMaze:
fix #2545


evap/evaluation/migrations/0159_merge_20251215_2018.py
niklasmohrin, line 1: this merge migration should not be here; instead, we should rename and change the new migration to be after whatever is the current one (so, at this time, yours should be 0159)

evap/staff/templates/staff_questionnaire_index_list.html
```diff
@@ -1,10 +1,10 @@
 {% if questionnaires %}
-    <div class="card collapsible {{ extra_classes }}">
-        <div class="card-header">
+    <!-- <div class="card collapsible {{ extra_classes }}"> -->
+        <!-- <div class="card-header">
```
niklasmohrin: (these changes should of course not be there when we merge)

niklasmohrin (commented):
do you want to get feedback yet? I took only a very quick look over the changes so far

ZoqkMaze:
> do you want to get feedback yet? I took only a very quick look over the changes so far

Yes, please. Hopefully, only one test is missing

evap/staff/views.py
```diff
@@ -1801,29 +1802,48 @@ def evaluation_preview(request, evaluation_id):
 
 @manager_required
 def questionnaire_index(request):
-    filter_questionnaires = get_parameter_from_url_or_session(request, "filter_questionnaires")
+    filters = ["all", "visible", "non-archived"]
+    filter_questionnaires = request.GET.get("filter_questionnaires", filters[0])
```
jooooosef: Please use `get_parameter_from_url_or_session` as before.

evap/staff/forms.py
```diff
@@ -436,7 +436,7 @@ def __init__(self, *args, requires_decided_main_language=False, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields["course"].queryset = Course.objects.filter(semester=semester)
 
-        visible_questionnaires = ~Q(visibility=Questionnaire.Visibility.HIDDEN)
+        visible_questionnaires = ~Q(visibility__in=[Questionnaire.Visibility.HIDDEN, Questionnaire.Visibility.ARCHIVED])
```
jooooosef: I think we can change this to use a non-negated query to make a little more readable. So use `Q(visibility__in=...)`.

evap/staff/templates/staff_questionnaire_index.html
```diff
@@ -56,14 +59,54 @@
...
+                            <span class="fas"></span>{% translate 'Dropout questionnaires' %}
+                        </a>
+                    </li>
+                </ul>
+            </div>
+
+            <div class="card-body tab-content">
+                <div class="tab-pane show active" id="top" role="tabpanel">
+                    <div class="row align-items-center mb-3">
+                        <!-- {% translate 'Top general questionnaires' as headline %} -->
```
ybrnr: Please remove these comments. You have these titles as the tab descriptions now.

evap/staff/templates/staff_questionnaire_index.html
```diff
@@ -125,4 +168,22 @@
...
         }
     </script>
+
+    <script type="module">
+        var tabName = localStorage['selected_q_tab'];
+        if (["topTab", "contributorTab", "bottomTab", "dropoutTab"].includes(tabName)) {
+            var tab = document.getElementById(tabName);
+            var tabTrigger = new bootstrap.Tab(tab);
+            tabTrigger.show();
+        }
```
jooooosef, line 173: Please be very careful when using `var` since the declaration is [hoisted](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#hoisting).
Use `let` or `const`.

evap/evaluation/tools.py
```diff
@@ -178,7 +178,7 @@ def get_parameter_from_url_or_session(request: HttpRequest, parameter: str, defa
     if result_str is None:  # if no parameter is given take session value
         result = request.session.get(parameter, default)
     else:
-        result = {"true": True, "false": False}.get(result_str.lower())  # convert parameter to boolean
+        result = {"true": True, "false": False}.get(result_str.lower(), default)  # convert parameter to boolean
```
jooooosef, line 178: You made a change here but never used this function. I think this change is useful to return the default value, if there parameter is not true/false @niklasmohrin ?

Should we leave this in here?

ybrnr (changes requested):
you should also run `./manage.py precommit`

evap/staff/views.py
```diff
@@ -1801,29 +1802,48 @@ def evaluation_preview(request, evaluation_id):
...
+    if filter_questionnaires not in filters:
+        filter_questionnaires = filters[0]
```
jooooosef: We have the `get_parameter_from_url_or_session` function which only works for booleans. I think this is fine for now since we do the same at other places but we could think about introducing a `get_string_parameter_from_url_or_session` or something @niklasmohrin @richardebeling ?

jooooosef: I think we typically raise a `BadRequest` or something here, since the user is not supposed to manually fiddle with these. These on the other hand mimics the `get_parameter_from_url_or_session` behaviour where we have a default. Idk if this is okay @niklasmohrin ?

evap/staff/templates/staff_questionnaire_index.html
```diff
@@ -125,4 +168,22 @@
...
+
+        function saveSelectedTab(tab) {
+            console.log("change tab to", tab)
```
richardebeling: Please go through the changes here once and remove everything that should not be merged (console.logs, prints, commented code, unintended whitespace changes)

evap/static/scss/_adjustments.scss
```diff
@@ -173,3 +173,34 @@ $alert-colors: (
...
+}
+
+.nav-item-new {
+    padding: 0pt;
+}
+
+.nav-tabs .nav-item .nav-link.smaller-padding {
+    padding-top: 0.25em;
+    padding-bottom: 0.25em;
+}
```
jooooosef: I think you can do a lot of that styling with [bootstrap classes](https://getbootstrap.com/docs/5.0/getting-started/introduction/) directly in the HTML. 
For example you could use the class `border-0` to not have a border on an item or `py-1` for `0.25` padding.

jooooosef (changes requested):

evap/evaluation/tools.py
niklasmohrin: my intuition is to not change it unless we need it for something; do we use this in this PR?

ZoqkMaze: no, we never use it in this pr. i found it and asked @ janno42 why it does not use the default value. if i remeber correctly, i should add it

niklasmohrin: Hm, I am actually not sure whether this change is correct. The change here makes it so that if users submit a malformed value (not "true" or "false") we ignore it and reset the parameter to the default. On the one hand, that makes sense, because we don't want users to trigger a server crash just by sending a funny http request. On the other hand, sometimes we developers are a little funny ourselves and mess up; in that case, I would like for the server to laugh out very loudly so that we notice haha

What do you think @richardebeling ?

evap/staff/templates/staff_questionnaire_index.html
```diff
@@ -56,14 +59,50 @@
...
-        {% include 'staff_questionnaire_index_list.html' with questionnaires=contributor_questionnaires headline=headline extra_classes='mb-3' type='contributor' %}
-        {% translate 'Bottom general questionnaires' as headline %}
-        {% include 'staff_questionnaire_index_list.html' with questionnaires=general_questionnaires_bottom headline=headline extra_classes='mb-3' type='bottom' %}
-        {% translate 'Dropout questionnaires' as headline %}
-        {% include 'staff_questionnaire_index_list.html' with questionnaires=dropout_questionnaires headline=headline extra_classes='' type='dropout' %}
+        <div class="card card-outline p-0">
+            <div class="d-flex bg-dark-subtle border-bottom border-secondary rounded-card-top">
+                <ul class="nav nav-tabs border-0 p-1 pb-0" role="tablist">
+                    <li class="nav-item p-0">
+                        <a class="nav-link active text-black hover-bg-light py-1" id="topTab" data-bs-toggle="pill" href="#top" role="tab">
```
niklasmohrin: Is there any reason that these should be links? They appear to link to `#top`, but nothing changes in the address bar of my browser. In the documentation that shows the javascript behavior, they use buttons: https://getbootstrap.com/docs/5.0/components/navs-tabs/#javascript-behavior

niklasmohrin: also, is `data-bs-toggle="pill"` what we want? in the documentation is seems that tabs and pills are different styles?

evap/staff/templates/staff_questionnaire_index.html
```diff
@@ -125,4 +164,22 @@
             button.addEventListener("click", () => changeLocked(button));
         }
     </script>
+
+    <script type="module">
+        let tabName = localStorage['selected_q_tab'];
+        if (["topTab", "contributorTab", "bottomTab", "dropoutTab"].includes(tabName)) {
```
niklasmohrin:
```suggestion
        const tabIds = ["topTab", "contributorTab", "bottomTab", "dropoutTab"];
        if (tabIds.includes(tabName)) {
```

evap/staff/templates/staff_questionnaire_index.html
```diff
@@ -125,4 +164,22 @@
...
+        }
+
+        function saveSelectedTab(tab) {
+            localStorage['selected_q_tab'] = tab;
+        }
+
+        document.getElementById("topTab").addEventListener("click", () => saveSelectedTab("topTab"));
+        document.getElementById("contributorTab").addEventListener("click", () => saveSelectedTab("contributorTab"));
+        document.getElementById("bottomTab").addEventListener("click", () => saveSelectedTab("bottomTab"));
+        document.getElementById("dropoutTab").addEventListener("click", () => saveSelectedTab("dropoutTab"));
```
niklasmohrin: ... so that we can do the following here:
```suggestion
        for (const id of tabIds) {
            document.getElementById(id).addEventListener("click", () => saveSelectedTab(id));
        }
```
and not repeat the ids in multiple places; otherwise we might sneak in a typo at some point :)

evap/staff/templates/staff_questionnaire_index.html
```diff
@@ -56,14 +59,50 @@
...
+                    <li class="nav-item p-0">
+                        <a class="nav-link text-black hover-bg-light py-1" id="dropoutTab" data-bs-toggle="pill" href="#dropout" role="tab">
+                            <span class="fas"></span>{% translate 'Dropout questionnaires' %}
+                        </a>
+                    </li>
+                </ul>
+            </div>
+
+            <div class="card-body tab-content">
+                <div class="tab-pane show active" id="top" role="tabpanel">
```
niklasmohrin: ah, I see, these ids here are what the href above targets; I still think the other way with button is slightly nicer because instead of href we can spell out data-bs-target - much more obvious what is going on.

additionally, I would prefer to use the ids `top-tab-content` or so to distinguish it from `topTab`. Using camelCase or kebab-case is both fine, as long as they are consistent :D

evap/staff/templates/staff_questionnaire_index.html
```diff
@@ -56,14 +59,50 @@
...
+                    <div class="row align-items-center mb-3">
+                        {% include 'staff_questionnaire_index_list.html' with questionnaires=general_questionnaires_top headline=headline extra_classes='mb-3' type='top' %}
+                    </div>
+                </div>
+                <div class="tab-pane" id="contributor" role="tabpanel">
+                    {% include 'staff_questionnaire_index_list.html' with questionnaires=contributor_questionnaires headline=headline extra_classes='mb-3' type='contributor' %}
```
niklasmohrin: `headline` is not a thing anymore, so we should also remove it here

(this is ~~an annoying quirk~~ a feature of Django templates: if a variable is not set or raises an error, then just put an empty string in its place)

evap/staff/templates/staff_questionnaire_index_list.html
```diff
@@ -1,10 +1,4 @@
 {% if questionnaires %}
```
niklasmohrin: I wonder whether we should still have this check. Before, it made sense because we would just omit an entire section of the long list with the heading and everything, but now it is a bit odd to go to a tab and see nothing at all.

An empty table would probably also look odd; maybe we should just add an else branch and put something like "There are no xyz"?

evap/staff/templates/staff_questionnaire_index.html
```diff
@@ -125,4 +164,22 @@
             button.addEventListener("click", () => changeLocked(button));
         }
     </script>
+
+    <script type="module">
+        let tabName = localStorage['selected_q_tab'];
+        if (["topTab", "contributorTab", "bottomTab", "dropoutTab"].includes(tabName)) {
+            let tab = document.getElementById(tabName);
+            let tabTrigger = new bootstrap.Tab(tab);
+            tabTrigger.show();
```
niklasmohrin, line 172: If I go to the bottom tab and reload the page (manually, or by changing the visibility settings above), then the page initially flickers because all of the top questionnaires are shown shortly before we switch to this tab again. Can we get rid of that in some way? Maybe hiding all rows initially before we know which tab to show? Or would that also be an annoying flicker?

evap/staff/views.py
niklasmohrin: yeah, SuspiciousOperation or something similar makes sense :)

niklasmohrin: yeah we lost the ability to save the state in the session here, it would be better if we could keep that, `get_string_from_url_or_session` makes sense

evap/staff/views.py
```diff
@@ -1800,29 +1801,48 @@ def evaluation_preview(request, evaluation_id):
...
+        Questionnaire.objects.dropout_questionnaires().prefetch_related(*prefetch_list),
+    ]
 
-    if filter_questionnaires:
-        general_questionnaires = general_questionnaires.exclude(visibility=Questionnaire.Visibility.HIDDEN)
-        contributor_questionnaires = contributor_questionnaires.exclude(visibility=Questionnaire.Visibility.HIDDEN)
+    if filter_questionnaires != "all":
+        questionnaires = [q_type.exclude(visibility=Questionnaire.Visibility.ARCHIVED) for q_type in questionnaires]
+        if filter_questionnaires == "visible":
+            questionnaires = [q_type.exclude(visibility=Questionnaire.Visibility.HIDDEN) for q_type in questionnaires]
```
niklasmohrin: How do you feel about something like the following?
```suggestion
    match filter_questionnaires:
        case "all":
            pass
        case "visible":
            questionnaires = [q_type.filter(...) for q_type in questionnaires]
        case "non-archived":
            questionnaires = [q_type.exclude(...) for q_type in questionnaires]
```
it's longer, but we avoid the whole "filter vs exclude for visible" thing that you fixed in the other file, and we also don't have the double `if`

evap/staff/views.py
```diff
@@ -1800,29 +1801,48 @@ def evaluation_preview(request, evaluation_id):
...
+
+    questionnaires = [
+        q_type.order_by(
+            Case(
+                When(visibility=Questionnaire.Visibility.ARCHIVED, then=Value(1)),
```
niklasmohrin: Can you add a short comment and test for this?

evap/static/scss/_adjustments.scss
```diff
@@ -173,3 +173,12 @@ $alert-colors: (
...
 }
+
+.hover-bg-light:hover {
+    background-color: $lighter-gray;
+}
+
+.rounded-card-top {
+    border-top-left-radius: $card-inner-border-radius;
+    border-top-right-radius: $card-inner-border-radius;
+}
```
niklasmohrin: there is no newline at the end of this file :)

niklasmohrin (commented):
looks good, but there are some rough corners we should still fix

evap/staff/views.py
ZoqkMaze: okay, i will change it. should there be an option to specify possible options?

evap/staff/templates/staff_questionnaire_index_list.html
ZoqkMaze: is the change okay?

evap/staff/templates/staff_questionnaire_index.html
ZoqkMaze: okay ... now nothing is active in default. It is still flickering on startup, but the topTabContent is no longer shown. I am not sure if this is what you wanted. Also, could it be possible that no tab is active?

evap/staff/views.py
ZoqkMaze: i found TestQuestionnaireIndexView.test_ordering. This is now outdated, isn't it?

evap/evaluation/tools.py
richardebeling: I think falling back to a default makes sense, the server ignoring an invalid GET param and using a default value should be easy to debug.

fekoch (changes requested):
@janno42 and me rethought about the "All" - "Visible" - "Non-Archived" filter-buttons. Can you change them to be the following:

 1. "All" (the default) should show all *non*-archived questionnaires
 2. "Visible": same as before
 3. "Archived": only archived questionaires

fekoch (approved):

evap/staff/templates/staff_questionnaire_index.html
```diff
@@ -56,14 +59,50 @@
...
+                    </div>
+                </div>
+                <div class="tab-pane" id="contributorTabContent" role="tabpanel">
+                    {% include 'staff_questionnaire_index_list.html' with questionnaires=contributor_questionnaires extra_classes='mb-3' type='contributor' %}
+                </div>
+                <div class="tab-pane" id="bottomTabContent" role="tabpanel">
+                    {% include 'staff_questionnaire_index_list.html' with questionnaires=general_questionnaires_bottom extra_classes='mb-3' type='bottom' %}
+                </div>
+                <div class="tab-pane" id="dropoutTabContent" role="tabpanel">
+                    {% include 'staff_questionnaire_index_list.html' with questionnaires=dropout_questionnaires extra_classes='' type='dropout' %}
```
richardebeling: ~~In tab layout, I think all `include`s here should get the same `extra_classes` passed. If we can make it work with no `extra_classes`, that would be nice, maybe we can simplify at some point in the future then.~~

Based on your changes in staff_questionnaire_index_list.html, I think these should just be removed, which also solves this.

evap/staff/templates/staff_questionnaire_index.html
```diff
@@ -125,4 +164,23 @@
             button.addEventListener("click", () => changeLocked(button));
         }
     </script>
+
+    <script type="module">
+        const tabIds = ["topTab", "contributorTab", "bottomTab", "dropoutTab"];
+        let tabName = localStorage.getItem('selected_q_tab') || tabIds[0];
```
richardebeling: The localstorage keys are shared across the full web page. Can we name it `staff_questionnaure_index_tab` or something else that clearly connects the key to this specific page?

(If the string is used in multiple places, put it in a constant somewhere and re-use that constant, so we prevent having a typo in one usage and can change it easily)

evap/staff/templates/staff_questionnaire_index.html
richardebeling: We would also prefer `const` whenever possible. We're not linting JS code in the html templates for lack of a suitable tool, so CI will not catch this, but it would still be nice to use `const` whenever possible.

evap/staff/templates/staff_questionnaire_index_list.html
```diff
@@ -1,10 +1,4 @@
...
-    <div class="card collapsible {{ extra_classes }}">
```
richardebeling, line 2: It looks like this was the only usage of `extra_classes` in this template. In that case, we should go through all "includers" and remove any explicit `extra_classes` we pass.

evap/staff/tests/test_views.py
```diff
@@ -2409,6 +2409,25 @@ def test_general_contribution_log_entry(self):
...
+        )
+
+        page = self.app.get(self.url, user=self.manager)
+
+        self.assertIn(selected_hidden_questionnaire.name, page)
+        self.assertIn(selected_archived_questionnaire.name, page)
+        self.assertNotIn(hidden_questionnaire.name, page)
+        self.assertNotIn(archived_questionnaire.name, page)
+
+        self.evaluation.general_contribution.questionnaires.set([self.general_questionnaire])
```
richardebeling: This should be unnecessary, the test framework is responsible for resetting any database modifications (both in the database and for in-memory python objects representing model instances) after a test finished. Did you experience any issues without this line?

evap/staff/tests/test_views.py
```diff
@@ -3185,6 +3204,31 @@ def test_no_second_update(self):
         # We should get redirected back to the questionnaire index.
         self.assertEqual(page.location, "/staff/questionnaire/")
 
+    def test_change_old_visibility(self):
```
richardebeling: Optimally, this test and `test_changes_old_title` would be merged so that we have one test that tests "making a new version of an existing questionnaire", including all the assertions of the final state of both questionnaires.

I think the code in the PR here is fine for now, but if you want, feel free to just merge them.

evap/staff/tests/test_views.py
```diff
@@ -3185,6 +3204,31 @@ def test_no_second_update(self):
...
+            Questionnaire.objects.get(pk=self.questionnaire.pk).visibility, Questionnaire.Visibility.ARCHIVED
+        )
+
+
+class TestQuestionnaireIndexView(WebTestStaffMode):
+    url = "/staff/questionnaire/"
+
+    @classmethod
+    def setUpTestData(cls):
+        cls.manager = make_manager()
```
richardebeling: You moved this test class up from below `TestQuestionnaireCreateView`. Not a super big problem here, but in general, for code reviews and code history tracking, we would like to not add unnecessary changes, including moving lines around. It creates noise in the git history and makes it harder to follow actual code changes.

evap/static/scss/_adjustments.scss
```diff
@@ -173,3 +173,12 @@ $alert-colors: (
 input[type="search"]::-webkit-search-cancel-button {
 	-webkit-appearance: none;
 }
+
+.hover-bg-light:hover {
+    background-color: $lighter-gray;
+}
```
richardebeling: Why is this needed on the buttons? We have lots of buttons on the website, hovering should work consistently for all of them.

If this works around some bootstrap bug, we should put a comment here explaining it (and, if possible, linking to an issue on their end so that we can easily find out if it has been fixed)

evap/static/scss/_adjustments.scss
richardebeling: Looks unused -> please remove

evap/staff/views.py
```diff
@@ -1804,29 +1805,50 @@ def evaluation_preview(request, evaluation_id):
...
+    questionnaires = [
+        Questionnaire.objects.general_questionnaires().prefetch_related(*prefetch_list),
+        Questionnaire.objects.contributor_questionnaires().prefetch_related(*prefetch_list),
+        Questionnaire.objects.dropout_questionnaires().prefetch_related(*prefetch_list),
+    ]
 
-    if filter_questionnaires:
-        general_questionnaires = general_questionnaires.exclude(visibility=Questionnaire.Visibility.HIDDEN)
-        contributor_questionnaires = contributor_questionnaires.exclude(visibility=Questionnaire.Visibility.HIDDEN)
+    match filter_questionnaires:
```
richardebeling, line 1830: Taking the named variables and making them a constant-index element in a three-element list makes this method a bit harder to read.

Django allows passing QuerySet filters around without "attaching" them to a specific query/QuerySet right away by using `Q` objects ([docs](https://docs.djangoproject.com/en/6.0/topics/db/queries/#complex-lookups-with-q-objects)). We can use this to factor out all the shared aspects of the queries, so that we can then directly create the lists we need. This allows to express the same logic with less code and without having to modify state so much. It would look somewhat like this:

```python3
match filter_questionnaires:
    case "all":
        questionnaires_filter = ~Q(visibility=Questionnaire.Visibility.ARCHIVED)
    case "visible":
        questionnaires_filter = ~Q(visibility__in=[Questionnaire.Visibility.ARCHIVED, Questionnaire.Visibility.HIDDEN])
    case "archived":
        questionnaires_filter = Q(visibility=Questionnaire.Visibility.ARCHIVED)

questionnaires = (
    Questionnaires.object.all()
    .filter(questionnaires_filter)
    .order_by("order", "pk")
    .prefetch_related("questions", "contributions__evaluation")
)

template_data = {
    "general_questionnaires_top": list(questionnaires.filter(type=Questionnaire.Type.TOP)),
    "general_questionnaires_bottom": list(questionnaires.filter(type=Questionnaire.Type.BOTTOM)),
    "contributor_questionnaires": list(questionnaires.filter(type=Questionnaire.Type.CONTRIBUTOR)),
    "dropout_questionnaires": list(questionnaires.filter(type=Questionnaire.Type.DROPOUT)),
    "filter_questionnaires": filter_questionnaires
}
```

evap/static/scss/_adjustments.scss
ZoqkMaze: It is necessary, but I do not know why

evap/static/scss/_adjustments.scss
ZoqkMaze: It is used: `<div class="d-flex bg-dark-subtle border-bottom border-secondary rounded-card-top">`. Without, there is a subtle visual bug at the top corners of the card.

evap/static/scss/_adjustments.scss
richardebeling: You're using the bootstrap components in "unintended" ways, with lots of explicit styling. That doesn't work nicely with the way the components were intended to be used. To fix that, you then have to add new styling/classes here.

Bootstrap tabs [are supposed to](https://getbootstrap.com/docs/5.2/components/navs-tabs/#tabs) have white background and link-colored text in a light theme, with spacious padding around the tabs. For this to work, the tabs need to be in the card body, the card is not supposed to have content in its "outer" shell.

I would very much prefer if we stick to the way bootstrap intends these components to be used, and re-use their default styling, without our explicit overrides. This allows for much easier support for dark-mode and also makes the responsiveness work as intended. I added a commit [here](https://github.com/richardebeling/EvaP/commits/q_management/) that does this.

Your version:
custom styling, wide screen, light mode
custom styling, wide screen, dark mode
custom styling, narrow screen, light mode


The simplified version:
default styling, wide screen, light mode
default styling, wide screen, dark mode
default styling, narrow screen, light mode

(If we wanted to override the bootstrap styling, we'd have to do that globally for all of EvaP, but I think the default styling is good enough for us here)

tl;dr: Let's simplify our usage, put the tab selectors into the card body, and let them look the way bootstrap intends them to look. Simpler code, more compatible styling, easier life :D

ZoqkMaze: Thank you very much! This is much better.
I'm sorry for producing such messy work. Next time I will take a closer look at the bootstrap documentation.

richardebeling (approved):

janno42 (changes requested):
Please remove the text "There are no matching questionnaires" and instead keep the empty table (with header). We will address the handling of empty tables in a separate issue.

evap/evaluation/tools.py
```diff
@@ -178,11 +178,19 @@ def get_parameter_from_url_or_session(request: HttpRequest, parameter: str, defa
...
     request.session[parameter] = result  # store value for session
     return result
 
 
+def get_string_from_url_or_session(request: HttpRequest, parameter: str, default=False) -> str:
```
niklasmohrin: Not sure if there is a sensible default, but the type annotation should surely be:
```suggestion
def get_string_from_url_or_session(request: HttpRequest, parameter: str, default: str) -> str:
```

I am not sure why the function passed typechecking in the first place, maybe the GET.get returns Any?

evap/staff/tests/test_live.py
```diff
@@ -178,6 +178,27 @@ def test_collapse_without_editor_approved(self) -> None:
         self.assertEqual(counter.text, "0")
 
 
+class QuestionnaireLiveTest(LiveServerTest):
```
niklasmohrin, line 229: Very nice, this is a very clear test :)

evap/staff/templates/staff_questionnaire_index.html
```diff
@@ -10,13 +10,16 @@
         </div>
         <div class="col-auto">
             <div class="btn-switch btn-switch-light">
-                <div class="btn-switch-label">{% translate 'Hidden questionnaires' %}</div>
+                <div class="btn-switch-label">{% translate 'Show questionnaires' %}</div>
                 <div class="btn-switch btn-group">
-                    <a href="{% url 'staff:questionnaire_index' %}?filter_questionnaires=false" role="button" class="btn btn-sm btn-light{% if not filter_questionnaires %} active{% endif %}">
-                    {% translate 'Show' %}
+                    <a href="{% url 'staff:questionnaire_index' %}?filter_questionnaires=all" role="button" class="btn btn-sm btn-light{% if filter_questionnaires == 'all' %} active{% endif %}">
+                    {% translate 'All' %}
```
niklasmohrin:
```suggestion
                        {% translate 'All' %}
```

evap/staff/templates/staff_questionnaire_index_list.html
```diff
@@ -1,10 +1,4 @@
-{% if questionnaires %}
-    <div class="card collapsible {{ extra_classes }}">
```
niklasmohrin, line 2: Can you remove the two leading spaces in the rest of the file? Currently, the whole file is indented

niklasmohrin (changes requested):
Everything seems to work as expected, some small comments on the code

niklasmohrin:
Note that we still have an "There are no questionnaires yet." with no table at all if there are no questionnaires in any of the tabs; @janno42 do you want this one to also go away?

evap/staff/templates/staff_questionnaire_index.html
```diff
@@ -56,14 +59,50 @@
...
+                            {% translate 'Dropout questionnaires' %}
+                        </button>
+                    </li>
+                </ul>
+
+                <div class="tab-content">
+                    <div class="tab-pane show" id="topTabContent" role="tabpanel">
+                        <div class="row align-items-center mb-3">
+                            {% include 'staff_questionnaire_index_list.html' with questionnaires=general_questionnaires_top type='top' %}
+                        </div>
```
janno42:
```suggestion
                        {% include 'staff_questionnaire_index_list.html' with questionnaires=general_questionnaires_top type='top' %}
```

evap/staff/templates/staff_questionnaire_index.html
```diff
@@ -56,14 +59,50 @@
...
 
-        {% translate 'Top general questionnaires' as headline %}
-        {% include 'staff_questionnaire_index_list.html' with questionnaires=general_questionnaires_top headline=headline extra_classes='mb-3' type='top' %}
-        {% translate 'Contributor questionnaires' as headline %}
-        {% include 'staff_questionnaire_index_list.html' with questionnaires=contributor_questionnaires headline=headline extra_classes='mb-3' type='contributor' %}
-        {% translate 'Bottom general questionnaires' as headline %}
-        {% include 'staff_questionnaire_index_list.html' with questionnaires=general_questionnaires_bottom headline=headline extra_classes='mb-3' type='bottom' %}
-        {% translate 'Dropout questionnaires' as headline %}
-        {% include 'staff_questionnaire_index_list.html' with questionnaires=dropout_questionnaires headline=headline extra_classes='' type='dropout' %}
+        <div class="card card-outline p-0">
```
janno42: why do we need `p-0`?

janno42 (changes requested):

evap/static/scss/_adjustments.scss
```diff
@@ -173,3 +173,10 @@ $alert-colors: (
 input[type="search"]::-webkit-search-cancel-button {
 	-webkit-appearance: none;
 }
+
+.card .card-header .nav-tabs .nav-item .nav-link {
+    border: none;
+    &:hover {
+        background-color: $white;
+    }
+}
```
janno42, line 183:
```suggestion
.card .card-header .nav-tabs .nav-item .nav-link {
    border: none;
    &:hover:not(.active) {
        background-color: $light-gray;
    }
}
```

janno42 (changes requested):
Looking good - except for this small thing. Then we can merge from my side :)

evap/static/scss/_adjustments.scss
```diff
@@ -174,6 +174,7 @@ input[type="search"]::-webkit-search-cancel-button {
 	-webkit-appearance: none;
 }
 
+// Fix hover background color of inactive tabs and remove bottom border line visuell glitch
```
janno42: Maybe just `// Fix tabs in card headers`?

janno42 (approved):
Manage all the questionnaires!

evap/evaluation/tools.py
```diff
@@ -175,11 +175,19 @@ def get_parameter_from_url_or_session(request: HttpRequest, parameter: str, defa
...
     request.session[parameter] = result  # store value for session
     return result
 
 
+def get_string_from_url_or_session(request: HttpRequest, parameter: str, default: str | None = None) -> str | None:
+    result = request.GET.get(parameter, None)
+    if result is None:
+        result = request.session.get(parameter, default)
+    request.session[parameter] = result
+    return result
```
richardebeling, line 188: The two helpers here are now basically the same thing, and they are named a bit weirdly.

We can do it as a follow-up issue to get this merged quickly, but I guess it would be nice to think about proper names here and how to make one of these use the other one.

richardebeling (approved):

niklasmohrin (approved):

PR #2588, opened 2025-12-15
Upgrade to django 6.0

### LINKED ISSUE #2537: Upgrade to Django 6.0
Django 6.0 [was released](https://www.djangoproject.com/weblog/2025/dec/03/django-60-released/) on 3rd of December. [Same procedure as every year](https://github.com/e-valuation/EvaP/issues/1387).

* [ ] We'll need to upgrade to Python 3.12
* [ ] Interesting new idea: Django will support [template partials](https://docs.djangoproject.com/en/dev/ref/templates/language/#template-partials), a mechanism to have repeatable/reusable blocks in templates. We should see in which places we have split up templates into multiple files to achieve this, and whether we can use template partials instead
* [ ] There were [changes to the email backend](https://docs.djangoproject.com/en/dev/releases/6.0/#adoption-of-python-s-modern-email-api), we should check whether we should migrate anything and whether everything still works.
* [ ] Possibly useful: In templates, we can now get `forloop.length`.

This looks interesting at first glance, but I think is not helpful for us:
* They [added support for running tasks outside of HTTP requests](https://docs.djangoproject.com/en/dev/releases/6.0/#background-tasks), but it seems to me that we cannot yet use this to replace our cron-based regularly scheduled tasks. It seems to be more tailored towards enqueuing tasks during processing of a HTTP request that then execute asynchronously after the request execution is finished

karyon:
fixes #2537 

Includes an update to python 3.13. Does not update to python 3.14, since that threw errors during installing. Something about setuptools in cffi (from mozilla-django-oidc ->  josepy -> cryptography -> cffi), hopefully is solved with the ext mozilla-django-oidc release.

Includes a workaround for https://code.djangoproject.com/ticket/36800. It has been labeled as release blocker, so it should be in django 6.0.1, I'd suggest to wait for that.

I looked into the template partials, even gathered a statistic about our template includes, but even for files that were included multiple times but only by a single other file, I thought having it in a separate file was actually nicer.


pyproject.toml
```diff
@@ -3,14 +3,14 @@ name = "evap"
...
-requires-python = ">=3.11.0"
+requires-python = ">=3.13.0"
 dependencies = [
     "django-extensions~=4.1.0",
     "django-fsm~=2.8.2",
-    "Django~=5.2.0",
+    "Django~=6.0.0",
     "mozilla-django-oidc~=4.0.1",
     "openpyxl~=3.1.5",
-    "psycopg~=3.2.3",
```
karyon: This was actually required. model-bakery and selenium I updated only because I thought it would help, but it didn't in the end, but I thought I'd leave it in, I believe python 3.14 would require the new selenium anyways.

tools/check_dist.py
```diff
@@ -28,13 +28,10 @@ def ensure_all_artifacts_included(pyproject, wheel_paths):
 
 
 def main(argv):
-    parser = argparse.ArgumentParser("check_dist", exit_on_error=False)
+    parser = argparse.ArgumentParser("check_dist")
     parser.add_argument("pyproject", type=Path)
     parser.add_argument("wheels", type=Path, nargs="*")
-    try:
-        args = parser.parse_args(argv)
-    except SystemExit:
```
karyon, line 36: I wasn't exactly sure about this. `exit_on_error=False` seemed to suggest that a system exit was unwanted, but the test `test_usage` suggested the exit behavior was wanted (only with exit_on_error the usage message is printed). Python 3.13 fixed a bug where exit_on_error=False wasn't respected, so now the test failed. I opted for using the test as expected behavior. @niklasmohrin?

evap/evaluation/tools.py
```diff
@@ -190,17 +189,16 @@ def translate(**kwargs):
     return property(lambda self: getattr(self, kwargs[get_language() or "en"]))
 
 
-EmailT = TypeVar("EmailT", str, None)
+def clean_email(email: str | None) -> str | None:
+    if email is None:
+        return None
 
-
-def clean_email(email: EmailT) -> EmailT:
```
niklasmohrin, line 196: The previous bound actually was tighter: EmailT is constrained to be either exactly str or exactly None, and the same type is returned; with the new signature we also permit str -> None and None -> str

I think we use the str->str guarantee in some auth code

tools/check_dist.py
niklasmohrin: Hm, I don't fully remember why I did it like this, I think it was because I didn't want to catch SystemExit in the test. Nowadays I don't really care

niklasmohrin (commented):
Agree on waiting for 6.0.1

pyproject.toml
```diff
@@ -23,11 +23,11 @@ psycopg-c = ["psycopg[c]"]
 
 [dependency-groups]
 dev = [
-    "coverage[toml]~=7.10.2",
+    "coverage~=7.10.2",
```
richardebeling: Out of curiosity: How did you notice this, did some tooling tell you?

richardebeling (commented):
lgtm when the 6.0.1 fix is in and the open discussions are resolved. Thanks!

evap/evaluation/tools.py
```diff
@@ -32,7 +32,6 @@
     except ImportError:
         StrOrPromise = Any  # on production setups, type alias to Any
 
-M = TypeVar("M", bound=Model)
 T = TypeVar("T")
 CellValue = str | int | float | None
 CV = TypeVar("CV", bound=CellValue)
```
Kakadus: any reason why we're keeping `T` and `CV`? `T` seems to be unused and the usages of `CV` can also be written using the new syntax

pyproject.toml
karyon: I don't recall, sorry :(

richardebeling (approved):
thanks

PR #2590, opened 2025-12-15
Fix readme papercuts, add process-compose.socket to .gitignore

karyon:
Small things I stumbled over while setting this up.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,7 @@ docs/_build
 coverage.xml
 htmlcov
 
-# seen once in WSL, possibly produced by our nix setup
+# Ignored by default, but seen as untracked in WSL once
 process-compose.socket
 
 # pip puts editable packages here


README.md
```diff
@@ -29,7 +29,7 @@ We use [nix](https://nixos.org/) to manage the development environment.
    ```
 5. Open a new terminal. Enter the development shell and start EvaP:
    ```
-   cd EvaP
```
karyon, line 32: It's not necessarily there but above it suggests `cd ~`.

README.md
```diff
@@ -46,8 +46,9 @@ Afterwards, `nix run .#services-full` will recreate a default development enviro
...
 
 Before committing, run the following commands:
-- `./manage.py test` (runs the test suite)
-- `./manage.py lint` (runs linting)
+- `./manage.py typecheck`
+- `./manage.py test`
+- `./manage.py lint`
 - `./manage.py format` (applies automatic code formatting)
 
 or, to combine all three, simply run `./manage.py precommit`.
```
richardebeling:
```suggestion
or, to combine them, simply run `./manage.py precommit`.
```

Maybe we should swap these and mention the precommit command first and then the list, idk

README.md
richardebeling: If we're touching this: We _just_ removed isort in https://github.com/e-valuation/EvaP/pull/2575, probably best to update this as well (add ruff, remove isort)

.gitignore
```diff
@@ -33,6 +33,9 @@ docs/_build
 coverage.xml
 htmlcov
 
+# our nix setup produces this
+process-compose.socket
```
richardebeling, line 37: Out of curiosity, what was your setup, Windows 11 with WSL 2? I think I have never seen this file mentioned by git. On my Linux, it's a socket, and git auto-ignores it.

README.md
richardebeling: Urgh, this is ugliniess coming in from the NTFS/Windows workarounds. All commands here currently work anywhere on disk. With an absolute path, we drop that, which I think makes it a bit worse

I see your point about us mentioning ~ above in the Windows-only step. The path there is just supposed to be a weak suggestion to give users _some_ option that works, since we'd expect them to be unfamiliar with the linux file system structure at that point.

I think it would be a bit better to try and reframe the contents of the windows-only point instead?

karyon: I didn't know what to reframe this to, I reverted this part

.gitignore
karyon: yes, that was the setup.

evap/evaluation/management/commands/precommit.py
```diff
@@ -18,9 +18,8 @@ def handle(self, *args, **options):
             sys.exit(1)
 
         call_command("typecheck")
+        call_command("format")
+        call_command("lint")
```
karyon, line 23: moved here because a failing test made them be skipped.

.gitignore
```diff
@@ -33,6 +33,9 @@ docs/_build
 coverage.xml
 htmlcov
 
+# seen once in WSL, possibly produced by our nix setup
```
niklasmohrin:
```suggestion
# Ignored by default, but seen as untracked in WSL once
```

PR #2592, opened 2025-12-16
Importer improvements (part 3)

janno42:

### DIFF
--- a/evap/staff/importers/json.py
+++ b/evap/staff/importers/json.py
@@ -93,8 +93,10 @@ import_dict_adapter = TypeAdapter(ImportDict)
 
 @dataclass
 class NameChange:
+    old_title: str
     old_last_name: str
     old_first_name_given: str
+    new_title: str
     new_last_name: str
     new_first_name_given: str
     email: str | None
@@ -141,7 +143,7 @@ class ImportStatistics:
         log += self._make_heading("Name Changes")
         log += self._make_total(len(self.name_changes))
         for name_change in self.name_changes:
-            log += f"- {name_change.old_first_name_given} {name_change.old_last_name} → {name_change.new_first_name_given} {name_change.new_last_name} (email: {name_change.email})\n"
+            log += f"- [{name_change.old_title}] {name_change.old_last_name}, {name_change.old_first_name_given} → [{name_change.new_title}] {name_change.new_last_name}, {name_change.new_first_name_given} (email: {name_change.email})\n"
         log += "\n"
 
         log += self._make_stats("New Courses", self.new_courses)
@@ -288,10 +290,12 @@ class JSONImporter:
 
     def _create_name_change_from_changes(self, user_profile: UserProfile, changes: dict[str, tuple[Any, Any]]) -> None:
         change = NameChange(
+            old_title=changes["title"][0] if changes.get("title") else user_profile.title,
             old_last_name=changes["last_name"][0] if changes.get("last_name") else user_profile.last_name,
             old_first_name_given=(
                 changes["first_name_given"][0] if changes.get("first_name_given") else user_profile.first_name_given
             ),
+            new_title=user_profile.title,
             new_last_name=user_profile.last_name,
             new_first_name_given=user_profile.first_name_given,
             email=user_profile.email,
--- a/evap/staff/tests/test_json_importer.py
+++ b/evap/staff/tests/test_json_importer.py
@@ -331,8 +331,10 @@ class TestImportUserProfiles(TestCase):
             importer.statistics.name_changes,
             [
                 NameChange(
+                    old_title="",
                     old_last_name="Doe",
                     old_first_name_given="Jane",
+                    new_title="",
                     new_last_name=self.students[0]["name"],
                     new_first_name_given=self.students[0]["callingname"],
                     email=self.students[0]["email"],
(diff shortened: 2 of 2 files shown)


hansegucker (approved):

evap/contributor/forms.py
```diff
@@ -48,6 +49,10 @@ def __init__(self, *args, **kwargs):
         self.fields["name_de_field"].initial = self.instance.full_name_de
         self.fields["name_en_field"].initial = self.instance.full_name_en
 
+        self.fields["exam_type"].disabled = True
+        if not self.instance.exam_type:
```
niklasmohrin:
```suggestion
        if self.instance.exam_type is None:
```

evap/contributor/views.py
```diff
@@ -227,6 +227,7 @@ def evaluation_edit(request, evaluation_id):
         "editable": True,
         "preview_html": preview_html,
         "questionnaires_with_answers_per_contributor": {},
+        "show_participant_cms_info": evaluation.exam_type is not None and evaluation.cms_id is not None,
```
niklasmohrin: Can we keep this decision in the form? Maybe we could set something like `self.fields["particpants"].cms_disclaimer = _(...)` and pick this up in the view?

niklasmohrin: even better would be if such an attribute would be included in our typing somehow, then we could also move the whole show-or-collapsed business to that.

not for this PR though, I am fine if we just keep the field specific things close to the field

evap/staff/importers/json.py
```diff
@@ -160,14 +160,19 @@ def get_log(self) -> str:
     def send_mail(self):
         subject = "[EvaP] JSON importer log"
 
-        managers = UserProfile.objects.filter(groups__name="Manager", email__isnull=False)
-        if not managers:
-            return
+        recipients = settings.JSON_IMPORTER_LOG_RECIPIENTS
+        if not recipients:
+            recipients = [a[1] for a in settings.ADMINS]
```
niklasmohrin, line 167: With #2499 we could do this fallback in the settings already :D

evap/staff/importers/json.py
```diff
@@ -141,7 +143,7 @@ def get_log(self) -> str:
         log += self._make_heading("Name Changes")
         log += self._make_total(len(self.name_changes))
         for name_change in self.name_changes:
-            log += f"- {name_change.old_first_name_given} {name_change.old_last_name} → {name_change.new_first_name_given} {name_change.new_last_name} (email: {name_change.email})\n"
+            log += f"- [{name_change.old_title}] {name_change.old_last_name}, {name_change.old_first_name_given} → [{name_change.new_title}] {name_change.new_last_name}, {name_change.new_first_name_given} (email: {name_change.email})\n"
```
niklasmohrin, line 146: is it intended that most names will then be like `[] Doe, John`?

niklasmohrin (commented):
makes sense, some remarks:

evap/staff/importers/json.py
janno42: yes. it's important to clearly separate the individual parts of the name and also explicitly see empty titles. we could also format it differently, i don't really care about the exact format. would you prefer something else?

niklasmohrin: no it's fine, just wanted to check so that we don't open "Log should not display [] if no title is set" after the next update :D

evap/contributor/views.py
janno42: adding attributes to fields is tricky, i now added it to the form. does that work for you?

richardebeling: Out of curiosity: What makes it tricky? I would have expected
```
self.fields["participants"].evap_participants_set_by_cms = True
```
to just work (apart from typechecking -> we can either ignore it, or add the four lines to have the custom field class with the attribute defined visibly for mypy).

I would also very much prefer to have this stored in the field, if possible just as a bool, with the translated string being added in the template.

evap/staff/importers/json.py
```diff
@@ -160,14 +162,19 @@ def get_log(self) -> str:
...
 
-        managers = UserProfile.objects.filter(groups__name="Manager", email__isnull=False)
-        if not managers:
-            return
+        recipients = settings.JSON_IMPORTER_LOG_RECIPIENTS
+        if not recipients:
+            recipients = [a[1] for a in settings.ADMINS]
+        bcc_addresses = []
+        if settings.SEND_ALL_EMAILS_TO_ADMINS_IN_BCC:
+            bcc_addresses = [a[1] for a in settings.ADMINS]
```
richardebeling, line 170: If user code can forget to consider this setting, and we have forgotten to consider this in the past, should we provide a `EvapEmailMultiAlternative` class that adds this automatically? (-> separate issue)

evap/contributor/views.py
janno42: i tried setting it for the field directly but `field.cms_disclaimer` is empty

evap/staff/importers/json.py
janno42: I don't think this is important. This setting is usually only True for a limited amount of time when testing new email functionality on production.

evap/contributor/views.py
richardebeling: Ok, we can merge as-is then from my side, I'd look into why it's not working or if we can find a better solution in the next days.

PR #2593, opened 2025-12-16
Set vote_end_date during json import only once

janno42:
currently the vote_end_date is set twice: on the import of the main evaluation and then again after the exam evaluation was imported. this leads to two database updates and two log entries even in cases where no update would be necessary at all and it is only changed back and forth.

this pr changes the import order:
1. course
2. exam evaluation(s)
3. main evaluation

so when the main evaluation is imported, the exam date is already known and the vote_end_date can be set correctly in the one and only change (if necessary).

### DIFF
--- a/evap/staff/importers/json.py
+++ b/evap/staff/importers/json.py
@@ -10,6 +10,7 @@ from typing import Any, NotRequired
 from django.conf import settings
 from django.core.mail import EmailMultiAlternatives
 from django.db import transaction
+from django.db.models import Min
 from django.utils.timezone import now
 from pydantic import TypeAdapter
 from typing_extensions import TypedDict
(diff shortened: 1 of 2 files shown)


evap/staff/importers/json.py
```diff
@@ -456,14 +460,35 @@ def _import_evaluation(self, course: Course, data: ImportEvent) -> Evaluation:
...
             evaluation_start_datetime = course_end.replace(hour=8, minute=0, second=0, microsecond=0) - timedelta(
                 weeks=1, days=course_end.weekday()
             )
             # End date is on the sunday in the week the event ends
             evaluation_end_date = (course_end + timedelta(days=6 - course_end.weekday())).date()
 
+            # Change evaluation period according to earliest exam date, if the course has an exam
+            if earliest_exam_date:
+                if earliest_exam_date <= evaluation_start_datetime.date():
+                    if not evaluation:
```
richardebeling, line 488: Why do we only create these warnings if the evaluation does not exist?

If we already have an evaluation, we would attempt to update it below, so it feels like the warning would also be appropriate?

I guess there is some reason in the data distribution we have -- can we add a short comment in that case that explains this `if`?

evap/staff/importers/json.py
```diff
@@ -572,12 +596,21 @@ def _import_events(self, data: list[ImportEvent]) -> None:  # noqa:PLR0912
...
+            course = self.courses_by_gguid.get(event["gguid"])
+            if course is not None:
+                exam_evaluations = [
+                    evaluation for evaluation in course.evaluations.all() if evaluation.exam_type is not None
+                ]
+                earliest_exam_date = None
+                if exam_evaluations:
+                    earliest_exam_date = min(
+                        evaluation.vote_start_datetime for evaluation in exam_evaluations
+                    ).date() - timedelta(days=1)
```
richardebeling: Can we do something like
```python3
min_vote_start_datetime = course.exam_evaluations.aggregate(Min("vote_start_datetime"))["vote_start_datetime__min"]
earliest_exam_date = min_vote_start_datetime.date() - timedelta(days=1) if min_vote_start_datetime else None
```
?

evap/staff/tests/test_json_importer.py
richardebeling, line 1: Can we add a short test that the change log entries in the previously problematic case are as expected?

evap/staff/importers/json.py
janno42: These types of warnings generate hundreds of warnings with every import. Once we have daily automated imports, a lot of time would be wasted parsing all the warnings to determine if action is necessary (in these cases, usually no further action is necessary after the initial import). If any values change, the updated evaluations are listed, which is sufficient.

evap/staff/importers/json.py
```diff
@@ -546,22 +578,21 @@ def _import_contribution(self, evaluation: Evaluation, data: ImportRelated) -> t
...
+            if not event["isexam"]:
+                non_exam_events.append(event)
+            elif event.get("relatedevents"):
+                exam_events.append(event)
+            else:
+                exam_events_without_related_non_exam_event.append(event)
 
+        # Only import courses in first step, don't create evaluations yet
         for event in non_exam_events:
             course = self._import_course(event)
```
niklasmohrin: We are not using the course in this loop anymore
```suggestion
            self._import_course(event)
```

PR #2594, opened 2025-12-16
Importer logging and attempted changes

janno42:
- fix attempted changes logging in json importer
- allow participant changes until evaluation begins

### DIFF
--- a/evap/staff/importers/json.py
+++ b/evap/staff/importers/json.py
@@ -1,7 +1,7 @@
 import json
 import logging
 import re
-from collections.abc import Iterable
+from collections.abc import Collection, Iterable
 from dataclasses import dataclass, field
 from datetime import date, datetime, timedelta
 from datetime import time as datetime_time
@@ -109,14 +109,17 @@ class WarningMessage:
     message: str
 
 
+# pylint: disable=too-many-instance-attributes
 @dataclass
 class ImportStatistics:
     name_changes: list[NameChange] = field(default_factory=list)
     new_courses: list[Course] = field(default_factory=list)
     new_evaluations: list[Evaluation] = field(default_factory=list)
     updated_courses: list[Course] = field(default_factory=list)
-    updated_evaluations: list[Evaluation] = field(default_factory=list)
-    attempted_changes: list[Evaluation] = field(default_factory=list)
+    updated_evaluations: set[Evaluation] = field(default_factory=set)
+    updated_participants: list[Evaluation] = field(default_factory=list)
+    attempted_evaluation_changes: list[Evaluation] = field(default_factory=list)
+    attempted_participant_changes: list[Evaluation] = field(default_factory=list)
     warnings: list[WarningMessage] = field(default_factory=list)
 
     @staticmethod
@@ -128,7 +131,7 @@ class ImportStatistics:
         return f"({total} in total)\n\n"
 
     @staticmethod
-    def _make_stats(heading: str, objects: list) -> str:
+    def _make_stats(heading: str, objects: Collection) -> str:
         log = ImportStatistics._make_heading(heading)
         log += ImportStatistics._make_total(len(objects))
         for obj in objects:
@@ -151,7 +154,9 @@ class ImportStatistics:
         log += self._make_stats("New Evaluations", self.new_evaluations)
         log += self._make_stats("Updated Courses", self.updated_courses)
         log += self._make_stats("Updated Evaluations", self.updated_evaluations)
-        log += self._make_stats("Attempted Changes", self.attempted_changes)
+        log += self._make_stats("Updated Participants", self.updated_participants)
+        log += self._make_stats("Attempted Changes", self.attempted_evaluation_changes)
+        log += self._make_stats("Attempted Participant Changes", self.attempted_participant_changes)
 
         log += self._make_heading("Warnings")
         log += self._make_total(len(self.warnings))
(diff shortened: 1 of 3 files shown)


evap/staff/importers/json.py
```diff
@@ -505,13 +510,25 @@ def _import_evaluation(self, course: Course, data: ImportEvent) -> Evaluation:
...
-            participant_changes = set(evaluation.participants.all()) != set(participants)
-            evaluation.participants.set(participants)
+        direct_changes = update_with_changes(evaluation, defaults, test=True)
+        if direct_changes:
+            if evaluation.state >= Evaluation.State.APPROVED:
+                self.statistics.attempted_changes.append(evaluation)
+            else:
+                direct_changes = update_with_changes(evaluation, defaults, test=False)
+                if not created:
+                    self.statistics.updated_evaluations.append(evaluation)
```
richardebeling: Reaching 3 nesting levels and running `update_with_changes` twice on the allow-changes path seems a bit complex to me.

We only ever expect `direct_changes` to be truthy when `created` is false (otherwise, the evaluation was already created with all up-to-date fields), right?

 What do you think about this?

```python3
allow_changes = evaluation.state < Evaluation.State.Approved
direct_changes = update_with_changes(evaluation, defaults, dry_run=not allow_changes)
assert not direct_changes or created
change_log = self.statistics.updated_evaluations if direct_changes and allow_changes else self.statistics.attempted_changes
change_log.append(evaluation)
```

(No strong opinion on using a conditional assignment for change_log vs having an if-else with two `append` calls)

We can re-use `allow_changes` below and I'd try to use a similar flow for participants.

evap/staff/importers/json.py
```diff
@@ -106,14 +106,17 @@ class WarningMessage:
...
 @dataclass
 class ImportStatistics:
     name_changes: list[NameChange] = field(default_factory=list)
     new_courses: list[Course] = field(default_factory=list)
     new_evaluations: list[Evaluation] = field(default_factory=list)
     updated_courses: list[Course] = field(default_factory=list)
     updated_evaluations: list[Evaluation] = field(default_factory=list)
+    updated_participants: list[Evaluation] = field(default_factory=list)
     attempted_changes: list[Evaluation] = field(default_factory=list)
+    attempted_participant_changes: list[Evaluation] = field(default_factory=list)
```
richardebeling: let's rename `attempted_changes` to `attempted_evaluation_changes` now that we have two of them

evap/staff/importers/json.py
```diff
@@ -520,11 +537,8 @@ def _import_evaluation(self, course: Course, data: ImportEvent) -> Evaluation:
                 for lecturer in data["lecturers"]:
                     __, lecturer_created = self._import_contribution(evaluation, lecturer)
                     any_lecturers_changed |= lecturer_created
-
-            if not created and (direct_changes or participant_changes or any_lecturers_changed):
-                self.statistics.updated_evaluations.append(evaluation)
-        else:
-            self.statistics.attempted_changes.append(evaluation)
+                if not created and any_lecturers_changed and evaluation not in self.statistics.updated_evaluations:
+                    self.statistics.updated_evaluations.append(evaluation)
```
richardebeling: I think we should make the list a `set` so we don't have to do `and evaluation not in x` in user code (and so we cannot possibly get this wrong). Semantically, we "mark this evaluation as changed", it should just work if you do this multiple times.

evap/staff/tests/test_json_importer.py
```diff
@@ -702,10 +702,40 @@ def test_import_courses_evaluation_approved(self):
...
+        evaluation.vote_end_date = datetime.now().date() + timedelta(days=1)
+        evaluation.begin_evaluation()
+        evaluation.save()
+
+        self.assertEqual(evaluation.participants.count(), 0)
+
+        importer = self._import()
+
+        self.assertEqual(len(importer.statistics.updated_participants), 0)
+        self.assertEqual(len(importer.statistics.attempted_participant_changes), 1)
```
richardebeling, line 753: Can we extend this so it also asserts that `attempted_changes` (-> then `attempted_evaluation_changes`) is non-empty?

evap/staff/tools.py
```diff
@@ -441,16 +441,17 @@ def update_or_create_with_changes(
     return obj, False, changes
 
 
-def update_with_changes(obj: Model, defaults: dict[str, Any]) -> dict[str, tuple[Any, Any]]:
+def update_with_changes(obj: Model, defaults: dict[str, Any], test: bool = False) -> dict[str, tuple[Any, Any]]:
```
richardebeling: I'd suggest `dry_run` as the parameter name, but no strong opinion

evap/staff/importers/json.py
janno42: since one is now a set and the other one a list, i used if-else

richardebeling: I would have changed both lists into sets :D but that also works for me

janno42: about the assertion: can't we have these three cases?:
- created, no direct_changes
- not created, direct_changes
- not created, no direct_changes

richardebeling: You're right, it should be `assert not direct_changes or not created`

(We want to express: If we have direct changes, we must have not created a new instance -- I don't know of a better way in python to express `direct changes => not created` logically)

richardebeling (approved):

PR #2595, opened 2025-12-23
Automate CMS Import

janno42:
Automate the CMS import with a new management command that
- downloads the JSON file from the CMS
- imports the downloaded JSON file

### DIFF
--- a/evap/evaluation/tests/test_commands.py
+++ b/evap/evaluation/tests/test_commands.py
@@ -1,14 +1,16 @@
+import os
 import random
 from collections import defaultdict
 from datetime import date, datetime, timedelta
 from io import StringIO
 from itertools import chain, cycle
+from tempfile import TemporaryDirectory
 from unittest.mock import MagicMock, call, patch
 
 from django.conf import settings
 from django.contrib.auth.hashers import make_password
 from django.core import mail, management
-from django.core.management import CommandError
+from django.core.management import CommandError, call_command
 from django.db.models import Sum
 from django.test.utils import override_settings
 from model_bakery import baker
@@ -565,8 +567,8 @@ class TestSendTextanswerRemindersCommand(TestCase):
 
 class TestImportCMSData(TestCase):
     @patch("requests.get")
-    @patch("evap.staff.importers.json.JSONImporter")
-    def test_import_cms_data(self, mock_json_importer, mock_get):
+    @patch("evap.evaluation.management.commands.import_cms_data.JSONImporter")
+    def test_download_import(self, mock_json_importer, mock_get):
         semester = baker.make(Semester, cms_name="WS 25/26", default_course_end_date=date(2026, 2, 28))
         baker.make(Semester, cms_name="WS 2025")
         baker.make(Semester, default_course_end_date=date(2026, 2, 28))
@@ -576,3 +578,36 @@ class TestImportCMSData(TestCase):
 
         mock_get.assert_called_once_with("https://example.com/download?semester=WS%2025/26", timeout=120)
         mock_json_importer.assert_called_once_with(semester, semester.default_course_end_date)
+
+    @patch("evap.evaluation.management.commands.import_cms_data.JSONImporter.import_json")
+    def test_file_import(self, mock_import_json):
+        semester = baker.make(Semester)
+        with TemporaryDirectory() as temp_dir:
+            test_filename = os.path.join(temp_dir, "test.json")
+            with open(test_filename, "w", encoding="utf-8") as f:
+                f.write("example contents")
+            call_command(
+                "import_cms_data",
+                "file",
+                "--semester-id",
+                semester.id,
+                "--default-course-end-date",
+                "2000-01-01",
+                test_filename,
+                stdout=StringIO(),
+            )
+
+            mock_import_json.assert_called_once_with("example contents")
+
+            with self.assertRaises(CommandError) as cm:
+                call_command(
+                    "import_cms_data",
+                    "file",
+                    "--semester-id",
+                    semester.id + 42,
+                    "--default-course-end-date",
+                    "2000-01-01",
+                    test_filename,
+                    stdout=StringIO(),
+                )
+            self.assertEqual(cm.exception.args, ("Semester does not exist.",))
--- a/evap/staff/tests/test_json_importer.py
+++ b/evap/staff/tests/test_json_importer.py
@@ -1,14 +1,9 @@
 import json
-import os
 from copy import deepcopy
 from datetime import date, datetime, timedelta
-from io import StringIO
-from tempfile import TemporaryDirectory
-from unittest.mock import patch
 
 from django.conf import settings
 from django.core import mail
-from django.core.management import CommandError, call_command
 from django.test import TestCase, override_settings
 from model_bakery import baker
 from pydantic import ValidationError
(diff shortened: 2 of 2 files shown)


evap/evaluation/management/commands/import_cms_data.py
```diff
@@ -0,0 +1,59 @@
+import logging
+import urllib.parse
+
+import requests
```
janno42, line 7: how do I fix the mypy error (`Library stubs not installed for "requests")`) ?

evap/evaluation/management/commands/import_cms_data.py
```diff
@@ -0,0 +1,59 @@
...
+            logger.info("Processing %s.", semester.name_en)
+
+            semester_string = urllib.parse.quote(semester.cms_name)
+            url = settings.CMS_DATA_DOWNLOAD_URL.format(semester_string)
+            filename = "CMS.json"
+            download_successful = download(url, filename)
+            if not download_successful:
+                continue
+
+            import_data(semester, filename)
```
niklasmohrin: Do we need the custom chunking logic? Otherwise we can greatly simplify this by using the builtin .content / .text. We also don't need to write to a file, since we load all of it back into memory later anyways:
```suggestion
            for _ in range(RETRIES):
                try:
                    json_contents = requests.get(url, timeout=TIMEOUT).text
                    break
                except requests.exceptions.Timeout:
                    logger.warning("Download timed out: %s", url)
            else:
                logger.warning("Giving up.")
                continue

            JSONImporter(semester, semester.default_course_end_date).import_json(json_contents)
```

evap/evaluation/management/commands/import_cms_data.py
```diff
@@ -0,0 +1,59 @@
...
+from evap.evaluation.management.commands.tools import log_exceptions
+from evap.evaluation.models import Semester
+from evap.staff.importers.json import JSONImporter
+
+logger = logging.getLogger(__name__)
+
+
+@log_exceptions
+class Command(BaseCommand):
+    help = "Downloads the JSON file with the CMS data for a given semester and imports it."
```
niklasmohrin: What is the relation of this command to the `json_import` command we already have in staff? I feel like they should be merged, or one of them should go away.

Staff also feels like a more appropriate place than evaluation. I don't care too much about the directory, except that there shouldn't be two commands in two different places

evap/evaluation/models.py
```diff
@@ -74,6 +74,14 @@ class Semester(models.Model):
     short_name_en = models.CharField(max_length=20, unique=True, verbose_name=_("short name (english)"))
     short_name = translate(en="short_name_en", de="short_name_de")
 
+    cms_name = models.CharField(
+        max_length=20,
+        blank=True,
+        verbose_name=_("CMS name"),
+        help_text=_("If set together with default course end date, CMS data is imported automatically every day."),
+    )
+    default_course_end_date = models.DateField(blank=True, null=True, verbose_name=_("default course end date"))
```
niklasmohrin: Is the following true, or are we planning something else for this field?
```suggestion
    default_course_end_date = models.DateField(blank=True, null=True, verbose_name=_("default course end date for CMS import"))
```

niklasmohrin: Do we need both blank and null?

niklasmohrin (changes requested):

evap/evaluation/management/commands/import_cms_data.py
niklasmohrin: We need to add it to the ignore list tool.mypy.overrides in pyproject.toml, or even better, install types-requests in our development dependencies

Further, if we use requests, we should also declare it in our regular project.dependencies list in pyproject.toml - currently, it is only a transitive one

janno42: So far so good. Let's say we don't just ignore it: How do I add a development dependency?

evap/evaluation/management/commands/import_cms_data.py
janno42: `json_import` is used for manually importing an existing file, which is still needed. We could combine them into one file but one is used on the command line, and the other is called from a cron job. One takes arguments, and the other doesn't. I'm not sure if combining them would make things easier to understand.

The other commands called by cron jobs are in evaluation (`update_evaluation_states`, `send_reminders`), so this felt the right place to put it.

evap/evaluation/management/commands/import_cms_data.py
niklasmohrin: You need to add it to the dev dependency group in pyproject.toml (where we already have mypy, model-bakery, etc) and then update uv.lock (the command for that is in the wiki, I think at the end of Development Tips)

Afterwards you want to restart your nix develop shell and editor so that they pick up the updated uv.lock file

niklasmohrin: You need to add it to the dev dependency group in pyproject.toml (where we already have mypy, model-bakery, etc) and then update uv.lock (the command for that is in the wiki, I think at the end of Development Tips)

Afterwards you want to restart your nix develop shell and editor so that they pick up the updated uv.lock file

evap/evaluation/management/commands/import_cms_data.py
niklasmohrin: I see, knowing that, it makes sense. I still think that it would be better to have both commands in one file so that it is easier to keep the two invocations of the importer close together, but it's also not a big deal.

For example, I just noticed that we are still specifying a course end date in the import-file command instead of using the field on the semester - is that intended? I guess we should at least print a warning if a date is set on the semester.

For the interface, I would see something like

```bash
./manage.py import_cms_data --download # current import_cms_data
./manage.py import_cms_data --semester xyz --file f.json # current json_import
```

What do you think? If you want, I can also do the merge of the commands, this PR or a later one

janno42: Sure, please go ahead and merge them into one, thanks. We can do it in this PR if you'd like.

However, the manual import should still allow us to specify a default end date, which can differ from the one set for the semester. This would allow us to import a subset of courses, e.g., block seminars during the lecture-free period, with a different date. It could be optional, and we could use the semester default otherwise.

niklasmohrin: @janno42 I pushed a commit that merges the two and applies some of my suggested refactors. I also did two things:
1. changed the log statements to something that fit the current structure; feel free to change them however you like
2. added the download url as a command line flag instead; I think this would be slightly nicer to not "bloat" the settings, but you can also change it back if you want to put it into the settings instead

evap/evaluation/models.py
janno42: null is sufficient when setting the form field as not required

evap/evaluation/management/commands/import_cms_data.py
janno42: thanks, looks good! we can keep the url as command line flag.

evap/evaluation/management/commands/import_cms_data.py
```diff
@@ -0,0 +1,72 @@
...
+                    url = options["url"].format(urllib.parse.quote(semester.cms_name))
+                    for _ in range(RETRIES):
+                        try:
+                            json_contents = requests.get(url, timeout=TIMEOUT).text
+                            break
+                        except requests.exceptions.Timeout:
+                            logger.warning("Download timed out: %s", url)
+                    else:
+                        logger.warning("Giving up.")
+                        continue
```
richardebeling, line 53: Fyi: There is a way to get automatic retries with requests, but configuring it takes almost as much code as just implementing it: https://requests.readthedocs.io/en/latest/user/advanced/#example-automatic-retries

niklasmohrin: yeah I saw, but I figured we don't need the Enterprise Retry Strategy :tm:

richardebeling: Retry Factory Visitor Adapter Bridge

janno42:
Let the automation begin!

PR #2599, opened 2026-01-05
Add usage overview for questionnaires

### LINKED ISSUE #2585: Questionnaire usage overview
Currently, the staff questionnaire overview page shows how frequently a questionnaire is used. Sometimes, managers want to see the context in which a questionnaire was previously used, to help them decide whether it could be used for similar purposes. Currently, this is only possible via a database query.

A new page should be added that lists all uses of a questionnaire, grouped by semester, and links to all evaluations in which it is currently used. This page should open when the 'used N times' text on the questionnaire overview page is clicked.

hansegucker:
Close #2585

### DIFF
--- a/evap/staff/templates/staff_questionnaire_index_list.html
+++ b/evap/staff/templates/staff_questionnaire_index_list.html
@@ -30,7 +30,7 @@
                                     <br />
                                     {% blocktranslate count questionnaire.questions.count as count %}{{ count }} question{% plural %}{{ count }} questions{% endblocktranslate %},
                                     <a href="{% url 'staff:questionnaire_usage' questionnaire.id %}">
-                                        {% blocktranslate count questionnaire.contributions.count as count %}used {{ count }} time{% plural %}used {{ count }} times{% endblocktranslate %}
+                                        {% blocktranslate count count=questionnaire.contributions.count %}used {{ count }} time{% plural %}used {{ count }} times{% endblocktranslate %}
                                     </a>
                                 </td>
                                 <td>
--- a/evap/staff/templates/staff_questionnaire_usage.html
+++ b/evap/staff/templates/staff_questionnaire_usage.html
@@ -9,23 +9,22 @@
     {{ block.super }}
     <h3>{{ questionnaire.name }}</h3>
     {% for semester, evaluations in evaluations_grouped_by_semester.items %}
-        <div class="card mb-3 {% if semester.is_active %}card-outline-primary{% endif %}">
+        <div class="card mb-3 {% if semester.is_active %}card-outline-primary{% endif %}" data-id="semester-{{ semester.id }}">
             <div class="card-header">
-                {% blocktrans with semester_name=semester.name count n=evaluations|length %}
-                    {{ n }} usage in {{ semester_name }}
-                {% plural %}
-                    {{ n }} usages in {{ semester_name }}
-                {% endblocktrans %}
+                {{ semester.name }}
+                <span class="badge rounded-pill bg-dark">{{ evaluations|length }}</span>
             </div>
             <div class="card-body">
                 <table class="table table-seamless-links">
                     {% for evaluation in evaluations %}
                         <tr class="hover-row hover-row-info" data-url="{% url 'staff:evaluation_edit' evaluation.id %}">
-                            <td>{{ evaluation }}</td>
+                            <td {% if forloop.last %}class="border-bottom-0"{% endif %}>{{ evaluation }}</td>
                         </tr>
                     {% endfor %}
                 </table>
             </div>
         </div>
+    {% empty %}
+        <div class="alert alert-warning">{% translate "This questionnaire hasn't been used yet." %}</div>
     {% endfor %}
 {% endblock %}
--- a/evap/staff/tests/test_views.py
+++ b/evap/staff/tests/test_views.py
@@ -3429,9 +3429,10 @@ class TestQuestionnaireUsageView(WebTestStaffModeWith200Check):
     def setUpTestData(cls):
         cls.user = make_manager()
 
-        questionnaire = baker.make(Questionnaire)
-        cls.question = baker.make(Question, questionnaire=questionnaire)
-        cls.url = reverse("staff:questionnaire_usage", args=[questionnaire.pk])
+        cls.questionnaire = baker.make(Questionnaire)
+        cls.question = baker.make(Question)
+        cls.questionnaire.questions.add(cls.question)
+        cls.url = reverse("staff:questionnaire_usage", args=[cls.questionnaire.pk])
 
         cls.semesters = baker.make(
             Semester,
@@ -3442,8 +3443,8 @@ class TestQuestionnaireUsageView(WebTestStaffModeWith200Check):
         )
         cls.semester_without_questionnaire = baker.make(Semester, name_de="Semester 4", name_en="Semester 4")
 
-        cls.evaluations = {}
-        for semester in cls.semesters:
+    def test_questionnaire_usage_view(self):
+        for semester in self.semesters:
             course = baker.make(Course, semester=semester)
             evaluations = baker.make(
                 Evaluation,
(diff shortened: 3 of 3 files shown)


evap/staff/templates/staff_questionnaire_base.html
```diff
@@ -4,7 +4,11 @@
     {{ block.super }}
     {% if questionnaire %}
         <li class="breadcrumb-item"><a href="{% url 'staff:questionnaire_index' %}">{% translate 'Questionnaires' %}</a></li>
-        <li class="breadcrumb-item">{{ questionnaire.name }}</li>
+        {% if disable_breadcrumb_questionnaire %}
+            <li class="breadcrumb-item">{{ questionnaire.name }}</li>
+        {% else %}
+            <li class="breadcrumb-item"><a href="{% url 'staff:questionnaire_edit' questionnaire.id %}">{{ questionnaire.name }}</a></li>
+        {% endif %}
```
richardebeling, line 11: Why? I don't think the link is a problem, on the contrary, I think not having the link can only make it harder for people.

Not having the link requires additional logic in the template here, we need to add a new template argument, which we need to find a name for ("disable_breadcrumb_questionnaire" is rather misleading).

I'd say let's not go this way, just keep the link, or am I missing something?

evap/staff/views.py
```diff
@@ -1850,6 +1850,22 @@ def questionnaire_view(request, questionnaire_id):
...
 
+@manager_required
+def questionnaire_usage(request: HttpRequest, questionnaire_id: int) -> HttpResponse:
+    questionnaire = get_object_or_404(Questionnaire, id=questionnaire_id)
+
+    evaluations = Evaluation.objects.filter(contributions__questionnaires=questionnaire).order_by(
+        "-course__semester__created_at", "vote_start_datetime"
+    )
+
+    evaluations_grouped_by_semester = unordered_groupby((e.course.semester, e) for e in evaluations)
```
richardebeling, line 1879: N+1 queries pattern -- let's prefetch the semester above (should be relatively simple)

evap/staff/templates/staff_questionnaire_base.html
richardebeling: offline discussion: We _disable_ the link for the edit page that the link points to. We do this for consistency because that's what we do everywhere else.

Okay, fine with me

evap/staff/tests/test_views.py
```diff
@@ -3339,6 +3339,53 @@ def test_preview_change_language(self):
...
+            evaluations = baker.make(
+                Evaluation,
+                course=course,
+                name_de=iter([f"Evaluation 1 - {semester.name}", f"Evaluation 2 - {semester.name}"]),
+                name_en=iter([f"Evaluation 1 - {semester.name}", f"Evaluation 2 - {semester.name}"]),
+                _quantity=2,
+                _bulk_create=True,
+            )
+            for evaluation in evaluations:
+                evaluation.save()
```
richardebeling: let's just not use _bulk_create above if we do this anyway

evap/staff/tests/test_views.py
```diff
@@ -3339,6 +3339,53 @@ def test_preview_change_language(self):
...
+class TestQuestionnaireUsageView(WebTestStaffModeWith200Check):
+    @classmethod
+    def setUpTestData(cls):
+        cls.user = make_manager()
+
+        questionnaire = baker.make(Questionnaire)
+        cls.question = baker.make(Question, questionnaire=questionnaire)
+        cls.url = reverse("staff:questionnaire_usage", args=[questionnaire.pk])
+
+        cls.semesters = baker.make(
```
richardebeling, line 3437: Let's add a fourth semester without the questionnaire and assert below that it doesn't show up on the page

evap/staff/templates/staff_questionnaire_base.html
```diff
@@ -4,7 +4,11 @@
     {{ block.super }}
     {% if questionnaire %}
         <li class="breadcrumb-item"><a href="{% url 'staff:questionnaire_index' %}">{% translate 'Questionnaires' %}</a></li>
-        <li class="breadcrumb-item">{{ questionnaire.name }}</li>
+        {% if disable_breadcrumb_questionnaire %}
```
niklasmohrin, line 7: could we get rid of having to know this template detail in the view by writing something like
```
{% url '...' as questionnaire_edit_link %}
<li>{{ link_to_elsewhere questionnaire.name questionnaire_edit_link}}</li>
```
where link_to_elsewhere expands to `<a>...` or just the plain text depending on whether we are already at the link given?

There are some similar things here: https://stackoverflow.com/q/10263482/13679671

niklasmohrin: this was resolved, but no change?

evap/staff/templates/staff_questionnaire_index_list.html
```diff
@@ -29,7 +29,9 @@
                                     <strong class="questionnaire-name">{{ questionnaire.name }}</strong>
                                     <br />
                                     {% blocktranslate count questionnaire.questions.count as count %}{{ count }} question{% plural %}{{ count }} questions{% endblocktranslate %},
-                                    {% blocktranslate count questionnaire.contributions.count as count %}used {{ count }} time{% plural %}used {{ count }} times{% endblocktranslate %}
+                                    <a href="{% url 'staff:questionnaire_usage' questionnaire.id %}">
+                                        {% blocktranslate count questionnaire.contributions.count as count %}used {{ count }} time{% plural %}used {{ count }} times{% endblocktranslate %}
```
niklasmohrin: By the way, shouldn't this be `count count=questionnaire.contributions.count`?

see https://docs.djangoproject.com/en/6.0/topics/i18n/translation/#std-templatetag-blocktranslate

evap/staff/templates/staff_questionnaire_usage.html
niklasmohrin, line 1: can we add some info text if the questionnaire is not used anywhere?

evap/staff/tests/test_views.py
```diff
@@ -3339,6 +3339,54 @@ def test_preview_change_language(self):
...
+                name_en=iter([f"Evaluation 1 - {semester.name}", f"Evaluation 2 - {semester.name}"]),
+                _quantity=2,
+            )
+            for evaluation in evaluations:
+                evaluation.save()
+                evaluation.general_contribution.questionnaires.set([questionnaire])
+
+    def test_questionnaire_usage_view(self):
+        page = self.app.get(self.url, user=self.user)
+        content = page.body.decode()
```
niklasmohrin:
```suggestion
        content = self.app.get(self.url, user=self.user).text
```

evap/staff/tests/test_views.py
```diff
@@ -3339,6 +3339,54 @@ def test_preview_change_language(self):
...
+
+        indexes = []
+        for semester in sorted(self.semesters, key=lambda s: s.created_at, reverse=True):
+            indexes.append(content.index(f"2 usages in {semester.name}"))
+            indexes += [
+                content.index(evaluation.full_name)
+                for evaluation in Evaluation.objects.filter(course__semester=semester).order_by("vote_start_datetime")
+            ]
+
+        self.assertListEqual(indexes, sorted(indexes))
```
niklasmohrin: if they are both lists, this method will automatically selected:
```suggestion
        self.assertEqual(indexes, sorted(indexes))
```

niklasmohrin (commented):
Looks mostly good, thanks!

evap/staff/templates/staff_questionnaire_base.html
hansegucker: Yeah, I think we spoke about this in person and decided not to change this, as this is the way it's done in EvaP in other places, too.

niklasmohrin: Hm, I don't recall talking about this before today, but whatever. I feel that we should try to not repeat a messy approach from before and instead check if we can do the nicer thing, but I am also fine for doing this in another issue. The questionnaire base template with its `if questionnaire` is already a bit hacky, I don't really want to extend this hack into views when the UI is conceptually very simple

richardebeling: Created https://github.com/e-valuation/EvaP/issues/2642, marking as resolved here

evap/staff/tests/test_views.py
richardebeling: Now that we've removed _bulk_create=True, the explicit `.save()` calls should be noops, baker should have already called `.save()` on every instance, right?

evap/staff/tests/test_views.py
richardebeling: 🤯 I've been writing `.body.decode()` like a stone age man for the last decade

evap/staff/templates/staff_questionnaire_usage.html
niklasmohrin, line 1: It looks a bit strange that all rows have a bottom border, but there is no top border. @janno42 can decide how this should look

evap/staff/templates/staff_questionnaire_usage.html
```diff
@@ -0,0 +1,33 @@
...
+    {{ block.super }}
+    <h3>{{ questionnaire.name }}</h3>
+    {% for semester, evaluations in evaluations_grouped_by_semester.items %}
+        <div class="card mb-3 {% if semester.is_active %}card-outline-primary{% endif %}">
+            <div class="card-header">
+                {% blocktrans with semester_name=semester.name count n=evaluations|length %}
+                    {{ n }} usage in {{ semester_name }}
+                {% plural %}
+                    {{ n }} usages in {{ semester_name }}
+                {% endblocktrans %}
```
janno42: In the header, please only include the semester name followed by a `badge rounded-pill` with the number of usages, as seen on the filter buttons on the semester index page.

evap/staff/templates/staff_questionnaire_usage.html
janno42: please remove the bottom border on the last `tr`

PR #2600, opened 2026-01-07
Restrict automated changes to new courses and evaluations

janno42:
changes to courses and evaluations should only be made by the cms importer as long as they are in preparation

### DIFF
--- a/evap/staff/importers/json.py
+++ b/evap/staff/importers/json.py
@@ -382,6 +382,7 @@ class JSONImporter:
 
         if course and course.evaluations.exclude(state=Evaluation.State.NEW).exists():
             self.statistics.attempted_course_changes.add(course)
+            # the course itself should not be updated but is still added to courses_by_gguid so that its evaluations will be processed
             self.courses_by_gguid[data["gguid"]] = course
             return course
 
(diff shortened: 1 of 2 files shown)


evap/staff/importers/json.py
```diff
@@ -373,6 +375,16 @@ def _import_course(self, data: ImportEvent, course_type: CourseType | None = Non
...
             return None
 
+        try:
+            course = Course.objects.get(cms_id=data["gguid"])
+        except Course.DoesNotExist:
+            course = None
+
+        if course and course.evaluations.exclude(state=Evaluation.State.NEW).exists():
+            self.statistics.attempted_course_changes.add(course)
+            self.courses_by_gguid[data["gguid"]] = course
```
richardebeling, line 386: In ~632, we have
```python3
            # Don't import if course was skipped
            course = self.courses_by_gguid.get(event["relatedevents"][0]["gguid"])
            if course is None:
                continue
```

Does adding the course to `courses_by_gguid` here make sense, given that this will make the logic below think the course was imported?

richardebeling (commented):
Change generally looks good to me, but I'm not yet convinced that we want to add the course to `self.courses_by_gguid`

evap/staff/tests/test_json_importer.py
```diff
@@ -686,6 +686,39 @@ def test_import_ignore_users(self):
...
+        evaluation.is_rewarded = False
+        evaluation.save()
+        evaluation.course.name_en = "Change"
+        evaluation.course.save()
+
+        importer = self._import()
+
+        evaluation = Evaluation.objects.get(pk=evaluation.pk)
+        self.assertFalse(evaluation.is_rewarded)
+        self.assertEqual(evaluation.course.name_en, "Change")
```
niklasmohrin, line 718: Does the following work, or does it also break because of #2576 ?
```suggestion
        with assert_no_database_modifications():
            importer = self._import()
```

evap/staff/tests/test_json_importer.py
```diff
@@ -686,6 +686,39 @@ def test_import_ignore_users(self):
...
+        evaluation.course.save()
+
+        importer = self._import()
+
+        evaluation = Evaluation.objects.get(pk=evaluation.pk)
+
+        self.assertTrue(evaluation.is_rewarded)
+        self.assertEqual(evaluation.course.name_en, "Process-oriented information systems")
+        self.assertEqual(len(importer.statistics.attempted_evaluation_changes), 0)
+        self.assertEqual(len(importer.statistics.attempted_course_changes), 0)
```
niklasmohrin:
```suggestion
        self.assertEqual(importer.statistics.attempted_evaluation_changes, set())
        self.assertEqual(importer.statistics.attempted_course_changes, set())
```

evap/staff/importers/json.py
janno42: If we don't add it, no evaluations will be imported/updated for the course at all, which is not what we want. Here, no changes will be made to the course in the database, and the importer will continue processing the evaluations as usual.

janno42: Skipping is intended for course types that we don't want to import in the first place.

evap/staff/importers/json.py
```diff
@@ -551,7 +563,7 @@ def _import_evaluation(  # noqa: PLR0912, PLR0915
                 **defaults,
             )
 
-        allow_changes = evaluation.state < Evaluation.State.APPROVED
+        allow_changes = evaluation.state == Evaluation.State.NEW
```
niklasmohrin: if we are touching these lines anyways, can we rename to `allow_evaluation_changes`, `allow_participant_changes`, and `allow_lecturer_changes` (or `allow_contributor_changes`)?

niklasmohrin (commented):
looks good, except for what richard said

evap/staff/importers/json.py
niklasmohrin: can you add a short comment about this on the line here?

evap/staff/tests/test_json_importer.py
janno42: yes, we have an empty log entry there

niklasmohrin (approved):

PR #2602, opened 2026-01-12
IgnoredEvaluations, evaluation notes, CMS link icons

janno42:
We need to keep track of which evaluations have been deleted so that we don't import them again -> `IgnoredEvaluation`
I'm happy to discuss the way it's implemented and naming; I don't have any strong feelings about them.

It would also help to have manager notes for evaluations, such as who said an evaluation could be deleted and when or any other things to remember for later preparation -> `Evaluation.notes`
I'd also like to see which courses and evaluations are linked to the CMS -> link icons


evap/evaluation/models.py
```diff
@@ -536,6 +536,8 @@ class State(models.IntegerChoices):
         verbose_name=_("campus management system id"), blank=True, null=True, unique=True, max_length=255
     )
 
+    staff_notes = models.TextField(verbose_name=_("staff_notes"), blank=True)
```
```suggestion
    staff_notes = models.TextField(verbose_name=_("staff notes"), blank=True)
```
or even something like "notes by staff users"?

evap/cms/management/commands/__init__.py
niklasmohrin, line 1: we also need a `__init__.py` file in the `management` directory, see #2607

evap/cms/json_importer.py
```diff
@@ -28,6 +28,8 @@
 from evap.evaluation.tools import clean_email
 from evap.staff.tools import update_m2m_with_changes, update_or_create_with_changes, update_with_changes
 
+from .models import IgnoredEvaluation
```
niklasmohrin:
```suggestion
from evap.cms.models import IgnoredEvaluation
```

evap/cms/models.py
```diff
@@ -0,0 +1,31 @@
...
+    # unique reference for import from campus management system
+    cms_id = models.CharField(
+        verbose_name=_("campus management system id"), blank=True, null=True, unique=True, max_length=255
+    )
+
+    name_de = models.CharField(max_length=1024, verbose_name=_("name (german)"), blank=True)
+    name_en = models.CharField(max_length=1024, verbose_name=_("name (english)"), blank=True)
+    name = translate(en="name_en", de="name_de")
+
+    course = models.ForeignKey(Course, models.CASCADE, verbose_name=_("course"), related_name="ignored_evaluations")
```
niklasmohrin, line 18: Doesn't that mean that if delete the entire course then all the `IgnoredEvaluation`s would also be deleted and we would import everything again? I am okay with it, just want to avoid surprises in the future :D

evap/cms/models.py
```diff
@@ -0,0 +1,31 @@
...
+from evap.evaluation.models import Course
+from evap.evaluation.tools import translate
+
+
+class IgnoredEvaluation(models.Model):
+    """Model for an evaluation that was deleted and should not be imported again from the CMS"""
+
+    # unique reference for import from campus management system
+    cms_id = models.CharField(
+        verbose_name=_("campus management system id"), blank=True, null=True, unique=True, max_length=255
```
niklasmohrin: For this model we should do the following, no?
```suggestion
        verbose_name=_("campus management system id"), unique=True, max_length=255
```

evap/evaluation/models.py
janno42: i don't have strong feelings about this but would choose shorter names if in doubt

evap/cms/json_importer.py
janno42: i assume you then also want to have this changed in the other two places where i imported from .something :)

evap/cms/models.py
janno42: Correct, the course cannot be deleted if we want to keep the information, but that's not a problem. Usually, we only delete some evaluations of a course, not all of them.

evap/cms/json_importer.py
niklasmohrin: yes :) we should also just add a lint

richardebeling (approved):
I have all files marked as "viewed" now, all code looked good to me at some point, but I'm not sure if everything is consistent within the "new app" approach now. Let me know if I should re-review or something

janno42:
The new app is currently on a separate branch and almost ready for its own PR. I'm just waiting for this PR to be merged :)
If you prefer a different approach and want to do everything in one PR, let me know.

evap/staff/templates/staff_evaluation_form.html
```diff
@@ -67,6 +67,9 @@ <h3 class="pt-2">
             </form>
         </div>
         <div class="col-12 col-xl-4 mh-100 overflow-auto">
+            <div class="mb-2">
+                <textarea id="{{ evaluation_form.staff_notes.id_for_label }}" name="{{ evaluation_form.staff_notes.name }}" rows="3" class="form-control" form="evaluation-form" placeholder="{% translate 'Notes' %}">{{ evaluation_form.staff_notes.value }}</textarea>
```
niklasmohrin: There is on way to update the notes if the evaluation itself cannot be updated anymore, but the textarea here is still editable. Should we disable it?

niklasmohrin (commented):
still need to check with a test file

evap/evaluation/models.py
niklasmohrin: I reopened this for the `_` -> ` ` change in the potentially user visible text

evap/cms/tests/test_json_importer.py
```diff
@@ -20,7 +21,6 @@
     UserProfile,
 )
 from evap.evaluation.models_logging import LogEntry
-from evap.staff.importers.json import ImportDict, JSONImporter, NameChange, WarningMessage, _clean_whitespaces
 
 EXAMPLE_DATA: ImportDict = {
```
niklasmohrin, line 25: note (to self): depending on whether we merge this or #2634 first, we need to move the file and update the module name in the import then

evap/cms/json_importer.py
```diff
@@ -426,7 +427,11 @@ def _import_course_from_unused_exam(self, data: ImportEvent) -> Course | None:
     # pylint: disable=too-many-locals
     def _import_evaluation(  # noqa: PLR0912, PLR0915
         self, course: Course, data: ImportEvent, earliest_exam_date: date | None = None
-    ) -> Evaluation:
+    ) -> Evaluation | None:
+        # Don't import ignored evaluations again
+        if IgnoredEvaluation.objects.filter(cms_id=data["gguid"]).exists():
+            return None
```
niklasmohrin, line 433: Not blocking, but we should add a note here that this evaluation was ignored; I guess a warning on `self.statistics` is a bit much, maybe ImportStatistics could get a new `info_messages` field?

I think some of our young contributors can figure this one out :)

PR #2606, opened 2026-01-12
Fix Evaluation Order

### LINKED ISSUE #2583: Fix evaluation order
Ordering evaluations on the staff semester page, results page, student index page, ... currently is case-sensitive and places umlauts (ÄÖÜ) after Z.
Instead, evaluations should be ordered case-insensitive and umlauts should be handled the same as their original vowel (DIN 5007 Alternative 1).

geromequa:
Closes #2583.

Updates `sortRows` inside the `datagrid.ts` component, which is utilized in all relevant views (see #2583). Initial sorting on `staff/semester` view was removed. Other initial sorts based on other values than 'full_name' were left untouched.

### DIFF
--- a/evap/staff/tests/test_live.py
+++ b/evap/staff/tests/test_live.py
@@ -264,7 +264,7 @@ class TextAnswerEditLiveTest(LiveServerTest):
 
         baker.make(
             TextAnswer,
-            question=question1,
+            assignment__question=question1,
             contribution=contribution1,
             answer=iter(f"this is a dummy answer {i}" for i in range(3)),
             original_answer=None,
@@ -274,7 +274,7 @@ class TextAnswerEditLiveTest(LiveServerTest):
 
         textanswer1 = baker.make(
             TextAnswer,
-            question=question1,
+            assignment__question=question1,
             contribution=contribution1,
             answer="this answer will be edited",
             original_answer=None,
@@ -283,7 +283,7 @@ class TextAnswerEditLiveTest(LiveServerTest):
 
         baker.make(
             TextAnswer,
-            question=question1,
+            assignment__question=question1,
             contribution=contribution1,
             answer=iter(f"this is a dummy answer {i}" for i in range(3, 6)),
             original_answer=None,


evap/static/ts/src/datagrid.ts
```diff
@@ -172,6 +172,18 @@ abstract class DataGrid {
         this.renderToDOM();
     }
 
+    // Applies DIN 5007 Alternative 1 to string
+    private normalizeValue(value: string) {
+        let normalized = value.toLowerCase();
+
```
styrix560: I would chain these together as in Python

evap/static/ts/src/datagrid.ts
```diff
@@ -189,9 +201,15 @@ abstract class DataGrid {
 
         this.rows.sort((a, b) => {
             for (const [column, order] of this.state.order) {
-                if (a.orderValues.get(column)! < b.orderValues.get(column)!) {
+                let valueA = a.orderValues.get(column);
+                let valueB = b.orderValues.get(column);
+                if (typeof valueA === "string" && typeof valueB === "string") {
```
styrix560: If there exists and invariant, that either both or neither of the values are strings, I would only check one of these conditions and assert the other.

evap/static/ts/src/datagrid.ts
```diff
@@ -189,9 +201,15 @@ abstract class DataGrid {
...
+                    valueA = this.normalizeValue(valueA);
+                    valueB = this.normalizeValue(valueB);
+                }
+                if (valueA! < valueB!) {
                     return order === "asc" ? -1 : 1;
-                } else if (a.orderValues.get(column)! > b.orderValues.get(column)!) {
+                } else if (valueA! > valueB!) {
```
styrix560, line 202: I would add a default return of 0 and _if you feel like it_ remove the else

geromequa: This is already the case. See line 209

evap/evaluation/models.py
```diff
@@ -1119,6 +1081,9 @@ def unlogged_fields(self):
             "dropout_count",
         ]
 
+    def full_name_ordering_key(self):
```
hansegucker:
```suggestion
    def full_name_ordering_key(self) -> str:
```

evap/evaluation/models.py
```diff
@@ -1081,6 +1081,9 @@ def unlogged_fields(self):
...
+        return self.full_name.lower().replace("ä", "a").replace("ö", "o").replace("ü", "u").replace("ß", "ss")
```
richardebeling: Let's not hardcode this information, but use https://docs.python.org/3/library/locale.html#locale.strcoll instead

evap/static/ts/src/datagrid.ts
```diff
@@ -172,6 +172,11 @@ abstract class DataGrid {
         this.renderToDOM();
     }
 
+    // Applies DIN 5007 Alternative 1 to string
+    private normalizeValue(value: string) {
+        return value.toLowerCase().replace("ä", "a").replace("ü", "u").replace("ö", "o").replace("ß", "ss");
```
richardebeling: For javascript, this is probably the way to go: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

evap/evaluation/models.py
niklasmohrin: The sad thing about Python's locale module is that it depends on the global locale setting which we might want to change depending on the request language. I haven't checked, but it seems like Django does not update this as part of their LocaleMiddleware.

Alternatively, we could use postgres locale sorting:
- https://docs.djangoproject.com/en/6.0/ref/models/database-functions/#collate
- https://www.postgresql.org/docs/current/collation.html

As far as I can tell, the correct sorting string would be something like "de-x-icu", I don't really know what sorting this summons, but it is what it says in the documentation. PR author should find this out for us :)

So it looks to me like postgres sorting is better because it does not depend on global state.

evap/static/ts/src/datagrid.ts
niklasmohrin: note the comment there about reusing one Collator instance when doing a lot of comparisons; this would be the case here

evap/evaluation/models.py
```diff
@@ -1119,6 +1119,9 @@ def unlogged_fields(self):
             "dropout_count",
         ]
 
+    def full_name_ordering_key(self) -> str:
```
richardebeling: now unused? If so, please remove

evap/static/ts/src/datagrid.ts
```diff
@@ -172,6 +172,11 @@ abstract class DataGrid {
         this.renderToDOM();
     }
 
+    // Applies DIN 5007 Alternative 1 to string
+    private normalizeValue(value: string) {
```
richardebeling: same, looks unused now -> should probably be removed

evap/staff/tests/test_views.py
```diff
@@ -733,32 +733,32 @@ def setUpTestData(cls):
...
             name_en="Evaluation 1",
-            course=baker.make(Course, name_de="A", name_en="B", semester=cls.semester),
+            course=baker.make(Course, name_de="A", name_en="Z", semester=cls.semester),
         )
         baker.make(
             Evaluation,
             name_de="Evaluation 2",
             name_en="Evaluation 2",
-            course=baker.make(Course, name_de="B", name_en="A", semester=cls.semester),
+            course=baker.make(Course, name_de="Ä", name_en="Ü", semester=cls.semester),
```
richardebeling: I think the added lines here are leftovers from the previous approach and no longer needed now

evap/staff/tests/test_live.py
```diff
@@ -178,6 +178,89 @@ def test_collapse_without_editor_approved(self) -> None:
...
+            name_de="Evaluation 4",
+            name_en="Evaluation 4",
+            course=baker.make(Course, name_de="ÜB", name_en="ÄB", semester=test_semester),
+        )
+        baker.make(
+            Evaluation,
+            name_de="Evaluation 5",
+            name_en="Evaluation 5",
+            course=baker.make(Course, name_de="Z", name_en="AE", semester=test_semester),
+        )
```
richardebeling, line 193:
```suggestion
        baker.make(
            Evaluation,
            _quantity=5,
            name_de=iter(f"Evaluation {i}" for i in range(1, 6)),
            name_en=iter(f"Evaluation {i}" for i in range(1, 6)),
            course__name_de=iter(("AE", "ÄB", "UE", "ÜB", "Z")),
            course__name_en=iter(("Z", "ÜB", "UE", "ÄB", "AE")),
            course__semester=test_semester
        )
```

evap/staff/tests/test_live.py
```diff
@@ -178,6 +178,89 @@ def test_collapse_without_editor_approved(self) -> None:
...
+
+        with self.enter_staff_mode():
+            self.selenium.get(self.live_server_url + reverse("staff:index"))
```
richardebeling: We should have some live server test helper for building the URL, something like `self.reverse()`, other tests should be using that.

evap/staff/tests/test_live.py
```diff
@@ -178,6 +178,89 @@ def test_collapse_without_editor_approved(self) -> None:
...
+
+            self.selenium.get(self.live_server_url + reverse("staff:index"))
+
+            language_en_button = self.selenium.find_element(
+                By.XPATH,
+                "//form[@action='/set_lang']//button[@data-set-spinner-icon='span-set-language-en']//parent::form",
+            )
+
+            language_en_button.submit()
+            self.selenium.get(self.live_server_url + reverse("staff:semester_view", args=[test_semester.id]))
```
richardebeling: Clicking the button should reload the page, an extra get request shouldn't be necessary. If it is necessary, we should fix that.

I think we should also pull out a "set language" helper, duplicating the xpath `//form[@action='/set_lang']//button[@data-set-spinner-icon='span-set-language-en']//parent::form` across tests will lead to pain in the future I think

evap/staff/tests/test_live.py
```diff
@@ -178,6 +178,89 @@ def test_collapse_without_editor_approved(self) -> None:
...
+
+            table = self.selenium.find_element(By.ID, "evaluation-table").find_elements(
+                By.XPATH, "//tbody//child::td[@data-col='name']"
+            )
+
+            self.assertEqual(table[0].get_attribute("data-order"), "ÄB – Evaluation 2")
+            self.assertEqual(table[1].get_attribute("data-order"), "AE – Evaluation 1")
+            self.assertEqual(table[2].get_attribute("data-order"), "ÜB – Evaluation 4")
+            self.assertEqual(table[3].get_attribute("data-order"), "UE – Evaluation 3")
+            self.assertEqual(table[4].get_attribute("data-order"), "Z – Evaluation 5")
```
richardebeling: It would probably make sense to use sample inputs that codify whether an umlaut is treated as equal to the non-umlaut character or not. Currently, both approaches would pass this test.

evap/staff/tests/test_live.py
geromequa: If you're referring to the redirects in lines 196 and 217, then yes. We need to wait for the page to fully load. There was no easy to find element to verify this, so the redirect was suggested by @hansegucker as a workaround. Also waiting for a random element seemed just as "hacky" as this solution.

Any suggestions on how to fix this, would be appreciated.

evap/static/ts/src/datagrid.ts
geromequa: Done. Should this have been caught by the linter? pylance in my VS Code highlighted it as unused, so the CI should have caught this IMHO

evap/staff/tests/test_live.py
geromequa: Would this not be caught by, e.g., "ÄB" < "AE"? If "Ä" was not treated as "A" (so either as "AE" or just as a separate character sorted after "Z") then "ÄB" > "AE". Same for the analog test cases for "ÜB" and "UE".

evap/staff/tests/test_live.py
geromequa: language change now lives in `set_page_language` helper in `LiveServerTest` Class.

evap/staff/tests/test_live.py
richardebeling: My point was: The comparison between an umlaut and a non-umlaut could have two results, based on the current tests:
1. The umlaut could sort before the non-umlaut ('Ä' < 'A'). Then, the ordering we assert in the test would hold: 'ÄB' < 'AE' and 'ÜB' < 'UE'. In this scenario, the second character in each string is not relevant, we would also have 'ÄZ' < 'AA'.
2. The umlaut could sort equal to the non-umlaut ('Ä' = 'A'). Then, the ordering we assert here would also still hold, since 'ÄB' = 'AB' < 'AE' and 'ÜB' = 'UB' < 'UE'. In this scenario, we arrive at this sorting only because of the second character.

So from my view, there's two possible behaviors that would make the test pass. A situation where they would behave differently is if you have the strings 'ÄC' and 'AB': What is their order?

(If the replacement 'Ä' -> 'AE' is in any way relevant to the sorting logic, we should also build test cases around this, e.g. 'AEA' < 'ÄB' < 'AEC')

evap/staff/tests/test_live.py
richardebeling: Helper is good, we can do more ugly stuff in there is we just don't know a better way since it's much easier to consistently replace in the future, then 👍

evap/evaluation/tests/tools.py
```diff
@@ -358,6 +358,15 @@ def setUpClass(cls) -> None:
...
         cls.selenium.set_window_size(*cls.window_size)
 
+    @contextmanager
+    def set_page_language(self, language: str):
+        language_button = self.selenium.find_element(
+            By.XPATH,
+            f"//form[@action='/set_lang']//button[@data-set-spinner-icon='span-set-language-${language}']//parent::form",
+        )
+
+        language_button.submit()
```
richardebeling: So if I understand https://github.com/e-valuation/EvaP/pull/2606#discussion_r2784178018 correctly, this currently does not wait for the page to be reloaded, right?

Looking at the other methods above this, `wait_until_page_reloads` seems very promising to me for this use case. Should we just do
```python3
with self.wait_until_page_reloads():
    language_button.submit();
```
or does this not work for some reason?

evap/evaluation/tests/tools.py
```diff
@@ -358,6 +358,15 @@ def setUpClass(cls) -> None:
         super().setUpClass()
         cls.selenium.set_window_size(*cls.window_size)
 
+    @contextmanager
```
richardebeling: This isn't a context manager

```suggestion
```

evap/static/ts/src/datagrid.ts
richardebeling: Very good point. Eslint has [no-unused-private-class-members](https://archive.eslint.org/docs/rules/no-unused-private-class-members), which should catch this. Our eslint config [derives from](https://github.com/e-valuation/EvaP/blob/9f10df883c5ece5a6659377e3a621767d847823a/evap/static/ts/eslint.config.js#L7) `eslint.configs.recommended`, for which I would have thought that it enables this rule, based on their source code [here](https://github.com/eslint/eslint/blob/8330d238ae6adb68bb6a1c9381e38cfedd990d94/packages/js/src/configs/eslint-recommended.js#L63).

Can you either open a PR that fixes this, or create an issue with a minimum reproduction (where to add a private method, so that, when running eslint, it still passes)?

niklasmohrin: we want to activate this: https://typescript-eslint.io/rules/no-unused-private-class-members/

evap/staff/tests/test_live.py
geromequa: I believe the first case ('Ä' < 'A') is not possible because there is no possible replacement, that implies this case. 'Ä' is either replaced by 'A', 'AE' or left as 'Ä' and sorted after 'Z'.

The testcase 'AE' vs. 'ÄB' covers 'A' -> 'AE' based on the order those two are sorted in the test.

Using the evaluation with 'Z' we ensure that an umlaut is not sorted after 'Z'.

'Ä'='A' is exactly the case we want.

richardebeling: > is not possible because there is no possible replacement

But this is an implementation detail :D My point is to make the test generic for all possible implementations, so that we don't have to change it even if we swap out the implementation.

> 'Ä'='A' is exactly the case we want.

Then let's assert this in the test, simply by having `AA`, `ÄB` and `AC` and asserting that they are ordered in this order.

geromequa: I moved some stuff around: AA, ÄB and AC are now asserted to be in that order.
UB < ÜC makes sure, that U=Ü and not Ü<U. This of course assumes that all umlauts are sorted the same relatively speaking.

evap/staff/tests/test_live.py
```diff
@@ -178,6 +177,63 @@ def test_collapse_without_editor_approved(self) -> None:
...
+            _quantity=7,
+            name_de=iter(f"Evaluation {i}" for i in range(1, 8)),
+            name_en=iter(f"Evaluation {i}" for i in range(1, 8)),
+            course__name_de=iter(("AA", "ÄB", "AC", "AE", "UB", "ÜC", "Z")),
+            course__name_en=iter(("Z", "ÜC", "UB", "AE", "AC", "ÄB", "AA")),
+            course__semester=test_semester,
+        )
+
+        with self.enter_staff_mode():
+            # self.selenium.get(self.reverse("staff:index"))
```
styrix560: We should remove this right?

evap/staff/tests/test_live.py
```diff
@@ -178,6 +177,63 @@ def test_collapse_without_editor_approved(self) -> None:
...
+            self.selenium.get(self.reverse("staff:semester_view", args=[test_semester.id]))
+            with self.wait_until_page_reloads():
+                self.set_page_language("de")
+
+            # self.wait.until(visibility_of_element_located((By.CSS_SELECTOR, "#evaluation-table .col-order-asc")))
```
styrix560: Same as above, remove this?

evap/staff/tests/test_live.py
```diff
@@ -178,6 +177,63 @@ def test_collapse_without_editor_approved(self) -> None:
...
+
+            self.assertEqual(table[0].get_attribute("data-order"), "AA – Evaluation 1")
+            self.assertEqual(table[1].get_attribute("data-order"), "ÄB – Evaluation 2")
+            self.assertEqual(table[2].get_attribute("data-order"), "AC – Evaluation 3")
+            self.assertEqual(table[3].get_attribute("data-order"), "AE – Evaluation 4")
+            self.assertEqual(table[4].get_attribute("data-order"), "UB – Evaluation 5")
+            self.assertEqual(table[5].get_attribute("data-order"), "ÜC – Evaluation 6")
+            self.assertEqual(table[6].get_attribute("data-order"), "Z – Evaluation 7")
+
+            # self.selenium.get(self.reverse("staff:index"))
```
styrix560: Remove this

evap/staff/tests/test_live.py
```diff
@@ -178,6 +177,63 @@ def test_collapse_without_editor_approved(self) -> None:
...
+            with self.wait_until_page_reloads():
+                self.set_page_language("en")
+
+            # self.selenium.get(self.reverse("staff:semester_view", args=[test_semester.id]))
```
styrix560: Remove this

evap/staff/tests/test_live.py
```diff
@@ -178,6 +177,63 @@ def test_collapse_without_editor_approved(self) -> None:
...
+            self.selenium.get(self.reverse("staff:semester_view", args=[test_semester.id]))
+            with self.wait_until_page_reloads():
+                self.set_page_language("de")
+
+            # self.wait.until(visibility_of_element_located((By.CSS_SELECTOR, "#evaluation-table .col-order-asc")))
+            table = self.selenium.find_element(By.ID, "evaluation-table").find_elements(
+                By.XPATH, "//tbody//child::td[@data-col='name']"
+            )
+
+            self.assertEqual(table[0].get_attribute("data-order"), "AA – Evaluation 1")
```
styrix560: I would probably do this in a loop. Something like:
```
order = ["AA - Evaluation 1", "ÄB - Evaluation 2", ...]

for i in range(8):
	self.assertEqual(table[i].get_attribute("data-order"), order[i])
```

evap/staff/tests/test_live.py
```diff
@@ -178,6 +177,63 @@ def test_collapse_without_editor_approved(self) -> None:
...
+        )
+
+        with self.enter_staff_mode():
+            # self.selenium.get(self.reverse("staff:index"))
+            self.selenium.get(self.reverse("staff:semester_view", args=[test_semester.id]))
+            with self.wait_until_page_reloads():
+                self.set_page_language("de")
+
+            # self.wait.until(visibility_of_element_located((By.CSS_SELECTOR, "#evaluation-table .col-order-asc")))
+            table = self.selenium.find_element(By.ID, "evaluation-table").find_elements(
```
styrix560: I think these are `table_rows` right? Or `table_items`?

evap/static/ts/src/datagrid.ts
```diff
@@ -187,11 +187,19 @@ abstract class DataGrid {
             header.classList.add(`col-order-${ordering}`);
         }
 
+        const Collator = new Intl.Collator(document.documentElement.lang, { caseFirst: "false" });
```
styrix560: Is there a reason, why this variable uses UpperCamelCase instead of lowerCamelCase?

styrix560 (commented):
Looks generally good, I just mentioned some things

evap/staff/tests/test_live.py
```diff
@@ -178,6 +177,69 @@ def test_collapse_without_editor_approved(self) -> None:
...
+            name_en=iter(f"Evaluation {i}" for i in range(1, 8)),
+            course__name_de=iter(("AA", "ÄB", "AC", "AE", "UB", "ÜC", "Z")),
+            course__name_en=iter(("Z", "ÜC", "UB", "AE", "AC", "ÄB", "AA")),
+            course__semester=test_semester,
+        )
+
+        with self.enter_staff_mode():
+            self.selenium.get(self.reverse("staff:semester_view", args=[test_semester.id]))
+            with self.wait_until_page_reloads():
+                self.set_page_language("de")
```
richardebeling: I don't think there is ever a use case to call `set_page_language` without wrapping it in `wait_until_page_reloads` -- in that case, I would really just move the "wait for page reload" into the "set page language"

PR #2607, opened 2026-01-12
Disallow PEP 420 implicit namespace packages

### LINKED ISSUE #2598: Linters not runnnig due to missing `__init__.py` files
I just noticed that pylint is not running on evap/staff/management/commands/json_import.py. The reason is that `evap.staff.management` is not a regular package, but instead an implicit namespace package, because it doesn't contain a `__init__.py` file.

Some tools also have options to traverse into namespace packages, I will check whether we should enable those anyways.

niklasmohrin:
Close #2598

It turns out that the one case I mentioned in the issue is fixed in 897fbed8c614e8859c6a5465e3bfbb42b6f88f1d, except that we forgot to delete the `__init__.py` file of the empty module.

Also, the ruff check would not have detected it right now, but that is fixed in preview mode right now. I would just leave it as is and expect that ruff will start detecting this correctly some time in the future.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -81,7 +81,7 @@ line-length = 120
 
 [tool.ruff.lint]
 # Those are interesting, but not yet addressed: PYI,Q,SIM,PTH,TRY,RUF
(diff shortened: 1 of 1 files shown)


richardebeling (approved):
oopsie whoopsie

PR #2608, opened 2026-01-12
Replace bootstrap-datetimepicker

### LINKED ISSUE #2546: Replace Datetime Picker
We currently use bootstrap-datetimepicker for date and datetime form fields. It requires jQuery and causes problems when used in modals. For the modal to create an exam evaluation we had to use additional logic to use the browser's native date picker because the bootstrap-datetimepicker did not work.

Ideally, we would replace it with an alternative that doesn't use jQuery and can be easily used in modals.

HannesClausnitzer:
fixes #2546


evap/contributor/forms.py
```diff
@@ -42,6 +42,11 @@ class Meta:
             "participants": UserModelMultipleChoiceField,
         }
 
+        widgets = {
+            "vote_start_datetime": forms.DateTimeInput(attrs={"type": "date", "class": "form-control"}),
```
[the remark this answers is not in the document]
hansegucker: I agree.

evap/staff/templates/staff_semester_import.html
```diff
@@ -45,16 +45,3 @@ <h3>{% translate 'Import semester data' %}</h3>
...
-{% block additional_javascript %}
-    {% include 'bootstrap_datetimepicker.html' %}
-    <script type="module">
-        import { copyHeaders } from "{% static 'js/copy-to-clipboard.js' %}";
-
-        document.getElementById("copy-headers-button").addEventListener("click", async () => await copyHeaders([
-            'Program', 'Participant last name', 'Participant first name', 'Participant email address', 'Course kind',
-            'Course is graded', 'Course name (de)', 'Course name (en)', 'Responsible title', 'Responsible last name',
-            'Responsible first name', 'Responsible email address'])
-        );
```
hansegucker, line 58: This is not a part of the datepicker. It probably wasn't your intention to remove this, was it?

hansegucker (changes requested):
Also, the file `evap/static/js/bootstrap-datetimepicker.min.js` still exists. 

But apart from that, this looks very good!

evap/contributor/forms.py
```diff
@@ -43,16 +42,17 @@ class Meta:
             "participants": UserModelMultipleChoiceField,
         }
 
+        widgets = {
+            "vote_start_datetime": forms.DateTimeInput(attrs={"type": "datetime-local", "class": "form-control"}),
```
niklasmohrin: I think that we should be able to do this automatically. Take a look at the comment above the `FORM_RENDERER` setting in `evap/settings.py`, we already automatically add the `form-control` to some widgets by placing files in `evaluation/templates/django/forms/widgets`.

I don't completely know how this works, but the Django documentation should be a good starting point for this. If it is too complicated though, we can also look at it later.

_Maybe_ we would have to define our own widget class that inherits from Django's widget.

cc @richardebeling (do you know how this works?)

niklasmohrin (commented):
Very cool!

We still mention bootstrap-datetimepicker in license.md and the places that are listed in the other comments

evap/contributor/forms.py
richardebeling: I was thinking the same thing, I also don't know how it works specifically, I would have to look it up as well. Talk to us if you want guidance/help, we're happy to help with this

janno42 (commented):
It would be nice to open the picker once the field receives focus.

Also, please investigate if we can force the date and time format to be our default format YYYY-MM-DD HH:mm.

richardebeling:
I'll re-mark as draft, if I recall correctly, we decided that a native HTML `datetime-local` input isn't viable because we had at least one scenario where Firefox would render the time with AM/PM although we didn't want it to (@janno42 right?)

evap/evaluation/templates/django/forms/widgets/date.html
```diff
@@ -0,0 +1,26 @@
...
+    <link rel="stylesheet" href="{% static 'tempus-dominus/tempus-dominus.min.css' %}" />
+    <!-- <script src="https://cdn.jsdelivr.net/npm/@popperjs/core@2.11.6/dist/umd/popper.min.js"></script> -->
+    <script src="{%static 'js/popper.min.js' %}"></script>
+    <script src="{% static 'tempus-dominus/tempus-dominus.min.js' %}"></script>
+    <!-- <script src="https://cdn.jsdelivr.net/npm/@eonasdan/tempus-dominus@6.9.4/dist/js/tempus-dominus.min.js" /> -->
+    <script src="{% static 'tempus-dominus/tempus-dominus-date-init.js' %}"></script>
+</head>
+
+<body>
+    <div class="input-group" id="datepicker1" data-td-target-input="nearest" data-td-target-toggle="nearest">
```
janno42: why do we need the number in the id?

evap/evaluation/templates/django/forms/widgets/datetime.html
```diff
@@ -0,0 +1,26 @@
+<head>
```
janno42: same here (head, ids)

evap/evaluation/templates/django/forms/widgets/date.html
```diff
@@ -0,0 +1,26 @@
+<head>
```
janno42: why do we need a head/body here? shouldn't all scripts and css be loaded by base.html?

evap/evaluation/templates/base.html
```diff
@@ -14,8 +14,12 @@
             <title>{% block title %}EvaP{% endblock %}</title>
 
             <link rel="stylesheet" href="{% static 'css/tom-select.bootstrap5.min.css' %}" />
-            <link rel="stylesheet" href="{% static 'css/bootstrap-datetimepicker.min.css' %}" />
             <link rel="stylesheet" href="{% static 'css/evap.css' %}" />
+            <link rel="stylesheet" href="{% static 'tempus-dominus/tempus-dominus.min.css' %}" />
+            <script src="{%static 'js/popper.min.js' %}"></script>
```
janno42: this should already be available via bootstrap, please check

evap/static/tempus-dominus/tempus-dominus-date-init.js
janno42, line 1: are the two tempus-dominus js files from us? then we should use typescript.

evap/static/tempus-dominus/tempus-dominus.min.css
```diff
@@ -0,0 +1,63 @@
...
...r-minute div:not(.no-highlight).disabled:hover,.tempus-dominus-widget.dark .time-container-second div:not(.no-highlight).disabled,.tempus-dominus-widget.dark .time-container-second div:not(.no-highlight).disabled:hover{color:var(--td-dark-disabled-color)}.tempus-dominus-widget.dark .date-container-days div:not(.no-highlight).today:before,.tempus-dominus-widget.dark .date-container-decades div:not(.no-highlight).today:before,.tempus-dominus-widget.dark .date-container-months div:not(.no-highlight).today:before,.tempus-dominus-widget.dark .date-container-years div:not(.no-highlight).today:before,.tempus-dominus-widget.dark .time-container-clock div:not(.no-highlight).today:before,.tempus-dominus-widget.dark .time-container-hour div:not(.no-highlight).today:before,.tempus-dominus-widget.dark .time-container-minute div:not(.no-highlight).today:before,.tempus-dominus-widget.dark .time-container-second div:not(.no-highlight).today:before{border-bottom-color:var(--td-dark-active-bg);border-top-color:var(--td-dark-secondary-border-color-rgba)}.tempus-dominus-widget.dark button{background-color:var(--td-dark-active-bg);border-color:var(--td-dark-active-bg);color:var(--td-dark-active-color)}
```
janno42, line 1: do we need all this? 😅

LICENSE.md
janno42, line 1: please add tempus dominus here with the correct license

janno42 (changes requested):
the picker doesn't always open when clicking on the input field.

evap/evaluation/templates/django/forms/widgets/date.html
HannesClausnitzer: it is included in the official documentation, but actually we don't need it, so i removed it.

evap/static/tempus-dominus/tempus-dominus.min.css
HannesClausnitzer: we need it for custom styling, but i moved it to a dedicated file so that it isn't included in the .min.css anymore.

evap/evaluation/templates/base.html
HannesClausnitzer: what do you mean exactly?

HannesClausnitzer: btw it seems popper js script is needed, when i remove it, the datetimepicker breaks...

evap/static/scss/components/_datetime-picker.scss
```diff
@@ -0,0 +1,68 @@
...
+    margin: 0 !important;
+    display: flex !important;
+    flex-direction: column;
+    justify-content: center !important;
+    align-items: center !important;
+}
+
+/* Hour/Minute display - WIDER to fill horizontal space */
+.tempus-dominus-widget.timepicker-sbs .time-container .clock .hour,
+.tempus-dominus-widget.timepicker-sbs .time-container .clock .minute {
```
janno42, line 37: I can't find the `.clock` and `.hour`/`.minute` classes in the HTML. Is this part correct?
There is a `.time-container-clock` in `.time-container` though.

janno42 (commented):
UI works

evap/evaluation/templates/django/forms/widgets/date.html
```diff
@@ -0,0 +1,14 @@
+<div class="input-group" id="datepicker1" data-td-target-input="nearest" data-td-target-toggle="nearest">
```
richardebeling, line 1: `id`s need to be unique in the full HTML document. This widget HTML snippet here may be used on one page multiple times, so we cannot put any `id`s in here.

Do we use this for anything? If not, let's just remove it. Otherwise, we have to use some other mechanism to identify these elements (probably classes)?

evap/staff/forms.py
```diff
@@ -100,8 +99,19 @@ def clean(self):
 class ImportForm(forms.Form):
     use_required_attribute = False
 
-    vote_start_datetime = forms.DateTimeField(label=_("Start of evaluation"), localize=True, required=False)
-    vote_end_date = forms.DateField(label=_("End of evaluation"), localize=True, required=False)
+    vote_start_datetime = forms.DateTimeField(
+        label=_("Start of evaluation"),
+        widget=forms.DateTimeInput(),
```
richardebeling, line 104: Is this necessary? I would have expected tht `forms.DateTimeInput` is the default widget for a `forms.DateTimeField`.

evap/staff/forms.py
```diff
@@ -100,8 +99,19 @@ def clean(self):
...
+        localize=True,
+        required=False,
+    )
+
+    vote_end_date = forms.DateField(
+        label=_("End of evaluation"),
+        widget=forms.DateInput(attrs={"type": "date", "class": "form-control"}),
```
richardebeling, line 111: What do we need `"type": "date`" for here? Should it just be part of the default for the widget? (And I guess the same for `"class": "form-control"`)

evap/static/ts/src/tempus-dominus-datetime-init.ts
richardebeling, line 1: The contents of our two `-init` files looks like it's mostly the same. Could we just put handling for both components in one file (and reuse all code that is currently duplicated)?

I think chances are if we change one of these files, we want to change the other one as well

evap/static/ts/src/tempus-dominus.ts
```diff
@@ -0,0 +1,9 @@
+declare global {
+    interface Window {
+        tempusDominus: any;
+    }
+}
```
richardebeling, line 5: The only purpose of this is to allow us to do `window.tempusDominus`? Is there an advantage of doing this over just doing `(window as any).tempusDominus`?

@niklasmohrin opinions on how to do this?

niklasmohrin: I would also just declare `any` wherever we use this

niklasmohrin:
@HannesClausnitzer do you want to keep working on this, or should we take over / assign this issue to someone else?

HannesClausnitzer:
at the moment I don't have time to work on this, so I'm totally fine with you taking over...

niklasmohrin:
Okay, I will close this PR for now because I think we need to get back to the drawing board on this one. This issue seems more complicated than we thought. We may come back to this PR if we decide to continue down the tempus dominus road. Thanks for all the work so far!

PR #2609, opened 2026-01-12
Misleading log entry when clearing field

### LINKED ISSUE #2603: Misleading log entry when clearing field
When clearing a field on a `LoggedModel` (tested on exam type in an evaluation), the log shows a message like `Exam type: Exam → <deleted object>`. This is incorrect because the new type was not actually deleted; it is simply an empty value. Therefore, the message should read `Exam type: Exam → <empty>`

Belissimo-T:
fixes #2603

### DIFF
--- a/evap/evaluation/tests/test_models_logging.py
+++ b/evap/evaluation/tests/test_models_logging.py
@@ -244,3 +244,24 @@ class TestLoggedModel(TestCase):
             evaluation.related_logentries().latest("id").field_context_data,
             {"exam_type": [FieldAction(label="Exam type", type="change", items=["<none>", exam_type.name_en])]},
         )
+
+    def test_deleted_object(self):
+        """Logging of deleted objects. (Regression test for #2603)"""
+        exam_type = baker.make(ExamType)
+        evaluation = baker.make(Evaluation, exam_type=exam_type)
+        exam_type2 = baker.make(ExamType)
+
+        evaluation = Evaluation.objects.get(pk=evaluation.pk)
+
+        evaluation.exam_type = exam_type2
+        evaluation.save()
+        exam_type.delete()
+
+        self.assertEqual(
+            evaluation.related_logentries().latest("id").field_context_data,
+            {
+                "exam_type": [
+                    FieldAction(label="Exam type", type="change", items=["<deleted object>", exam_type2.name_en])
+                ]
+            },
+        )


evap/evaluation/models_logging.py
niklasmohrin, line 1: We should add a test for the two cases described in the issue

evap/evaluation/models_logging.py
```diff
@@ -78,9 +78,15 @@ def _field_actions_for_field(field, actions):
     for field_action_type, items in actions.items():
         if field.many_to_many or field.many_to_one or field.one_to_one:
             # convert item values from primary keys to string-representation for relation-based fields
-            related_objects = field.related_model.objects.filter(pk__in=items)
-            missing = len(items) - related_objects.count()
-            items = [str(obj) for obj in related_objects] + [_("<deleted object>")] * missing
+
+            items_not_none = [i for i in items if i is not None]
```
niklasmohrin: cant we just use `items` below? `None` being in there should not make a difference

evap/evaluation/models_logging.py
```diff
@@ -78,9 +78,15 @@ def _field_actions_for_field(field, actions):
...
+
+            pk_to_obj = {obj.pk: obj for obj in field.related_model.objects.filter(pk__in=items_not_none)}
+
+            items = [
+                _("<unset>") if item is None else str(pk_to_obj[item]) if item in pk_to_obj else _("<deleted object>")
```
niklasmohrin: I don't get it, the issue says `<empty>`, but this does not appear here?

Belissimo-T: I talked with @janno42 and we decided on `<none>`.

evap/evaluation/models_logging.py
Belissimo-T: Thanks, you're right.

evap/evaluation/tests/test_models_logging.py
```diff
@@ -205,3 +205,33 @@ def test_m2m_logging_respects_unlogged_fields(self):
             participant.evaluations_participating_in.clear()
             participant.evaluations_participating_in.add(self.evaluation)
             self.assertFalse(any("participants" in entry.data for entry in self.evaluation.related_logentries()))
+
+    def test_clear_field(self):
+        """Test that clearing a field shows '<empty>' instead of '<deleted object>'"""
```
jooooosef: We are setting it to `<none>` not empty.

I also think that this just describes exactly what happens in the test, so we could just omit that und write something like "Regression test for #2603" (same below)

evap/evaluation/tests/test_models_logging.py
```diff
@@ -205,3 +205,33 @@ def test_m2m_logging_respects_unlogged_fields(self):
...
+        evaluation.exam_type = None
+        evaluation.save()
+
+        self.assertEqual(
+            evaluation.related_logentries().latest("id").field_context_data,
+            {"exam_type": [FieldAction(label="Exam type", type="change", items=[exam_type.name_en, "<none>"])]},
+        )
+
+    def test_set_field(self):
+        """Test that setting a field from empty shows correct order with '<empty>' as old value"""
```
jooooosef: We are setting it to `<none>` not empty.

evap/evaluation/tests/test_models_logging.py
```diff
@@ -205,3 +205,33 @@ def test_m2m_logging_respects_unlogged_fields(self):
...
+
+        evaluation.exam_type = None
+        evaluation.save()
+
+        self.assertEqual(
+            evaluation.related_logentries().latest("id").field_context_data,
+            {"exam_type": [FieldAction(label="Exam type", type="change", items=[exam_type.name_en, "<none>"])]},
+        )
+
+    def test_set_field(self):
```
jooooosef: I think that this name is a bit too generic and we could highlight the fact that we set an empty field to something not empty. Something like test_set_field_from_empty.

evap/evaluation/models_logging.py
```diff
@@ -78,9 +78,13 @@ def _field_actions_for_field(field, actions):
...
-            related_objects = field.related_model.objects.filter(pk__in=items)
-            missing = len(items) - related_objects.count()
-            items = [str(obj) for obj in related_objects] + [_("<deleted object>")] * missing
+
+            pk_to_obj = {obj.pk: obj for obj in field.related_model.objects.filter(pk__in=items)}
+
+            items = [
+                _("<none>") if item is None else str(pk_to_obj[item]) if item in pk_to_obj else _("<deleted object>")
+                for item in items
+            ]
```
jooooosef: I think this looks more complicated than it needs to be (with the ternary-logic)?
We could use the [dictionary get](https://www.w3schools.com/python/ref_dictionary_get.asp) to set a default value, so `str(pk_to_obj.get(item, _("<deleted object>")))`

Belissimo-T: I like the ternary expression more, though. When using `str(pk_to_obj.get(item, _("<deleted object>")))`, we evaluate `_("<deleted object>")` even when the item is in the dict and if the item is not in the dict, we redundantly convert the translated string to a `str`.

evap/evaluation/tests/test_models_logging.py
```diff
@@ -205,3 +205,33 @@ def test_m2m_logging_respects_unlogged_fields(self):
             participant.evaluations_participating_in.clear()
             participant.evaluations_participating_in.add(self.evaluation)
             self.assertFalse(any("participants" in entry.data for entry in self.evaluation.related_logentries()))
+
+    def test_clear_field(self):
+        """Regression test for #2603"""
```
richardebeling:
```suggestion
        """Regression test for #2603: When clearing a m2m field, the log should say "none" as the new value"""
```

(When running the tests in verbose mode, _only_ the docstring is shown, we don't show the test names anymore. Having a list of hundreds of test, each just saying "Regression test for #xxxx" doesn't really help in that case)

evap/evaluation/tests/test_models_logging.py
```diff
@@ -205,3 +205,33 @@ def test_m2m_logging_respects_unlogged_fields(self):
...
+        evaluation.exam_type = None
+        evaluation.save()
+
+        self.assertEqual(
+            evaluation.related_logentries().latest("id").field_context_data,
+            {"exam_type": [FieldAction(label="Exam type", type="change", items=[exam_type.name_en, "<none>"])]},
+        )
+
+    def test_set_field_from_none(self):
+        """Regression test for #2603"""
```
richardebeling:
```suggestion
        """Regression test for #2603: Logging when adding values to an empty m2m field"""
```

evap/evaluation/tests/test_models_logging.py
niklasmohrin, line 1: We should also have a test that triggers a log with "deleted object" in it. Currently, I can change the `.get` to just `pk_to_obj[item]` and all tests still pass. Thus I can't really be confident that our code does what we think it does.

evap/evaluation/tests/test_models_logging.py
jooooosef: :o

richardebeling (approved):

PR #2610, opened 2026-01-12
Hide Dropout Questionnaire with no results

### LINKED ISSUE #2601: Dropout questionnaire should only be shown on results page if answers were given
Dropout questionnaires are currently shown on the results page, even when no participants have answered the questions:
Image

The questionnaire and the surrounding card should only be included on the page if answers were given.

geromequa:
Closes #2601

Add simple check in `results_evaluation_detail.html` and expand `test_dropout_results_shown` test.

### DIFF
--- a/evap/results/templates/results_evaluation_detail.html
+++ b/evap/results/templates/results_evaluation_detail.html
@@ -230,13 +230,9 @@
                 </div>
             </div>
             <div class="card-body">
-                {% if dropout_questionnaire_results|length > 0 %}
-                    {% for questionnaire_result in dropout_questionnaire_results %}
-                        {% include 'results_evaluation_detail_questionnaires.html' %}
-                    {% endfor %}
-                {% else %}
-                    <i>{% translate 'No additional answers were given.' %}</i>
-                {% endif %}
+                {% for questionnaire_result in dropout_questionnaire_results %}
+                    {% include 'results_evaluation_detail_questionnaires.html' %}
+                {% endfor %}
             </div>
         </div>
     {% endif %}
--- a/evap/results/tests/test_views.py
+++ b/evap/results/tests/test_views.py
@@ -538,7 +538,7 @@ class TestResultsSemesterEvaluationDetailView(WebTestStaffMode):
         self.assertContains(response, "test-dropout-question-text")
         self.assertContains(response, "test-dropout-questionnaire-title")
 
-        # Assert Dropout Section is not shown if no dropout_results exist
+        # Assert Dropout Section is not shown if nobody dropped out
         self.evaluation.dropout_count = 0
         self.evaluation.save()
 


evap/results/templates/results_evaluation_detail.html
```diff
@@ -220,7 +220,7 @@ <h3>{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</h3>
         </div>
     {% endif %}
 
-    {% if evaluation.is_dropout_allowed or evaluation.dropout_count > 0 or dropout_questionnaire_results %}
+    {% if evaluation.is_dropout_allowed and evaluation.dropout_count > 0 and dropout_questionnaire_results|length > 0 %}
```
niklasmohrin: I don't really see why `is_dropout_allowed` is in there, if that would be false, then there cannot be any dropout questionnaire results either. And then, I feel that we should display _something_ if at least one person dropped out _or_ submitted something.

@janno42 Is there ever a case where we someone would drop out, and then staff would change the evaluation to not allow dropout anymore? No right?

Then I think what makes the most sense to me here is
```suggestion
    {% if evaluation.dropout_count > 0 or dropout_questionnaire_results|length > 0 %}
```

Regardless of what exactly we do, we should make sure that our tests cover the entire decision we do here; changing `and` to `or` or vice versa should make the tests fail.

janno42: We delete answers for questionnaires that are removed during a running evaluation. So if staff would remove the dropout questionnaire, `is_dropout_allowed` would be `False` and no answers for the dropout questionnaires would exist anymore.

jooooosef: Just on a sidenote, as we discussed this just now:
`and` could be explicitly false here since when someone always clicks "No answer" for the questions and submits, then the `dropout_count` would (rightfully so) be > 0, but there would not be a `dropout_questionnaire_result`.

evap/results/templates/results_evaluation_detail.html
```diff
@@ -229,11 +229,17 @@ <h3>{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</h3>
...
             <div class="card-body">
                 {% for questionnaire_result in dropout_questionnaire_results %}
                     {% include 'results_evaluation_detail_questionnaires.html' %}
                 {% endfor %}
             </div>
+            {% else %}
+            <i>
+                {% translate 'No additional answers were given.' %}
+            </i>
+            {% endif %}
```
jooooosef: The second part "No additional answers were given" needs to be in the card body div, otherwise it would look like this:

Image

evap/results/templates/results_evaluation_detail.html
```diff
@@ -229,11 +229,17 @@ <h3>{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</h3>
                     <span class="fas fa-user"></span> {{ evaluation.dropout_count }}
                 </div>
             </div>
+            {% if dropout_questionnaire_results|length > 0 %}
```
jooooosef: I do like this change to add this here :)

evap/results/templates/results_evaluation_detail.html
```diff
@@ -220,7 +220,7 @@ <h3>{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</h3>
         </div>
     {% endif %}
 
-    {% if evaluation.is_dropout_allowed or evaluation.dropout_count > 0 or dropout_questionnaire_results %}
+    {% if evaluation.dropout_count > 0 %}
```
jooooosef, line 223: We now assume that this always gets updated correctly, but this prob fine @niklasmohrin ?

jooooosef (changes requested):

evap/results/templates/results_evaluation_detail.html
niklasmohrin: Yes, we can rely on dropout count always being increased if someone drops out

evap/results/templates/results_evaluation_detail.html
```diff
@@ -234,12 +234,12 @@ <h3>{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</h3>
...
-            </div>
-            {% else %}
-            <i>
+                {% else %}
+                <i>
                 {% translate 'No additional answers were given.' %}
-            </i>
+                </i>
             {% endif %}
+            </div>
```
jooooosef: the `<div class="card-body">` is tied to the if/else statement. so when `dropout_questionnaire_results|length > 0` then `div` is opened and closed later. But if not (else-case) the div is not opened. You need to move the `<div class="card-body">` outside the if/else statement.

evap/results/tests/test_views.py
```diff
@@ -538,6 +538,14 @@ def test_dropout_results_shown(self):
         self.assertContains(response, "test-dropout-question-text")
         self.assertContains(response, "test-dropout-questionnaire-title")
 
+        # Assert Dropout Section is not shown if no dropout_results exist
```
niklasmohrin:
```suggestion
        # Assert Dropout Section is not shown if nobody dropped out
```

evap/results/templates/results_evaluation_detail.html
```diff
@@ -230,9 +230,13 @@ <h3>{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</h3>
                 </div>
             </div>
             <div class="card-body">
-                {% for questionnaire_result in dropout_questionnaire_results %}
-                    {% include 'results_evaluation_detail_questionnaires.html' %}
-                {% endfor %}
+                {% if dropout_questionnaire_results|length > 0 %}
+                    {% for questionnaire_result in dropout_questionnaire_results %}
```
niklasmohrin: you can simplify this using the `{% empty %}` directive of Django's loops

evap/results/templates/results_evaluation_detail.html
```diff
@@ -230,9 +230,13 @@ <h3>{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</h3>
...
+                        {% include 'results_evaluation_detail_questionnaires.html' %}
+                    {% endfor %}
+                {% else %}
+                    <i>{% translate 'No additional answers were given.' %}</i>
```
richardebeling: This would be shown if at least one person marked the evaluation as "I dropped out", but the number of people answering the dropout questionnaire questions with anything else than "No answer" is less than the publishing threshold, right?

It would probably make sense to also test this case in the unit test, the setup should also just be 2-5 lines.

niklasmohrin (approved):
Makes sense, thanks!

evap/results/templates/results_evaluation_detail.html
niklasmohrin: I don't even think this would happen then, if I delete all ratinganswercounts the results still exists, just with all counters zero; I believe this case here would only happen if no questionnaires would be selected, but then dropout would not be allowed

richardebeling: Then we shouldn't have the code here :D

niklasmohrin: When will Python finally get dependent type theory and the non-empty-list type :smiling_face_with_tear:

richardebeling (approved):

PR #2612, opened 2026-01-19
exclude evaluation responsibles from editor review notice email cc

### LINKED ISSUE #2589: Send Editor Review Notice only once
When staff users perform the `ReadyForEditorsOperation` (making evaluations ready for editors to approve), emails (Editor Review Notice) are sent to the evaluations' responsible persons (each one individually) with all contributors who have edit rights copied in CC. Users who already receive these emails in their role as responsible persons should not receive emails addressed to other responsible persons in CC.

Belissimo-T:
fixes #2589

### DIFF
--- a/evap/staff/views.py
+++ b/evap/staff/views.py
@@ -373,10 +373,11 @@ class ReadyForEditorsOperation(EvaluationOperation):
             ).format(len(evaluations)),
         )
         if email_template:
-            evaluations_by_responsible: dict[UserProfile, list[Evaluation]] = defaultdict(list)
-            for evaluation in evaluations:
-                for responsible in evaluation.course.responsibles.all():
-                    evaluations_by_responsible[responsible].append(evaluation)
+            evaluations_by_responsible = unordered_groupby(
+                (responsible, evaluation)
+                for evaluation in evaluations
+                for responsible in evaluation.course.responsibles.all()
+            )
 
             for responsible, responsible_evaluations in evaluations_by_responsible.items():
                 body_params = {"user": responsible, "evaluations": responsible_evaluations}


evap/staff/views.py
```diff
@@ -383,7 +383,7 @@ def apply(
                 editors = UserProfile.objects.filter(
                     contributions__evaluation__in=responsible_evaluations,
                     contributions__role=Contribution.Role.EDITOR,
-                ).exclude(pk=responsible.pk)
+                ).exclude(pk__in=[responsible.pk for responsible in evaluations_by_responsible])
```
jooooosef: This would lead to the following scenario:
`responsible_1` is responsible for eval1.
`responsible_2` is responsible for eval2 and `responsible_1` is contributor here. 
Then `responsible_1` would receive a mail for eval1 but would **not** be in cc for eval2. 
This is not wanted behaviour. 

In general would any reponsible not be in any cc for any mail, because you exlcude every responsible from any evaluation from the cc.

Belissimo-T: Ahh you're right. It's wrong. Thanks. I hope it's fixed now.

evap/staff/tests/test_views.py
```diff
@@ -1807,6 +1807,20 @@ def test_operation_prepare_does_not_put_responsible_into_cc(self):
...
+
+        evaluation = baker.make(Evaluation, state=Evaluation.State.NEW, course=self.course)
+        baker.make(Contribution, evaluation=evaluation, contributor=other_responsible, role=Contribution.Role.EDITOR)
+
+        url_options = f"?evaluation={evaluation.pk}&target_state={Evaluation.State.PREPARED}"
+        actual_emails = self.submit_operation_prepare_form(url_options)
+
+        self.assertEqual(len(actual_emails), 2)
+        email_to_responsible = next(email for email in actual_emails if email["user"] == self.responsible)
+        self.assertEqual(email_to_responsible["additional_cc_users"], set())
```
richardebeling: Neither email should have any CC receivers, right? In that case, I'd just
```python3
self.assertEqual([email["additional_cc_users"] for email in actual_emails], [set(), set()])
```

evap/staff/tests/test_views.py
```diff
@@ -1807,6 +1807,20 @@ def test_operation_prepare_does_not_put_responsible_into_cc(self):
         self.assertEqual(len(actual_emails), 1)
         self.assertEqual(actual_emails[0]["additional_cc_users"], set())
 
+    def test_operation_prepare_does_not_put_other_responsible_into_cc_if_they_are_editor(self):
+        other_responsible = baker.make(UserProfile, email="co-responsible@example.com")
```
richardebeling: In all existing tests, we typically name additional instances `[...]2` or `[...]_b`. For easier readability, I wouldn't introduce new `other_` and `co-` prefixes and just reuse the existing naming patterns.

evap/staff/views.py
```diff
@@ -373,17 +373,18 @@ def apply(
             ).format(len(evaluations)),
         )
         if email_template:
-            evaluations_by_responsible = {}
+            evaluations_by_responsible: dict[UserProfile, list[Evaluation]] = defaultdict(list)
             for evaluation in evaluations:
                 for responsible in evaluation.course.responsibles.all():
-                    evaluations_by_responsible.setdefault(responsible, []).append(evaluation)
+                    evaluations_by_responsible[responsible].append(evaluation)
```
richardebeling: fyi: strangely, we have to build such "inverted groupings" relatively often, so we built a helper for this operation at some point in the past: [`unordered_groupby`](https://github.com/e-valuation/EvaP/blob/ead55bbc6297706ed8e8f1b905b398fd965bb8fe/evap/tools.py#L18). This could be written as:
```python3
evaluations_by_responsible = unordered_groupby(
    (responsible, evaluation)
    for evaluation in evaluations
    for responsible in evaluation.course.responsibles.all()
)
```

Two small advantages: Typing should just work without explicit hints, and we don't "leak" the defaultdict into the remaining context.

niklasmohrin (approved):
I agree with the other comments and have nothing else to add

evap/staff/views.py
Belissimo-T: Ahh nice. Thanks.

PR #2615, opened 2026-01-25
Use faster nix installer action and cache for CI

niklasmohrin:
I saw that there is this action to cache the nix store and in their README they also mentioned this other nix installer action. Both look very promising, but let's see how they perform in our CI for this PR.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,5 +1,4 @@
 {
-
   description = "EvaP";
 
   inputs = {


niklasmohrin:
I removed the prefix-restoring stuff now because I feel that we can only mess this up. If a normal job restores from, for example, the frontend test shell, then all jobs will pick up this bigger cache going forward. I think it's fine if we rebuild everything if an input changes. Then, we don't need to do gc either

niklasmohrin:
I switched to the cachix installer action now because they just use the nixos.org installer and it is only seconds slower. One problem I am foreseeing is that by using different caches we quickly use up our free 10 Gb of caches. From this PR, we have 5 caches with roughly 1GB and 1 small one (see https://github.com/e-valuation/EvaP/actions/caches). So if we have a renovate PR with automatic rebases lurking around, our caches will basically always be purged and the builds are slow again.

The cache for the macos workflow is a bit silly because if we hit the cache then we should have just skipped the job entirely. For the others, we could, for example, always include the files for `.#services` in the store and restore the slightly bigger cache even if we don't need the database running (for example when checking formatting). Or, we could not save the caches for some of the jobs and always rebuild a little bit on top of the cache, for example for build_wheel we need the normal dev shell plus a second shell that has the `build` Python module installed. For backup_process, looking at this run from main: https://github.com/e-valuation/EvaP/actions/runs/21843459302/job/63033840236 I guess we don't need any caching for the job because downloading and starting the database takes only 20s anyways.

Thoughts? @richardebeling

.github/setup_evap/action.yml
```diff
@@ -18,7 +18,12 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - uses: DeterminateSystems/nix-installer-action@main
+    - name: Install nix
+      uses: cachix/install-nix-action@v31
+    - name: Restore and save Nix store
+      uses: nix-community/cache-nix-action@v7
+      with:
+        primary-key: nix-${{ runner.os }}-${{ hashFiles('**/*.nix', 'flake.lock', 'uv.lock', 'evap/development/manage_autocompletion.sh') }}-${{ inputs.shell }}-${{ inputs.start-db }}
```
richardebeling, line 30: Regarding caching limits:

For my understanding: We have 5*1GB because we intentionally have different hash keys for the different runs, but most of the time, we would expect them to be ~90% the same contents across runs, right?

Radical approach: Could we just write to caches for runs on `main` and have the caches read-only for pull requests to circumvent cache trashing? If your PR bumps a dependency, you'll have to wait a bit, I think that's fair (we're talking about 1-2 minutes here, right?).

> The cache for the macos workflow is a bit silly because if we hit the cache then we should have just skipped the job entirely.

How hard would it be to build that logic, i.e. "cache just a success bool for a given input hash, and succeed early if it exists (otherwise, build from scratch)"? I feel like it should be doable in a handful of job configuration lines (but if not, let's not overengineer it)

> For the others, we could, for example, always include the files for .#services in the store and restore the slightly bigger cache even if we don't need the database running (for example when checking formatting). Or, we could not save the caches for some of the jobs and always rebuild a little bit on top of the cache, for example for build_wheel we need the normal dev shell plus a second shell that has the build Python module installed.

Sounds very reasonable to me, I'd be in favor: PRs need to wait for the slowest run anyway, which is likely pylint and/or tests, so if formatting takes 5 seconds extra, that doesn't cause much harm.

> For backup_process, looking at this run from main: https://github.com/e-valuation/EvaP/actions/runs/21843459302/job/63033840236 I guess we don't need any caching for the job because downloading and starting the database takes only 20s anyways.

Agree

PR #2621, opened 2026-01-26
Make global evaluation progress feature generic

### LINKED ISSUE #2617: Customizable Progress Bar Texts
On the student index page, a progress bar (GLOBAL_EVALUATION_PROGRESS) with customizable steps can be displayed. The long description text in the dropdown can also be customized, but the heading ("Fundraising") and the dropdown header ("Read more about our fundraising goal.") cannot be customized yet.
All texts for the progress bar should be customizable so that it can be used for other purposes.

hansegucker:
Close #2617

### DIFF
--- a/evap/development/localsettings.template.py
+++ b/evap/development/localsettings.template.py
@@ -42,18 +42,23 @@ SECRET_KEY = "$SECRET_KEY"  # nosec
 ALLOWED_HOSTS = ["localhost", "127.0.0.1"]
 
 ### Evaluation progress rewards
-GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[Fraction, str]] = [
-    (Fraction("0"), "0€"),
-    (Fraction("0.25"), "1.000€"),
-    (Fraction("0.6"), "3.000€"),
-    (Fraction("0.7"), "7.000€"),
-    (Fraction("0.9"), "10.000€"),
+GLOBAL_EVALUATION_PROGRESS_REWARDS: list[tuple[Fraction, dict[str, str]]] = [
+    (Fraction("0"), {"de": "0 €", "en": "0€"}),
+    (Fraction("0.25"), {"de": "1.000 €", "en": "1,000€"}),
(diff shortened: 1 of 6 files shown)


evap/development/localsettings.template.py
```diff
@@ -51,6 +51,11 @@
 ]
 GLOBAL_EVALUATION_PROGRESS_EXCLUDED_COURSE_TYPE_IDS: list[int] = []
 GLOBAL_EVALUATION_PROGRESS_EXCLUDED_EVALUATION_IDS: list[int] = []
+GLOBAL_EVALUATION_PROGRESS_TITLE = {"de": mark_safe("Spendenaktion"), "en": mark_safe("Fundraising")}
```
niklasmohrin: we don't need mark_safe for these, the only reason we used it below is because there is html in the strings

evap/student/views.py
```diff
@@ -83,21 +85,25 @@ def from_settings() -> "GlobalRewards | None":
...
 
-        return GlobalRewards(
+        return GlobalEvaluationProgress(
             vote_count=vote_count,
             participation_count=participation_count,
             max_reward_votes=max_reward_votes,
             bar_width_votes=min(vote_count, max_reward_votes),
             last_vote_datetime=last_vote_datetime,
             rewards_with_progress=rewards_with_progress,
+            title=settings.GLOBAL_EVALUATION_PROGRESS_TITLE[get_language()],
```
niklasmohrin: can we call `get_language` just once above?

evap/student/views.py
```diff
@@ -54,10 +54,12 @@ class RewardProgress:
     bar_width_votes: int
     last_vote_datetime: datetime.datetime
     rewards_with_progress: list[RewardProgress]
+    title: str
+    info_title: str
```
niklasmohrin: hm, the names are not super great; with the lint warning, maybe we should just make a small dataclass "Campaign" or so inside GlobalEvaluationProgress that has fields title, callout, and description? Then we can also just have one setting of type GlobalEvaluationProgress.Campaign. What do you think?

PR #2632, opened 2026-02-02
Json importer: hyphen normalization, main language, default textanswer visibility

### LINKED ISSUE #2623: Further JSON Importer Improvements
The following changes should be made to the CMS data JSON importer:
- Dashes in Course and Evaluation names should be cleared: All ` - ` (hyphen) surrounded by whitespace should be replaced by ` – ` (en dash), surrounded by one space on each side.
- The main language should be set in exam evaluations, similar to main evaluations. Currently, it is not defined by the importer.
- Textanswer visibility for contributors in all created evaluations should be set to "Own and General."

geromequa:
Closes #2623.

### DIFF
--- a/evap/cms/tests/test_json_importer.py
+++ b/evap/cms/tests/test_json_importer.py
@@ -686,6 +686,7 @@ class TestImportEvents(TestCase):
         evaluation.general_contribution.questionnaires.add(
             baker.make(Questionnaire, type=Questionnaire.Type.CONTRIBUTOR)
         )
+        evaluation.main_language = "de"
         evaluation.manager_approve()
         evaluation.vote_start_datetime = datetime.now()
         evaluation.vote_end_date = datetime.now().date() + timedelta(days=1)


evap/cms/json_importer.py
```diff
@@ -188,9 +188,9 @@ def send_mail(self):
         mail.send()
 
 
-def _clean_whitespaces(text: str) -> str:
+def _clean_whitespaces_and_hyphens(text: str) -> str:
```
niklasmohrin, line 192:
```suggestion
def _clean_whitespaces_and_hyphens(text: str) -> str:
    """Normalize to single space characters and en-dashes."""
```

evap/staff/importers/json.py
```diff
@@ -596,14 +601,18 @@ def _import_contribution(self, evaluation: Evaluation, data: ImportRelated) -> t
...
+            role=Contribution.Role.EDITOR,
+            textanswer_visibility=Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS,
         )
         return contribution, created
 
     def _import_events(self, data: list[ImportEvent]) -> None:  # noqa:PLR0912
         # Divide in multiple lists to handle individually
         non_exam_events, exam_events, exam_events_without_related_non_exam_event = [], [], []
         for event in data:
+            self.events_dict[event["gguid"]] = event
```
niklasmohrin: how about `events_by_gguid`?

evap/staff/importers/json.py
```diff
@@ -523,6 +524,10 @@ def _import_evaluation(  # noqa: PLR0912, PLR0915
             is_rewarded = True
 
         main_language = LANGUAGE_MAP.get(data["language"], Evaluation.UNDECIDED_MAIN_LANGUAGE)
+        if main_language == Evaluation.UNDECIDED_MAIN_LANGUAGE and data.get("relatedevents"):
+            related_main_evaluation = self.events_dict[data["relatedevents"][0]["gguid"]]
```
niklasmohrin: not super stoked about having the `relatedevents[0]` dance in a second place now; in the other place we have the assertion at least. can we maybe extract this lookup into a small helper that also has the assert?

evap/cms/tests/test_json_importer.py
```diff
@@ -891,9 +897,15 @@ def test_exam_type_different_name(self):
         self.assertEqual(exam_evaluation.name_en, exam_evaluation.exam_type.name_de)
 
     def test_clean_whitespaces(self):
```
niklasmohrin, line 813: please also add `test_clean_hyphens`

evap/cms/tests/test_json_importer.py
niklasmohrin, line 1: the textanswer_visibility change is not tested

evap/staff/importers/json.py
geromequa: See `def get_related_events` in `json.py` on new commit.

evap/cms/tests/test_json_importer.py
geromequa: see `def test_textanswer_visibility` in `test_json_importer.py`

niklasmohrin: see yourself :^)

geromequa: Look again

evap/cms/tests/test_json_importer.py
```diff
@@ -809,9 +810,31 @@ def test_exam_type_different_name(self):
...
+        self.assertEqual(_clean_whitespaces_and_hyphens("Course - Evaluation"), "Course – Evaluation")
+        self.assertEqual(_clean_whitespaces_and_hyphens("-"), "-")
+
+    def test_exam_missing_language(self):
+        self._import(EXAMPLE_DATA)
+
+        evaluation = Evaluation.objects.get(cms_id=EXAMPLE_DATA["events"][1]["gguid"])
+        self.assertEqual(evaluation.main_language, "de")
+
+    def test_textanswer_visibility(self):
```
richardebeling: What additional benefit does this test have, compared to `def test_import_courses` (which you expended with what looks to me like the same assertion)

niklasmohrin (commented):
in person discussion:
- helper changes
- comments for new events_by_gguid and course dict

evap/cms/tests/test_json_importer.py
```diff
@@ -809,9 +810,19 @@ def test_exam_type_different_name(self):
...
         # non-breaking whitespace
-        self.assertEqual(_clean_whitespaces("inbetween  inbetween"), "inbetween inbetween")
+        self.assertEqual(_clean_whitespaces_and_hyphens("inbetween  inbetween"), "inbetween inbetween")
+
+    def test_clean_hyphens(self):
+        self.assertEqual(_clean_whitespaces_and_hyphens("Course - Evaluation"), "Course – Evaluation")
+        self.assertEqual(_clean_whitespaces_and_hyphens("-"), "-")
+
+    def test_exam_missing_language(self):
+        self._import(EXAMPLE_DATA)
```
niklasmohrin, line 826: for our sanity: can you add the following?
```suggestion
    def test_exam_missing_language(self):
        self._import(EXAMPLE_DATA)
        self.assertEqual(EXAMPLE_DATA["events"][1]["language"], "", "precondition for the test")
```

evap/cms/json_importer.py
```diff
@@ -242,9 +242,18 @@ def __init__(self, semester: Semester, default_course_end: date) -> None:
         self.course_type_cache = ImportCache(CourseType)
         self.exam_type_cache = ImportCache(ExamType)
         self.program_cache = ImportCache(Program)
-        self.courses_by_gguid: dict[str, Course] = {}
         self.statistics = ImportStatistics()
 
+        # raw events
+        self.events_by_gguid: dict[str, ImportEvent] = {}
+        # courses already parsed as events
```
niklasmohrin: Doesn't the following make more sense?
```suggestion
        # events already parsed as courses
```

evap/cms/json_importer.py
```diff
@@ -627,10 +644,9 @@ def _import_events(self, data: list[ImportEvent]) -> None:  # noqa:PLR0912
         for event in exam_events:
             # Exam events have the non-exam event as a single entry in the relatedevents list
             # We lookup the Course from this non-exam event (the main evaluation) to add the exam evaluation to the same Course
```
niklasmohrin: I think the first comment does not fit here but in the helper instead; we should use this comment in the helper instead of the shorter one we currently have

The second comment needs to be slightly updated then because "this" does not refer to anything anymore
```suggestion
            # We lookup the Course from the main evaluation to add the exam evaluation to the same Course
```

evap/cms/json_importer.py
```diff
@@ -242,9 +242,18 @@ def __init__(self, semester: Semester, default_course_end: date) -> None:
...
         self.program_cache = ImportCache(Program)
-        self.courses_by_gguid: dict[str, Course] = {}
         self.statistics = ImportStatistics()
 
+        # raw events
+        self.events_by_gguid: dict[str, ImportEvent] = {}
+        # courses already parsed as events
+        self.courses_by_gguid: dict[str, Course] = {}
+
+    def get_main_evaluation(self, exam_event: ImportEvent) -> ImportEvent:
```
niklasmohrin: now that I see it, I feel like it should be 
```suggestion
    def get_main_evaluation_data(self, exam_event: ImportEvent) -> ImportEvent:
```
as I would otherwise expect that it returns `Evaluation`; what do you think?

evap/cms/json_importer.py
```diff
@@ -532,6 +541,10 @@ def _import_evaluation(  # noqa: PLR0912, PLR0915
             is_rewarded = True
 
         main_language = LANGUAGE_MAP.get(data["language"], Evaluation.UNDECIDED_MAIN_LANGUAGE)
+        if main_language == Evaluation.UNDECIDED_MAIN_LANGUAGE and data.get("relatedevents"):
```
niklasmohrin: hmm, I feel that from the issue and our usage of `get_main_evaluation` below, we should really do
```suggestion
        if main_language == Evaluation.UNDECIDED_MAIN_LANGUAGE and data["isexam"]:
```

on the other hand, it is also fair to say "try all related events to find a language"; not sure, the current combination seems odd though

what do you think? I think I slightly prefer the isexam check

evap/cms/json_importer.py
```diff
@@ -189,9 +189,9 @@ def send_mail(self):
         mail.send()
 
 
-def _clean_whitespaces(text: str) -> str:
-    # Use regex for cleaning to also include non-ASCII whitespaces like non-breaking whitespaces
-    return re.sub(r"\s+", " ", text.strip())
+def _clean_whitespaces_and_hyphens(text: str) -> str:
+    """Normalize to single space characters and en-dashes."""
+    return re.sub(r"\s+", " ", text.strip()).replace(" - ", " – ")
```
niklasmohrin: For clarity: Is there any good way to make this more explicit, like `f" {EN_DASH} "`? Maybe this already exists as a constant in the unicode module or so, or we make our own in this module

niklasmohrin (commented):
okay only some minor stuff left; core stuff works and makes sense :+1:

evap/cms/json_importer.py
geromequa: Yes :D

evap/cms/tests/test_json_importer.py
```diff
@@ -809,9 +811,20 @@ def test_exam_type_different_name(self):
...
-        self.assertEqual(_clean_whitespaces("back "), "back")
-        self.assertEqual(_clean_whitespaces("inbetween  inbetween"), "inbetween inbetween")
-        self.assertEqual(_clean_whitespaces("inbetween \n inbetween"), "inbetween inbetween")
+        self.assertEqual(_clean_whitespaces_and_hyphens(" front"), "front")
+        self.assertEqual(_clean_whitespaces_and_hyphens("back "), "back")
+        self.assertEqual(_clean_whitespaces_and_hyphens("inbetween  inbetween"), "inbetween inbetween")
+        self.assertEqual(_clean_whitespaces_and_hyphens("inbetween \n inbetween"), "inbetween inbetween")
         # non-breaking whitespace
-        self.assertEqual(_clean_whitespaces("inbetween  inbetween"), "inbetween inbetween")
+        self.assertEqual(_clean_whitespaces_and_hyphens("inbetween  inbetween"), "inbetween inbetween")
```
niklasmohrin, line 819: by the way, do we even have a non breaking space here? as far as I can tell, we don't, and not on main either @richardebeling

niklasmohrin (approved):
Nice, thanks!

evap/cms/tests/test_json_importer.py
richardebeling: If we want to test with a non-breaking space, let's make it explicitly visible in all editors by using `\u00A0` or whatever the appropriate string literal escape sequence is.

PR #2635, opened 2026-02-09
Make JSONImporter pass assert_no_database_modifications on nop runs

### LINKED ISSUE #2628: CMS Importer: "No-op" runs shouldn't modify data in the database
Follow-up to https://github.com/e-valuation/EvaP/issues/2576

In https://github.com/e-valuation/EvaP/pull/2547#discussion_r2574989166, we wanted to use our `with assert_no_database_modifications()` context around importer runs that we don't expect to perform any changes. Specifically, this affects `test_import_courses_evaluation_approved`, and maybe some other tests.

At the time, we thought this wasn't possible due to a bug in the logging framework, which logs m2m field changes even if nothing semantically changes. The logging side was fixed in github.com/e-valuation/EvaP/pull/2584, but in that PR, we couldn't re-enable the "no database modifications" assertions because there are other data-modifying queries in the importer.

We should look into these queries, probably fix them, and come to a state where those tests pass without database modifications.

niklasmohrin:
Close #2628

### DIFF
--- a/evap/cms/json_importer.py
+++ b/evap/cms/json_importer.py
@@ -479,9 +479,13 @@ class JSONImporter:
                 wait_for_grade_upload_before_publishing = True
             else:
                 wait_for_grade_upload_before_publishing = any(grade["scale"] for grade in data["courses"])
-            course.evaluations.all().update(
+
+            if course.evaluations.exclude(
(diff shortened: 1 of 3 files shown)


niklasmohrin:
I would go for a merge commit here so that the first commit message sticks around

PR #2643, opened 2026-02-15
Replace login keys with one-time passwords

### LINKED ISSUE #1128: Consider hashing the login_key
External user's login key is effectively their password, even though they cannot choose it themselves and we embed it in the urls so it will appear in the browsers history.

However, i couldn't think of a good reason why we should not hash it server-side so a database breach does not give the attacker immediate access to all external accounts.

karyon:
Fixes https://github.com/e-valuation/EvaP/issues/1128

What changes:
* The OTPs are hashed and we don't have the plaintext available.
* As a result, we can't reuse existing valid OTPs. So whenever an external user gets an email, or when an external user requests one, or when exporting login URLs for an evaluation, a *new* OTP and login URL is generated for that user.
* To ensure that multiple emails to one user can contain valid login URLs, a user can have multiple valid OTPs at the same time.
* The login URL is `/otp/<plainext OTP>`, similar to `/key/` before. Old login URLs break. We could use the old url and migrate existing keys, but I thought that's not worth it.
* In #1128 we planned to salt the passwords like you would usually do, and add the email address to the login link to identify the user. I thought that 40-char random passwords are safe enough without salting and skipped this part of the design, it made a couple things easier.
* The OTPs are valid for two weeks (configurable) instead of 7 months.
* The login URL export has a separate "typeable" alphabet and shorter length (8 chars, configurable). These might still be brute-force-able with db access, but at least you have only two weeks for that and you can't reasonably brute-force the login form anymore.

TODO naming consistency. There's a mix of "login link", "login url" and "login key" in user-facing messages. Login keys are gone, but I also wouldn't mention "OTP" or "password" to the user. @janno42? After deciding for one, I'd also align the code. A small complication: the email templates contain at least `user.needs_login_key`, so I'd leave that in for backwards compatibility. To be absolutely sure I would need the templates used on prod.


evap/settings.py
```diff
@@ -33,7 +33,13 @@
 
 ### EvaP logic
 
-LOGIN_KEY_VALIDITY = 210  # days, so roughly 7 months
+OTP_VALIDITY_HOURS = 14 * 24  # validity period for long (emailed) one-time passwords
```
karyon: For user-requested login links, we could reduce this to an hour or even minutes. I wasn't sure whether that's worth the additional code.

evap/evaluation/tests/test_auth.py
```diff
@@ -40,70 +38,77 @@ def setUpTestData(cls):
...
         self.assertRedirects(self.app.get(reverse("contributor:index")), "/?next=/contributor/")
 
-        url_with_key = reverse("evaluation:login_key_authentication", args=[self.external_user.login_key])
-        old_login_key = self.external_user.login_key
-        old_login_key_valid_until = self.external_user.login_key_valid_until
-        page = self.app.get(url_with_key)
-        self.external_user.refresh_from_db()
-        self.assertEqual(old_login_key, self.external_user.login_key)
-        self.assertEqual(old_login_key_valid_until, self.external_user.login_key_valid_until)
+        url = self.external_user.generate_login_url()
```
richardebeling: Should we run through this twice, once for `typeable=False` and once for `typeable=True`? (Maybe we event want to parameterize a set of tests here, so they run for both "methods"?)

evap/evaluation/models.py
```diff
@@ -2131,6 +2105,80 @@ def get_sorted_due_evaluations(self):
...
+        Enforces the per-user OTP limit.
+
+        If typeable is True, uses the TYPEABLE variants of
+        OTP_ALPHABET, OTP_LENGTH, and OTP_VALIDITY_HOURS.
+        """
+        # Keep only the newest (limit - 1) OTPs, delete everything else (expired + excess)
+        ids_to_keep = (
+            cls.objects.filter(user=user)
+            .order_by("-valid_until")
+            .values_list("id", flat=True)[: settings.MAX_OTPS_PER_USER - 1]
```
richardebeling: Is `.values_list` required here? Would expect that it works without, but maybe I'm mistaken.

evap/evaluation/views.py
```diff
@@ -132,25 +133,18 @@ def login_key_authentication(request, key):
...
-        user.save()
-    elif user:
-        # A user exists, but the login key is not valid anymore. Send the user a new one.
-        user.ensure_valid_login_key()
+        otp_hash.invalidate()
+    else:
         EmailTemplate.send_login_url_to_user(user)
         messages.warning(request, _("The login URL is not valid anymore. We sent you a new one to your email address."))
-    else:
-        messages.warning(request, _("Invalid login URL. Please request a new one below."))
```
richardebeling, line 153: Was this intentionally dropped? (I'm fine with it, not sure if @janno42 would want to keep the message though, one could say UX is worse if you're just redirected with no feedback)

evap/staff/tools.py
```diff
@@ -242,6 +242,7 @@ def merge_users(  # noqa: PLR0915  # This is much stuff to do. However, splittin
     merged_user["represented_users"] = UserProfile.objects.filter(delegates__in=[main_user, other_user]).distinct()
     merged_user["cc_users"] = UserProfile.objects.filter(ccing_users__in=[main_user, other_user]).distinct()
     merged_user["ccing_users"] = UserProfile.objects.filter(cc_users__in=[main_user, other_user]).distinct()
+    merged_user["otp_hashes"] = OtpHash.objects.filter(user__in=[main_user, other_user]).distinct()
```
richardebeling, line 245: This doesn't adhere to the max-opt-count limit, does it? Can we call some "delete expired otp entries" helper methods here to make the guarantee hold?

richardebeling (commented):
Code looks good to me, thanks. A few minor comments

evap/settings.py
richardebeling: I'd also be in favor of drastically reducing the time, but I think one use case is us sending emails to a user without them asking for it, we'd also include a login link then. By the time they see the email (possibly 1-3 days I'd say), it would then likely be expired.

evap/evaluation/models.py
```diff
@@ -2131,6 +2105,80 @@ def get_sorted_due_evaluations(self):
...
+        )
+        cls.objects.filter(user=user).exclude(id__in=ids_to_keep).delete()
+
+        if typeable:
+            alphabet = settings.OTP_ALPHABET_TYPEABLE
+            length = settings.OTP_LENGTH_TYPEABLE
+            validity = timedelta(hours=settings.OTP_VALIDITY_HOURS_TYPEABLE)
```
niklasmohrin: can we make the setting a timedelta directly?

evap/evaluation/models.py
```diff
@@ -2131,6 +2105,80 @@ def get_sorted_due_evaluations(self):
...
+        else:
+            alphabet = settings.OTP_ALPHABET
+            length = settings.OTP_LENGTH
+            validity = timedelta(hours=settings.OTP_VALIDITY_HOURS)
+
+        while True:
```
niklasmohrin: Should we do something like
```suggestion
        for _ in range(10):
```
so that we eventually notice unexpected errors?

niklasmohrin (commented):
Makes sense, I will do a more thorough read once the vocabulary is decided

evap/evaluation/views.py
karyon: This moved to the top.

evap/evaluation/tests/test_auth.py
karyon: I thought it's enough to test that with test_login_url_generation, that's the only place where there is any typeable-specific logic 🤷

Let me know if you want this.

evap/staff/tools.py
karyon: The goal of the limit was to prevent unbounded growth of the table and to prevent easy brute-force attacks by generating large amounts of keys. Both aren't affected by the merge. But no strong opinions here either, let me know.

richardebeling: If it's as simple as calling a `enforce_otp_count_limit(merged_user)` after this line, I would enforce it here as well. Keeping the constraint enforced gives us the ability to also model the constraint as a database constraint and possibly use it in the future.

I see your point though and agree that it's not critical for correctness or safety, I just think it may become handy in the future. Feel free to just resolve the discussion if this isn't easy to implement or you disagree

evap/evaluation/tests/test_auth.py
richardebeling: Hmm, I see your point, but I think for potential future changes, having an assertion for "a login URL generated with typeable=True can be used for logging in" still would make sense and would add value here, even though the OTP generation and login validation should be unrelated.

Idk, what do others think (@niklasmohrin @Kakadus)?

niklasmohrin: I agree that doing the same thing with typeable=True makes sense, don't want to think about whether one working implies the other

evap/staff/views.py
```diff
@@ -1590,10 +1590,11 @@ def evaluation_login_key_export(_request, evaluation_id):
     writer = csv.writer(response, delimiter=";", lineterminator="\n")
     writer.writerow([_("Last name"), _("First name"), _("Email"), _("Login key")])
 
-    external_participants = (participant for participant in evaluation.participants.all() if participant.is_external)
-    for participant in external_participants:
-        participant.ensure_valid_login_key()
-        writer.writerow([participant.last_name, participant.first_name, participant.email, participant.login_url])
+    for participant in evaluation.participants.all():
+        if not participant.needs_login_key:
```
niklasmohrin, line 1594: Before, we would not include proxy users without emails, now they would be included

I guess the new behavior makes more sense?

evap/evaluation/models.py
```diff
@@ -2157,6 +2131,80 @@ def get_sorted_due_evaluations(self):
...
+    user = models.ForeignKey(UserProfile, on_delete=models.CASCADE, related_name="otp_hashes")
+    otp_hash = models.CharField(max_length=256, unique=True, verbose_name=_("Hashed OTP"))
+    valid_until = models.DateTimeField(verbose_name=_("Valid until"))
+
+    class Meta:
+        verbose_name = _("OTP hash")
+        verbose_name_plural = _("OTP hashes")
+
+    @classmethod
+    def enforce_otp_count_limit(cls, user: UserProfile) -> None:
```
niklasmohrin, line 2148: when we call this from the user merge code we want to keep `limit`, not `limit - 1`; should we make this a parameter here and call it with `settings.MAX_OTPS_PER_USER - 1` in `create`, _or_ should we always use `settings.MAX_OTPS_PER_USER` here and call this only after creating the new otp?

evap/evaluation/models.py
```diff
@@ -2157,6 +2131,80 @@ def get_sorted_due_evaluations(self):
...
+        """Keep only the newest (limit - 1) OTPs for the user, delete the rest."""
+        ids_to_keep = cls.objects.filter(user=user).order_by("-valid_until")[: settings.MAX_OTPS_PER_USER - 1]
+        cls.objects.filter(user=user).exclude(id__in=ids_to_keep).delete()
+
+    @classmethod
+    def create(cls, user: UserProfile, *, typeable: bool = False) -> str:
```
niklasmohrin, line 2154: we also have the `= False` default in the userprofile, should we remove it here?

evap/evaluation/models.py
```diff
@@ -2157,6 +2131,80 @@ def get_sorted_due_evaluations(self):
...
+            return cls.objects.select_related("user").get(otp_hash=otp_hash)
+        except cls.DoesNotExist:
+            return None
+
+    def is_valid(self) -> bool:
+        return self.valid_until >= now()
+
+    def invalidate(self) -> None:
+        # send it way back so enforce_otp_count_limit deletes it before any active OTPs
+        self.valid_until = now() - timedelta(days=365)
```
niklasmohrin, line 2204: ~~should we assert somewhere that settings.OTP_VALIDITY is not longer than this?~~ nvm, anything before now() is enough

janno42 (commented):
> TODO naming consistency. There's a mix of "login link", "login url" and "login key" in user-facing messages. Login keys are gone, but I also wouldn't mention "OTP" or "password" to the user. @janno42? After deciding for one, I'd also align the code. A small complication: the email templates contain at least user.needs_login_key, so I'd leave that in for backwards compatibility. To be absolutely sure I would need the templates used on prod.

Why wouldn't you use "one-time password"? I'd say that's a common name for this feature and would be okay to use it in the UI.
We also don't need to keep `needs_login_key`, I'll make sure the templates will be updated accordingly.

PR #2644, opened 2026-02-16
Skip MacOS CI check if nix files are unchanged

niklasmohrin:
Came up during #2615, but useful already before I figure that one out.

Notably, we don't even request the MacOS runner if the files are not changed.


niklasmohrin:
See the last two commits for the behavior

PR #2645, opened 2026-02-16
Display current version in footer

### LINKED ISSUE #2638: Display current version in footer
The footer of the website should show the currently running version. In particular, the "Github project" link should be removed and instead it should say "Version: xxxx.xx.xx" and link to the corresponding github release (for example https://github.com/e-valuation/EvaP/releases/tag/2026.2.0). The version should be retrieved according to our pyproject.toml file and using whatever method is standard for this purpose (a little research for this should be done).

suttrop:
closes #2638

### DIFF
--- a/evap/evaluation/templates/footer.html
+++ b/evap/evaluation/templates/footer.html
@@ -16,8 +16,8 @@
                 </li>
                 <li class="d-none d-sm-block nav-item">
                     <a class="nav-link" href="https://github.com/e-valuation/EvaP/releases/tag/{% release_version %}" target="_blank">
-                        <span class="fas fa-arrow-up-right-from-square me-1"></span>
-                        <span> Version: {% release_version %}</span>
+                        <span> Version: {% release_version %} </span>
+                        <span class="fas fa-arrow-up-right-from-square ps-1"></span>
                     </a>
                 </li>
             </ul>


evap/evaluation/templatetags/footer_templatetags.py
```diff
@@ -0,0 +1,10 @@
+from importlib.metadata import version
```
niklasmohrin: nice

evap/evaluation/templates/footer.html
```diff
@@ -13,7 +15,7 @@
                     <a class="nav-link nav-link-multiline" href="{% url 'evaluation:legal_notice' %}">{% translate 'Legal Notice' %}<br />{% translate 'Data Privacy' %}</a>
                 </li>
                 <li class="d-none d-sm-block nav-item">
-                    <a class="nav-link" href="https://github.com/e-valuation/EvaP" target="_blank"><span class="fas fa-arrow-up-right-from-square"></span> {% translate 'GitHub project' %}</a>
+                    <a class="nav-link" href="https://github.com/e-valuation/EvaP/releases/tag/{% release_version %}" target="_blank"><span class="fas fa-arrow-up-right-from-square"></span> Version: {% release_version %}</a>
```
niklasmohrin: Can we try the following?

Make it so that the icon stays centered as is, but "Version" and the version number are two lines like "legal notice" and "data privacy"?

I feel like it would look slightly nicer :)

evap/evaluation/templatetags/footer_templatetags.py
```diff
@@ -0,0 +1,10 @@
...
+
+from django.template import Library
+
+register = Library()
+
+
+@register.simple_tag
+def release_version():
+    return version("evap")
```
jooooosef: This could throw a `PackageNotFoundError` when evap is not installed. This is unlikely but since this is on every page, entire EvaP would be unusable.

We could just do a try except @niklasmohrin ?

niklasmohrin: I don't think this is a problem because in the dev setup we have an editable install and in production we have a proper installation. I guess if things would go wrong, it would be helpful to get the error immediately and not only when rendering a template. So we should move this out of the function and do this call for a module level constant

niklasmohrin: as discussed in person with @suttrop , we should probably also not compute this multiple times because it is a bit costly: it traverses the python path for all packages until a matching one is found

it's not noticeably slow, but I also don't want to think about it becoming slow if we have lots of packages installed (for example in the system site-packages)

PR #2649, opened 2026-02-23
Importer evaluation merge and other things

janno42:
see commit messages, best reviewed individually

### DIFF
--- a/evap/development/fixtures/test_data.json
+++ b/evap/development/fixtures/test_data.json
@@ -788,7 +788,6 @@
     "type": 1,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       1
     ],
@@ -807,7 +806,6 @@
     "type": 2,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       2
     ],
@@ -826,7 +824,6 @@
     "type": 1,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       2
     ],
@@ -845,7 +842,6 @@
     "type": 3,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       1
     ],
@@ -864,7 +860,6 @@
     "type": 3,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       2
     ],
@@ -883,7 +878,6 @@
     "type": 1,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       2
     ],
@@ -902,7 +896,6 @@
     "type": 4,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       1
     ],
@@ -921,7 +914,6 @@
     "type": 2,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       1
     ],
@@ -940,7 +932,6 @@
     "type": 2,
     "is_private": true,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       2
     ],
@@ -959,7 +950,6 @@
     "type": 1,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       2
     ],
@@ -978,7 +968,6 @@
     "type": 2,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       2
     ],
@@ -997,7 +986,6 @@
     "type": 2,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       2
     ],
@@ -1016,7 +1004,6 @@
     "type": 1,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       2
     ],
@@ -1035,7 +1022,6 @@
     "type": 2,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       1
     ],
@@ -1054,7 +1040,6 @@
     "type": 3,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       1
     ],
@@ -1073,7 +1058,6 @@
     "type": 2,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       1
     ],
@@ -1092,7 +1076,6 @@
     "type": 2,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       2
     ],
@@ -1111,7 +1094,6 @@
     "type": 3,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       2
     ],
@@ -1130,7 +1112,6 @@
     "type": 2,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       2
     ],
@@ -1149,7 +1130,6 @@
     "type": 3,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       2
     ],
@@ -1168,7 +1148,6 @@
     "type": 3,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       2
     ],
@@ -1187,7 +1166,6 @@
     "type": 2,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       2
     ],
@@ -1206,7 +1184,6 @@
     "type": 2,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       2
     ],
@@ -1225,7 +1202,6 @@
     "type": 2,
     "is_private": false,
     "gets_no_grade_documents": false,
-    "cms_id": null,
     "programs": [
       2
     ],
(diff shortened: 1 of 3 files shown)


evap/cms/tests/test_json_importer.py
```diff
@@ -334,7 +334,7 @@ def test_import_courses(self):
         course = Course.objects.get()
 
         self.assertEqual(course.semester, self.semester)
-        self.assertEqual(course.cms_id, EXAMPLE_DATA["events"][0]["gguid"])
+        self.assertEqual(course.course_links.first().cms_id, EXAMPLE_DATA["events"][0]["gguid"])
```
niklasmohrin: wouldn't it be nicer if we would use `course.cms_links` as the reverse name? (same for evaluation)

niklasmohrin: also, I suppose it should be `list(course.course_links.values_list(flat=True)) == [...]`

evap/cms/templatetags/cms_filters.py
```diff
@@ -0,0 +1,19 @@
...
+
+register = Library()
+
+
+@register.filter
+def has_active_links(links):
+    for link in links:
+        if link.is_active:
+            return True
+    return False
```
niklasmohrin:
```suggestion
    return any(link.is_active for link in links)
```

evap/cms/templatetags/cms_filters.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+@register.filter
+def has_deactivated_links(links):
+    for link in links:
+        if not link.is_active:
+            return True
+    return False
```
niklasmohrin:
```suggestion
    return any(not link.is_active for link in links)
```

evap/cms/json_importer.py
```diff
@@ -365,38 +367,36 @@ def _import_course(self, data: ImportEvent, course_type: CourseType | None = Non
             )
             return None
 
+        if not data["title_en"]:
+            data["title_en"] = data["title"]
+
         try:
-            course = Course.objects.get(cms_id=data["gguid"])
+            course = Course.objects.get(course_links__cms_id=data["gguid"])
+            if CourseLink.objects.get(cms_id=data["gguid"]).is_active:
```
niklasmohrin: Can't we `.get` the link first and then have `course = link.course`?

evap/cms/json_importer.py
```diff
@@ -426,14 +426,18 @@ def _import_course_from_unused_exam(self, data: ImportEvent) -> Course | None:
 
     # pylint: disable=too-many-locals
     def _import_evaluation(  # noqa: PLR0912, PLR0915
-        self, course: Course, data: ImportEvent, earliest_exam_date: date | None = None
+        self, course: Course, data: ImportEvent, all_events: list[ImportEvent], earliest_exam_date: date | None = None
```
niklasmohrin: we have something very similar in #2632 as we also need to access other events there to find the corresponding main evaluation of an exam

I guess we should merge this one first and then rebase here and use `self.events_by_gguid` here too?

evap/cms/views.py
```diff
@@ -16,3 +26,63 @@ def ignored_evaluation_delete(request):
     )
     ignored_evaluation.delete()
     return HttpResponseNoContent()
+
+
+@require_POST
+@manager_required
+def toggle_course_link_activation(request):
```
niklasmohrin: Can we make them _not_ toggle and instead give them "set to x" semantics? This way, the request is idempotent and we avoid two users toggling with the same intention

We also did this for textanswers/update_flag for example

evap/cms/templates/cms_evaluation_link_badges.html
```diff
@@ -0,0 +1,13 @@
+{% if evaluation.evaluation_links.exists %}
+    <div class="d-flex">
+        {% for evaluation_link in evaluation.evaluation_links.all %}
+            <form id="evaluation-link-activation-form" method="POST" action="{% url 'cms:toggle_evaluation_link_activation' %}">
```
niklasmohrin: not used:
```suggestion
            <form method="POST" action="{% url 'cms:toggle_evaluation_link_activation' %}">
```

evap/cms/views.py
```diff
@@ -16,3 +26,63 @@ def ignored_evaluation_delete(request):
...
+    course_link = get_object_from_dict_pk_entry_or_logged_40x(CourseLink, request.POST, "course_link_id")
+
+    course_link.is_active = not course_link.is_active
+    course_link.save()
+
+    return HttpResponseRedirect(request.META.get("HTTP_REFERER"))
```
niklasmohrin: Can we return just a normal 200 instead and use `reload-on-success` on the form in the badge?

evap/cms/templatetags/cms_filters.py
niklasmohrin, line 1: Why is it `active/deactivated` and not `active/inactive` or `activated/deactivated`?

evap/cms/tests/test_json_importer.py
janno42: wouldn't it be confusing that `course.cms_links` returns `CourseLink`s?

niklasmohrin: Hm, fair point. I feel that from the pov of a course, I cannot grasp what a course link is, it could well be a link to the moodle or something. That's why I feel the word CMS must be in there. Now if a course has a "CMS link", I would be fine if that is a specialized type evap.cms.models.CourseLink

I would also be fine with `cms_course_links`

evap/cms/templates/cms_course_link_icons.html
richardebeling, line 1: Note to self: We'll have to ensure this is always prefetched. I don't know if we can easily assert that this was prefetched

evap/cms/templates/cms_evaluation_merge_selection.html
```diff
@@ -0,0 +1,31 @@
+{% extends 'staff_base.html' %}
+
+{% block title %}
+   {% translate 'Merge evaluations' %}
+    {{ block.super }}
```
richardebeling: indentation

evap/cms/forms.py
```diff
@@ -0,0 +1,27 @@
...
+        self.fields["main_evaluation"].queryset = Evaluation.objects.filter(course__semester__pk=semester_id).distinct()
+        self.fields["main_evaluation"].initial = main_evaluation.pk
+        self.fields["other_evaluation"].queryset = Evaluation.objects.filter(
+            course__semester__pk=semester_id
+        ).distinct()
+
+    def clean(self):
+        super().clean()
+        if self.cleaned_data.get("main_evaluation") == self.cleaned_data.get("other_evaluation"):
+            raise ValidationError(_("You must select two different evaluations."))
```
richardebeling, line 27: Is this necessary if we set the queryset of the field to not include the main evaluation? I would have expected django to disallow this

evap/cms/json_importer.py
```diff
@@ -567,21 +582,24 @@ def _import_evaluation(  # noqa: PLR0912, PLR0915
...
-                cms_id=data["gguid"],
                 # we only want to set names on creation to avoid renumbering
                 name_de=name_de,
                 name_en=name_en,
                 **defaults,
             )
+            EvaluationLink.objects.create(evaluation=evaluation, cms_id=data["gguid"])
 
-        allow_evaluation_changes = evaluation.state == Evaluation.State.NEW
+        # Only allow changes for new evaluations and if they have not more than one evaluation link
```
richardebeling, line 592: The comment also sounds like it covers "participants", but we accept those. I don't know if `non-participant changes" makes it clearer

evap/cms/views.py
```diff
@@ -16,3 +26,71 @@ def ignored_evaluation_delete(request):
...
+
+    if form.is_valid():
+        main_evaluation = form.cleaned_data["main_evaluation"]
+        other_evaluation = form.cleaned_data["other_evaluation"]
+
+        with transaction.atomic():
+            for evaluation_link in EvaluationLink.objects.filter(evaluation=other_evaluation):
+                evaluation_link.evaluation = main_evaluation
+                evaluation_link.save()
+
```
richardebeling: should we just
```suggestion
            EvaluationLink.objects.filter(evaluation=other_evaluation).update(evaluation=main_evaluation)´
```
?

evap/cms/views.py
```diff
@@ -16,3 +26,71 @@ def ignored_evaluation_delete(request):
...
+            participant_importer_log = import_persons_from_evaluation(
+                ImportType.PARTICIPANT, main_evaluation, test_run=False, source_evaluation=other_evaluation
+            )
+            contributor_importer_log = import_persons_from_evaluation(
+                ImportType.CONTRIBUTOR, main_evaluation, test_run=False, source_evaluation=other_evaluation
+            )
```
richardebeling, line 80: These two methods also mark the users as active. Is that intended here?

evap/evaluation/migrations/0163_migrate_cms_links.py
```diff
@@ -0,0 +1,61 @@
...
+    EvaluationLink = apps.get_model("cms", "EvaluationLink")
+    CourseLink = apps.get_model("cms", "CourseLink")
+
+    for evaluation_link in EvaluationLink.objects.all():
+        evaluation_link.evaluation.cms_id = evaluation_link.cms_id
+        evaluation_link.delete()
+
+    for course_link in CourseLink.objects.all():
+        course_link.course.cms_id = course_link.cms_id
+        course_link.delete()
```
richardebeling: Note, so we keep this fully in mind: This is not lossless, we may have ended up with multiple links per evaluation, we're now essentially selecting one at random

evap/cms/tests/test_json_importer.py
janno42: `assertQuerysetEqual`

evap/staff/templates/staff_semester_view_evaluation.html
janno42, line 1: Changes:  
- Action buttons equidistant
- Preview in the middle (so only left can be empty if no other action)
- `ellipsis-vertical` for dropdown button icon

evap/cms/views.py
janno42: I'd say that's fine since all of the participants and contributors of the currently used evaluations are active anyway.

evap/cms/forms.py
janno42: As discussed, both fields can be changed in the form and need the complete queryset.

PR #2657, opened 2026-03-02
Update contact modal

### LINKED ISSUE #2129: Use web components for contact and direct delegation modals
In #1985, we updated all confirmation modals to a browser-native web component approach. It would be nice to do something similar with the contact and direct delegation modals. The PR had an attempt at the direct delegation modal but it had some issues with the dropdown select field.

styrix560:
Fixes #2129


evap/student/tests/test_live.py
```diff
@@ -28,7 +28,6 @@ def setUp(self) -> None:
         voting_user2 = baker.make(UserProfile, email="voting_user2@institution.example.com")
         contributor1 = baker.make(UserProfile, email="contributor1@institution.example.com")
         contributor2 = baker.make(UserProfile, email="contributor2@institution.example.com")
-
```
fekoch, line 31: accidental change?

evap/evaluation/templates/contact_modal.html
```diff
@@ -1,55 +1,47 @@
...
 <script type="module">
-    import { ContactModalLogic } from "{% static 'js/contact_modal.js' %}";
+   const messageTextElement = document.querySelector("#{{ modal_id }}MessageText");
 
-    new ContactModalLogic("{{ modal_id }}", "{{ title|escapejs }}").attach();
+    function onSuccess() {
+        messageTextElement.value = "";
+        window.location.reload();
+        console.log("here");
+    }
```
fekoch: debugging code?

fekoch (changes requested):

evap/static/scss/components/_buttons.scss
```diff
@@ -145,9 +145,6 @@ a:not([href]):not(.disabled) {
 
 .feedback-button {
     @include media-breakpoint-up(md) {
-        position: fixed;
```
styrix560, line 148: I will revert this also

evap/evaluation/templates/contact_modal.html
```diff
@@ -1,55 +1,46 @@
...
-            <div class="modal-body">
-                {% translate 'Your message was successfully sent.' %}
+<form custom-success id="contact-modal-form-{{ modal_id }}" method="POST" action="{% url 'evaluation:contact' %}">
+    {% csrf_token %}
+    <input hidden name="title" value="{{ title }}"/>
+
+    <confirmation-modal type="submit">
+        <span slot="title" id="{{ modal_id }}Title">{{ title }}</span>
+        <span slot="question" id="{{ modal_id }}Teaser">{{ teaser }}</span>
+        <span slot="action-text" id="{{ modal_id }}Teaser">{% translate 'Send Message' %}</span>
```
richardebeling: These two ids are not required anymore I think (and reusing the same ID twice is not legal in HTML)

evap/evaluation/templates/base.html
```diff
@@ -30,14 +30,6 @@
 
         {% include "custom_elements.html" %}
 
-        {% block modals %}
```
richardebeling, line 33: If we remove the block here, we must ensure that noone else "uses"/"overrides" this block anymore. At least the index page still mentions this

evap/evaluation/templates/contribution_formset.html
```diff
@@ -5,9 +5,10 @@
         <h5 class="card-title me-auto">{% translate 'Contributors' %}</h5>
         {% if not manager %}
             <div>
-                <button type="button" class="btn btn-sm btn-light" id="createAccountRequestModalShowButton">
-                    {% translate 'Request creation of new account' %}
-                </button>
+                {% blocktranslate asvar title with evaluation_name=evaluation.full_name %}Request account creation for {{ evaluation_name }}{% endblocktranslate %}
+                {% translate 'Please tell us which new account we should create. We need the name and email for all new accounts.' as teaser %}
+                {% translate 'Request creation of new account' as show_button_text %}
+                {% include 'contact_modal.html' with modal_id='createContributorAccountRequestModal' user=request.user title=title teaser=teaser show_button_text=show_button_text show_button_class="btn btn-sm btn-light mb-3"%}
```
richardebeling, line 11: I'm not sure including the modal .html (including the javascript block) makes sense here -- this file will be included multiple times into an outer page, right?

evap/evaluation/templates/contact_modal.html
```diff
@@ -1,55 +1,46 @@
...
+    </confirmation-modal>
+</form>
 <script type="module">
-    import { ContactModalLogic } from "{% static 'js/contact_modal.js' %}";
+   const messageTextElement = document.querySelector("#{{ modal_id }}MessageText");
 
-    new ContactModalLogic("{{ modal_id }}", "{{ title|escapejs }}").attach();
+    function onSuccess() {
+        messageTextElement.value = "";
+        window.location.reload();
```
richardebeling: The old logic didn't have a forced page reload -- can we get by without it? I would expect that the "old" mechanism to show a success message should just as well work here, or is there some problem with it?

styrix560: We changed to way we display the success message to the banner on top. Therefore, we need to reload here to show it.

evap/evaluation/templates/contribution_formset.html
styrix560: I don't think so. What makes you say that?

richardebeling: Discussed offline: if we re-use a "reloadOnSuccess" modal tag, we don't need custom java-script, then including this ~25ish times is probably ok

evap/evaluation/templates/contact_modal.html
richardebeling: Discussed offline: reload-on-success was approved by @janno42 and @niklasmohrin, fine with me then

evap/evaluation/tests/test_live.py
```diff
@@ -11,11 +11,11 @@ def test_contact_modal(self) -> None:
         self.selenium.find_element(By.ID, "feedbackModalShowButton").click()
         self.wait.until(visibility_of_element_located((By.ID, "feedbackModalMessageText")))
         self.selenium.find_element(By.ID, "feedbackModalMessageText").send_keys("Test message")
-        self.selenium.find_element(By.ID, "feedbackModalActionButton").click()
+        self.selenium.find_element(By.CSS_SELECTOR, "span[slot='action-text']").click()
```
niklasmohrin: is there any selector that ensures that we click on the _contact_ modal?

evap/contributor/tests/test_views.py
```diff
@@ -246,7 +246,7 @@ def test_contact_modal_escape(self):
         self.evaluation.save()
         page = self.app.get(self.url, user=self.responsible, status=200)
 
-        self.assertIn("changeEvaluationRequestModalLabel", page)
```
niklasmohrin, line 249: Why?

evap/contributor/templates/contributor_evaluation_form.html
niklasmohrin, line 1: Image

niklasmohrin (commented):
I think the end goal should be to not use django include for modals at all; this is definitely a step in the right direction.

evap/evaluation/templates/contribution_formset.html
```diff
@@ -5,9 +5,10 @@
         <h5 class="card-title me-auto">{% translate 'Contributors' %}</h5>
         {% if not manager %}
             <div>
-                <button type="button" class="btn btn-sm btn-light" id="createAccountRequestModalShowButton">
-                    {% translate 'Request creation of new account' %}
-                </button>
+                {% blocktranslate asvar title with evaluation_name=evaluation.full_name %}Request account creation for {{ evaluation_name }}{% endblocktranslate %}
```
richardebeling, line 8: It seems to me that this isn't used on the staff evaluation edit page. On the contributor evaluation edit page, the button looks like this for me, now:

Image

Looks like it's unintended?

evap/evaluation/templates/contact_modal.html
```diff
@@ -1,55 +1,36 @@
...
-            <div class="modal-header">
-                <h5 class="modal-title" id="successMessageModalLabel_{{ modal_id }}">{% translate 'Message sent' %}</h5>
-            </div>
-            <div class="modal-body">
-                {% translate 'Your message was successfully sent.' %}
-            </div>
-        </div>
-    </div>
-</div>
+<form reload-on-success id="contact-modal-form-{{ modal_id }}" method="POST" action="{% url 'evaluation:contact' %}">
```
richardebeling, line 3: Looks like this ID is now unused -- can we remove it?

evap/evaluation/templates/contact_modal.html
```diff
@@ -1,55 +1,36 @@
...
+    {% csrf_token %}
+    <input hidden name="title" value="{{ title }}"/>
```
richardebeling, line 5: Before, we passed the title to a hidden input with `id="{{ modal_id }}Subject"` and no name. We still have the corresponding input below, but now we also have `name="title"`. Is this intended and required?

richardebeling (changes requested):

evap/contributor/tests/test_views.py
styrix560: Before, this was the code `<h5 class="modal-title" id="{{ modal_id }}Label">{{ title }}</h5>`. Now, the id also reflects, that this is the title and not some kind of label: `<span slot="title" id="{{ modal_id }}Title">{{ title }}</span>`

evap/evaluation/templates/contact_modal.html
styrix560: No, it is required for the test `test_contact_modal_escape` in `contributor:test_views`

evap/contributor/tests/test_views.py
niklasmohrin: but the "change evaluation" part changed too, I think this is the more interesting change

styrix560: I think I know why this is happening and the reason is a bit hilarious.

Our test was (kinda) supposed to fail always. Before, we included the modals unconditionally and only showed the show-buttons conditionally. Now, we also show modals conditionally. Where before both modals lay in the DOM, now only the `createContributorAccountRequestModal` appears.

evap/evaluation/templates/contact_modal.html
styrix560: Before, we had custom logic to fire the http request, that passed the title value with key 'title' as body parameters. Because we now use html forms for this, `name="title"` submits the title the same way as before.
The input you mention was only for UI and was never submitted, because it had no name.

evap/evaluation/tests/test_live.py
styrix560: There is now :)

evap/evaluation/templates/contact_modal.html
richardebeling: But then let's just use one input instead of having two of them? I'd propose to remove this one here and juts add `name="title"` to the one below in line 26

evap/evaluation/templates/contact_modal.html
richardebeling: It does not use `contact-modal-form-...`?

evap/evaluation/templates/contact_modal.html
```diff
@@ -1,55 +1,36 @@
...
                     </div>
-                </div>
-            </form>
+                {% else %}
+                    <input type="text" class="form-control mx-auto text-break" id="{{ modal_id }}SenderName" disabled value="{{ user.full_name }}"/>
+                {% endif %}
+                <label for="{{ modal_id }}Subject" class="control-label my-auto pe-4 text-break">{% translate 'Subject' %}</label>
+                <input type="text" class="form-control mx-auto" id="{{ modal_id }}Subject" disabled value="{{ title }}"/>
+            </div>
+            <label for="{{ modal_id }}MessageText" class="d-none">Nachricht</label>
```
niklasmohrin, line 28: German text here and below, should be `translate`

evap/evaluation/tests/test_live.py
niklasmohrin: I don't see it yet :))

styrix560: Uhh, I think we click on the button in the confirmation-modal now? Or what do you mean? Maybe I'm skissuing again though

evap/evaluation/templates/contact_modal.html
styrix560: Disregard everything I said. I removed it now

evap/evaluation/tests/test_live.py
```diff
@@ -11,11 +11,11 @@ def test_contact_modal(self) -> None:
         self.selenium.find_element(By.ID, "feedbackModalShowButton").click()
```
richardebeling, line 12: I think this doesn't reliably work, we don't have a guarantee that our custom javascript handlers  have loaded when this is clicked

evap/evaluation/tests/test_live.py
```diff
@@ -11,11 +11,11 @@ def test_contact_modal(self) -> None:
...
         self.wait.until(visibility_of_element_located((By.ID, "feedbackModalMessageText")))
         self.selenium.find_element(By.ID, "feedbackModalMessageText").send_keys("Test message")
```
richardebeling: You can combine the "wait for element" and "get element" into one operation

evap/contributor/tests/test_views.py
richardebeling: But _you_ also changed the test to specify `self.evaluation.allow_editors_to_edit = True` :D

With this set to false, we show one modal, and with it set to true, we show the other modal.

https://github.com/e-valuation/EvaP/blob/69005caf86b0f20ae7c9efdba7effa34b9155cdf/evap/contributor/templates/contributor_evaluation_form.html#L67-L79

styrix560: This was a recent change, because I wanted to make it explicit, which one we show. I swapped it to False now and we show the other modal now. Is this fine for you?

evap/evaluation/tests/test_live.py
styrix560: I will fix this, but just so you know, this is not my code :)

evap/evaluation/templates/contact_modal.html
```diff
@@ -1,55 +1,34 @@
...
-                            <input type="text" class="form-control mx-auto text-break" id="{{ modal_id }}SenderName" disabled value="{{ user.full_name }}"/>
-                        {% endif %}
-                        <label for="{{ modal_id }}Subject" class="control-label my-auto pe-4 text-break">{% translate 'Subject' %}</label>
-                        <input type="text" class="form-control mx-auto" id="{{ modal_id }}Subject" disabled value="{{ title }}"/>
-                    </div>
-                    <textarea autofocus class="form-control modal-textfield my-4" id="{{ modal_id }}MessageText"></textarea>
-                    <div class="modal-submit-group">
-                        <button type="button" class="btn btn-light me-1" data-bs-dismiss="modal">{% translate 'Cancel' %}</button>
-                        <button type="submit" id="{{ modal_id }}ActionButton" class="btn btn-primary ms-1">{% translate 'Send Message' %}</button>
+    <confirmation-modal type="submit">
```
niklasmohrin: Currently the modals don't have any title, should we put something like "Contact the EvaP team" or so? @janno42

evap/evaluation/templates/contact_modal.html
```diff
@@ -1,55 +1,34 @@
...
                     </div>
-                </div>
-            </form>
+                {% else %}
+                    <input type="text" class="form-control mx-auto text-break" id="{{ modal_id }}SenderName" disabled value="{{ user.full_name }}"/>
+                {% endif %}
+                <label for="{{ modal_id }}Subject" class="control-label my-auto pe-4 text-break">{% translate 'Subject' %}</label>
+                <input name="title" type="text" class="form-control mx-auto" id="{{ modal_id }}Subject" disabled value="{{ title }}"/>
+            </div>
+            <label for="{{ modal_id }}MessageText" class="d-none">Nachricht</label>
```
niklasmohrin: German text is still there, please check the diff for more occurrences (there is a second one below)

evap/evaluation/templates/contact_modal.html
```diff
@@ -1,55 +1,34 @@
...
+            <textarea autofocus class="form-control modal-textfield my-4" id="{{ modal_id }}MessageText" name="message" placeholder="{% translate 'Nachricht' %}"></textarea>
         </div>
-    </div>
-</div>
-<script type="module">
-    import { ContactModalLogic } from "{% static 'js/contact_modal.js' %}";
 
-    new ContactModalLogic("{{ modal_id }}", "{{ title|escapejs }}").attach();
-</script>
+        <div class="{% if modal_id == "feedbackModal" %}feedback-button{% endif %}" slot="show-button">
```
niklasmohrin: I feel like we shouldn't have any `if id == ...` with this new setup, the place where we include this with `id == feedbackModal` should have a class around the modal and in the css we can use a selector like `#feedback-modal .show-button` or so instead of `.feedback-button`.

evap/evaluation/templates/contact_modal.html
```diff
@@ -1,55 +1,34 @@
...
+        <span slot="action-text">{% translate 'Send Message' %}</span>
+
+        <div slot="extra-inputs">
+            <div class="modal-grid">
+                <label for="{{ modal_id }}SenderName" class="control-label my-auto pe-4">{% translate 'Sender' %}</label>
+                {% if modal_id == "feedbackModal" and allow_anonymous_feedback_messages %}
+                    <div class="btn-group text-wrap" role="group" aria-label="{{ modal_id }} Radio Group">
+                        <input type="radio" class="btn-check" name="anonymous" value="false" id="{{ modal_id }}SenderName" checked />
+                        <label class="btn btn-sm btn-outline-primary text-break" for="{{ modal_id }}SenderName">{{ user.full_name }}</label>
+                        <input type="radio" class="btn-check" name="anonymous" value="true" id="{{ modal_id }}AnonymousName"/>
```
niklasmohrin: Hm, this is not necessarily something for this PR but we don't need the `{{ modal_id }}Something` ids because we can just as well have an id on the element that wraps this modal and also use nested selectors

niklasmohrin (changes requested):
Tests look all good, some final touches:

evap/evaluation/templates/contact_modal.html
styrix560: I think I tried this a while back and I think it broke a few tests. Either way is fine for me

evap/evaluation/templates/contact_modal.html
styrix560: What do you think of the added `show_button_class` attribute for feedback-modal?

evap/evaluation/templates/contact_modal.html
styrix560: I mistakenly removed the title at some point. I added it back again. Though the modal looks a bit weird now. should we set some kind of width?
image

niklasmohrin: I guess we just want the subject line to be as wide as is needed for the given text, something like `width: min(80ch, fit-content)` or so, not sure if this is correct

evap/evaluation/templates/contact_modal.html
niklasmohrin: makes sense

evap/static/scss/components/_buttons.scss
```diff
@@ -143,21 +143,22 @@ a:not([href]):not(.disabled) {
...
+
+    [slot=show-button] {
+        @include media-breakpoint-up(md) {
+            position: fixed;
+            right: 10px;
+            bottom: 8px;
+            z-index: $zindex-fixed;
+        }
+
+        button {
```
niklasmohrin:
```suggestion
        > button {
```
or just add an explicit class on the show button and write `.feedback-modal-show-button` here; no strong opinion

evap/evaluation/templates/contact_modal.html
```diff
@@ -1,55 +1,35 @@
...
                     </div>
-                </div>
-            </form>
+                {% else %}
+                    <input type="text" class="form-control mx-auto text-break" id="{{ modal_id }}SenderName" disabled value="{{ user.full_name }}"/>
+                {% endif %}
+                <label for="{{ modal_id }}Subject" class="control-label my-auto pe-4 text-break">{% translate 'Subject' %}</label>
+                <input name="title" type="text" class="form-control mx-auto" id="{{ modal_id }}Subject" disabled value="{{ title }}"/>
+            </div>
+            <label for="{{ modal_id }}MessageText" class="d-none">{% translate 'Nachricht' %}</label>
```
niklasmohrin: This is not how we do this :D 
```suggestion
            <label for="{{ modal_id }}MessageText" class="d-none">{% translate 'Message' %}</label>
```

also below

richardebeling (approved):

evap/evaluation/templates/contact_modal.html
richardebeling: fit-content doesn't work for input fields. Since the current contact modal just hardcodes a width, I said it's probably fine if we continue hardcoding a width for now.

(Alternative would have been to evaluate the text width in javascript and update the width of the element manually, but this seemed more annoying and brittle to me, given that we have only a fixed set of constant subjects that could be shown)

niklasmohrin: hm, there is also `field-sizing`, but it doesnt work in firefox. sure, lets hardcode something that looks okay then :)

janno42 (changes requested):
- Submitting the "Request changes" modal results in a `MultiValueDictKeyError`.
- The modal behaves strangely on a small screen with a low Y resolution. It should be made scrollable instead of the contents being placed out of view.

evap/contributor/templates/contributor_evaluation_form.html
```diff
@@ -69,15 +69,19 @@ <h5 class="card-title">{% translate 'Course data' %}</h5>
                     <h5 class="card-title me-auto">{% translate 'Evaluation data' %}</h5>
                     {% if evaluation.allow_editors_to_edit %}
                         <div>
-                            <button type="button" class="btn btn-sm btn-light mb-3 createAccountRequestModalShowButton">
-                                {% translate 'Request creation of new account' %}
-                            </button>
+                            {% blocktranslate asvar title with evaluation_name=evaluation.full_name %}Request account
```
niklasmohrin: if a blocktranslate contains line breaks like this, then we should use `blocktranslate trimmed`; then, we should format this like
```
{% blocktranslate ... %}
    Request ...
{% endblocktranslate %}
```

evap/evaluation/templates/contact_modal.html
niklasmohrin, line 1: So we don't need `modal_id` anymore right? All the labels and inputs we can use implicit association (putting the input _inside_ the label) and all the other ids are not needed. I think this would make this a bit nicer :)

evap/static/ts/src/confirmation-modal.ts
```diff
@@ -24,6 +24,11 @@ export class ConfirmationModal extends HTMLElement {
         const confirmButtonExtraClass = this.getAttribute("confirm-button-class") ?? "btn-primary";
         confirmButton.className += " " + confirmButtonExtraClass;
 
+        const dialogExtraClass = this.getAttribute("dialog-extra-class");
```
niklasmohrin, line 27: we use this only one time to add overflow-scroll; should we just always put `overflow: scoll` into the css for the confirmation modal, or does that break anything?

styrix560: Incidentally, this is my last PR where I put `overflow: show` into the css so that the TomSelect input does not get clipped at the confirmation modal border

evap/evaluation/templates/contact_modal.html
styrix560: So I referred to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/label) when I wrote this code. It says

> Generally, we recommend using explicit association with the for attribute, to ensure compatibility with external tools and assistive technologies. In fact, you can simultaneously nest and provide id/for for maximum compatibility.

If this does not matter for us, I think we can remove the ID

niklasmohrin: implicit association all the way

evap/static/ts/src/confirmation-modal.ts
niklasmohrin: okay, let's keep it like this for now

evap/evaluation/templates/contact_modal.html
styrix560: We looked at this, but the CSS classes all assume that labels and inputs lie next to each other, which makes this refactoring harder than we thought. So we decided to postpone this refactoring for now.

PR #2513, opened 2025-09-24
Separate questions and questionnaires

Kakadus:
close #457 

featuring... 
:ballot_box_with_check: ~11 year old issue
:ballot_box_with_check: >24k changed lines (nearly all in test data / tests though)
:ballot_box_with_check: touch every file in the git tree
:ballot_box_with_check: some non-trivial migrations

I await your comments or questions of questions with questions :)

PR #2665, opened 2026-03-08
Fix crash when deleting a heading question

karyon:
This got introduced in https://github.com/e-valuation/EvaP/pull/2513. `answer_class` throws on heading questions.

The tests didn't explicitly cover deleting a heading question, but test_delete_question would still occasionally do it depending on the status of the RNG when running with --shuffle, because it didn't specify the question type so model baker would choose one at random. I saw this failing in a PR check for #2643.

### DIFF
--- a/evap/staff/tests/test_views.py
+++ b/evap/staff/tests/test_views.py
@@ -3374,6 +3374,7 @@ class TestQuestionnaireEditView(WebTestStaffModeWith200Check):
 
     def test_delete_question(self) -> None:
         other_questionnaire = baker.make(Questionnaire)
+        # Use HEADING to test for a regression, see https://github.com/e-valuation/EvaP/pull/2665
         other_question = baker.make(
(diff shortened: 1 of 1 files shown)


evap/evaluation/models.py
```diff
@@ -1374,9 +1374,10 @@ class Meta:
         unique_together = [("question", "questionnaire")]
 
     def delete(self, using=None, keep_parents=False) -> tuple[int, dict[str, int]]:
-        assert not self.question.answer_class.objects.filter(assignment=self).exists(), (
-            "cannot delete question with answers"
-        )
+        if not self.question.is_heading_question:
+            assert not self.question.answer_class.objects.filter(assignment=self).exists(), (
+                "cannot delete question with answers"
+            )
```
Kakadus, line 1380: we could check both answer classes directly, but this is also fine I guess.

PR #2666, opened 2026-03-08
Use --parallel when running tests on CI and in precommit

karyon:
On my machine, wall clock time for "./manage.py run test" drops from 43 to 20s with 6 threads. More threads increased the overhead for creating databases more than it reduced the time for the actual tests. The github runners only have 4 cores.


.github/workflows/tests.yml
```diff
@@ -43,7 +45,7 @@ jobs:
           start-db: true
 
       - name: Run tests
-        run: python manage.py test --shuffle
+        run: python manage.py test --shuffle --parallel 4
```
richardebeling: Django uses the number of physical CPU cores when you just specify `--parallel`, or `--parallel auto` ([docs](https://docs.djangoproject.com/en/6.0/ref/django-admin/#envvar-DJANGO_TEST_PROCESSES))

I think it would make sense to just use the auto detection instead of hardcoding the 4 and the 6 here. For 4, you said that this is equal to the number of cores. If GitHub runners change in the future, with "auto", everything keeps working optimally.

For the dev environment, we have people running in different setups, so I'd also say that using the core count is a better approximation. How many cores does your system have, 8? How much slower was it with 8 threads compared to 6?

.github/workflows/tests.yml
niklasmohrin, line 1: Are we going to regret this when trying to reproduce CI failures locally? It says in the docs that

> Django distributes test cases — [unittest.TestCase](https://docs.python.org/3/library/unittest.html#unittest.TestCase) subclasses — to subprocesses. If there are fewer test case classes than configured processes, Django will reduce the number of processes accordingly.

but does that mean if I know the shuffle seed and use `parallel=4` that I will get the same test order?

.github/workflows/tests.yml
niklasmohrin: Definitely agree for CI, for dev I am not so sure, don't want any surface laptops to go up in flames :D Maybe cpucount/2 or so?

richardebeling: Django can only do "auto", anything else would require code from us. I think "auto" will be fine, even for the weaker laptops. Note that this is only `precommit`. For `test`, we don't change the default.

.github/workflows/tests.yml
richardebeling: Django's parallel test execution parallelizes by test class. We will get all test methods inside one test class executed in the same multiprocessing worker process. Test order within a class will be identical to what it would be with non-parallel execution. So, for every cross-test interaction within the same class, we will get the same behavior with parallel and serial execution.

Side-effects that are caused from cross-test-class interactions (python, database, cache) will be dependent on the actual execution timing. But I don't recall ever having problems with this, and it would also be an obvious first thing to check if a test starts being flaky.

We have code for resetting `baker` to the same seed before every test in https://github.com/e-valuation/EvaP/pull/2691, which would allow reproducing the exact test baker-created setting for an individual test method, independently of whether other test classes were running in parallel, or what the test filter was.

I think this will be fine, so I'd suggest to merge parallel=auto here and see if we get any problems. We can always go back if something starts breaking in a way that we absolutely cannot figure out.

karyon:
I put a hardcoded value there because on my machine with a i7-13800H (6 P-cores with hyperthreadig, 8 E-cores) it would start 20 processes and actually run quite a bit slower while burning much more CPU cycles than with 6 processes. When I opened the PR I thought 6 was an "organic" sweet spot based on my measurements, but it's suspiciously equal to the number of P-cores :) I would still think that in the absence of a percentage option a hardcoded 4-8 would be better for most, but I don't mind much. Feel free to edit the PR and merge it :)

.github/workflows/tests.yml
richardebeling, line 1: Test `test_evaluation_grid_sorting` is flaky I think, should be fixed with the rewrite in https://github.com/e-valuation/EvaP/pull/2691

PR #2669, opened 2026-03-16
Remove `@types/jquery`

niklasmohrin:
We only use jquery in one place and convert to DOM immediately. We cannot update the types because we do not update the vendored jquery. In conclusion, these types do not spark any joy and cause only pain. I want them gone. Also see https://github.com/e-valuation/EvaP/pull/2655#issuecomment-4027362643

### DIFF
--- a/evap/static/ts/src/sortable-form.ts
+++ b/evap/static/ts/src/sortable-form.ts
@@ -28,12 +28,14 @@ export function makeFormSortable(
     }
 
     // This is the only remaining jQuery usage, since formset requires jQuery
+    // eslint-disable-next-line
     $(`#${tableId} tbody tr`).formset({
         prefix: prefix,
         deleteCssClass: removeAsButton ? "btn btn-danger btn-sm" : "delete-row",
         deleteText: removeAsButton ? '<span class="fas fa-trash"></span>' : window.gettext("Delete"),
         addText: window.gettext("add another"),
         added: function (rowJQuery: any) {
+            // eslint-disable-next-line
             const row = rowJQuery.get()[0];
             assert(row instanceof HTMLTableRowElement);
             row.querySelectorAll<HTMLInputElement>("input[id$=-order]").forEach(input => {


evap/static/ts/src/sortable-form.ts
```diff
@@ -47,8 +33,9 @@ export function makeFormSortable(
         deleteCssClass: removeAsButton ? "btn btn-danger btn-sm" : "delete-row",
         deleteText: removeAsButton ? '<span class="fas fa-trash"></span>' : window.gettext("Delete"),
         addText: window.gettext("add another"),
-        added: function (rowJQuery: JQuery<HTMLTableRowElement>) {
+        added: function (rowJQuery: any) {
             const row = rowJQuery.get()[0];
```
Kakadus, line 39: we now get some eslint warnings due to the added any types. The `unknown` type fixes one. @typescript-eslint/no-unsafe-call and @typescript-eslint/no-unsafe-member-access can be disabled in this file I think

```suggestion
            const row: unknwon = rowJQuery.get()[0];
```

niklasmohrin: Good point, I feel that `any` makes more sense for the interaction with jquery though because we are not declaring a strict typescript interface that should work for all kinds of types but instead want to say "we are doing a quick detour into javascript here, trust us". Does that make sense, or am I misunderstanding unknown vs. any?

Kakadus: Effectively for us, unknown creates a compiler error when it is used, any a warning. I was suggesting unkwown because assigning any (which we do anyway without type annotations) creates another warning. When doing the ignore next line, it doesn't matter, and we do the assertion directly after that anyway.

niklasmohrin: The thing is I don't want to spell out what the type of `$` or `rowJQuery` is here; for `$` I can see us doing `($ as any)(...).formset(...)` but for the other I want to call methods on it right away and if I cast it to `any` there too then it seems silly to not write `rowJQuery: any` to begin with. Is there a better way?

Kakadus: ah, I did not mean to type `$` or `rowJQuery` as `unknown`, only `row`. `any` is fine and the warnings are what we get / want. I don't see a good way to expect / check the function types.

niklasmohrin: aaaah, makes sense; since we do the assertion right after I would just save the characters here, but if there were other branches I would also prefer unknown! Will keep in mind!

Issue #2652, opened 2026-02-23
Textanswer redirect test seems flaky


geromequa:
`test_edit_textanswer_redirect` in `evap/staff/tests/test_live.py` seems to be flaky. The test fails on a few CI runs for unrelated PRs e.g. [22320022227](https://github.com/e-valuation/EvaP/actions/runs/22320022227), [62417088524](https://github.com/e-valuation/EvaP/actions/runs/21651556025/job/62417088524) or [63018645758](https://github.com/e-valuation/EvaP/actions/runs/21839213274/job/63018645758).

richardebeling:
Code in question: https://github.com/e-valuation/EvaP/blob/9b14de119ef672406803b0c56435d59e0433a1df/evap/staff/tests/test_live.py#L233-L238

The last line fails. `self.selenium.get()` [waits for `onload` to fire](https://selenium-python.readthedocs.io/navigating.html). The span is [part of the HTML template](https://github.com/e-valuation/EvaP/blob/9b14de119ef672406803b0c56435d59e0433a1df/evap/staff/templates/staff_evaluation_textanswers_quick.html#L135) and not created by javascript, so it should exist when `onload` fires. Selenium suggest [three common error cases](https://www.selenium.dev/documentation/webdriver/troubleshooting/errors/#nosuchelementexception), I think we're looking for the correct selector at the right time, so perhaps something went wrong with the previous page load.

richardebeling:
Playing around with our frontend tests, I found various problems:
* self.selenium.get() is documented to wait for the HTML `load` event, but the specification says that this should fire as soon as resources have **loaded**. It seems that this can fire before javascript has actually _executed_, causing the test to run too early.
* Selenium has an "implicit wait" feature for any `find_element` calls. By default, this should be disabled, but it seems that some part of our infrastructure actually enables it. Removing it makes various tests fail in situations where they actually depend on waiting behavior (getting a tomselect input, wait for bootstrap carousel element, wait for anything created/removed by our javascript, wait until our click handlers are registered before clicking something, ...)
* The `test_edit_textanswer_redirect` test here was flaky on one hand because of such a timing flakiness, but on the other hand, it randomly also runs into assertion errors inside the view logic returning a 500 error page from the server. The test framework of `LiveServerTestCase` doesn't assert the return code, so we just get selenium operating with the 500 error page. The assertion we trigger is in the following stack:
  ```
  evap/staff/views.py:1670 in evaluation_textanswers:
[...]

PR #2676, opened 2026-03-16
create custom tooltip for questionnaire

### LINKED ISSUE #2591: Questionnaire preview tooltip positioning
On the staff and contributor evaluation edit pages, hovering over a questionnaire name displays the questions in a tooltip. Currently, the tooltip is not always aligned correctly next to the questionnaire labels.

Before the "plain_page" changes (b2cc6691306521d9b664dd72a7618111e050e9f2), the tooltip appeared to the left of the labels. When there was not enough space available, which happened more often after the "plain_page" was introduced, the tooltips jumped around.

#2631 moved the tooltips to the right, where there is more space, so the jumping occurs less often.
However, there are still cases in which Bootstrap/Popper thinks there isn't enough space to place the tooltip, even when there is - because Bootstrap/Popper wants to center the tooltip to the hovered element.
To fully resolve this issue, we would need to improve the positioning.

geromequa:
Closes #2591.

### DIFF
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -98,7 +98,7 @@
                     return config;
                 }
             };
-            activateTooltips("","questionnaire-tooltip", questionnaireTooltipOptions);
+            activateTooltips("", "questionnaire-tooltip", questionnaireTooltipOptions);
 
             // These are workarounds around incompatibilities of django, django-dynamic-formsets, and tomselect.
             // tomselect can't handle already tomselect'd forms that were copies made by django-dynamic-formsets'
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -91,6 +91,29 @@
             };
             activateTooltips();
 
+            activateQuestionnaireTooltips = function(selector = "") {
+                var tooltipTriggerList = [].slice.call(document.querySelectorAll(selector + ' [data-bs-toggle="questionnaire-tooltip"]'))
+                var tooltipList = tooltipTriggerList.map(function (tooltipTriggerEl) {
+                    var container = 'body';
+                    if (tooltipTriggerEl.dataset.container){
+                        container = tooltipTriggerEl.closest(tooltipTriggerEl.dataset.container);
+                    }
+                    return new bootstrap.Tooltip(tooltipTriggerEl, {container: container, html: true, trigger: 'hover', popperConfig: function (defaultBsPopperConfig) {
+                        return {
+                            placement: 'right',
+                            modifiers: [{
+                                name: 'preventOverflow',
+                                altAxis: true,
+                                rootBoundary: 'viewport',
+                            },
+                        ],
+                        };
+                    }})
+                })
+
+            };
+            activateQuestionnaireTooltips();
+
             // These are workarounds around incompatibilities of django, django-dynamic-formsets, and tomselect.
             // tomselect can't handle already tomselect'd forms that were copies made by django-dynamic-formsets'
             // "add another" button, so we take a copy of a non-tomselect'd form here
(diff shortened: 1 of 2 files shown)


geromequa:
Failed Test is known flaky test (see #2652).

evap/evaluation/templates/base.html
```diff
@@ -91,6 +91,28 @@
             };
             activateTooltips();
 
+            activateQuestionnaireTooltips = function(selector = "") {
+                var tooltipTriggerList = [].slice.call(document.querySelectorAll(selector + ' [data-bs-toggle="questionnaire-tooltip"]'))
```
richardebeling:
```javascript
[].slice.call(
```
what is this cursed scripture? :D Why can't we just assign the result of `querySelectorAll` to a (`const`!) variable, and then iterate it with a for loop like a normal grug-brain developer?

Edit: I just saw that this is fully copy-pasta from above. That's a no-no from my side. Can we reuse one function, maybe by using advanced language concepts like function call arguments? :P

I don't think we need to define this function only to then immediately invoke it? But I see that this was probably done for code structuring purposes -- talking about code structuring: can we port this to typescript? Bonus points if this code sequence doesn't appear:
```js
                                },
                            ],
                        };
                    }})
                })

            };
```

evap/evaluation/templates/base.html
```diff
@@ -91,6 +91,28 @@
...
+                        return {
+                            placement: 'right',
+                            modifiers: [
+                                {
+                                    name: 'preventOverflow',
+                                },
+                            ],
+                        };
+                    }})
+                })
```
richardebeling: This currently takes `defaultBsPopperConfig` and ignores it -- I would assume we would need to merge our config with theirs?

evap/evaluation/templates/base.html
```diff
@@ -78,19 +78,28 @@
...
-                })
-
+                    return new bootstrap.Tooltip(tooltipTriggerEl, {container: container, html: true, trigger: 'hover', ...additionalOptions});
+                });
             };
             activateTooltips();
 
+            const questionnaireTooltipOptions = {
+                template: '<div class="tooltip py-2" role="tooltip"><div class="tooltip-inner"></div></div>',
+                popperConfig: config => {
```
richardebeling, line 95: Looking up the actual documentation: It seems like we may not actually change the object we get here, but should create a new object.

https://getbootstrap.com/docs/5.0/components/popovers/#using-function-with-popperconfig

I still don't really get what specific usage pattern they want us to use here, guess we'll have to check their code a bit.

evap/evaluation/templates/base.html
```diff
@@ -78,19 +78,28 @@
...
+                    config.placement = 'right';
+                    config.modifiers = [{name: 'preventOverflow'}];
+                    return config;
+                }
+            };
+            activateTooltips("","questionnaire-tooltip", questionnaireTooltipOptions);
```
richardebeling:
```suggestion
            activateTooltips("", "questionnaire-tooltip", questionnaireTooltipOptions);
```

richardebeling (approved):
code lgtm, haven't tested the page

janno42 (approved):

evap/evaluation/templates/base.html
```diff
@@ -78,19 +78,28 @@
...
             };
             activateTooltips();
 
+            const questionnaireTooltipOptions = {
+                template: '<div class="tooltip py-2" role="tooltip"><div class="tooltip-inner"></div></div>',
+                popperConfig: config => {
+                    config.placement = 'right';
+                    config.modifiers = [{name: 'preventOverflow'}];
+                    return config;
+                }
```
richardebeling:
```suggestion
                popperConfig: {
                    placement: 'right',
                    modifiers: [{name: 'preventOverflow'}],
                }
```

Does this work? This is what I think [their tests do](https://github.com/twbs/bootstrap/blob/cb80079fc9c644957828f9c24fd9190ef3c24678/js/tests/unit/tooltip.spec.js#L159). In that case, let's just use this.

richardebeling (approved):

PR #2677, opened 2026-03-16
Use server side search for user fields

### LINKED ISSUE #1681: User selection field update
When editing an evaluation and adding participants or contributors, non-existing users first need to be created separately. It is then necessary to reload the evaluation form to have the new users available in the user selection fields.

Instead, these user selection fields should either receive their options via an AJAX request when typing a name or should get a button that reloads the list of available users.

hansegucker:
Close #1681

### DIFF
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -3,9 +3,10 @@ from datetime import datetime
 from django import forms
 from django.db.models import Q
 from django.forms.widgets import CheckboxSelectMultiple
+from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
-from evap.evaluation.forms import UserModelChoiceField, UserModelMultipleChoiceField
+from evap.evaluation.forms import ServerSearchSelect, UserModelChoiceField, UserModelMultipleChoiceField
 from evap.evaluation.models import Course, Evaluation, Questionnaire, UserProfile
 from evap.evaluation.tools import vote_end_datetime
 from evap.staff.forms import ContributionForm
@@ -155,6 +156,14 @@ class EditorContributionForm(ContributionForm):
 
 
 class DelegateSelectionForm(forms.Form):
+    @staticmethod
+    def get_delegate_to_queryset():
+        return UserProfile.objects.exclude(is_active=False).exclude(is_proxy_user=True)
+
     delegate_to = UserModelChoiceField(
-        label=_("Delegate to"), queryset=UserProfile.objects.exclude(is_active=False).exclude(is_proxy_user=True)
+        label=_("Delegate to"), queryset=get_delegate_to_queryset(), widget=ServerSearchSelect()
     )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["delegate_to"].widget.search_url = reverse("contributor:fetch_delegate_to_user_profiles")
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -1,12 +1,17 @@
 from datetime import datetime
 
 from django import forms
-from django.db.models import Q
+from django.db.models import Q, QuerySet
 from django.forms.widgets import CheckboxSelectMultiple
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
-from evap.evaluation.forms import ServerSearchSelect, UserModelChoiceField, UserModelMultipleChoiceField
+from evap.evaluation.forms import (
+    ServerSearchSelect,
+    ServerSearchSelectMultiple,
+    UserModelChoiceField,
+    UserModelMultipleChoiceField,
+)
 from evap.evaluation.models import Course, Evaluation, Questionnaire, UserProfile
 from evap.evaluation.tools import vote_end_datetime
 from evap.staff.forms import ContributionForm
@@ -43,6 +48,7 @@ class EvaluationForm(forms.ModelForm):
         field_classes = {
             "participants": UserModelMultipleChoiceField,
         }
+        widgets = {"participants": ServerSearchSelectMultiple()}
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -70,10 +76,8 @@ class EvaluationForm(forms.ModelForm):
         self.fields["vote_start_datetime"].localize = True
         self.fields["vote_end_date"].localize = True
 
-        queryset = UserProfile.objects.exclude(is_active=False)
-        if self.instance.pk is not None:
-            queryset = (queryset | self.instance.participants.all()).distinct()
-        self.fields["participants"].queryset = queryset
+        self.fields["participants"].queryset = self.get_participants_queryset(self.instance)
+        self.fields["participants"].widget.search_url = reverse("contributor:fetch_participants_user_profiles")
 
         if general_contribution := self.instance.general_contribution:
             self.fields["general_questionnaires"].initial = [
@@ -90,6 +94,13 @@ class EvaluationForm(forms.ModelForm):
         if self.instance.exam_type is not None and self.instance.cms_id is not None:
             self.fields["participants"].disabled = True
             self.cms_disclaimer = _("Participants are regularly updated with exam registrations from the CMS.")
+
+    @classmethod
+    def get_participants_queryset(cls, evaluation: Evaluation | None) -> QuerySet[UserProfile]:
+        queryset = UserProfile.objects.exclude(is_active=False)
+        if evaluation.pk is not None:
+            queryset = (queryset | evaluation.participants.all()).distinct()
+        return queryset
 
     def clean(self):
         super().clean()
@@ -157,13 +168,13 @@ class EditorContributionForm(ContributionForm):
 
 class DelegateSelectionForm(forms.Form):
     @staticmethod
-    def get_delegate_to_queryset():
+    def get_delegates_queryset():
         return UserProfile.objects.exclude(is_active=False).exclude(is_proxy_user=True)
 
     delegate_to = UserModelChoiceField(
-        label=_("Delegate to"), queryset=get_delegate_to_queryset(), widget=ServerSearchSelect()
+        label=_("Delegate to"), queryset=get_delegates_queryset(), widget=ServerSearchSelect()
     )
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.fields["delegate_to"].widget.search_url = reverse("contributor:fetch_delegate_to_user_profiles")
+        self.fields["delegate_to"].widget.search_url = reverse("contributor:fetch_delegates_user_profiles")
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -3,9 +3,10 @@ from datetime import datetime
 from django import forms
 from django.db.models import Q
 from django.forms.widgets import CheckboxSelectMultiple
+from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
-from evap.evaluation.forms import UserModelChoiceField, UserModelMultipleChoiceField
+from evap.evaluation.forms import ServerSearchSelect, UserModelChoiceField, UserModelMultipleChoiceField
 from evap.evaluation.models import Course, Evaluation, Questionnaire, UserProfile
 from evap.evaluation.tools import vote_end_datetime
 from evap.staff.forms import ContributionForm
@@ -157,6 +158,14 @@ class EditorContributionForm(ContributionForm):
 
 
 class DelegateSelectionForm(forms.Form):
+    @staticmethod
+    def get_delegate_to_queryset():
+        return UserProfile.objects.exclude(is_active=False).exclude(is_proxy_user=True)
+
     delegate_to = UserModelChoiceField(
-        label=_("Delegate to"), queryset=UserProfile.objects.exclude(is_active=False).exclude(is_proxy_user=True)
+        label=_("Delegate to"), queryset=get_delegate_to_queryset(), widget=ServerSearchSelect()
     )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["delegate_to"].widget.search_url = reverse("contributor:fetch_delegate_to_user_profiles")
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -166,7 +166,7 @@ class EditorContributionForm(ContributionForm):
 
 
 class DelegateSelectionForm(forms.Form):
-    delegate_to = UserModelChoiceField(label=_("Delegate to"), queryset=UserProfile.objects.get_delegates())
+    delegate_to = UserModelChoiceField(label=_("Delegate to"), queryset=UserProfile.objects.get_delegate_options())
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -1,12 +1,17 @@
 from datetime import datetime
 
 from django import forms
-from django.db.models import Q
+from django.db.models import Q, QuerySet
 from django.forms.widgets import CheckboxSelectMultiple
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
-from evap.evaluation.forms import ServerSearchSelect, UserModelChoiceField, UserModelMultipleChoiceField
+from evap.evaluation.forms import (
+    ServerSearchSelect,
+    ServerSearchSelectMultiple,
+    UserModelChoiceField,
+    UserModelMultipleChoiceField,
+)
 from evap.evaluation.models import Course, Evaluation, Questionnaire, UserProfile
 from evap.evaluation.tools import vote_end_datetime
 from evap.staff.forms import ContributionForm
@@ -43,6 +48,7 @@ class EvaluationForm(forms.ModelForm):
         field_classes = {
             "participants": UserModelMultipleChoiceField,
         }
+        widgets = {"participants": ServerSearchSelectMultiple()}
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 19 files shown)


evap/evaluation/templates/django/forms/widgets/server_select.html
```diff
@@ -0,0 +1,11 @@
+<select class="form-select 2" name="{{ widget.name }}"{% include 'django/forms/widgets/attrs.html' %}>
+    {% for group_name, group_choices, group_index in widget.optgroups %}
+        {% if group_name %}
+            <optgroup label="{{ group_name }}">
+        {% endif %}
+        {% for option in group_choices %}{% if option.selected %}{% include option.template_name with widget=option %}{% endif %}{% endfor %}
+        {% if group_name %}
+            </optgroup>
+        {% endif %}
+    {% endfor %}
```
richardebeling, line 10: Why is all this required? It looks like we're listing options and option groups here, which would be against the intent of this widget.

If this is required, can we maybe add a short comment on what we are doing here / why this is required?

evap/evaluation/templates/base.html
```diff
@@ -161,8 +161,25 @@
...
                         },
+                        ...(element.hasAttribute("data-tomselect-server-search") ? {
+                            shouldLoad: (query) => query.length >= minimumInputLength,
+                            load: (query, callback) => {
+                                var url = element.getAttribute("data-tomselect-server-search") + "?query=" + encodeURIComponent(query);
+                                fetch(url)
+                                    .then(response => response.json())
+                                    .then(json => {
+                                        callback(json.items);
+                                    }).catch(() => {
```
richardebeling, line 175: is "catch all" appropriate here? What kinds of errors, other than network/server unresponsive do we anticipate? In other parts of the site, we signal "server unresponsive" to the user as a blocking popup, we should probably do this consistently

evap/evaluation/forms.py
```diff
@@ -13,6 +14,26 @@
 logger = logging.getLogger(__name__)
 
 
+class ServerSearchSelect(forms.Select):
```
richardebeling, line 17: Nothing here indicates that this is a widget, not a form. Can we make it inherit from `forms.widgets.Select` instead?

(repeated below)

evap/evaluation/forms.py
```diff
@@ -13,6 +14,26 @@
...
+    template_name = "django/forms/widgets/server_select.html"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.attrs["autocomplete"] = "off"
+
```
richardebeling, line 23: Let's just put `autocomplete="off"` in the html, simpler code, simpler life

evap/evaluation/forms.py
```diff
@@ -108,20 +129,20 @@ def label_from_instance(self, obj):
...
         fields = ("title", "first_name_chosen", "first_name_given", "last_name", "email", "delegates")
-        field_classes = {
-            "delegates": UserModelMultipleChoiceField,
-        }
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         for field in ("title", "first_name_given", "last_name", "email"):
             self.fields[field].disabled = True
+        self.fields["delegates"].widget.search_url = reverse("contributor:fetch_delegates_user_profiles")
```
richardebeling, line 145: Why not just pass `search_url=...` to the widget constructor above?

evap/evaluation/views.py
```diff
@@ -247,3 +249,32 @@ def set_startpage(request):
...
+
+
+class TomSelectSearchView(View):
+    def get(self, request, *args, **kwargs):
+        query = request.GET.get("query")
+        if query is None:
+            raise BadRequest("Please provide a search query")
+        items = self.search(query, request, *args, **kwargs)
+        items = [{"id": item.id, "label": self.get_item_label(item)} for item in items]
+        return JsonResponse({"items": items})
```
richardebeling, line 261: This has the abstraction boundary "Given a model instance, return the id and label for the model instance". This requires that we build a full model instance in python for every object first. For many model classes, we may just want to get the two fields, `id` and `name` from the database, and could skip python object construction.

Can we make the abstraction boundary instead "Given a query set over models, return a list of id-label objects", to nudge future PRs towards not building objects unless necessary?

(Given that we _only_ have this for user models, it may also be simpler to just not introduce a base class yet, and make the first version specific for user profiles. I don't see us adding this for other models so soon).

evap/evaluation/views.py
```diff
@@ -247,3 +249,32 @@ def set_startpage(request):
...
+
+    def get_item_label(self, item) -> str:
+        return str(item)
```
richardebeling, line 264: I would not add base class implementations that are unused. If in doubt, remove.

evap/evaluation/views.py
```diff
@@ -247,3 +249,32 @@ def set_startpage(request):
...
+
+    def search(self, query, request, *args, **kwargs) -> QuerySet:
```
richardebeling, line 266: If we use `@abstractmethod`, I think we will get better type checking for this. Can we use that?

evap/evaluation/views.py
```diff
@@ -247,3 +249,32 @@ def set_startpage(request):
...
+class UserProfileSearchView(TomSelectSearchView):
+    def get_item_label(self, item) -> str:
+        return item.full_name_with_additional_info
+
+    def search(self, query, request, *args, **kwargs) -> QuerySet[UserProfile]:
+        return UserProfile.objects.filter(
+            Q(first_name_given__icontains=query)
+            | Q(first_name_chosen__icontains=query)
+            | Q(last_name__icontains=query)
+            | Q(email__icontains=query)
```
richardebeling, line 279: 1. We lose fuzziness of the match here. Before, searching for `om ner` would find Oma Abner. @janno42 I guess we are okay with that?
2. What about database performance here? What does a postgres execution plan look like, full table scan for each query? Would we benefit from having a specific string index? If we want/need that anyway, maybe we can get fuzzier matching with postgres as well (see https://www.postgresql.org/docs/current/pgtrgm.html)

evap/staff/forms.py
```diff
@@ -616,7 +633,10 @@ def __init__(self, *args, evaluation=None, **kwargs):
         super().__init__(*args, **kwargs)
 
         if self.instance.contributor:
-            self.fields["contributor"].queryset |= UserProfile.objects.filter(pk=self.instance.contributor.pk)
+            self.fields["contributor"].queryset = self.get_contributor_queryset() | UserProfile.objects.filter(
+                pk=self.instance.contributor.pk
+            )
+            self.fields["contributor"].widget.search_url = reverse("staff:fetch_contributor_user_profiles")
```
richardebeling, line 639: Does this mean that if you accidentally remove someone from the selection, you'll not be able to add them back?

How does TomSelect handle a mixture of HTML-supplied and server-supplied options? Does merge the server-options with the HTML options on an update, or will the HTML options be overwritten? If the former, maybe we can remove a large part of the specialization logic we have for "keep allowing existing participants" etc.

evap/staff/forms.py
```diff
@@ -1117,12 +1146,21 @@ def save(self, *args, **kw):
 
 
 class UserMergeSelectionForm(forms.Form):
-    main_user = UserModelChoiceField(UserProfile.objects.all())
-    other_user = UserModelChoiceField(UserProfile.objects.all())
+    main_user = UserModelChoiceField(UserProfile.objects.all(), widget=ServerSearchSelect())
+    other_user = UserModelChoiceField(UserProfile.objects.all(), widget=ServerSearchSelect())
```
richardebeling, line 1150: We should be consistent with how we define these fields and their widgets. This PR mixes various approaches (form.meta, explicit instance methods).

If we have various places where we want dynamically fetched user profile selects, and it is **always** user profiles, I think it would make sense to have a separate `UserModuleChoiceField`-ish class for that, that defines the widget internally, and also sets the search url.

evap/staff/views.py
```diff
@@ -2693,3 +2695,27 @@ def enter_staff_mode(request):
 def exit_staff_mode(request):
     staff_mode.exit_staff_mode(request)
     return redirect("evaluation:index")
+
+
+@staff_permission_required
+class ParticipantsUserProfileSearchView(UserProfileSearchView):
+    def search(self, query, request, *args, **kwargs) -> QuerySet[UserProfile]:
+        evaluation = get_object_or_404(Evaluation, id=kwargs["evaluation"]) if kwargs.get("evaluation") else None
```
richardebeling, line 2703: The argument should be `evaluation_id` to be consistent with the rest of the code base. We should unpack it from `kwargs` in the signature (-> `self, query, request, *args, evaluation_id=None, **kwargs`

evap/evaluation/templates/base.html
```diff
@@ -161,8 +161,25 @@
...
                                     return '<div class="no-results">{% translate "No results found" %}</div>';
                                 }
-                            }
+                            },
+                            not_loading: (data, escape) => `<div class="no-results">{% translate "Please enter ${ minimumInputLength } characters or more..." %}</div>`
                         },
+                        ...(element.hasAttribute("data-tomselect-server-search") ? {
+                            shouldLoad: (query) => query.length >= minimumInputLength,
+                            load: (query, callback) => {
+                                var url = element.getAttribute("data-tomselect-server-search") + "?query=" + encodeURIComponent(query);
```
richardebeling, line 170: We should iterate on the names we're introducing here.
* I think "server search" is pretty non-descriptive in general, and would likely be interpreted by someone unfamiliar with the code base as "I entered something in a search bar, now I get a search results page", especially when talking about views.
* The name "data-tomselect-server-search" does not convey that this contains a URL where we can send a request to get mode select options.
* The name `fetch_delegates_user_profiles`, to me, described a view that returns delegates of a user profile, which is not what it does.
* `ParticipantsUserProfileSearchView` the "UserProfile" is redundant since participants are user profiles.
* `UserProfileSearchView` could be mistaken for a full-blown actual view. Maybe add `Base`?

There is already quite a large set of terminology around this, I would try to reuse it and not invent new words. The entries of select inputs are "options", so I could see our views being called `participant_options` or something like that.

Widgets, at least if defined in `forms.py`, should probably have something in their name that makes it clear that they are widgets, not forms.

evap/contributor/forms.py
richardebeling, line 1: We need frontend tests that ensure this works as expected.

richardebeling (commented):
Looks good, thanks. My biggest concern is that we need to find some descriptive terminology around this (and tests), otherwise just nits. Marking as draft until we have tests.

niklasmohrin:
I think this is very cool! I didn't have time to look at it closely yet, but now I will let you settle Richard's comments first :)

evap/evaluation/views.py
hansegucker: Currently, we use some logic from the `UserProfile` model to build the full label (`full_name_with_additional_info`, `full_name`). Fetching only the required columns would require to duplicate this logic.

evap/evaluation/views.py
hansegucker: yes, changed

evap/evaluation/views.py
hansegucker: changed

evap/staff/forms.py
hansegucker: The HTML options don't get overriden.

evap/staff/forms.py
hansegucker: I tried to make this more consistent.

evap/evaluation/forms.py
hansegucker: Because URLs can not be resolved at this time (circular imports).

evap/evaluation/forms.py
hansegucker: Done

evap/evaluation/forms.py
hansegucker: Done

evap/evaluation/templates/base.html
hansegucker: Tried to find some more consistence

evap/evaluation/templates/django/forms/widgets/server_select.html
hansegucker: Comment added

evap/evaluation/views.py
richardebeling: Yes, for UserProfile, we have to build the model instances to be able to call `full_name`.

My point is about _other_ model classes, and the abstraction layer you introduced here. The way the interfaces of `TomSelectSearchView` are defined (`search` returns a `QuerySet`, `get_item_label` takes a model instance), we currently force instance creation, even for model classes that don't have this limitation that UserProfile has. I suggest we a) either don't build the abstraction yet, or b) build it in such a way that it nudges towards better performance in the future.

evap/staff/forms.py
richardebeling: Does this work then?
> maybe we can remove a large part of the specialization logic we have for "keep allowing existing participants" etc.

I would imagine that if we just provide all current participants as in-HTML options, we can simplify the view logic drastically, leading to cacheability etc

evap/evaluation/models.py
```diff
@@ -1821,6 +1811,9 @@ def create_superuser(self, *, email, password=None, first_name_given=None, last_
         user.groups.add(Group.objects.get(name="Manager"))
         return user
 
+    def get_delegates(self):
+        return self.exclude(is_active=False).exclude(is_proxy_user=True)
```
richardebeling: Those are not "delegates" as defined in https://github.com/e-valuation/EvaP/wiki/Terminology -- should we just name it `get_delegate_options`?

evap/evaluation/views.py
```diff
@@ -247,3 +249,22 @@ def set_startpage(request):
     user.save()
 
     return redirect("evaluation:index")
+
+
+class UserProfileOptionsBaseView(View):
```
richardebeling, line 254: What are the access permissions here? Would it suffice to only make this available to staff and editors (or maybe contributors)? Can we limit the query to contain at least three letters?

evap/evaluation/views.py
hansegucker: I've now tried `pgtrgm` and the result looks quite good:

```
 Sort  (cost=51.39..51.40 rows=1 width=254) (actual time=7.833..7.835 rows=6 loops=1)
   Sort Key: (lower(NULLIF((evaluation_userprofile.last_name)::text, ''::text))), (lower(COALESCE(NULLIF((evaluation_userprofile.first_name_chosen)::text, ''::text), NULLIF((evaluation_userprofile.first_name_given)::text, ''::text)))), (lower((evaluation_userprofile.email)::text))
   Sort Method: quicksort  Memory: 26kB
   ->  Nested Loop  (cost=0.28..51.38 rows=1 width=254) (actual time=0.709..7.750 rows=6 loops=1)
         ->  Seq Scan on evaluation_userprofile  (cost=0.00..43.06 rows=1 width=158) (actual time=0.669..7.686 rows=6 loops=1)
               Filter: (((first_name_given)::text % 'cheng'::text) OR ((first_name_chosen)::text % 'cheng'::text) OR ((last_name)::text % 'cheng'::text) OR ((email)::text % 'cheng'::text))
               Rows Removed by Filter: 947
         ->  Index Scan using evaluation_userprofile_pkey on evaluation_userprofile u0  (cost=0.28..8.29 rows=1 width=4) (actual time=0.007..0.007 rows=1 loops=6)
               Index Cond: (id = evaluation_userprofile.id)
               Filter: (is_active AND (NOT is_proxy_user))
 Planning Time: 3.356 ms
 Execution Time: 8.036 ms
```

https://explain.depesz.com/s/m40y

hansegucker: But still, there is a sequential scan, so we maybe could add a index, too.

evap/contributor/forms.py
```diff
@@ -90,6 +92,13 @@ def __init__(self, *args, **kwargs):
             self.fields["participants"].disabled = True
             self.cms_disclaimer = _("Participants are regularly updated with registrations from the CMS.")
 
+    @classmethod
+    def get_participants_queryset(cls, evaluation: Evaluation | None) -> QuerySet[UserProfile]:
+        queryset = UserProfile.objects.exclude(is_active=False)
+        if evaluation is not None and evaluation.pk is not None:
+            queryset = (queryset | evaluation.participants.all()).distinct()
```
richardebeling, line 99: I had hoped we would not need this anymore (by just passing the initial options in HTML) -- why is this still needed here?

evap/evaluation/templates/base.html
```diff
@@ -169,8 +161,25 @@
...
+                            shouldLoad: (query) => query.length >= minimumInputLength,
+                            load: (query, callback) => {
+                                const url = element.getAttribute("data-tomselect-server-side-options-endpoint") + "?query=" + encodeURIComponent(query);
+                                fetch(url)
+                                    .then(response => response.json())
+                                    .then(json => {
+                                        callback(json.options);
+                                    }).catch(() => {
+                                        alert(window.gettext("The server is not responding."));
+                                    });
```
richardebeling: Why are we using a braced body here once? Could we just do this?
```suggestion
                                fetch(url)
                                    .then(response => response.json())
                                    .then(json => callback(json.options))
                                    .catch(() => alert(window.gettext("The server is not responding.")));
```

evap/staff/views.py
```diff
@@ -2722,3 +2724,27 @@ def enter_staff_mode(request):
 def exit_staff_mode(request):
     staff_mode.exit_staff_mode(request)
     return redirect("evaluation:index")
+
+
+@staff_permission_required
+class ParticipantsUserProfileSearchView(UserProfileOptionsBaseView):
```
richardebeling: `OptionsView` instead of `SearchView`? (Here and below)

evap/evaluation/views.py
richardebeling: Feel free to add one if we can declare this easily in django, otherwise, I think 10-ish milliseconds of execution time is good enough

niklasmohrin (commented):
("reviewing" to get this out of the queue, feel free to re-request)

evap/evaluation/views.py
hansegucker: We have permissions with the specific options views (because it's not always limited to contributors/staff). But I limited the query to at least three characters.

hansegucker:
@niklasmohrin @richardebeling Could you please have a look at the failing test (or the failing tests, depending on the shuffled execution order different tests fail)?

Issue #2235, opened 2024-06-30
Student index page / `TestStudentIndexView` performance


richardebeling:
TestStudentIndexView.test_num_queries_is_constant already is pretty slow (2s on my machine) and becomes even slower when increasing the constant 100 evaluations to 1000.

Interrupting the computation, the traceback indicated that the code was stuck in retrieving the `course_id_evaluation_weight_sum_pairs` in `get_evaluations_with_course_result_attributes`. This should be a pretty simple and fast query. Maybe the Course subquery isn't optimized out properly by postgres.

In general, the student index page shouldn't load longer than the results index page, but it currently does (1.4s for me vs 0.6s on results)

niklasmohrin:
It seems like we are actually spending a lot of time in Python code. The Django debug toolbar does not list a lot of time in SQL or caches. Here is an overview from pyinstrument:

![Image](https://github.com/user-attachments/assets/80693de8-592b-4110-90d4-6c6a8a3f6697)

LukasMansour:
So after spending quite a bit of time, I am relatively confident that the main bulk of this issue can be attributed to something happening in the Django templating engine, which can take upwards of 20-30 seconds for 6000 objects on my machine. I do not think that it can be SQL queries in the rendering, not only do we limit the number of queries to be constant (which is the whole point of one of the tests), but also analysing the stack during profiling only shows queries for the user profile during rendering.

Although I also don't agree that a templating engine shouldn't have issues concatenating 6000 strings, it seems to me that Django's templating engine is written pretty badly (in regards to our use-case):

(A) A while ago I tried to rewrite it using Jinja2 and that also didn't seem to improve the performance significantly, however since then I have attained more knowledge, so I would not definitively rule it out as a possible solution, it's possible that I only benchmarked a small number of objects in that rewrite.

Here are some proofs, or at least similar experiences, of this claim:
- https://mentorcruise.com/blog/how-easily-make-existing-django-app-3x-faster/ (Hotspot: Templating Engine, we have exactly such an include within a for loop, in-fact we do a little bit more nesting)
- https://stackoverflow.com/a/26592207 ("Don't use fancy templating features")
- https://medium.com/@officialscoutapm/troubleshooting-inefficient-template-rendering-in-django-29262c35a102 ("Slow rendering when querying large datasets from the database")
[...]

PR #2678, opened 2026-03-16
Slightly improve evaluation query performance in student index view

LukasMansour:
When setting the student index view test to 6000 objects (instead of 100), on average 500ms improvement for the query.

This is work towards #2235, though it is not nearly enough improvement, for a small amount of objects almost negligible.

I did not test it yet to see if it is slower with a small number of elements.

However, the improvement makes sense, as ```prefetch_related``` will do the join in python and ```select_related``` will do the join on the database resulting in a slight improvement, unfortunately it cannot be applied to reverse foreign key relations and the many to many relations.

### DIFF
--- a/evap/student/views.py
+++ b/evap/student/views.py
@@ -9,7 +9,7 @@ from django.conf import settings
 from django.contrib import messages
 from django.core.exceptions import PermissionDenied, SuspiciousOperation
 from django.db import transaction
-from django.db.models import Exists, F, Max, OuterRef, Q, Sum
+from django.db.models import Exists, F, Max, OuterRef, Sum
 from django.http import HttpRequest, HttpResponse
 from django.shortcuts import get_object_or_404, render
 from django.urls import reverse
@@ -125,16 +125,26 @@ class GlobalEvaluationProgress:
 def index(request):
     query = (
         Evaluation.objects.annotate(
-            participates_in=Exists(Evaluation.objects.filter(id=OuterRef("id"), participants=request.user))
+            participates_in=Exists(
+                Evaluation.participants.through.objects.filter(
+                    evaluation_id=OuterRef("pk"), userprofile_id=request.user.id
+                )
+            )
         )
-        .annotate(voted_for=Exists(Evaluation.objects.filter(id=OuterRef("id"), voters=request.user)))
-        .filter(~Q(state=Evaluation.State.NEW), course__evaluations__participants=request.user)
+        .annotate(
+            voted_for=Exists(
+                Evaluation.voters.through.objects.filter(evaluation_id=OuterRef("pk"), userprofile_id=request.user.id)
+            )
+        )
+        .filter(course__evaluations__participants=request.user)
         .exclude(state=Evaluation.State.NEW)
-        .prefetch_related(
+        .select_related(
             "course",
             "course__semester",
-            "course__grade_documents",
             "course__type",
+        )
+        .prefetch_related(
+            "course__grade_documents",
             "course__evaluations",
             "course__responsibles",
             "course__programs",


niklasmohrin:
The prefetch -> select makes sense to me

Did you observe any difference between the join with Evaluation vs Evaluation.participants.through? If not, I would prefer the old version for simplicity

In general, can you post the benchmark code here?

evap/student/views.py
```diff
@@ -125,16 +125,26 @@ def from_settings() -> "GlobalEvaluationProgress | None":
 def index(request):
     query = (
         Evaluation.objects.annotate(
-            participates_in=Exists(Evaluation.objects.filter(id=OuterRef("id"), participants=request.user))
+            participates_in=Exists(
+                Evaluation.participants.through.objects.filter(
+                    evaluation_id=OuterRef("pk"), userprofile_id=request.user.id
+                )
+            )
+        )
```
richardebeling, line 133: This simplifies a very fast subquery by a tiny bit (direct scan vs join), which saves a tiny bit of time: `.explain(analyze=True)` says 1ms improvement, for a total of 44ms for the query, with 5000 extra participations added to the test data for the evap test user.

The "main" query here runs in 20-30ms (measured by `.explain(analyze=True)`). Evaluating the list comprehension  `[evaluation for evaluation in query if evaluation.can_be_seen_by(request.user)]` (which includes the additional prefetch queries and building all the python objects) below takes ~550ms on my notebook. The full page load takes ~37000ms. There must be things we can do that have more impact here for #2235

Approving here because checking the existence of an "participation" is what we want to do here. In some way, the code is more canonical, even if this requires knowledge of django through models now.

The other changes below look good to me and make the code better, I think.

richardebeling (approved):

LukasMansour:
Here's the profiling code, containing the old code commented and the new code:

replaced in student/views.py#index
```
    from pyinstrument import Profiler
    profiler = Profiler()
    profiler.start()
    query = (
        Evaluation.objects.annotate(
            participates_in=Exists(
                Evaluation.participants.through.objects.filter(
                    evaluation_id=OuterRef("pk"), userprofile_id=request.user.id
                )
            )
        )
        .annotate(
            voted_for=Exists(
                Evaluation.voters.through.objects.filter(evaluation_id=OuterRef("pk"), userprofile_id=request.user.id)
            )
        )
        .filter(course__evaluations__participants=request.user)
        .exclude(state=Evaluation.State.NEW)
        .select_related(
            "course",
            "course__semester",
            "course__type",
        )
        .prefetch_related(
            "course__grade_documents",
            "course__evaluations",
            "course__responsibles",
            "course__programs",
        )
        .distinct()
    )
    #query = (
    #        Evaluation.objects.annotate(
    #            participates_in=Exists(Evaluation.objects.filter(id=OuterRef("id"), participants=request.user))
    #        )
    #        .annotate(voted_for=Exists(Evaluation.objects.filter(id=OuterRef("id"), voters=request.user)))
    #        .filter(~Q(state=Evaluation.State.NEW), course__evaluations__participants=request.user)
    #        .exclude(state=Evaluation.State.NEW)
    #        .prefetch_related(
    #            "course",
    #            "course__semester",
    #            "course__grade_documents",
    #            "course__type",
    #            "course__evaluations",
    #            "course__responsibles",
    #            "course__programs",
    #        )
    #        .distinct()
    #    )
    query = Evaluation.annotate_with_participant_and_voter_counts(query)
    evaluations = [evaluation for evaluation in query if evaluation.can_be_seen_by(request.user)]

    profiler.stop()
    profiler.write_html("./profile_new.html")
    # profiler.write_html("./profile_old.html")
```

Running this a few times, you should (hopefully) see a speedup on average, it may be better to increase to 8000 objects, which on first glance seems to be a more extreme difference on my machine at least (used for the images below).

An example flamegraph below (however, in the original PR I averaged this over 20 runs old and new respectively).

I want to note explicitly that its really difficult to know if this really speeds up, or might even slow us done for a more reasonable/realistic number of objects, I guess we are talking about asymptotics here to some degree.

OLD:
image

NEW:
image

niklasmohrin (approved):
Change is good so I will merge, but feel free to keep discussing here. Thanks!

PR #2682, opened 2026-03-21
General contribution can be created lazily

### LINKED ISSUE #1847: Invariant for an evaluation's general contribution
In #1839, it seemed unclear what we want our invariant regarding the general contribution to be.

Thoughts / observations so far:
* If we require a general contribution to exist for all evaluations stored in the database, this should be a db constraint (-> #1800)
* Some views expect general contributions to exist. As an example in #1839, we had the "assign questionnaires" manager function.
* Lazy creation
    * Could work with a `get-or-create` helper function.
    * Some code uses `contributions.get(contributor=None)` to access general contributions. This would likely break. `contributor=None` occurs 17 times in the codebase.
* Eager creation on database inserts
    * This is what the current `save()` method on `Evaluation` attempts to do, so I'd guess we have quite some code relying on this.
        * The cause for this discussion is that the `save()` override does not work when `bulk_` operations are used.
    * #1843 didn't add a test (because responsibilities were unclear). If we enforce eager creation, we'd probably want to add a regression test.
    * Since we can have model instances that were not yet saved to database, basically all code that can not assure that a evaluation instance is stored in the db has to assume the general contribution doesn't exist. Likely, we currently do not do this, and most code just works because we mostly work with instances stored to the database.
        * As a result, e.g. the `general_contribution` property checks for existance and returns `None` if the general contribution doesn't exist. Note that it currently handles both cases, "Evaluation not yet in DB" and "Contribution doesn't exist", explicitly.
        * Since we have to handle the "does not exist" case anyway, the lazy code may not be more complex?
        * We could add "exists in DB" as a pre-condition for the `general_contribution` property.
            * This wouldn't "solve" cases where we always "accidentally" have instances that are stored in the DB, an assertion would always succeed, but we'd still have to add checks. I'd estimate this is the case for most places that should check, but don't check (otherwise, they would likely already fail due to `None` being returned)
    * `test_has_enough_questionnaires` uses `bulk_create` to explicitly circumvent creating the general contribution, and then asserts that some methods still work in some expected way. This would be invalid. We may just want to test with a not-yet-stored-to-db instance?

We considered allowing not to have a general contribution in https://github.com/e-valuation/EvaP/issues/1246, it seems there was no good fundamental reason to enfore existence of the general contribution, it just happened to be what we first implemented.

Related (and funny) PR / issue combination: https://github.com/e-valuation/EvaP/pull/1259 and https://github.com/e-valuation/EvaP/issues/1260.

In https://github.com/e-valuation/EvaP/issues/1839#issuecomment-1368086138, @niklasmohrin voted pro eager creation.

richardebeling:
Relates to #1847, #1839, #1846

This defensively still creates a general contribution on `.save()`. If we remove that, we get lots of test failures, but it's mostly tests that expect that it was created. I would make this conditional on `not DEBUG`, so that we don't accidentally break production but have some time to check for problems in dev environments.

The big gain we get from this is that `general_contribution` is now typed to be able to return `None`, which should, in typed code, catch problems when we try to access it in an unsafe way.

* [x] One TODO in code
* [x] See what the change set would be to make the tests pass when not calling `ensure_general_contribution` in `Evaluation.save` anymore
    * ~70 tests require us to explicitly create a general contribution -- I'll do this separately

### DIFF
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -100,9 +100,9 @@ class EvaluationForm(forms.ModelForm):
             self.add_error("vote_end_date", _("The first day of evaluation must be before the last one."))
 
         # Ensure locked questionnaires cannot be deselected and only unlocked ones can be added
-        selected_questionnaires = (
-            self.instance.ensure_general_contribution().questionnaires.filter(is_locked=True).distinct()
-        )
+        selected_questionnaires = Questionnaire.objects.filter(
+            is_locked=True, contributions__contributor=None, contributions__evaluation=self.instance
+        ).distinct()
         selected_questionnaires |= self.cleaned_data.get("general_questionnaires").filter(is_locked=False)
         selected_questionnaires |= self.cleaned_data.get("dropout_questionnaires").filter(is_locked=False)
 
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -100,7 +100,9 @@ class EvaluationForm(forms.ModelForm):
             self.add_error("vote_end_date", _("The first day of evaluation must be before the last one."))
 
         # Ensure locked questionnaires cannot be deselected and only unlocked ones can be added
-        selected_questionnaires = self.instance.general_contribution.questionnaires.filter(is_locked=True).distinct()
+        selected_questionnaires = (
+            self.instance.ensure_general_contribution().questionnaires.filter(is_locked=True).distinct()
+        )
         selected_questionnaires |= self.cleaned_data.get("general_questionnaires").filter(is_locked=False)
         selected_questionnaires |= self.cleaned_data.get("dropout_questionnaires").filter(is_locked=False)
 
@@ -133,7 +135,7 @@ class EvaluationForm(forms.ModelForm):
         selected_questionnaires = self.cleaned_data.get("general_questionnaires") | self.cleaned_data.get(
             "dropout_questionnaires"
         )
-        evaluation.general_contribution.questionnaires.set(selected_questionnaires)
+        evaluation.ensure_general_contribution().questionnaires.set(selected_questionnaires)
         return evaluation
 
 
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -919,7 +919,7 @@ class Evaluation(LoggedModel):
         if "general_contribution" in self.__dict__ and self.general_contribution is not None:
             return self.general_contribution
 
-        contribution, created = self.contributions.get_or_create(contributor=None)
+        contribution, __ = self.contributions.get_or_create(contributor=None)
         self.__dict__["general_contribution"] = contribution
         return assert_not_none(self.general_contribution)
 
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -50,7 +50,7 @@ from evap.evaluation.tools import (
     translate,
     vote_end_datetime,
 )
-from evap.tools import date_to_datetime
+from evap.tools import assert_not_none, date_to_datetime
 
 logger = logging.getLogger(__name__)
 
@@ -567,7 +567,7 @@ class Evaluation(LoggedModel):
         exam_evaluation.participants.set(self.participants.all())
         for contribution in self.contributions.exclude(contributor=None):
             exam_evaluation.contributions.create(contributor=contribution.contributor)
-        exam_evaluation.general_contribution.questionnaires.set(settings.EXAM_QUESTIONNAIRE_IDS)
+        exam_evaluation.ensure_general_contribution().questionnaires.set(settings.EXAM_QUESTIONNAIRE_IDS)
 
     class TextAnswerReviewState(Enum):
         do_not_call_in_templates = enum.nonmember(True)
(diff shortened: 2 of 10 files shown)


evap/student/views.py
```diff
@@ -282,7 +282,12 @@ def render_vote_page(
 
     assert preview or not all(form.is_valid() for form_group in form_groups.values() for form in form_group)
 
-    evaluation_form_group = form_groups.pop(evaluation.general_contribution, default=[])
+    # TODO: Can we do this more elegantly? dict.pop requires that first input type is key type (=Contribution), but
+    # passing any other type (especially None) is fine and falls back to the default.
+    if evaluation.general_contribution is None:
+        evaluation_form_group = []
+    else:
+        evaluation_form_group = form_groups.pop(evaluation.general_contribution, default=[])
```
richardebeling: @niklasmohrin any clever ideas for this? I couldn't come up with anything

niklasmohrin: not really, I think that with typing in mind, the current code is the most sensible

evap/contributor/forms.py
```diff
@@ -100,7 +100,9 @@ def clean(self):
             self.add_error("vote_end_date", _("The first day of evaluation must be before the last one."))
 
         # Ensure locked questionnaires cannot be deselected and only unlocked ones can be added
-        selected_questionnaires = self.instance.general_contribution.questionnaires.filter(is_locked=True).distinct()
+        selected_questionnaires = (
+            self.instance.ensure_general_contribution().questionnaires.filter(is_locked=True).distinct()
```
niklasmohrin, line 104: hm, I feel we shouldn't create a contribution in `clean` here; can't we use an empty queryset if no general contribution exists?

evap/evaluation/models.py
```diff
@@ -913,8 +913,18 @@ def unpublish(self):
...
 
+    def ensure_general_contribution(self) -> "Contribution":
+        assert self.pk is not None
+
+        if "general_contribution" in self.__dict__ and self.general_contribution is not None:
+            return self.general_contribution
+
+        contribution, __ = self.contributions.get_or_create(contributor=None)
+        self.__dict__["general_contribution"] = contribution
+        return assert_not_none(self.general_contribution)
```
niklasmohrin, line 924: Is there a reason not to do
```suggestion
        return contribution
```
?

evap/staff/forms.py
```diff
@@ -533,11 +533,11 @@ def save(self, *args, **kw):
         selected_questionnaires = self.cleaned_data.get("general_questionnaires") | self.cleaned_data.get(
             "dropout_questionnaires"
         )
-        removed_questionnaires = set(self.instance.general_contribution.questionnaires.all()) - set(
+        removed_questionnaires = set(self.instance.ensure_general_contribution().questionnaires.all()) - set(
```
niklasmohrin, line 536: can we do `general_contribution = evaluation.ensure_general_contribution()` just once here?

also, shouldn't self.instance and evaluation be the same?

evap/evaluation/models.py
richardebeling: Yes,
> evap/evaluation/models.py:924: error: Incompatible return value type (got "Contribution | None", expected "Contribution")  [return-value]

niklasmohrin: Huh, not on my computer :D I didn't check this when writing the comment initially, but I would expect the return of `get_or_create` to be non-None, and now that I checked this is also what I get when using `reveal_type(contribution)` right after get_or_create

richardebeling: Ah, I tried `return self.general_contribution`, not `return contribution`. Yeah, that should work.

evap/staff/forms.py
```diff
@@ -534,26 +530,29 @@ def clean(self):
...
-        evaluation.general_contribution.remove_answers_to_questionnaires(removed_questionnaires)
-        evaluation.general_contribution.questionnaires.set(selected_questionnaires)
-        if hasattr(self.instance, "old_course"):
-            if self.instance.old_course != evaluation.course:
-                update_template_cache_of_published_evaluations_in_course(self.instance.old_course)
+        removed_questionnaires = set(general_contribution.questionnaires.all()) - set(selected_questionnaires)
+        general_contribution.remove_answers_to_questionnaires(removed_questionnaires)
+        general_contribution.questionnaires.set(selected_questionnaires)
+
+        if hasattr(evaluation, "old_course"):
```
niklasmohrin, line 542: What do we think are the chances that (in a future Django release) a new Python object is created in super().save() instead and we lose our old_course attribute? Isn't this whole `self.instance.old_course = ...` business super whack anyways? Can we just put `old_course` on the form itself?

niklasmohrin (approved):
Very nice improvement, the type hint is a win. It won't help in Django templates, but I guess there every access through an un-created `evaluation.general_contribution.frobnicate` would just cascade to None or empty string or whatever nully value Django comes up with then. So it's less risky then untyped Python code, which we have less and less of :+1:

evap/staff/forms.py
richardebeling: I think this is rather unlikely: Creating a new object requires CPU time, possibly a database request. Without a good reason, I don't think django would do that, and if they do:
* I'd expect it to be listed as a breaking change in the release notes
* I expect one of our tests to fail afterwards (we specifically have `test_course_change_updates_cache` for this)

PR #2687, opened 2026-03-30
Moar types

richardebeling:
because moar = better

I think we can conclude by now that python type hints are not going anywhere, so we may just as well annotate our interfaces to benefit from better LSP understanding and type checking.

Split into two commits, the first one should be very quick to review, containing only type changes. The second one contains everything that needs to touch actual production logic, mostly by adding assertions.

### DIFF
--- a/evap/evaluation/auth.py
+++ b/evap/evaluation/auth.py
@@ -1,11 +1,13 @@
 import inspect
 from collections.abc import Callable, Iterable
 from functools import wraps
+from typing import Any
 
 from django.conf import settings
 from django.contrib.auth.backends import ModelBackend
 from django.core.exceptions import PermissionDenied
 from django.core.mail import EmailMessage
+from django.http import HttpRequest
 from django.utils.decorators import method_decorator
 from mozilla_django_oidc.auth import OIDCAuthenticationBackend
 
@@ -25,7 +27,7 @@ class RequestAuthUserBackend(ModelBackend):
 
     # Having a different method signature is okay according to django documentation:
     # https://docs.djangoproject.com/en/3.0/topics/auth/customizing/#writing-an-authentication-backend
-    def authenticate(self, request, key):  # pylint: disable=arguments-differ
+    def authenticate(self, request: HttpRequest, key: int) -> UserProfile | None:  # type: ignore[override] # pylint: disable=arguments-differ
         if not key:
             return None
 
@@ -37,7 +39,9 @@ class RequestAuthUserBackend(ModelBackend):
 
 class EmailAuthenticationBackend(ModelBackend):
     # https://docs.djangoproject.com/en/3.1/topics/auth/customizing/#writing-an-authentication-backend
-    def authenticate(self, request, email=None, password=None):  # pylint: disable=arguments-differ,arguments-renamed
+    def authenticate(  # type: ignore[override] # pylint: disable=arguments-differ,arguments-renamed
+        self, request: HttpRequest | None, email: str | None = None, password: str | None = None
+    ) -> UserProfile | None:
         assert password_login_is_active()
         try:
             user = UserProfile.objects.get(email=email)
@@ -79,57 +83,57 @@ def class_or_function_check_decorator(test_func: Callable[[UserProfile], bool]):
 
 
 @class_or_function_check_decorator
-def internal_required(user):
+def internal_required(user: UserProfile) -> bool:
     return not user.is_external
 
 
 @class_or_function_check_decorator
-def staff_permission_required(user):
+def staff_permission_required(user: UserProfile) -> bool:
     return user.has_staff_permission
 
 
 @class_or_function_check_decorator
-def manager_required(user):
+def manager_required(user: UserProfile) -> bool:
     return user.is_manager
 
 
 @class_or_function_check_decorator
-def reviewer_required(user):
+def reviewer_required(user: UserProfile) -> bool:
     return user.is_reviewer
 
 
 @class_or_function_check_decorator
-def grade_publisher_required(user):
+def grade_publisher_required(user: UserProfile) -> bool:
     return user.is_grade_publisher
 
 
 @class_or_function_check_decorator
-def grade_publisher_or_manager_required(user):
+def grade_publisher_or_manager_required(user: UserProfile) -> bool:
     return user.is_grade_publisher or user.is_manager
 
 
 @class_or_function_check_decorator
-def grade_downloader_required(user):
+def grade_downloader_required(user: UserProfile) -> bool:
     return user.can_download_grades
 
 
 @class_or_function_check_decorator
-def responsible_or_contributor_or_delegate_required(user):
+def responsible_or_contributor_or_delegate_required(user: UserProfile) -> bool:
     return user.is_responsible_or_contributor_or_delegate
 
 
 @class_or_function_check_decorator
-def editor_or_delegate_required(user):
+def editor_or_delegate_required(user: UserProfile) -> bool:
     return user.is_editor_or_delegate
 
 
 @class_or_function_check_decorator
-def participant_required(user):
+def participant_required(user: UserProfile) -> bool:
     return user.is_participant
 
 
 @class_or_function_check_decorator
-def reward_user_required(user):
+def reward_user_required(user: UserProfile) -> bool:
     return can_reward_points_be_used_by(user)
 
 
@@ -142,13 +146,13 @@ class OpenIDAuthenticationBackend(OIDCAuthenticationBackend):
         if previous_domain := settings.OIDC_EMAIL_TRANSITIONS.get(domain):
             yield f"{name}@{previous_domain}"
 
-    def get_userinfo(self, *args, **kwargs):
+    def get_userinfo(self, *args, **kwargs) -> Any:
         claims = super().get_userinfo(*args, **kwargs)
         if "email" in claims:
             claims["email"] = clean_email(claims["email"])
         return claims
 
-    def filter_users_by_claims(self, claims):
+    def filter_users_by_claims(self, claims: dict[str, Any]) -> list[UserProfile]:
         assert openid_login_is_active()
         email = claims.get("email")
         if not email:
@@ -162,7 +166,7 @@ class OpenIDAuthenticationBackend(OIDCAuthenticationBackend):
 
         return []
 
-    def create_user(self, claims):
+    def create_user(self, claims: dict[str, Any]) -> UserProfile:
         assert openid_login_is_active()
         return self.UserModel.objects.create(
             email=claims.get("email"),
@@ -170,7 +174,7 @@ class OpenIDAuthenticationBackend(OIDCAuthenticationBackend):
             last_name=claims.get("family_name", ""),
         )
 
-    def update_user(self, user, claims):
+    def update_user(self, user: UserProfile, claims: dict[str, Any]) -> UserProfile:
         assert openid_login_is_active()
         if not user.first_name_given:
             user.first_name_given = claims.get("given_name", "")
--- a/evap/evaluation/auth.py
+++ b/evap/evaluation/auth.py
@@ -147,7 +147,7 @@ class OpenIDAuthenticationBackend(OIDCAuthenticationBackend):
         if previous_domain := settings.OIDC_EMAIL_TRANSITIONS.get(domain):
             yield f"{name}@{previous_domain}"
 
-    def get_userinfo(self, *args, **kwargs) -> Any:
+    def get_userinfo(self, *args, **kwargs) -> dict[str, Any]:
         claims = super().get_userinfo(*args, **kwargs)
         if "email" in claims:
             claims["email"] = clean_email(claims["email"])
--- a/evap/evaluation/auth.py
+++ b/evap/evaluation/auth.py
@@ -64,7 +64,8 @@ def class_or_function_check_decorator(test_func: Callable[[UserProfile], bool]):
 
     def function_decorator(func):
         @wraps(func)
-        def wrapped(request, *args, **kwargs):
+        def wrapped(request: HttpRequest, *args, **kwargs) -> Any:
+            assert isinstance(request.user, UserProfile)
             if not test_func(request.user):
                 raise PermissionDenied
             return func(request, *args, **kwargs)
--- a/evap/evaluation/forms.py
+++ b/evap/evaluation/forms.py
@@ -33,8 +33,8 @@ class LoginEmailForm(forms.Form):
         super().__init__(*args, **kwargs)
 
     @sensitive_variables("password")
-    def clean_password(self):
-        email = self.cleaned_data.get("email")
+    def clean_password(self) -> str | None:
+        email = assert_not_none(self.cleaned_data.get("email"))
         password = self.cleaned_data.get("password")
 
         email = email.lower()
@@ -69,8 +69,8 @@ class NewKeyForm(forms.Form):
 
         super().__init__(*args, **kwargs)
 
-    def clean_email(self):
-        email = self.cleaned_data.get("email")
+    def clean_email(self) -> str:
+        email: str = assert_not_none(self.cleaned_data.get("email"))
 
         if not UserProfile.email_needs_login_key(email):
             raise forms.ValidationError(
--- a/evap/evaluation/forms.py
+++ b/evap/evaluation/forms.py
@@ -4,11 +4,13 @@ from django import forms
 from django.conf import settings
 from django.contrib.auth import authenticate
 from django.core.exceptions import ValidationError
+from django.http import HttpRequest
 from django.utils.translation import gettext_lazy as _
 from django.views.decorators.debug import sensitive_variables
 
 from evap.evaluation.models import Evaluation, UserProfile
 from evap.results.tools import STATES_WITH_RESULTS_CACHING, cache_results
+from evap.tools import assert_not_none
 
 logger = logging.getLogger(__name__)
 
@@ -19,7 +21,7 @@ class LoginEmailForm(forms.Form):
     email = forms.CharField(label=_("Email"), max_length=254, widget=forms.EmailInput(attrs={"autofocus": True}))
     password = forms.CharField(label=_("Password"), widget=forms.PasswordInput)
 
-    def __init__(self, request, *args, **kwargs):
+    def __init__(self, request: HttpRequest, *args, **kwargs) -> None:
         """
         If request is passed in, the form will validate that cookies are
         enabled. Note that the request (a HttpRequest object) must have set a
@@ -27,7 +29,7 @@ class LoginEmailForm(forms.Form):
         running this validation.
         """
         self.request = request
-        self.user_cache = None
+        self.user_cache: UserProfile | None = None
         super().__init__(*args, **kwargs)
 
     @sensitive_variables("password")
@@ -44,26 +46,26 @@ class LoginEmailForm(forms.Form):
         self.check_for_test_cookie()
         return password
 
-    def check_for_test_cookie(self):
+    def check_for_test_cookie(self) -> None:
         if self.request and not self.request.session.test_cookie_worked():
             raise forms.ValidationError(
                 _("Your Web browser doesn't appear to have cookies enabled. Cookies are required for logging in.")
             )
 
-    def get_user_id(self):
+    def get_user_id(self) -> int | None:
         if self.user_cache:
             return self.user_cache.id
         return None
 
-    def get_user(self):
+    def get_user(self) -> UserProfile | None:
         return self.user_cache
 
 
 class NewKeyForm(forms.Form):
     email = forms.EmailField(label=_("Email address"))
 
-    def __init__(self, *args, **kwargs):
-        self.user_cache = None
+    def __init__(self, *args, **kwargs) -> None:
+        self.user_cache: UserProfile | None = None
 
         super().__init__(*args, **kwargs)
 
@@ -90,19 +92,19 @@ class NewKeyForm(forms.Form):
 
         return email
 
-    def get_user(self):
+    def get_user(self) -> UserProfile | None:
         return self.user_cache
 
 
 class UserModelChoiceField(forms.ModelChoiceField):
-    def label_from_instance(self, obj):
+    def label_from_instance(self, obj: UserProfile) -> str:
         return obj.full_name_with_additional_info
 
 
 class UserModelMultipleChoiceField(forms.ModelMultipleChoiceField):
     widget = forms.SelectMultiple(attrs={"data-tomselect-fullwidth": ""})
 
-    def label_from_instance(self, obj):
+    def label_from_instance(self, obj: UserProfile) -> str:
         return obj.full_name_with_additional_info
 
 
@@ -118,12 +120,12 @@ class ProfileForm(forms.ModelForm):
             "delegates": UserModelMultipleChoiceField,
         }
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
         for field in ("title", "first_name_given", "last_name", "email"):
             self.fields[field].disabled = True
 
-    def save(self, *args, **kwargs):
+    def save(self, *args, **kwargs) -> None:
         super().save(*args, **kwargs)
 
         if "first_name_chosen" in self.changed_data:
@@ -138,7 +140,7 @@ class ProfileForm(forms.ModelForm):
 
         logger.info('User "%s" edited the settings.', self.instance.email)
 
-    def clean_first_name_chosen(self):
+    def clean_first_name_chosen(self) -> str:
         name = self.cleaned_data["first_name_chosen"]
 
         for character in name:
@@ -149,7 +151,7 @@ class ProfileForm(forms.ModelForm):
 
 
 class NotebookForm(forms.ModelForm):
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
         self.fields["notes"].widget.attrs.update({"class": "notebook-textarea"})
 
--- a/evap/evaluation/forms.py
+++ b/evap/evaluation/forms.py
@@ -10,6 +10,7 @@ from django.views.decorators.debug import sensitive_variables
 
 from evap.evaluation.models import Evaluation, UserProfile
 from evap.results.tools import STATES_WITH_RESULTS_CACHING, cache_results
+from evap.tools import assert_not_none
 
 logger = logging.getLogger(__name__)
 
--- a/evap/evaluation/forms.py
+++ b/evap/evaluation/forms.py
@@ -10,7 +10,6 @@ from django.views.decorators.debug import sensitive_variables
 
 from evap.evaluation.models import Evaluation, UserProfile
 from evap.results.tools import STATES_WITH_RESULTS_CACHING, cache_results
-from evap.tools import assert_not_none
 
 logger = logging.getLogger(__name__)
 
--- a/evap/evaluation/models.py
+++ b/evap/evaluation/models.py
@@ -8,8 +8,7 @@ from dataclasses import dataclass
 from datetime import date, datetime, time, timedelta
 from enum import Enum, auto
 from functools import partial
-from numbers import Real
-from typing import Any
+from typing import TYPE_CHECKING, Any, cast
 
 from django.conf import settings
 from django.contrib import messages
@@ -50,7 +49,12 @@ from evap.evaluation.tools import (
     translate,
     vote_end_datetime,
 )
-from evap.tools import date_to_datetime
+from evap.tools import assert_not_none, date_to_datetime
+
+if TYPE_CHECKING:
+    import builtins
+
+    from evap.grades.models import GradeDocument
 
 logger = logging.getLogger(__name__)
 
@@ -99,11 +103,11 @@ class Semester(models.Model):
         verbose_name = _("semester")
         verbose_name_plural = _("semesters")
 
-    def __str__(self):
+    def __str__(self) -> str:
         return self.name
 
     @property
-    def can_be_deleted_by_manager(self):
+    def can_be_deleted_by_manager(self) -> bool:
         if self.is_active:
             return False
 
@@ -113,21 +117,21 @@ class Semester(models.Model):
         return self.participations_are_archived and self.grade_documents_are_deleted and self.results_are_archived
 
     @property
-    def participations_can_be_archived(self):
+    def participations_can_be_archived(self) -> bool:
         return not self.participations_are_archived and all(
             evaluation.participations_can_be_archived for evaluation in self.evaluations.all()
         )
 
     @property
-    def grade_documents_can_be_deleted(self):
+    def grade_documents_can_be_deleted(self) -> bool:
         return not self.grade_documents_are_deleted
 
     @property
-    def results_can_be_archived(self):
+    def results_can_be_archived(self) -> bool:
         return not self.results_are_archived
 
     @transaction.atomic
-    def archive(self):
+    def archive(self) -> None:
         if not self.participations_can_be_archived:
             raise NotArchivableError
         for evaluation in self.evaluations.all():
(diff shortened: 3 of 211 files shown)


evap/evaluation/models.py
```diff
@@ -384,6 +388,13 @@ class Course(LoggedModel):
         verbose_name=_("campus management system id"), blank=True, null=True, unique=True, max_length=255
     )
 
+    if TYPE_CHECKING:
+        not_all_evaluations_are_published: bool | None = None
+        distribution: tuple[float, ...] | None = None
+        evaluation_count: int | None = None
+        evaluation_weight_sum: int | None = None
+        avg_grade: float | None = None
```
niklasmohrin, line 396: But this is only the case if we have added these attributes right? I don't think we should have this, we might access one of these somewhere and get an attribute error

evap/evaluation/models_logging.py
```diff
@@ -58,28 +59,29 @@ class LogJSONEncoder(JSONEncoder):
...
+def _choice_to_display(field: models.Field, choice: Any) -> Any:
+    assert isinstance(field.choices, Sequence)  # does not support nested choices
     for key, label in field.choices:
         if key == choice:
             return label
     return choice
 
 
-def _field_actions_for_field(field, actions):
+def _field_actions_for_field(field: models.Field | Any, actions: dict[str, Any]) -> Iterable[FieldAction]:
```
niklasmohrin, line 77: Is there a benefit over this?
```suggestion
def _field_actions_for_field(field: Any, actions: dict[str, Any]) -> Iterable[FieldAction]:
```

evap/evaluation/models_logging.py
```diff
@@ -109,16 +111,17 @@ class Meta:
         ordering = ["-datetime", "-id"]
 
     @property
-    def field_context_data(self):
+    def field_context_data(self) -> dict[str, Any]:
```
niklasmohrin, line 114:
```suggestion
    def field_context_data(self) -> dict[str, list[FieldAction]]:
```

evap/evaluation/auth.py
```diff
@@ -142,13 +147,13 @@ def possible_existing_account_emails(email: str) -> Iterable[str]:
         if previous_domain := settings.OIDC_EMAIL_TRANSITIONS.get(domain):
             yield f"{name}@{previous_domain}"
 
-    def get_userinfo(self, *args, **kwargs):
+    def get_userinfo(self, *args, **kwargs) -> Any:
```
niklasmohrin: I think this is true?
```suggestion
    def get_userinfo(self, *args, **kwargs) -> dict[str, Any]:
```

evap/evaluation/forms.py
```diff
@@ -19,20 +21,20 @@ class LoginEmailForm(forms.Form):
...
         self.request = request
-        self.user_cache = None
+        self.user_cache: UserProfile | None = None
         super().__init__(*args, **kwargs)
 
     @sensitive_variables("password")
-    def clean_password(self):
-        email = self.cleaned_data.get("email")
+    def clean_password(self) -> str | None:
+        email = assert_not_none(self.cleaned_data.get("email"))
```
niklasmohrin:
```suggestion
        email = self.cleaned_data["email"]
```

evap/evaluation/forms.py
```diff
@@ -19,20 +21,20 @@ class LoginEmailForm(forms.Form):
...
         """
         self.request = request
-        self.user_cache = None
+        self.user_cache: UserProfile | None = None
         super().__init__(*args, **kwargs)
 
     @sensitive_variables("password")
-    def clean_password(self):
-        email = self.cleaned_data.get("email")
+    def clean_password(self) -> str | None:
```
niklasmohrin, line 36: The `password` variable has type Any inside, I think both email and password could get a `: str` annotation to pull them out of Any

evap/evaluation/tools.py
```diff
@@ -111,7 +113,7 @@ def get_object_from_dict_pk_entry_or_logged_40x[M: Model](
         raise SuspiciousOperation from e
 
 
-def is_prefetched(instance, attribute_name: str) -> bool:
+def is_prefetched(instance: Any, attribute_name: str) -> bool:
```
niklasmohrin: Any reason not do this?
```suggestion
def is_prefetched(instance: Model, attribute_name: str) -> bool:
```

evap/evaluation/tools.py
```diff
@@ -357,7 +361,7 @@ def write_empty_row_with_styles(self, styles: Iterable[str]) -> None:
     def export_impl(self, *args, **kwargs) -> None:
         """Specify the logic to insert the data into the sheet here."""
 
-    def export(self, response, *args, **kwargs) -> None:
+    def export(self, response: HttpResponse | typing.BinaryIO, *args, **kwargs) -> None:
```
niklasmohrin, line 364: Hm, fair, we really want this to be https://docs.python.org/3/library/io.html#io.Writer but it is new in Python 3.14 :(

evap/evaluation/views.py
```diff
@@ -116,7 +116,7 @@ def index(request):
 
 
 @no_login_required
-def login_key_authentication(request, key):
+def login_key_authentication(request: HttpRequest, key: str) -> HttpResponse:
```
niklasmohrin: This should be 
```suggestion
def login_key_authentication(request: HttpRequest, key: int) -> HttpResponse:
```
no? A bit of a bad sign that typing didn't catch this, I would have expected typeguard to go off here

niklasmohrin (commented):
some more files checked off; should we maybe split this into multiple PRs so that reviewed changes can get merged before they get conflicts?

evap/evaluation/models.py
richardebeling: Removed again, see https://github.com/e-valuation/EvaP/pull/2687/commits/764ab189dc8b7ffb30fdc0b5adc4bf529e29c15f for the fallout -- not that much, I think it's acceptable for now

tools/check_dist.py
```diff
@@ -3,13 +3,15 @@
...
+from collections.abc import Iterable
 from pathlib import Path
+from typing import Any
 from zipfile import ZipFile
 
 from pathspec.patterns.gitwildmatch import GitWildMatchPattern
 
 
-def ensure_all_artifacts_included(pyproject, wheel_paths):
+def ensure_all_artifacts_included(pyproject: Any, wheel_paths: Iterable[str]) -> int:
```
niklasmohrin:
```suggestion
def ensure_all_artifacts_included(pyproject: Any, wheel_paths: Iterable[Path]) -> int:
```

evap/evaluation/models.py
```diff
@@ -1108,24 +1116,26 @@ def update_evaluations(cls):
...
 
-        participant_count_subquery = subquery.annotate(
+        participant_count_subquery = subquery.annotate(  # type: ignore[no-redef,misc]
             num_participants=Coalesce("_participant_count", Count("participants")),
         ).values("num_participants")
 
-        voter_count_subquery = subquery.annotate(
+        voter_count_subquery = subquery.annotate(  # type: ignore[no-redef,misc]
             num_voters=Coalesce("_voter_count", Count("voters")),
         ).values("num_voters")
```
niklasmohrin: Here we could also avoid the ignore by changing the name of the annotation:
```suggestion
        participant_count_subquery = subquery.annotate(
            annotated_num_participants=Coalesce("_participant_count", Count("participants")),
        ).values("annotated_num_participants")

        voter_count_subquery = subquery.annotate(
            annotated_num_voters=Coalesce("_voter_count", Count("voters")),
        ).values("annotated_num_voters")
```

evap/evaluation/models.py
```diff
@@ -1136,15 +1146,15 @@ def unlogged_fields(self):
 
 
 @receiver(post_transition, sender=Evaluation)
-def evaluation_state_change(instance, source, **_kwargs):
+def evaluation_state_change(instance: Evaluation, source: int, **_kwargs: Any) -> None:
```
niklasmohrin: This is pretty wild, why do we annotate `instance` just to cast it to Any, and shouldn't `source` be Evaluation.State? We use it like it is in `state_changed_to` above

edit: I think this is a bug above 🤔

evap/evaluation/models.py
```diff
@@ -1389,7 +1399,7 @@ class Meta:
 
 @receiver(post_delete, sender=QuestionAssignment)
 @transaction.atomic
-def _question_assignment_post_delete(*, instance: QuestionAssignment, **_kwargs) -> None:
+def _question_assignment_post_delete(*, instance: QuestionAssignment, **_kwargs: Any) -> None:
```
niklasmohrin, line 1410: (todo: continue reading from here)

evap/evaluation/models_logging.py
```diff
@@ -109,16 +111,17 @@ class Meta:
...
         return {
             field_name: list(_field_actions_for_field(model._meta.get_field(field_name), actions))
             for field_name, actions in self.data.items()
         }
 
     @property
-    def message(self):
-        match self.action_type:
+    def message(self) -> str:
+        match cast("InstanceActionType", self.action_type):
```
niklasmohrin:
```suggestion
        match InstanceActionType[self.action_type]:
```

evap/evaluation/models_logging.py
```diff
@@ -177,7 +184,7 @@ def _as_dict(self):
         ]
         return model_to_dict(self, fields)
 
-    def _get_change_data(self, action_type: InstanceActionType):
+    def _get_change_data(self, action_type: InstanceActionType) -> dict[str, Any]:
```
niklasmohrin:
```suggestion
    def _get_change_data(self, action_type: InstanceActionType) -> dict[str, dict[FieldActionType, list[Any]]]::
```

evap/evaluation/models_logging.py
```diff
@@ -341,14 +354,22 @@ def object_to_attach_logentries_to(self):
...
+def _m2m_changed(  # noqa: PLR0912
+    sender: Any,
+    instance: models.Model | None,
+    action: str | None,
+    reverse: bool,
+    model: type[models.Model],
+    pk_set: set[int] | None,
+    **kwargs: Any,
+) -> None:
+    model_class = cast("type[models.Model]", model if reverse else type(instance))
```
niklasmohrin:
```suggestion
    model_class = model if reverse else type(assert_not_none(instance))
```

evap/evaluation/models_logging.py
```diff
@@ -379,22 +400,22 @@ def _m2m_changed(sender, instance, action, reverse, model, pk_set, **kwargs):  #
         if action_type == FieldActionType.M2M_CLEAR:
             action_type = FieldActionType.M2M_REMOVE
         if pk_set:
-            related_instances = model.objects.filter(pk__in=pk_set)
+            related_instances = cast("Any", model_class).objects.filter(pk__in=pk_set)
```
niklasmohrin:
```suggestion
            related_instances = model_class._default_manager.filter(pk__in=pk_set)
```

evap/rewards/forms.py
```diff
@@ -104,6 +107,7 @@ def save(self) -> list[RewardPointRedemption]:
         return created
 
 
-RewardPointRedemptionFormSet = forms.formset_factory(
-    RewardPointRedemptionForm, formset=BaseRewardPointRedemptionFormSet, extra=0
+RewardPointRedemptionFormSet = cast(
```
niklasmohrin, line 110: hm sad, dependent types when

evap/rewards/views.py
```diff
@@ -28,13 +29,15 @@
...
+def index(request: HttpRequest) -> HttpResponse:
     status = 200
 
+    assert isinstance(request.user, UserProfile)
+
     events = RewardPointRedemptionEvent.objects.filter(redeem_end_date__gte=date.today()).order_by("date")
 
     # pylint: disable=unexpected-keyword-arg
-    formset = RewardPointRedemptionFormSet(
+    formset: Any = RewardPointRedemptionFormSet(
```
niklasmohrin: This works for me:
```suggestion
    formset = RewardPointRedemptionFormSet(
```

evap/staff/importers/base.py
```diff
@@ -198,7 +205,7 @@ def __init__(self, skip_first_n_rows: int, row_cls: type[InputRow], importer_log
         self.row_cls = row_cls
         self.importer_log = importer_log
 
-    def map(self, file_content: bytes):
+    def map(self, file_content: bytes) -> list:
```
niklasmohrin: if you change
1. the class def to `class ExcelFileRowMapper[R: InputRow]`
2. InputRow.from_cells to return Self (from typing)

then we can return `list[R]` here

pyproject.toml
niklasmohrin, line 1: I get a warning
> pyproject.toml: note: unused section(s): module = ['django_sendfile.*']

evap/evaluation/models.py
richardebeling: My thought was that `instance` _is_ an evaluation, so I think we should annotate it as such, independently of what we do with it. The fact that we monkey-annotate some attribute that isn't part of the static "class interface" is orthogonal (-> hence cast to Any). I don't care much about this one-liner here, though, so I can just change it back to `Any`.

Regarding `source`/"bug above": ~~In all our tests, this is only ever called with an `Evaluation.State` value, I'll narrow the type. Would that solve the issue you saw here, or is there still something open then?~~

Edit: Never mind, we have tests where it is called with a `source: int` value. I think this is not immediately a problem, since `Evaluation.State` is a IntegerChoices instance, which I guess can be used somewhat interchangeably with raw ints?
```python3
>>> int(Evaluation.State.NEW)
10
>>> 6 == Evaluation.State.NEW
False
>>> 10 == Evaluation.State.NEW
True
>>> 10 in [Evaluation.State.NEW]
True
```

evap/evaluation/models_logging.py
richardebeling: We have the underlying string value in `self.action_type` (here: `"create"`, `"change"`, `"delete"`), but this construction syntax only accepts the python enum element names (`"CREATE"`, `"CHANGE"`, `"DELETE"`).

```python3
>>> from evap.evaluation.models_logging import InstanceActionType
>>> InstanceActionType["create"]
  File "<console>", line 1, in <module>
  File "/nix/store/829wb290i87wngxlh404klwxql5v18p4-python3-3.13.7/lib/python3.13/enum.py", line 793, in __getitem__
    return cls._member_map_[name]
           ~~~~~~~~~~~~~~~~^^^^^^
KeyError: 'create'
>>> InstanceActionType["CREATE"]
<InstanceActionType.CREATE: 'create'>
```

So with this change, we get test failures

evap/staff/importers/base.py
richardebeling: Hmm, with diff
```diff
diff --git evap/staff/importers/base.py evap/staff/importers/base.py
index 632b6f7e..bbb187bb 100644
--- evap/staff/importers/base.py
+++ evap/staff/importers/base.py
@@ -5,7 +5,7 @@ from collections.abc import Iterable, Iterator
 from dataclasses import dataclass
 from enum import Enum
 from io import BytesIO
-from typing import Any
+from typing import Any, Self
 
 import openpyxl
 from django.conf import settings
@@ -191,11 +191,11 @@ class InputRow(ABC):
         pass
 
     @classmethod
-    def from_cells(cls, location: ExcelFileLocation, cells: Iterable[str]) -> "InputRow":
+    def from_cells(cls, location: ExcelFileLocation, cells: Iterable[str]) -> Self:
         return cls(location, *cells)
 
 
-class ExcelFileRowMapper:
+class ExcelFileRowMapper[R: InputRow]:
     """
     Take a excel file and map it to a list of row_cls instances
     """
@@ -205,7 +205,7 @@ class ExcelFileRowMapper:
         self.row_cls = row_cls
         self.importer_log = importer_log
 
-    def map(self, file_content: bytes) -> list:
+    def map(self, file_content: bytes) -> list[R]:
         try:
             book = openpyxl.load_workbook(BytesIO(file_content))
         except Exception as e:  # noqa: BLE001
```

I get
```bash
 $ ./manage.py typecheck
Executing mypy
evap/staff/importers/base.py:257: error: Incompatible return value type (got "list[InputRow]", expected "list[R]")  [return-value]
evap/staff/importers/user.py:328: error: Need type annotation for "excel_mapper"  [var-annotated]
evap/staff/importers/enrollment.py:707: error: Need type annotation for "input_rows" (hint: "input_rows: list[<type>] = ...")  [var-annotated]
Found 3 errors in 3 files (checked 508 source files)
```

---

This diff works, but it's still somewhat ugly because we now pass the class two times, once for the type system and once to be used as a run-time value. Do we want that?
```diff
diff --git evap/staff/importers/base.py evap/staff/importers/base.py
index 632b6f7e..ed5a4297 100644
--- evap/staff/importers/base.py
+++ evap/staff/importers/base.py
@@ -5,7 +5,7 @@ from collections.abc import Iterable, Iterator
 from dataclasses import dataclass
 from enum import Enum
 from io import BytesIO
-from typing import Any
+from typing import Any, Self
 
 import openpyxl
 from django.conf import settings
@@ -191,21 +191,21 @@ class InputRow(ABC):
         pass
 
     @classmethod
-    def from_cells(cls, location: ExcelFileLocation, cells: Iterable[str]) -> "InputRow":
+    def from_cells(cls, location: ExcelFileLocation, cells: Iterable[str]) -> Self:
         return cls(location, *cells)
 
 
-class ExcelFileRowMapper:
+class ExcelFileRowMapper[R: InputRow]:
     """
     Take a excel file and map it to a list of row_cls instances
     """
 
-    def __init__(self, skip_first_n_rows: int, row_cls: type[InputRow], importer_log: ImporterLog):
+    def __init__(self, skip_first_n_rows: int, row_cls: type[R], importer_log: ImporterLog):
         self.skip_first_n_rows = skip_first_n_rows
         self.row_cls = row_cls
         self.importer_log = importer_log
 
-    def map(self, file_content: bytes) -> list:
+    def map(self, file_content: bytes) -> list[R]:
         try:
             book = openpyxl.load_workbook(BytesIO(file_content))
         except Exception as e:  # noqa: BLE001
diff --git evap/staff/importers/enrollment.py evap/staff/importers/enrollment.py
index a36cb5f5..36a66e8e 100644
--- evap/staff/importers/enrollment.py
+++ evap/staff/importers/enrollment.py
@@ -704,7 +704,7 @@ def import_enrollments(
     importer_log = ImporterLog()
 
     with ConvertExceptionsToMessages(importer_log):
-        input_rows = ExcelFileRowMapper(
+        input_rows = ExcelFileRowMapper[EnrollmentInputRow](
             skip_first_n_rows=1,
             row_cls=EnrollmentInputRow,
             importer_log=importer_log,
diff --git evap/staff/importers/user.py evap/staff/importers/user.py
index 51f82e99..fd3f60d1 100644
--- evap/staff/importers/user.py
+++ evap/staff/importers/user.py
@@ -325,7 +325,7 @@ def import_users(excel_content: bytes, test_run: bool) -> tuple[list[UserProfile
     importer_log = ImporterLog()
 
     with ConvertExceptionsToMessages(importer_log):
-        excel_mapper = ExcelFileRowMapper(skip_first_n_rows=1, row_cls=UserInputRow, importer_log=importer_log)
+        excel_mapper = ExcelFileRowMapper[UserInputRow](skip_first_n_rows=1, row_cls=UserInputRow, importer_log=importer_log)
         raw_rows = excel_mapper.map(excel_content)
         importer_log.raise_if_has_errors()
```

niklasmohrin: I would expect that the class-wide type would be inferred if you pass `row_cls=UserInputRow` / I thought that it did infer it when I tried it out; I also like the second diff as-is if it's not inferred

evap/evaluation/models.py
niklasmohrin: okay fair enough; I feel that we should promote to our enum as early as possible at runtime, but as you show it's not broken so no concern for this PR

evap/evaluation/models_logging.py
niklasmohrin: ah, the typechecker fooled me into believing this works; what we want is this?
```suggestion
        match InstanceActionType(self.action_type):
```

evap/staff/importers/base.py
richardebeling: Hmm, works now, used the smaller version :shrug: Maybe I tried with an outdated mypy version last time.

evap/evaluation/models.py
```diff
@@ -1399,13 +1409,13 @@ def _question_assignment_post_delete(*, instance: QuestionAssignment, **_kwargs)
 @dataclass
 class Choices:
     css_class: str
-    values: tuple[Real]
+    values: tuple[int]
     colors: tuple[str]
-    grades: tuple[Real]
+    grades: tuple[int]
```
niklasmohrin: not an int below, but we ignore it

evap/evaluation/models.py
```diff
@@ -2060,68 +2087,66 @@ def is_student(self):
             return True
 
         last_semester_participated = (
-            Semester.objects.filter(courses__evaluations__participants=self).order_by("-created_at").first()
+            Semester.objects.filter(courses__evaluations__participants=self).order_by("-created_at").get()
```
niklasmohrin: naaaaah, students can participate in more than one semester

evap/results/views.py
```diff
@@ -28,40 +33,47 @@
 )
 from evap.tools import unordered_groupby
 
+if TYPE_CHECKING:
+    from typing import Any
```
niklasmohrin, line 37: We already import `typing` to import `TYPE_CHECKING`, so we can afford to move `Any` there too :D

(also `PublishedRatingResult`)

evap/staff/staff_mode.py
```diff
@@ -41,19 +48,20 @@ def middleware(request):
...
 
 
-def is_in_staff_mode(request):
+def is_in_staff_mode(request: HttpRequest) -> bool:
     return "staff_mode_start_time" in request.session
 
 
-def update_staff_mode(request):
+def update_staff_mode(request: HttpRequest) -> None:
+    assert isinstance(request.user, UserProfile)
```
niklasmohrin, line 64: this one had me thinking a moment to convince myself that unauthenticated users cannot end up in the middleware here. it would be much nicer if this would be enforced in the enter_staff_mode view and in the middleware directly. if I recall correctly, we haven't found a good way yet to get a `AuthenticatedHttpRequest` protocol working, right?

evap/staff/views.py
```diff
@@ -706,14 +706,15 @@ def semester_import(request, semester_id):
     import_type = ImportType.SEMESTER
 
     importer_log = None
+    user_id = assert_not_none(request.user.id)
```
niklasmohrin: we can just do `assert request.user.id is not None` and leave the lines below untouched (same in the other views)

evap/staff/views.py
```diff
@@ -2322,15 +2329,15 @@ def user_index(request):
 
 
 @manager_required
-def user_list(request):
-    filter_users = get_bool_parameter_from_url_or_session(request, "filter_users")
+def user_list(request: HttpRequest) -> HttpResponse:
+    filter_users = get_string_parameter_from_url_or_session(request, "filter_users")
```
niklasmohrin: why the change to string? seems to be used as a bool

evap/staff/views.py
```diff
@@ -2691,14 +2700,14 @@ class InfotextsView(SuccessMessageMixin, SaveValidFormMixin, FormsetView):
...
-def download_sample_file(_request, filename):
+def download_sample_file(_request: HttpRequest, filename: str) -> HttpResponse:
     email_placeholder = "institution.com"
 
     if filename not in ["sample.xlsx", "sample_user.xlsx"]:
         raise SuspiciousOperation("Invalid file name.")
 
     book = openpyxl.load_workbook(filename=settings.STATICFILES_DIRS[0] / filename)
-    for sheet in book:
+    for sheet in book:  # type: ignore[attr-defined]
```
niklasmohrin, line 2705: sad, should we switch to typeshed for the stubs? see https://redirect.github.com/MartinThoma/openpyxl-stubs/issues/30

evap/staff/forms.py
```diff
@@ -53,27 +56,27 @@ class CharArrayField(forms.Field):
...
         return BoundCharArrayField(form, self, field_name)
 
 
 class BoundCharArrayField(forms.BoundField):
-    def as_widget(self, widget=None, attrs=None, only_initial=False):
+    def as_widget(self, widget: Widget | None = None, attrs: Any = None, only_initial: bool = False) -> SafeString:
         widget = widget or self.field.widget
         # Inject all current values as choices so they don’t get discarded
-        if self.value():
+        if self.value() and hasattr(widget, "choices"):
```
niklasmohrin, line 79: hm, not really in this code right now, but isn't is possible that we just give the initial definition of `choices` here?

evap/results/tools.py
```diff
@@ -69,14 +79,18 @@ def create_rating_result(question: Question, answer_counters, additional_text_re
 
 class RatingResult:
     @classmethod
-    def is_published(cls, rating_result) -> TypeGuard["PublishedRatingResult"]:
+    def is_published(
+        cls, rating_result: "RatingResult | TextResult | HeadingResult"
```
niklasmohrin, line 82: from the name I would think that we only intend to pass `RatingResult` here, do we pass the other types too?

evap/results/tools.py
```diff
@@ -100,7 +119,7 @@ def __init__(self, question: Question, answer_counters, additional_text_result=N
         for answer_counter in answer_counters:
             assert counts[answer_counter.answer][0] == 0
             counts[answer_counter.answer][0] = answer_counter.count
-        self.counts = tuple(count for count, _, _, _ in counts.values())
+        self.counts = tuple(cast("int", count) for count, _, _, _ in counts.values())
```
niklasmohrin, line 121: can't we just give a type annotation to `counts` above?

evap/results/tools.py
```diff
@@ -295,12 +314,13 @@ def normalized_distribution(distribution):
...
+def unipolarized_distribution(result: PublishedRatingResult) -> Distribution:
+    summed_distribution: list[float] = [0, 0, 0, 0, 0]
 
     if not result.counts:
         return None
 
+    grade: float | Real
     for counts, grade in zip(result.counts, result.choices.grades, strict=True):
         grade_fraction, grade = modf(grade)
         grade = int(grade)
```
niklasmohrin: strange, can we do the following instead?
```suggestion
    for counts, grade in zip(result.counts, result.choices.grades, strict=True):
        grade_fraction, grade_whole = modf(grade)
        grade = int(grade_whole)
```

evap/results/tools.py
```diff
@@ -311,39 +331,49 @@ def unipolarized_distribution(result):
     return normalized_distribution(summed_distribution)
 
 
-def avg_distribution(weighted_distributions):
+def avg_distribution(weighted_distributions: Iterable[tuple[tuple[float, ...] | None, float]]) -> Distribution:
```
niklasmohrin:
```suggestion
def avg_distribution(weighted_distributions: Iterable[tuple[Distribution, float]]) -> Distribution:
```

niklasmohrin (commented):
looked at everything now

evap/staff/views.py
richardebeling: seems like I have messed up the rebase. Fixed, thanks

evap/results/tools.py
richardebeling: > evap/results/exporters.py:303: error: Argument 1 to "has_answers" of "RatingResult" has incompatible type "RatingResult | TextResult | HeadingResult"; expected "RatingResult"  [arg-type]

niklasmohrin: agh, the big pot of results that is evaluations_with_results :(

evap/results/tools.py
richardebeling: But what do we annotate it as? `dict[int, list[X]]`? The list has homogeneous elements here. We cannot trivially make it a tuple because we modify it below:
```python3
            counts[answer_counter.answer][0] = answer_counter.count
```

evap/staff/forms.py
richardebeling: My reasoning here was that anything that respects choices is something that also sets a `choices` attribute on itself, see `ChoiceWidget` source code:
https://github.com/django/django/blob/fb1137658ff597ffac758c3246a9bf8f0e90c753/django/forms/widgets.py#L742. I think it makes sense for us to _not_ annotate choices to a widget that doesn't have choices.

`Select` [inherits](https://github.com/django/django/blob/fb1137658ff597ffac758c3246a9bf8f0e90c753/django/forms/widgets.py#L869), `SelectMultiple` [inherits from `Select`](https://github.com/django/django/blob/fb1137658ff597ffac758c3246a9bf8f0e90c753/django/forms/widgets.py#L949), `RadioSelect` [inherits](https://github.com/django/django/blob/fb1137658ff597ffac758c3246a9bf8f0e90c753/django/forms/widgets.py#L965), `CheckboxSelectMultiple` [inherits from `RadioSelect`](https://github.com/django/django/blob/fb1137658ff597ffac758c3246a9bf8f0e90c753/django/forms/widgets.py#L982), so every widget with choices [from the docs](https://docs.djangoproject.com/en/6.0/ref/forms/widgets/) is covered.

In theory, the error scenario is possible, so if you would like to be defensive, we can also be defensive.

evap/results/tools.py
niklasmohrin: okay, so let's keep the cast, but I feel that at some point we should refactor this instead to avoid the mutation altogether

evap/staff/views.py
richardebeling: I'm open to it, but for the sake of stopping the rebasing madness here, maybe let's think about this separately

niklasmohrin (approved):
Thanks!

evap/staff/staff_mode.py
richardebeling: You're thinking about adding the protocol for a `HttpRequest` that always has `request.user` set and be a non-`AnonymousUser`, and then make the `RequireLoginMiddleware` stuff that protocol into the middleware call chain, so that the type-system basically is the proof that the logged-in-ness of the user was asserted some time before?

I don't think I've tried, I like the idea, but maybe I just forget the last failure :D

evap/staff/forms.py
niklasmohrin: Fair, your new version seems more reasonable than what we had before

evap/staff/staff_mode.py
niklasmohrin: everything until "be a non-Anonymoususer": yes; everything after: I don't think Python has the necessary type system features to let RequireLoginMiddleware declare this proof (sounds dependent-type-y to me)

but no one is checking the type of view arguments right now anyways (I think); the usage site of the views is the call to `path` in `urls.py`, and this accepts any callable returning a response (https://github.com/typeddjango/django-stubs/blob/085ae56c4fb116673909e0ecf5b35a62e238047d/django-stubs/urls/conf.pyi#L17). so we could just annotate this on any view that doesn't have the `no_login_required` decorator. if we want to be thorough, we can annotate `no_login_required` to require that the view handles the AnonymousUser case, but I am again not sure if there is a nice Python typing way to formulate this

richardebeling: I see. My idea was to really use the type system as a proof that an `assert request.user is not None` was executed. We shouldn't need dependent types for this, but we need for our sub-type to be properly passed through into the view function. If there is infrastructure that drops our more specialized type again, we cannot have that.

I would like to keep the actual assert, I just wanted to move it to a central place and have it typechecked by mypy. "Just annotating any view that doesn't have `no_login_required`" does not achieve that, so I'd be worried that we miss the check then.

I think keeping the asserts is fine, for now

evap/results/views.py
richardebeling: A few of these are only used in Strings, i.e., as `cast("Any", ...)`. Ruff enforces this:
```
 TC006 [*] Add quotes to type expression in `typing.cast()`
   --> evap/results/views.py:168:14
    |
166 |     )
167 |     for course, annotated_course in zip(courses_and_evaluations.keys(), annotated_courses, strict=True):
168 |         cast(Any, course).num_evaluations = annotated_course.num_evaluations
    |              ^^^
169 |
170 |     programs = Program.objects.filter(courses__pk__in=course_pks).distinct()
    |
```

And if only used in Strings, pylint fails if we import them unconditionally:

```
Executing pylint evap
************* Module evap.results.views
evap/results/views.py:3:0: W0611: Unused Any imported from typing (unused-import)
evap/results/views.py:19:0: W0611: Unused PublishedRatingResult imported from evap.results.tools (unused-import)
```

PR #2688, opened 2026-03-30
2604 preparation reminder for internal users

### LINKED ISSUE #2604: Preparation reminder for internal users
On the staff page `preparation_reminder` there is an option to "Remind all". Additionally, there should be an analogous option to remind only internal users ("Remind all internal").
This will send reminders to all users with `is_external==False`.

ybrnr:
closes #2604: On staff:preparation_reminder, next to "Remind all", adds a "Remind all internal" button.


evap/staff/views.py
```diff
@@ -959,13 +959,16 @@ def semester_questionnaire_assign(request, semester_id):
...
+    internal_only = request.POST.get("internal_only") == "true"
 
     evaluations = semester.evaluations.filter(
         state__in=[Evaluation.State.PREPARED, Evaluation.State.EDITOR_APPROVED]
     ).prefetch_related("course__programs")
 
     prepared_evaluations = semester.evaluations.filter(state=Evaluation.State.PREPARED)
     responsibles = UserProfile.objects.filter(courses_responsible_for__evaluations__in=prepared_evaluations).distinct()
+    if internal_only:
+        responsibles = responsibles.exclude(pk__in=[r.pk for r in responsibles if r.is_external])
```
richardebeling, line 971: `is_external` is implemented in python logic, so django cannot convert it into a SQL query. That means we don't "win" anything by adding this filter to the query early. Note that the code here currently fetches responsibles, computes `is_external` for each of them, and then amends the original query with "where id is not in [list of external IDs]" -- this is just one extra database round trip without real benefit.

Above, you access `request.POST`, but outside the `if request.method == 'POST'` block below.

I think for maintainability, it makes sense to put this check inside this block. That way, we keep the queries simple, save one database roundtrip, don't build large SQL, and the world is greener.

The code change would then just be
```python3
if request.POST.get("internal_only") == "true" and responsible.is_external:
    continue
```

evap/staff/tests/test_views.py
```diff
@@ -1064,6 +1064,27 @@ def test_remind_all(self, email_template_mock) -> None:
...
+        )
+
+        email_template_mock.objects.get.return_value = email_template_mock
+        email_template_mock.EDITOR_REVIEW_REMINDER = EmailTemplate.EDITOR_REVIEW_REMINDER
+
+        self.app.post(self.url, params={"internal_only": "true"}, user=self.manager, status=200)
+
+        recipients = [call[0][0] for call in email_template_mock.send_to_user.call_args_list]
+        self.assertIn(self.user, recipients)
+        self.assertNotIn(external_responsible, recipients)
```
richardebeling, line 1086: Would it be less code to not patch the email templates and just check mail.outbox afterwards? If so, let's do that (otherwise, ignore).

evap/staff/templates/staff_semester_preparation_reminder.html
```diff
@@ -15,19 +15,37 @@ <h3>
...
+
+                        <input type="hidden" name="semester_id" value="{{ semester.id }}" />
+                        <input type="hidden" name="internal_only" value="true">
+
+                        <confirmation-modal type="submit">
+                            <span slot="title">{% translate 'Remind all internal users' %}</span>
+                            <span slot="action-text">{% translate 'Remind all internal users' %}</span>
+                            <span slot="question">{% translate 'Do you really want to remind all internal users?' %}</span>
+
+                            <button slot="show-button" type="button" id="remindAllInternalUsersButton" class="btn btn-sm btn-light">{% translate 'Remind all internal users' %}</button>
```
richardebeling, line 44: The `id` here, `remindAllInternalUsersButton`, is never used. Let's remove it.

While you're at it, the same is true for `remindAllButton` above.

evap/staff/views.py
```diff
@@ -981,7 +984,7 @@ def semester_preparation_reminder(request: HttpRequest, semester_id: int) -> Htt
         for responsible, evaluations, __ in responsible_list:
             body_params = {"user": responsible, "evaluations": evaluations}
             template.send_to_user(responsible, subject_params={}, body_params=body_params, use_cc=True, request=request)
-        messages.success(request, _("Successfully sent reminders to everyone."))
+        messages.success(request, _("Successfully sent reminders."))
```
janno42, line 987: two different success messages for the two cases would be nice

richardebeling (changes requested):
Marking as "changes requested" to clean-up our "to review" lists

PR #2691, opened 2026-04-01
Tests debuggability improvements. Fix flaky `test_edit_textanswer_redirect`. Improve live test speed.

### LINKED ISSUE #2652: Textanswer redirect test seems flaky

richardebeling:
Fixes #2652.

Best reviewed by commit.

* Removes implicit wait on selenium web driver find interactions. This made tests flaky before, we had scenarios where we actually needed to wait, but we didn't use an explicit wait with a long enough timeout. Rationale: We should be explicit and strict about whether interactions need to wait or whether elements should already be there. I also reduced our explicit wait polling wait time from the default of 500ms to 100ms. Most of our javascript is extremely lightweight, if something isn't there immediately, most likely, after 100ms, it will be there.
* Moves out `get_open_modals` into a generic helper function.
* Fixes instances where the tests get stuck waiting needlessly.
* Makes LiveServerTests produce a stacktrace / drop into debugger when we run into an internal server error. Before, we returned our generic production "Something went wrong" page, which caused later test assertions to fail in a way that made in unnecessarily hard to debug.
* Adds a mechanism to re-run tests with the same baker-generated values using test command line flag `--baker-seed`. UX is copied from django's `--shuffle` feature. This should allow reproducing flaky test failures much more easily.
* Fixes flakiness of `test_edit_textanswer_redirect`: There was a timing issue, and the server would fail an assertion when the question type happened to be randomly generated as `HEADING`. I ran all tests with a default question type `HEADING` to see if there are more affected tests, but every other test passes.

Runtime (`time ./manage.py test --tag selenium`)
* before: 138s
* after: 44s

### DIFF
--- a/evap/evaluation/tests/tools.py
+++ b/evap/evaluation/tests/tools.py
@@ -1,3 +1,4 @@
+import random
 import time
 from collections.abc import Iterator, Sequence
 from contextlib import contextmanager
@@ -44,10 +45,11 @@ from evap.evaluation.models import (
 class EvapTestRunner(DiscoverRunner):
     """Skips selenium tests by default, if no other tags are specified."""
 
-    def __init__(self, *args: Any, headed=False, **kwargs: Any) -> None:
+    def __init__(self, *args: Any, headed: bool, baker_seed: int, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
 
         self.__headed = headed
+        self.__baker_seed = baker_seed
 
         if not self.tags and not self.exclude_tags:
             self.exclude_tags = {"selenium"}
@@ -62,10 +64,21 @@ class EvapTestRunner(DiscoverRunner):
             action="store_true",
         )
 
+        parser.add_argument(
+            "--baker-seed",
+            nargs="?",
+            type=int,
+            default=random.getrandbits(32),
+            help="Force model-bakery to use this seed to generate random values. Default: random value",
+        )
+
     def setup_test_environment(self, **kwargs):
         super().setup_test_environment(**kwargs)
 
         LiveServerTest.headless = not self.__headed
+
+        baker.seed(self.__baker_seed)
+        self.log(f"Using baker seed: {self.__baker_seed}")
 
 
 class ResetLanguageOnTearDownMixin:
--- a/evap/evaluation/tests/tools.py
+++ b/evap/evaluation/tests/tools.py
@@ -77,8 +77,8 @@ class EvapTestRunner(DiscoverRunner):
 
         LiveServerTest.headless = not self.__headed
 
-        baker.seed(self.__baker_seed)
         self.log(f"Using baker seed: {self.__baker_seed}")
+        settings.EVAP_TEST_BAKER_SEED = self.__baker_seed
 
 
 class ResetLanguageOnTearDownMixin:
@@ -87,15 +87,32 @@ class ResetLanguageOnTearDownMixin:
         super().tearDown()
 
 
-class TestCase(ResetLanguageOnTearDownMixin, django.test.TestCase):
+class SeedBakerMixin:
+    @classmethod
+    def setUpClass(cls):
+        # runs before `setUpTestData`
+        baker.seed(settings.EVAP_TEST_BAKER_SEED)
+        super().setUpClass()
+
+    @classmethod
+    def _pre_setup(cls):
+        # runs before `setUp`
+        # Same seed would imply the same value sequence, which causes problems:
+        # * uniqueness constraints fail if setUpTestData and setUp both generate an instance of the same model class
+        # * "assert name not in page"-style asserts fail for shared names with non-unique fields
+        baker.seed(settings.EVAP_TEST_BAKER_SEED + 1)
+        super()._pre_setup()
+
+
+class TestCase(SeedBakerMixin, ResetLanguageOnTearDownMixin, django.test.TestCase):
     pass
 
 
-class SimpleTestCase(ResetLanguageOnTearDownMixin, django.test.SimpleTestCase):
+class SimpleTestCase(SeedBakerMixin, ResetLanguageOnTearDownMixin, django.test.SimpleTestCase):
     pass
 
 
-class WebTest(ResetLanguageOnTearDownMixin, django_webtest.WebTest):
+class WebTest(SeedBakerMixin, ResetLanguageOnTearDownMixin, django_webtest.WebTest):
     pass
 
 
@@ -317,7 +334,7 @@ def assert_no_database_modifications(*args, **kwargs):
 
 # For the average LiveServerTest, if the "server" has an internal error, we want to abort/stacktrace/drop into debugger
 @override_settings(DEBUG_PROPAGATE_EXCEPTIONS=True)
-class LiveServerTest(SeleniumTestCase):
+class LiveServerTest(SeedBakerMixin, SeleniumTestCase):
     browser = "firefox"
     selenium: WebDriver
     headless = True
--- a/evap/evaluation/tests/tools.py
+++ b/evap/evaluation/tests/tools.py
@@ -14,6 +14,7 @@ from django.contrib.auth.models import Group
 from django.contrib.staticfiles.handlers import StaticFilesHandler
 from django.db import DEFAULT_DB_ALIAS, connections
 from django.http.request import HttpRequest, QueryDict
+from django.test import override_settings
 from django.test.runner import DiscoverRunner
 from django.test.selenium import SeleniumTestCase
 from django.test.utils import CaptureQueriesContext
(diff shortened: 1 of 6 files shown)


evap/results/tests/test_live.py
```diff
@@ -130,8 +130,10 @@ def test_results_filter_with_search_input(self):
         self.selenium.get(self.url)
         self.wait.until(visibility_of_element_located((By.CLASS_NAME, "reset-button"))).click()
 
+        course_c_span = self.selenium.find_element(By.XPATH, "//span[contains(text(),'Course C')]")
+
         self.selenium.find_element(By.CSS_SELECTOR, "input[name=search]").send_keys("Exam")
-        self.wait.until(invisibility_of_element_located((By.XPATH, "//span[contains(text(),'Course C')]")))
+        self.wait.until(invisibility_of_element_located(course_c_span))
```
richardebeling, line 136: `invisibility_of_element_located`, together with this XPATH selector here, breaks polling. This test was stuck here for 10 seconds before it continued. Selecting the DOM element before and then waiting for invisibility of the DOM element is more efficient in general, and works with polling.

evap/staff/tests/test_live.py
```diff
@@ -296,31 +297,31 @@ def test_edit_textanswer_redirect(self):
...
-                WebDriverWait(self.selenium, 1).until(
-                    visibility_of_element_located((By.ID, f"textanswer-{str(textanswer1.pk)}"))
+            while True:
+                textanswer_element_visible = self.wait.until(
+                    visibility_of_element_located((By.CSS_SELECTOR, ".slider-item.card-body.active[id^=textanswer-]"))
                 )
+                if textanswer_element_visible.get_attribute("id") != f"textanswer-{str(textanswer1.pk)}":
+                    next_textanswer_btn.click()
+                    continue
                 break
```
richardebeling, line 298: This skips waiting 3 * 1 second needlessly (cc @geromequa)

evap/staff/tests/test_live.py
```diff
@@ -296,31 +297,31 @@ def test_edit_textanswer_redirect(self):
...
-            except TimeoutException:
-                next_textanswer_btn.click()
 
-        with self.enter_staff_mode():
             edit_btn.click()
 
-        textanswer_field = self.selenium.find_element(By.XPATH, "//textarea[@name='answer']")
-        submit_btn = self.selenium.find_element(By.ID, "textanswer-edit-submit-button")
+            textanswer_field = self.selenium.find_element(By.XPATH, "//textarea[@name='answer']")
```
richardebeling, line 302: Code here was indented so that everything runs inside one `with self.enter_staff_mode()`. I think this makes sense, realistically, users aren't entering and leaving staff mode repeatedly during this interaction here.

We probably want a `LiveTestStaffMode` for staff-mode pages in general. Currently, we cannot use the decorator on test classes, since it needs a `self` instance.

evap/evaluation/tests/tools.py
```diff
@@ -61,11 +64,22 @@ def add_arguments(cls, parser):
...
+            default=random.getrandbits(32),
+            help="Force model-bakery to use this seed to generate random values. Default: random value",
+        )
+
     def setup_test_environment(self, **kwargs):
         super().setup_test_environment(**kwargs)
 
         LiveServerTest.headless = not self.__headed
 
+        baker.seed(self.__baker_seed)
```
richardebeling: I think it would be even more convenient to re-seed the rng before each test. If a test fails only with a specific set of random data, with the setup here, you'd have to run all previous tests as well to get the RNG into the same state. If we reseed for every test, you can easily debug individual tests.

It doesn't seem like the Runner has a method that is called once per test, though. I don't know if the TestCase class could get the seed from the runner to do this in `setUp` or similar. I also don't think we have consistent `super()` calls in these methods.

evap/evaluation/tests/tools.py
```diff
@@ -313,6 +329,7 @@ def setUp(self) -> None:
         super().setUp()
         self.request = self.make_request()
         self.manager = make_manager()
+        self.selenium.implicitly_wait(0)
```
richardebeling: set `implicit_wait = 0` on the class instead of calling this here. Django does `self.selenium.implicitly_wait(cls.implicit_wait)`

evap/evaluation/tests/tools.py
```diff
@@ -301,10 +332,13 @@ def assert_no_database_modifications(*args, **kwargs):
                     raise AssertionError("Unexpected modifying query found: " + query["sql"])
 
 
-class LiveServerTest(SeleniumTestCase):
+# For the average LiveServerTest, if the "server" has an internal error, we want to abort/stacktrace/drop into debugger
+@override_settings(DEBUG_PROPAGATE_EXCEPTIONS=True)
+class LiveServerTest(SeedBakerMixin, SeleniumTestCase):
```
richardebeling, line 339: Currently, each LiveServerTest class spawns a new firefox process. We have 8 subclasses.

If we reuse the same firefox process between test classes, we save 16 seconds (out of 44 seconds, 36%) of test runtime.

I've added the code changes that achieves this to https://github.com/e-valuation/EvaP/commit/fef7b8101d76b14aaaffb6a183f0a77620d9038f, but it's a bit hacky, I don't know if we want this.

niklasmohrin: Is teardown_databases called on all threads? Otherwise, as far as I understand, we call quit() only on one of the threads

niklasmohrin: Either way, we can do this in another PR

evap/evaluation/tests/tools.py
```diff
@@ -61,27 +64,55 @@ def add_arguments(cls, parser):
...
+            help="Force model-bakery to use this seed to generate random values. Default: random value",
+        )
+
     def setup_test_environment(self, **kwargs):
         super().setup_test_environment(**kwargs)
 
         LiveServerTest.headless = not self.__headed
 
+        self.log(f"Using baker seed: {self.__baker_seed}")
+        settings.EVAP_TEST_BAKER_SEED = self.__baker_seed
```
niklasmohrin: Why a setting? I feel that it would be easier to set `SeedBakerMixin.seed = ...` instead

evap/evaluation/tests/tools.py
```diff
@@ -301,10 +332,13 @@ def assert_no_database_modifications(*args, **kwargs):
                     raise AssertionError("Unexpected modifying query found: " + query["sql"])
 
 
-class LiveServerTest(SeleniumTestCase):
+# For the average LiveServerTest, if the "server" has an internal error, we want to abort/stacktrace/drop into debugger
+@override_settings(DEBUG_PROPAGATE_EXCEPTIONS=True)
```
niklasmohrin, line 338: Does this stall CI if it happens there? I would expect that stdin is just eof, but do we know?

evap/staff/tests/test_live.py
```diff
@@ -180,63 +181,48 @@ def test_collapse_without_editor_approved(self) -> None:
...
+        semester = baker.make(Semester)
 
         baker.make(
             Evaluation,
             _quantity=7,
-            name_de=iter(f"Evaluation {i}" for i in range(1, 8)),
-            name_en=iter(f"Evaluation {i}" for i in range(1, 8)),
+            name_en=baker.seq("Evaluation"),
+            name_de=baker.seq("Evaluation"),
+            course__name_en=iter(("AA", "ÄB", "AC", "AE", "UB", "ÜC", "Z")),
```
niklasmohrin, line 191: Just so we are on the same page: Before, the German names were the other way around and we expected the same order; now they are the same and we expect the reverse order for German?

But after setting the language and reloading the page, the order selection is cached and restored so the test asserts that the order is the same regardless of language?

evap/student/tests/test_live.py
```diff
@@ -130,12 +132,19 @@ def test_checking_bottom_confirm_checkbox_check_top_but_keeps_bottom_visible(sel
     def test_resolving_submit_errors_clears_warning(self) -> None:
         self.selenium.get(self.url)
         with self.wait_until_page_reloads():
-            self.wait.until(presence_of_element_located((By.ID, "vote-submit-btn"))).click()
+            self.selenium.find_element(By.ID, "vote-submit-btn").click()
+
+        # wait for all java script to full execute, so our click handlers are registered
```
niklasmohrin:
```suggestion
        # wait for all javascript to fully execute, so our click handlers are registered
```

richardebeling: oops

evap/evaluation/tests/tools.py
richardebeling: No particular reason. Will put it into SeedBakerMixin.seed

evap/evaluation/tests/tools.py
richardebeling: It just works: We get the stack trace of the error dumped in the log, and the test fails, the test invocation returns non-null exit code. See https://github.com/e-valuation/EvaP/actions/runs/24357097442/job/71126797312?pr=2691#step:5:934.

evap/staff/tests/test_live.py
richardebeling: basically, yes. Added comment, as discussed offline.

niklasmohrin (approved):

Issue #2296, opened 2024-10-03
Content Security Policy


richardebeling:
See #2099.

* [ ] Add django-csp
* [ ] All html script tags should be annotated with a nonce (`nonce="{{ CSP_NONCE }}"`), then we can enforce nonces for scripts. For template tags, we probably have to [manually forward the context ](https://github.com/e-valuation/EvaP/pull/2099/files#diff-6a4f496a5384e23594e4f21829be4f03f8a5ccf739e75b9c14b09d54b96caaddR9). CSP Configuration should then look something like [this](https://github.com/e-valuation/EvaP/pull/2099/files#diff-819b34c938f55401df6712f972da4abc15e76811cc5382922fa799345c03c10eR407).

* [ ] To fix CSS, we need to figure out how we can handle dynamically generated colors. Using `attr()` in CSS doesn't work with current browsers, at least for colors. The only workaround I currently see is having custom javascript that translates `data-X` helper attributes for color into "inline" style (using the `.style` attribute) -- seems a bit ugly to me

* [ ] To fix images, we'd need to move the `data:` images we currently have (3 svg paths in CSS files) into separate files. These are currently inlined into CSS to use our color definitions.

* [ ] Consider setting up CSP failure reporting, maybe using some external service that tracks / aggregates them for us? Would leak user data though. mozilla-django-csp at least had a report-view in the past, but the documentation looks they don't want to maintain that anymore

PR #2697, opened 2026-04-13
Add copy name option

### LINKED ISSUE #2673: Copy evaluation name to clipboard
This should be implemented after #2649 (with [condense staff semester evaluation action button area](https://github.com/e-valuation/EvaP/pull/2649/changes/6ffafdcdcd135d83e4c0a724510897d2d3630248)) was merged.

On the staff semester page, a new action should be added to the dropdown for each evaluation row: "Copy Name to Clipboard," which ... copies the evaluation's name to the clipboard.
On the staff evaluation edit page, clicking on the heading should also copy the evaluation's name (without the semester) to the clipboard. A tooltip explaining this should appear when hovering over the heading.

ZoqkMaze:
fix #2673

### DIFF
--- a/evap/staff/templates/staff_course_form.html
+++ b/evap/staff/templates/staff_course_form.html
@@ -6,9 +6,9 @@
     {{ block.super }}
     <h3>
         {% if course_form.instance.id %}
-            <a data-bs-toggle="tooltip" data-bs-placement="right" title="{% translate 'Click to copy name' %}" onclick="navigator.clipboard.writeText('{{ course.name }}')">
+            <span data-bs-toggle="tooltip" data-bs-placement="right" title="{% translate 'Click to copy name' %}" onclick="navigator.clipboard.writeText('{{ course.name }}')">
                 {{ course.name }} ({{ course.semester.name }})
-            </a>
+            </span>
         {% else %}
             {% translate 'Create course' %}
         {% endif %}
--- a/evap/staff/templates/staff_course_form.html
+++ b/evap/staff/templates/staff_course_form.html
@@ -6,7 +6,7 @@
     {{ block.super }}
     <h3>
         {% if course_form.instance.id %}
-            <span data-bs-toggle="tooltip" data-bs-placement="right" title="{% translate 'Click to copy name' %}" onclick="navigator.clipboard.writeText('{{ course.name }}')">
+            <span data-bs-toggle="tooltip" data-bs-placement="right" title="{% translate 'Click to copy name' %}" data-copy-on-click="{{ course.name }}">
                 {{ course.name }} ({{ course.semester.name }})
             </span>
         {% else %}
--- a/evap/staff/templates/staff_course_form.html
+++ b/evap/staff/templates/staff_course_form.html
@@ -6,7 +6,9 @@
     {{ block.super }}
     <h3>
         {% if course_form.instance.id %}
-            {{ course.name }} ({{ course.semester.name }})
+            <a data-bs-toggle="tooltip" data-bs-placement="right" title="{% translate 'Click to copy name' %}" onclick="navigator.clipboard.writeText('{{ course.name }}')">
+                {{ course.name }} ({{ course.semester.name }})
+            </a>
         {% else %}
             {% translate 'Create course' %}
         {% endif %}
--- a/evap/staff/templates/staff_course_form.html
+++ b/evap/staff/templates/staff_course_form.html
@@ -6,7 +6,7 @@
     {{ block.super }}
     <h3>
         {% if course_form.instance.id %}
-            <span data-bs-toggle="tooltip" data-bs-placement="right" title="{% translate 'Click to copy name' %}" data-copy-on-click="{{ course.name }}">
+            <span data-bs-toggle="tooltip" data-bs-placement="right" title="{% translate 'Click to copy name' %}" copyOnClick="{{ course.name }}">
                 {{ course.name }} ({{ course.semester.name }})
             </span>
         {% else %}
--- a/evap/staff/templates/staff_course_form.html
+++ b/evap/staff/templates/staff_course_form.html
@@ -6,7 +6,7 @@
     {{ block.super }}
     <h3>
         {% if course_form.instance.id %}
-            <span data-bs-toggle="tooltip" data-bs-placement="right" title="{% translate 'Click to copy name' %}" copyOnClick="{{ course.name }}">
+            <span data-bs-toggle="tooltip" data-bs-placement="right" title="{% translate 'Click to copy name' %}" data-copy-on-click="{{ course.name }}">
                 {{ course.name }} ({{ course.semester.name }})
             </span>
         {% else %}
--- a/evap/staff/templates/staff_evaluation_form.html
+++ b/evap/staff/templates/staff_evaluation_form.html
@@ -12,13 +12,13 @@
                 </div>
             {% endif %}
             <h3 class="pt-2">
-                <a data-bs-toggle="tooltip" data-bs-placement="right" title="{% translate 'Click to copy name' %}" onclick="navigator.clipboard.writeText('{{ evaluation.full_name }}')">
+                <span class="btn-clipboard" data-bs-toggle="tooltip" data-bs-placement="right" title="{% translate 'Click to copy name' %}" onclick="navigator.clipboard.writeText('{{ evaluation.full_name }}')">
                     {% if evaluation_form.instance.id %}
                         {{ evaluation.full_name }} ({{ evaluation.course.semester.name }})
                     {% else %}
                         {% translate 'Create evaluation' %}
                     {% endif %}
-                </a>
+                </span>
             </h3>
             {% if evaluation_form.instance.id %}
                 <div class="mb-2">
--- a/evap/staff/templates/staff_evaluation_form.html
+++ b/evap/staff/templates/staff_evaluation_form.html
@@ -12,11 +12,13 @@
                 </div>
             {% endif %}
             <h3 class="pt-2">
-                {% if evaluation_form.instance.id %}
-                    {{ evaluation.full_name }} ({{ evaluation.course.semester.name }})
-                {% else %}
-                    {% translate 'Create evaluation' %}
-                {% endif %}
+                <a onclick="navigator.clipboard.writeText('{{ evaluation.full_name }}')">
+                    {% if evaluation_form.instance.id %}
+                        {{ evaluation.full_name }} ({{ evaluation.course.semester.name }})
+                    {% else %}
+                        {% translate 'Create evaluation' %}
+                    {% endif %}
+                </a>
             </h3>
             {% if evaluation_form.instance.id %}
                 <div class="mb-2">
--- a/evap/staff/templates/staff_evaluation_form.html
+++ b/evap/staff/templates/staff_evaluation_form.html
@@ -12,7 +12,7 @@
                 </div>
             {% endif %}
             <h3 class="pt-2">
-                <a onclick="navigator.clipboard.writeText('{{ evaluation.full_name }}')">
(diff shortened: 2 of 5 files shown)


evap/staff/templates/staff_course_form.html
```diff
@@ -6,7 +6,9 @@
     {{ block.super }}
     <h3>
         {% if course_form.instance.id %}
-            {{ course.name }} ({{ course.semester.name }})
+            <a data-bs-toggle="tooltip" data-bs-placement="right" title="{% translate 'Click to copy name' %}" onclick="navigator.clipboard.writeText('{{ course.name }}')">
```
niklasmohrin, line 9: Why is this a `<a>` tag? It doesn't link to anything, so a `<span>` should probably be fine too?

niklasmohrin, line 9: hm, the `onclick=` is a bit hard to read. I think it would be nicer if the code here looked like `data-copy-on-click="{{ course.name }}"` and we would add a corresponding click handler to all elements that have this attribute in, for example, base-template.ts

niklasmohrin (changes requested):

evap/staff/templates/staff_course_form.html
richardebeling: We also need this to eventually enable [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP) in #2296.

ZoqkMaze: Is this what you imagined (except the failing frontend tests)?

richardebeling: Yes, the approach matches what we meant, thanks (failing frontend tests is what our nightmares are made of 😅-- the failure here looks like it may be unrelated to your changes)

evap/static/ts/src/base-template.ts
```diff
@@ -11,3 +11,24 @@ document.addEventListener("shown.bs.modal", e => {
         autofocusEl.focus();
     }
 });
+
+document.addEventListener("click", e => {
+    if (!e.target) {
+        return;
+    }
+    const modalEl = e.target as HTMLElement;
```
richardebeling: `modalEl` is used above because the HTMLElement there is a [bootstrap modal dialog](https://getbootstrap.com/docs/5.3/components/modal/). In this case here, we're just looking for any element emitting a "click" element, so `modalEl` is not really appropriate. Let's just call it `element`

evap/static/ts/src/base-template.ts
```diff
@@ -11,3 +11,24 @@ document.addEventListener("shown.bs.modal", e => {
...
+    if (copy_string) {
+        navigator.clipboard.writeText(copy_string);
+        // console.log("copied");
+        // console.log(modalEl.getAttributeNames());
+        // console.log(modalEl.getAttribute("data-bs-original-title"))
+        // if (modalEl.hasAttribute("data-bs-original-title")) {
+        //     modalEl.setAttribute("data-bs-original-title", "Copied to clipboard");
+        //     console.log("data-bs-original-title");
+        // }
+    }
```
richardebeling: There's a simpler way to access `data-` attributes, using [`.dataset`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset). We can just do something like:

```suggestion
    if ("copyOnClick" in element.dataset) {
        copyToClipboard(element.dataset.copyOnClick);
    }
```

with `copyToClipboard` being defined in [evap/static/ts/src/copy-to-clipboard.ts](https://github.com/e-valuation/EvaP/blob/1ec1427299cfa56e226c0c332a2a104c121073a3/evap/static/ts/src/copy-to-clipboard.ts#L1), where we can then also use it in `copyHeaders`.

(Note that there is [eternal async pain](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/) waiting for us here since there is [no synchronous interface](https://w3c.github.io/clipboard-apis/#dom-clipboard-writetext) to copy something to the clipboard, so we won't know if the operation succeeds or fails. I don't know if there's anything we want to do if the operation fails. It may be fine to just ignore the returned promise, but we may need to be explicit about it to satisfy some linting check, because forgetting to check a returned promise is often a bug)

richardebeling (commented):
Looking generally good, thanks. We'll have to work around a few of the quirks of accessing the clipboard, but it should be manageable

evap/staff/templates/staff_course_form.html
richardebeling, line 1: (random file)

The frontend test failure was flaky. For reference: [The failed run](https://github.com/e-valuation/EvaP/actions/runs/25018369985/job/73272139693?pr=2697) failed with:
```
======================================================================
ERROR: setUpClass (evap.contributor.tests.test_live.ContributorDelegationLiveTest)
----------------------------------------------------------------------
  File "/nix/store/829wb290i87wngxlh404klwxql5v18p4-python3-3.13.7/lib/python3.13/unittest/suite.py", line 166, in _handleClassSetUp
    setUpClass()
    ^^^
  File "/home/runner/work/EvaP/EvaP/evap/evaluation/tests/tools.py", line 402, in setUpClass
    cls.selenium.set_window_size(*cls.window_size)
    ^^^^^^^^^^^^^^^
  File "/nix/store/1bx8jycx2qc53www9mnwkjxidyp1wa4i-evap-dev-env/lib/python3.13/site-packages/selenium/webdriver/remote/webdriver.py", line 928, in set_window_size
    self.set_window_rect(width=int(width), height=int(height))
    ^^^^^^^^^^^^^^^
  File "/nix/store/1bx8jycx2qc53www9mnwkjxidyp1wa4i-evap-dev-env/lib/python3.13/site-packages/selenium/webdriver/remote/webdriver.py", line 1000, in set_window_rect
    return self.execute(Command.SET_WINDOW_RECT, {"x": x, "y": y, "width": width, "height": height})["value"]
    ^^^^^^^^^^^^^^^
  File "/nix/store/1bx8jycx2qc53www9mnwkjxidyp1wa4i-evap-dev-env/lib/python3.13/site-packages/selenium/webdriver/remote/webdriver.py", line 446, in execute
    self.error_handler.check_response(response)
    ^^^^^^^^^^^
  File "/nix/store/1bx8jycx2qc53www9mnwkjxidyp1wa4i-evap-dev-env/lib/python3.13/site-packages/selenium/webdriver/remote/errorhandler.py", line 232, in check_response
    raise exception_class(message, screen, stacktrace)
    ^^^^^^^^^^^^^^^
selenium.common.exceptions.WebDriverException: Message: Failed to decode response from marionette
```

which I don't find very helpful. The internet hints that this may be a memory availability issue. We recently started [running the tests in parallel](https://github.com/e-valuation/EvaP/pull/2666/changes#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR136), maybe spanning 4 instances of firefox will exhaust main memory on the GitHub Actions runner.

---

It passed now, you don't have to worry about it in this PR.

evap/static/ts/src/base-template.ts
niklasmohrin: I think as long as an error with writing to the clipboard prints the error to the console it's fine; anything we do won't be more helpful than just looking at what the browser said went wrong

The only thing that I can see in the linked spec document is missing permissions, we can try out this error case by forbidding clipboard access manually

niklasmohrin: To that end, I think that uncaught promise errors are printed to the console? If not, we can catch errors and call `console.error` ourselves

evap/static/ts/src/base-template.ts
```diff
@@ -11,3 +11,24 @@ document.addEventListener("shown.bs.modal", e => {
         autofocusEl.focus();
     }
 });
+
+document.addEventListener("click", e => {
+    if (!e.target) {
```
niklasmohrin: Usually we add such event handlers on the elements themselves instead of the document, like
```javascript
for (const element of document.querySelectorAll("[data-copy-on-click]")) {
    element.addEventListener("click", () => {
        foobar(element.dataset.copyOnClick);
    });
}
```
This has the advantage that not all click on the whole page go through this click handler, and that we don't have to think about event bubbling at all since we use `element` from the outer scope instead of `event.target`.

Was the global event handler an intended choice here? If so, what are the reasons?

richardebeling: I considered this as well and thought that it may be beneficial to have a single global click handler whenever we have dynamically generated DOM elements, i.e., in modals, in bootstrap carousels, or in formsets. If we add a new button with `data-copy-on-click` after page load, the "global" click handler just works without hassling about re-registration

Given that this can bail out very fast with "nothing to do here", I don't think running it for clicks on other elements is problematic. Bubbling, I think, also shouldn't be problematic: Having an intended-to-be-clicked element inside another intended-to-be-clicked element shouldn't happen with sane UI designs.

Consistency with existing code is a valid argument though, so if we don't need this (which I think we currently don't), both approaches are fine with me.

richardebeling (approved):

ZoqkMaze:
Currently, there is no feedback when copying names to the clipboard (especially when clicking on the heading). The existing copy functionality (email templates) also gives no feedback, but at least there is a mouse pointer icon. Should there be some kind of feedback and should the existing code be rewritten using data-copy-on-click?

evap/staff/templates/staff_evaluation_form.html
```diff
@@ -12,11 +12,13 @@
...
-                {% if evaluation_form.instance.id %}
-                    {{ evaluation.full_name }} ({{ evaluation.course.semester.name }})
-                {% else %}
-                    {% translate 'Create evaluation' %}
-                {% endif %}
+                <span class="btn-clipboard" data-bs-toggle="tooltip" data-bs-placement="right" title="{% translate 'Click to copy name' %}" data-copy-on-click="{{ evaluation.full_name }}">
+                    {% if evaluation_form.instance.id %}
+                        {{ evaluation.full_name }} ({{ evaluation.course.semester.name }})
+                    {% else %}
+                        {% translate 'Create evaluation' %}
```
richardebeling: Do we want to add the "click to copy"-thingy also for the "Create evaluation" text?

richardebeling (approved):

evap/staff/templates/staff_evaluation_form.html
```diff
@@ -12,11 +12,13 @@
                 </div>
             {% endif %}
             <h3 class="pt-2">
-                {% if evaluation_form.instance.id %}
-                    {{ evaluation.full_name }} ({{ evaluation.course.semester.name }})
-                {% else %}
-                    {% translate 'Create evaluation' %}
-                {% endif %}
+                <span class="btn-clipboard" data-bs-toggle="tooltip" data-bs-placement="right" title="{% translate 'Click to copy name' %}" data-copy-on-click="{{ evaluation.full_name }}">
```
niklasmohrin: The copy-on-click makes no sense in the case that there is no evaluation, then it just says "Creation evaluation"

niklasmohrin (changes requested):
Almost everything looks good

richardebeling (approved):

evap/staff/templates/staff_evaluation_form.html
```diff
@@ -13,7 +13,9 @@
             {% endif %}
             <h3 class="pt-2">
                 {% if evaluation_form.instance.id %}
-                    {{ evaluation.full_name }} ({{ evaluation.course.semester.name }})
+                    <span class="btn-clipboard" data-bs-toggle="tooltip" data-bs-placement="right" title="{% translate 'Click to copy name' %}" data-copy-on-click="{{ evaluation.full_name }}">
```
niklasmohrin: what does btn-clipboard do here? It doesn't seem to have an effect

ZoqkMaze: it was an artifact and should be removed

niklasmohrin (approved):

janno42 (approved):
📋

PR #2700, opened 2026-04-13
Student vote page: red border for erroneous rows is missing #2692

### LINKED ISSUE #2692: Student vote page: red border for erroneous rows is missing
Go to any student vote page, e.g. http://localhost:8000/student/vote/1673, and submit the form with a missing answer. The answer button row will look like this:
Image

but it is supposed to look like this:
Image

Note that the red border coloring is currently missing. We have codified the red border color in CSS here:
https://github.com/e-valuation/EvaP/blob/865aeb586e9ff8bd66d5ee93054d80fa2a097532/evap/static/scss/components/_button-group.scss#L17-L19

but the vote-btn background-color of medium gray is more important due to the `!important` annotation:
https://github.com/e-valuation/EvaP/blob/865aeb586e9ff8bd66d5ee93054d80fa2a097532/evap/static/scss/components/_buttons.scss#L193

anjaton:
Fixes #2692

### DIFF
--- a/evap/static/scss/components/_button-group.scss
+++ b/evap/static/scss/components/_button-group.scss
@@ -3,7 +3,7 @@ div.vote-inputs {
     flex-wrap: wrap;
 }
 
-div.vote-inputs > textarea {
+div.vote-inputs>textarea {
     width: 100%;
     min-height: 8rem;
 }
--- a/evap/static/scss/components/_button-group.scss
+++ b/evap/static/scss/components/_button-group.scss
@@ -3,7 +3,7 @@ div.vote-inputs {
     flex-wrap: wrap;
 }
 
-div.vote-inputs > textarea {
+div.vote-inputs>textarea {
     width: 100%;
     min-height: 8rem;
 }
@@ -15,5 +15,5 @@ div.vote-inputs > textarea {
 }
 
 div.vote-inputs label.choice-error {
-    border-color: $evap-red;
+    border-color: $evap-red !important;
 }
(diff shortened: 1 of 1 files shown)


evap/static/scss/components/_button-group.scss
```diff
@@ -3,7 +3,7 @@ div.vote-inputs {
     flex-wrap: wrap;
 }
 
-div.vote-inputs > textarea {
+div.vote-inputs>textarea {
```
richardebeling: Please revert the whitespace changes here

evap/static/scss/components/_button-group.scss
```diff
@@ -15,5 +15,5 @@ div.vote-inputs > textarea {
 }
 
 div.vote-inputs label.choice-error {
-    border-color: $evap-red;
-}
+    border-color: $evap-red !important;
+}
```
richardebeling, line 19: You removed a trailing newline character here. Please add it back

richardebeling (commented):
Looks mostly good, thanks! There's two small white space changes, can you revert them?

evap/static/scss/components/_button-group.scss
richardebeling: The line terminator for the last line is still missing. For details on why we want it, see https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline.

I would guess that your editor automatically removed the last newline character. If you're using VSCode, check that it doesn't have `trimFinalNewlines` enabled.

janno42 (approved):
Thanks! Good first PR :)
🔳

richardebeling (approved):

PR #2701, opened 2026-04-13
Run `./manage.py run`, TS, SCSS in process-compose; README changes

niklasmohrin:
See the README change. I kept the old services-full name around so that we don't break setups in the transition period.

If any of these services should be stopped or restarted, this can be done in the process-compose TUI with the keyboard shortcuts listed at the bottom.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ To exit the development shell, press `Ctrl-D` or type `exit`.
 
 If you start your code editor from the `nix develop` shell, it should automatically pick up all dependencies. If this does not work automatically, try using the `nix/nix-python` script in the EvaP project as the Python interpreter in your IDE.
 
-After quitting the services, you can run the command `nix run .#clean-setup` to remove persistent state (database, node modules, localsettings).
+After quitting `nix run`, you can run the command `nix run .#clean-setup` to remove persistent state (database, node modules, localsettings).
 Afterwards, `nix run` will recreate everything when you run it the next time.
 
 Before committing, run `./manage.py precommit` or alternatively, the individual commands:
--- a/README.md
+++ b/README.md
@@ -14,12 +14,12 @@ For the documentation, please see our [wiki](https://github.com/e-valuation/EvaP
 
 ## Development Setup
 
-We use [nix](https://nixos.org/) to manage the development environment.
+We use [nix](https://nixos.org/) to manage the development environment. To get a local version of EvaP running, follow these steps:
 
 1. Windows only: Install the Windows Subsystem for Linux (WSL) using `wsl --install -d Ubuntu-24.04` (you may have to restart your computer and run this command again). Enter the WSL environment using the `wsl` command. On your first entry, you need to choose a username and password - anything works (for example: username "evap", password "evap"). Perform the next step outside of `/mnt`, for example by going to your home directory (`cd ~`).
 2. Install [git](https://git-scm.com/downloads). Run the following commands to clone and enter the EvaP repository:
    ```bash
-   git clone --recurse-submodules https://github.com/e-valuation/EvaP.git
+   git clone https://github.com/e-valuation/EvaP.git
    cd EvaP
    ```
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -18,28 +18,28 @@ We use [nix](https://nixos.org/) to manage the development environment.
 
 1. Windows only: Install the Windows Subsystem for Linux (WSL) using `wsl --install -d Ubuntu-24.04` (you may have to restart your computer and run this command again). Enter the WSL environment using the `wsl` command. On your first entry, you need to choose a username and password - anything works (for example: username "evap", password "evap"). Perform the next step outside of `/mnt`, for example by going to your home directory (`cd ~`).
 2. Install [git](https://git-scm.com/downloads). Run the following commands to clone and enter the EvaP repository:
-   ```
+   ```bash
    git clone --recurse-submodules https://github.com/e-valuation/EvaP.git
```
richardebeling: We can also remove the `--recurse-submodules` here, our init script will do `git submodule update --initialize`.

README.md
```diff
@@ -18,28 +18,28 @@ We use [nix](https://nixos.org/) to manage the development environment.
...
-   ./manage.py run
+3. On Linux and WSL, install nix by running `./nix/setup-nix`. On MacOS, install nix using the [Determinate Nix Installer](https://install.determinate.systems/). Afterwards, if you get any errors when running nix, restart your computer.
+4. Start EvaP and wait until you see a table view and the "evap" row shows "Running":
+   ```bash
+   nix run
    ```
-6. Open your web browser at http://localhost:8000/ and login with email `evap@institution.example.com` and password `evap`.
+5. Open your web browser at http://localhost:8000/ and login with email `evap@institution.example.com` and password `evap`.
+
+When developing EvaP, you can open a shell with all dependencies available:
```
richardebeling: Should we maybe add something that gives a bit more context on `nix run`, to allow people to get up-to-speed on what's going on quicker?

> Note that `nix run` opens an orchestrator for the components required for development, such as a database, an in-memory cache, and the processes actually processing our code: A django development server, an scss compiler and a typescript compiler. The output of these three will contain information about errors in the code, and are thus a good place to start debugging if something doesn't work.

richardebeling (approved):
Good stuff, thanks

nix/services.nix
```diff
@@ -78,4 +78,29 @@
...
+          depends_on = {
+            "init-django".condition = "process_completed_successfully";
+            "npm-ci".condition = "process_completed_successfully";
+          };
+        };
+      in
+      {
+        ts = make-dev-command "ts" "./manage.py ts compile --fresh --watch";
+        scss = make-dev-command "scss" "./manage.py scss --watch";
+        evap = make-dev-command "evap" "./manage.py run";
```
niklasmohrin, line 104: @richardebeling The reloading of Python code does not work as I expect, if I add a print statement to the student index view then for some reason these prints only show up the _next_ time I change a Python file. The normal HTTP logs appear immediately, but my print seems to be buffered somehow, perhaps because stdout of the Django server is now piped to process-compose and not to a pty?

richardebeling: Ah, yeah, perhaps it thinks it's writing to a file and buffers for more than a line or so. Maybe we can disable or reduce output buffering. I'll try to look into it some time.

Do we consider this blocking? For debugging, I would advice people to run shutdown the server in process-componse and run `manage.py run` in a separate shell anyway, and then interact with it through gdb.

richardebeling: Adding `runtimeEnv = { PYTHONUNBUFFERED = "1"; };` to `writeShellApplication` seems to do the trick for me on Windows with WSL

niklasmohrin: Actually I also found something else now: Process compose has a `is_tty` setting which we can just set to true

richardebeling: 🤯

niklasmohrin:
@janno42 Please take a look at the readme changes

richardebeling (approved):

janno42 (approved):

PR #2702, opened 2026-04-15
Stop recommending the Determinate Nix installer; eagerly install nix config

niklasmohrin:
The Determinate Nix installer is no longer the Determinate (Nix Installer) but rather the (Determinate Nix) Installer, that is, it installs the proprietary version of nix built by Determinate Systems. I don't think there is anything wrong with their nix build, but I don't want to recommend installing proprietary software.

One benefit of the Determinate installer is that it had some sane default settings (the same that we include in our Linux installation script). With this PR, we now let the nix install script set up these settings even if installing nix does not work. This way, for Mac users the process is "run installer", "installer installs config", "installer says no package manager found and suggests manual download", "user does manual download". I think this is pretty okay. If the Mac users complain, we can add support for homebrew or so to our installer script.

### DIFF
--- a/nix/setup-nix
+++ b/nix/setup-nix
@@ -2,16 +2,32 @@
 
 set -e
 
+XDG_CONFIG_HOME=${XDG_CONFIG_HOME:-$HOME/.config}
+NIX_CONF="$XDG_CONFIG_HOME/nix/nix.conf"
+
+if test -e "$NIX_CONF"; then
+    echo "Nix config already found at $NIX_CONF"
+    echo "Please make sure it contains the following settings:"
+    echo
+    cat $(dirname $0)/nix.conf
+    echo
+else
+    echo "Installing nix config at $NIX_CONF"
+    mkdir -p "$XDG_CONFIG_HOME/nix/"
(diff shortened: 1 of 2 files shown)


niklasmohrin:
Marked as draft because I want to merge #2701 first

niklasmohrin:
Also note that existing installations of nix (on mac computers) continue to work

nix/setup-nix
```diff
@@ -2,17 +2,31 @@
...
     exit 1
 fi
 
-if apt --version &> /dev/null; then
+if apt-get --version &> /dev/null; then
     set -x
     sudo apt-get update -y
     sudo apt-get install -y nix
     sudo usermod -aG nix-users $USER
-    newgrp nix-users
```
richardebeling, line 15: `newgrp nix-users` worked fine for me here on ubuntu 25.10 and ubuntu 24.04 in WSL. I wouldn't remove it, it doesn't do harm.

richardebeling (approved):

nix/setup-nix
niklasmohrin: Hm, interesting, I thought that on Monday there was someone for whom this did not work. Sure, we can also keep it

PR #2704, opened 2026-04-20
Add dynamic CSS "empty" disclaimer to all datagrids

### LINKED ISSUE #2662: Use CSS for "no matching entries" tables
We have a number of tables that are searchable (using `DataGrid`, `TableGrid`), for example in `grades_semester_view`. On other pages with lists generated in templates, we sometimes have an extra "empty" check that adds a disclaimer "no results found" or something similar.

Instead, we should add such disclaimers to searchable tables dynamically in the frontend so that it works both when there are no results initially as well as when there are no results because of a search. I feel that this should work by having an extra div with class `empty-disclaimer` that is hidden by default but shown by a rule like the following:
```css
table:not(:has(:not(.empty-disclaimer))) .empty-disclaimer {
    display: block;
}
```

Maybe it is easier to do it the other way around and only hide it if there are elements like
```css
table:has(:not(.empty-disclaimer)) .empty-disclaimer {
    display: none;
}
```

niklasmohrin:
Closes #2662

I made two separate commits, the first has the meaningful changes and the second fixes the indentation.

### DIFF
--- a/evap/evaluation/templates/profile.html
+++ b/evap/evaluation/templates/profile.html
@@ -44,6 +44,7 @@
                                         </td>
                                     </tr>
                                 {% endif %}
+                                <tr>
                                     <td>
                                         <strong>{% translate 'CC all my emails to' %}</strong><br />
                                         {% for cc_user in cc_users %}
--- a/evap/staff/templates/staff_questionnaire_index.html
+++ b/evap/staff/templates/staff_questionnaire_index.html
@@ -34,7 +34,6 @@
         </div>
     </div>
 
-    {% if general_questionnaires_top or general_questionnaires_bottom or contributor_questionnaires %}
         <form id="questionnaire-deletion-form" method="POST" action="{% url 'staff:questionnaire_delete' %}">
             {% csrf_token %}
         </form>
@@ -99,12 +98,6 @@
                 </div>
             </div>
         </div>
-
-    {% else %}
-        <p>
-            {% translate 'There are no questionnaires yet.' %}
-        </p>
-    {% endif %}
 {% endblock %}
 
 {% block additional_javascript %}
--- a/evap/staff/templates/staff_questionnaire_index.html
+++ b/evap/staff/templates/staff_questionnaire_index.html
@@ -54,7 +54,7 @@
                 fadeOutThenRemove(document.querySelector(`#questionnaire-row-${body.get("questionnaire_id")}`));
             }).catch(error => {
                 window.alert("{% translate 'The server is not responding.' %}");
-            });;
+            });
         });
     </script>
 
--- a/evap/staff/templates/staff_questionnaire_index.html
+++ b/evap/staff/templates/staff_questionnaire_index.html
@@ -34,70 +34,70 @@
         </div>
     </div>
 
-        <form id="questionnaire-deletion-form" method="POST" action="{% url 'staff:questionnaire_delete' %}">
-            {% csrf_token %}
-        </form>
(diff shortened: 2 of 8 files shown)


janno42 (approved):
I can see the 🎆

evap/staff/templates/staff_questionnaire_index.html
```diff
@@ -31,44 +31,38 @@
...
+            const request = fetch(
+                form.action,
+                { method: form.method, body },
+            );
+            request.then(response => {
+                assert(response.ok);
+                fadeOutThenRemove(document.querySelector(`#questionnaire-row-${body.get("questionnaire_id")}`));
+            }).catch(error => {
+                window.alert("{% translate 'The server is not responding.' %}");
+            });;
```
fekoch:
```suggestion
            });
```

evap/evaluation/templates/profile.html
```diff
@@ -44,6 +44,7 @@ <h4 class="card-title mb-4">{% translate 'Delegates and emails' %}</h4>
                                         </td>
                                     </tr>
                                 {% endif %}
+                                <tr>
```
richardebeling, line 47: hah

fekoch (changes requested):

richardebeling (approved):

niklasmohrin:
Waiting for #2587 to spare them the rebase

niklasmohrin:
I think we are running into a firefox bug on the questionnaire index. The first time (per tab) that I filter for something, if it matches nothing (i.e. you have to enter something like adljaladljad right away without a table update from a prefix), the disclaimer does not show up. The disclaimer shows up if:
1. this is not the first time that the table contents change
2. the table is initially empty (for example reload page with something in search box)

It does work again if you remove the filter, switch to another questionnaire tab, and enter alkjhdf again. Very strange. In chromium everything works as expected

niklasmohrin:
Oh, I can also get it to break in other places. Semester view, enter something that doesn't match anything (evaluation or course list, doesn't matter). Reload page. Empty table, no disclaimer. Funnily enough, this is exactly the opposite of when it breaks on the questionnaire page

niklasmohrin:
Probably firefox is tripping up with our table contents remove and append here https://github.com/e-valuation/EvaP/blob/88074281169c81c7e8407c67ad5cffb957a49b18/evap/static/ts/src/datagrid.ts#L211-L216 and fails to apply CSS rules?

fekoch:
> Oh, I can also get it to break in other places. Semester view, enter something that doesn't match anything (evaluation or course list, doesn't matter). Reload page. Empty table, no disclaimer. Funnily enough, this is exactly the opposite of when it breaks on the questionnaire page

Interesting, I can reproduce it, but not consistently. When reloading multiple times, the notice sometimes shows and sometimes is missing.

Additionally, I can not reproduce it, if I have my devtools open :upside_down_face: ...

fekoch:
I opened a ticket at bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=2035307

niklasmohrin:
What do we do with this in the meantime? I guess the new disclaimer always shows up when the old one would have, that is, if there are no items to begin with. Do we still want to merge it then, or do we think with the FF bug it would be distracting? Personally I don't think that one would usually notice

richardebeling:
Yeah I think we can merge already and just get a better feature automagically when Firefox fixes their CSS logic

niklasmohrin:
(we talked in person)

PR #2707, opened 2026-04-20
Fix misplaced form error tooltip

### LINKED ISSUE #2646: Misplaced form error tooltip
When a mandatory form field is left empty, the browser displays an error tooltip next to the field. TomSelect seems to cause strange positioning, as seen in this example on the staff course creation page:

_Firefox_
Image

_Chrome_
Image

Instead, the message should appear next to the respective field (here, the "Responsibles" field), similar to how the message appears for text fields.
Image

geromequa:
Closes #2646. Removed height attribute was introduced [here](https://github.com/e-valuation/EvaP/blob/4e07a0b4c61e0468513a6d0c033ddf9b3664c5e0/evap/static/css/evap.css).

### DIFF
--- a/evap/static/scss/components/_forms.scss
+++ b/evap/static/scss/components/_forms.scss
@@ -10,12 +10,9 @@
     }
 }
 
-select[multiple='multiple'] {
-    height: 0px;
-}
-
-select[multiple] {
-    min-height: 40px;
+// place tooltip for tomselect based on form-element height
+.tomselected {
+    height: 38px;
 }
 
 .email-form {
--- a/evap/static/scss/components/_forms.scss
+++ b/evap/static/scss/components/_forms.scss
@@ -11,7 +11,7 @@
 }
 
 select[multiple='multiple'] {
-    height: 200px;
+    height: 0px;
 }
 
 select[multiple] {
--- a/evap/static/scss/components/_forms.scss
+++ b/evap/static/scss/components/_forms.scss
@@ -10,7 +10,7 @@
     }
 }
 
-// place tooltip for tomselect based on form-element height
+// place tooltip for tomselect based on form-element height. See https://github.com/orchidjs/tom-select/issues/1038
 .tomselected {
     height: 38px;
 }


evap/static/scss/components/_forms.scss
richardebeling, line 19: Should we just remove these two? It feels like this doesn't make sense anymore.

geromequa: When removing this we get a distance of 40px for the tooltip, i.e., the min-height set in line 18. This is not the behavior we want. When also removing the min-height, we still have a gap between tooltip and form element.

janno42: height 0 together with min-height 40px is weird though. wouldn't height 40px make more sense? and maybe we can only define this for `.tomselected` elements?

evap/static/scss/components/_forms.scss
```diff
@@ -10,12 +10,9 @@
...
-select[multiple='multiple'] {
-    height: 200px;
-}
-
-select[multiple] {
-    min-height: 40px;
+// place tooltip for tomselect based on form-element height
+.tomselected {
+    height: 38px;
 }
```
richardebeling: This works for me for now, but I think we're working around a TomSelect bug here: The browser, understandably, renders the information popup relative the the `select` HTML element that has the `required` attribute. Tomselect [makes](https://github.com/orchidjs/tom-select/blob/3db44e7453a4d6276dbaff04bafd3f2149c6797f/src/scss/tom-select.scss#L172) this select element 0x0-sized and invisible. They should do something to make the browser tooltip show at the correct position, we shouldn't have to fix this.

It is also broken on their example page: https://tom-select.js.org/examples/validation/

Can we file an issue with them about this, and link it here, to make it clear that this is a workaround for a tomselect bug, not actual styling we want to apply?

(Side note @janno42: What do we think about their bootstrap form error display mechanism for form validation, do we like it? See "Bootstrap" [here](https://tom-select.js.org/examples/validation/))

richardebeling (approved):

evap/static/scss/components/_forms.scss
geromequa: https://github.com/orchidjs/tom-select/issues/1038

richardebeling: Nice. Can we add a short comment in code referencing this issue?

janno42: The bootstrap error display would be even nicer - because it would be consistent with the other fields.

geromequa: I have created #2737, so that this can be implemented in future.

janno42 (approved):
looks good :)

PR #2708, opened 2026-04-20
Average grades in export [REBORN]

### LINKED ISSUE #2001: Add averages to results exports

Belissimo-T:
closes #2001


evap/results/tests/test_exporters.py
```diff
@@ -606,3 +606,73 @@ def test_text_answer_export(self):
...
+        cache_results(evaluation_2)
+
+        binary_content = BytesIO()
+        ResultsExporter().export(
+            binary_content,
+            [evaluation_1.course.semester, evaluation_2.course.semester],
+            [
+                (
+                    [course_program.id for course_program in evaluation_1.course.programs.all()]
+                    + [course_program.id for course_program in evaluation_2.course.programs.all()],
```
richardebeling: This is just `[program]`? (or, if we want to test the duplication, `[program, program]`)

evap/results/tests/test_exporters.py
```diff
@@ -606,3 +606,73 @@ def test_text_answer_export(self):
...
+                    [evaluation_1.course.type.id, evaluation_2.course.type.id],
+                )
+            ],
+            True,
+            True,
+        )
+        binary_content.seek(0)
+        workbook = xlrd.open_workbook(file_contents=binary_content.read())
+
+        self.assertAlmostEqual(workbook.sheets()[0].row_values(5)[1], 1.5833333333333335)
```
richardebeling: Please add a short comment explaining where this number comes from

evap/results/tests/test_exporters.py
```diff
@@ -606,3 +606,73 @@ def test_text_answer_export(self):
...
+
+        self.assertEqual(
+            workbook.sheets()[0].row_values(0)[1],
+            "Average result for this question over all published evaluations in all semesters",
+        )
+
+        # average for second questionnaire must be the value from evaluation2 (since evaluation1 doesn't have the questionnaire)
+        self.assertEqual(float(workbook.sheets()[0].row_values(8)[1]), float(workbook.sheets()[0].row_values(8)[3]))
+        # average field next to the questionnaire title must be empty
+        self.assertEqual("", workbook.sheets()[0].row_values(8)[2])
```
richardebeling: The discussions from https://github.com/e-valuation/EvaP/pull/2366#discussion_r1987979850 still apply

evap/results/tools.py
```diff
@@ -121,11 +121,15 @@ def approval_count(self) -> int:
...
-    def average(self) -> float:
+    def grade_average(self) -> float:
         return (
             sum(grade * count for count, grade in zip(self.counts, self.choices.grades, strict=True)) / self.count_sum
         )
 
+    @property
+    def approval_average(self) -> float:
+        return self.approval_count / self.count_sum
+
```
richardebeling, line 132: Is https://github.com/e-valuation/EvaP/wiki/Calculation-of-Results still up to date, or do we need to update anything there?

evap/results/exporters.py
```diff
@@ -271,58 +308,112 @@ def write_overall_results(
...
+                continue
+
+            average_grade, average_approval = cls._get_average_grade_and_approval(
+                question, results_by_questionnaire_id[questionnaire_id]
+            )
+
+            average_grade_averager.record_value(average_grade)
+
+            if question.is_yes_no_question:
+                average_approval_averager.record_value(typing.cast("float", average_approval))
```
richardebeling: Does this need a `cast`, or would `assert is not None` also do it?

evap/results/tools.py
Belissimo-T: Yes, I think so.

evap/results/tests/test_exporters.py
```diff
@@ -606,3 +607,172 @@ def test_text_answer_export(self):
...
+            course__type=course_type,
+            course__programs=[program],
+            course__semester=semester,
+            state=Evaluation.State.PUBLISHED,
+            _participant_count=6,
+            _voter_count=6,
+        )
+        evaluation.general_contribution.questionnaires.set([questionnaire])
+        make_rating_answer_counters(question_assignment, evaluation.general_contribution, [2, 0, 0, 0, 0])
+        cache_results(evaluation)
```
richardebeling: this block is repeated 7 times here, 13 lines each time -- can we maybe make a helper function, and bring it down to just one line (pass the rating answer counters and every baker.make argument that is _not_ the default)?

We should be able to get this test to fit on one screen page, and resolve the disable=too-many-locals pylint ignore, there isn't that much going on here.

evap/results/tests/test_exporters.py
```diff
@@ -606,3 +607,172 @@ def test_text_answer_export(self):
...
+        binary_content = BytesIO()
+        ResultsExporter().export(
+            binary_content,
+            semesters=[semester],
+            selection_list=[([program.id], [course_type.id])],
+            include_not_enough_voters=True,
+            include_unpublished=True,
+        )
+        binary_content.seek(0)
+        workbook = xlrd.open_workbook(file_contents=binary_content.read())
```
richardebeling: Any evaluations you create after this will be trivially not included in the export, because they didn't exist at the time of the export. If we want to assert that the data we create below does not influence the exporter result, we have to create it before running the export.

evap/results/exporters.py
```diff
@@ -1,8 +1,8 @@
-import warnings
-from collections import OrderedDict, defaultdict
+import math
+import typing
+from collections import defaultdict
 from collections.abc import Iterable, Sequence
 from itertools import chain, repeat
-from typing import Any, TypeVar
```
richardebeling, line 5: My vote would be to keep using `Any` and `TypeVar`, I don't see an advantage in having to spell out `typing.` in every usage

evap/results/exporters.py
```diff
@@ -75,23 +98,17 @@ def init_grade_styles(cls) -> None:
...
-            )
-            return
+        if cls.COLOR_MAPPINGS:
+            raise RuntimeError("ResultsExporter.init_grade_styles has been called twice.")
 
-        grade_base_style = "pattern: pattern solid, fore_colour {}; alignment: horiz centre; font: bold on; borders: left medium, right medium"
+        grade_base_style = (
+            "pattern: pattern solid, fore_colour {}; alignment: horiz centre; font: bold on; "
+            "borders: left medium, right medium"
+        )
```
richardebeling: where does this change come from? The string contents are identical, right?

evap/results/exporters.py
```diff
@@ -271,58 +308,113 @@ def write_overall_results(
...
 
+            grade_average, approval_average = self._get_average_of_average_grade_and_approval(
+                all_evaluations_with_results, questionnaire.id, question
+            )
+            if math.isnan(grade_average) or math.isnan(approval_average or 0):
+                self.write_cell("", "border_left_right")
+            elif question.is_yes_no_question:
+                self.write_cell(f"{approval_average:.0%}", self.grade_to_style(grade_average))
+            else:
+                self.write_cell(grade_average, self.grade_to_style(grade_average))
```
richardebeling: This logic is now copied two times in this function -- can we deduplicate?

evap/results/exporters.py
```diff
@@ -353,17 +443,27 @@ def export_impl(
                 include_not_enough_voters,
             )
 
+            all_evaluations_with_results, __, ___ = self.filter_evaluations(
```
richardebeling: why three underscores?

evap/results/exporters.py
```diff
@@ -353,17 +443,27 @@ def export_impl(
...
+                semesters=None,
+                evaluation_states=[Evaluation.State.PUBLISHED],
+                program_ids=None,
+                course_type_ids=None,
+                contributor=None,
+                include_not_enough_voters=False,
+            )
```
richardebeling: @janno42 this now fetches all results from all published evaluations from all semesters, this will probably be human-noticeable slow, I don't know if it will be cannot-be-used slow.

Prior discussion in https://github.com/e-valuation/EvaP/pull/2366#discussion_r1922713988
> shouldn't take longer than ~10 seconds (with ~100 evaluations in the semester).

We're looking at all semesters here, so the "100 evaluations in the semester" only given an implicit bound.

evap/results/tests/test_exporters.py
Belissimo-T: 😭

evap/results/exporters.py
Belissimo-T:
```
evap/results/exporters.py:441: error: Incompatible types in assignment (expression has type "bool", variable has type "list[Questionnaire]")  [assignment]
```

evap/results/exporters.py
Belissimo-T: This came from my PyCharm formatter.

niklasmohrin (commented):
please re-request once the youngsters have done their reviews :^)

PR #2710, opened 2026-04-27
rename helper function

ZoqkMaze:
In #2587, there were still problems with helper function names. Is this an acceptable solution?


evap/evaluation/tools.py
```diff
@@ -180,7 +180,9 @@ def get_parameter_from_url_or_session(request: HttpRequest, parameter: str, defa
     return result
```
richardebeling: Naming was one thing, thinking about whether we can reuse one of these functions instead of implementing the logic twice was the other consideration. Semantically, retrieving the parameter as a non-string value could be thought of as retrieving it as a string value and then converting the value to the target type.

What do you think of this?

```python3
def get_bool_parameter_from_url_or_session(request: HttpRequest, parameter: str, default: bool = False) -> bool:
    return {"true": True, "false": False}.get(
        get_string_parameter_from_url_or_session(request, parameter, "").lower(),
        default
    )
```

richardebeling (approved):
thanks

niklasmohrin (approved):
what richard said

richardebeling (approved):

evap/evaluation/tools.py
```diff
@@ -170,17 +170,15 @@ def vote_end_datetime(vote_end_date: datetime.date) -> datetime.datetime:
...
+def get_bool_parameter_from_url_or_session(request: HttpRequest, parameter: str, default: bool = False) -> bool:
+    return {"true": True, "false": False}.get(
+        get_string_parameter_from_url_or_session(request, parameter, "").lower(), default
+    )
 
 
-def get_string_from_url_or_session(request: HttpRequest, parameter: str, default: str | None = None) -> str | None:
+def get_string_parameter_from_url_or_session[DefaultT: (typing.Any, None)](
+    request: HttpRequest, parameter: str, default: DefaultT = None
+) -> str | DefaultT:
```
richardebeling:
```suggestion
@overload
def get_string_parameter_from_url_or_session(request: HttpRequest, parameter: str) -> str | None: ...
@overload
def get_string_parameter_from_url_or_session(request: HttpRequest, parameter: str, default: str) -> str: ...

def get_string_parameter_from_url_or_session(request: HttpRequest, parameter: str, default=None):
```

(see https://mypy.readthedocs.io/en/stable/generics.html and https://typing.python.org/en/latest/spec/overload.html)

niklasmohrin (approved):

richardebeling (approved):

PR #2712, opened 2026-04-27
Legal Notice Link

### LINKED ISSUE #2668: Change Legal Notice Link
The `LEGAL_NOTICE_TEXT` in the settings should be changed to a `LEGAL_NOTICE_LINK` that specifies a URL.
The link on the left side in the footer should read "Legal Notice" and link to this URL. The link should have the "external link" icon behind the link text.

suttrop:
closes #2668


evap/settings.py
```diff
@@ -189,7 +189,7 @@ class ManifestStaticFilesStorageWithJsReplacement(ManifestStaticFilesStorage):
 
 CONTACT_EMAIL = "webmaster@localhost"
 ALLOW_ANONYMOUS_FEEDBACK_MESSAGES = True
-LEGAL_NOTICE_TEXT = "Objection! (this is a default setting that the administrators should change, please contact them)"
+LEGAL_NOTICE_LINK = "http://example.com/legalnotice"
```
richardebeling: Probably doesn't matter much, but if we specify a protocol (do we have to?), I would just default to using https instead

```suggestion
LEGAL_NOTICE_LINK = "https://example.com/legalnotice"
```

richardebeling (approved):
Thanks :)

niklasmohrin (approved):

PR #2714, opened 2026-04-29
prefetch participant and voter counts in contributor index

Kakadus:
relates to #1229

Directly reduces n database hits.
When testing with 200 evaluations 10 times, the accumulated test-runtime was reduced from 50s to 44s on my laptop. Even with 10 evaluations, I see a speedup (4.820s -> 3.689s).

### DIFF
--- a/evap/contributor/tests/test_views.py
+++ b/evap/contributor/tests/test_views.py
@@ -89,7 +89,6 @@ class TestContributorView(WebTestWith200Check):
 
     def test_num_queries_is_not_constant(self):
         """Current defect, see https://github.com/e-valuation/EvaP/issues/1229#issuecomment-4239035495."""
-        url = "/contributor/"
         represented = baker.make(UserProfile, email="represented@example.com")
         self.responsible.represented_users.add(represented)
         evaluations = baker.make(
@@ -108,7 +107,7 @@ class TestContributorView(WebTestWith200Check):
             _bulk_create=True,
         )
         with self.assertNumQueries(FuzzyInt(80, 100)):
-            self.app.get(url, user=self.responsible)
+            self.app.get(self.url, user=self.responsible)
 
 
 class TestContributorEvaluationView(WebTestWith200Check):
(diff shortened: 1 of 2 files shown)


richardebeling (commented):
Lgtm. Can you add a test that asserts that we have < N queries? (I would assume that, for each relation we currently `prefetch_related`, the test fails if we remove it).

evap/contributor/views.py
```diff
@@ -77,9 +77,17 @@ def index(request):
...
+        delegated_evaluations = Evaluation.annotate_with_participant_and_voter_counts(
+            Evaluation.objects.filter(course__in=delegated_courses)
+        ).prefetch_related(
+            "course",
+            "course__evaluations",
+            "course__programs",
+            "course__type",
+            "course__semester",
+            "course__responsibles",
         )
```
richardebeling: Some of these could be `select_related`s (course, course__type, course__semester). You could check whether it improves anything, if yes, we can use that

niklasmohrin (approved):

Kakadus:
finally found some time to revisit this.

> Can you add a test that asserts that we have < N queries?

No, because we do not have < N queries yet. With 100 evaluations, contributions, we have 540 queries. Without this PR, we have ~200 more. For a test, I can cut down the number if objects (say 10 each) and assert that we are not constant (currently 90 queries, so assert we make 80-100 queries). For reference, the constant PoC at #2715 has 43 queries.
That way, we see degradations and monitor the current state, although margins are quite small.

evap/contributor/tests/test_views.py
```diff
@@ -82,8 +83,32 @@ class TestContributorView(WebTestWith200Check):
 
     @classmethod
     def setUpTestData(cls):
-        users = create_evaluation_with_responsible_and_editor()
-        cls.test_users = [users["editor"], users["responsible"]]
+        result = create_evaluation_with_responsible_and_editor()
+        cls.responsible = result["responsible"]
+        cls.test_users = [result["editor"], cls.responsible]
+
+    def test_num_queries_is_not_constant(self):
```
richardebeling, line 90: haha :D

richardebeling (approved):
Nice, thanks. The fixed contributor page (via https://github.com/e-valuation/EvaP/pull/2715) is going to be epic.

niklasmohrin (approved):
Thanks!

PR #2715, opened 2026-04-29
Constant number of db queries in contributor index

Kakadus:
concerns #1229 
1. prefetch patch, discussed in PR #2714. (200x10 evaluation: 50s to 44s)
2. annotate evaluations in calculate_average_course_distribution
  - I _think_ this may even cause a significant slowdown, but does reduce the number of db queries. The problem is that `course.evaluations.all()` does not have the voter and participant counts which are fetched in `calculate_average_distribution`.
  - The patch allows callers of calculate_average_distribution to prefetch & annotate all relevant evaluations beforehand (+ some sorting magic to avoid grouping the evaluations by course). I'm sure something smarter is possible. Maybe you have some thoughts about this already?
3. Very crude implementation of the Evaluation.can_be_seen_by(x) functions as django Q objects to allow annotating those attributes on the evaluation objects. This reduced the 200x10 test time significantly to just 13s (-70%). However, the Q functions aren't readable at all.

2 & 3 are very hacky and were just results of me experimenting around in order to start a discussion.


evap/evaluation/models.py
```diff
@@ -731,6 +740,24 @@ def can_be_seen_by(self, user):
...
+
+        base_q = ~Q(state=Evaluation.State.NEW)
+        if user.is_external:
+            return base_q & Evaluation.user_is_responsible_or_contributor_or_delegate_Q(user) | Q(participants=user)
+
+        return base_q & (
+            ~Q(course__is_private=True)
+            | Evaluation.user_is_responsible_or_contributor_or_delegate_Q(user)
+            | Q(participants=user)
+        )
```
richardebeling, line 759: There is no way to evaluate a `Q` instance on a python model instance, right? If we can in any way prevent having to re-implement this logic, I would like to prefer that. Hunting after subtle semantic difference bugs between the SQL world and the python world is on my not-to-do-list for this year.

One approach that should always work is: Fetch a list of instances (e.g. evaluations visible to the user) without any prefetching (should be a few milliseconds), and then generate a list of IDs in python and pass that to follow-up queries (which now do "heavy" prefetching)

evap/contributor/templates/contributor_index.html
richardebeling, line 1: (random line): What is the struggle with the 200x10 case? We would fetch 2000 evaluation instances, and we'd have to prefetch 200 courses, ~20 course types, ~10 programs, ~400 evaluations(instances are already in memory), 1 semester, ~200 responsible. All of this are _toy numbers_. I suspect the total database query run time for the main query and all the prefetch queries is in the single to two-digit milliseconds.

None of this justifies almost a minute of runtime. Touching on the discussion in https://github.com/e-valuation/EvaP/issues/2235: For _all test_data.json_, we render _all result rows_ for _all evaluations_ (thousands) in a few seconds. There is nothing fancy in [the code for that](https://github.com/e-valuation/EvaP/blob/1ec1427299cfa56e226c0c332a2a104c121073a3/evap/results/views.py#L64). How can fetching and rendering a few hundred evaluations possibly be slower than that?

PR #2719, opened 2026-05-04
Card Header child alignment fixed

### LINKED ISSUE #2670: Card Headers: Responsive Design for Left and Right Texts
In many places, we have card headers that contain two pieces of information, one left-aligned and one right-aligned. For example, on the student index page:

The elements:
* "Evaluation rate" / "Last Evaluation: ..."
* "Summer Term 2022" / "The results of this semester have been archived"

These texts are currently fixed to left-right layout.

On small screens, we should instead of in a top-bottom layout. I'd hope there is some canonical mechanism in bootstrap to achieve this. Since this applies basically to the whole page, I would assume that we would likely want to have and use some CSS helper classes for this

Jakobus400:
fixes #2670

### DIFF
--- a/evap/staff/templates/staff_semester_preparation_reminder.html
+++ b/evap/staff/templates/staff_semester_preparation_reminder.html
@@ -63,8 +63,8 @@
     {% if interactive %}
         {% for responsible, evaluations, delegates in responsible_list %}
             <div class="card{% if not forloop.last %} mb-3{% endif %}">
-                <div class="card-header d-flex">
-                    <span class="ps-1 me-auto">
+                <div class="card-header">
+                    <span class="ps-1">
                         <a href="{% url 'staff:user_edit' responsible.id %}">{{ responsible.full_name }}</a>,
                         {% translate 'Delegates' %}:
                         {% for delegate in delegates %}
--- a/evap/staff/templates/staff_semester_view.html
+++ b/evap/staff/templates/staff_semester_view.html
@@ -162,14 +162,14 @@
         <div class="card-header d-flex">
             <a id="collapseOverviewCardButton" class="collapse-toggle collapsed" data-bs-toggle="collapse" href="#overviewCard" aria-expanded="false" aria-controls="overviewCard">{% translate 'Overview' %}</a>
             {% if request.user.is_manager %}
-                <div class="ms-auto">
+                <div class="">
                     {% if not semester.participations_are_archived %}
                         <a href="{% url 'staff:semester_preparation_reminder' semester.id %}" class="btn btn-sm btn-light">{% translate 'Preparation reminder' %}</a>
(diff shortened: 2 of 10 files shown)


evap/staff/templates/staff_semester_view.html
```diff
@@ -162,14 +162,14 @@ <h3 class="m-0 me-1">{{ semester.name }}</h3>
         <div class="card-header d-flex">
             <a id="collapseOverviewCardButton" class="collapse-toggle collapsed" data-bs-toggle="collapse" href="#overviewCard" aria-expanded="false" aria-controls="overviewCard">{% translate 'Overview' %}</a>
             {% if request.user.is_manager %}
-                <div class="ms-auto">
+                <div class="">
```
richardebeling: If we have no classes left, let's just remove the full `class=""` as well -- does the `<div>` have some purpose anymore now, or could we also remove it completely?

richardebeling (approved):

evap/staff/templates/staff_semester_view.html
```diff
@@ -162,14 +162,14 @@ <h3 class="m-0 me-1">{{ semester.name }}</h3>
         <div class="card-header d-flex">
```
janno42: The card below this one also needs fixing.

janno42: And there is another card on the student index page (for archived semesters).

janno42 (changes requested):
Ideally, the right-aligned content would stay right-aligned (in a new line when it doesn't fit in one line with the other content), and the padding between the elements would stay as it is now.

niklasmohrin:
@Jakobus400 do you want to continue working on this, or should we take over / give this issue to someone else?

evap/staff/templates/staff_semester_preparation_reminder.html
```diff
@@ -63,8 +63,8 @@ <h3>
     {% if interactive %}
         {% for responsible, evaluations, delegates in responsible_list %}
             <div class="card{% if not forloop.last %} mb-3{% endif %}">
-                <div class="card-header d-flex">
-                    <span class="ps-1 me-auto">
+                <div class="card-header">
+                    <span class="ps-1">
```
richardebeling, line 67: probably want to remove `ps-1` here as well

evap/static/scss/components/_card.scss
```diff
@@ -1,5 +1,14 @@
 .card-header {
     font-weight: bold;
+
+    // prepare for two elements, left- and right aligned, with flex wrap
+    // should have no effect with just a single child. We don't expect more than two children
```
richardebeling, line 5: update comment: first element grows -> allow n elements, first one is left, everything else is right

PR #2722, opened 2026-05-08
Optimize question migration

Kakadus:
redesign evaluation 0161 and 0162 migration so that all question objects are not stored in python.

- use bulk_create with `batch_size` 2000 on plain querysets only
- update answer objects with single update + subqueries for joined fields

followup for #2513

### DIFF
--- a/evap/evaluation/migrations/0161_temporary_newquestion_relation.py
+++ b/evap/evaluation/migrations/0161_temporary_newquestion_relation.py
@@ -2,6 +2,7 @@
 
 import django.db.models.deletion
 from django.db import migrations, models
+from django.db.models import OuterRef, Subquery
 
 
 def questions_to_question_assignments(apps, _schema_editor):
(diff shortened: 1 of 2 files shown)


evap/evaluation/migrations/0161_temporary_newquestion_relation.py
```diff
@@ -10,33 +11,58 @@ def questions_to_question_assignments(apps, _schema_editor):
...
+        # added "pk" to order_by and values for test_data migration
+        (
+            NewQuestion(**v)
+            for v in Question.objects.all()
+            .order_by("text_de", "text_en", "allows_additional_textanswers", "type")
+            .values("text_de", "text_en", "allows_additional_textanswers", "type")
+            .distinct("text_de", "text_en", "allows_additional_textanswers", "type")
+            .iterator()
+        ),
+        batch_size=2000,
```
richardebeling, line 25: Let's not do hasty changes to code for issues we don't fully understand. @janno42 do we have a specific understanding what part of this migration is problematic?

I would like to first have some test data for which we can reproduce the issue, and then properly check that a change here actually solves the issue. My mental model is that we have hundreds of questionnaires and maybe ten thousands of questions in production. This should be in the range of megabytes of data. We should easily be able to load those model instances, especially if no related models are ever loaded/accessed.

I could see that we have more answers than fit in memory, then we should do a targeted fix for that.

Kakadus:
AFAIK this is not needed anymore. Feel free to reopen if the problems of the current migrations persist

niklasmohrin:
I think we likely will want to have this, we just want to have the evidence from production first. But closing until we have it seems like the right thing for our PR list :+1:

PR #2729, opened 2026-05-13
Fix mixed query/create_defaults for Contribution in CMS import

### LINKED ISSUE #2728: JSON Importer: Contribution import error
The JSON Importer fails in `_import_contribution` if a contribution already exists (IntegrityError: duplicate key value violates unique constraint).

If a contribution already exists, it should not be modified.
If a contribution does not yet exist, it should be created with the default values.

niklasmohrin:
Fixes #2728

This is a pretty obvious mistake, but it slipped through in two PRs. I wonder how hard it would be to write a lint for this, maybe with pylint_django. I don't see how `update_or_create(**kwargs)` where `kwargs` is a strict superset of a key can ever be correct. For our codebase, this lint should also cover `update_or_create_with_changes`.

### DIFF
--- a/evap/cms/json_importer.py
+++ b/evap/cms/json_importer.py
@@ -639,10 +639,10 @@ class JSONImporter:
         if user_profile.email in settings.NON_RESPONSIBLE_USERS:
             return None, False
 
-        contribution, created = Contribution.objects.update_or_create(
+        contribution, created = Contribution.objects.get_or_create(
             evaluation=evaluation,
             contributor=user_profile,
-            create_defaults={
+            defaults={
                 "role": Contribution.Role.EDITOR,
                 "textanswer_visibility": Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS,
             },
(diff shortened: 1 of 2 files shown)


niklasmohrin:
cc @hansegucker @geromequa

evap/cms/json_importer.py
```diff
@@ -642,8 +642,10 @@ def _import_contribution(self, evaluation: Evaluation, data: ImportRelated) -> t
         contribution, created = Contribution.objects.update_or_create(
             evaluation=evaluation,
             contributor=user_profile,
-            role=Contribution.Role.EDITOR,
-            textanswer_visibility=Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS,
+            create_defaults={
+                "role": Contribution.Role.EDITOR,
+                "textanswer_visibility": Contribution.TextAnswerVisibility.GENERAL_TEXTANSWERS,
+            },
```
richardebeling: If we don't intend to update anything, my vote would be to use `get_or_create`, to make this more intuitively clear

hansegucker (approved):

richardebeling (approved):

janno42 (approved):
▶️

PR #2736, opened 2026-06-01
Add `djangofmt` HTML formatter

### LINKED ISSUE #1777: Template linting
This seems to resurface regularly, so making this issue as a place where we can write stuff down

* A generic html linter can not work, we need a litner that correctly parses HTML, possible with nested JS, that contains django's template tags both in the html as well as the js parts. Template tags need to be understood because control flow and block tags require indentation.
* [djhtml](https://github.com/rtts/djhtml) seems to only do indentation.
* [djlint](https://github.com/Riverside-Healthcare/djLint) is proposed by @Kakadus in https://github.com/e-valuation/EvaP/pull/1770#discussion_r913711367 with this example code:
```django
<div class="modal fade"
     id="successMessageModal_{{ modal_id }}"
     tabindex="-1"
     role="dialog"
     aria-labelledby="successMessageModalLabel_{{ modal_id }}"
     aria-hidden="true">
    <div class="modal-dialog" role="document">
        <div class="modal-content">
            <div class="modal-header">
                <h5 class="modal-title" id="successMessageModalLabel_{{ modal_id }}">{% trans 'Message sent' %}</h5>
            </div>
            <div class="modal-body">{% trans 'Your message was successfully sent.' %}</div>
        </div>
    </div>
</div>
```
this looks relatively ok to me, but we'd have to investigate whether it works for most of our templates, or if it fails in some edge cases.

niklasmohrin:
Closes #1777


evap/results/templates/result_widget_tooltip.html
```diff
@@ -2,7 +2,8 @@
 
 {% if weight_info %}
     <p>
-        {% blocktranslate with percentage=weight_info %}This evaluation contributes {{ percentage }} to the final grade of the course.{% endblocktranslate %}
+        {% blocktranslate with percentage=weight_info %}This evaluation contributes {{ percentage }} to the final grade
+            of the course.{% endblocktranslate %}
```
richardebeling: We need to somehow get `trimmed` in

evap/staff/templates/staff_semester_preparation_reminder.html
```diff
@@ -70,18 +90,28 @@ <h3>
                         </thead>
                         <tbody>
                             {% for evaluation in evaluations %}
-                                <tr{% if evaluation.state != evaluation.State.PREPARED %} class="deactivate"{% endif %}>
+                                <tr {% if evaluation.state != evaluation.State.PREPARED %}class="deactivate"{% endif %}>
```
richardebeling, line 93: This was intentionally different before. I'm fine with both versions, just to call it out explicitly

evap/staff/templates/staff_semester_preparation_reminder.html
```diff
@@ -99,7 +129,9 @@ <h3>
                 {% for responsible, evaluations, _ in responsible_list %}
                     <b>{{ responsible.full_name }}</b><br />
                     {% for evaluation in evaluations %}
-                        - {{ evaluation.full_name }} (<span class="fst-italic">{{ evaluation.course.type }}</span>)<br />
+                        - {{ evaluation.full_name }} (<span
+                            class="fst-italic"
+                        >{{ evaluation.course.type }}</span>)<br />
```
richardebeling, line 134: This may be a bit worse -- but I think it's still fine

evap/contributor/templates/contributor_evaluation_form.html
```diff
@@ -18,9 +18,12 @@
 
     <div class="callout callout-info small">
         {% if editable %}
-            {% blocktranslate %}Please review the evaluation's details below, add all contributors and select suitable questionnaires. Once everything is okay, please approve the evaluation on the bottom of the page.{% endblocktranslate %}
+            {% blocktranslate %}Please review the evaluation's details below, add all contributors and select suitable
```
niklasmohrin: we noticed these breaks in blocktranslates, we should probably add `trimmed` then; will look into contributing or reporting this to djangofmt

richardebeling (commented):
50 files viewed, looks generally good, thanks

evap/contributor/templates/contributor_evaluation_form.html
niklasmohrin: @richardebeling So there is a builtin lint that all blocktranslates should be `trimmed` which we can apply with `djangofmt check . --fix`. Not sure if that is what we want though, seems to add more noise. Maybe we should contribute a setting so that trimmed is only needed/added if the blocktranslate spans more than one line?

evap/evaluation/templates/base.html
```diff
@@ -1,10 +1,10 @@
...
-{% load evaluation_filters %}
-{% load navbar_templatetags %}
-
-{% endspaceless %}<!DOCTYPE html>
+    {% load static %}
+    {% load cache %}
+    {% load evaluation_filters %}
+    {% load navbar_templatetags %}
+{% endspaceless %}
+<!DOCTYPE html>
```
niklasmohrin: Is this just wrong?

evap/grades/templates/grades_semester_view.html
```diff
@@ -63,59 +78,122 @@ <h3 class="col-8 mb-0">
...
+                                            value="0"
+                                            confirm-button-class="btn-primary"
+                                        >
                                             <span slot="title">{% translate 'Will final grades be uploaded?' %}</span>
                                             <span slot="action-text">{% translate 'Confirm' %}</span>
                                             <span slot="question">
                                                 {% blocktranslate trimmed with course_name=course.name %}
-                                                    Please confirm that a grade document for the course <strong>{{ course_name }}</strong> will be uploaded later on.
+                                                    Please confirm that a grade document for the course <strong
+                                                    >{{ course_name }}</strong> will be uploaded later on.
```
richardebeling: We have a few linebreaks like thus, inside the strong tag. A bit weird, but doesn't happen too often, so 🤷‍♂️

richardebeling (commented):
103 files look good to me. 17 files need blocktranslate trimmed fixed, then they are also good.

evap/evaluation/templates/base.html
richardebeling: I think both the old and the new code are correct. Whitespace before a `DOCTYPE` are [legal](https://html.spec.whatwg.org/multipage/syntax.html#writing). We could also juts remove the spaceless block here.

evap/contributor/templates/contributor_evaluation_form.html
richardebeling: I think always putting in `trimmed` is fine: We only use blocktranslate, if we expect the contents to be larger/more complex, and in those cases, `trimmed` is basically always appropriate.

We have 117 occurrences of `blocktranslate`, I think it would be manageable. We can do it separately (PR/commit).

evap/staff/templates/staff_course_type_merge.html
```diff
@@ -15,9 +15,10 @@ <h3>{% translate 'Merge course types' %}</h3>
     <div class="card mb-3">
         <div class="card-body">
             <p>
-                {% blocktranslate %}The following courses currently have the course type <strong>{{ other_type }}</strong> and will
-                get the type <strong>{{ main_type }}</strong> once you merge the course types. The course type <strong>
-                {{ other_type }}</strong> will be deleted after that.{% endblocktranslate %}
+                {% blocktranslate trimmed %}The following courses currently have the course type <strong
+                    >{{ other_type }}</strong> and will get the type <strong>{{ main_type }}</strong> once you merge the
+                    course types. The course type <strong>
+                        {{ other_type }}</strong> will be deleted after that.{% endblocktranslate %}
```
richardebeling: technically, I think there shouldn't be a space here between `<strong>` and `{{ other_type }}`, but it probably doesn't matter much

evap/evaluation/templates/choice_button.html
```diff
@@ -1,12 +1,24 @@
...
+    {% if choice.data.selected %}checked{% endif %}
+/>
+<label
+    class="
+        btn btn-sm btn-outline-primary
         {% if formelement.field.disabled or disabled %} disabled{% endif %}
-
-        {% if formelement.errors %} choice-error{% endif %}"
-        for="{{ choice.id_for_label }}" data-bs-toggle="tooltip" data-bs-placement="top" title="{{ tooltip }}">
+        {% if formelement.errors %} choice-error{% endif %}
```
richardebeling, line 16: with line breaks, we don't need the initial spaces here (but not super important)

niklasmohrin (commented):
I looked at everything now

evap/staff/templates/staff_semester_view_evaluation.html
```diff
@@ -2,265 +2,474 @@
...
 
-<td data-col="name" data-order="{{ evaluation.full_name }}">
-    <div>
-    {% if not info_only and request.user.is_manager %}
-        <a href="{% url 'staff:evaluation_edit' evaluation.id %}">{{ evaluation.full_name }}</a>
-        {% include 'cms_evaluation_link_icons.html' %}
-        <a class="small no-underline" href="{% url 'staff:course_edit' evaluation.course.id %}" data-bs-toggle="tooltip" data-bs-placement="right" title="{% translate 'Edit course' %}">
-            <span class="fas fa-chalkboard-user"></span>
-        </a>
+    {% if evaluation.textanswer_review_state == evaluation.TextAnswerReviewState.NO_TEXTANSWERS or evaluation.textanswer_review_state == evaluation.TextAnswerReviewState.NO_REVIEW_NEEDED %}
```
richardebeling, line 79: interesting that this doesn't get a line break -- maybe worth a bug report?

richardebeling (approved):
thanks

evap/evaluation/templates/choice_button.html
niklasmohrin: I think I will leave this to a future update of djangofmt

evap/staff/templates/staff_semester_view_evaluation.html
richardebeling: Created https://github.com/UnknownPlatypus/djangofmt/issues/379. The code here is simple, but I think the problem can get very complex given that we can have arbitrary expressions. Currently, djangofmt doesn't really look into / parse the tag expression here, except for newline removal. Let's see what the maintainers think.

PR #2738, opened 2026-06-01
Use template tag for breadcrumb items

### LINKED ISSUE #2642: Refactoring: link-if-not-self-link breadcrumb helper template block tag
For the breadcrumbs at the top of the page, we want elements to be clickable links _unless_ the user is already on the page that we would link to. Currently, we typically implement this with extra state passed to the template:

https://github.com/e-valuation/EvaP/blob/a68bace07b5d5e150ca87c287f44d87db2e29128/evap/staff/templates/staff_course_base.html#L15-L19

https://github.com/e-valuation/EvaP/blob/a68bace07b5d5e150ca87c287f44d87db2e29128/evap/grades/templates/grades_course_base.html#L8-L12

https://github.com/e-valuation/EvaP/blob/a68bace07b5d5e150ca87c287f44d87db2e29128/evap/staff/templates/staff_semester_base.html#L14-L18

and one instance is being added in https://github.com/e-valuation/EvaP/pull/2599 in evap/staff/templates/staff_questionnaire_base.html

This requires juggling additional state. It should be possible to [build a custom django template block tag](https://docs.djangoproject.com/en/6.0/howto/custom-template-tags/#simple-block-tags) that is just passed the target URL and then selectively adds the HTML anchor only if the user isn't already on this page.

Some more context:
* Related stackoverflow question for a similar problem: https://stackoverflow.com/questions/10263482/compare-request-path-with-a-reversed-url-in-django-template
* Related discussion: https://github.com/e-valuation/EvaP/pull/2599/changes#r2716402428

LukasMansour:
This PR provides a solution to #2642 by adding a custom Django template tag that adds breadcrumbs that can automagically figure out if they are self-referencial ("link-if-not-self") and therefore disabled or not respectively.

One advantage of this is more standardisation for breadcrumbs across the website.
Furthermore, we can avoid using some of the "disable_breadcrumb_tags" in the project, slightly improving the backend code. 

Closes #2642

### DIFF
--- a/evap/grades/templates/grades_base.html
+++ b/evap/grades/templates/grades_base.html
@@ -1,4 +1,5 @@
 {% extends 'base.html' %}
+{% load staff_templatetags %}
 
 {% block title %}{% translate 'Grade publishing' %} - {{ block.super }}{% endblock %}
 
@@ -6,10 +7,14 @@
     <div class="breadcrumb-bar">
         <ul class="breadcrumb">
             {% block breadcrumb %}
-                {% if disable_breadcrumb_grades or not user.is_grade_publisher %}
-                    <li class="breadcrumb-item">{% translate 'Grade publishing' %}</li>
+                {% if not user.is_grade_publisher %}
+                    {% breadcrumb_item %} 
+                        {% translate 'Grade publishing' %}
+                    {% endbreadcrumb_item %}
                 {% else %}
-                    <li class="breadcrumb-item"><a href="{% url 'grades:index' %}">{% translate 'Grade publishing' %}</a></li>
+                    {% breadcrumb_item_url 'grades:index' %} 
+                        {% translate 'Grade publishing' %}
+                    {% endbreadcrumb_item_url %}
                 {% endif %}
             {% endblock %}
         </ul>
--- a/evap/grades/templates/grades_base.html
+++ b/evap/grades/templates/grades_base.html
@@ -9,11 +9,11 @@
             {% block breadcrumb %}
                 {% if not user.is_grade_publisher %}
                     {% breadcrumb_item %}
-                    {% translate 'Grade publishing' %}
+                        {% translate 'Grade publishing' %}
                     {% endbreadcrumb_item %}
                 {% else %}
                     {% breadcrumb_item_url 'grades:index' %}
-                    {% translate 'Grade publishing' %}
+                        {% translate 'Grade publishing' %}
                     {% endbreadcrumb_item_url %}
                 {% endif %}
             {% endblock %}
--- a/evap/staff/templates/staff_course_base.html
+++ b/evap/staff/templates/staff_course_base.html
@@ -12,17 +12,10 @@
 
 {% block breadcrumb %}
     {{ block.super }}
-
     {% if course %}
-        {% if disable_breadcrumb_course %}
-            {% breadcrumb_item %} 
-                {{ course.name }}
-            {% endbreadcrumb_item %}
-        {% else %}
-            {% breadcrumb_item_url 'staff:course_edit' course.id %} 
-                {{ course.name }}
-            {% endbreadcrumb_item_url %}
-        {% endif %}
+        {% breadcrumb_item_url 'staff:course_edit' course.id %} 
+            {{ course.name }}
+        {% endbreadcrumb_item_url %}
     {% elif evaluation.course %}
         {% breadcrumb_item_url 'staff:course_edit' evaluation.course.id %} 
             {{ evaluation.course.name }}
--- a/evap/staff/templates/staff_course_base.html
+++ b/evap/staff/templates/staff_course_base.html
@@ -15,13 +15,13 @@
 
     {% if course %}
         {% url 'staff:course_edit' course.id as course_edit_url %}
-        {% create_breadcrumb course_edit_url %} 
+        {% breadcrumb_item course_edit_url %} 
             {{ course.name }}
-        {% endcreate_breadcrumb %}
+        {% endbreadcrumb_item %}
     {% elif evaluation.course %}
         {% url 'staff:course_edit' evaluation.course.id as course_edit_url %}
-        {% create_breadcrumb course_edit_url %} 
+        {% breadcrumb_item course_edit_url %} 
             {{ evaluation.course.name }}
-        {% endcreate_breadcrumb %}
+        {% endbreadcrumb_item %}
     {% endif %}
 {% endblock %}
--- a/evap/staff/templates/staff_course_base.html
+++ b/evap/staff/templates/staff_course_base.html
@@ -1,4 +1,5 @@
 {% extends 'staff_semester_base.html' %}
+{% load staff_templatetags %}
 
 {% block title %}
     {% if course %}
@@ -11,13 +12,16 @@
 
 {% block breadcrumb %}
     {{ block.super }}
+
     {% if course %}
-        {% if disable_breadcrumb_course %}
-            <li class="breadcrumb-item">{{ course.name }}</li>
-        {% else %}
-            <li class="breadcrumb-item"><a href="{% url 'staff:course_edit' course.id %}">{{ course.name }}</a></li>
-        {% endif %}
+        {% url 'staff:course_edit' course.id as course_edit_url %}
+        {% create_breadcrumb course_edit_url %} 
+            {{ course.name }}
+        {% endcreate_breadcrumb %}
     {% elif evaluation.course %}
-        <li class="breadcrumb-item"><a href="{% url 'staff:course_edit' evaluation.course.id %}">{{ evaluation.course.name }}</a></li>
+        {% url 'staff:course_edit' evaluation.course.id as course_edit_url %}
+        {% create_breadcrumb course_edit_url %} 
+            {{ evaluation.course.name }}
+        {% endcreate_breadcrumb %}
     {% endif %}
 {% endblock %}
--- a/evap/staff/templates/staff_course_base.html
+++ b/evap/staff/templates/staff_course_base.html
@@ -14,11 +14,11 @@
     {{ block.super }}
     {% if course %}
         {% breadcrumb_item_url 'staff:course_edit' course.id %}
-        {{ course.name }}
+            {{ course.name }}
         {% endbreadcrumb_item_url %}
     {% elif evaluation.course %}
         {% breadcrumb_item_url 'staff:course_edit' evaluation.course.id %}
-        {{ evaluation.course.name }}
+            {{ evaluation.course.name }}
         {% endbreadcrumb_item_url %}
     {% endif %}
 {% endblock %}
--- a/evap/staff/templates/staff_course_base.html
+++ b/evap/staff/templates/staff_course_base.html
@@ -14,9 +14,15 @@
     {{ block.super }}
 
     {% if course %}
-        {% breadcrumb_item_url 'staff:course_edit' course.id %} 
-            {{ course.name }}
-        {% endbreadcrumb_item_url %}
+        {% if disable_breadcrumb_course %}
+            {% breadcrumb_item %} 
+                {{ course.name }}
+            {% endbreadcrumb_item %}
+        {% else %}
+            {% breadcrumb_item_url 'staff:course_edit' course.id %} 
+                {{ course.name }}
+            {% endbreadcrumb_item_url %}
+        {% endif %}
     {% elif evaluation.course %}
         {% breadcrumb_item_url 'staff:course_edit' evaluation.course.id %} 
             {{ evaluation.course.name }}
--- a/evap/staff/templates/staff_course_base.html
+++ b/evap/staff/templates/staff_course_base.html
@@ -14,14 +14,12 @@
     {{ block.super }}
 
     {% if course %}
-        {% url 'staff:course_edit' course.id as course_edit_url %}
-        {% breadcrumb_item course_edit_url %} 
+        {% breadcrumb_item_url 'staff:course_edit' course.id %} 
             {{ course.name }}
-        {% endbreadcrumb_item %}
+        {% endbreadcrumb_item_url %}
     {% elif evaluation.course %}
-        {% url 'staff:course_edit' evaluation.course.id as course_edit_url %}
-        {% breadcrumb_item course_edit_url %} 
+        {% breadcrumb_item_url 'staff:course_edit' evaluation.course.id %} 
             {{ evaluation.course.name }}
-        {% endbreadcrumb_item %}
+        {% endbreadcrumb_item_url %}
     {% endif %}
 {% endblock %}
--- a/evap/staff/templates/staff_evaluation_base.html
+++ b/evap/staff/templates/staff_evaluation_base.html
@@ -13,11 +13,11 @@
     {% if evaluation %}
         {% if evaluation.name %}
             {% breadcrumb_item_url 'staff:evaluation_edit' evaluation.id %}
-            {{ evaluation.name }}
+                {{ evaluation.name }}
             {% endbreadcrumb_item_url %}
         {% else %}
             {% breadcrumb_item_url 'staff:evaluation_edit' evaluation.id %}
-            {{ evaluation.course.name }}
+                {{ evaluation.course.name }}
             {% endbreadcrumb_item_url %}
         {% endif %}
     {% endif %}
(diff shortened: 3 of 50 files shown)


evap/staff/templatetags/staff_templatetags.py
```diff
@@ -9,3 +10,16 @@
...
+def create_breadcrumb(context, content, url=None):
+    request = context["request"]
+    current_path = request.path
+    if url is None or current_path == url:
+        html = f'<li class="breadcrumb-item">{content}</li>'
+    else:
+        # create the href="link" string.
+        href_str = f'href="{url}"'
+        html = f'<li class="breadcrumb-item"><a {href_str}>{content}</a></li>'
+    return mark_safe(html)
```
richardebeling: `content` wasn't HTML-escaped -- whenever we use `mark_safe`, we have to make 100% sure we've escaped any kind of user input properly. There's also `format_html` for these kinds of use cases

evap/staff/templatetags/staff_templatetags.py
```diff
@@ -9,3 +10,16 @@
 def create_exam_evaluation_form(evaluation):
     form_id = f"exam_creation_form_{evaluation.id}"
     return ExamEvaluationForm(None, evaluation=evaluation, form_id=form_id)
+
+
+@register.simple_block_tag(takes_context=True)
+def create_breadcrumb(context, content, url=None):
```
richardebeling: Does the `create_` prefix have any meaningful purpose here, or could the tag just be `{% breadcrumb %}` and `{% endbreadcrumb %}`?

evap/staff/templates/staff_user_bulk_update.html
```diff
@@ -1,9 +1,16 @@
...
+
 {% block breadcrumb %}
     {{ block.super }}
-    <li class="breadcrumb-item"><a href="{% url 'staff:user_index' %}">{% translate 'Users' %}</a></li>
-    <li class="breadcrumb-item">{% translate 'Bulk update' %}</li>
+    {% url 'staff:user_index' as user_index_url %}
+    {% create_breadcrumb user_index_url %} 
+        {% translate 'Users' %}
+    {% endcreate_breadcrumb %}
+    {% create_breadcrumb request.path %}
```
richardebeling: What's the purpose of passing `request.path` as url here?

evap/staff/templates/staff_semester_base.html
```diff
@@ -9,12 +10,10 @@
...
     {{ block.super }}
-    <li class="breadcrumb-item">{% translate 'Semesters' %}</li>
     {% if semester %}
-        {% if disable_breadcrumb_semester %}
-            <li class="breadcrumb-item">{{ semester.name }}</li>
-        {% else %}
-            <li class="breadcrumb-item"><a href="{% url 'staff:semester_view' semester.id %}">{{ semester.name }}</a></li>
-        {% endif %}
+        {% url 'staff:semester_view' semester.id as semester_view_url %}
+        {% create_breadcrumb url=semester_view_url %}
```
richardebeling: What's the difference between `{% create_breadcrumb url=semester_view_url %} ` and `{% create_breadcrumb semester_view_url %}`, and which one do we want to use?

evap/staff/templates/staff_infotexts.html
```diff
@@ -1,8 +1,14 @@
 {% extends 'staff_base.html' %}
 
+{% load staff_templatetags %}
+
 {% block breadcrumb %}
     {{ block.super }}
-    <li class="breadcrumb-item">{% translate 'Infotexts' %}</li>
+    
+    {% url request.resolver_match.view_name as self_url %}
+    {% create_breadcrumb self_url %}
```
richardebeling: What's the intention of resolving `request.resolver_match.view_name` and passing it to the block here?

evap/staff/templates/staff_evaluation_base.html
```diff
@@ -11,9 +12,15 @@
...
-        	<li class="breadcrumb-item">{{ evaluation.name }}</li>
+            {% url 'staff:evaluation_edit' evaluation.id as eval_edit_id %}
+            {% create_breadcrumb eval_edit_id %} 
+                {{ evaluation.name }}
+            {% endcreate_breadcrumb %}
     	{% else %}
-    		<li class="breadcrumb-item">{{ evaluation.course.name }}</li>
+            {% url 'staff:evaluation_edit' evaluation.id as eval_edit_id %}
+            {% create_breadcrumb eval_edit_id %} 
+                {{ evaluation.course.name }}
```
richardebeling, line 20: If we need this often, could we have two block tags, one `breadcrumb_url` that forwards all inputs for `url` and then uses the output as the link target, and one `breadcrumb` that takes the link target directly?

Code-wise, one could call the other, so it should just be one more line of code in python, but we could skip the additional line in each template to define the URL (and we don't have to think about a variable name -- `eval_edit_id` is not great, but coming up with a good name is also hard here).

What do you think about this? (cc @niklasmohrin)

niklasmohrin: Makes sense to me

Some time ago, I was thinking about something more radical even, but I am not sure if it works for all cases. (probably not)

Just give the object directly to the breadcrumb helper and let the helper do
```python
if isinstance(obj, Evaluation):
    url = reverse("staff:evaluation_view", ...)
elif isinstance(obj, Course):
    url = ...
```
and so on. Probably not exactly like this, but similar (we can probably use a `match` statement to make it a lot nicer).

evap/staff/templates/staff_user_bulk_update.html
niklasmohrin: Seems like a way to always make the helper go into the "dont put a url" branch? Seems like a bad idea to me, the idea behind the breadcrumb system is to make it easy to inherit from other templates and get their breadcrumbs for free, but it we put request.path, then the breadcrumb item will be missing (=wrong) in the derived template. I think the straightforward way to say "dont put url" would be to not provide the url argument instead, I suppose the block code can mark arguments as optional. If so, let's do that?

richardebeling: Yeah, if it's just about going into the `url is None or current_path == url` branch, I would very much prefer to just not pass a URL and use the `url is None` aspect. Some "invocations" of the new block tag in this PR already do not provide a URL.

evap/staff/templates/staff_infotexts.html
LukasMansour: If only I knew. Should be removed and more clear now.

evap/staff/templates/staff_semester_base.html
LukasMansour: This was left over from testing something.

evap/staff/templates/staff_user_bulk_update.html
LukasMansour: Yeah no idea what I was thinking here. Fixed it.

evap/staff/templatetags/staff_templatetags.py
LukasMansour: It's now called breadcrumb_item

evap/staff/templatetags/staff_templatetags.py
LukasMansour: Done.

niklasmohrin (changes requested):
What do you think about https://github.com/e-valuation/EvaP/pull/2738#discussion_r3375423495 ?

evap/staff/templates/staff_evaluation_base.html
LukasMansour: Would be possible, however, would probably have to come with some interesting behavior for many of the edge-cases for breadcrumbs. I will look into if it is a nice solution, first intuition is no though.

EDIT: Richard's idea will work fine, not sure if the radical approach is really worth it.

evap/staff/templates/staff_semester_base.html
```diff
@@ -9,12 +10,9 @@
 
 {% block breadcrumb %}
     {{ block.super }}
-    <li class="breadcrumb-item">{% translate 'Semesters' %}</li>
```
LukasMansour, line 12: I would argue that the breadcrumbs look a lot better without this singular breadcrumb. This breadcrumb links to nowhere, just makes the breadcrumb bar longer and looks ugly when something clickable appears to the right of it in the breadcrumb bar.

evap/staff/templates/staff_course_base.html
```diff
@@ -11,13 +12,14 @@
 
 {% block breadcrumb %}
     {{ block.super }}
+
     {% if course %}
-        {% if disable_breadcrumb_course %}
```
niklasmohrin, line 15: we still set this to true in some places, let's keep this behavior

evap/staff/templates/staff_semester_base.html
```diff
@@ -9,12 +10,9 @@
...
     {% if semester %}
-        {% if disable_breadcrumb_semester %}
```
niklasmohrin, line 14: same here, although I think that we only set it to True if the link would be to the same page now, so I guess we can just remove the variable in the views

evap/staff/templatetags/staff_templatetags.py
```diff
@@ -9,3 +11,26 @@
 def create_exam_evaluation_form(evaluation):
     form_id = f"exam_creation_form_{evaluation.id}"
     return ExamEvaluationForm(None, evaluation=evaluation, form_id=form_id)
+
+
+@register.simple_block_tag(takes_context=True)
+def breadcrumb_item(context, content, url=None):
```
niklasmohrin: can we make it the other way around, that is, have the logic for urls in breadcrumb_item_url and only pass `url=None` in breadcrumb_item? otherwise, we might accidentally pass an unresolved url name with breadcrumb_item

niklasmohrin (changes requested):
Most changes look good, I have a small number of comments:

evap/staff/templatetags/staff_templatetags.py
LukasMansour: Resolving this, have a look if it is what you had in mind.

evap/staff/templates/staff_semester_base.html
LukasMansour: done.

evap/staff/templates/staff_course_base.html
LukasMansour: done.

evap/staff/templates/staff_questionnaire_base.html
```diff
@@ -1,20 +1,20 @@
...
+        {% endbreadcrumb_item_url %}
+        {% breadcrumb_item_url 'staff:questionnaire_edit' questionnaire.id %}
+            {{ questionnaire.name }}
+        {% endbreadcrumb_item_url %}
     {% else %}
-        <li class="breadcrumb-item">{% translate 'Questionnaires' %}</li>
+        {% breadcrumb_item %}
+            {% translate 'Questionnaires' %}
+        {% endbreadcrumb_item %}
     {% endif %}
```
richardebeling, line 14: Could this just be
```jinja
{% breadcrumb_item_url 'staff:questionnaire_index' %}
    {% translate 'Questionnaires' %}
{% endbreadcrumb_item_url %}
{% if questionnaire %}
    {% breadcrumb_item_url 'staff:questionnaire_edit' questionnaire.id %}
        {{ questionnaire.name }}
    {% endbreadcrumb_item_url %}
{% endif %}
```
?

evap/staff/templates/staff_user_form.html
```diff
@@ -1,14 +1,21 @@
...
-        <li class="breadcrumb-item">{{ form.instance.full_name }}</li>
+        {% breadcrumb_item %}
+            {{ form.instance.full_name }}
+        {% endbreadcrumb_item %}
     {% else %}
-        <li class="breadcrumb-item">{% translate 'Create user' %}</li>
+        {% breadcrumb_item %}
+            {% translate 'Create user' %}
+        {% endbreadcrumb_item %}
     {% endif %}
```
richardebeling: Could we do
```jinja
    {% breadcrumb_item %}
        {% if form.instance.id %}{{ form.instance.full_name }}{% else %}{% translate 'Create user' %}{% endif %}
    {% endbreadcrumb_item %}

**Review by richardebeling — APPROVED** [review]

File: evap/staff/templates/staff_semester_base.html (hunk 7 above)
Comment (richardebeling) [reply evap/staff/templates/staff_semester_base.html:12]: @janno42 opinions?

File: evap/grades/templates/grades_base.html [hunk 13]
```diff
@@ -1,17 +1,20 @@
...
 
 {% block title %}{% translate 'Grade publishing' %} - {{ block.super }}{% endblock %}
 
 {% block breadcrumb_bar %}
     <div class="breadcrumb-bar">
         <ul class="breadcrumb">
             {% block breadcrumb %}
-                {% if disable_breadcrumb_grades or not user.is_grade_publisher %}
-                    <li class="breadcrumb-item">{% translate 'Grade publishing' %}</li>
+                {% if not user.is_grade_publisher %}
```
niklasmohrin, line 10: I wonder if we should also add access checks to the template tag so that we can remove the branch here too. If we want it, we can do it another PR though. @richardebeling Thoughts?

evap/staff/templates/staff_evaluation_textanswer_edit.html
```diff
@@ -1,13 +1,16 @@
...
 {% block breadcrumb %}
     {{ block.super }}
-    <li class="breadcrumb-item">
-        <a
-            href="{% url 'staff:evaluation_textanswers' evaluation.id %}#{{ textanswer.id }}"
-        >{% translate 'Text answers' %}</a>
-    </li>
-    <li class="breadcrumb-item">{{ textanswer.id }}</li>
+
+    {% breadcrumb_item_url 'staff:evaluation_textanswers' evaluation.id fragment=textanswer.id|stringformat:"s" %}
```
niklasmohrin: this is the same as before, but doesn't work; when clicking the save button we go to `#textanswer-<id>`

niklasmohrin (commented):
Looks good, one more thing before we merge :)

evap/staff/templates/staff_semester_base.html
janno42: Sure, let's get rid of it

evap/grades/templates/grades_base.html
LukasMansour: New PR because of my departure.

evap/staff/templates/staff_user_merge.html
```diff
@@ -1,11 +1,16 @@
...
 {% load evaluation_filters %}
 
 {% block breadcrumb %}
     {{ block.super }}
-    <li class="breadcrumb-item"><a href="{% url 'staff:user_index' %}">{% translate 'Users' %}</a></li>
-    <li class="breadcrumb-item">{% translate 'Merge users' %}</li>
+    {% breadcrumb_item_url 'staff:user_index' %}
+        {% translate 'Users' %}
+    {% endbreadcrumb_item_url %}
+    {% breadcrumb_item %}
```
janno42: could be a url

evap/staff/templates/staff_program_merge.html
```diff
@@ -1,11 +1,18 @@
...
 {% block breadcrumb %}
     {{ block.super }}
-    <li class="breadcrumb-item"><a href="{% url 'staff:program_index' %}">{% translate 'Programs' %}</a></li>
-    <li class="breadcrumb-item">{% translate 'Merge programs' %}</li>
+
+    {% breadcrumb_item_url 'staff:program_index' %}
+        {% translate 'Programs' %}
+    {% endbreadcrumb_item_url %}
+
+    {% breadcrumb_item %}
```
janno42: could be a url

janno42 (changes requested):

niklasmohrin:
Can you also add a more descriptive PR title / description?

LukasMansour:
> Can you also add a more descriptive PR title / description?

Edited. Does this suite you, or more detailed?

evap/staff/templates/staff_user_merge.html
LukasMansour: agreed. done.

evap/staff/templates/staff_program_merge.html
LukasMansour: agreed done.

janno42 (approved):

niklasmohrin (approved):
Thanks!

PR #2739, opened 2026-06-08
Remove no-op `utils.js` and `csrf-utils.js` script tag in `base.html`

niklasmohrin:
Just noticed this while looking at #2718. Currently, `utils.js` and `csrf-utils.js` have no side effects, so having it like this does not do anything. We added the former in 93cf2413fcd59f6bf92d1020056c7259b5224796 because we added global side effects, but I removed those in 7ac051c16885e08ce9d38a091451466d086daf93.

### DIFF
--- a/evap/evaluation/templates/base.html
+++ b/evap/evaluation/templates/base.html
@@ -73,9 +73,6 @@
         <script type="text/javascript" src="{% static 'js/plugins/jquery.formset.js' %}"></script>
         <script type="text/javascript" src="{% static 'js/Sortable.min.js' %}"></script>
 
-        <script type="module" src="{% static 'js/csrf-utils.js' %}"></script>
-        <script type="module" src="{% static 'js/utils.js' %}"></script>
-
         <script type="module" src="{% static 'js/base-template.js' %}"></script>
         <script type="text/javascript">
             activateTooltips = function(selector = "", tooltipType = "tooltip", additionalOptions) {


evap/evaluation/templates/base.html
```diff
@@ -73,9 +73,6 @@
         <script type="text/javascript" src="{% static 'js/plugins/jquery.formset.js' %}"></script>
         <script type="text/javascript" src="{% static 'js/Sortable.min.js' %}"></script>
 
-        <script type="module" src="{% static 'js/csrf-utils.js' %}"></script>
-        <script type="module" src="{% static 'js/utils.js' %}"></script>
-
```
richardebeling, line 78: Wait, the purpose of these was to make the exported symbols from these files available in the "global" namespace in javascript (e.g., `export const CSRF_HEADERS = { "X-CSRFToken": csrftoken };`). Does this not work? How does it work if we remove the includes here?

niklasmohrin: No, not as far as I know. The `export` doesn't do anything global, it just adds the given item to the module's exports so that they can be used in `import { something } from "themodule"`. The global change was only due to `(globalThis as any).something = something`. But now that we explicitly import those items instead of assuming them in the global namespace, we don't need to add them to it anymore (this is already merged) and now running this script doesn't have any effect

The only thing is that the browser runs each module only once, and we might run the script only later when importing from it the first time, but I think this is irrelevant for us

richardebeling: I see, makes sense, thanks

richardebeling (approved):

fekoch (approved):

PR #2743, opened 2026-06-15
Use `QuestionAssignment.order` where we expect it

### LINKED ISSUE #2741: Incorrect Question Order
The order of the questions in the questionnaire editing form differs from the order on the preview and voting pages.
In the editing form, the questions can be reordered and the new order is saved correctly. After reloading the page, the questions are also correctly ordered.
However, on the preview and voting pages, the order of some of the questions is incorrect and different to that on the editing page. This issue was probably introduced in #2513, and I assume that an ordering step is missing somewhere along the process.

Steps to reproduce: Create a new questionnaire containing a question and save it. Open the edit page again, add another question, move it to the top, then save. The question will now be at the top of the edit page, but at the bottom of the preview page.

niklasmohrin:
Closes #2741

When we introduced `QuestionAssignment` we moved the `order` field from `Question` to `QuestionAssignment`. Semantically, this makes sense, because the same question might be ordered differently in different questionnaires. However, specifying an ordering on a through-model like this does not result in Django ordering the related instances as we expect (see https://code.djangoproject.com/ticket/30460). Instead, the order of the related model is used, which in our case (on `Question`) was unspecified, so the database just returned some arbitrary order.

The fix is to _not_ iterate over `questionnaire.questions` and instead prefer going via `questionnaire.question_assignments`. This ensures that the `QuestionAssignment` ordering is used. Another option would be to always prefetch with custom `Prefetch` objects like `Prefetch("questions", queryset=Question.objects.order_by("assignments"))`. However, this seems less robust then explicitly going via assignments in the right places and also uses the `order_by("through-model")` hack from the Django ticket which, as far as I can tell, is not documented.

The places from the issue seem to be fixed by this PR. I checked for other places by searching for the strings `.questions` and `questions"`.

### DIFF
--- a/evap/student/forms.py
+++ b/evap/student/forms.py
@@ -59,7 +59,8 @@ class QuestionnaireVotingForm(forms.Form):
         super().__init__(*args, **kwargs)
         self.questionnaire = questionnaire
 
-        for question in self.questionnaire.questions.all():
+        for assignment in self.questionnaire.question_assignments.all():
+            question = assignment.question
             if question.is_text_question:
                 field = TextAnswerField.from_question(question)
             elif question.is_rating_question:
--- a/evap/contributor/forms.py
+++ b/evap/contributor/forms.py
@@ -57,13 +57,13 @@ class EvaluationForm(forms.ModelForm):
             Questionnaire.objects.general_questionnaires()
             .filter(Q(visibility=Questionnaire.Visibility.EDITORS) | Q(contributions__evaluation=self.instance))
             .distinct()
-            .prefetch_related("questions")
+            .prefetch_related("question_assignments", "question_assignments__question")
         )
         self.fields["dropout_questionnaires"].queryset = (
             Questionnaire.objects.dropout_questionnaires()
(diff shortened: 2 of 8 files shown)


evap/student/forms.py
niklasmohrin, line 1: Do we want to have tests that check this ordering? I think that we decided in previous cases that ordering is something that is usually just right if we have the correct `Meta.ordering` set, but I feel that this is not the case here

janno42: I agree that this is a bit more tricky here, and we should add a test. The wrong order of questions can make a questionnaire almost unusable.

janno42 (approved):
🔧

richardebeling (approved):
Changes here lgtm, although I can definitely see that people will stumble over this again in the future. Let's discuss options in person some time (I'm thinking of asking django to respect M2M ordering, I feel like this is a reasonable ask, and maybe building some workaround in the meantime)

richardebeling (approved):

PR #2744, opened 2026-06-15
Use evap-dev shell for `./manage.py run` process

niklasmohrin:
I wanted to use `debug_toolbar`, but our process-compose services used the minimal `evap` shell instead of the `evap-dev` one (which is the default for `nix develop`), so `debug_toolbar` was not available.

This makes all development dependencies available in the process-compose Python interpreter.


richardebeling (approved):
makes sense, thanks

PR #2745, opened 2026-06-15
Fix dynamic formset prefix on staff questionnaire form

### LINKED ISSUE #2742: Questions are not created correctly
Not all of the questions added to a questionnaire are currently saved.

Steps to reproduce: Create a new questionnaire containing two questions and save it. Open the edit page again and you will see that only one question was saved.

niklasmohrin:
Closes #2742

Ough, this was annoying to debug (see the issue). So the problem is that the underlying ids changed so our call to `makeFormSortable` passed the wrong prefix which meant that the dynamic formset code did not identify the numbering in the form field ids correctly and kept generating rows with the same ids. For some reason, our backend just ignored the fact that it received an array of values where it expected only a single one and decided to just sweep all but one value under the table.

I also changed the `makeFormSortable` callback to remove the `form-template` class from newly created rows because the template we use stays in place and the rows are not the template, although I think that the dynamic formset code queries the template node only once in the beginning, otherwise we would be seeing rows with duplicate values created in all kinds of places.

### DIFF
--- a/evap/staff/templates/staff_questionnaire_form.html
+++ b/evap/staff/templates/staff_questionnaire_form.html
@@ -100,7 +100,7 @@
                 applyTomSelect(row.querySelectorAll("select"));
                 registerSelectChangedHandlers();
             };
-            makeFormSortable("question_table", "questions", rowChanged, rowAdded, true, true);
+            makeFormSortable("question_table", "question_assignments", rowChanged, rowAdded, true, true);
 
             const selectChangedHandler = function(e) {
                 const checkbox = this.closest("td.question-type").querySelector("input[type=checkbox]");
--- a/evap/staff/tests/test_live.py
+++ b/evap/staff/tests/test_live.py
@@ -1,7 +1,9 @@
 from datetime import date, datetime
 
 from model_bakery import baker
+from selenium.webdriver.common.action_chains import ActionChains
 from selenium.webdriver.common.by import By
+from selenium.webdriver.common.keys import Keys
 from selenium.webdriver.support.expected_conditions import (
     element_to_be_clickable,
     invisibility_of_element_located,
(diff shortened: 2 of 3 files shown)


evap/staff/templates/staff_questionnaire_form.html
niklasmohrin, line 1: Again, is there something we can meaningfully test here? For this one I am learning more towards "this just has to work", but we manged to not catch this once in review, so I don't know

niklasmohrin: On the bright side, at least the AI also didn't catch this, so we can all sleep tight :^)

janno42 (approved):
➕

evap/staff/templates/staff_questionnaire_form.html
richardebeling: If we want a test to ensure this doesn't stop working, I think live/frontend test, re-order, assert "order" field in database is what we should do. Should be doable in ~10ish lines, assuming it's somewhat possible to simulate a drag+drop for reordering in Selenium.

richardebeling (approved):
lgtm, thanks

niklasmohrin:
So I wrote a frontend test which moves them using the nice `ActionChains(...).drag_and_drop` API. This worked, but then I figured we might want to check something more involved than just two elements and one move. Like, five elements and two moves. An hour later, I finally notice why it doesn't work, after the first move the dragged element is still a "ghost" and waiting for a "dragend" event. Oh noes.

So turns out that this whole dragging thing is still absolutely broken, and not just in selenium but in the official webdriver interface. See https://redirect.github.com/w3c/webdriver/issues/1488 and https://redirect.github.com/SeleniumHQ/selenium/issues/12525

I will just revert to the easy test case :(

richardebeling (approved):

evap/staff/tests/test_live.py
```diff
@@ -247,6 +249,60 @@ def test_questionnaire_selection(self):
...
+            _quantity=5,
+        )
+
+    def test_add_multiple(self):
+        with self.enter_staff_mode():
+            self.selenium.get(self.url)
+            add_another = self.selenium.find_element(By.CLASS_NAME, "add-row")
+            add_another.click()
+            add_another.click()
+            add_another.click()
```
fekoch, line 272: Is there any reason why you click three times on the button? I would think once should be enough or do you try to trigger something else?

niklasmohrin: Just simulating my personal interaction with the site :D

The real reason is that I wanted to check what happens if there are empty rows at the end and inbetween (notice that I use rows 5 and 7 instead of 5 and 6 below)

PR #2747, opened 2026-06-18
Import merged evaluation

### LINKED ISSUE #2746: Import fails after merging Evaluation into other Course
After merging an Evaluation with another Evaluation from a different Course, the next CMS import fails.
This occurs because, in `_import_evaluation`, we assume that the combination of Course and EvaluationLink either exists or that the Evaluation is new. In this case, however, the combination no longer exists, yet the Evaluation with the respective `cms_id` still exists. Therefore, it cannot be created (`IntegrityError: duplicate key value violates unique constraint "cms_evaluationlink_cms_id_key"`).

The importer needs to be changed to correctly handle this use case.

janno42:
fix #2746; support importing evaluations that have been merged into another course

### DIFF
--- a/evap/cms/json_importer.py
+++ b/evap/cms/json_importer.py
@@ -434,7 +434,7 @@ class JSONImporter:
 
     # pylint: disable=too-many-locals
     def _import_evaluation(  # noqa: PLR0912, PLR0915
-        self, course: Course, data: ImportEvent, earliest_exam_date: date | None = None
+        self, cms_course: Course, data: ImportEvent, earliest_exam_date: date | None = None
     ) -> Evaluation | None:
         # Don't import ignored evaluations again
         if IgnoredEvaluation.objects.filter(cms_id=data["gguid"]).exists():
@@ -444,8 +444,11 @@ class JSONImporter:
         if EvaluationLink.objects.filter(cms_id=data["gguid"], is_active=False).exists():
             return None
 
+        course = cms_course  # by default, we use the course listed in the cms as the evaluation's course
         try:
-            evaluation = Evaluation.objects.get(course=course, cms_evaluation_links__cms_id=data["gguid"])
+            evaluation = Evaluation.objects.get(cms_evaluation_links__cms_id=data["gguid"])
+            # if the evaluation already exists, we use its course in case it was merged into a different course
+            course = evaluation.course
         except Evaluation.DoesNotExist:
             evaluation = None
 
--- a/evap/cms/json_importer.py
+++ b/evap/cms/json_importer.py
@@ -379,17 +379,19 @@ class JSONImporter:
 
         try:
             cms_course_link = CourseLink.objects.get(cms_id=data["gguid"])
-            if cms_course_link.is_active:
-                course = cms_course_link.course
-                changes = update_with_changes(
-                    course,
-                    {
-                        "name_de": _clean_whitespaces_and_hyphens(data["title"]),
-                        "name_en": _clean_whitespaces_and_hyphens(data["title_en"]),
-                    },
-                )
-                if changes:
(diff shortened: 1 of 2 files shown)


evap/cms/json_importer.py
niklasmohrin, line 1: We should add a test for this behavior

evap/cms/json_importer.py
```diff
@@ -445,7 +445,10 @@ def _import_evaluation(  # noqa: PLR0912, PLR0915
             return None
 
         try:
-            evaluation = Evaluation.objects.get(course=course, cms_evaluation_links__cms_id=data["gguid"])
+            evaluation = Evaluation.objects.get(cms_evaluation_links__cms_id=data["gguid"])
+            # get the course from the evaluation in case it was merged into a different course than defined in the
+            # import data
+            course = evaluation.course
```
niklasmohrin, line 453: Is there any chance that we can choose the correct course once before calling `import_evaluation`? I think it is hard to track what course belongs to what now. In the calling code in `import_events` we have some `if course is None: continue` and I am not sure how this should interact with this new assignment here (my intuition is that if the original course is None but we find a course here then we should still run the import).

If this is not possible, then one (or both) of the two `course` variables needs a new name, I guess the course before this is something like `course_from_event` and this here is like a `selected_course`? Although `selected_course` really should just always be `evaluation.course`, so I guess we could also just use this everywhere?

niklasmohrin (commented):
I have some trouble understanding how all the parts come together here. In the issue it says that we have an assumption that two things exist together when they don't really have to, but here it is not obvious how this is split into two now

richardebeling (approved):
code change looks good but I agree that a test would be nice

evap/cms/json_importer.py
janno42: I tried looking up the course outside of `_import_evaluation`, but that would require a larger refactoring. So, I renamed the parameter to `cms_course`, the default value for `course` that is overwritten if the evaluation already exists.

evap/cms/json_importer.py
janno42: I would have expected `test_import_merged_evaluations` to fail in this case and don't really understand why it didn't.

janno42:
Turns out we already have `test_import_merged_evaluations` and it doesn't fail. I couldn't figure out why the error occurred on production.
However, I still think the changes from this PR are valid :)

richardebeling (approved):

evap/cms/tests/test_json_importer.py
```diff
@@ -871,12 +884,16 @@ def test_import_merged_evaluations(self):
...
         evaluation = Evaluation.objects.get(cms_evaluation_links__cms_id="0x10")
+        # the evaluation should have three participants (because student "1" has no email and is skipped)
         self.assertEqual(
             set(evaluation.participants.values_list("last_name", flat=True)),
             {"2", "3", "7"},
         )
         self.assertEqual(len(importer.statistics.updated_participants), 0)
 
+        other_evaluation = Evaluation.objects.get(cms_evaluation_links__cms_id="0x44")
+        self.assertEqual(evaluation.course, other_evaluation.course)
```
niklasmohrin: as we modified the cms link above, we now have `evaluation.id == other_evaluation.id` here, so this line always passes

niklasmohrin (approved):
In person discussion:
- the usage of a course other than `evaluation.course` outside of `_import_evaluation` is probably wrong
- long term, we probably want to get rid of `course_by_gguid` altogether because this is really just an ad hoc mapping of evaluation's gguids to courses; cms doesn't distinguish between courses and evaluations
- the notable change regarding the imported evaluation is the removal of the `course=course` when looking for linked evaluations

PR #2748, opened 2026-06-22
Update grades_-/staff_semester_view to css grid

fekoch:
Refactors the DataGrid class to provide different factory functions for different HTML representations of the tables.

Updates the tables in:
- `grades_semester_view.html`
- `staff_semester_view.html`

### DIFF
--- a/evap/static/ts/src/datagrid.ts
+++ b/evap/static/ts/src/datagrid.ts
@@ -23,10 +23,6 @@ interface State {
 
 interface BaseParameters {
     storageKey: string;
-    searchInput: HTMLInputElement;
-    resetSearch?: HTMLButtonElement;
-    filterButtons: HTMLButtonElement[];
-    resetFilterButton?: HTMLButtonElement;
 }
 
 interface DataGridParameters extends BaseParameters {
@@ -36,6 +32,7 @@ interface DataGridParameters extends BaseParameters {
 }
 
 export class DataGrid {
+    // TODO: split into LegacyDataGrid & DataGrid for faster integration
     private _rows?: Row[];
     private delayTimer: number | undefined;
     protected state: State = {
@@ -44,15 +41,12 @@ export class DataGrid {
         search: "",
         order: this.defaultOrder,
     };
-    protected readonly filterButtons: HTMLButtonElement[] = [];
-    protected readonly resetFilterButton?: HTMLButtonElement;
+    private _setSearchValue: (newSearchValue: any) => void = _ => {};
 
     protected constructor(
         private readonly storageKey: string,
         protected readonly sortableHeaders: Map<string, HTMLElement>,
         public readonly container: HTMLElement,
-        private readonly searchInput: HTMLInputElement,
-        protected readonly resetSearch: HTMLButtonElement | undefined,
         private readonly getRowElements: () => HTMLElement[] = () => [...this.container.children] as HTMLElement[],
         protected readonly defaultOrder: Order = [],
     ) {}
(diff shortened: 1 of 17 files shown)


niklasmohrin:
There are still TODOs in the code. Is there something that we should already look at?

fekoch:
> There are still TODOs in the code. Is there something that we should already look at?

Good point, I overlooked those. Something, something self review....

Still, @niklasmohrin I would appreciate a review. Resolving the TODOs should only result in a minimal diff. (I switched it to draft anyway.)

evap/static/ts/src/datagrid.ts
```diff
@@ -23,50 +25,201 @@ interface BaseParameters {
...
-        this.resetSearch = resetSearch;
-        this.state = this.restoreStateFromStorage();
+    protected state: State = {
+        equalityFilter: new Map<string, string[]>(),
+        rangeFilter: new Map<string, Range>(),
+        search: "",
+        order: this.defaultOrder,
+    };
+    protected readonly filterButtons: HTMLButtonElement[] = [];
+    protected readonly resetFilterButton?: HTMLButtonElement;
```
niklasmohrin: no one is providing any value for these two attributes, right?

evap/static/ts/src/datagrid.ts
```diff
@@ -23,50 +25,201 @@ interface BaseParameters {
...
+
+    public get rows() {
+        this._rows ??= this.fetchRows(this.getRowElements());
+        return this._rows;
+    }
+
+    public static buildSortableHeadersMap(headerContainer: HTMLElement): Map<string, HTMLElement> {
+        const sortableHeaders = new Map<string, HTMLElement>();
+        for (const orderElement of headerContainer.querySelectorAll<HTMLElement>(".col-order")) {
+            sortableHeaders.set(orderElement.dataset.col!, orderElement);
```
niklasmohrin: Same in old version, but: Why are we not querying for `[data-col]` here?

evap/static/ts/src/datagrid.ts
```diff
@@ -23,50 +25,201 @@ interface BaseParameters {
...
 }
 
-abstract class DataGrid {
-    private readonly storageKey: string;
-    protected sortableHeaders: Map<string, HTMLElement>;
-    protected container: HTMLElement;
-    private searchInput: HTMLInputElement;
-    protected readonly resetSearch?: HTMLButtonElement;
-    protected rows: Row[] = [];
+export class DataGrid {
```
niklasmohrin: It seems that the base data grid is now even more complicated and the responsibilities are less clear. I wonder if there is some way to decompose the responsibilities into more modules/classes. Intuitively, I would say that we have
a) sorting of rows b) filtering of rows c) save and restore of state d) mirroring internal state to the DOM. These responsibilities are somewhat intertwined though.

Maybe one possible way to split this is to have one class that does all the sorting, filtering, state management, but does not interact with the DOM, and another class drives an instance according to what users to in the DOM. Not sure though.

niklasmohrin:
Discussion today:
- CSS grid is not as great as we think, but we still want it in favor of tables for more flexible sizing
- we want to have table grid functionality in base datagrid, but the code is rather convoluted
- we want to refactor datagrid to separate concerns (filtering, sorting, rendering, initialization), perhaps with a factory function and by attaching several `Filter` instances to a grid; filters are tied to DOM elements and tell the grid to update, the grid re-renders itself
- long term: all filters fetch their range of possible values from the data in the DOM and create html inputs accordingly (similar to RangeSlider's includeValues today); edit: nevermind, we want to keep the chosen order of, for example, course types as in the database, and the dom doesn't know about that
- we develop this new data grid in parallel to the current data grid to avoid having a huge PR that changes all grids; this way, we can port HTML to CSS grid and js to new datagrid at once without doing everything in one huge PR

PR #2753, opened 2026-07-06
Add scheduled workflow to tag stale PRs

niklasmohrin:
From the docs, I think this is the correct configuration. There is a debug-only (= dry run) option, but I think we would have to merge this PR with it enabled to see what it does. We could do that and remove it once we saw what it does.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,16 @@
+name: 'Tag stale PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,16 @@
+name: 'Tag stale PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
```
richardebeling, line 4: Is there anything these magical "computers" cannot do? Incredible...

richardebeling (approved):
thanks :)

.github/workflows/stale.yml
niklasmohrin: yes, it appears that they are not yet able to decrypt complex directives like "daily"

PR #2755, opened 2026-07-06
Global evaluation progress markers on small screens

### LINKED ISSUE #2671: Global Evaluation Progress: Milestone Spacing on Narrow Screens
Image

The labels are anchored to a horizontal position. We should layout them in a way that doesn't overlap. One possible solution would be to have multiple vertical "lines" that we use, and have each label drop down to the line that allows it to exist without overlapping with the label on its left.

Best case, we find a CSS-only solution that only requires a small amount of CSS

suttrop:
closes #2671


niklasmohrin (approved):

richardebeling (approved):

janno42:
It would be nice if the texts can't overlap the numbers, like here:
image

Instead, we could put all the numbers in one row on top and only the texts in steps. Perhaps the lines could continue below the numbers and down to the texts, either overlapping them or, better yet, placing them in the correct z-order and underlaying the texts with a background.

evap/static/scss/components/_progress.scss
```diff
@@ -73,10 +73,35 @@
...
     }
 }
 
+@include media-breakpoint-up(md) {
+    .spaced {
+        display: none;
+    }
+}
+
+@include media-breakpoint-down(md) {
```
niklasmohrin: this is not needed, we can set the height even if `display: none` is set

evap/static/scss/components/_progress.scss
```diff
@@ -73,10 +73,35 @@
...
+    .progress-step:nth-of-type(3n + 1) {
+        .spaced {
```
niklasmohrin: you can merge these as
```suggestion
    .progress-step:nth-of-type(3n + 1) .spaced {
```

niklasmohrin (commented):
site looks good to me

evap/static/scss/components/_progress.scss
```diff
@@ -73,10 +74,28 @@
...
+}
+
+.progress-step:nth-of-type(3n + 2) .spaced {
+    height: 18px;
+}
+
+
+.progress-step:nth-of-type(3n) .spaced {
+    height: 36px;
+}
```
niklasmohrin: missing final newline :^)

janno42 (approved):
✅ UI checked

evap/static/scss/components/_progress.scss
suttrop: :smiling_face_with_tear:

richardebeling (approved):

PR #2757, opened 2026-07-13
Replace deprecated tseslint.config() call.

### LINKED ISSUE #2639: Replace deprecated `tseslint.config(...)` call
Our configuration for eslint with ts-eslint uses their helper method `tseslint.config`. This method is apparently deprecated in favor of something in core eslint. We should update our config to use this new thing. See https://typescript-eslint.io/packages/typescript-eslint/#config-deprecated

LukasMansour:
Reading through https://typescript-eslint.io/packages/typescript-eslint/#migrating-to-defineconfig, it seems that this achieves exactly what we would like.

Closes #2639

### DIFF
--- a/evap/static/ts/eslint.config.js
+++ b/evap/static/ts/eslint.config.js
@@ -1,9 +1,10 @@
 // @ts-check
 
 import eslint from "@eslint/js";
+import { defineConfig } from "eslint/config";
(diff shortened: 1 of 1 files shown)


richardebeling (approved):

niklasmohrin (approved):
Thanks

PR #2758, opened 2026-07-13
Formatting for scss

geromequa:
Adds formatting for `.scss` using prettier.

### DIFF
--- a/evap/static/scss/.prettierrc.json
+++ b/evap/static/scss/.prettierrc.json
@@ -1,0 +1,5 @@
+{
+    "tabWidth": 4,
+    "arrowParens": "avoid",
+    "printWidth": 120
+}
--- a/evap/static/scss/.prettierrc.json
+++ b/evap/static/scss/.prettierrc.json
@@ -1,5 +1,4 @@
 {
     "tabWidth": 4,
-    "arrowParens": "avoid",
     "printWidth": 120
 }
--- a/evap/static/scss/_mixins.scss
+++ b/evap/static/scss/_mixins.scss
@@ -1,7 +1,13 @@
 @import "../bootstrap/scss/mixins";
 
 @mixin bar-shadow($color) {
-    text-shadow: 0 0 2px $color, 0 0 2px $color, 0 0 2px $color, 0 0 2px $color, 0 0 2px $color, 0 0 2px $color;
+    text-shadow:
+        0 0 2px $color,
+        0 0 2px $color,
+        0 0 2px $color,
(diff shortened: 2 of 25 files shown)


niklasmohrin (changes requested):
tests failing

evap/static/scss/_mixins.scss
```diff
@@ -35,7 +41,10 @@
         @content;
     } @else {
         $print-index: index(map-keys($breakpoints), $print-breakpoint);
-        $print: index(map-keys($breakpoints), $lower) <= $print-index and $print-index <= index(map-keys($breakpoints), $upper);
+        $print: index(map-keys($breakpoints), $lower) <=
+            $print-index and
+            $print-index <=
+            index(map-keys($breakpoints), $upper);
```
richardebeling, line 47: This is weird -- why can't we format it as
> index(map-keys($breakpoints), $lower) <= $print-index
> and $print-index <= index(map-keys($breakpoints), $upper);

? This should fit within 120 columns with ease

evap/static/scss/.prettierrc.json
```diff
@@ -0,0 +1,5 @@
+{
+    "tabWidth": 4,
+    "arrowParens": "avoid",
```
richardebeling: What does "arrowParens" do in scss?

geromequa: Probably nothing, so i removed it.

evap/static/scss/.prettierrc.json
niklasmohrin, line 1: why do we even have this? can't we just use the same file as for typescript? and if not, we should just use a symlink

niklasmohrin (commented):
Looks good :)

evap/static/scss/.prettierrc.json
geromequa: without this file prettier does not work properly for scss. symlink could work, but by now the scss config does not have the `arrowPens` option anymore, which we want in the ts config. See https://github.com/e-valuation/EvaP/pull/2758#discussion_r3574404179.

richardebeling: symlinking one file is also fine for me (I just don't want a full separate copy of the file that includes attributes that don't make sense)

niklasmohrin: I would expect that we can just have some global config for prettier. And a simple Google search reveals: yes, we can. according to https://prettier.io/docs/configuration, we can just specify everything in package.json and this config applies to all target files

PR #2760, opened 2026-07-20
Rename nix/nix-python to nix/python

fekoch:
Allows newer PyCharm versions to accept it as a local python interpreter.


niklasmohrin (approved):
Github useful as usual 

Image

PR #2762, opened 2026-07-20
Fix contributor index switch button label white space preservation

richardebeling:
In https://github.com/e-valuation/EvaP/pull/2736, we [introduced additional in-html whitespace](https://github.com/e-valuation/EvaP/pull/2736/changes?diff=split#diff-4995b7ca622ef8bbd0be7e45487de4a809bd094c539e558f2468b4a8037011f2R20-R22) in the switch button label on the contributor index page. However, in https://github.com/e-valuation/EvaP/pull/2054, we added the classes `text-break break-spaces` to the button label, which applies this CSS:
```css
.break-spaces {
    white-space: break-spaces;
}
```
Now, this unintuitively does not mean "break my text at whitespaces". It also preserves original breaking whitespaces (aka newlines). So, the two newlines our formatting added are also included in the div, causing the switch button to look like this:
image

I think we should get rid of this `break-spaces` class since it does not do what people would expect it to do. To still make our text wrap here, I discussed with @janno42 and we agreed that we should probably just allow text wrapping in `.btn-switch-label`.

### DIFF
--- a/evap/staff/templates/staff_semester_view.html
+++ b/evap/staff/templates/staff_semester_view.html
@@ -278,7 +278,7 @@
                                     <button
                                         slot="show-button"
                                         type="button"
-                                        class="btn btn-sm btn-light{% if rewards_active %} active{% endif %}"
+                                        class="btn btn-sm btn-light{% if rewards_active %} active{% endif %} h-100"
                                     >
                                         <span class="fas fa-check" aria-hidden="true"></span>
                                     </button>
--- a/evap/staff/templates/staff_semester_view.html
+++ b/evap/staff/templates/staff_semester_view.html
@@ -261,6 +261,7 @@
                                     type="submit"
                                     name="activation_status"
                                     value="on"
+                                    class="d-contents"
                                     confirm-button-class="btn-primary"
                                 >
                                     <span slot="title">{% translate 'Activate reward points' %}</span>
@@ -278,7 +279,7 @@
                                     <button
                                         slot="show-button"
                                         type="button"
-                                        class="btn btn-sm btn-light{% if rewards_active %} active{% endif %} h-100"
+                                        class="btn btn-sm btn-light{% if rewards_active %} active{% endif %}"
                                     >
                                         <span class="fas fa-check" aria-hidden="true"></span>
                                     </button>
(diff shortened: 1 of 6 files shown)


niklasmohrin (changes requested):
Student vote looks worse now: 

Before:  Image

After: Image

niklasmohrin:
semester view also has this, which is more functional than before, but also funny: 
image

evap/staff/templates/staff_semester_view.html
```diff
@@ -278,7 +278,7 @@ <h3 class="m-0 me-1">{{ semester.name }}</h3>
                                     <button
                                         slot="show-button"
                                         type="button"
-                                        class="btn btn-sm btn-light{% if rewards_active %} active{% endif %}"
+                                        class="btn btn-sm btn-light{% if rewards_active %} active{% endif %} h-100"
```
niklasmohrin: sus, the other button doesn't need this

from playing around in the inspector, I wonder if we should always use `display: contents` on our confirmation modals. I am really not sure though if that messes things up in other places

evap/student/templates/student_vote.html
```diff
@@ -68,10 +68,10 @@
             </small>
         </div>
     {% endif %}
-    <h3 class="mb-3 d-flex justify-content-between flex-column-mobile">
-        {{ evaluation.full_name }} ({{ evaluation.course.semester.name }})
+    <h3 class="mb-3 d-flex gap-2 justify-content-between align-items-end flex-column flex-md-row">
+        <span>{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</span>
 
-        <div class="float-lg-end mt-3 mt-lg-0 d-flex justify-content-end height-fit-content">
+        <div class="d-flex flex-shrink-0 justify-content-end height-fit-content">
```
fekoch: Maybe you could also remove the `d-flex` here, as this container has only a single child. (But you should check that, maybe it is needed for the `justify-content-end`.)

niklasmohrin (commented):
The sites look good to me, I have one concern

janno42 (approved):
Sure it'll be fine :)

evap/staff/templates/staff_semester_view.html
richardebeling: changed to `d-contents` -- good now? Or do we want to change something else?

niklasmohrin (approved):

PR #2764, opened 2026-07-20
Fix quoting in included `result_widget_tooltip` template

niklasmohrin:
In #2736 we introduced template formatting, which (among other things) normalizes the quotes used for element attributes like `class="..."`. In particular, the formatter changes the style to use double quotes by default.

In normal HTML, the type of quotes is insignificant. However, in `result_widget_tooltip`, we _need_ to use single quotes for the span's class attribute, because the entire file is included in other files as `<div title="{% include 'result_widget_tooltip.html' with ... %}">`. Thus, with single quotes, we get something like `title="<span class='abc'>"` whereas with double quotes we get `title="<span class="abc">"`, so the `title` attribute is terminated early and the `abc">"` spills into the surrounding HTML. This leads to the result page looking like this: 
image

I checked:
- all occurrences of `{% include` on current main, we only include `result_widget_tooltip.html` in this way
- all occurrences of `='` in the diff of #2736, other places look fine

### DIFF
--- a/evap/results/templates/result_widget_tooltip.html
+++ b/evap/results/templates/result_widget_tooltip.html
@@ -16,13 +16,16 @@
     {% with question_result.question.is_bipolar_likert_question as is_bipolar %}
         <p>
             {% for count, name, color, value in question_result.zipped_choices %}
+                {# djangofmt:ignore #}
                 <span
-                    class="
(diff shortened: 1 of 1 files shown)


richardebeling (approved):

janno42 (approved):
👁️‍🗨️

PR #2765, opened 2026-07-26
Validate all generated HTML in tests

niklasmohrin:
Implements the idea from https://github.com/e-valuation/EvaP/issues/1777#issuecomment-2230499050

We add a middleware that checks all HTTP responses with HTML content against a validator. This ensures that not only our templates are linted, but also the final HTML. This catches bugs that occur only when combining multiple template files, for example duplicate html ids in a loop or cursed string quoting like #2764 (this is actually caught in the initial version of this PR until the fix is merged).

There don't seem to be many maintained HTML validators, the only serious one appears to be the Nu Html Checker (also called validator NU, VNU, and W3 Markup Validator): https://validator.nu/about.html It is associated with W3, so it's probably as good as it gets.

This pulls in a headless Java runtime (about 500MB) into our dev setup, but I think it is well worth it. To avoid having to start the JVM for every validation, we run a server as part of our process compose services with `nix run`. So the whole Java business is completely hidden from the developers, they get the Java env and the running service with the normal `nix run` as usual.


niklasmohrin:
(marking as draft because #2764 should be merged first, but other than that everything can be reviewed)

evap/evaluation/tests/tools.py
```diff
@@ -114,8 +116,76 @@ class SimpleTestCase(SeedBakerMixin, ResetLanguageOnTearDownMixin, django.test.S
...
+            data=html,
+        ).json()["messages"]
+
+        if errors:
+            raise InvalidHtmlError(path, html, errors)
+
+
 class WebTest(SeedBakerMixin, ResetLanguageOnTearDownMixin, django_webtest.WebTest):
-    pass
+    app_class = ValidatingTestApp
```
niklasmohrin: In retrospect, I feel this would be better modeled by a middleware. What do you think?

evap/results/templates/results_evaluation_detail.html
```diff
@@ -106,7 +106,6 @@ <h3>{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</h3>
                 <a
                     class="btn btn-sm btn-light d-print-none"
                     href="{% url 'results:evaluation_text_answers_export' evaluation.id %}?view_general_results={{ view_general_results.value }}&view_contributor_results={{ view_contributor_results.value }}{% if contributor_id is not None %}&contributor_id={{ contributor_id }}{% endif %}"
-                    type="button"
```
niklasmohrin, line 114: For this, see https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#type

Our usage of `type` here doesn't make any sense, we probably meant to put a `role=button`? I am hesitant to put in more accessibility attributes because we don't really have a clear guideline for when to do so

evap/staff/forms.py
```diff
@@ -567,29 +567,27 @@ def __init__(self, data=None, instance=None):
...
-    base_evaluation = forms.ModelChoiceField(Evaluation.objects.all(), required=True, widget=forms.HiddenInput())
     exam_date = forms.DateField(
         label=_("Exam date"),
         required=True,
         localize=True,
     )
     exam_type = forms.ModelChoiceField(ExamType.objects.all(), required=True, label=_("Exam type"))
 
-    def __init__(self, *args, evaluation=None, form_id=None, **kwargs):
+    def __init__(self, *args, evaluation: Evaluation, form_id=None, **kwargs):
```
niklasmohrin: This was a bigger change: I don't really see why we had both `evaluation` and `base_evaluation`, I think they are always equal in our current usage? Anyways, I removed `base_evaluation` and passed the evaluation id in the url instead because the evaluation is needed to generate the prefix with which the form is created, so that it knows the names in `request.POST` to read its data from

evap/evaluation/tests/tools.py
```diff
@@ -114,8 +116,76 @@ class SimpleTestCase(SeedBakerMixin, ResetLanguageOnTearDownMixin, django.test.S
...
+            s += f"context:\n{error['extract']}"
+        return s
+
+    def __str__(self) -> str:
+        errors = "\n".join(self.format_error(e) for e in self.errors)
+        return f"html validation failed for {self.path}:\n{errors}"
+
+
+class ValidatingTestApp(django_webtest.DjangoTestApp):
+    IGNORED_ERROR_PATTERNS = (
```
niklasmohrin: I would make a follow up issue with sub-issues for some of these; in particular the accessibility errors, the "not allowed as child of span", and the table stuff

Note that currently, we only ask for _errors_, but the validator also has _warnings_ for other things :D

evap/evaluation/tests/tools.py
richardebeling: Both fine with me.

evap/evaluation/tests/tools.py
richardebeling: I would also put relaxing the 200 HTTP status code check on that list: If we say content type is text/html, it should be valid HTML, no matter what the status code is

evap/development/middleware.py
```diff
@@ -0,0 +1,76 @@
...
+        assert isinstance(settings.VNU_URL, str)
+        self.vnu_url = settings.VNU_URL
+
+    def __call__(self, request: HttpRequest) -> HttpResponseBase:
+        response = self.get_response(request)
+        if (
+            isinstance(response, HttpResponse)
+            and response.headers.get("Content-Type", "").startswith("text/html")
+            and response.text
+        ):
```
niklasmohrin, line 59: coverage (locally and on codecov) reports that this condition is always true; if I add `else: print("gotcha")`, both branches are detected as taken :(

niklasmohrin: if I turn off typeguard, everything is reported correctly; see https://redirect.github.com/agronholm/typeguard/issues/356

Kakadus: urgh. I mean we do have two test runs: shuffled and coverage. How about disabling typeguard in coverage tests?

niklasmohrin: Yes, I was thinking the same thing. I am not sure whether we should then also turn if off in the frontend/live test job, because we track coverage for this one too and combine it with the normal one. If typeguard is messing up the mapping-to-lines in one of two coverage files, that can't be good. Or maybe codecov figures it out magically, but that seems unlikely to me (probably they just do `coverage combine` too).

I don't think that we (should) lose any coverage by not tracking in live tests, except for the code in `test_live.py` itself. We could just ignore that though.

On the other hand, maybe losing coverage due to typeguard is not so bad and we would rather have typeguard always on instead of correct coverage. @richardebeling What do you think?

richardebeling: How big is the misreporting we get / what is the difference between a coverage run with typeguard enabled vs disabled, over the whole test suite? If it's just 0.x%, I'd say we can live with it.

We can disable typeguard for the coverage run, that shouldn't have any downsides, right? For the live server tests, I think we could also disable it if we gain anything, but I'm more inclined to just remove them from coverage reporting instead, then. We shouldn't have many code paths that are tested by live server tests but by nothing else (and adding non-liveserver tests for those cases would add value, I think)

evap/student/templates/student_vote.html
```diff
@@ -73,8 +73,8 @@
             </small>
         </div>
     {% endif %}
-    <h3 class="mb-3 d-flex gap-2 justify-content-between align-items-end flex-column flex-md-row">
-        <span>{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</span>
+    <div class="mb-3 d-flex gap-2 justify-content-between align-items-end flex-column flex-md-row">
+        <h3 class="mb-0">{{ evaluation.full_name }} ({{ evaluation.course.semester.name }})</h3>
```
richardebeling, line 77: heh, I was thinking about asking you whether that's legal, but got distracted in some way.

evap/development/middleware.py
```diff
@@ -0,0 +1,76 @@
+from collections.abc import Callable
+
+import requests
+from django.conf import settings
+from django.http.request import HttpRequest
+from django.http.response import HttpResponse, HttpResponseBase
+
+
+class InvalidHtmlError(Exception):
```
richardebeling, line 9: How well does this plug into the unittest framework, if I run the test with `--pdb`, do I end up at a place where I can still inspect the html? (Wondering if we should use `AssertionError` instead)

evap/development/middleware.py
```diff
@@ -0,0 +1,76 @@
...
+            self.validate_html(request.path, response.text)
+        return response
+
+    def validate_html(self, path: str, html: str) -> None:
+        errors = requests.post(
+            self.vnu_url,
+            params={
+                "out": "json",
+                "level": "error",
+                "filterpattern": "|".join(f".*{p}.*" for p in self.IGNORED_ERROR_PATTERNS),
```
richardebeling: This builds the string every time -- is that significant, or does it not matter in the scheme of things?

What is your full end-to-end runtime of `./manage.py test --keepdb --parallel` with and without this change here? I think you mentioned some numbers at some point but I forgot them

richardebeling (approved):

evap/student/templates/student_vote.html
niklasmohrin: straight to jail

evap/development/middleware.py
niklasmohrin: Yes, you end up on the line `raise InvalidHtmlError` (which is tens of calls deep in the middleware, but still). In fact, the only reason I added the `html` attribute is to inspect the full html in this case, it is not used elsewhere.

richardebeling: Ok, works for me

evap/development/middleware.py
niklasmohrin: yeah I think I had some numbers at the beginning, but that was trash because the tests were failing earlier with the validator throwing errors

on this branch, the timings were 13.5, 12.4, and 12.0 seconds, and on main they were 8.8, 9.1, and 9.0

creating this string once in `__init__` gives 11.9, 11.8, and 12.0

evap/development/middleware.py
Kakadus: On main, we lose about 0.5% coverage, almost all of which is branch coverage. Interestingly, 8 more lines are covered through typeguard, 6 of which are random uncovered lines reported as covered. I have nearly enough for a PR at typeguard, so hopefully all of this can be resolved upstream.

richardebeling: Nice, that would be really cool

janno42 (approved):
much validation

PR #2770, opened 2026-08-03
Add copy name feature in textanswer review view

ZoqkMaze:
In addition to #2697


evap/staff/templates/staff_evaluation_textanswers.html
niklasmohrin, line 1: the tooltip is centered on the div, not the title; the span should be in the h3 (or the span should not exist in the first place)

niklasmohrin (commented):
can you give a more detailed title? :D

niklasmohrin (approved):

richardebeling (approved):

PR #2772, opened 2026-08-05
Fix tests in debug mode

Kakadus:
see https://github.com/e-valuation/EvaP/issues/1719#issuecomment-5181194424 following.

If we want to enable the debug mode in CI again is separate.

25b575f266ef3e9913a557ce3a88410878cae2be is unrelated to debug mode. It also triggers if `assert_no_database_modifications` is used after something else the activates a `CaptureQueriesContext`, e.g. the num queries asserter, whose queries would always leak into later calls to `assert_no_database_modifications`.

### DIFF
--- a/evap/evaluation/tests/test_auth.py
+++ b/evap/evaluation/tests/test_auth.py
@@ -240,7 +240,7 @@ class LoginTests(WebTest):
             self.assertIsInstance(response.context["user"], AnonymousUser)
             self.assertNotContains(response, "Logout")
 
-        # correct password while password login is disabled. Disable the debug page
+        # correct password while password login is disabled. Disable the debug internal server error page
         with override_settings(ACTIVATE_OPEN_ID_LOGIN=True, DEBUG=False):
             self.assertFalse(auth.password_login_is_active())
 
--- a/evap/evaluation/tests/test_auth.py
+++ b/evap/evaluation/tests/test_auth.py
@@ -240,8 +240,8 @@ class LoginTests(WebTest):
             self.assertIsInstance(response.context["user"], AnonymousUser)
             self.assertNotContains(response, "Logout")
 
-        # correct password while password login is disabled
-        with override_settings(ACTIVATE_OPEN_ID_LOGIN=True):
+        # correct password while password login is disabled. Disable the debug page
+        with override_settings(ACTIVATE_OPEN_ID_LOGIN=True, DEBUG=False):
             self.assertFalse(auth.password_login_is_active())
 
             password_form["password"] = "evap"  # nosec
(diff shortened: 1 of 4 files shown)


evap/evaluation/tests/tools.py
richardebeling, line 1: I see, that explains it :D

evap/evaluation/tests/test_models.py
```diff
@@ -248,7 +248,9 @@ def test_evaluation_ended(self):
             wait_for_grade_upload_before_publishing=False,
         )
 
-        with patch("evap.evaluation.models.Evaluation.end_evaluation") as mock:
+        with patch(
+            "evap.evaluation.models.Evaluation.end_evaluation", autospec=True, side_effect=Evaluation.end_evaluation
+        ) as mock:
             Evaluation.update_evaluations()
```
richardebeling: How did this break? If we just mocked the method, I would have expected that it is just a noop, but that shouldn't break any of the state-machine field logic, right?

evap/evaluation/tests/test_auth.py
```diff
@@ -240,8 +240,8 @@ def test_passworduser_login(self):
             self.assertIsInstance(response.context["user"], AnonymousUser)
             self.assertNotContains(response, "Logout")
 
-        # correct password while password login is disabled
-        with override_settings(ACTIVATE_OPEN_ID_LOGIN=True):
+        # correct password while password login is disabled. Disable the debug page
+        with override_settings(ACTIVATE_OPEN_ID_LOGIN=True, DEBUG=False):
```
richardebeling: The reason here is that if we run with debug mode enabled, we get an internal server error page instead? Maybe say that in the comment instead

evap/evaluation/tests/test_models.py
richardebeling: Ah, I guess we first call `end_evaluation` and then immediately `end_review`, but since `end_evaluation` was a noop, it breaks?

Do we maybe want to mock `end_review` as well instead (and assert that it was called)?

evap/evaluation/tests/tools.py
```diff
@@ -314,11 +314,11 @@ def assert_no_database_modifications(*args, **kwargs):
     allowed_prefixes = ["select", "savepoint", "release savepoint", "rollback to savepoint"]
 
     conn = connections[DEFAULT_DB_ALIAS]
-    with CaptureQueriesContext(conn):
+    with CaptureQueriesContext(conn) as context:
         try:
             yield
         finally:
-            for query in conn.queries_log:
+            for query in context.captured_queries:
```
niklasmohrin, line 321: note: if I put `[]` here instead, no test fails; should we add one? (outside this PR)

niklasmohrin (approved):
Thanks :)

richardebeling (approved):

niklasmohrin (approved):

PR #2774, opened 2026-08-07
prevent path coverage drop on thurdays

Kakadus:
in memorandum of "tests fail at specific time of day"

No test actually forces a reminder _not_ to be sent, and on a thursday, the reminder is always sent in all tests. So on thursday, there is a missing branch early returning from 
https://github.com/e-valuation/EvaP/blob/0a5f134af470b3264e354e018799d69c016168b7/evap/evaluation/management/commands/send_reminders.py#L73-L74


niklasmohrin (approved):
hahaha, our very own "open office doesn't print on tuesdays"

Good stuff!

Issue #2763, opened 2026-07-20
EvaP light


janno42:
The settings should allow you to choose whether or not you want to see the following things. For each of the items below, we can have a setting like `ENABLE_COURSE_TYPES` that is either `True` or `False`.

Please create separate PRs for each of these things.

1. Course Type
  - A default value "Other" should be set for all courses
  - The management page should not be linked in the navbar
  - The labels on evaluations/courses should not be shown
  - The form field should not be shown
  - The "Assign questionnaires" button on the staff semester page should not be shown
  - The course types on the results index page filters should not be shown
2. Program
  - A default value "Other" should be set for all courses
  - The management page should not be linked in the navbar
  - The labels on evaluations/courses should not be shown
  - The form field should not be shown
  - The programs on the results index page filters should not be shown
3. Exam Type
  - The management page should not be linked in the navbar
  - The default value should always be "None"
  - The form field should not be shown
  - The "Create exam evaluation" dropdown entry for evaluations on the staff semester page should not be shown
4. Grades
  - The app "grades" should be disabled
  - `wait_for_grade_upload_before_publishing` should be False by default
  - The form field should not be shown
  - "Delete grade documents" should not have a button in the "Archiving" card on the staff semester page
  - The "Is midterm evaluation" form field in the evaluation form should not be shown
5. Staff semester overview card
  -  The "Overview" card on the staff semester page and all buttons in its header should not be shown
6. Reward Points
  - The app "rewards" should be disabled
  - Reward Point Redemption Events should not be linked in the navbar
  - The "Is rewarded" form field in the evaluation form should not be shown
  - The rewards overview on the student index page should not be shown
7. Editors/Delegations
  - The "Allow editors to edit" form field should not be shown
  - The "Delegate preparation" button on the contributor index page should not be shown
8. Email Templates
  - Email templates should not be linked in the navbar
  - Email templates should not be shown on the staff index page
  - The "Send email" dropdown entry for evaluations on the staff semester page should not be shown
9. CMS links
  - The app "cms" should be disabled
  - All CMS link icons and badges should not be shown
  - The "merge with other evaluation" dropdown entry for evaluations on the staff semester page should not be shown
10. Approval by managers
  - Clicking on "Save and approve" by editors should set the evaluation's state to `APPROVED` (skipping `EDITOR_APPROVED`)
11. Publishing
  - When the evaluation period ends, the `ReviewDecision` of all text answers should be set to `PUBLIC` and the results should be published immediately

PR #2778, opened 2026-08-13
Configurable image path

janno42:
For the "EvaP light" instance described in #2763, we would also like to be able to change the branding.
This moves images to their own directory, which can be replaced in the settings so we can use files from a different directory if needed.


janno42:
I need assistance with the mypy errors :/

niklasmohrin:
I think we could make the type error go away by changing the setting to be a tuple instead of a list.

However, it seems more sensible to me to delete the view in development (I think it is left from our old way of doing frontend tests), and use a "get static" helper from Django in the staff view

janno42:
What do you mean by "helper from Django"?

niklasmohrin:
I think we should use finders.find from https://docs.djangoproject.com/en/6.1/ref/contrib/staticfiles/#finders-module instead of assuming that that the sample excel file is in exactly the first static directory from the setting

janno42:
We also use `settings.STATICFILES_DIRS[0]` in these files:

- evaluation/management/commands/scss.py
- evaluation/management/commands/ts.py
- evaluation/tests/test_commands.py

Should these also be changed?

niklasmohrin (approved):
I think for the development commands the change is not as important (we don't expect a changed value for `STATICFILES_DIRS`); we should think about it, but we can do that in another PR

PR #2781, opened 2026-09-02
Introduce Rewards Status

janno42:
Earned reward points can now also count as "status points". The number of status points earned determines each student's rewards status.
Detailed information about reward and status points, as well as their respective progress, is shown on the student and rewards index pages.

Best reviewed per commit.

### DIFF
--- a/evap/rewards/forms.py
+++ b/evap/rewards/forms.py
@@ -1,11 +1,13 @@
 from collections.abc import Generator
 from contextlib import contextmanager
-from datetime import date
+from datetime import date, datetime
 from typing import Any, cast
 
 from django import forms
+from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.db import transaction
+from django.db.models import Sum
 from django.utils.translation import gettext as _
 
 from evap.evaluation.models import UserProfile
@@ -84,6 +86,12 @@ class BaseRewardPointRedemptionFormSet(forms.BaseFormSet):
 
         total_points_available = reward_points_of_user(self.user)
         total_points_redeemed = sum(form.cleaned_data["points"] for form in self.forms)
+        points_redeemed_this_year = (
+            RewardPointRedemption.objects.filter(
+                user_profile=self.user, redemption_time__year=datetime.now().year
+            ).aggregate(Sum("value"))["value__sum"]
+            or 0
+        )
         assert all(form.cleaned_data["points"] >= 0 for form in self.forms)
 
         if total_points_redeemed <= 0:
@@ -91,6 +99,13 @@ class BaseRewardPointRedemptionFormSet(forms.BaseFormSet):
 
         if total_points_redeemed > total_points_available:
             raise ValidationError(_("You don't have enough reward points."))
+
+        if points_redeemed_this_year + total_points_redeemed > settings.MAX_REDEEMED_POINTS_PER_YEAR:
+            raise ValidationError(
+                _("You can only redeem up to {points} points per year.").format(
+                    points=settings.MAX_REDEEMED_POINTS_PER_YEAR
+                )
+            )
 
     def save(self) -> list[RewardPointRedemption]:
         assert self.locked
--- a/evap/rewards/templates/rewards_index.html
+++ b/evap/rewards/templates/rewards_index.html
@@ -10,12 +10,57 @@
 
     {% show_infotext "rewards_index" %}
 
-    <div class="card card-outline-primary mb-3">
-        <div class="card-header">
-            {% translate 'Redeem points' %}
+    {% if status_points_progress %}
+        <div class="card mb-3">
+            <div class="card-header d-flex">
+                {% translate 'Your status' %}: {{ status_points_progress.current_status.name }}
+                <div class="ms-auto">
+                    <span class="badge bg-light">
+                        {% translate 'Your status points' %}: {{ status_points_progress.earned_status_points }}
+                    </span>
+                </div>
+            </div>
+            <div class="card-body">
+                <div class="rewards-progress-bar pt-0 px-4">
+                    <span class="progress-container">
+                        <span class="progress m-0 bg-info">
+                            {% with earned_status_points=status_points_progress.earned_status_points|min_value:status_points_progress.max_status_points %}
+                                <span
+                                    class="progress-bar width-percent-{% widthratio earned_status_points status_points_progress.max_status_points 100 %}"
+                                    style="background: {{ status_points_progress.current_status.color2 }}; background: linear-gradient(180deg, {{ status_points_progress.current_status.color2 }} -10%, {{ status_points_progress.current_status.color1 }} 10%, {{ status_points_progress.current_status.color2 }} 70%)"
+                                ></span>
+                            {% endwith %}
+                        </span>
+                    </span>
+                    {% for status in status_points_progress.statuses %}
+                        <div class="position-relative progress-step progress-step-top left-percent-{% widthratio status.points status_points_progress.max_status_points 100 %}">
+                            <div class="progress-step-number">
+                                {{ status.points }}
+                            </div>
+                            <div class="progress-step-label">{{ status.name }}</div>
+                        </div>
+                        {% if not forloop.first %}
+                            <div class="position-relative progress-step progress-step-bottom left-percent-{% widthratio status.points status_points_progress.max_status_points 100 %}">
+                                <div class="progress-step-number"></div>
+                                <div class="progress-step-label">{{ status.reward }}</div>
+                            </div>
+                        {% endif %}
+                    {% endfor %}
+                </div>
+            </div>
+        </div>
+    {% endif %}
+
+    <div class="card card-outline-primary">
+        <div class="card-header d-flex">
+            {% translate 'Redeem reward points' %}
+            <div class="ms-auto">
+                <span class="badge bg-primary">
+                    {% translate 'Available reward points' %}: {{ total_points_available }}
+                </span>
+            </div>
         </div>
         <div class="card-body table-responsive">
-            <p><b>{% translate 'Total points available' %}: {{ total_points_available }}</b></p>
             {% if total_points_available > 0 %}
                 {% if events %}
                     <form id="reward-redemption-form" action="#" method="POST" class="form-horizontal multiselect-form">
@@ -29,7 +74,7 @@
                             <thead>
                                 <tr>
                                     <th class="width-percent-20">{% translate 'Date' %}</th>
-                                    <th class="width-percent-40">{% translate 'Event' %}</th>
+                                    <th class="width-percent-40">{% translate 'Offer' %}</th>
                                     <th class="width-percent-20">{% translate 'Available until' %}</th>
                                     <th class="width-percent-20">{% translate 'Redeem points' %}</th>
                                 </tr>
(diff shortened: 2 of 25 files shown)


evap/static/scss/components/_progress.scss
```diff
@@ -54,48 +70,104 @@
...
+        .progress-step.progress-step-bottom {
+            &:nth-of-type(3n + 1) .progress-step-label::before {display: none;}
+            &:nth-of-type(3n + 2) .progress-step-label::before {height: 18px;}
+            &:nth-of-type(3n) .progress-step-label::before {height: 36px;}
+        }
+        .progress-step.progress-step-top {
+            &:nth-of-type(3n + 1) .progress-step-label::after {display: none;}
+            &:nth-of-type(3n + 2) .progress-step-label::after {height: 18px;}
+            &:nth-of-type(3n) .progress-step-label::after {height: 36px;}
+        }
```
niklasmohrin, line 171: I think this does the same:
```suggestion
        .progress-step:is(.progress-step-top, .progress-step-bottom) {
            &:nth-of-type(3n + 1) .progress-step-label::before {display: none;}
            &:nth-of-type(3n + 2) .progress-step-label::before {height: 18px;}
            &:nth-of-type(3n) .progress-step-label::before {height: 36px;}
        }
```

evap/rewards/forms.py
```diff
@@ -92,6 +100,13 @@ def clean(self) -> None:
         if total_points_redeemed > total_points_available:
             raise ValidationError(_("You don't have enough reward points."))
 
+        if points_redeemed_this_year + total_points_redeemed > settings.MAX_REDEEMED_POINTS_PER_YEAR:
+            raise ValidationError(
+                _("You can only redeem up to {points} points per year.").format(
```
niklasmohrin, line 105: Colloquial use of "only", "redeem only up to" sounds odd, maybe rephrase to something like this? Also fine to keep colloquial if we prefer that
```suggestion
                _("You can't redeem more than {points} points per year.").format(
```

evap/rewards/tests/test_views.py
```diff
@@ -114,6 +114,27 @@ def test_inconsistent_previous_redemption_counts(self):
         form2.submit(status=409)
         self.assertEqual(1, RewardPointRedemption.objects.filter(user_profile=self.student).count())
 
+    @override_settings(MAX_REDEEMED_POINTS_PER_YEAR=4)
+    def test_redeem_more_than_allowed(self):
```
niklasmohrin, line 118: can we add a check or modify the test to ensure that the previously redeemed points are taken into accounts? right now, we are redeeming more than MAX_REDEEMED_POINTS_PER_YEAR at once

evap/settings.py
```diff
@@ -69,10 +69,28 @@
...
+    (50 / 100, 3),
+    (55 / 100, 4),
+    (60 / 100, 5),
+    (65 / 100, 6),
+    (70 / 100, 7),
+    (75 / 100, 8),
+    (80 / 100, 9),
+    (90 / 100, 10),
 ]
+MAX_REDEEMED_POINTS_PER_YEAR = 50
```
niklasmohrin, line 83: Should we add a note like
```suggestion
# Equivalent to 50€ per student
MAX_REDEEMED_POINTS_PER_YEAR = 50
```
Or do we feel that this is too closely tied to our model? (and is it even accurate in the new rewards model?)

evap/rewards/templates/rewards_index.html
```diff
@@ -1,26 +1,66 @@
...
+                </div>
+            </div>
+            <div class="card-body">
+                <div class="rewards-progress-bar pt-0 px-4">
+                    <span class="progress-container">
+                        <span class="progress m-0 bg-info">
+                            {% with earned_status_points=status_points_progress.earned_status_points|min_value:status_points_progress.max_status_points %}
+                                <span
+                                    class="progress-bar width-percent-{% widthratio earned_status_points status_points_progress.max_status_points 100 %}"
+                                    style="background: {{ status_points_progress.current_status.color2 }}; background: linear-gradient(180deg, {{ status_points_progress.current_status.color2 }} -10%, {{ status_points_progress.current_status.color1 }} 10%, {{ status_points_progress.current_status.color2 }} 70%)"
```
niklasmohrin, line 30: Why are there two `background` rules? Is it as a fallback rule? I don't think we have something like this anywhere else

evap/student/templates/student_index.html
```diff
@@ -8,16 +8,20 @@
...
                 {% include 'startpage_button.html' with page='ST' btnClass='btn-sm' %}
             </div>
-        </div>
-    {% endif %}
+        {% endif %}
+    </div>
 
-    {% include 'student_global_evaluation_progress.html' with global_evaluation_progress=global_evaluation_progress %}
+    {% if show_reward_card %}
+        {% include 'student_rewards_card.html' with global_evaluation_progress=global_evaluation_progress %}
```
niklasmohrin, line 23: the included template has two essential parameters, semester_rewards_stats and global_evaluation_progress, why is one of them passed with `with` and the other implicitly? can we make it consistent?

evap/rewards/tools.py
```diff
@@ -83,12 +86,101 @@ def grant_eligible_reward_points_for_semester(request: HttpRequest, semester: Se
...
 
+@dataclass
+class StatusPointsProgress:
+    @dataclass
+    class Status:
+        points: int  # points required for this status
+        name: str
+        reward: str
+        color1: str
+        color2: str
```
niklasmohrin, line 97: What is the meaning of these two? Can we give more descriptive names like "start_color" and "finished_color" or so? (I am not sure what they mean, so my suggested names might make no sense)

Another possible way could be `color_transition: tuple[str, str]` if the colors are meant for only a transition.

Regardless, a comment would be nice here

evap/rewards/tools.py
```diff
@@ -83,12 +86,101 @@ def grant_eligible_reward_points_for_semester(request: HttpRequest, semester: Se
...
+
+        language = get_language()
+
+        max_status_points = max(status[0] for status in settings.STATUS_POINTS)
+        statuses = [
+            StatusPointsProgress.Status(
+                points=points, name=name[language], reward=reward[language], color1=color1, color2=color2
+            )
+            for points, name, reward, color1, color2 in settings.STATUS_POINTS
+        ]
```
niklasmohrin, line 117:
```suggestion
        statuses = [
            StatusPointsProgress.Status(
                points=points, name=name[language], reward=reward[language], color1=color1, color2=color2
            )
            for points, name, reward, color1, color2 in settings.STATUS_POINTS
        ]
        max_status_points = max(status.points for status in statuses)
```

evap/rewards/tools.py
```diff
@@ -83,12 +86,101 @@ def grant_eligible_reward_points_for_semester(request: HttpRequest, semester: Se
...
+            .aggregate(Sum("value"))
+        )
+        status_points_progress.earned_status_points = query["value__sum"] or 0
+
+        status_points_progress.current_status = status_points_progress.statuses[0]
+        for status in status_points_progress.statuses:
+            if status.points <= status_points_progress.earned_status_points:
+                status_points_progress.current_status = status
+            else:
+                break
```
niklasmohrin, line 142: not sure if we like this better, but I couldn't help but replace this loop:
```suggestion
        status_points_progress.current_status = max(
            (
                status
                for status in status_points_progress.statuses
                if status.points <= status_points_progress.earned_status_points
            ),
            key=lambda status: status.points,
        )
```
now it also doesn't rely on ordering of `statuses` and on `statuses` being non-empty

evap/rewards/tools.py
```diff
@@ -83,12 +86,101 @@ def grant_eligible_reward_points_for_semester(request: HttpRequest, semester: Se
...
+        points: int  # points required for this status
+        name: str
+        reward: str
+        color1: str
+        color2: str
+
+    max_status_points: int
+    statuses: list[Status]
+    current_status: Status | None = None
+    earned_status_points: int = 0
```
niklasmohrin, line 102: I think it is clear what kind of points they are from context
```suggestion
    earned_points: int = 0
```
(I found it a bit annoying how often I typed "status" in my `max(...)` code before)

evap/rewards/tools.py
```diff
@@ -83,12 +86,101 @@ def grant_eligible_reward_points_for_semester(request: HttpRequest, semester: Se
...
+                status_points_progress.current_status = status
+            else:
+                break
+
+        return status_points_progress
+
+
+@dataclass
+class UserSemesterRewardsStats:
+    semester: Semester
```
niklasmohrin, line 149: it's a bit strange that in the `UserSemesterSomethingSomething` struct we store the semester but not the user; feels like the data structure that stores these objects should be a `dict[Semester, UserSemesterRewardsStats]` instead of `list[UserSemesterRewardsStats]` (dicts preserve ordering too, so that's no problem)

evap/rewards/tools.py
```diff
@@ -83,12 +86,101 @@ def grant_eligible_reward_points_for_semester(request: HttpRequest, semester: Se
...
+        for semester in Semester.objects.all()
+        if semester.evaluations.filter(participants=user, is_rewarded=True).count() > 0
+    ]
+
+
 # Signal handlers
 
 
 @receiver(Evaluation.evaluation_evaluated)
 def grant_reward_points_after_evaluate(request: HttpRequest, semester: Semester, **_kwargs) -> None:
```
niklasmohrin, line 180: should we rename?
```suggestion
def grant_reward_and_status_points_after_evaluate(request: HttpRequest, semester: Semester, **_kwargs) -> None:
```

evap/student/templates/student_rewards_card.html
```diff
@@ -0,0 +1,104 @@
...
+                            </span>
+                        </span>
+                    </span>
+                    {% for reward_points_step in reward_points_steps %}
+                        <div class="position-relative progress-step progress-step-bottom left-percent-{% widthratio reward_points_step.0 1 100 %}">
+                            <div class="progress-step-number">
+                                {{ reward_points_step.0|percentage:1 }}
+                            </div>
+                            {% if reward_points_step.1 %}
+                                <div class="progress-step-label">{{ reward_points_step.1 }}</div>
```
niklasmohrin, line 53: I think this is a bit hard to parse in the UI right now, there is percentages, these I understand because it says "1 of 5 evaluations submitted". But then there are just unlabeled numbers; I get now that those are the points that I get, but it doesn't say "point" anywhere in the rewards card

I don't have a good idea on how to fix it right now

evap/rewards/templates/rewards_index.html
```diff
@@ -1,26 +1,66 @@
...
+{% load infotext_templatetags %}
 
 {% block title %}{% translate 'Rewards' %} - {{ block.super }}{% endblock %}
 
 {% block content %}
     {{ block.super }}
 
-    <div class="callout callout-info small">
-        {% blocktranslate trimmed %}
-            You will get reward points by taking part in evaluations. These points can be redeemed for different events.
```
niklasmohrin, line 12: The current info text in the test data doesn't explain how status points work. As far as I understand, I always get status points and reward points paired together, but I keep my status points when redeeming reward points. I think in our explanation, we should explain status points first, and then say something like "in addition, for every status point you earn, you also get a reward point. you can use reward points to ..."

niklasmohrin (commented):
first 5 commits look good; I checked the last commit only roughly for now

if we want to get the first 5 commits in through a separate PR already, that would be fine by me

